PR #5498, opened 2023-11-17
[OSCI][Fix] ScopedHistory out of navigation scope

### LINKED ISSUE #5476: [BUG] ScopedHistory instance has fell out of scope for basePath: `/app/data-explorer`
**Describe the bug**

When we switch the query language in discover page, the console outputs the below error. On top of that, the action of navigating from discover page to another page will output a series of the following error as well. 

It seems like the scope history in data explorer needs to be better handled. 

```
scoped_history.ts:252 Uncaught (in promise) Error: ScopedHistory instance has fell out of navigation scope for basePath: /app/data-explorer
    at ScopedHistory.verifyActive (scoped_history.ts:252:1)
    at get location [as location] (scoped_history.ts:207:1)
    at getCurrentUrl (parse.ts:37:1)
    at getPendingUrl (osd_url_storage.ts:146:1)
    at flush (osd_url_storage.ts:135:1)
    at osd_url_storage.ts:163:1
```

**To Reproduce**
1. Go to discover page
2. Click on the toggle on query language selector on the top navigation bar. 
3. See error

thanhinhchtom:
### Description

Catch the error due to race conditions when the window reloads before the history comes. Name of error is ScopedHistory fell out of navigation scope.

### Issues Resolved

Fix #5476 

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/88635563/2e87ce91-7873-45fc-91bd-945c32680509

image

### Check List

- [] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 - Fix navigation issue across dashboards ([#5435](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5435))
 - [Discover] Fix table panel auto-sizing ([#5441](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5441))
+- [Fix] ScopedHistory out of navigation scope ([#5498](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5498))
 
 ### 🚞 Infrastructure
 
--- a/src/plugins/opensearch_dashboards_utils/public/state_management/url/osd_url_storage.ts
+++ b/src/plugins/opensearch_dashboards_utils/public/state_management/url/osd_url_storage.ts
@@ -213,7 +213,12 @@ export const createOsdUrlControls = (
   return {
     listen: (cb: () => void) =>
       history.listen(() => {
-        cb();
+        try {
+          cb();
+        } catch {
+          // TODO: Do something when ScopedHistory
+          // instance has fell out of navigation scope
+        }
       }),
     update: (newUrl: string, replace = false) => updateUrl(newUrl, replace),
     updateAsync: (updater: UrlUpdaterFnType, replace = false) => {


abbyhu2000:
Hi @thanhinhchtom, thanks for the PR! However, i think this issue needs a little more diving deep on it.

For example, if we type something in the query bar on the data explorer page, the `scoped history fell out of scope` error will pop out. We want to find out why this error appear since typing a query should not cause any console error; adding a try catch here only handles the error but does not really solve the bug. 

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/43937633/3074c7ef-570a-4c34-a900-52ab44764a62

Do you want to dive a little deeper on this bug and document your findings? Thank you!

thanhinhchtom:
> Hi @thanhinhchtom, thanks for the PR! However, i think this issue needs a little more diving deep on it.
> 
> For example, if we type something in the query bar on the data explorer page, the `scoped history fell out of scope` error will pop out. We want to find out why this error appear since typing a query should not cause any console error; adding a try catch here only handles the error but does not really solve the bug.
> 
>  Screen.Recording.2023-12-01.at.11.33.57.AM.mov 
> Do you want to dive a little deeper on this bug and document your findings? Thank you!

thanhinhchtom:
> Hi @thanhinhchtom, thanks for the PR! However, i think this issue needs a little more diving deep on it.
> 
> For example, if we type something in the query bar on the data explorer page, the `scoped history fell out of scope` error will pop out. We want to find out why this error appear since typing a query should not cause any console error; adding a try catch here only handles the error but does not really solve the bug.
> 
>  Screen.Recording.2023-12-01.at.11.33.57.AM.mov 
> Do you want to dive a little deeper on this bug and document your findings? Thank you!

I already looked up a little bit and I think this error coming from the code from Elastic Search. From what I understand, it is caused because the window reloads faster than the scoped history added, so it results in an error. And because of that, when I did the try catch one, I already tried to make some changes to lots of other places, to make sure that the history will work fine. However, I failed to do so and I think the easiest way is just to add try catch statement here to handle the error with history.

abbyhu2000:
@thanhinhchtom Hi, I do not think this is coming from the elastic search code since this error only occurs after we replace from the legacy discover to the current data explorer. Before the change, the scoped history for discover pages works fine and will not output errors like this to the console. I think the error prob originated somewhere in the code of de-angular discover to data explorer. Thanks for taking your time on this issue, and please comment more of your findings if you are still interested in digging deeper. Meanwhile I will convert this PR to draft for now.

PR #5505, opened 2023-11-17
[feat/2.11/os-assistant] Make new homepage configurable

BSFishy:
iteration to make the homepage configurable

### DIFF
--- a/src/plugins/home/config.ts
+++ b/src/plugins/home/config.ts
@@ -33,6 +33,7 @@ import { schema, TypeOf } from '@osd/config-schema';
 export const configSchema = schema.object({
   disableWelcomeScreen: schema.boolean({ defaultValue: false }),
   disableNewThemeModal: schema.boolean({ defaultValue: false }),
+  newHomepage: schema.boolean({ defaultValue: false }),
   prompts: schema.arrayOf(
     schema.object({
       prompt: schema.string(),
--- a/src/plugins/home/public/application/components/home_app.js
+++ b/src/plugins/home/public/application/components/home_app.js
@@ -35,7 +35,7 @@ import { Home } from './home';
 import { FeatureDirectory } from './feature_directory';
 import { TutorialDirectory } from './tutorial_directory';
 import { Tutorial } from './tutorial/tutorial';
-import { OllyHomepage, PlaygroundHomepage } from './homepage';
+import { Homepage } from './homepage';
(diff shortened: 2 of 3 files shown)


src/plugins/home/config.ts
```diff
@@ -33,6 +33,7 @@ import { schema, TypeOf } from '@osd/config-schema';
 export const configSchema = schema.object({
   disableWelcomeScreen: schema.boolean({ defaultValue: false }),
   disableNewThemeModal: schema.boolean({ defaultValue: false }),
+  newHomepage: schema.boolean({ defaultValue: false }),
```
kavilla, line 36: nit: can maybe mimic the other configurations like `disableNewHomePage`

I wonder since since the default prompts are verify specific, i wonder if we should just make default `true`.

kavilla (approved):

src/plugins/home/public/application/components/home_app.js
```diff
@@ -57,9 +57,11 @@ export function HomeApp({ directories, solutions }) {
     addBasePath,
     environmentService,
     telemetry,
+    homeConfig,
   } = getServices();
   const environment = environmentService.getEnvironment();
   const isCloudEnabled = environment.cloud;
+  const isOlly = homeConfig.newHomepage;
```
lezzago, line 64: Lets remove Olly specific references in the OSD code and make it more generic since that is creating a psuedo dependency on Olly from OSD core which is bad.

This can be renamed as `isNewHomePageEnabled`.

src/plugins/home/public/application/components/home_app.js
```diff
@@ -110,6 +106,9 @@ export function HomeApp({ directories, solutions }) {
               telemetry={telemetry}
             />
           </Route>
+          <Route exact path="/">
+            <Homepage olly={isOlly} />
```
lezzago, line 110: ideally `olly` reference would be removed here, but that can be a followup since that was not caused by this PR.

src/plugins/home/config.ts
lezzago: +1, lets follow current scheme of the config naming here.

src/plugins/home/public/application/components/home_app.js
kavilla: +1. ive started created a backlog of generalizing this component

lezzago (approved):

Issue #5440, opened 2023-11-07
[BUG][Discover] Discover table panel won't adjust the size automatically when time range changes


ananzh:
**Describe the bug**
When change time range, discover table panel won't adjust the size automatically. Only refresh will work.

<img width="1728" alt="Screenshot 2023-11-07 at 09 00 27" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/79961084/2a588ef8-8e17-42c2-923f-8f0d53c41c9e">

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/79961084/ef90f833-555e-4feb-871a-e52b58ece833

**OpenSearch Version**
2.10, 2.11 and main

**Dashboards Version**
2.10, 2.11 and main

**Plugins**

ananzh:
Bug is fixed in 2.12. Close this issue.

Issue #4851, opened 2023-08-30
[Data Explorer][Discover 2.0] Break down data fetch logic to multiple data streams or use Expression for histogram


ananzh:
Currently in discover 2.0, we are re-fetch data when histogram interval is updated. Histogram interval only updates aggregation which won't affect total data fetch. It is a bit of waste to send the fetch query to OS. Solution could be breaking down the current data stream to separate streams for histogram and table or use Expression for histogram.

PR #5514, opened 2023-11-20
[BUG] Make table size adjustable according to table rows and columns

ananzh:
## Description
Auto-resized issue is caused by we have a central discover context created useSearch.
When it is updated, both panel and canvas which use useDiscoverContext will render.
The update on panel will trigger useSearch and cause canvas render before update is done.
Therefore when table might not be fully re-rendered and just use the current table space. 
It appears that the main challenge is ensuring that DiscoverTable fully re-renders when the rows changes. 
    
* This PR uses the length of the rows and a combination of last column that uniquely identify the state of the rows as the key.
* The last column name is helpful to make sure add/remove column will trigger fully re-render. For example, since `_source` takes multiple lines so we will see extra empty space at the bottom when we add a column.  Since add/remove column always update last one, by add last column name as a key, we won't see extra space by fully re-render.

We could do more optimization after 2.12. 
* investigate whether we have to call fetch in use_search when add/remove a column
* investigate whether we could not reload panel with canvas
* breaking down histogram and table fetch: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4851

## Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5440

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/79961084/263913f9-625b-4707-ab16-761e1642fe51

### Check List

- [ ] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [Discover] Fix missing index pattern field from breaking Discover [#5626](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5626)
 - [BUG] Remove duplicate sample data as id 90943e30-9a47-11e8-b64d-95841ca0b247 ([5668](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5668))
 - [BUG][Multiple Datasource] Fix datasource testing connection unexpectedly passed with wrong endpoint [#5663](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5663)
+- [BUG][Discover] Make table size adjustable according to table rows and columns ([#5514](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5514))
 
 ### 🚞 Infrastructure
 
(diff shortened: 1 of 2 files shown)


ananzh:
There are two functional tests broken by reload. If we think this could be a tmp fix, I could fix the tests as well.

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [Discover] Fix missing index pattern field from breaking Discover [#5626](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5626)
 - [BUG] Remove duplicate sample data as id 90943e30-9a47-11e8-b64d-95841ca0b247 ([5668](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5668))
 - [BUG][Multiple Datasource] Fix datasource testing connection unexpectedly passed with wrong endpoint [#5663](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5663)
+- [BUG][Discover] Make table size adjustable according to table rows and columns ([#5514](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5514))
```
AMoo-Miki, line 56:
```suggestion
- [Discover] Make table size adjustable according to table rows and columns ([#5514](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5514))
```

AMoo-Miki (changes requested):

ananzh:
close this since the issue is fixed

Issue #5527, opened 2023-11-21
[BUG][Disocver] Timefield of an indexPattern would get added to selected field in the side nav when removing a column from the canvas


MadaniKK:
**Describe the bug**
Everytime we remove a column from the canvas, the timefield of the indexpattern get added to the selected field in the side nav.

**To Reproduce**
1. Go to Discover, selected some columns
2. Click on 'remove column'
3. Look at the selected fields in left side bar
4. See the timestamp/order date column got added to the selected field
![image](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/71857976/f4755f8d-b298-47f8-8bfe-2ce82b5bd1f5)
![image](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/71857976/c9569ea2-5dc8-4826-98fb-c7cf1cb5333b)
![image](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/71857976/fe28c55a-8de6-4f43-aaaf-86ca8529b8df)

**Expected behavior**
The time field shouldn't get added to the selected field automatically.

**OpenSearch Version**
Please list the version of OpenSearch being used.
3.0.0

**Dashboards Version**
Please list the version of OpenSearch Dashboards being used.

**Plugins**

Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

ananzh:
@MadaniKK Thanks for opening. I have assigned it to you.

MadaniKK:
I suggest keeping the time column in the side nav bar because it exists in the legacy cover, and when the hide time column is off, we can still add the time column to the selected field and use drag-drop to arrange column order.

ananzh:
Close it as it is resolve in #5538

PR #5537, opened 2023-11-27
[OSCI][Fix][Discover]Prevent Adding Timefield to the Side Nav Selected Fields on Column Removal

MadaniKK:
### Description
Issue #5527: Now the time field of an index pattern would not get wrongly added to the selected field when we click `remove column` on canvas.

Issue #5538:  When doc_table:hideTimeColumn is enabled, removing the last columns from the canvas would automatically add "_source" to the selected fields in the left nav bar.

### Issues Resolved
issue #5527 #5538 

## Screenshot
For #5527
image
image

For #5538 
 
image

image

### Check List

- [x ] All tests pass
  - [x ] `yarn test:jest`
  - [ x] `yarn test:jest_integration`
- [ x] Commits are signed per the DCO using --signoff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,11 +31,11 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fix `maps.proxyOpenSearchMapsServiceInMaps` config definition so it can be set ([#5170](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5170))
 - [Discover] Fix inactive state on 'Discover' tab in side navigation menu ([#5432](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5432))
 - [BUG] Add platform "darwin-arm64" to unit test ([#5290](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5290))
-- [BUG][Dev Tool] Add dev tool documentation link to dev tool's help menu [#5166](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5166)
+- [BUG][dev tool] Add dev tool documentation link to dev tool's help menu [#5166](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5166)
 - Fix missing border for header navigation control on right ([#5450](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5450))
 - [BUG] Fix filtering issue in data source selector ([5484](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5484))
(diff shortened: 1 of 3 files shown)


joshuarrrr:
@MadaniKK Can you add a changelog entry?

MadaniKK:
> @MadaniKK Can you add a changelog entry?

Hi, Josh, I found another bug (issue #5538) when I wrote this PR, and I found out I could fix it within the same PR. should I make a new PR or fix these two issues in the same PR? ( I just needed to reorder two lines of codes that I wrote in this PR and it will fix issue #5538  too)

ananzh:
@MadaniKK love these fixes and thanks a lot. I will re-run the failed cypress tests and update you.

ananzh:
@MadaniKK, for https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5538, your understanding seems not quite right.

This advanced setting `doc_table:hideTimeColumn` is to allow Discover and Saved Search in Dashboard to not display `timestamp` column (first column) when enabled for a time based index pattern . Without this setting, for time based index pattern, display `timestamp` column is default and user can't remove it.

For the legacy Discover, we could have a better understanding

* after enabling `doc_table:hideTimeColumn`
Screenshot 2023-12-01 at 4 53 02 PM

The buildColumns fun itself will make sure when no columns `return ['_source']`:
```
export function buildColumns(columns: string[]) {
  if (columns.length > 1 && columns.indexOf('_source') !== -1) {
    return columns.filter((col) => col !== '_source');
  } else if (columns.length !== 0) {
    return columns;
  }
  return ['_source'];
}
```

Therefore, the problem is why enabling `doc_table:hideTimeColumn` will cause https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5538 and how could fix it. I think we haven't implement the logic for this advanced settings.

ananzh:
@MadaniKK I would suggest targeting https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5527 in this PR and re-check if all logics are needed if just fixing this specific issue. 

Then start another PR to make sure we have time field removed when enabling `doc_table:hideTimeColumn`. What do you think?

ananzh (changes requested):
Would like to research on `doc_table:hideTimeColumn` and find out the root cause of the break. Thanks a lot for carefully checking these settings.

ananzh (dismissed):
I think this is an acceptable change to fix all the bugs together. @MadaniKK could you select and add more explains here from public slack? Also there is a conflict on CHANGELOG. Thanks.

ananzh (approved):

ashwin-pc (approved):

PR #5546, opened 2023-11-28
[OSCI] Fix type errors in Multiple Data Sources

CMDWillYang:
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_source_management/public/components/create_data_source_wizard/components/create_form/create_data_source_form.tsx
```diff
@@ -162,7 +162,7 @@ export class CreateDataSourceForm extends React.Component<
   };
 
   validateUsername = () => {
-    const isValid = !!this.state.auth.credentials.username?.trim().length;
+    const isValid = !!this.state.auth.credentials.username?.toString().trim().length;
```
ruanyl, line 165: I guess we expect `username` to be a non-empty string? I'd do this validation instead:

```typescript
const isValid = typeof this.state.auth.credentials.username === 'string' && this.state.auth.credentials.username.trim().length > 0;
```

src/plugins/data_source_management/public/components/create_data_source_wizard/components/create_form/create_data_source_form.tsx
```diff
@@ -206,7 +206,7 @@ export class CreateDataSourceForm extends React.Component<
   };
 
   validateRegion = () => {
-    const isValid = !!this.state.auth.credentials.region?.trim().length;
+    const isValid = !!this.state.auth.credentials.region?.toString().trim().length;
```
ruanyl, line 209: The same here, if we expect `region` to be a non-empty string, I'd do this validation:

```typescript
const isValid = typeof this.state.auth.credentials.region === 'string' && this.state.auth.credentials.region.trim().length > 0;

```

src/plugins/data_source_management/public/components/create_data_source_wizard/components/create_form/create_data_source_form.tsx
```diff
@@ -365,7 +365,7 @@ export class CreateDataSourceForm extends React.Component<
                   }
                 )}
                 isInvalid={!!this.state.formErrorsByField.createCredential.username.length}
-                value={this.state.auth.credentials.username || ''}
+                value={this.state.auth.credentials.username?.toString() || ''}
```
ruanyl, line 368: If we expect the text field input to be a string, maybe better to do this?
```typescript
value={typeof this.state.auth.credentials.username === 'string' ? this.state.auth.credentials.username : ''}
```
The same for `password` `region` `accessKey` and `secretKey`

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
```diff
@@ -204,7 +204,7 @@ export class EditDataSourceForm extends React.Component<EditDataSourceProps, Edi
   };
 
   validateUsername = () => {
-    const isValid = !!this.state.auth.credentials.username?.trim().length;
+    const isValid = !!this.state.auth.credentials.username?.toString().trim().length;
```
ruanyl, line 207: The same here:
```typescript
    const isValid =
      typeof this.state.auth.credentials.username === 'string' &&
      this.state.auth.credentials.username.trim().length > 0;
```

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
```diff
@@ -263,7 +263,7 @@ export class EditDataSourceForm extends React.Component<EditDataSourceProps, Edi
   };
 
   validateRegion = () => {
-    const isValid = !!this.state.auth.credentials.region?.trim().length;
+    const isValid = !!this.state.auth.credentials.region?.toString().trim().length;
```
ruanyl, line 266: Same:
```typescript
    const isValid =
      typeof this.state.auth.credentials.region === 'string' &&
      this.state.auth.credentials.region.trim().length > 0;
```

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
```diff
@@ -881,8 +884,8 @@ export class EditDataSourceForm extends React.Component<EditDataSourceProps, Edi
             type={'dual'}
             value={
               this.props.existingDataSource.auth.type === AuthType.SigV4
-                ? this.maskedPassword
-                : this.state.auth.credentials?.secretKey
+                ? this.maskedPassword.toString()
```
ruanyl, line 887: `this.maskedPassword` is already `string` type

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
```diff
@@ -853,8 +856,8 @@ export class EditDataSourceForm extends React.Component<EditDataSourceProps, Edi
             type={'dual'}
             value={
               this.props.existingDataSource.auth.type === AuthType.SigV4
-                ? this.maskedPassword
-                : this.state.auth.credentials?.accessKey
+                ? this.maskedPassword.toString()
```
ruanyl, line 859: `this.maskedPassword` is already `string` type

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
```diff
@@ -944,8 +947,8 @@ export class EditDataSourceForm extends React.Component<EditDataSourceProps, Edi
                 type={'dual'}
                 value={
                   this.props.existingDataSource.auth.type === AuthType.UsernamePasswordType
-                    ? this.maskedPassword
-                    : this.state.auth.credentials?.password
+                    ? this.maskedPassword.toString()
```
ruanyl, line 950: `this.maskedPassword` is already `string` type

CMDWillYang:
Many thanks for the detailed suggestions @ruanyl. I haven't worked with typescript much before and just tried everything to get rid of the type errors, which led to some wrong assumptions/code. Thanks for pointing them out and showing me why that's the case and what a better approach might be. I've gone through and made changes based on your feedback. Please let me know what you think :)

ananzh:
might need to check if we need to backport to 1.3

ananzh:
@CMDWillYang  I 
Correcting TypeScript type errors often falls under code quality improvement and refactoring. I have updated the CHANGELOG to make this PR to refactor section.

src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
```diff
@@ -448,9 +448,7 @@ export class DataSourceSavedObjectsClientWrapper {
     auth: T,
     encryptionContext: EncryptionContext
   ) {
-    const {
-      credentials: { username, password },
-    } = auth;
+    const { username, password } = auth as UsernamePasswordTypedContent;
```
AMoo-Miki, line 451: I am puzzled by this. Does this mean `username` and `password` were props of `auth` and not props of `credentials` which was a prop of `auth`? If so, how was this working until now?

src/plugins/data_source/server/legacy/configure_legacy_client.ts
```diff
@@ -165,10 +165,10 @@ const callAPI = async (
       return request.then(resolve, reject);
     });
   } catch (err) {
-    if (!options.wrap401Errors || err.statusCode !== 401) {
+    if (!options.wrap401Errors || (err as any).statusCode !== 401) {
```
SuZhou-Joe, line 168:
```suggestion
    if (!options.wrap401Errors || (err as unknown).statusCode !== 401) {
```

`unknown` will be preferred

src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
```diff
@@ -462,9 +460,7 @@ export class DataSourceSavedObjectsClientWrapper {
   }
 
   private async encryptSigV4Credential<T = unknown>(auth: T, encryptionContext: EncryptionContext) {
-    const {
-      credentials: { accessKey, secretKey, region, service },
-    } = auth;
+    const { accessKey, secretKey, region, service } = auth as SigV4Content;
```
SuZhou-Joe, line 463: Same question as Miki that we should destruct auth.credentials to get all the credentials related fields.

CMDWillYang: Many thanks for pointing this out. would the following work?

`    const {
      credentials: { accessKey, secretKey, region, service },
    } = auth as { credentials: SigV4Content };`

src/plugins/data_source/server/legacy/configure_legacy_client.ts
ruanyl: `err` is `any` with `"useUnknownInCatchVariables": false`, casting to unknown like this `(err as unknown).statusCode` actually results in TS error.

src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
CMDWillYang: Many thanks for pointing this out. It was my mistake and got confused about the structure. Would this work?

`    const {
      credentials: { username, password },
    } = auth as { credentials: UsernamePasswordTypedContent };
`

src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
CMDWillYang: Many thanks for pointing this out. It was my mistake and got confused about the structure. Would this work?

`    const {
      credentials: { accessKey, secretKey, region, service },
    } = auth as { credentials: SigV4Content };
`

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
```diff
@@ -566,8 +573,8 @@ export class EditDataSourceForm extends React.Component<EditDataSourceProps, Edi
 
         {this.state.showUpdateAwsCredentialModal ? (
           <UpdateAwsCredentialModal
-            region={this.state.auth.credentials!.region}
-            service={this.state.auth.credentials!.service}
+            region={this.state.auth.credentials!.region?.toString() || ''}
+            service={this.state.auth.credentials!.service as SigV4ServiceName}
```
ruanyl, line 577: These non-null assertion seems unnecessary
```suggestion
            region={this.state.auth.credentials.region?.toString() || ''}
            service={this.state.auth.credentials.service as SigV4ServiceName}
```

ruanyl (commented):
Thank you for your patient to continuously improve this PR, I left a few more comment 👍

src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
CMDWillYang: Sorry, but I'm still a bit confused. Even after changing the signature, there's still the below error unless I cast credentials to UsernamePasswordTypedContent.

Screenshot 2023-12-18 at 10 27 45 PM

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
```diff
@@ -69,7 +69,7 @@ export interface EditDataSourceState {
 
 export class EditDataSourceForm extends React.Component<EditDataSourceProps, EditDataSourceState> {
   static contextType = contextType;
-  public readonly context!: DataSourceManagementContextValue;
+  declare context: React.ContextType<typeof contextType>;
```
CMDWillYang, line 72: Many thanks @ruanyl  for giving me such great suggestions and feedback! Another problem I'm getting is test failing with the below error. I'm unsure what that means but from some research, it seems like I should make some changes to the config file. For context, I added declare to get rid of the type error. If this is the right approach, would changing the config file be the right approach?

potential fix: https://stackoverflow.com/questions/70690029/typescript-declare-fields-must-first-be-transformed-by-babel-plugin-transform 

failed test: https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/7256678119/job/19769311750?pr=5546 

image

src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
ruanyl: I checked your latest code, I'm fine with your current solution, the check of `AuthType` should make sure `encryptBasicAuthCredential` to only receive `UsernamePasswordTypedContent` type of credential

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
ruanyl: I did some investigation, but couldn't find a straight-forward solution, I might leave it as it was for now.

src/plugins/data_source/server/legacy/configure_legacy_client.ts
```diff
@@ -194,14 +194,14 @@ const getBasicAuthClient = async (
...
-  const { accessKey, secretKey, region, service } = credential;
+  const { accessKey, secretKey, region } = credential;
   const client = new LegacyClient({
     connectionClass: HttpAmazonESConnector,
     awsConfig: new Config({
       region,
       credentials: new Credentials({ accessKeyId: accessKey, secretAccessKey: secretKey }),
     }),
-    service,
+
```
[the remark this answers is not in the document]
CMDWillYang: I believe it is not needed. The only usage of `getAWSClient` in the repo is line 120. const `awsClient` is used in `addClientToPool` and `callAPI.bind(null, awsClient)` which `getQueryClient` then returns. None of them should use service to my knowledge/investigation, and also type `legacyClient` / class `Client` does not have this service property so that should be a good indication.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/bca1dcfcc783ccf2837c26f41bd6c5305aa9f869/src/plugins/data_source/server/legacy/configure_legacy_client.ts#L120

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
```diff
@@ -118,16 +118,16 @@ export class EditDataSourceForm extends React.Component<EditDataSourceProps, Edi
       this.setState({
         title,
         description: description || '',
-        endpoint,
+        endpoint: endpoint || '', // Fix: Assign a default value of an empty string if endpoint is undefined
```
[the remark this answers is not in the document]
CMDWillYang: changed, thanks!

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
```diff
@@ -118,16 +118,16 @@ export class EditDataSourceForm extends React.Component<EditDataSourceProps, Edi
...
           type: auth.type,
           credentials: {
             username: authTypeCheckResults.isUserNamePassword ? auth.credentials?.username : '',
             password: authTypeCheckResults.isUserNamePassword ? this.maskedPassword : '',
             service: authTypeCheckResults.isSigV4
-              ? auth.credentials?.service || SigV4ServiceName.OpenSearch
-              : '',
-            region: authTypeCheckResults.isSigV4 ? auth.credentials!.region : '',
+              ? (auth.credentials?.service as SigV4ServiceName) || SigV4ServiceName.OpenSearch
+              : SigV4ServiceName.OpenSearch,
```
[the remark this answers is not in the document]
CMDWillYang: changed, thanks!

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
```diff
@@ -380,7 +384,10 @@ export class EditDataSourceForm extends React.Component<EditDataSourceProps, Edi
     this.setState({ isLoading: true });
     const isNewCredential = !!(this.state.auth.type !== this.props.existingDataSource.auth.type);
 
-    let credentials = this.state.auth.credentials;
+    let credentials = this.state.auth.credentials as
+      | UsernamePasswordTypedContent
+      | SigV4Content
+      | undefined;
```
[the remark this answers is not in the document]
CMDWillYang: changed, thanks!

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
ruanyl: So instead of changing the config file, I might just not trying to fix the type error for now.

CMDWillYang: Got it, I already reverted this change in a commit about an hour ago. 
Thanks!

src/plugins/data_source/server/legacy/configure_legacy_client.ts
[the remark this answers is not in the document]
ruanyl: Thanks for the clarification!

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
```diff
@@ -118,16 +118,16 @@ export class EditDataSourceForm extends React.Component<EditDataSourceProps, Edi
...
           credentials: {
             username: authTypeCheckResults.isUserNamePassword ? auth.credentials?.username : '',
             password: authTypeCheckResults.isUserNamePassword ? this.maskedPassword : '',
             service: authTypeCheckResults.isSigV4
-              ? auth.credentials?.service || SigV4ServiceName.OpenSearch
-              : '',
-            region: authTypeCheckResults.isSigV4 ? auth.credentials!.region : '',
+              ? (auth.credentials?.service as SigV4ServiceName) || SigV4ServiceName.OpenSearch
+              : ('' as SigV4ServiceName),
+            region: authTypeCheckResults.isSigV4 ? String(auth.credentials?.region) || '' : '',
```
ruanyl, line 130: This expression `String(auth.credentials?.region) || ''` evaluates to `"undefined"` string when `auth.credentials?.region = undefined` which may not quite as you expected. I guess what you intend to do was:
```typescript
String(auth.credentials?.region || '')
```

src/plugins/data_source/server/legacy/configure_legacy_client.ts
```diff
@@ -165,10 +165,10 @@ const callAPI = async (
       return request.then(resolve, reject);
     });
   } catch (err) {
-    if (!options.wrap401Errors || err.statusCode !== 401) {
+    if (!options.wrap401Errors || (err as unknown).statusCode !== 401) {
```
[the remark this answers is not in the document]
SuZhou-Joe: Oh, thanks @ruanyl for pointing it out. Actually `any` should not be explicitly declared as there is a lint to check any `any` used across OSD. So as what @ruanyl suggested, we can remove the as unknown as err is already an `any` type.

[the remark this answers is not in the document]
CMDWillYang: thanks for the info. changed!

src/plugins/data_source/server/legacy/configure_legacy_client.ts
```diff
@@ -165,10 +165,10 @@ const callAPI = async (
...
       throw err;
     }
-    throw LegacyOpenSearchErrorHelpers.decorateNotAuthorizedError(err);
+    throw LegacyOpenSearchErrorHelpers.decorateNotAuthorizedError(err as Error);
```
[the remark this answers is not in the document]
CMDWillYang: fixed, thanks!

src/plugins/data_source/server/routes/test_connection.ts
```diff
@@ -78,7 +78,7 @@ export const registerTestConnectionRoute = (
             success: true,
           },
         });
-      } catch (err) {
+      } catch (err: Error | any) {
```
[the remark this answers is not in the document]
CMDWillYang: fixed, thanks!

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
CMDWillYang: fixed, thanks!

src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx
```diff
@@ -118,16 +118,16 @@ export class EditDataSourceForm extends React.Component<EditDataSourceProps, Edi
...
           type: auth.type,
           credentials: {
             username: authTypeCheckResults.isUserNamePassword ? auth.credentials?.username : '',
             password: authTypeCheckResults.isUserNamePassword ? this.maskedPassword : '',
             service: authTypeCheckResults.isSigV4
-              ? auth.credentials?.service || SigV4ServiceName.OpenSearch
-              : '',
-            region: authTypeCheckResults.isSigV4 ? auth.credentials!.region : '',
+              ? (auth.credentials?.service as SigV4ServiceName) || SigV4ServiceName.OpenSearch
+              : ('' as SigV4ServiceName),
```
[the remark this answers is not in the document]
CMDWillYang: fixed, I didn't do it because I wasn't sure if there was some unexpected behavior since the original code has `''` instead of `undefined`.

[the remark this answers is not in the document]
ruanyl: @CMDWillYang I think that's valid concern, I checked the usage of `this.state.auth.credentials?.service`, it looks logically fine. And I also tested locally with editing an existing data source connection, it works fine.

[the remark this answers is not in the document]
CMDWillYang: got it, thanks!

src/plugins/data_source/server/legacy/configure_legacy_client.ts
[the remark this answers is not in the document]
CMDWillYang: I did notice, however, that there is a test expecting `service` to still be included. 
I have yet to figure out why this might be needed. still researching.

https://github.com/CMDWillYang/OpenSearch-Dashboards/blob/1c0f44422994a9e5da13a4807dc5812a35b20026/src/plugins/data_source/server/legacy/configure_legacy_client.test.ts#L167-L179

ashwin-pc:
Closing this since this is quite old and a lot has changed since. Thanks for the PR @CMDWillYang

PR #5549, opened 2023-11-29
[TEST] Change to use electron to run cypress

### LINKED ISSUE #5495: [BUG][Data Explorer] Table rendering flakiness
**Describe the bug**
The discover table rendering seems to be flaky, and it causes some tests fail like below. The tests depend on 'docTable' (the discover table) to render correctly, however, it seems like the table fails to render so element `[data-test-subj="docTable"]` is never found.

The tests all pass correctly locally, however, it is flaky on the CI.

```
1) discover app
       nested query
         "before all" hook for "should support querying on nested fields":
     AssertionError: Timed out retrying after 60000ms: Expected to find element: `[data-test-subj="docTable"]`, but never found it.

Because this error occurred during a `before all` hook we are skipping the remaining tests in the current suite: `nested query`
      at Context.eval (http://localhost:5601/__cypress/tests?p=cypress/support/index.js:461:13)

  2) discover app
       data-shared-item
         should have correct data-shared-item title and description:
     AssertionError: Timed out retrying after 60000ms: Expected to find element: `[data-test-subj="docTable"]`, but never found it.
      at Context.eval (http://localhost:5601/__cypress/tests?p=cypress/support/index.js:[461](https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/6855751194/job/18764087133#step:17:462):13)
```

abbyhu2000:
### Description
Change from using chromium to electron to see if it helps with the flaky data explorer cypress tests.

### Issues Resolved
#5495

### DIFF
--- a/.github/workflows/cypress_workflow.yml
+++ b/.github/workflows/cypress_workflow.yml
@@ -32,7 +32,7 @@ env:
   FTR_PATH: 'ftr'
   START_CMD: 'node ../scripts/opensearch_dashboards --dev --no-base-path --no-watch --savedObjects.maxImportPayloadBytes=10485760 --server.maxPayloadBytes=1759977 --logging.json=false --data.search.aggs.shardDelay.enabled=true'
   OPENSEARCH_SNAPSHOT_CMD: 'node ../scripts/opensearch snapshot -E cluster.routing.allocation.disk.threshold_enabled=false'
-  CYPRESS_BROWSER: 'chromium'
+  CYPRESS_BROWSER: 'electron'
   CYPRESS_VISBUILDER_ENABLED: true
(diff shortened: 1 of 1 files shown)


ananzh (approved):

ananzh:
My local tests do show better results with electron. I approved base on local results.
But will this prevent us finding errors in chrome? @AMoo-Miki @abbyhu2000 @kavilla

abbyhu2000:
Originally this PR is for testing if switching from chromium to electron would improve flaky data explorer cypress tests; however, based on the result: https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/7037458640/job/19152268943, the same test are still flaky. Will close this PR as electron has not show improvement on the CI.

PR #5572, opened 2023-12-05
Custom Branding: Relative URL should be allowed

Alankarsharma:
### Description

We have open search running behind a reverse proxy along with other service (Logo is deployed here). From OpenSearch server external URLs are not accessible so if we give this value then Dashboard on startup are not able to fetch the logo and revert to original, if we enter internal URL then server starts properly but logo are not loaded as internal URL is not accessible from external network.

Here if relative URL is passed returning true so that server can start properly and logos are also loaded properly in browser.

Workaround like hostname mapping work only if both internal and external service are deployed on HTTP or HTTPS, but in case of SSL offloading this also fails.
 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/server/rendering/rendering_service.tsx
+++ b/src/core/server/rendering/rendering_service.tsx
@@ -171,7 +171,7 @@ export class RenderingService {
     };
   }
 
-  public async stop() { }
+  public async stop() {}
 
   /**
    * Setups HTTP Agent if SSL is enabled to pass SSL config
--- a/src/core/server/rendering/rendering_service.tsx
+++ b/src/core/server/rendering/rendering_service.tsx
@@ -366,7 +366,7 @@ export class RenderingService {
    */
   public isUrlValid = async (url: string, configName?: string): Promise<boolean> => {
     if (url === '/') {
-      return false;
+      return true;
     }
     if (url.match(/\.(png|svg|gif|PNG|SVG|GIF)$/) === null) {
       this.logger.get('branding').error(`${configName} config is invalid. Using default branding.`);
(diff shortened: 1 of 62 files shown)


kavilla:
Hello @Alankarsharma, thanks for opening!

Could you please sign the commit? https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/CONTRIBUTING.md#developer-certificate-of-origin.

I merged the branch with `main` to recheck the tests so you might have to pull. But you can amend the commit.

src/core/server/rendering/rendering_service.tsx
```diff
@@ -171,7 +171,7 @@ export class RenderingService {
     };
   }
 
-  public async stop() {}
+  public async stop() { }
```
abbyhu2000, line 174: Could you delete the empty space here? it causes a linter error @Alankarsharma

abbyhu2000:
Just a linter error, otherwise LGTM!

src/core/server/rendering/rendering_service.tsx
Alankarsharma: done, thanks for the review

abbyhu2000:
@Alankarsharma Could you please also add the change in the Changelog.md?

Alankarsharma:
> @Alankarsharma Could you please also add the change in the Changelog.md?

Done

SuZhou-Joe (approved):
Overall, LGTM and relative url should not introduce any security concern like ssrf as we've already do a suffix check beforehand and this config can only be set by cluster admin.

ananzh (approved):

Issue #5423, opened 2023-11-03
[BUG]`match_all` query filters applied despite courier:ignoreFilterIfFieldNotInIndex setting enabled


ananzh:
**Describe the bug**
The `match_all` query filter is applied and included in the OpenSearch request when `courier:ignoreFilterIfFieldNotInIndex` is enabled. This setting is used to skip filters that apply to fields that don’t exist in the index for a visualization. However, we see filter applied even if the field the filter applies to does not exist in the index pattern of the visualization, which contradicts the expected behavior as per the setting's definition.

<img width="1695" alt="Screenshot 2023-11-03 at 10 56 38" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/79961084/2c71d0c5-1e2a-40ec-a0de-70b8d483d0ed">

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/79961084/d58f433d-8c92-406b-be15-739aa301863e

**To Reproduce**
* Navigate to `Stack Management` > `Advanced Settings` > `courier:ignoreFilterIfFieldNotInIndex`.
* Enable the setting `courier:ignoreFilterIfFieldNotInIndex` and save the changes.
* Create a visualization of type DataTable and choose an index pattern.
* In Buckets, split rows, choose aggregation as "Terms", select a field of type string, and click Update.
* Add a normal filter (match_all query) that targets a non-existent field in the chosen index pattern.
* Click `Inspect` -> `View: Requests` -> Click `request`:  Here the custom query DSL that we created is still sent in the request. 

**Expected behavior**
**with courier:ignoreFilterIfFieldNotInIndex Enabled**
Filters that target fields not present in the index pattern should be skipped, and not included in the request. Both `prefix` query and `match_all` query should not send the request for non-existent fields.

**OpenSearch Version**
any

**Dashboards Version**
any

**Plugins**

na

ananzh:
https://github.com/opensearch-project/dashboards-visualizations/issues/281

anand3493:
@ananzh  I believe you misunderstood my issue raised in the bug https://github.com/opensearch-project/dashboards-visualizations/issues/281

Re-explaining:
courier:ignoreFilterIfFieldNotInIndex is enabled now

Irrespective of the type of filter (either custom query DSL or normal filter), the filter should be applied if the field is present in the chosen index.

The filter should not be applied only when the field is not present on the index pattern

[...]

PR #5577, opened 2023-12-06
[BUG][Data] Support for custom filters with heterogeneous data fields

kavilla:
### Description

When enabling the advanced setting `courier:ignoreFilterIfFieldNotInIndex` Custom OpenSearch Query DSL filters could technically be applied to index patterns that map to indices that are not exactly the same. Since the custom query filter is a user input then users can really type anything that they need. Or any field that they know is present but we do not know for sure.

Therefore, we can check if the id which is the index pattern title to check if we should apply the filter or not.

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5581
I believe issue:
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5423

Should closed as that is expected functionality.

## Screenshot

Request before:
Screenshot 2023-12-06 at 2 16 05 AM

Request after:
Screenshot 2023-12-06 at 2 19 56 AM

## Testing the changes

Using the sample data web logs and created a data table. 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ananzh (approved):

ananzh:
Issue Resolve https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5423

src/plugins/data/common/opensearch_query/opensearch_query/filter_matches_index.ts
```diff
@@ -31,14 +31,14 @@
 import { IIndexPattern, IFieldType } from '../../index_patterns';
 import { Filter } from '../filters';
 
-/*
- * TODO: We should base this on something better than `filter.meta.key`. We should probably modify
- * this to check if `filter.meta.index` matches `indexPattern.id` instead, but that's a breaking
- * change.
- */
```
joshuarrrr, line 38: Let's make an issue that captures the uncertainty we still have here, where we can discuss the other underlying issues, including lack of validation on the custom filter inputs.

joshuarrrr (approved):

kavilla:
> What do we think about backporting this to 1.3?

i think so

kavilla:
Verified newest commit locally.

ZilongX (approved):

PR #5338, opened 2023-10-19
Replace `node-sass` with `sass-embedded`

AMoo-Miki:
### Description

Replace `node-sass` with `sass-embedded` in preparation for OUI 1.4.0.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [X] Commits are signed per the DCO using --signoff

PR #5593, opened 2023-12-11
[Manual Backport 2.x] Replace `node-sass` with `sass-embedded` (#5338)

AMoo-Miki:
Replace node-sass with sass-embedded #5338 
* Dart's `sass` uses a lot more memory that `node-sass` which causes failures with CI on Windows when building platform plugins: pagefile size was bumped.

(cherry picked from commit f822702cfc4466201f6c48d805513f5d8c82d52e)


AMoo-Miki:
CHANGELOG seems to have the entry needed; skipping the workflow for it.

package.json
```diff
@@ -128,7 +128,7 @@
   "dependencies": {
     "@aws-crypto/client-node": "^3.1.1",
     "@elastic/datemath": "5.0.3",
-    "@elastic/eui": "npm:@opensearch-project/oui@1.3.0",
+    "@elastic/eui": "npm:@opensearch-project/oui@1.4.0-alpha.2",
```
joshuarrrr, line 131: I don't think we want alpha versions in `2.x` branch? In `main` it makes sense, but we should keep `2.x` releasable.

packages/osd-interpreter/package.json
```diff
@@ -27,7 +27,7 @@
     "del": "^6.1.1",
     "getopts": "^2.2.5",
     "pegjs": "0.10.0",
-    "sass-loader": "npm:@amoo-miki/sass-loader@10.4.1-node-sass-9.0.0-libsass-3.6.5",
+    "sass-loader": "npm:@amoo-miki/sass-loader@10.4.1-node-sass-9.0.0-libsass-3.6.5-with-sass-embedded.rc1",
```
joshuarrrr, line 30: This patch is getting increasingly gross 🤢

packages/osd-optimizer/src/integration_tests/__snapshots__/basic_optimization.test.ts.snap
joshuarrrr, line 1: Anything notable about the diff here? It's not a particularly great snapshot...

joshuarrrr (commented):
@AMoo-Miki Mostly looks good to me. My only hesitation is about the OUI version bump to a non-released version. Is that strictly necessary here?

package.json
ZilongX: + 1

packages/osd-interpreter/package.json
ZilongX: since it's an npm for, can we give it a more friendly name, also are we able to keep `sass-loader` up to date since the official one is already at `13.3.2` https://github.com/webpack-contrib/sass-loader/tree/v13.3.2, it might cause future troubles in case we need to bump `sass-loader` version down the road

package.json
AMoo-Miki: This is temporary until OUI 1.4 is released.

packages/osd-interpreter/package.json
AMoo-Miki: @ZilongX we can only use sass-loader@9 with our ancient webpack@4.

packages/osd-optimizer/src/integration_tests/__snapshots__/basic_optimization.test.ts.snap
AMoo-Miki: I diffed it manually. Two types of changes: (1) some \\n removed and (2) color string to hex change

package.json
joshuarrrr: Yeah, but it makes the 2.x branch unreleasable, which I thought we wanted to avoid.

packages/osd-optimizer/src/integration_tests/__snapshots__/basic_optimization.test.ts.snap
joshuarrrr: thanks - I figured, but that was the assurance I was looking for 😁

joshuarrrr (approved):

SuZhou-Joe (approved):

Issue #5504, opened 2023-11-17
[Feature] Unified query language selection in Data Explorer


shanilpa:
## Context
The following issue describes a step towards unification towards a common [Data Explorer](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4165) in OpenSearch Dashboards. Currently (v2.11), in OpenSearch Dashboards there is support for several querying languages including DQL (Dashboards Query Language), Lucene, PPL (Piped Processing Language), and SQL. However, these query languages exist in separate applications with no clear way for users to access them outside of just knowing what is supported where. This results in a poor user experience. This issue existed for data sources which we resolved through an improved [data source selector component](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4482). Now we seek to do the same for query languages.

## The challenge
Part of this solution has been implemented in Observability Logs (v2.11) where users can select PPL, SQL and DQL as a query language from a query language selector - DQL redirects users to Discover. Within Discover users do not have a way to select the other available languages. This issue seeks to address this while improving, and unify the experience in Discover and Observability Logs.

Query picker in Observability Logs 2.11
<img width="1696" alt="image" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/113368824/a3aa5208-ee59-421d-86ac-b361d056528d">

### Discover:
<img width="1777" alt="image" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/113368824/bf30060b-e1d3-404f-a66a-0591f9570b2a">

### Observability Logs:
<img width="1767" alt="image" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/113368824/361df5c2-ab34-412a-becb-e6633679fdce">

The proposed solution places the query language switcher in the same location and with the same look and feel in both applications so users know where to find the functionality and expect consistent actions and behaviors.

## Additional details
1. Query language is dependent on the data source selected. Some data sources may not allow for querying with certain languages. As a result the query language picker needs to be dynamic and only show the querying languages supported for the selected data source
2. Short term the query language acts as a redirection mechanism between Discover and Observability Logs applications.
3. In future releases we will seek to break the query language selector and query bar from the composed component that contains the time range selector and execute button. This will allow users more space to query but also modularize the querying component for use elsewhere in the application. [More information on the proposed North Star for Data Explorer](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5507).
[...]

PR #5623, opened 2023-12-15
[Data Explorer] Adding PPL to query language selector

abbyhu2000:
### Description
A quick solution for adding navigation from data explorer page to log explorer page by adding the option of choosing PPL in the query language selector in the top nav bar.

### Issues Resolved
resolves part of #5504 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/types.ts
+++ b/src/plugins/data/public/types.ts
@@ -42,7 +42,7 @@ import { IndexPatternsContract } from './index_patterns';
 import { IndexPatternSelectProps, StatefulSearchBarProps } from './ui';
 import { UsageCollectionSetup } from '../../usage_collection/public';
 import { DataSourceStart } from './data_sources/datasource_services/types';
-import { ObservabilitySetup } from '../../../../plugins/dashboards-observability/public/types';
+
 export interface DataPublicPluginEnhancements {
   search: SearchEnhancements;
 }
--- a/src/plugins/data/public/types.ts
+++ b/src/plugins/data/public/types.ts
@@ -50,7 +50,7 @@ export interface DataPublicPluginEnhancements {
 export interface DataSetupDependencies {
   expressions: ExpressionsSetup;
   uiActions: UiActionsSetup;
-  usageCollection?: UsageCollectionSetup;
+  usageCollection?: UsageCollectionSetup
 }
 
 export interface DataStartDependencies {
--- a/src/plugins/data/public/types.ts
+++ b/src/plugins/data/public/types.ts
@@ -42,7 +42,7 @@ import { IndexPatternsContract } from './index_patterns';
 import { IndexPatternSelectProps, StatefulSearchBarProps } from './ui';
 import { UsageCollectionSetup } from '../../usage_collection/public';
 import { DataSourceStart } from './data_sources/datasource_services/types';
-
+import { ObservabilitySetup } from '../../../../plugins/dashboards-observability/public/types';
 export interface DataPublicPluginEnhancements {
   search: SearchEnhancements;
 }
@@ -50,7 +50,7 @@ export interface DataPublicPluginEnhancements {
 export interface DataSetupDependencies {
   expressions: ExpressionsSetup;
   uiActions: UiActionsSetup;
-  usageCollection?: UsageCollectionSetup
+  usageCollection?: UsageCollectionSetup;
 }
 
 export interface DataStartDependencies {
--- a/src/plugins/data/public/ui/query_string_input/language_switcher.tsx
+++ b/src/plugins/data/public/ui/query_string_input/language_switcher.tsx
@@ -45,14 +45,30 @@ import { FormattedMessage } from '@osd/i18n/react';
 import React, { useState } from 'react';
 import { useOpenSearchDashboards } from '../../../../opensearch_dashboards_react/public';
 import { useCallback } from 'react';
+import { Observable } from 'rxjs';
+import { useObservable } from 'react-use';
 
 interface Props {
   language: string;
   onSelectLanguage: (newLanguage: string) => void;
   anchorPosition?: PopoverAnchorPosition;
+  currentApp$?: Observable<string|undefined>
 }
 
 export function QueryLanguageSwitcher(props: Props) {
+  const dataExplorerOptions = [
+    {
+      label: 'DQL'
+    },
+    {
+      label: 'Lucene',
+    },
+    {
+      label: 'PPL'
+    }
+  ]
+  const [selectedLanguage, setSelectedLanguage] = useState([dataExplorerOptions[0]])
+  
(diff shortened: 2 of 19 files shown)


src/plugins/data/public/types.ts
```diff
@@ -42,7 +42,7 @@ import { IndexPatternsContract } from './index_patterns';
 import { IndexPatternSelectProps, StatefulSearchBarProps } from './ui';
 import { UsageCollectionSetup } from '../../usage_collection/public';
 import { DataSourceStart } from './data_sources/datasource_services/types';
-
+import { ObservabilitySetup } from '../../../../plugins/dashboards-observability/public/types';
```
ashwin-pc, line 45: This import breaks when the observability plugin isn't installed right? Isn't there a better way to check if the plugin is installed? The data source selector uses an API to do it I believe

ashwin-pc (commented):
Can we make this a lot simpler? Like all we want is a simple if condition where the existing pop over is placed to be replaced with a new popp er with the new options if the current app is data explorer and observability APIs are available. Feel like the whole change should be contained in just a single file that we can remove when this hard coded logic isn't needed.

Also add a to-do with an issue that call out that this is a temporary solution and needs to be removed ASAP and add the issue URL to the code comment

src/plugins/data/public/ui/query_string_input/language_switcher.tsx
```diff
@@ -42,15 +43,32 @@ import {
...
+    {
+      label: 'lucene',
+    },
+    {
+      label: 'PPL',
+    },
+  ];
+  const [selectedLanguage, setSelectedLanguage] = useState([dataExplorerOptions[0]]);
+
   const osdDQLDocs = useOpenSearchDashboards().services.docLinks?.links.opensearchDashboards.dql
```
ruanyl, line 72: Maybe we can access `application?.currentAppId$;` here directly instead of passing `currentApp$` all the way down to `QueryLanguageSwitcher`?

src/plugins/discover/public/build_services.ts
```diff
@@ -123,6 +124,7 @@ export function buildServices(
     toastNotifications: core.notifications.toasts,
     uiSettings: core.uiSettings,
     visualizations: plugins.visualizations,
+    useNewQuerySelector: !!plugins.observabilityDashboards,
```
ruanyl, line 127: Shall we leverage `capabilities` service to check if `PPL` is enabled instead of relying on the existence of observabilityDashboards plugin?

kavilla: does observability dashboards populate the capabilities service?

ideally, i think the best route would be a public registry, (if it's not available already) for the query language. for which plugins can populate with their language and routes like how lucene and dql does. so when a plugin starts up it registers its query language and the routes it should hit.

however, i think this is done in its current state due to priorities in external plugins and that the observability plugin is suppose to support the functionality of re-routing. in attempt to avoid confusion where i can go from [observability data explorer](https://playground.opensearch.org/app/observability-logs#/explorer/Zr-XIowB4v5SU7kSv17P) to [discover data explorer ](https://playground.opensearch.org/app/data-explorer/discover#?_a=(discover:(columns:!(_source),isDirty:!f,sort:!()),metadata:(indexPattern:d3d7af60-4c81-11e8-b3d7-01146121b73d,view:discover))&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15h,to:now))&_q=(filters:!(),query:(language:kuery,query:''))). Where when the user selects DQL it takes you to the Discover view (even though there appears to be a bug that prevents that feature.

cc: @abbyhu2000 @ashwin-pc is that correct?

src/plugins/data/public/types.ts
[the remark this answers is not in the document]
abbyhu2000: removed

abbyhu2000:
Created a follow-up issue for this: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5628

src/plugins/data/public/types.ts
abbyhu2000: Updated it using application.application$ service to determine if observability plugin is intalled.

src/plugins/discover/public/build_services.ts
abbyhu2000: Yea when user selects DQL on observability page, it will take user to data explorer. 

Also updated to use application service to determine if observability plugin is installed.

src/plugins/data/public/ui/query_string_input/language_switcher.tsx
```diff
@@ -51,6 +55,32 @@ interface Props {
 }
 
 export function QueryLanguageSwitcher(props: Props) {
+  const opensearchDashboards = useOpenSearchDashboards<IDataPluginServices>();
+  const { application } = opensearchDashboards.services;
+  const currentApp$ = application?.currentAppId$;
+  let useNewQuerySelector;
+  application?.applications$.subscribe((applications) => {
```
SuZhou-Joe, line 62: This subscribe should be wrapped inside a useEffect I suppose?

src/plugins/data/public/ui/query_string_input/language_switcher.tsx
```diff
@@ -51,6 +55,32 @@ interface Props {
...
+    applications.forEach((applicationEntry) => {
+      if (applicationEntry.id === 'observability-dashboards') {
+        useNewQuerySelector = true;
+        return;
+      }
+    });
+  });
+
+  const dataExplorerOptions = [
```
SuZhou-Joe, line 71: Can we add an enum for these language options like enum ENUM_LANGUAGES { DQL = "DQL" } ? And I suggest adding an id field inside options though under context of language, label no need to wrapped by i18n.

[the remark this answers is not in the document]
abbyhu2000: Added SQL which will also redirect to observability page.

`dataExplorerOptions` is used by `<EuiComboBox>` to show the dropdown list. Do we want to change how the label is shown on the dropdown from `lucene` to `luceneLabel` etc? It looks kinda crowded to me.
Screenshot 2024-01-03 at 12 15 18 PM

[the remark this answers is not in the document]
ashwin-pc: oh no, i meant that on line 90 you have an internationalized version of these labels. Can we reuse them? and add new labels for the missing languages?

ashwin-pc: We will still need i18n since its purpose is to individually identify the language name. e.g. `data.query.queryBar.dqlLanguageName`

ashwin-pc (commented):
Just 2 minor changes, the rest of the code looks good to me :)

abbyhu2000:
Close this PR since we want to build a more unified experience for query language.

Issue #5639, opened 2023-12-22
[RFC] Clickjacking Mitigation


tianleh:
**_(This RFC will now depend on https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5796 and its content will be updated.)_**

# Problem Statement <a href="#problem-statement" id="problem-statement">#</a>
Clickjacking is a vulnerability when an application allows its content to be loaded inside a frame of an attacker-controlled webpage. The attacker-controlled webpage contains UI elements aligned in such a way that when the customer thinks they are clicking on a UI element in the visible page, they are clicking on the invisible element from the framed page floating above it. Security testers find this vulnerability existing in OpenSearch Dashboards. The root cause is that OSD by default allows dashboards to be shared as embed code in iFrame (no `x-frame-options` HTTP headers or `frame-ancestors` in CSP rules). All downstream OSD offerings running on top of open source OSD inherit the same vulnerability. 

We will evaluate multiple approaches to mitigate this issue by minimizing the impact to user experiences and avoiding duplicated work in downstream OSD offerings.

# Desired State <a href="#desired-state" id="desired-state">#</a>

We want to resolve the vulnerability with the approval from Security testers. In the meanwhile, we continue to support the existing customers who use iFrame features. A self service option is preferred to minimize the operational load to OSD system administrator. Our solution shall simplify the effort for all downstream OSD offering.

# Out of Scope <a href="#out-of-scope" id="out-of-scope">#</a>
We do not plan to backport the changes to versions already released to minimize impact to existing customers using iFrame features. Only new versions will have this mitigation.

We do not plan to provide new UIs as long as existing UIs could meet the requirements.

# Challenges <a href="#challenges" id="challenges">#</a>
There are seven challenges for this project. 

1. OSD YML change requires a server restart to take effect. Any approach proposing YML change will have the risk of service availability interruption. Furthermore, for multi-tenancy service, it is infeasible to support dynamic customer configurations through the current YML file.

2. The existing request handlers in core OSD only rely on the OSD YML file. However, a proposed approach may need to read OSD indices and thus requires an instance of OpenSearch client which may not be ready yet during service setup. The existing OSD Security plugin is able to perform authentication for every request. It gives potential directions to structure the new feature into a plugin (inside core) or postpone the registration of the new handler after the readiness of OpenSearch client.

[...]

Issue #4713, opened 2023-08-10
Plugins Usage of Opensearch Core System and Hidden Indices


Rishikesh1159:
### Description/Concept of System Indices and Hidden Indices on Opensearch core:
**System Index** - An index must extend [system index plugin](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/plugins/SystemIndexPlugin.java) for a index to be called as system index.
Example: Security Plugin correctly extends and uses system Indices, more info [here](https://github.com/search?q=org%3Aopensearch-project+systemindex+repo%3Aopensearch-project%2Fsecurity&type=code)

**Hidden index** -  An index must have hidden [SETTING_INDEX_HIDDEN = "index.hidden"](https://github.com/opensearch-project/OpenSearch/blob/7278f434cd83f095623e34c2105f1b79d247af18/server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java#L598) set on the index setting to call it as hidden. It doesn't matter if it starts with "." or not.
Example: Asynchronous-Search plugin correctly sets the index setting value [here](https://github.com/search?q=org%3Aopensearch-project+SETTING_INDEX_HIDDEN+repo%3Aopensearch-project%2Fasynchronous-search&type=code)

**Misconception:**
Many plugins still misunderstand the actual definition of system and hidden indices. Usual misconception is that any index starting with `.` like `.indexName` is a system or hidden index, but this is incorrect. Any user can create an index with `.indexName` which is neither a system or hidden index. So users can mistakenly still create index starting with "." there is nothing stopping them from doing it.

To avoid this misconception all plugins should adopt/on-board with concept of system and hidden indices defined in opensearch core.

### **Goal:**
The main ask of this issue is to make sure all plugins having/using system and hidden indices must on-board/adopt with concept defined in opensearch core. 

Any plugin already on-board with opensearch core defined concept of system and hidden indices can ignore this issue and close the issue as completed on the plugin repo.

### **Open questions**
In case of any questions or issues, please post it in https://github.com/opensearch-project/OpenSearch/issues/9239 issue

AMoo-Miki:
While plugins should be using the SavedObjects API and not create their own indices, OSD should look into how they define their metadata-store.

Rishikesh1159:
@kavilla @AMoo-Miki do we have any update on this issue? If not can you guys please prioritize this one.

kavilla:
[...]

PR #5641, opened 2023-12-23
Support dynamic CSP rules to mitigate clickjacking

tianleh:
### Description

This PR is to mitigate the Clickjacking vulnerability as stated https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5639.

~~**Notes for the repo maintainers**: please add backport label to the branch `2.x`. We are targeting at `2.12.0` release for this feature.~~ 

We will target at `2.13.0`. Please help add the backport label to the branch `2.x`.

**12/27/2023**
Unit tests are being added.

**12/23/2023**
For now, I am sending the PR without any tests to collect early feedback. More tests will be added later.

### Issues Resolved

fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5639

### Test Case 1

Step 0: There are no csp rules configured in OSD yml.

Step 1: Verify that there is no index `.opensearch_dashboards_config` by default.
Run command `GET .opensearch_dashboards_config/_doc/csp.rules`

Receive this response.
Screenshot 2023-12-26 at 2 06 45 PM

Step 2: Verify the response header has added `frame-ancestors 'self'` by default.

Screenshot 2023-12-26 at 2 08 33 PM

Step 3: Verify through a local test html file.
```
<html>
<head>
<title>Clickjack test page</title>
</head>
<body>
<p>Test clickjacking!</p>
<iframe src="http://localhost:5601/app/home#/" width="500" height="500"></iframe>
</body>
</html>
```

Open the test html in browser and expect to see the following page.

Screenshot 2023-12-26 at 2 18 46 PM

Step 4: Update the index `.opensearch_dashboards_config` to allow embedding inside local file.

```
PUT .opensearch_dashboards_config/_doc/csp.rules
{
  "value": "frame-ancestors 'self' file://* filesystem:"
}
```

Step 5: Verify that the test html can open now.

Screenshot 2023-12-26 at 2 17 39 PM

### Test Case 2

Step 0: There are CSP rules configured in OSD YML.
```
csp.rules: ["frame-ancestors 'self' file://* filesystem:"]

```

Step 1:
Verify that by default there is no index `.opensearch_dashboards_config`.

```
GET .opensearch_dashboards_config/_doc/csp.rules

```

Returns `index_not_found_exception`

Step 2:

Verify that the values from the OSD YML are used.

Screenshot 2023-12-28 at 8 06 02 PM

Step 3:

Update the `.opensearch_dashboards_config` with different CSP rules from the YML.

```

PUT .opensearch_dashboards_config/_doc/csp.rules
{
  "value": "frame-ancestors 'self' http://example-test.com"
}
```

Step 4:

Verify that the values from the index are used.
Screenshot 2023-12-28 at 8 12 27 PM

Step 5: 

Update `.opensearch_dashboards_config` with empty rules.

```
PUT .opensearch_dashboards_config/_doc/csp.rules
{
  "value": ""
}

```

Step 6:

Verify that the values from the YML are used.

Screenshot 2023-12-28 at 8 14 42 PM

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


tianleh:
The one failed https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/7353558209/job/20019665599?pr=5641 to build on macOS ARM64 failed but was successful in previous run

https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/7353233916

Will monitor in next run. 

**Could anyone with permission rerun it for me or grant me the re-run workflow permission?**

tianleh:
Hi team @opensearch-project/opensearch-dashboards-core , could you please help review?

cc @seraphjiang @wbeckler @zengyan-amazon

src/plugins/csp_configuration_provider/server/csp_handlers.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    const myClient = getCspClient(coreStart.opensearch.client.asInternalUser);
+
+    const existsValue = await myClient.exists(OPENSEARCH_DASHBOARDS_CONFIG_INDEX_NAME);
+
+    if (!existsValue) {
+      return toolkit.next({});
+    }
+
+    const cspRules = await myClient.get(
```
ruanyl, line 35: The http service has `csp` exposed and you can read it via `http.csp`, shall we use that instead of manually call query it from the index?

src/plugins/csp_configuration_provider/server/provider.ts
```diff
@@ -0,0 +1,65 @@
...
+import { CspClient } from './types';
+
+export class OpenSearchCspClient implements CspClient {
+  private client: OpenSearchClient;
+
+  constructor(inputOpenSearchClient: OpenSearchClient) {
+    this.client = inputOpenSearchClient;
+  }
+
+  async exists(configurationName: string): Promise<boolean> {
```
SuZhou-Joe, line 17: Do we need this method? It seems that `get` can cover `exists`.

src/plugins/csp_configuration_provider/README.md
```diff
@@ -0,0 +1,64 @@
+# CspConfigurationProvider
+
+A OpenSearch Dashboards plugin
+
+This plugin introduces the support of self service dynamic configuration of Content Security Policy (CSP) rules without requiring a server restart. It registers a pre-response handler to `HttpServiceSetup` which can get CSP rules from a new index `.opensearch_dashboards_config` and then rewrite to CSP header. OSD users could update the content of the index with new CSP rules and the change will take effect immediately.
```
SuZhou-Joe, line 5: Do we have to store the config into an index? If taking security plugin into consideration, some users may not be able to get the correct config because of lack of permission and fallback to default value. Then different users may have different Csp config.

src/plugins/csp_configuration_provider/server/plugin.ts
```diff
@@ -0,0 +1,100 @@
...
+  private setCspClient(inputCspClient: CspClient) {
+    this.cspClient = inputCspClient;
+  }
+
+  private getCspClient(inputOpenSearchClient: OpenSearchClient) {
+    if (this.cspClient) {
+      return this.cspClient;
+    }
+
+    return new OpenSearchCspClient(inputOpenSearchClient);
```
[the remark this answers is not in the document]
tianleh: Good suggestion! Updated.

src/plugins/csp_configuration_provider/server/provider.ts
tianleh: If we only use `get`, it will throw exception when there is no this index.

Also it will generate an error log in OSD log.

```
server    log   [22:47:21.193] [error][data][opensearch] [index_not_found_exception]: no such index [.opensearch_dashboards_config]
```

This will cause confusion and waste system resource to handle exception.

As a contrast, `exists` will handle the same case smoothly.

src/plugins/csp_configuration_provider/server/types.ts
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface CspConfigurationProviderPluginSetup {}
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface CspConfigurationProviderPluginStart {}
+
+export interface CspClient {
+  exists(configurationName: string): Promise<boolean>;
+
+  get(configurationName: string, cspRulesName: string): Promise<string>;
```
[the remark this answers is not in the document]
tianleh: When not existing, the code will return an empty string. Then we will use a default value. 

I do notice that there is one case that the return value will be `undefined`. I just updated the code to default to empty string.

tianleh:
cc current oncall @AMoo-Miki for visibility. Please help review and add the backport labels `2.x`.

seraphjiang:
@Flyingliuhub @bandinib-amzn @ZilongX would you help to review

kavilla:
pulling down

src/plugins/csp_configuration_provider/README.md
kavilla, line 5: on start up I believe we can give the option to populate this index and give the mapping.

One thing I'm worried about is the whole `.` usage in OSD not being actually treated as a system index and any issues if OpenSearch does any changes to enforce to pre-pending `.`

src/plugins/csp_configuration_provider/opensearch_dashboards.json
```diff
@@ -0,0 +1,9 @@
+{
+  "id": "cspConfigurationProvider",
+  "version": "1.0.0",
+  "opensearchDashboardsVersion": "opensearchDashboards",
+  "server": true,
+  "ui": false,
+  "requiredPlugins": ["navigation"],
```
kavilla, line 7: does this need navigation plugin?

src/plugins/csp_configuration_provider/server/csp_handlers.ts
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { CoreSetup, OnPreResponseHandler, OpenSearchClient } from '../../../core/server';
+import { CspClient } from './types';
+
+const OPENSEARCH_DASHBOARDS_CONFIG_INDEX_NAME = '.opensearch_dashboards_config';
```
kavilla, line 9: to verify, the ideal of putting it into a new config was to mitigate the current issue of private tenancy and global tenancy? would this be too generic of a name since it seems it can house more than just csp rules? perhaps we can make this configurable to avoid any potential conflicts?

like: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/opensearch_dashboards_config.ts#L50

by default it will defined as whatever we decide.

src/plugins/csp_configuration_provider/opensearch_dashboards.json
```diff
@@ -0,0 +1,9 @@
+{
+  "id": "cspConfigurationProvider",
```
kavilla, line 2: This gets automatically transformed to `csp_configuration_provider` for configurations. Do we want to consider providing a `configPath` or do we want to consider just naming this plugin more generic so that other stuff can be stuffed into it.

kavilla (commented):
Nice!

@AMoo-Miki I think you were looking into CSP stuff is this kind of the idea you had in mind?

The biggest thing for me is the OSD index that is defined externally which will make future iterations a little bit difficult.

src/plugins/csp_configuration_provider/opensearch_dashboards.json
tianleh: Actually not. It was automatically added by the plugin creation command. Let me remove it.

src/plugins/csp_configuration_provider/server/csp_handlers.ts
tianleh: 1 Yes. The key motivation is to avoid the issue of private/global tenancies. 
2 It is intended to have a generic name for housing more configurations which have similar issue.
3 Great suggestion! I just made a change to make the name configurable

src/plugins/csp_configuration_provider/opensearch_dashboards.json
tianleh: The new index could house more configurations in the future. However, this plugin is only for CSP rules. Its logic of modifying the request handler is unique to CSP rules. Other configurations may have different use cases and may be used in other existing or new plugins or in core.

src/plugins/csp_configuration_provider/README.md
tianleh: 1. I intentionally do not create the new index to avoid race condition. (The existing `.kibana` often runs into this problem when there are multiple nodes). Furthermore, for customers who do not want to change the default behavior, there is no need for the new index to exist. 

Currently there is only one doc in the index with only one field. It looks like that an index mapping may not be necessary yet. Furthermore, it may be infeasible to define a mapping for a non-exist index.

2. I read through that issue about system index and notice this comment https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4713#issuecomment-1835865775 The way we name the new index will make it match the current regular expression and get the benefits provided by OpenSearch core.

tianleh:
@kavilla Thanks for all the comments! I have replied/addressed all of them. Please help take a look again.

Also there is a failed workflow about macos build. https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/7454245516/job/20281220506?pr=5641 I believe it is a transient error. Could you also help re-run it? I don't have the re-run permission.

tianleh:
> @Flyingliuhub @bandinib-amzn @ZilongX would you help to review

Hi, @Flyingliuhub @bandinib-amzn @ZilongX Could you please take a look? The context is detailed the related RFC issue. Also your thoughts from the perspective of dashboards anywhere/metadata storage are appreciated.

tianleh:
> Nice!
> 
> @AMoo-Miki I think you were looking into CSP stuff is this kind of the idea you had in mind?
> 
> The biggest thing for me is the OSD index that is defined externally which will make future iterations a little bit difficult.

Hi, @AMoo-Miki Would you please share your thoughts so that we can merge this PR? cc @kavilla

tianleh:
Need help from maintainers to re-run the failed workflow "Build and test / Build min release artifacts on macOS x64"
https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/7470103512/job/20328249268?pr=5641

Also created an issue to track this flaky workflow https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5677

ZilongX:
So basically now we're disabling the iframe embedding by default (through config in `csp/config.ts`) then leverage yml to specify any diff-from-default CSP config which will be persisted in index `.opensearch_dashboards_config` ?

tianleh:
> So basically now we're disabling the iframe embedding by default (through config in `csp/config.ts`) then leverage yml to specify any diff-from-default CSP config which will be persisted in index `.opensearch_dashboards_config` ?

If the index `.opensearch_dashboards_config` has specified the CSP rules, then the system will use what is in the index.

If If the index `.opensearch_dashboards_config` doesn't exist, then the system will use whatever the yml has specified (whose default value has been updated to only allow self embedding).

tianleh:
@ZilongX see my reply in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5641#issuecomment-1885756725

ZilongX:
> > So basically now we're disabling the iframe embedding by default (through config in `csp/config.ts`) then leverage yml to specify any diff-from-default CSP config which will be persisted in index `.opensearch_dashboards_config` ?
> 
> If the index `.opensearch_dashboards_config` has specified the CSP rules, then the system will use what is in the index.
> 
> If If the index `.opensearch_dashboards_config` doesn't exist, then the system will use whatever the yml has specified (whose default value has been updated to only allow self embedding).

I don't see any entries added in yml, do you mean the default value sits in `src/core/server/csp/config.ts` aka 
```
        `script-src 'unsafe-eval' 'self'`,
        `worker-src blob: 'self'`,
        `style-src 'unsafe-inline' 'self'`,
        `frame-ancestors 'self'`,
```

tianleh:
> > > So basically now we're disabling the iframe embedding by default (through config in `csp/config.ts`) then leverage yml to specify any diff-from-default CSP config which will be persisted in index `.opensearch_dashboards_config` ?
> > 
> > 
> > If the index `.opensearch_dashboards_config` has specified the CSP rules, then the system will use what is in the index.
> > If If the index `.opensearch_dashboards_config` doesn't exist, then the system will use whatever the yml has specified (whose default value has been updated to only allow self embedding).
> 
> I don't see any entries added in yml, do you mean the default value sits in `src/core/server/csp/config.ts` aka
> 
> ```
>         `script-src 'unsafe-eval' 'self'`,
>         `worker-src blob: 'self'`,
>         `style-src 'unsafe-inline' 'self'`,
>         `frame-ancestors 'self'`,
> ```

Yes. @ZilongX

ZilongX:
> > > > So basically now we're disabling the iframe embedding by default (through config in `csp/config.ts`) then leverage yml to specify any diff-from-default CSP config which will be persisted in index `.opensearch_dashboards_config` ?
> > > 
> > > 
> > > If the index `.opensearch_dashboards_config` has specified the CSP rules, then the system will use what is in the index.
> > > If If the index `.opensearch_dashboards_config` doesn't exist, then the system will use whatever the yml has specified (whose default value has been updated to only allow self embedding).
> > 
> > 
> > I don't see any entries added in yml, do you mean the default value sits in `src/core/server/csp/config.ts` aka
> > ```
> >         `script-src 'unsafe-eval' 'self'`,
> >         `worker-src blob: 'self'`,
> >         `style-src 'unsafe-inline' 'self'`,
> >         `frame-ancestors 'self'`,
> > ```
> 
> Yes. @ZilongX

LGTM, we would also need to make the default CSP rule more robust but maybe out of the scope for this one ~

ZilongX (dismissed):

src/plugins/csp_configuration_provider/server/csp_handlers.ts
```diff
@@ -0,0 +1,50 @@
...
+    const existsValue = await myClient.exists(dynamicConfigIndex);
+
+    if (!existsValue) {
+      return toolkit.next({});
+    }
+
+    const cspRules = await myClient.get(
+      dynamicConfigIndex,
+      OPENSEARCH_DASHBOARDS_CONFIG_DOCUMENT_NAME
+    );
```
bandinib-amzn, line 38: In case of non-opensearch client (e.g. dynamodb, postgres etc), `dynamicConfigIndex` would be table name and `OPENSEARCH_DASHBOARDS_CONFIG_DOCUMENT_NAME` would be key or column, right? If my understanding is correct, we are kinda forcing implementation to follow same naming. Can we make more flexible to check whether dynamic configuration is available or not or get csp rule irrespective of table name or key name? Let plugin decide how they want to store dynamic configuration. If needed we can do validation in core OSD to check CSP rule is passing validation (maybe using regex).

bandinib-amzn:
Overall looks good to me. Thanks for the change. I left one comment about the interfaces for CSP client. Also I have one question:

Suppose User A running OSD version 2.11 (or any version older than 2.12). They are using iframe (not self) to load OSD web page. Now they upgrade to their OSD 2.12. OSD inside iframe would fail, right? How can we make easy for user to understand next course of actions in order to fix it?

tianleh:
Got a comment to still provide an `update` API and customers could call the API to make the change.

src/plugins/csp_configuration_provider/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+```
+"script-src 'unsafe-eval' 'self'; worker-src blob: 'self'; style-src 'unsafe-inline' 'self'; frame-ancestors 'self'"
+```
+(Note that `frame-ancestors 'self'` is used to only allow self embedding and prevent clickjacking by default.)
+
+For OSD users who want to make changes, e.g allow this site `https://example-site.org` to embed OSD pages, they can update the index from DevTools Console.
+
+```
+PUT .opensearch_dashboards_config/_doc/csp.rules
```
seraphjiang, line 25: Direct database access introduce security risk and by pass validation. We should expose config update via api.

src/plugins/csp_configuration_provider/server/csp_handlers.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    const myClient = getCspClient(coreStart.opensearch.client.asInternalUser);
+
+    const existsValue = await myClient.exists(dynamicConfigIndex);
+
+    if (!existsValue) {
+      return toolkit.next({});
+    }
+
+    const cspRules = await myClient.get(
```
seraphjiang, line 35: need to handle error and exception

src/plugins/csp_configuration_provider/server/plugin.ts
```diff
@@ -0,0 +1,80 @@
...
+      return this.cspClient;
+    }
+
+    const openSearchCspClient = new OpenSearchCspClient(inputOpenSearchClient);
+    this.setCspClient(openSearchCspClient);
+    return this.cspClient;
+  }
+
+  public async setup(core: CoreSetup) {
+    this.logger.debug('CspConfigurationProvider: Setup');
```
seraphjiang, line 53: Remove debug code

src/plugins/csp_configuration_provider/server/provider.test.ts
```diff
@@ -0,0 +1,165 @@
...
+  OpenSearchClientMock,
+  opensearchClientMock,
+} from '../../../core/server/opensearch/client/mocks';
+import { OpenSearchCspClient } from './provider';
+
+const INDEX_NAME = 'test_index';
+const INDEX_DOCUMENT_NAME = 'csp_doc';
+const ERROR_MESSAGE = 'Service unavailable';
+
+describe('Provider', () => {
```
seraphjiang, line 17: Could we be more descriptive about test suite name to help locate code when test fail

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [CVE-2023-45133] Bump all babel dependencies from `7.16.x` to `7.22.9` to fix upstream vulnerability ([#5428](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5428))
 - [CVE-2023-45857] Bump `axios` from `0.27.2` to `1.6.1` ([#5470](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5470))
 - [CVE-2023-26159] Bump `follow-redirects` from `1.15.2` to `1.15.4` ([#5669](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5669))
+- Support dynamic CSP rules to mitigate clickjacking ([#5641](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5641))
```
seraphjiang, line 20: This is also a breaking change, since iframe will be disable by default if not from same original domain

src/plugins/csp_configuration_provider/server/provider.ts
```diff
@@ -0,0 +1,65 @@
...
+    try {
+      const data = await this.client.search({
+        index: configurationName,
+        body: query,
+      });
+
+      value = data?.body?.hits?.hits[0]?._source?.value || '';
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.error(
```
seraphjiang, line 58: Should we use logger

src/plugins/csp_configuration_provider/server/provider.ts
```diff
@@ -0,0 +1,65 @@
...
+            query: cspRulesName,
+          },
+        },
+      },
+    };
+
+    let value = '';
+
+    try {
+      const data = await this.client.search({
```
seraphjiang, line 50: Is doc id uniqued, if so why not use doc api instead of search api

src/plugins/csp_configuration_provider/server/provider.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    try {
+      const exists = await this.client.indices.exists({
+        index: configurationName,
+      });
+
+      value = exists.body;
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.error(
```
seraphjiang, line 28: Why output to console ?

src/plugins/csp_configuration_provider/README.md
tianleh: In progress to expose APIs.

CHANGELOG.md
tianleh: added to the `### 💥 Breaking Changes` as well

src/plugins/csp_configuration_provider/server/csp_handlers.ts
tianleh: Let me make the interface generic and irrespective of table name or key name

src/plugins/csp_configuration_provider/server/plugin.ts
tianleh: It is automatically generated by the plugin generation helper. Other plugins have this debug statement as well. See advanced settings for example https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/advanced_settings/server/plugin.ts#L48

src/plugins/csp_configuration_provider/server/csp_handlers.ts
tianleh: will add

src/plugins/csp_configuration_provider/server/provider.test.ts
tianleh: will add

src/plugins/csp_configuration_provider/server/provider.ts
tianleh: will pass in `logger` and remove console.

src/plugins/csp_configuration_provider/server/provider.ts
tianleh: will switch to doc api

src/plugins/csp_configuration_provider/server/provider.ts
tianleh: updating.

src/plugins/csp_configuration_provider/opensearch_dashboards.json
tianleh: After receiving from another team about the possibility to use this index, I think your point is valid that we need to name the plugin more generic. Updated.

tianleh:
> Overall looks good to me. Thanks for the change. I left one comment about the interfaces for CSP client.
> 
> I also have two questions:
> 
>     1. Suppose User A running OSD version 2.11 (or any version older than 2.12). They are using iframe (not self) to load OSD web page. Now they upgrade to their OSD 2.12. OSD inside iframe would fail, right? How can we make easy for user to understand next course of actions in order to fix it?
> 
>     2. Did we test upgrade scenario / race condition in multi node cluster?

1: Yes. This is a breaking change. I updated the changelog to mention such. Also I have a document issue https://github.com/opensearch-project/documentation-website/issues/6139 to add instructions.

2. Different from the current `.kibana` index, we will not create this index by default. Only customers would create it when needed. Thus race condition is not an concern.

tianleh:
Below are the results of the API testing.

```
curl 'http://localhost:5601/api/configuration/getCspRules'

{"cspRules":"script-src unsafe-eval self; worker-src blob: self; style-src unsafe-inline self; frame-ancestors self https://example-site.org22"}%
```

```

curl 'http://localhost:5601/api/configuration/existsCspRules'

{"exists":true}%
```

```

curl 'http://localhost:5601/api/configuration/updateCspRules' -X POST -H 'Accept: application/json' -H 'Content-Type: application/json' -H 'osd-xsrf: osd-fetch' -H 'Sec-Fetch-Dest: empty' --data-raw '{"value":"script-src 'unsafe-eval' 'self'; worker-src blob: 'self'; style-src 'unsafe-inline' 'self'; frame-ancestors 'self' https://example-site.com"}'

{"updatedRules":"script-src unsafe-eval self; worker-src blob: self; style-src unsafe-inline self; frame-ancestors self https://example-site.com"}%
```


```
curl 'http://localhost:5601/api/configuration/getCspRules'
{"cspRules":"script-src unsafe-eval self; worker-src blob: self; style-src unsafe-inline self; frame-ancestors self https://example-site.com"}%
```


```
curl 'http://localhost:5601/api/configuration/deleteCspRules' -X POST -H 'osd-xsrf: osd-fetch' -H 'Sec-Fetch-Dest: empty'

{"deletedCspRulesName":"csp.rules"}%

```


```
curl 'http://localhost:5601/api/configuration/existsCspRules'
{"exists":false}%
```

tianleh:
While I am working with security reviewers for this change, the current status is ready for another round of PR review. @kavilla @bandinib-amzn @ZilongX

src/core/server/plugins/plugin_context.test.ts
```diff
@@ -98,6 +98,7 @@ describe('createPluginInitializerContext', () => {
     expect(configObject).toStrictEqual({
       opensearchDashboards: {
         index: '.kibana',
+        dynamic_config_index: '.opensearch_dashboards_config',
```
kavilla, line 101: nitpick: `dynamic_config_index` feels long. I would be satisfied with `config_index`

kavilla: for the second point nvm: https://github.com/opensearch-project/security/blob/main/src/main/resources/static_config/static_roles.yml#L103

it actually is a lot better to do `.opensearch_dashboards_*` because kibana server role because it is all setup for that.

src/plugins/configuration_provider/server/routes/index.ts
```diff
@@ -0,0 +1,163 @@
...
+import { ConfigurationClient } from '../types';
+
+export function defineRoutes(
+  router: IRouter,
+  getConfigurationClient: (inputOpenSearchClient: IScopedClusterClient) => ConfigurationClient,
+  logger: Logger
+) {
+  router.get(
+    {
+      path: '/api/configuration/existsCspRules',
```
kavilla, line 23: i think get `api/configuration/csp` or `api/config/csp` would be ok and then use `get` `post` put`

src/plugins/configuration_provider/server/provider.ts
```diff
@@ -0,0 +1,107 @@
...
+import { ConfigurationClient } from './types';
+
+const CSP_RULES_DOC_ID = 'csp.rules';
+
+export class OpenSearchConfigurationClient implements ConfigurationClient {
+  private client: IScopedClusterClient;
+  private configurationIndexName: string;
+  private readonly logger: Logger;
+
+  constructor(
```
kavilla, line 17: I think we can make this a service and then plugin developers can make this an optional plugin within their repo and then call to the DynamicConfigService(key, value) to insert or DynamicConfigService(key) to get value. Then we can get other plugins onboarded if they want to with little effort.

src/plugins/configuration_provider/server/routes/index.ts
tianleh: Sure. Updated the path structure.

src/core/server/plugins/plugin_context.test.ts
tianleh: Since point 2 and 3 are nvm, I just addressed point 1 to have a simple name.

src/plugins/configuration_provider/server/provider.ts
tianleh: I am evaluating this.

tianleh:
Still evaluating Rocky's comment https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5641#discussion_r1464172739

tianleh:
During feature review with PMs, we got a feedback to have the feature disabled by default until OSD 3.0.0 since this is a breaking change. Thus I am making the changes to restore the original default values. Also I am making this new plugin disabled by default.

tianleh:
Based on current development progress, we will not be able to catch the 2.12.0 release. Please help remove the label 2.12.0 to avoid confusion.

@kavilla

tianleh:
The PR has been evolving for a while. To summarize the current status, I have made the new plugin **disabled** by default. **Thus this is no longer a breaking change.** 

For @kavilla 's comment https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5641#discussion_r1464172739 about making this a service, it looks like a bigger scope. I have started a separate RFC issue https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5796 (still in progress)

I am proposing to make incremental progress on this by first merging current changes into 3.0.0. Then we will go through security review. After that, we backport to `2.13.0` after `2.12.0` is over. In parallel, we will collect feedback and work on the application level configuration service.

Let me know how you think @kavilla @ZilongX @bandinib-amzn @Flyingliuhub 

cc @seraphjiang

tianleh:
There is no need to have the label `backport 2.x` at this moment. Please help remove as well. @AMoo-Miki

tianleh:
We will pause the work on this PR and switch to work on the dependent issue https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5796

tianleh:
Since the logic of application configuration has been moved to this PR https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855, I am in progress to clean up this current PR so that it contains only Clickjacking (csp) related changes.

tianleh:
The related PR has been merged https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855

Syncing the latest changes to this PR

tianleh:
Updated test cases:

1. Initially there is no dynamic configurations for CSP rules.

Verify this by running

```
curl 'http://localhost:5601/api/appconfig/csp.rules'
```

2. Check the CSP rules for document request.

```
content-security-policy
	frame-ancestors 'self'; script-src 'unsafe-eval' 'self'; worker-src blob: 'self'; style-src 'unsafe-inline' 'self'
```

Screenshot 2024-03-04 at 4 37 35 PM


We can see that `frame-ancestors` has been automatically added.

3. Verify in a `iframe` tag

Screenshot 2024-03-04 at 4 39 17 PM

tianleh:
Find a strange issue that updating through the API will lose the single quotes around `self`. After some research, when calling the APIs, there shall be escaping wrapper around them. See this reference link https://stackoverflow.com/questions/32122586/curl-escape-single-quote

This is a valid call for reference.

```

curl 'http://localhost:5601/api/appconfig/csp.rules' -X POST -H 'Accept: application/json' -H 'Content-Type: application/json' -H 'osd-xsrf: osd-fetch' -H 'Sec-Fetch-Dest: empty' --data-raw '{"newValue":"frame-ancestors '\''self'\'' file://* filesystem:"}'

{"newValue":"frame-ancestors 'self' file://* filesystem:"}%
```

tianleh:
Please help take a look @kavilla @ZilongX @bandinib-amzn @BionIT 

cc @seraphjiang

src/plugins/application_config/server/index.ts
```diff
@@ -20,4 +20,8 @@ export function plugin(initializerContext: PluginInitializerContext) {
   return new ApplicationConfigPlugin(initializerContext);
 }
 
-export { ApplicationConfigPluginSetup, ApplicationConfigPluginStart } from './types';
+export {
+  ApplicationConfigPluginSetup,
```
seraphjiang, line 24: this code pretty/format change should not be part of csp PR right?

src/plugins/csp_handler/server/csp_handlers.ts
```diff
@@ -0,0 +1,71 @@
...
+  Logger,
+  OnPreResponseHandler,
+  OnPreResponseInfo,
+  OnPreResponseToolkit,
+  OpenSearchDashboardsRequest,
+} from '../../../core/server';
+
+const CSP_RULES_CONFIG_KEY = 'csp.rules';
+
+export function createCspRulesPreResponseHandler(
```
seraphjiang, line 19: it is worth to add description/jsdoc for this method

src/plugins/csp_handler/server/csp_handlers.ts
```diff
@@ -0,0 +1,71 @@
...
+  getConfigurationClient: (inputOpenSearchClient: IScopedClusterClient) => ConfigurationClient,
+  logger: Logger
+): OnPreResponseHandler {
+  return async (
+    request: OpenSearchDashboardsRequest,
+    response: OnPreResponseInfo,
+    toolkit: OnPreResponseToolkit
+  ) => {
+    try {
+      const shouldCheckDest = ['document', 'frame', 'iframe', 'embed', 'object'];
```
seraphjiang, line 31: where are these element comes from, add a link if any official doc

src/plugins/csp_handler/server/csp_handlers.ts
```diff
@@ -0,0 +1,71 @@
...
+
+      return toolkit.next({ headers: additionalHeaders });
+    } catch (e) {
+      logger.error(`Failure happened in CSP rules pre response handler due to ${e}`);
+      return updateFrameAncestors(cspHeader, toolkit);
+    }
+  };
+}
+
+function updateFrameAncestors(cspHeader: string, toolkit: OnPreResponseToolkit) {
```
seraphjiang, line 61: this method named updateXXX, but i didn't see any update code, would you elaborate a little, maybe add description for function or rename to more appropriate name

src/plugins/csp_handler/server/csp_handlers.ts
```diff
@@ -0,0 +1,71 @@
...
+  if (cspHeader.includes('frame-ancestors')) {
+    return toolkit.next({});
+  }
+
+  const additionalHeaders = {
```
seraphjiang, line 66: what does addtionalHeaders mean?

src/plugins/application_config/server/index.ts
tianleh: It adds a type `ConfigurationClient,` to the list. Thus it is not just format change.

Since the new plugin needs to import this from ApplicationConfigPlugin, lint checker will complain if this is not exposed at `index.ts` level. See the following lint error for reference.


```

ERROR 
      /home/ubuntu/workspace/clickjacking/osd-dev/OpenSearch-Dashboards/src/plugins/csp_handler/server/csp_handlers.ts
        6:37  error  Unexpected path "../../application_config/server/types" imported in restricted zone. Plugins may only import from top-level public and server modules  @osd/eslint/no-restricted-paths
      
      ✖ 1 problem (1 error, 0 warnings)
      
ERROR [eslint] errors
```

src/plugins/csp_handler/server/csp_handlers.ts
tianleh: Good callout! Just added.

src/plugins/csp_handler/server/csp_handlers.ts
tianleh: Added

src/plugins/csp_handler/server/csp_handlers.ts
tianleh: Renamed for better readability and added description as well.

src/plugins/csp_handler/server/csp_handlers.ts
tianleh: A request handler could potentially change many headers. E.g `osd-name`.  We call the variable `addtionalHeaders` to let people know that the next call `toolkit.next({ headers: additionalHeaders })` just updates the headers specified in the variable and will not flush out all other unmentioned headers. With this said, `addtionalHeaders` means **additional** headers on top of the existing unmentioned headers.

ZilongX (approved):

bandinib-amzn:
Can you please rebase to main branch? Also is it going to be breaking change?

config/opensearch_dashboards.yml
```diff
@@ -36,6 +36,10 @@
 # Set the value of this setting to true to enable plugin application config. By default it is disabled.
 # application_config.enabled: false
 
+# Set the value of this setting to true to enable plugin CSP handler. By default it is disabled.
+# It requires the application config plugin as its dependency.
+# csp_handler.enabled: false
```
bandinib-amzn, line 41: Why need extra configuration? Can we enable by default and if user adds record if config index and this plugin should enforce csp rules otherwise do nothing?

tianleh:
> Can you please rebase to main branch? Also is it going to be breaking change?

Just rebased. This is not a breaking change since it is disabled by default. @bandinib-amzn

src/plugins/csp_handler/README.md
```diff
@@ -0,0 +1,51 @@
+# CspHandler
+
+A OpenSearch Dashboards plugin
+
+This plugin is to support updating Content Security Policy (CSP) rules dynamically without requiring a server restart. It registers a pre-response handler to `HttpServiceSetup` which can get CSP rules from a dependent plugin `applicationConfig` and then rewrite to CSP header. Users are able to call the API endpoint exposed by the `applicationConfig` plugin directly, e.g through CURL. Currently there is no new OSD page for ease of user interactions with the APIs. Updates to the CSP rules will take effect immediately. As a comparison, modifying CSP rules through the key `csp.rules` in OSD YAML file would require a server restart.
+
+By default, this plugin is disabled. Once enabled, the plugin will first use what users have configured through `applicationConfig`. If not configured, it will check whatever CSP rules aggregated by the values of `csp.rules` from OSD YAML file and default values. If the aggregated CSP rules don't contain the CSP directive `frame-ancestors` which specifies valid parents that may embed OSD page, then the plugin will append `frame-ancestors 'self'` to prevent Clickjacking.
```
bandinib-amzn, line 7: " it will check whatever CSP rules aggregated by the values of `csp.rules` from OSD YAML file and default values" - This should be default behavior whether plugin enable or disable.

CHANGELOG.md
```diff
@@ -20,7 +21,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [BUG][MD]Fix schema for test connection to separate validation based on auth type ([#5997](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5997))
 - [Discover] Enable 'Back to Top' Feature in Discover for scrolling to top ([#6008](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6008))
 - [Discover] Fix lazy loading of the legacy table from getting stuck ([#6041](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6041))
-- [BUG][MD]Expose picker using function in data source management plugin setup([#6030](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6030)) 
+- [BUG][MD]Expose picker using function in data source management plugin setup([#6030](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6030))
```
tianleh, line 24: The code formatter removes an extra whitespace. Looks like a valid change. Thus I will keep it.

src/plugins/csp_handler/server/csp_handlers.ts
```diff
@@ -0,0 +1,86 @@
...
+ */
+function appendFrameAncestorsWhenMissing(cspHeader: string, toolkit: OnPreResponseToolkit) {
+  if (cspHeader.includes('frame-ancestors')) {
+    return toolkit.next({});
+  }
+
+  const additionalHeaders = {
+    'content-security-policy': "frame-ancestors 'self'; " + cspHeader,
+  };
+
```
bandinib-amzn, line 84: This would be breaking change, right? Also where are we handling values from YML file.

config/opensearch_dashboards.yml
tianleh: Once the feature is enabled which means enforcing Clickjacking mitigation, it will become a breaking change. We have received guidance from product team to keep the feature disabled in 2.x and only enable in 3.x https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5639#issuecomment-1930756997

So, to your question, the extra configuration is for the switch on/off.

src/plugins/csp_handler/README.md
tianleh: The sentence has a if condition where there is no configuration in the index. If the dynamic config exists, then it will take precedence.

tianleh:
Notice there is a failed workflow https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/8177983078/job/22360904964?pr=5641

This workflow was successful before merging the `main` branch changes. See the previous run 
https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/8165519624

src/plugins/csp_handler/server/csp_handlers.ts
tianleh: When adding the reference link to code, it triggered a link check failure https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/8177983078/job/22360904964?pr=5641

Thus let me directly share the reference link here
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

config/opensearch_dashboards.yml
tianleh: We will enable by default for sure in 3.x. This is to ensure that the product the org ships is secure by default. Whether we remove it could be decided later.

src/plugins/csp_handler/server/csp_handlers.ts
```diff
@@ -0,0 +1,85 @@
...
+ * @param core Context passed to the plugins `setup` method
+ * @param cspHeader The CSP header from YML
+ * @param getConfigurationClient The function provided by application config plugin to retrieve configurations
+ * @param logger The logger
+ * @returns The pre-response handler
+ */
+export function createCspRulesPreResponseHandler(
+  core: CoreSetup,
+  cspHeader: string,
+  getConfigurationClient: (scopedClusterClient: IScopedClusterClient) => ConfigurationClient,
```
bandinib-amzn, line 33: not blocker for this PR, but we expose getConfigurationClient to plugins, right? If yes, then why do they need pass `IScopedClusterClient` to get `ConfigurationClient`. After all plugin or component who implements interfaces, will build their own client.

src/plugins/csp_handler/server/plugin.ts
```diff
@@ -0,0 +1,42 @@
...
+  constructor(initializerContext: PluginInitializerContext) {
+    this.logger = initializerContext.logger.get();
+  }
+
+  public async setup(core: CoreSetup, { applicationConfig }: AppPluginSetupDependencies) {
+    core.http.registerOnPreResponse(
+      createCspRulesPreResponseHandler(
+        core,
+        core.http.csp.header,
+        applicationConfig.getConfigurationClient,
```
bandinib-amzn, line 29: Can we pass Client here, instead of passing function?

src/plugins/csp_handler/server/csp_handlers.ts
```diff
@@ -0,0 +1,85 @@
...
+
+      const currentDest = request.headers['sec-fetch-dest'];
+
+      if (!shouldCheckDest.includes(currentDest)) {
+        return toolkit.next({});
+      }
+
+      const [coreStart] = await core.getStartServices();
+
+      const myClient = getConfigurationClient(coreStart.opensearch.client.asScoped(request));
```
BionIT, line 52: can we have a better name like configClient instead of myClient?

xinruiba: +1

tianleh: updated

src/plugins/csp_handler/server/csp_handlers.ts
tianleh: Good call out! Let me elaborate more.

When an external configuration client is registered, there is no need to use `IScopedClusterClient`. We just return the external client. 
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/application_config/server/plugin.ts#L54

The reason why `IScopedClusterClient` is needed for the default OpenSearch backend is that we need to check permissions by using the `asCurrentUser` from the scoped client. This is only available from request context and this is also the reason why we need a function instead of the client. https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5641/files#r1517079679

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/application_config/server/opensearch_config_client.ts#L50

src/plugins/csp_handler/server/plugin.ts
tianleh: Replied as part of https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5641/files#r1518231638

src/plugins/csp_handler/server/csp_handlers.ts
bandinib-amzn: I understand that. suppose I am implementing my own Configuration Client in external component say Plugin. 
```
Myconfig: ConfigurationClient {
...
}
```

in plugin setup register the configuration client 
```
applicationconfig.registerConfigurationClient(Myconfig);
```

Now I'm trying to get Configuraton Client using `getConfigurationClient`.
How to call this API without IScopedClusterClient as IScopedClusterClient is required param of getConfigurationClient. How the plugin would get IScopedClusterClient? And they don't need to pass IScopedClusterClient, right? Because when they implement their own ConfiguratonClient, IScopedClusterClient doesn't even come into picture. Correct me if I'm wrong or missing something here?

src/plugins/csp_handler/server/plugin.ts
bandinib-amzn: Can't we `getConfigurationClient(coreStart.opensearch.client.asScoped(request)` call this before` core.http.registerOnPreResponse()`?

src/plugins/csp_handler/server/csp_handlers.ts
tianleh: Customers could pass null to this field when there is no such thing. The if check will be true since there will be an external client registered.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/application_config/server/plugin.ts#L54

src/plugins/csp_handler/server/plugin.ts
tianleh: This block of code is in `setup`. There is no access to `coreStart` yet. Furthermore, there is no `request` coming in yet. The code is written in this way so that we can leverage the request context inside the request handler.

src/plugins/csp_handler/server/csp_handlers.ts
bandinib-amzn: instead forcing user to pass null, I think it would be better to make optional parameter? anyway this is not in scope of this PR. This is something we might fix in ApplicationConfig. If you agree to make optional parameter, you can create an issue and resolve this conversation.

src/plugins/csp_handler/server/plugin.ts
bandinib-amzn: Got it.

bandinib-amzn (approved):

src/plugins/csp_handler/server/csp_handlers.ts
tianleh: Great suggestion!

Created the follow up issue https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6100

@bandinib-amzn

ZilongX (approved):

Issue #4462, opened 2023-06-30
[Look & Feel] Add dark mode option that respects browser settings


joshuarrrr:
Administrators should have to option to configure OpenSearch Dashboards to respect individual user browser settings for light mode/dark mode.

Currently, switching between light and dark mode theme variants can only be done via the Advanced Settings. Whichever mode is selected affects all users of that instance, regardless of their own preferences or system settings. We'd like to add a third option to that setting "Respect user browser settings", which would allow user preferences to override the default light/dark behavior.

Ideally, if users have their system or browser configured to switch between light and dark mode at various times of day, the application would adjust automatically. But due to the current implementation, we'll start with the requirement that at page load/refresh, the application gets dark/light mode preferences from the browser and uses the preferred theme accordingly.

In absence of any user/browser preferences, dark mode will be used by default.

Related:
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4454

Issue #4454, opened 2023-06-30
[Look & Feel] OpenSearch Dashboards should use dark mode themes by default


joshuarrrr:
Currently, OpenSearch Dashboards administrators can use the Advanced Settings in Stack Management to toggle the theme between light and dark mode, with light mode set by default. As a preliminary part of https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4298, we'll make light mode the default.

Users will retain the ability to override the theme to light mode.

wbeckler:
I think #4298 proposes dark mode by default, not light mode.

joshuarrrr:
@wbeckler @dagneyb Confirming - is this still planned/scoped for 2.10 or is it now targeted for 2.11?

wbeckler:
Dark mode optional in 2.10 and default in 2.11. I think the open question is whether it's "Next" theme in 2.10

wbeckler:
Just to clarify, I think the plan is to do "Next" theme in 2.10.

joshuarrrr:
> Just to clarify, I think the plan is to do "Next" theme in 2.10.

OK, I've labeled this for 2.11 and that issue for 2.11: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4677

hdhalter:
@bbarani - Can you please update the release train also to 2.13?

Alameyo:
It would be nice if by default this would take a setting from the user system/browser through `prefers-color-scheme`? And then make it possible to override making this 3 way setting.

PR #4702, opened 2023-08-09
Refactor logo usage

AMoo-Miki:
### Description

Refactor logo usage
Also:
* Move logos to a central location
* Make the loading spinner color-scheme-aware
* Recreate `OverviewPageHeader`, `HomeIcon`, `HeaderLogo`, `SolutionTitle`, `Welcome`, `Overview` tests
* Enhance `ExitFullScreenButton`, `Header` tests
* Tinified favicon assets
* Introduce a custom object snapshot serializer for testing

### Details

This change is merely to refactor logo, mark, and loading spinner imagery in one place. 

* Many components had logic of their own for parsing the branding and choosing imagery; this change pulls that logic to a central `logos`. This change also removes all of the individual unit tests that dealt with the image selection.
* The `logos` need to be made available to both `server`, for rendering the `<Template />`, and `public`, for use across the frontend; hence it is created in `common`.
* The `logos` are exposed to the frontend via the `chrome` service but depend on the branding metadata that are compiled by the `rendering` service.
* The next step is to move the parsing, validation, and serving of custom branding into a service of their own and use schema and its validation capabilities.
    * In this change, `colorScheme` is attached to logos and `branding.darkMode` is still used in some places; the branding service will take that over and expose it directly on `chrome`.
    * `branding.applicationTitle` continues to be messed up and will be taken over by the new branding service.
    * `favicon` in general is a messed up across browsers; OSD's use of favicons is also messed up. Branding's favicon is also very limiting as no user will be able to customize them with only an SVG or only a PNG. While `logos` will own the delivery of the favicon in the upcoming change, we will have to deal with the favicons differently later.

### To reviewers

While I have tried to make the tests do real testing, as opposed to just taking snapshots, there is a lot hidden in the snapshots; please start your review by looking at `getLogos` and then validate that snapshots are doing what the tests expect them to be doing.

### Issues Resolved

TBD

### Check List

- [X] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
  - [ ] `yarn test:ftr`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [X] Commits are signed per the DCO using --signoff

PR #5652, opened 2024-01-02
Feat (core): Make theme settings user-specific and user-configurable

joshuarrrr:
### Description

The purpose of this PR is to make theme settings (dark mode and theme version) settable by device, rather than cluster-wide. This allows different users (or users on different devices) to use different themes and light/dark mode. It also adds a new global nav element for global access to theming controls. These changes, however, can be toggled off in advanced settings if the previous behavior is desired/preferred.

This is an iterative step toward user-specific settings that can be saved/shared across devices, and it should unblock https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4454.

The biggest change is that it's no longer possible for any server system or component to know the theme version or dark mode settings, because the source of truth for theme settings resides client-side (stored in localstorage).

Notes for reviewers:
1. Start with the new theming documentation/diagram, which provides some background for why the implementation looks like it does.
2. Look at `core/server/rendering` to see how server-defined templates and styles are made theme/mode agnostic
3. Then see how the theme-specific logic has been moved to a client-side script under `legacy/ui/ui_render`. Unlike the existing script. This startup script is required to provide the theming info needed by the loading/error screen prior to starting the react app. Theme info not required by the loading page stays in the existing bootstrap script, which also pulls in some of the conditional theming logic that can no longer be done in the templates.
4. See how the `ui_settings` service was updated (both server and client). We define a new property `preferBrowserSetting` which indicates settings that have a local storage source of truth. The `public` client then reads and writes those settings appropriately. Note that this means even editing theme settings from the Advanced Setting page will no longer update the cluster setting - it will also just be considered device-specific.
5. Finally see how the branding prop is injected in the bootstrap process and how the new header control is registered and defined.

### Issues Resolved

Fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4462

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/1679762/4b0139a9-066d-4782-a7ee-571672bc8da8

## Testing the changes

See video above, but some scenarios to verify:
1. For new instance, user appearance menu should appear, and it should work to set either theme, color mode or both
2. When choosing "Use browser settings", the dark mode should match browser settings (after page reload)
3. For existing instances, user appearance menu should appear, but with default values that match what was previously set via advanced settings.
4. In advanced settings, theming settings are disabled with explanatory message
5. When the `Enable user control` setting is disabled, all theming behavior should work as previously (advanced settings controls enabled), with no user appearance menu.
6. Any changes made to theme settings in advanced settings menu should persist as defaults if `Enable user control` is enabled
7. Any callers of the client-side `ui_settings_client` will get the theme values that represent the state of those settings, not necessarily the settings value (for backwards compatibility). E.g. if user settings are enabled and set to "Use browser settings", consumers calling `uiClient.get('theme:darkMode')` should get back `true` if that was the browser setting at page load/application start.
8. Callers of the server-side `ui_settings_client` will always get server/saved object default values, which may not reflect app state

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -198,8 +225,10 @@ You can use \`IUiSettingsClient.get("${key}", defaultValue)\`, which will just r
     this.setLocally(key, newVal);
 
     try {
-      const { settings } = (await this.api.batchSet(key, newVal)) || {};
-      this.cache = defaultsDeep({}, defaults, settings);
+      const { settings } = this.cache[key]?.preferBrowserSetting
+        ? this.setBrowserStoredSettings(key, newVal)
+        : (await this.api.batchSet(key, newVal)) || {};
+      this.cache = defaultsDeep({}, defaults, this.getBrowserStoredSettings(), settings);
```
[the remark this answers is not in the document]
joshuarrrr: That's the correct interpretation, and the intended behavior.

> default is `darkMode: true`

Yep, and this is coming from the configs - under this new implementation, there's no longer the UI ability to set a cluster-wide default.

> I have `preferBrowserSetting: false`

This setting is irrelevant for the purpose of this client and not read. It's only purpose is to change/update the localStorage settings on bootstrap/startup

> my browser's localStorage also says `darkMode: false`

This is the source of truth for any UISetting that has the `preferBrowserSetting` flag prop set to `true`. As an implementation detail, the bootstrap/startup script are responsible for making sure these values are set (and setting them correctly based on defaults or preferred browser settings.

> My savedObjects doesn't have a `DarkMode`.

Yep - in the current implementation, we wouldn't look there. We may change this in the future so that we use/store a persistent source of truth there which can be utilized by the bootstrap/startup initialization; but the source of truth would remain the local storage value.

seraphjiang:
correct if i'm wrong, this is per browser per device. that means, if i'm using different browser on same device, the settings won't be share cross browser.

docs/theme.md
```diff
@@ -0,0 +1,131 @@
...
+        core/server->>core/server/rendering: setup rendering service
+        core/server/rendering->>core/server: provide render() method
+        core/server->>core/server: setup legacy service
+        core/server->>legacy: create legacy server
+        legacy->>legacy: start ui mixin to<br>handle special routes
+        core/server->>core/server/core_app: setup core app
+        core/server/core_app->>core/server/core_app: register default routes
+        core/server/core_app->>core/server/core_app: register static asset dir routes
+    end
+    Browser->>core/server: OSD page request (e.g. /app/home#/ )
```
BSFishy, line 54: The `#` causes rendering to get cut off

2024-01-17 14_43_35-OpenSearch-Dashboards_docs_theme md at f56628a7e24051b22308c52e6fb64d7af78002c4

src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -173,6 +178,28 @@ You can use \`IUiSettingsClient.get("${key}", defaultValue)\`, which will just r
...
+  private setBrowserStoredSettings(key: string, newVal: any) {
+    const oldSettings = this.getBrowserStoredSettings();
+    const newSettings = cloneDeep(oldSettings);
+    if (newVal === null) {
+      delete newSettings[key];
+    } else {
+      newSettings[key] = { userValue: newVal };
+    }
+    window.localStorage.setItem(`uiSettings`, JSON.stringify(newSettings));
+    return { settings: newSettings };
```
BSFishy, line 200: Why not just return `newSettings` like `getBrowserStoredSettings`?

src/core/types/ui_settings.ts
```diff
@@ -75,6 +75,7 @@ export interface UiSettingsParams<T = unknown> {
   optionLabels?: Record<string, string>;
   /** a flag indicating whether new value applying requires page reloading */
   requiresPageReload?: boolean;
+  preferBrowserSetting?: boolean;
```
BSFishy, line 78: Docs for this would be useful here

src/plugins/advanced_settings/public/header_user_theme_menu.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+  // TODO: make i18n, check all translation ids
+  // TODO: fix focus behavior
+  const appearanceContent = (
+    <div style={{ maxWidth: 300 }}>
+      <EuiCallOut color="warning">
+        These settings apply to only this user account. To change settings for the entire
+        application, {/* TODO: use navigateToApp instead? */}
+        <EuiLink
+          href={basePath.prepend('/app/management/opensearch-dashboards/settings#appearance')}
```
BSFishy, line 140: Or use `getUrlForApp`

src/legacy/ui/ui_render/bootstrap/startup.js.hbs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+var useBrowserColorScheme = window.localStorage.getItem('useBrowserColorScheme') === 'true' || false;
+
+var browserDarkMode = uiSettings['theme:darkMode'] || {};
+var browserThemeVersion = uiSettings['theme:version'] || {};
+
+var rawDarkMode = typeof browserDarkMode.userValue !== 'boolean' ? {{configDarkMode}} : browserDarkMode.userValue;
+
+debugger;
```
BSFishy, line 17: Was this left in intentionally?

src/core/public/ui_settings/ui_settings_client.ts
BSFishy: Just noticed this probably just matches the return for `this.api.batchSet`

joshuarrrr:
> correct if i'm wrong, this is per browser per device. that means, if i'm using different browser on same device, the settings won't be share cross browser.

Yes, that's exactly correct. That's one reason I chose `preferBrowserSetting` for the new property, to emphasize that it's not a true user-preference. The intent here is to unblock dark mode by default (https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4454) by giving all users an easy escape hatch.

It also lays the groundwork for future iterations, where we have an actual user concept in OSD core (rather than only in `security-dashboards`), and user-specific settings saved objects (and potentially other preference/settings layers, such as workspaces). @AMoo-Miki and I have discussed a potential design for that scenario where a user settings saved object and local storage both keep a timestamp of the most recent update and then sync as part of the startup proces.

docs/theme.md
joshuarrrr: wow - well spotted 👁️! I'll update on the next revision.

src/legacy/ui/ui_render/bootstrap/startup.js.hbs
joshuarrrr: nope 🤦 - will remove.

src/plugins/advanced_settings/public/header_user_theme_menu.tsx
joshuarrrr: Right, thanks for the reminder! I think this message may be removed altogether, now that the advanced settings page actually does the exact same thing (update for that browser, not for the whole cluster).

src/core/types/ui_settings.ts
joshuarrrr: agreed, will add.

src/core/public/ui_settings/ui_settings_client.ts
joshuarrrr: Yeah - this was not really the ideal shape, but I did it this way to avoid some duplication. But I'm definitely open for ways to clean this up more and make it more readable and less weird.

wbeckler:
Does this add a setting to the advanced settings page, in addition to offering a settings widget in the theme submenu in the top right corner? If so, it could be confusing to mix domain-wide settings and browser/user-specific settings in the same page that is only accessible to admins. To avoid this confusion, I would propose that these settings do not get listed on the advanced settings page.

BSFishy:
I agree with @wbeckler. It was certainly confusing to me trying to figure out the semantics and difference between changing these settings in advanced settings and the theme submenu. I feel like advanced settings should almost blanket filter out `preferBrowserSetting`, then if we find ourselves using `preferBrowserSetting` a lot, we could add a "User settings" page that only has those settings and just have some sort of callout at the top that says something like "These settings are applied on a per-browser basis. Changes to them won't apply to any other devices."

src/plugins/advanced_settings/public/header_user_theme_menu.tsx
```diff
@@ -0,0 +1,206 @@
...
+    </EuiHeaderSectionItemButton>
+  );
+
+  // TODO: make i18n, check all translation ids
+  // TODO: fix focus behavior
+  const appearanceContent = (
+    <div style={{ maxWidth: 300 }}>
+      <EuiCallOut color="warning">
+        These settings apply to only this user account. To change settings for the entire
+        application, {/* TODO: use navigateToApp instead? */}
```
ashwin-pc, line 138: yes we should use `navigateToApp` instead

ashwin-pc (commented):
The change looks good overall. have a few comments and questions:

1. Why didnt you just make the automatic option a part of the UI setting as well? That part was a lil confusing to understand
2. The updated UI Settings logic could definitely use some documentation (Not a blocker but will be invaluable in future)
3. Looks like the wording here dont match your implementation. browser settings can never be set at the cluster level right?

Screenshot 2024-02-06 at 1 53 31 AM

docs/theme.md
AMoo-Miki: @joshuarrrr would you have time to address this?

src/core/public/ui_settings/ui_settings_client.ts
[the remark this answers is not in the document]
AMoo-Miki: > That's the correct interpretation, and the intended behavior.

@joshuarrrr, when `preferBrowserSetting: false`, I would expect nothing from the browser to be honored at all; i would expect that the default `darkMode: true` will be applied.

docs/theme.md
kavilla: @AMoo-Miki i have created an issue for non-blockers to which i added this comment tagged for 2.13: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5831

src/core/server/rendering/rendering_service.tsx
```diff
@@ -94,31 +94,18 @@ export class RenderingService {
...
         const brandingAssignment = await this.assignBrandingConfig(
-          darkMode,
           opensearchDashboardsConfig as OpenSearchDashboardsConfigType
         );
 
         const metadata: RenderingMetadata = {
           strictCsp: http.csp.strict,
           uiPublicUrl,
           bootstrapScriptUrl: `${basePath}/bootstrap.js`,
+          startupScriptUrl: `${basePath}/startup.js`,
```
kavilla, line 106: i feel like we desperately needed this and it allows us to do some cools tuff

src/legacy/ui/ui_render/ui_render_mixin.js
```diff
@@ -92,47 +96,21 @@ export function uiRenderMixin(osdServer, server, config) {
       );
       const uiSettings = osdServer.newPlatform.start.core.uiSettings.asScopedToClient(soClient);
 
-      const darkMode =
+      const configDarkMode =
         !authEnabled || request.auth.isAuthenticated
```
kavilla, line 100: i feel like this logic isn't hooked into from the security plugin

kavilla: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5831

docs/theme.md
```diff
@@ -0,0 +1,131 @@
...
+3. Theme computed SASS variables as JSON (e.g. `@elastic/eui/dist/eui_theme_dark.json`). Consumed by [/packages/osd-ui-shared-deps/theme.ts](/packages/osd-ui-shared-deps/theme.ts) and made available to other components via the mode and theme aware `euiThemeVars`. In general, these should not be consumed by any other component directly.
+4. Theme type definition file for SASS variables as JSON (e.g. `@elastic/eui/dist/eui_theme_dark.json.d.ts`)
+
+Note that all of these artifacts should ideally only be imported or used directly in one place (by `osd-ui-shared-deps`).
+
+In addition to these artifacts, OpenSearch Dashboards also makes heavy use of the theme SASS variables and mixins as defined in the source files (e.g. `@elastic/eui/src/theme_dark.scss`).
+
+### Theme definitions in OpenSearch Dashboards
+
+1. Theme tags are defined in [/packages/osd-optimizer/src/common/theme_tags.ts](/packages/osd-optimizer/src/common/theme_tags.ts) corresponding to each mode (light/dark) of each OUI theme.
```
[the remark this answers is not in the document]
kavilla: added this to: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5831

src/core/public/ui_settings/ui_settings_client.ts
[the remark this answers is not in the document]
kavilla: added to: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5831

src/core/public/ui_settings/ui_settings_client.ts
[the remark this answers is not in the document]
kavilla: is this a blocker @AMoo-Miki ?

src/legacy/ui/ui_render/bootstrap/startup.js.hbs
kavilla: will you have time for this? @joshuarrrr

src/core/types/ui_settings.ts
kavilla: add to https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5831

src/plugins/advanced_settings/public/header_user_theme_menu.tsx
kavilla: added to check https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5652/files#r1460030358

cc: @kgcreative

src/core/public/ui_settings/ui_settings_client.ts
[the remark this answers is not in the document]
AMoo-Miki: Rocky, I think the unexpected result will confuse users enough to make it a problem.

src/core/server/rendering/views/fonts.tsx
```diff
@@ -532,9 +531,7 @@ export const Fonts: FunctionComponent<Props> = ({ url, theme }) => {
       url('${url}/fonts/inter_ui/Inter-UI-italic.var.woff2') format('woff2');
   }
   */
-  const fontText = theme === 'v7' ? interUi : sourceSans3;
-  const fontCode = theme === 'v7' ? roboto : sourceCodePro;
-  const fontsDefinitionRules = [fontText, fontCode]
+  const fontsDefinitionRules = [interUi, sourceSans3, roboto, sourceCodePro]
```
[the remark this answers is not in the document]
AMoo-Miki: @ashwin-pc I believe the goal of this change was to not bother filtering the definitions of the fonts here and allow each theme to use which ever they want. It doesn't add too much burden considering this would be compressed in transit to the browser but I don't like it much.

src/core/server/rendering/views/fonts.tsx
```diff
@@ -557,36 +554,11 @@ export const Fonts: FunctionComponent<Props> = ({ url, theme }) => {
...
-
-    code, pre, kbd, samp {
-      font-family: var(--font-code);
-    }
-    html, input, textarea, select, button {
-      font-family: var(--font-text);
-    }
-
-  `;
-
```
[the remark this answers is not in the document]
AMoo-Miki: These were moved to startup.js.hbs to be injected via js. However, I am concerned about FOUT.

src/legacy/ui/ui_render/bootstrap/startup.js.hbs
```diff
@@ -0,0 +1,106 @@
...
+  rawDarkMode = window.matchMedia('(prefers-color-scheme: dark)').matches;
+}
+
+var rawThemeVersion = browserThemeVersion.userValue || '{{configThemeVersion}}'
+
+// TODO: source of truth for mapping should be elsewhere
+var darkMode = rawDarkMode ? 'dark' : 'light';
+var themeVersion = rawThemeVersion === 'v7' ? 'v7' : 'v8';
+
+console.log(themeVersion, darkMode, window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light');
```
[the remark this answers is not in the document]
AMoo-Miki: removed

src/core/server/rendering/views/fonts.tsx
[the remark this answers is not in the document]
joshuarrrr: It's not FOUT because that's render blocking JS (although, it turns out the bootstrap.js is also effectively render blocking, because the page content is set to `display:none` until it finishes)

src/core/server/rendering/rendering_service.tsx
joshuarrrr: It's handy, but we definitely want to change how it's built so that we can use our proper TS/JS tooling, rather than writing JS in a big template.

src/legacy/ui/ui_render/ui_render_mixin.js
joshuarrrr: Yeah, that's a good idea, although all I changed here was the naming.

src/plugins/advanced_settings/public/header_user_theme_menu.tsx
```diff
@@ -0,0 +1,206 @@
...
+    {
+      value: 'dark',
+      text: 'Dark mode',
+    },
+    {
+      value: 'automatic',
+      text: 'Use browser settings',
+    },
+  ];
+  const prefersAutomatic =
```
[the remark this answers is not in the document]
joshuarrrr: So, the existing type signature for `theme:darkMode` is a boolean that indicates whether the current mode is light or dark. I decided that the safest and easiest approach was not to break/change that contract, because that's what all `.get` callers actually care about. Instead, preferring automatic is an alternative way to _set_ that setting, not a setting itself.

src/plugins/advanced_settings/public/header_user_theme_menu.tsx
joshuarrrr: This callout needs to be removed now anyway, because the behavior in both places is the same.

kavilla (dismissed):

kavilla:
@ananzh will check on following up on this.

src/core/server/rendering/views/__snapshots__/template.test.tsx.snap
joshuarrrr, line 1: This diff is useful for reviewers to review, as it demonstrates some of the changes.

src/core/server/rendering/views/styles.tsx
```diff
@@ -177,6 +155,10 @@ export const Styles: FunctionComponent<Props> = ({ theme, darkMode }) => {
             max-width: 100%;
           }
 
+          .darkOnly, .lightOnly {
+            display: none;
+          }
+
```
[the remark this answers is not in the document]
joshuarrrr: These static styles are the initial rendering state - the display value will be overridden/unset as part of the render-blocking `startup.js` once dark mode has been determined. See L61-63 in `src/legacy/ui/ui_render/bootstrap/startup.js.hbs`

src/core/server/ui_settings/settings/theme.ts
```diff
@@ -43,8 +43,10 @@ export const getThemeSettings = (): Record<string, UiSettingsParams> => {
         defaultMessage: `Enable a dark mode for the OpenSearch Dashboards UI. A page refresh is required for the setting to be applied.`,
       }),
       requiresPageReload: true,
+      preferBrowserSetting: true,
       category: ['appearance'],
       schema: schema.boolean(),
+      type: 'boolean',
```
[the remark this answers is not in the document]
joshuarrrr: It seems like usage is inconsistent, but there are code paths that only check for `type`, so I think it's more correct to have both.

src/core/server/ui_settings/settings/theme.ts
```diff
@@ -34,6 +34,19 @@ import { UiSettingsParams } from '../../../types';
...
 export const getThemeSettings = (): Record<string, UiSettingsParams> => {
   return {
+    'theme:enableUserControl': {
+      name: i18n.translate('core.ui_settings.params.enableUserControlTitle', {
+        defaultMessage: 'Enable user control',
+      }),
+      value: true,
+      description: i18n.translate('core.ui_settings.params.enableUserControlText', {
+        defaultMessage: `Enable users to control theming and dark or light mode via "Appearance" control in top navigation. When true, those settings can no longer be set globally by administrators.`,
+      }),
```
joshuarrrr, line 44: @kgcreative Some new copy to review. One thing to consider is the fact that all UI settings within the same group ("Appearance", in this case), will be listed alphabetically by the key `theme:enableUserControl`. Ideally, I think this setting would precede the other two `theme:` prefixed settings on the advanced settings page. So maybe just `enableUserControl` instead, even though it currently just controls theme-related settings?

src/core/types/ui_settings.ts
joshuarrrr: I'm not at all convinced this is a good name, but unfortunately anything "user"-related here would arguably be even more confusing, because it's already in use.

src/plugins/advanced_settings/public/management_app/components/field/field.tsx
```diff
@@ -443,6 +444,15 @@ export class Field extends PureComponent<FieldProps> {
           />
         </EuiText>
       );
+    } else if (setting.preferBrowserSetting) {
+      return (
+        <EuiText size="xs">
+          <FormattedMessage
+            id="advancedSettings.field.browserSettingHelpText"
+            defaultMessage="This setting is overridden by user or browser preferences."
```
joshuarrrr, line 452: @kgcreative More copy to review - this explains why these advanced settings are disabled when `theme:enableUserControl`is set to `true`.

joshuarrrr:
For the failing tests:

```
     │1)    dashboard app
     │       using current data
     │         dashboard snapshots
     │           compare TSVB snapshot:
     │
     │      Error: expected 0.021522 to be below 0.02
     │       at Assertion.assert (packages/osd-expect/expect.js:111:11)
     │       at Assertion.lessThan.Assertion.below (packages/osd-expect/expect.js:347:8)
     │       at Function.lessThan (packages/osd-expect/expect.js:542:15)
     │       at Context.<anonymous> (test/functional/apps/dashboard/dashboard_snapshots.js:78:39)
     │       at processTicksAndRejections (node:internal/process/task_queues:95:5)
     │       at Object.apply (packages/osd-test/src/functional_test_runner/lib/mocha/wrap_function.js:95:16)
```

This looks like a flaky test to me - has it failed on other PRs? I don't see how theming changes could make this value round incorrectly. And the CI test group 3 passed on the Windows run.

```
│1)    visualize app
     │       
     │         OpenSearch Dashboards branding configuration
     │           should render home page
     │             in default mode
     │               with customized logo in header bar:
     │
     │      Error: expected 'https://opensearch.org/ASSETS/BRAND/SVG/LOGO/OPENSEARCH_LOGO_DARKMODE.SVG' to equal 'https://opensearch.org/ASSETS/BRAND/SVG/LOGO/OPENSEARCH_LOGO_DEFAULT.SVG'
     │       at Assertion.assert (packages/osd-expect/expect.js:111:11)
     │       at Assertion.equal (packages/osd-expect/expect.js:238:8)
     │       at GlobalNav.logoExistsOrFail (test/functional/services/global_nav.ts:90:44)
     │       at processTicksAndRejections (node:internal/process/task_queues:95:5)
     │       at Context.<anonymous> (test/functional/apps/visualize/_custom_branding.ts:128:11)
     │       at Object.apply (packages/osd-test/src/functional_test_runner/lib/mocha/wrap_function.js:95:16)
```

@AMoo-Miki Can you help me look at this? As far as I can tell the behavior of this PR is actually what we intend/want to happen, since merging https://github.com/opensearch-project/OpenSearch-Dashboards/pull/4702. So I think the test case itself is wrong, but it's puzzling that it has been passing in its current state. Just want your confirmation before updating the test case to match the behavior (where we always use the dark mode version of the logo since it's always on a dark background component)

AMoo-Miki:
@joshuarrrr 

```
     │1)    dashboard app
     │       using current data
     │         dashboard snapshots
     │           compare TSVB snapshot:
     │
     │      Error: expected 0.021522 to be below 0.02
```
> This looks like a flaky test to me - has it failed on other PRs? I don't see how theming changes could make this value round incorrectly. And the CI test group 3 passed on the Windows run.

This is the result of visual diff. It is saying that too much has changed. The solution is to update the base image after validating that the visual changes are desired.


```
     │1)    visualize app
     │       
     │         OpenSearch Dashboards branding configuration
     │           should render home page
     │             in default mode
     │               with customized logo in header bar:
     │
     │      Error: expected 'https://opensearch.org/ASSETS/BRAND/SVG/LOGO/OPENSEARCH_LOGO_DARKMODE.SVG'
     |      to equal 'https://opensearch.org/ASSETS/BRAND/SVG/LOGO/OPENSEARCH_LOGO_DEFAULT.SVG'
```
> As far as I can tell the behavior of this PR is actually what we intend/want to happen, since merging https://github.com/opensearch-project/OpenSearch-Dashboards/pull/4702. So I think the test case itself is wrong, but it's puzzling that it has been passing in its current state.

I too believe the test is wrong. The expanded header shows up with a dark background irrespective of the color scheme and because of that, we will always have a dark logo there.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/c2c4f53f140a655705e0e8830ea14e8d958a8de9/src/core/public/chrome/ui/header/header.tsx#L131

A collapsed header uses the theme's background color which is why the next test passes.

joshuarrrr:
> A collapsed header uses the theme's background color which is why the next test passes.

AMoo-Miki (approved):

ananzh (approved):

seraphjiang:
do we have functional test automation for test coverage?

Issue #5656, opened 2024-01-04
[BUG] Datasource testing connection don't validate endpoints with path


xinruiba:
### Task List 
- [ ] https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5707
- [ ] https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5706

**Describe the bug**

If cx adds endpoint with path appended (for example: `https://OpenSearchEndpoint/_dashboards`) as a datasource, connection test will pass, but cx can't create index pattern or add sample data; 

We may need to validate endpoint's path is empty when testing connection

**To Reproduce**
1. Go to "Dashboards Management" ---> "Data Sources"
2. Click on "Create data source connection"
3. Fill in the form with DataSource Title. eg: "test"
4. Fill in the form with DataSource Endpoint followed with a path name. eg: "https://OpenSearchEndPoint/_dashboard"
5. Click on "Test Connection" the test will pass
6. Go to "Home" ---> "Add Sample Data" ---> "Data Source" choose Test --->
7. Error shows up: sample data not able to be added

<img width="1772" alt="Screenshot 2023-11-30 at 8 17 39 PM" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/139305463/8373fe31-f341-4cdb-9622-ef143fe284be">

**Expected behavior**
DataSource should not be able to created with an endpoint appended with a path.
We should do URL validation to make sure not path name appended to a datasource endpoint.

**OpenSearch Version**
2.11

**Dashboards Version**
2.11

**Plugins**

Please list all plugins currently enabled.
N/A

**Screenshots**

**Host/Environment (please complete the following information):**
 - OS: Linux
 - Browser and version: Chrome Version 120.0.6099.129

**Additional context**

xinruiba:
I have a proposal to fix this issue. Creating a PR now.

PR #5663, opened 2024-01-05
Fix Datasource testing connection don't validate endpoints with path

### LINKED ISSUE #5702: [BUG] Multiple datasource test connection didn't test correctly
**Describe the bug**

The multiple datasource test connection didn't test correctly in some cases, like as "Username & Password", and serverless cases

**To Reproduce**
1. Go to 'Data source'
2. Click on 'Create data source'
3. Fill out endpoint URL "https://google.com"
4. Choose "Username & Password"
5. Fill out any Username and Password
6. Click 'Test connection'
7. It is showing up "Connecting to the endpoint using the provided authentication method was successful."

**Expected behavior**
It should throw error that "failed to connection"

**OpenSearch Version**
2.11

**Dashboards Version**
2.11

**Plugins**

N/A

**Screenshots**

image

 - OS: MacOS
 - Browser and version: firefox 115.5.0esr

**Additional context**

xinruiba:
### Task List 
- [x] https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5707
- [ ] https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5706
### Description
Test connection should only success when opensearch-client response status code is 200 and 
1. response body not empty (with serverless collection) 
2. response body have cluster_name field (with server cluster)
Please check this ticket for more details https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5707

## 01/19/2024 Update:
Create an initial fix based on discussion, and more details can be found under this issue: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5707

## 01/09/2024 Update:
Discussed with team (checkout: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5663#issuecomment-1883594471)

We are going to take a further investigation and we put the first proposal under the issue description:
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5656 

CC @Flyingliuhub @seraphjiang @bandinib-amzn please feel free to leave comments~ Thanks
### Issues Resolved

fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5656

Fixes #5702 

## Screenshot

**Before:**
Test connection with endpoint: "https://google.com" will pass which is unexpected

Before-TestGoogle

**After:**
Test connection with endpoint: "https://google.com" will be failed to connect with expected response message:

After-TestGoogle

Also verified endpoint with path name will fail test connection:

After-WithPath

Validate cluster endpoint still able to pass connection testing:
After-NoPath

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


src/plugins/data_source_management/public/components/utils.ts
Flyingliuhub, line 1: can you add the change to CHANGELOG.md too? Thanks

xinruiba: Sure, working on it. Thanks for the comment~

Flyingliuhub (approved):

src/plugins/data_source_management/public/components/utils.ts
xinruiba: CHANGELOG.md updated. Resolving this comment. Thanks

Flyingliuhub (dismissed):

src/plugins/data_source_management/public/components/validation/datasource_form_validation.test.ts
```diff
@@ -49,6 +53,16 @@ describe('DataSourceManagement: Form Validation', () => {
...
     });
+    test('should fail validation when endpoint path is not empty', () => {
+      form.endpoint = mockDataSourceAttributesWithInvalidEndpoint.endpoint;
+      const result = performDataSourceFormValidation(form, [], '');
+      expect(result).toBe(false);
+    });
+    test('should NOT fail validation when endpoint path is empty', () => {
+      form.endpoint = mockDataSourceAttributesWithValidEndpoint.endpoint;
+      const result = performDataSourceFormValidation(form, [], '');
+      expect(result).toBe(false);
```
bandinib-amzn, line 64: The expected result here should be true.

bandinib-amzn (changes requested):

bandinib-amzn:
Also I think it would be great if we can run this fix by UX team to see if it is ok from users perspective.

src/plugins/data_source_management/public/components/validation/datasource_form_validation.test.ts
xinruiba: Good catch~

seraphjiang:
> Also I think it would be great if we can run this fix by UX team to see if it is ok from users perspective.

@kgcreative who could help work with us on this.

kgcreative:
I'm happy to review async here. To that end, i have a few questions before i make a recommendation.

1. When we say "OpenSearch Endpoint" we mean the cluster endpoint, and not an OSD endpoint, correct? If so, then "test connection" should handshake with the endpoint, and there should be some sort of response that validates that this is indeed the correct endpoint.

2. when it comes to there being a path, is there a legitimate usecase for an endpoint having a path in it? for example: `https://www.bank.com/opensearch` and if so, how do we differentiate between that and say `https://opensearch.bank.com`

3. I don't think we should disable the `test endpoint` button, i think if there is an appended path, `test connection` should return an error

src/plugins/data_source_management/public/components/validation/datasource_form_validation.test.ts
```diff
@@ -49,6 +53,18 @@ describe('DataSourceManagement: Form Validation', () => {
...
     });
+    test('should fail validation when endpoint path is not empty', () => {
+      form.endpoint = mockDataSourceAttributesWithInvalidEndpoint.endpoint;
+      const result = performDataSourceFormValidation(form, [], '');
+      expect(result).toBe(false);
+    });
+    test('should NOT fail validation when endpoint path is empty', () => {
+      form.auth.type = AuthType.NoAuth;
+      form.title = 'test';
+      form.endpoint = mockDataSourceAttributesWithValidEndpoint.endpoint;
```
wanglam, line 64: It seems the `mockDataSourceAttributesWithValidEndpoint.endpoint` is not empty here, shall we update the unit tests description?

xinruiba:
> I'm happy to review async here. To that end, i have a few questions before i make a recommendation.
> 
> 1. When we say "OpenSearch Endpoint" we mean the cluster endpoint, and not an OSD endpoint, correct? If so, then "test connection" should handshake with the endpoint, and there should be some sort of response that validates that this is indeed the correct endpoint.

**Answer:** 
Yes, it's cluster endpoint.

> 2. when it comes to there being a path, is there a legitimate usecase for an endpoint having a path in it? for example: `https://www.bank.com/opensearch` and if so, how do we differentiate between that and say `https://opensearch.bank.com`

**Answer:** 
For now, we don't have a user case that supporting an endpoint having a path in it (@bandinib-amzn @Flyingliuhub rectify me if I'm wrong). In the future will we supported endpoint with path? I need to have more discussion with my team. 

> 3. I don't think we should disable the `test connection` button, instead, i think if there is an appended path, `test connection` should return an error, at which point the endpoint URL input should change to failed (red)

**Answer:** 
Thanks for the comment. I think your proposed user experience is:

Before user clicking the "Test Connection" Button:
1. User fill in the datasource creation form (with an invalidate cluster endpoint)
2. Test connect button still available 
3.  Endpoint URL input still green

After user clicking the "Test Connection" Button:
1. Banner shows up to say connection failed
2. Endpoint URL input becomes red

Is that correct?

For this proposal, me and @Flyingliuhub have a discussion and @Flyingliuhub already provide a response. 
We will have a further discussion and keep you updated.

Thanks for the comment.

kgcreative:
I think the green underline is just a focus indicator, so when you tab out of the field it should return to it's unfocused state.

src/plugins/data_source_management/public/components/validation/datasource_form_validation.test.ts
xinruiba: Thanks for the comment. Working on updating.

src/plugins/data_source/server/routes/data_source_connection_validator.ts
```diff
@@ -14,10 +14,21 @@ export class DataSourceConnectionValidator {
...
-      return await this.callDataCluster.info();
+      if (
+        this.dataSourceAttr.auth?.credentials?.service === SigV4ServiceName.OpenSearchServerless
+      ) {
+        validationResponse = await this.callDataCluster.cat.indices();
+      } else {
+        validationResponse = await this.callDataCluster.info();
+      }
+
+      if (validationResponse?.statusCode === 200) {
```
Flyingliuhub, line 27: Can we add one more check here if have clustername or version for opensearch, and body if empty for serverless

xinruiba: That's a good idea. I will update the validation logic into following conditions:

1. Server cluster:
test connection pass when: (response.statuscode == 200 && response.body.cluster_name.isNotEmpty())

2. Serverless collection:
test connection pass when: (response.statuscode == 200 && response.body.isNotEmpty())

cc @Flyingliuhub @bandinib-amzn

bandinib-amzn (approved):

src/plugins/data_source_management/public/components/validation/datasource_form_validation.test.ts
xinruiba: This test case is not needed. Resolving this comment. Thanks~

bandinib-amzn:
Failed cypress test is not related to this change, so we can ignore.

Flyingliuhub (approved):

PR #4612, opened 2023-07-24
[Decouple] Allow plugin manifest config to define semver compatible OpenSearch plugin and verify if it is installed on the cluster

manasvinibs:
### Description

This PR is an enhancement to the already existing plugin manifest ability to allow defining backend OpenSearch plugin dependencies. As part of decoupling Dashboards with OpenSearch Dashboards plugins and then with OpenSearch, we need a way to establish Dashboards plugin dependency with its counter part. 
This PR adds following changes -
* Modifying existing manifest config key `requiredOpenSearchPlugins` to define list of OpenSearch Plugin object and its version compatibility with that plugin.
* In the first iteration we can define version range in semver format only. 
* Compatible version defined in the manifest should be a valid version range.
* Validates if version compatible plugin is installed on the cluster or not.
* In this initial implementation, we are only logging the warning if plugin is not available and does not fail to start.

In the follow-up PRs,
* Add ability to disable Dashboards plugin when its counterpart is not installed on the cluster.
* Export function/API that plugins can extend to decide what they want to do when their OpenSearch counterpart plugin is not installed or is incompatible. 

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4611

### Check List

- [X] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
  - [ ] `yarn test:ftr`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [X] Update [CHANGELOG.md](./../CHANGELOG.md)
- [X] Commits are signed per the DCO using --signoff

PR #5675, opened 2024-01-09
[PM] Enhance single version requirements imposed during bootstrapping

AMoo-Miki:
Add `--single-version` switch to `bootstrap` to manipulate the behavior of single-version validation

### Check List

- [X] All tests pass
  - [ ] `yarn test:jest`
- [X] New functionality has been documented.

### DIFF
--- a/packages/osd-pm/README.md
+++ b/packages/osd-pm/README.md
@@ -141,6 +141,40 @@ yarn osd
 Bootstrapping also calls the `osd:bootstrap` script for every included project.
 This is intended for packages that need to be built/transpiled to be usable.
 
+#### Single-version validation
+
+Bootstrapping, by default, applies a `strict` single-version validation where it requires all the dependencies defined
+more than once to have the same version-range in the `package.json` files of Dashboards, packages, and plugins. If a
+violation is identified, bootstrapping terminates with an error. This behavior can be controlled using the 
+`--single-version` switch. Using any switch other than the default can result in the installation of versions of the
+dependencies that were never tested and this could lead to unexpected results.
+
+```
+yarn osd bootstrap --single-version=loose
+```
+In `loose` mode, bootstrapping reconciles the various versions installed as a result of having multiple ranges for a 
+dependency, by choosing one that satisfies all said ranges. Even though installing the chosen version updates the 
+`yarn.lock` files, no `package.json` changes would be needed.
+
+```
+yarn osd bootstrap --single-version=force
+```
+In `force` mode, bootstrapping acts like `loose` for each dependency. If despite that, a suitable version was not found,
+it switches to behave like the `brute-force` mode (see below).
+
+```
+yarn osd bootstrap --single-version=brute-force
+```
+In `brute-force` mode, bootstrapping chooses the newest of the various versions installed, irrespective of whether it 
+satisfies any of the ranges. Installing the chosen version, bootstrapping updates the `yarn.lock` files and applies a 
+range, in the form of `^<version>`, to all `package.json` files that declared the dependency.
+
+```
+yarn osd bootstrap --single-version=ignore
+```
+In `ignore` mode, bootstrapping behaves very similar to the `strict` mode by showing errors when different ranges of a 
+package are marked as dependencies, but without terminating.
+
 ### Running scripts
 
 Some times you want to run the same script across multiple packages and plugins,
--- a/packages/osd-pm/src/utils/project.ts
+++ b/packages/osd-pm/src/utils/project.ts
@@ -28,8 +28,8 @@
  * under the License.
  */
 
-import fs from 'fs';
-import Path from 'path';
+import { existsSync, unlinkSync } from 'fs';
+import { resolve, relative } from 'path';
 import { inspect } from 'util';
 
 import { CliError } from './errors';
@@ -43,6 +43,7 @@ import {
 } from './package_json';
 import {
   installInDir,
+  patchFile,
   runScriptInPackage,
   runScriptInPackageStreaming,
   yarnWorkspacesInfo,
@@ -97,9 +98,9 @@ export class Project {
     this.json = Object.freeze(packageJson);
     this.path = projectPath;
 
-    this.packageJsonLocation = Path.resolve(this.path, 'package.json');
-    this.nodeModulesLocation = Path.resolve(this.path, 'node_modules');
-    this.targetLocation = Path.resolve(this.path, 'target');
+    this.packageJsonLocation = resolve(this.path, 'package.json');
+    this.nodeModulesLocation = resolve(this.path, 'node_modules');
+    this.targetLocation = resolve(this.path, 'target');
 
     this.version = this.json.version;
     this.productionDependencies = this.json.dependencies || {};
@@ -130,7 +131,7 @@ export class Project {
     if (dependentProjectIsInWorkspace) {
       expectedVersionInPackageJson = project.json.version;
     } else {
-      const relativePathToProject = normalizePath(Path.relative(this.path, project.path));
+      const relativePathToProject = normalizePath(relative(this.path, project.path));
       expectedVersionInPackageJson = `link:${relativePathToProject}`;
     }
 
@@ -168,7 +169,7 @@ export class Project {
    * instead of everything located in the project directory.
    */
   public getIntermediateBuildDirectory() {
-    return Path.resolve(this.path, this.getBuildConfig().intermediateBuildDirectory || '.');
+    return resolve(this.path, this.getBuildConfig().intermediateBuildDirectory || '.');
   }
 
   public getCleanConfig(): CleanConfig {
@@ -196,14 +197,14 @@ export class Project {
 
     if (typeof raw === 'string') {
       return {
-        [this.name]: Path.resolve(this.path, raw),
+        [this.name]: resolve(this.path, raw),
       };
     }
 
     if (typeof raw === 'object') {
       const binsConfig: { [k: string]: string } = {};
       for (const binName of Object.keys(raw)) {
-        binsConfig[binName] = Path.resolve(this.path, raw[binName]);
+        binsConfig[binName] = resolve(this.path, raw[binName]);
       }
       return binsConfig;
     }
@@ -262,6 +263,51 @@ export class Project {
   }
 
   /**
+   * Install a specific version of a dependency and update the package.json.
+   * When a range is not specified, ^<version> is used. The range is then
+   * placed in the package.json with intentionally no validation.
+   */
+  public async installDependencyVersion(
+    depName: string,
+    version: string,
+    dev: boolean = false,
+    range?: string
+  ) {
+    log.info(`[${this.name}] running yarn to install ${depName}@${version}`);
+
+    log.write('');
+
+    const rangeToUse = range || `^${version}`;
+
+    const extraArgs = [`${depName}@${version}`];
+    if (dev) extraArgs.push('--dev');
+
+    if (this.isWorkspaceProject) {
+      await installInDir(this.path);
+    } else {
+      await installInDir(this.path, extraArgs, true);
+    }
+
+    log.info(`[${this.name}] updating manifests with ${depName}@${rangeToUse}`);
+
+    await patchFile(
+      this.packageJsonLocation,
+      `"${depName}": "${version}"`,
+      `"${depName}": "${rangeToUse}"`
+    );
+    // The lock-file of workspace packages are symlinked to the root project's and editing the one in the project suffices
+    await patchFile(
+      resolve(this.path, 'yarn.lock'),
+      `${depName}@${version}`,
+      `${depName}@${rangeToUse}`
+    );
+
+    log.write('');
+
+    await this.removeExtraneousNodeModules();
+  }
+
+  /**
    * Yarn workspaces symlinks workspace projects to the root node_modules, even
    * when there is no depenency on the project. This results in unnecicary, and
    * often duplicated code in the build archives.
@@ -283,13 +329,13 @@ export class Project {
 
     unusedWorkspaces.forEach((name) => {
       const { dependencies, devDependencies } = this.json;
-      const nodeModulesPath = Path.resolve(this.nodeModulesLocation, name);
+      const nodeModulesPath = resolve(this.nodeModulesLocation, name);
       const isDependency = dependencies && dependencies.hasOwnProperty(name);
       const isDevDependency = devDependencies && devDependencies.hasOwnProperty(name);
 
-      if (!isDependency && !isDevDependency && fs.existsSync(nodeModulesPath)) {
+      if (!isDependency && !isDevDependency && existsSync(nodeModulesPath)) {
         log.debug(`No dependency on ${name}, removing link in node_modules`);
-        fs.unlinkSync(nodeModulesPath);
+        unlinkSync(nodeModulesPath);
       }
     });
   }
--- a/packages/osd-pm/src/utils/validate_dependencies.ts
+++ b/packages/osd-pm/src/utils/validate_dependencies.ts
@@ -29,9 +29,12 @@
  */
 
 // @ts-expect-error published types are useless
-import { stringify as stringifyLockfile } from '@yarnpkg/lockfile';
+import { stringify as stringifyLockfile, parse as parseLockFile } from '@yarnpkg/lockfile';
 import dedent from 'dedent';
 import chalk from 'chalk';
+import path from 'path';
+import { readFileSync } from 'fs';
+import { satisfies, rcompare } from 'semver';
 
 import { writeFile } from './fs';
 import { OpenSearchDashboards } from '../utils/opensearch_dashboards';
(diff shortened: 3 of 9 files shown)


peterzhuamazon:
As of now, build repo during building process, will clone OSD, then pull each plugin into the plugins directory of OSD, before running plugin level bootstrap and build. There is no cleanup of the plugin repos within plugins directory, after the build is completed. We think there are two ways to solve the error on build_workflow code now:

1. Add `yarn osd bootstrap --single-version=loose` to every build script including core and plugins if needed.
2. In build_workflow code, do a cleanup of plugins directory after one plugin completed the build, before next plugin repo get cloned there.

Thanks.

ZilongX:
Nice move towards version decoupling ! So with this newly introduced loose options, we'll be able to avoid the version-bump-only releases for bunch of components correct ? Also given this will ease the building check during bootstrap, would there be any concern during run time ? Usually a match major.minor should work seamlessly though (say osd core 2.11.1 with plugin 2.11.2)

packages/osd-pm/src/utils/project.ts
```diff
@@ -261,6 +262,49 @@ export class Project {
...
+   * When a range is not specified, ^<version> is used. The range is then
+   * placed in the package.json with intentionally no validation.
+   */
+  public async installDependencyVersion(
+    depName: string,
+    version: string,
+    dev: boolean = false,
+    range?: string
+  ) {
+    log.info(`[${this.name}] running yarn to install ${depName}@${version}`);
```
SuZhou-Joe, line 276: Should we log range info here as well?

ruanyl:
> As of now, build repo during building process, will clone OSD, then pull each plugin into the plugins directory of OSD, before running plugin level bootstrap and build. There is no cleanup of the plugin repos within plugins directory, after the build is completed. We think there are two ways to solve the error on build_workflow code now:
> 
> 1. Add `yarn osd bootstrap --single-version=loose` to every build script including core and plugins if needed.
> 2. In build_workflow code, do a cleanup of plugins directory after one plugin completed the build, before next plugin repo get cloned there.
> 
> Thanks.

@peterzhuamazon I observed the following behaviour while assembling OSD docker image with `opensearch-build`
1. The plugin folder did get removed after the plugin been built successfully
2. However, the plugin folder did **NOT** get removed if the plugin build was failed

So if one plugin build failed, then the single version check might be failed for the next plugin pulled into `plugins` folder as the previous plugin still in `plugins` folder.

I think for `opensearch-build`, I agree that it should always removed the plugin folder to the make the behaviour consistent no matter the plugin build was succeed or failed.

packages/osd-pm/src/utils/validate_dependencies.ts
```diff
@@ -138,21 +178,166 @@ export async function validateDependencies(osd: OpenSearchDashboards, yarnLock:
...
+        log.warning(dedent`
+
+            [single_version_dependencies] Multiple version ranges for package "${depName}"
+            were found across different package.json files. A version, v${suitableVersion}, was identified as the most recent
+            already installed replacement. All package.json files have been updated to indicate a dependency on \`${depName}@${suitableRange}\`.
+
+            The conflicting version ranges are:
+              ${conflictingRanges}
+          `);
+      } else if (singleVersionResolution === SingleVersionResolution.LOOSE) {
```
SuZhou-Joe, line 272: Get lost here, can we move line 181-325 to a help function like `getViolatingSingleVersionDepRangesByDepRanges`, in that case we can fail fast inside the function.
```javascript
if (installedVersions.size === 0) {
  xxx
  return violatingSingleVersionDepRanges;
}

switch (singleVersionResolution)
case "x": do something
```
I believe it will be easier to understand the complex logic inside this huge block.

peterzhuamazon:
> > As of now, build repo during building process, will clone OSD, then pull each plugin into the plugins directory of OSD, before running plugin level bootstrap and build. There is no cleanup of the plugin repos within plugins directory, after the build is completed. We think there are two ways to solve the error on build_workflow code now:
> > 
> > 1. Add `yarn osd bootstrap --single-version=loose` to every build script including core and plugins if needed.
> > 2. In build_workflow code, do a cleanup of plugins directory after one plugin completed the build, before next plugin repo get cloned there.
> > 
> > Thanks.
> 
> @peterzhuamazon I observed the following behaviour while assembling OSD docker image with `opensearch-build`
> 
>     1. The plugin folder did get removed after the plugin been built successfully
> 
>     2. However, the plugin folder did **NOT** get removed if the plugin build was failed
> 
> 
> So if one plugin build failed, then the single version check might be failed for the next plugin pulled into `plugins` folder as the previous plugin still in `plugins` folder.
> 
> I think for `opensearch-build`, I agree that it should always remove the plugin folder to the make the behaviour consistent no matter the plugin build was succeed or failed.

Ok, then it is a bi-product of continue-on-error incremental build I think.
Thanks for observing. cc @zelinh

AMoo-Miki:
> There is no cleanup of the plugin repos within plugins directory, after the build is completed. 

https://github.com/opensearch-project/opensearch-build/blob/main/scripts%2Fdefault%2Fopensearch-dashboards%2Fbuild.sh#L95

This is the cleanup of the plugin.

AMoo-Miki:
> we'll be able to avoid the version-bump-only releases for bunch of components correct ? 

This change deals purely with bootstrapping but what you mentioned is covered by #4612 

> Also given this will ease the building check during bootstrap, would there be any concern during run time ? Usually a match major.minor should work seamlessly though (say osd core 2.11.1 with plugin 2.11.2)

Using `loose`, we ...
1. stop unnecessary and pointless errors when ranges defined in `package.json` are different but the actual dep installed via `yarn.lock` are the same, and
2. if we have differing ranges in`package.json` and versions in `yarn.lock`, it will find something that works for ALL of the ranges. This makes sure that the version it settles on is compatible with what the packages ask for.
    If some package has a range, not version, like `1.2.3` and another `1.2.4`, there is no way we can find a package that satisfies both of them so "loose" will give up and throw an error.
    However, you are right because in practice deps are either prefixed by `~` or `^`, "loose" will do a great job of not causing pain.

AMoo-Miki:
> So if one plugin build failed, then the single version check might be failed for the next plugin pulled into `plugins` folder as the previous plugin still in `plugins` folder.
> 
> I think for `opensearch-build`, I agree that it should always remove the plugin folder to the make the behavior consistent no matter the plugin build was succeed or failed.

@ruanyl, that is great observation. However, this cleanup was a compromise established at the time to be able to make releases. In practice, doing individual builds defies the single-version validation. 

I believe single-version validation was implemented to make sure no differing versions of packages end up in a browser bundle with the potential for conflicting logic that would break the experience or at the very least, wastefully enlarge the bundles. However, the way it was implemented applies to all packages because, during bootstrap, there is no way to know what dep ends up where. So today, only a tiny portion of the errors thrown in the name of single-version are actually problems and the 90% are unnecessary; that is to say 10% are actually important!

By building plugins one at a time and cleaning up after them, we are in fact circumventing the single-version validation. We are also making the build process unnecessarily long. There is also the side-effect of a previous build updating yarn.lock in a way that would compromise the building of the plugins after it.

To summurize that all:
* I believe building plugins one-by-one and cleaning up after each is wrong and can potentially be problematic.
* We should checkout all of our plugins and bootstrap them together because they are all included in the release, together.

packages/osd-pm/src/utils/validate_dependencies.ts
AMoo-Miki: > in that case we can fail fast inside the function

We are going out of our way to not fail fast. The goal is to be able to show all of the violating dep-range combos.

I can do a `continue` instead of using `else` but then someone else might find that less readable; readability is highly subjective. 

However, I had a real dream of having a 5th mode which does loose + brute-force, and for that I would need to break the logic apart. I will give it a shot and if it works, I will update the PR.

AMoo-Miki: I updated this logic to include what I had dreamed. Hope you find it more readable.

packages/osd-pm/src/utils/validate_dependencies.ts
```diff
@@ -138,21 +182,181 @@ export async function validateDependencies(osd: OpenSearchDashboards, yarnLock:
...
+        if (deps?.[`${depName}@${range}`]?.version) {
+          installedVersions.add(deps[`${depName}@${range}`].version);
+          installedDepVersionsCache.set(
+            `${project.name}#${depName}`,
+            deps[`${depName}@${range}`].version
+          );
+        } else {
+          log.warning(`Failed to find the installed version for ${depName}@${range}`);
+          // If we cannot read any one of the installed versions of a depName, there is no point in continuing with it
+          installedVersions.clear();
```
wanglam, line 216: These operation will clear installed versions in other projects, is that by design?

packages/osd-pm/src/utils/validate_dependencies.ts
```diff
@@ -138,21 +182,181 @@ export async function validateDependencies(osd: OpenSearchDashboards, yarnLock:
...
+          desiredRanges.clear();
+          break rangesLoop;
+        }
+      }
+
+      desiredRanges.set(range, projects);
+    }
+
+    // More than one range is used but couldn't get all the installed versions: call out violation
+    if (installedVersions.size === 0) {
```
wanglam, line 226: If `installedVersions.size` equal 1, shall we check if the installed versions was from different ranges?

packages/osd-pm/src/utils/validate_dependencies.ts
SuZhou-Joe: Thanks @AMoo-Miki .As for me it looks much easier to under as we compare the Resolution and only the Resolution in the first level. One suggestion is maybe we need to add unit test for all these `if else` branches.

AMoo-Miki: I will have to think of a way to test these. They depend on what Yarn does to complete so <fingers crossed>.

packages/osd-pm/src/utils/project.ts
```diff
@@ -261,6 +262,51 @@ export class Project {
...
+
+    log.write('');
```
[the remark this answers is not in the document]
AMoo-Miki: They are part of the original bootstrap logic of separation.

packages/osd-pm/src/utils/project.ts
```diff
@@ -261,6 +262,51 @@ export class Project {
...
+
+    const rangeToUse = range || `^${version}`;
+
+    const extraArgs = [`${depName}@${version}`];
+    if (dev) extraArgs.push('--dev');
+
+    if (this.isWorkspaceProject) {
+      await installInDir(this.path);
```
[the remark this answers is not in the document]
AMoo-Miki: With `isWorkspaceProject`, we don't add a package and we just install. However, there is a piece of code missing here.

packages/osd-pm/src/utils/validate_dependencies.ts
AMoo-Miki: Yes. I am using an empty `installedVersions` as an indicator of something going wrong.

seraphjiang:
could we elaborate why the change is associated with package.json, my understand none of osd plugins dependencies are declare and managed in package.json 

> > we'll be able to avoid the version-bump-only releases for bunch of components correct ?
> 
> This change deals purely with bootstrapping but what you mentioned is covered by #4612
> 
> > Also given this will ease the building check during bootstrap, would there be any concern during run time ? Usually a match major.minor should work seamlessly though (say osd core 2.11.1 with plugin 2.11.2)
> 
> Using `loose`, we ...
> 
> 1. stop unnecessary and pointless errors when ranges defined in `package.json` are different but the actual dep installed via `yarn.lock` are the same, and
> 2. if we have differing ranges in`package.json` and versions in `yarn.lock`, it will find something that works for ALL of the ranges. This makes sure that the version it settles on is compatible with what the packages ask for.
>    If some package has a range, not version, like `1.2.3` and another `1.2.4`, there is no way we can find a package that satisfies both of them so "loose" will give up and throw an error.
>    However, you are right because in practice deps are either prefixed by `~` or `^`, "loose" will do a great job of not causing pain.

packages/osd-pm/src/utils/validate_dependencies.ts
```diff
@@ -40,8 +43,43 @@ import { log } from './log';
...
+  IGNORE = 'ignore',
+}
+
+export async function validateDependencies(
+  osd: OpenSearchDashboards,
+  yarnLock: YarnLock,
+  /* `singleVersionResolution` controls how violations of single-version-dependencies is applied.
+   *    STRICT (default): throw an error and exit
+   *    LOOSE: identify and install a single version that satisfies all ranges
+   *    BRUTE_FORCE: identify and install the newest version
```
seraphjiang, line 60: UES_NEWEST is more clear

packages/osd-pm/src/utils/validate_dependencies.ts
```diff
@@ -40,8 +43,43 @@ import { log } from './log';
...
+   *
+   * `BRUTE_FORCE`:
+   *          With no care for reconciliation, the newest of the various versions installed is chosen, irrespective of
+   *          whether it satisfies any of the ranges. Installing the chosen version updates the lock-files and a range
+   *          in the form of `^<version>` is applied to all `package.json` files that declared the dependency.
+   *
+   * `FORCE`:
+   *          For each dependency, first LOOSE resolution is attempted but if that fails, BRUTE_FORCE is applied.
+   *
+   * `IGNORE`:
```
seraphjiang, line 76: Any benefit by using this option more than get error as warning.  could we still run osd by ignore the error/warning. If not, should we callout specific dev scenario to leverage this option

packages/osd-pm/README.md
```diff
@@ -141,6 +141,40 @@ yarn osd
 Bootstrapping also calls the `osd:bootstrap` script for every included project.
 This is intended for packages that need to be built/transpiled to be usable.
 
+#### Single-version validation
+
+Bootstrapping, by default, applies a `strict` single-version validation where it requires all the dependencies defined
```
seraphjiang, line 146: could we clarify what does single version validation mean and probably giving an example. that will help. let's say HelloPlugin

ruanyl:
> this cleanup was a compromise established at the time to be able to make releases.

Great to know, thanks for the background! @AMoo-Miki 

> So today, only a tiny portion of the errors thrown in the name of single-version are actually problems and the 90% are unnecessary; that is to say 10% are actually important!

Right, in my past experience, nearly all single-version error cases that I encountered in development are about dev dependencies, it would be nice to allow plugins to choose whichever version of dev dependencies they want to use, like husky, lint-staged, etc. I believe dev dependencies won't get bundled. So I'm thinking, does it make sense to introduce an option to skip the validation of dev dependencies?

> I believe building plugins one-by-one and cleaning up after each is wrong and can potentially be problematic.
> We should checkout all of our plugins and bootstrap them together because they are all included in the release, together.

Totally understand your points, but does it imply plugins are NOT independent at build time? I'm a little concern the consequence, will that "encourage" people to write code which statically depends on other plugins? like `import bar from '../../other-plugin-name/foo.ts'`? I believe the long term vision is to make OSD plugins to be able to build independently,  is that correct? Ideally, the plugin should be built correctly no matter it's with other plugins or without other plugins.

packages/osd-pm/src/utils/validate_dependencies.ts
AMoo-Miki: You can still run OSD with `IGNORE` and it will more than likely work just fine.

packages/osd-pm/README.md
AMoo-Miki: This is really all that it is:
```
Bootstrapping, by default, applies a `strict` single-version validation where it requires all 
the dependencies defined more than once to have the same version-range in the `package.json`
files of Dashboards, packages, and plugins. If a violation is identified, bootstrapping terminates
with an error.
```

Unfortunately, I don't know the exact reasons behind its existence; I have some theories but I am having a hard time defending them.

AMoo-Miki:
> So I'm thinking, does it make sense to introduce an option to skip the validation of dev dependencies?

`--single-version=loose` does just that. My goals is to make `loose` the default in the next major release.


> ... but does it imply plugins are NOT independent at build time? 

Since we build and package a bunch of plugins together with each OSD release, I am pushing for them to ALSO be compiled (aka bootstrapped) together to make sure they don't have any conflicts with each other. Of course they are all independent of each other and can be compiled without the other plugins. However, being compiled together will be one extra validation to boost out confidence that users with multiple plugins can do a `bootstrap` on their own as well.

> I'm a little concern the consequence, will that "encourage" people to write code which statically depends on other plugins? like `import bar from '../../other-plugin-name/foo.ts'`?

We have (or should have) linting rules that would prevent that from happening. @kavilla, please tell me I am right.

> I believe the long term vision is to make OSD plugins to be able to build independently, is that correct? 

Yes. The hope is that one day, before I turn 100, plugins can be compiled and packaged with OSD being in the neighborhood.

> Ideally, the plugin should be built correctly no matter it's with other plugins or without other plugins.

Indeed. Being compiled alone is not too much of a challenge but being compiled side-by-side of another plugin, is complicated due to the possible impacts of any common deps with different versions.

packages/osd-pm/src/utils/validate_dependencies.ts
seraphjiang: well, not that scary to me ^o^

seraphjiang:
i got it.

maybe change
All of OSD and plugin dependencies are declared and managed via `the` `package.json`
to
All of OSD and plugin dependencies are declared and managed via `their` `package.json`

help disambiguate which package.json

> > could we elaborate why the change is associated with package.json, my understand none of osd plugins dependencies are declare and managed in package.json
> 
> All of OSD and plugin dependencies are declared and managed via the `package.json`, and during bootstrap, `package.json` is the source of truth about the dependencies. If you are thinking about `yarn.lock`, it is used by Yarn for maintaining the "installed" versions. In other words, changes to `package.json` will change `yarn.lock` but not the other way around. Lemme know if that was not what you were referring to.

kavilla (approved):

ZilongX (approved):

Issue #5720, opened 2024-01-22
[Tests] Add Github workflow for Test Orchestrator in FT Repo to run cypress tests within Dashboards repo


manasvinibs:
This workflow gets triggered on a dispatch event sent from a FT repo test orchestrator. The success or failure of the workflow run status will be sent back to the orchestrator. Test results will be uploaded in the given path within source code and will be available within GitHub workflow.

PR #5725, opened 2024-01-22
[Tests] Add Github workflow for Test Orchestrator in FT Repo to run cypress tests within Dashboards repo

manasvinibs:
### Description

 * Adds Github workflow which gets triggered on dispatch event sent from FT Repo Orchestrator. Currently this workflow defaults to using release bundle artifact for Opensearch and Dashboards. 
 * In this iteration, pulling out few of the dashboards sanity test from FT repo into the Dashboards repo -  https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/cypress/integration/common/dashboard_sample_data_spec.js
 * Introduces Cypress dependency into the package json to run cypress tests within repo. Currently, I'm pulling the version which matches the one in FT repo. 
 * Adds cypress config file.

ToDo:
Add fallback mechanism for using snapshot URL when release bundle url is not accessible.
Enable this workflow for each PR/push event.

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5720

## Screenshot

https://github.com/manasvinibs/OpenSearch-Dashboards/actions/runs/7618260248/job/20749040514

### Check List

- [X] All tests pass
  - [ ] `yarn test:jest`


scripts/cypress_tests.sh
SuZhou-Joe, line 1: Do we need to add this sh file in every plugin if any of them want to onboard the test Orchestrator?

SuZhou-Joe:
Will we pull all of the tests out of test repo to OpenSearch Dashboards repo?

cypress.config.js
```diff
@@ -0,0 +1,28 @@
...
+    SECURITY_ENABLED: false,
+    AGGREGATION_VIEW: false,
+    username: 'admin',
+    password: 'myStrongPassword123!',
+    ENDPOINT_WITH_PROXY: false,
+    MANAGED_SERVICE_ENDPOINT: false,
+    VISBUILDER_ENABLED: true,
+    DATASOURCE_MANAGEMENT_ENABLED: false,
+    ML_COMMONS_DASHBOARDS_ENABLED: true,
+    WAIT_FOR_LOADER_BUFFER_MS: 0,
```
SuZhou-Joe, line 26: It seems only `openSearchUrl` is needed based on the migrated test?

package.json
```diff
@@ -77,7 +77,9 @@
     "docs:acceptApiChanges": "scripts/use_node --max-old-space-size=6144 scripts/check_published_api_changes.js --accept",
     "osd:bootstrap": "scripts/use_node scripts/build_ts_refs && scripts/use_node scripts/register_git_hook",
     "spec_to_console": "scripts/use_node scripts/spec_to_console",
-    "pkg-version": "scripts/use_node -e \"console.log(require('./package.json').version)\""
+    "pkg-version": "scripts/use_node -e \"console.log(require('./package.json').version)\"",
+    "cypress:run-without-security": "env TZ=America/Los_Angeles NO_COLOR=1 cypress run --headless --env SECURITY_ENABLED=false"
```
SuZhou-Joe, line 81: Should we introduce `SECURITY_ENABLED` concept into OSD core as security plugin is a plugin.

package.json
```diff
@@ -355,6 +359,7 @@
     "chromedriver": "^119.0.1",
     "classnames": "2.3.1",
     "compare-versions": "3.5.1",
+    "cypress": "9.5.4",
```
SuZhou-Joe, line 362: Will all of the plugins honor the cypress version here? Or it will be difficult to upgrade Cypress version and as far as I know, latest Cypress version has reached v13. Though this is the version we used in Cypress test repo, we have to add this version to a centralized place for upgrade in the future.

scripts/cypress_tests.sh
manasvinibs: I think so. Every plugin has to have their GitHub workflow file which runs cypress tests in their repo along with setup scripts and config files to be able to run cypress tests inside the repo. I can help adding readme doc/example plugin setup once we are done with onboarding Dashboards.

manasvinibs:
> Will we pull all of the tests out of test repo to OpenSearch Dashboards repo?

I think that's the end goal, but till we reach there the idea was to have two states where in future tests will be written in the components repository while also maintaining the current test setup in Functional Test repo as it is. Once we are comfortable with the new orchestrator setup and plugins starts onboarding to write new tests in their own repo, we can also start pulling out all the old ones into its respective repos.

package.json
manasvinibs: I think this is for this first iteration. The plan is to add Functional test repo as a dependency to consume cypress version and its config in the component repositories.

cypress.config.js
manasvinibs: Yes but as the plan is to right away start writing new tests in this repo, I've added/matched mostly what is in FTRepo.

kavilla:
related to: https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/988

package.json
manasvinibs: I think so because these are Cypress env variables that needs to be set in order to override some of the configurations/commands based on if security plugin is installed or not. 
https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/cypress/utils/commands.js#L75

scripts/cypress_tests.sh
SuZhou-Joe: Sure, maybe add some helper function such as polling script into the test-library as well so that we can keep the cypress_tests.sh as simple as possible for onboarding.

package.json
SuZhou-Joe: Sounds good, please try to keep only one place to store the cypress version and some common configs.

package.json
SuZhou-Joe: Just feel a little bit weird as for OSD core, it has no knowledge on authentication. @kavilla What's your opinion?

SuZhou-Joe (dismissed):

manasvinibs:
@SuZhou-Joe @kavilla Resolved merge conflict. Can you pls re-approve?

AMoo-Miki (approved):

scripts/bwctest_osd.sh
```diff
@@ -5,9 +5,9 @@
 
 set -e
 
-. scripts/bwc/utils.sh
-. scripts/bwc/opensearch_service.sh
-. scripts/bwc/opensearch_dashboards_service.sh
+. scripts/common/utils.sh
```
kavilla, line 8: could we create an issue into adding these to: https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/packages/osd-test

.github/workflows/dashboards_cypress_workflow.yml
```diff
@@ -0,0 +1,143 @@
...
+        required: false
+        type: string
+
+env:
+  TEST_REPO: ${{ inputs.test_repo != '' && inputs.test_repo || github.repository }}
+  TEST_BRANCH: "${{ inputs.test_branch != '' && inputs.test_branch || github.base_ref }}"
+  OSD_PATH: 'osd'
+  CYPRESS_BROWSER: 'chromium'
+  JOB_ID: ${{ inputs.UNIQUE_ID}}
+  OPENSEARCH:  ${{ inputs.OS_URL != '' && inputs.OS_URL  || 'https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/$VERSION/latest/linux/x64/tar/dist/opensearch/opensearch-$VERSION-linux-x64.tar.gz' }}
```
kavilla, line 46: would we be able to do a matrix and pass in like linux, windows and x64, arm

.github/workflows/dashboards_cypress_workflow.yml
```diff
@@ -0,0 +1,143 @@
+name: Orchestrator cypress workflow
+run-name: dashboards_cypress_workflow ${{ inputs.UNIQUE_ID != '' && inputs.UNIQUE_ID || '' }} # Unique id number appended to the workflow run-name to reference the run within the orchestrator.
```
kavilla, line 2: Is a better name for this file perhaps `release_cypress_workflow` since it will execute the tests within this repo against a release artifact.

.github/workflows/dashboards_cypress_workflow.yml
```diff
@@ -0,0 +1,143 @@
...
+        run: |
+          cd ${{ env.OSD_PATH }}
+          echo "VERSION=$(yarn --silent pkg-version)" >> $GITHUB_ENV
+
+      - name: Run bootstrap
+        run: |
+          cd ${{ env.OSD_PATH }}
+          yarn osd bootstrap
+
+      - name: Download and extract Opensearch artifacts
```
kavilla, line 111: nit: OpenSearch

.github/workflows/dashboards_cypress_workflow.yml
```diff
@@ -0,0 +1,143 @@
...
+        run: |
+          CWD=$(pwd)
+          mkdir -p $CWD/${{ env.OPENSEARCH_DIR }}
+          source ${{ env.OSD_PATH }}/scripts/common/utils.sh
+          open_artifact $CWD/${{ env.OPENSEARCH_DIR }} ${{ env.OPENSEARCH }}
+      
+      - name: Download and extract Opensearch Dashboards artifacts
```
kavilla, line 118: NIT: OpenSearch

.github/workflows/dashboards_cypress_workflow.yml
```diff
@@ -0,0 +1,143 @@
...
+      - name: Run Cypress tests 
+        run: |
+          source ${{ env.OSD_PATH }}/scripts/cypress_tests.sh
+          run_dashboards_cypress_tests  
+
+      # Screenshots are only captured on failures
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: osd-cypress-screenshots
```
kavilla, line 134: same as the above comment: indicate that these tests were executed against a release artifact so `release-osd-cypress-screenshots` or something if folks were looking at a glance.

.github/workflows/dashboards_cypress_workflow.yml
```diff
@@ -0,0 +1,143 @@
...
+          npm uninstall -g yarn
+          npm i -g yarn@1.22.10
+          yarn config set network-timeout 1000000 -g
+
+      - name: Get Cypress version
+        id: cypress_version
+        run: |
+          echo "name=cypress_version::$(cat ./${{ env.OSD_PATH }}/package.json | jq '.devDependencies.cypress' | tr -d '"')" >> $GITHUB_OUTPUT
+    
+      - name: Cache Cypress
```
kavilla, line 90: totally unrelated: but does this cache help the re-run of stuff?

cypress.config.js
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const { defineConfig } = require('cypress');
+
+module.exports = defineConfig({
```
kavilla, line 8: IMO, I think if we are adding this already we should create a `cypress` folder in [test folder](https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/test), move this [spec](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5725/files#diff-b2eea1186135b946e8d070f62b93e527ddc1f56ae90e8ae295630efd6c80ef55) to  that folder.

Sort of how like selenium does it where it defines the config at the functional folder level and adds subdirectories below that.

kavilla:
We should also update this: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/.github/workflows/cypress_workflow.yml to execute the cypress tests within this repo.

This way we can get the release testing verification added in this PR that is triggered with a dispatch event with a specified build. While the existing cypress test workflow will address the code within opened pull requests and run the tests.

While we are at it, is there an easy way to code cov from cypress and add the diff to PRs: https://app.codecov.io/gh/opensearch-project/OpenSearch-Dashboards

kavilla (commented):
some comments

kavilla (approved):
todo: create an issue and add follow up feedback.

manasvinibs:
> todo: create an issue and add follow up feedback.

Created an issue to address PR comments https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5892

package.json
cwperks: The security plugin needs Cypress >= 12 for cross-origin testing. I opened a [META](https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/801) to track the upgrade of Cypress in the FTR repo, but in order to provide adequate testing for cross-origin cases, the security-dashboards-plugin added cypress tests directly in the security-dashboards-plugin repo and uses [Cypress 13](https://github.com/opensearch-project/security-dashboards-plugin/blob/main/package.json#L30).

Can we allow dashboards-plugins to have flexibility with the version of cypress required for their testing without forcing all of them to use a version of cypress that does not support a feature needed for their plugin's testing needs. The reason security-dashboards-plugin needs cross-origin testing is to test single sign on behavior with different authentication providers like SAML and OpenID.

manasvinibs:
Manual backport https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6095

Issue #5717, opened 2024-01-19
[MD] Data source picker should have same behavior in devtool, tutorial, etc where it is used


BionIT:
**Is your feature request related to a problem? Please describe.**

I feel confused when seeing different behavior for the data source picker based on where it is located

**Describe the solution you'd like**

The datasource picker should be one component to ensure the behavior is the same. Then, if we need to change the behavior for the picker, we don't need to go through entire code base to find the duplicated code in order to make change. The component should be reusable by other plugins
To unify the behavior, the default option for the picker when first opening the page with the picker should be `Local cluster`. After removing the option, it should display `Select a data source`. Change the `DataSource` to `Data source`

**Describe alternatives you've considered**

Stay where we are and just change every places every time that a small text adjustment is needed for the picker. 

**Additional context**
When data source plugin is enabled, if we go to dev tool page or add sample data page, the behavior of the data source picker is different: 
- dev tools shows `Select a data source` and the prepend is `Data source`
![Screenshot 2024-01-19 at 2 41 45 PM](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/8d764a79-8c7f-45bd-a0af-8fa8bda057c6)

- add sample data tutorial shows a default value and the prepend is `DataSource`
![Screenshot 2024-01-19 at 2 41 55 PM](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/e131a3e8-17bd-4a89-a5fa-f4b10087e931)
after removing the default value it shows `Select a Data Source`
![Screenshot 2024-01-19 at 2 42 02 PM](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/13c39fea-2349-4117-b41e-b25808e13c9a)

BionIT:
Happy to work on it

bandinib-amzn:
> After removing the option, it should display Select a data source. Change the Data source or DataSource to Destination

Does that mean all occurrences of `Select a data source` will change to `Select a destination`?

bandinib-amzn:
Adding here one feedback I received from UX team while reviewing these pages:

Make first created data source as default data source and set default data source as default selected option on dev tool and sample data screens instead of `Select a data source`.

BionIT:
> > After removing the option, it should display Select a data source. Change the Data source or DataSource to Destination
> 
> Does that mean all occurrences of `Select a data source` will change to `Select a destination`?

[...]

PR #5756, opened 2024-01-30
Add datasource picker component and use it in devtools and tutorial page when multiple datasource is enabled

BionIT:
### Description

This resolves https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5717. 
The datasource picker has different behavior in devtools and tutorial page when multiple datasource is enabled, it should be one component to ensure the behavior is the same. 

This change adds datasource picker component and use it in devtools and tutorial page when multiple datasource is enabled to unify the behavior: the default option for the picker when first opening the page with the picker should be `Local cluster`. After removing the option, it should display `Select a data source`. Also, the prepend should be `Data source` instead of `DataSource`

### Issues Resolved

fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5717

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/920260dd-885d-4774-83a5-2174c64e67f9

## Testing the changes
1. If data_source.enabled is false, then data source picker doesn't show up in devtools nor add sample data page. 
2. Change data_source.enabled to true will turn on multiple datasource
3. Go to devtools page, the datasource picker should appear on top right, with `Local cluster` chosen as the default option. Choose the local data source, query data should be successful, choose remote data source,  query data should be successful
4. Go to add sample data page, the datasource picker should appear, with `Local cluster` chosen as the default option. Choose the local data source, add sample data should be successful, choose remote data source,  add sample data should be successful

### Check List

- [ ] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
+++ b/src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
@@ -57,7 +57,7 @@ export class DataSourcePicker extends React.Component {
       });
   }
 
-  onChang(e) {
+  onChange(e) {
     if (!this._isMounted) return;
     this.setState({
       selectedOption: e,
@@ -77,8 +77,10 @@ export class DataSourcePicker extends React.Component {
         singleSelection={{ asPlainText: true }}
         options={this.state.dataSources}
         selectedOptions={this.state.selectedOption}
-        onChange={(e) => this.onChang(e)}
-        prepend="Data source"
+        onChange={(e) => this.onChange(e)}
+        prepend={i18n.translate('dataSourceComboBoxPrepend', {
+          defaultMessage: 'Data source',
+        })}
         compressed
         isDisabled={this.props.disabled}
       />
--- a/src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
+++ b/src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
@@ -1,0 +1,87 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from 'react';
+import { getDataSources } from '../utils';
+import { i18n } from '@osd/i18n';
+import { EuiComboBox } from '@elastic/eui';
+
+export const LocalCluster = {
+  label: i18n.translate('dataSource.localCluster', {
+    defaultMessage: 'Local cluster',
+  }),
+  id: '',
+};
+
+export class DataSourcePicker extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      dataSources: [],
+      selectedOption: [LocalCluster],
+    };
+  }
+
+  componentWillUnmount() {
+    this._isMounted = false;
+  }
+
+  async componentDidMount() {
+    this._isMounted = true;
+    getDataSources(this.props.savedObjectsClient)
+      .then((fetchedDataSources) => {
+        if (fetchedDataSources?.length) {
+          const dataSourceOptions = fetchedDataSources.map((dataSource) => ({
+            id: dataSource.id,
+            label: dataSource.title,
+          }));
+
+          dataSourceOptions.push(LocalCluster);
+
+          if (!this._isMounted) return;
+          this.setState({
+            ...this.state,
+            dataSources: dataSourceOptions,
+          });
+        }
+      })
+      .catch(() => {
+        this.props.notifications.addWarning(
+          i18n.translate('dataSource.fetchDataSourceError', {
+            defaultMessage: 'Unable to fetch existing data sources',
+          })
+        );
+      });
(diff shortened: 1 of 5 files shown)


ZilongX (approved):

src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
```diff
@@ -0,0 +1,87 @@
...
+      .catch(() => {
+        this.props.notifications.addWarning(
+          i18n.translate('dataSource.fetchDataSourceError', {
+            defaultMessage: 'Unable to fetch existing data sources',
+          })
+        );
+      });
+  }
+
+  onChang(e) {
```
bandinib-amzn, line 60: nit: spelling mistake

src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
```diff
@@ -0,0 +1,87 @@
...
+    if (!this._isMounted) return;
+    this.setState({
+      selectedOption: e,
+    });
+    this.props.onSelectedDataSource(e);
```
bandinib-amzn, line 65: How UI will look like if enters in catch block?

src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  render() {
+    return (
+      <EuiComboBox
+        aria-label={i18n.translate('dataSourceComboBoxAriaLabel', {
+          defaultMessage: 'Select a data source',
```
bandinib-amzn, line 72: Oh, I thought we are changing it to `Select a destination`. Did we finally land on `Select a data source`?

src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
```diff
@@ -0,0 +1,87 @@
...
+        })}
+        placeholder={i18n.translate('dataSourceComboBoxPlaceholder', {
+          defaultMessage: 'Select a data source',
+        })}
+        singleSelection={{ asPlainText: true }}
+        options={this.state.dataSources}
+        selectedOptions={this.state.selectedOption}
+        onChange={(e) => this.onChang(e)}
```
bandinib-amzn, line 80: nit: spelling mistake

src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
```diff
@@ -0,0 +1,87 @@
...
+        prepend="Data source"
```
bandinib-amzn, line 81: Can you double check if we need to i18n this text as well?

bandinib-amzn:
Thanks @BionIT for making this change. I really appreciate your efforts to unify the data source picker component and remove repetitive code. I can see the great value of this change from maintenance, consistent experience and reusability perspective.

src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
BionIT: Yeah, since destination indicates current dashboard rather than remote data source, thus we changed back to data source

src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
BionIT: Good catch!

src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
BionIT: Thanks for pointing it out! Looked through the repo and found different handling of the prepend, some wrapped with i18n while others are not. I will wrap it so that when we have translation support, we don't left it behind.

src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
BionIT: For devtools and add sample data, a toast will show up

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/96545e0b-bd33-4fc4-a0d5-6f9f31780452

src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
BionIT: Resolved in new commit

src/plugins/data_source_management/public/components/data_source_picker/data_source_picker.js
BionIT: Resolved in new commit

bandinib-amzn (approved):

Flyingliuhub (approved):

PR #5776, opened 2024-02-02
Routing for surrounding doc link should work without a question mark appending to the end

### LINKED ISSUE #5774: [BUG] Discover routing without query params does not route correctly
**Describe the bug**

Wen visiting the single document or context pages, if the url does not contain query parameters, the routing breaks. e.g.

```sh
# Works
http://localhost:5601/app/discover#/context/90943e30-9a47-11e8-b64d-95841ca0b247/oyj5iIwBAXmWFklaNiuy?

# Does not work
http://localhost:5601/app/discover#/context/90943e30-9a47-11e8-b64d-95841ca0b247/oyj5iIwBAXmWFklaNiuy
```

The issue is with how discover maps the legacy URL's to the new data explorer route. The url parameters after the `?` are optional, but the regex currently requires the `?` to be present to route the URL correctly. 

**To Reproduce**

**Expected behavior**

Routes `http://localhost:5601/app/discover#/context/<doc-id>/<index-id>` correctly

**OpenSearch Version**
Please list the version of OpenSearch being used.

**Dashboards Version**
Please list the version of OpenSearch Dashboards being used.

**Plugins**

**Screenshots**

**Additional context**

abbyhu2000:
### Description
URLs such as /app/discover#/context/<document-id>/<index-id> which is a valid URL for the context page should work.

Previously the link will only work if there is a question mark append to it.

### Issues Resolved

closes #5774 

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/43937633/b0a4ba77-9d9d-4764-b4a2-1e4bf9a3de7e

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/migrate_state.ts
+++ b/src/plugins/discover/public/migrate_state.ts
@@ -66,7 +66,7 @@ export function migrateUrlState(oldPath: string, newPath = '/'): string {
   let path = newPath;
   const pathPatterns = [
     {
-      pattern: '#/context/:indexPattern/:id?:appState?',
+      pattern: '#/context/:indexPattern/:id',
       extraState: { docView: 'context' },
       path: `context`,
     },
--- a/src/plugins/discover/public/migrate_state.ts
+++ b/src/plugins/discover/public/migrate_state.ts
@@ -71,7 +71,7 @@ export function migrateUrlState(oldPath: string, newPath = '/'): string {
       path: `context`,
     },
     {
-      pattern: '#/doc/:indexPattern/:index\\?:appState?',
+      pattern: '#/doc/:indexPattern/:id',
       extraState: { docView: 'doc' },
       path: `doc`,
     },
@@ -99,6 +99,7 @@ export function migrateUrlState(oldPath: string, newPath = '/'): string {
     case `doc`:
     case `context`:
       path = oldPath;
+      break;
     case `discover`:
     case `savedSearch`:
       const params = matchPath<DiscoverParams>(oldPath, {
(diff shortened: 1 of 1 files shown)


src/plugins/discover/public/migrate_state.ts
```diff
@@ -66,7 +66,7 @@ export function migrateUrlState(oldPath: string, newPath = '/'): string {
   let path = newPath;
   const pathPatterns = [
     {
-      pattern: '#/context/:indexPattern/:id\\?:appState?',
+      pattern: '#/context/:indexPattern/:id?:appState?',
```
ashwin-pc, line 69: This works because the a route like `#/context/ip/abcd` results in `indexPattern = ip`, `id = a` and `appState = bcd`

But it looks like we dont need the `appState` so lets just change the paths to 

```
    {
      pattern: '#/context/:indexPattern/:id',
      extraState: { docView: 'context' },
      path: `context`,
    },
    {
      pattern: '#/doc/:indexPattern/:index',
      extraState: { docView: 'doc' },
      path: `doc`,
    },
```

Also you forgot to add the break statement after line 101

abbyhu2000: I dont think we need to add the break statement on line 101? since it will return path anyway @ashwin-pc

ashwin-pc: So the way a switch statement works is that once it matches a particular switch case, everything under that is executed until it hits a break or a default statement. So in this case, once the dog or context parts are matched, all commands under that are exterior until it hits the final break statement, which means that everything inside the discover path is also executed. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

ashwin-pc: Can you also remove the appstate from the doc route?

abbyhu2000: fixed

ashwin-pc (approved):

Issue #5712, opened 2024-01-19
[Feature] Support Multiple Data Source when Import saved objects from uploading files


yujin-emma:
**Is your feature request related to a problem? Please describe.**

When multiple data source enabled, user can add [sample data](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/4412/files) and use the [dev tool](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/3806) with one specified data source. Import saved objects accepts uploaded files, but the API does not know the data source. These objects cannot display since the it does not know where to fetch the data.

![Screenshot 2024-01-18 at 12 48 19](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/112784385/3538a0ee-f225-438d-93b8-dc60c806a56c)

We want to provide user the ability to import saved objects after selecting one data source. When user import saved object from  cluster where not enabled multiple data source feature, the imported saved objects will attach the dataSourceId with the objectId; while user import saved objects from some cluster with multiple data source feature enabled, then the objectId would already contains data source info, for this case, we want to replace the dataSourceId. The id schema are exactly same as when user add sample data and dev tools with multiple data source enabled. 

![Screenshot 2024-01-18 at 12 26 38](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/112784385/ed00a3d8-70ae-4832-b6b6-9ef21259538b)
![Screenshot 2024-01-18 at 12 27 48](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/112784385/2827a7cd-4541-4ecd-be89-a72e4e828680)

**Challenges**

There are two modes to import saved objects, one is create new copy, the other is check conflict and override. We need to support both modes and cannot break the existing behavior. 
Current Import API support create new copy by using the exact objectId to construct the importIdMap, support check conflict and override by checking if there is anything error and only push overrides to construct final import saved object request. 
For these two modes, we need to resolve carefully, below are the detailed approach to support multiple data source when import saved objects.

**Describe the solution you'd like**

***Mode - create new copy***

When choosing create new copy, after user select one data source, attach the dataSourceId to the saved objects. The `_`  is between dataSourceId and objectId. 
```
if (dataSourceId) {
      return acc.set(`${object.type}:${object.id}`, {
        id: `${dataSourceId}_${uuidv4()}`,
        omitOriginId: true,
      });
    }
[...]

PR #5167, opened 2023-10-02
[Feature] Datasource selector of multiple datasources 

mengweieric:
### Description

Datasource selector support in OpenSearch Dashboard Discover

### Issues Resolved

fixes #4482, #4991 

## Screenshot

Screenshot 2023-10-01 at 8 13 59 PM

### Check List

- [ ] All tests pass
  - [x] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
  - [ ] `yarn test:ftr`
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #5781, opened 2024-02-02
Add datasource picker to import saved object flyout when multiple data source is enabled

BionIT:
### Description

When multiple data source plugin is enabled, user can connect to remote data sources in addition to the local cluster, to support importing saved objects from remote data source via file upload(https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5712), we need to have an option in the flyout for user to select where the file comes from. 

Via discussion in the issue, we are altering the order of the import option, and have the `Create new objects with random IDs`  as the default choice. When multiple data source is enabled, we will have the data source picker under `Import options` section then the `Conflict management` section will contain the options to address conflicts.

### Issues Resolved

Fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5712 

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/1bdda995-47f0-425e-820c-ce031f80fbf8

## Testing the changes

In the video above, we tested the following scenarios
when multiple data source is disabled:
- upload same ndjson with `Create new objects with unique IDs` option is successful when there are duplicate objects exists
- upload same ndjson with `Automatically overwrite conflicts` option is successful when there are duplicate objects
- upload same ndjson with `Request action on conflict` option will prompt user to handle on the conflict
- upload same json with `Automatically overwrite conflicts` option twice is successful when there are duplicate objects exists
- upload same json with `Request action on conflict` option will prompt user to handle on the conflict
when multiple data source is enabled:
- upload ndjson with `Create new objects with unique IDs` option is successful 
- upload ndjson with `Automatically overwrite conflicts` option is successful when there are duplicate objects
- upload same ndjson with `Request action on conflict` option will prompt user to handle on the conflict
- upload same json with `Automatically overwrite conflicts` option twice is successful when there are duplicate objects exists
- upload same json with `Request action on conflict` option will prompt user to handle on the conflict

### Check List

- [x] All tests pass
  - [ ] `yarn test:jest`


ashwin-pc (changes requested):
We already have a datasource selector component in the data plugin. We should use that and not create a new picker.

bandinib-amzn:
I'm coming here after watching video attached in screenshot. Does it make sense to add a checkbox or another option to select 'overwrite all' instead of having to click 'overwrite' 15 times? cc: @kgcreative

src/plugins/saved_objects_management/public/management_section/objects_table/components/flyout.tsx
```diff
@@ -78,8 +79,10 @@ import { FailedImportConflict, RetryDecision } from '../../../lib/resolve_import
 import { OverwriteModal } from './overwrite_modal';
 import { ImportModeControl, ImportMode } from './import_mode_control';
 import { ImportSummary } from './import_summary';
+// eslint-disable-next-line @osd/eslint/no-restricted-paths
```
bandinib-amzn, line 82: Do we need this?

src/plugins/saved_objects_management/public/management_section/objects_table/components/flyout.tsx
```diff
@@ -78,8 +79,10 @@ import { FailedImportConflict, RetryDecision } from '../../../lib/resolve_import
...
+import { DataSourcePicker } from '../../../../../data_source_management/public/components/data_source_picker/data_source_picker.js';
```
bandinib-amzn, line 83: nit: consider adding index file in `data_source_management/public/components/data_source_picker` and export from there

src/plugins/saved_objects_management/public/management_section/objects_table/components/flyout.tsx
```diff
@@ -78,8 +79,10 @@ import { FailedImportConflict, RetryDecision } from '../../../lib/resolve_import
...
 
-const CREATE_NEW_COPIES_DEFAULT = false;
+const CREATE_NEW_COPIES_DEFAULT = true;
```
bandinib-amzn, line 85: why are we changing the value of this const? won't it impact current behaviour?

bandinib-amzn:
> We already have a datasource selector component in the data plugin. We should use that and not create a new picker.

Hi @ashwin-pc 
I guess you are talking about this PR - https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5167 and wondering why we don't use the same selector.

Let's consider the use case from above PR which is Datasource selector support in OpenSearch Dashboard Discover.
The **Discover** feature in OpenSearch Dashboards is designed for searching, exploring, and visualizing data stored in **OpenSearch indices** or **Remote cluster indices**. 

The **Dev Tool** feature in OpenSearch Dashboard provides an environment for developers and administrators to interact with **OpenSearch** or **Remote cluster** with query DSL and API directly.

I took discover and dev tool here as an example to show the difference between two pickers here. They serve different purposes. #5167 has  picker for selecting data source for searching, exploring data in the selected data source. Whereas data source picker in this PR is listing the data sources (basically remote cluster) to interact with them directly for fetch sample data, fetch indices to create index patterns etc.

IMO, they both serves different purpose, so I would prefer separate picker. @kgcreative I would like to here your thoughts as well.

src/plugins/saved_objects_management/public/management_section/objects_table/components/flyout.tsx
BionIT: Yeah, as discussed with @kgcreative , we want to have create new copies with unique IDs as default option, and find screenshot here https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5712#issuecomment-1922650607, and this option is switched up in order as well

src/plugins/saved_objects_management/public/management_section/objects_table/components/flyout.tsx
BionIT: Yes, this is required to avoid lint error

BionIT:
> I'm coming here after watching video attached in screenshot. Would it be possible to consider adding a checkbox or an alternative option for selecting 'overwrite all,' instead of having to click 'overwrite' 15 times? This could potentially enhance the user experience. cc: @kgcreative

BionIT:
> I'm coming here after watching video attached in screenshot. Would it be possible to consider adding a checkbox or an alternative option for selecting 'overwrite all,' instead of having to click 'overwrite' 15 times? This could potentially enhance the user experience. cc: @kgcreative

@bandinib-amzn The reason user has to act on the action multiple times is because `Request action on conflict` is selected.  If user selects `Automatically overwrite conflicts` then all the conflict objects will be overwritten without any additional actions. This is the current behavior of the import function

kgcreative:
> I'm coming here after watching video attached in screenshot. Would it be possible to consider adding a checkbox or an alternative option for selecting 'overwrite all,' instead of having to click 'overwrite' 15 times? This could potentially enhance the user experience. cc: @kgcreative

Yes, this is ~PAINFUL~ today, however, i'm hesitant to add to the scope unless the dev effort is reasonable. Ideally, we should show a selectable list, with easy "Select all/Deselect all/multiselect" options. Presumably each modal in the current experience triggers only when there is a conflict, so this would require a re-think of the flow.

src/plugins/saved_objects_management/public/management_section/objects_table/components/flyout.tsx
ashwin-pc: +1

BionIT:
> We already have a datasource selector component in the data plugin. We should use that and not create a new picker.

As discussed with @ashwin-pc and @mengweieric , we will switch to use the data source selector in 2.13 when it supports the feature requested in https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5790

ashwin-pc:
> IMO, they both serves different purpose, so I would prefer separate picker. @kgcreative I would like to here your thoughts as well.

Thanks for that example. That was a very useful distinction! Then i think my concern is around the confusion in naming. We are calling this and the other component both datasource pickers but have very different uses for each. This one is used specifically to select an OpenSearch cluster while the other is to select what is in thes clusters but also data from other sources. Can we rename this to be a `cluster selector` instead to disambiguate the two?

BionIT:
> > IMO, they both serves different purpose, so I would prefer separate picker. @kgcreative I would like to here your thoughts as well.
> 
> Thanks for that example. That was a very useful distinction! Then i think my concern is around the confusion in naming. We are calling this and the other component both datasource pickers but have very different uses for each. This one is used specifically to select an OpenSearch cluster while the other is to select what is in thes clusters but also data from other sources. Can we rename this to be a `cluster selector` instead to disambiguate the two?

@ashwin-pc I hear you, I don't have a strong opinion, since I will change to use the new data source selector in 2.13, should we go with this name for now? Happy to change it if we feel necessary

ashwin-pc:
> @ashwin-pc I hear you, I don't have a strong opinion, since I will change to use the new data source selector in 2.13, should we go with this name for now? Happy to change it if we feel necessary

based on Bandini's reply i'm actually convinced that the two might not need to be the same thing if the only thing you are selecting is the cluster and not the index/indexPattern. But if thats true then we do need to rename this since the two wont be combined in 2.13 or beyond

src/plugins/saved_objects_management/public/management_section/objects_table/components/flyout.tsx
bandinib-amzn: I think if we take care of [below comment](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5781#discussion_r1476707363), then we won't have lint errors.

bandinib-amzn:
> e import function

Yes, I completely agree. It doesn't need to be within the scope of this PR. It's an enhancement that we can address in the future. I'll create an issue to track it.

src/plugins/saved_objects_management/public/management_section/objects_table/components/flyout.tsx
Flyingliuhub: I'm wondering if there any tests related to this in the [function test](https://github.com/opensearch-project/opensearch-dashboards-functional-test) repo if we changed the default behavior

Flyingliuhub, line 85: I'm wondering if there any tests related to this in the [function test ](https://github.com/opensearch-project/opensearch-dashboards-functional-test)repo if we changed the default behavior

src/plugins/saved_objects_management/public/management_section/objects_table/components/flyout.tsx
BionIT: Moved export to top level and this line is removed in new commit

src/plugins/saved_objects_management/public/management_section/objects_table/components/flyout.tsx
BionIT: Changed component name to cluster selector and also moved export to top level in new commit

src/plugins/saved_objects_management/public/management_section/objects_table/components/flyout.tsx
BionIT: The FT repo doesn't have tests for import saved objects, and the dashboard repo has functional tests which have been fixed to adopt the change of default behavior

BionIT:
> Can we also add some tests for the code added here?

Added tests in new commit

Flyingliuhub (approved):

PR #5851, opened 2024-02-08
[Multi data source] Add interfaces to register add-on authentication method from plug-in module

### LINKED ISSUE #5692: [Proposal] Refactoring data source plugin to support add-on authentication method with plug-in module
# Overview
The purpose of this design document is to outline the approach and specifications for adding an interface in multi data source plugin that allows the seamless integration of authentication method through plugin. This enhancement aims to provide a standardized and user-friendly mechanism for extending the authentication in multi data source plugin.

# User stories
Currently in multi data source, we support these auth types

* No Auth
* Basic Auth
* AWS SigV4 

In AWS SigV4, we only support user based access policies where IAM user comes with IAM access key & secret key that can be used to sign the request.

As a user I want additional support for role based authentication for my OpenSearch domain. Data source plugin already supports AWS SigV4 for authenticating request. I want to add required UI fields and generate access toke, secret key and session token in external plugin and pass them to data source plugin to authenticating request. As a user I want interface in data source plugin to register my own authentication method.

# Requirement
## Functional
* **Modularity**:  Enable OpenSearch-Dashboards to easily accommodate additional authentication methods through external plugins.
* **Standardization**: Establish set of APIs to ensure uniformity and compatibility with existing authentication method.

## Non-functional

* **Scalability**: The system should be able to handle a growing number of authentication methods through plugins without a significant degradation in performance. It should support a scalable architecture that can accommodate an increasing concurrently active plugins.
* **Usability**:  The interface should have a consistent and intuitive design, making it easy for users to manage plugins.
* **Security and Compliance:** Implement security best practices to protect against potential vulnerabilities in plugins. Ensure that the interface and plugins comply with relevant security standards and regulations.

## Architecture Proposal
To meet the outlined requirements, we will introduce a new interface in data source and data source management plugins. These plugins are OpenSearch Dashboards core plugins. This will extend the current multi data source plugin’s capabilities, enabling it to be able to support multiple authentication method.

### Components
* **Server side auth registry** : This map will hold all the details required for signing the request.
* **DataSourcePlugin** : The proposed architecture will add method to DataSourcePluginSetup to register authentication method and to DataSourcePluginStart to retrieve auth registry . This will be in charge of returning OpenSearch client for provided credentials by plugin.
* **Client side auth registry** : This map will hold UI component required for authentication method.
* **DataSourceManagementPlugin**: The proposed architecture will add method to DataSourceManagementPluginSetup to register authentication method with UI elements and to DataSourceManagementPluginStart to retrieve auth registry. This will be in charge of mounting UI component for user input.
* **Plugin**: Plugin should add dataSource and dataSourceManagement as requiredPlugins in plugin’s JSON manifest file named opensearch_dashboards.json. requiredPlugins is an optional list of the other plugins that must be installed and enabled for this plugin to function properly. During setup in both browser-side and server-side plugin, it should register the authentication method to above core plugin.

### Location of Datasource Codebase

The proposed architecture will comprise a set of back-end functionalities sitting within the server [folder](https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/src/plugins/data_source/server) of the core data_source plugin and a set of front-end functionalities sitting within the public [folder](https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/src/plugins/data_source_management/public) of the core data_source_management plugin. The interfaces to register authentication method into the registry and get authentication registry will be made accessible via the setup and start hook in plugin.ts. Any core components or plugins that list dataSource and dataSourceManagement as a dependency will have access to these interfaces. During plugin setup stage, it initiates registry to store authentication methods and provide function to register authentication method. When the plugin starts, it returns function to retrieve the authentication registry. During setup stage of both  dataSource 
and  dataSourceManagement  plugins, we need authentication registry to pass to async handler. We are using core. getStartService which allows plugins to get access to APIs available in start inside async handlers. Promise will not resolve until Core and plugin dependencies have completed start. This should only be used inside handlers registered during setup that will only be executed after start lifecycle.

```
# src/plugins/data_source/server/plugin.ts

public async setup(core: CoreSetup<DataSourcePluginStart>) {
    const authRegistryPromise = core.getStartServices().then(([, , selfStart]) => {
      const dataSourcePluginStart = selfStart as DataSourcePluginStart;
      return dataSourcePluginStart.getAuthenticationMethodRegistery();
    });
    .
    .
    .
    const registerCredentialProvider = (method: AuthenticationMethod) => {
      if (this.started) {
        throw new Error('cannot call `registerCredentialProvider` after service startup.');
      }
      this.authMethodsRegistry.registerAuthenticationMethod(method);
    };

    return {
      createDataSourceError: (e: any) => createDataSourceError(e),
      registerCredentialProvider, #This is new API
    };
}

public start(core: CoreStart) {
    this.started = true;
    return {
      getAuthenticationMethodRegistery: () => this.authMethodsRegistry,
    };
  }
```
```
# src/plugins/data_source_management/public/plugin.ts

public setup(core: CoreSetup<DataSourceManagementPluginStart>) {
    const registerAuthenticationMethod = (authMethod: AuthenticationMethod) => {
      if (this.started) {
        throw new Error(
          'cannot call `registerAuthenticationMethod` after data source management startup.'
        );
      }
      this.authMethodsRegistry.registerAuthenticationMethod(authMethod);
    };

    return { registerAuthenticationMethod };
}

public start(core: CoreStart) {
    this.started = true;
    return {
      getAuthenticationMethodRegistery: () => this.authMethodsRegistry,
    };
}
```

#### Interfaces and Classes
**Authentication method registry:**
We will maintain two registry to hold authentication method information. One in server which holds the credential provider and one in public which holds the UI elements to mount on web page.
```
# Registry for server-side

export type IAuthenticationMethodRegistery = Omit<
  AuthenticationMethodRegistery,
  'registerAuthenticationMethod'
>;

export class AuthenticationMethodRegistery {
  private readonly authMethods = new Map<string, AuthenticationMethod>();
  /**
   * Register a authMethods with function to return credentials inside the registry.
   * Authentication Method can only be registered once. subsequent calls with the same method name will throw an error.
   */
  public registerAuthenticationMethod(method: AuthenticationMethod) {
    if (this.authMethods.has(method.name)) {
      throw new Error(`Authentication method '${method.name}' is already registered`);
    }
    this.authMethods.set(method.name, deepFreeze(method) as AuthenticationMethod);
  }

  public getAllAuthenticationMethods() {
    return [...this.authMethods.values()];
  }

  public getAuthenticationMethod(name: string) {
    return this.authMethods.get(name);
  }
}
```

```
# Registry for browser-side

export type IAuthenticationMethodRegistery = Omit<
  AuthenticationMethodRegistery,
  'registerAuthenticationMethod'
>;

export class AuthenticationMethodRegistery {
  private readonly authMethods = new Map<string, AuthenticationMethod>();
  /**
   * Register a authMethods with function to return credentials inside the registry.
   * Authentication Method can only be registered once. subsequent calls with the same method name will throw an error.
   */
  public registerAuthenticationMethod(method: AuthenticationMethod) {
    if (this.authMethods.has(method.name)) {
      throw new Error(`Authentication method '${method.name}' is already registered`);
    }
    this.authMethods.set(method.name, deepFreeze(method) as AuthenticationMethod);
  }

  public getAllAuthenticationMethods() {
    return [...this.authMethods.values()];
  }

  public getAuthenticationMethod(name: string) {
    return this.authMethods.get(name);
  }
}
```

**data_source plugin**
1] **DataSourceCredentialsProvider**: Instance of DataSourceCredentialsProvider will be responsible for providing credentials to sign the request either with Basic auth or AWS SigV4 which are only supported request signing mechanism.
```
export interface DataSourceCredentialsProviderOptions {
  dataSourceAttr: DataSourceAttributes;
  request?: OpenSearchDashboardsRequest;
  cryptography?: CryptographyServiceSetup;
}

export type DataSourceCredentialsProvider = (
  options: DataSourceCredentialsProviderOptions
) => Promise<UsernamePasswordTypedContent | SigV4Content>;
```

2] **DataSourceAttributes**:  Right now we only support [UsernamePasswordTypedContent](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source/common/data_sources/types.ts#L31-L34), [SigV4Content](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source/common/data_sources/types.ts#L24-L29) and undefined as type for auth credentials. We will need to support more generic type along with existing types to allow credentials in the form of key:value pair. Also we will add new field name: AuthType | string  in   [DataSourceAttributes](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source/common/data_sources/types.ts#L8-L17) to identify the authentication method. To ensure compatibility with existing auth methods, we will make default value same as type for Basic and SigV4 authentication method. auth.type in DataSourceAttributes will have only values of AuthType. auth.type will tell data_source plugin, using which method request needs to be signed. Right now we only have two ways: Basic authentication and AWS SigV4 authentication to sign the request.

```
export interface AuthTypeContent {
  [key: string]: string;
}

export interface DataSourceAttributes extends SavedObjectAttributes {
  title: string;
  description?: string;
  endpoint: string;
  auth: {
    type: AuthType;
    credentials: UsernamePasswordTypedContent | SigV4Content | undefined | **AuthTypeContent**;
  };
  lastUpdatedTime?: string;
  **name: AuthType | string;**
}
```
3] **SigV4Content**: This proposed architecture will also extend the ability of multi data source to support authentication using IAM role. Current architecture only supports authenticating as IAM user. As IAM role session requires temporary security credentials through assuming role, we will need additional field for sessionToken in [SigV4Content](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source/common/data_sources/types.ts#L24-L29) interface.
```
export interface SigV4Content extends SavedObjectAttributes {
  accessKey: string;
  secretKey: string;
  region: string;
  service?: SigV4ServiceName;
  **sessionToken?: string;** # Newly added field to support temporary security credentials.
}
```
4] **Client**: Based on authentication method type, client will look for authentication method in auth registry to get credentials. Client pooling and caching will be handled by data_source core plugin.
```
# src/plugins/data_source/server/client/configure_client.ts

export const getQueryClient = async (): Promise<Client> => {
   awsCredential = authMethod?.credentialProvider({ dataSourceAttr, request, cryptography });
   type = authMethod?.authType;
   .
   .
   .
   return client; # bases on auth type.
}
```
**data_source_management plugin**
1] **AuthenticationMethod**: To support custom authentication method from plugin, browser side we need option for selection and then form to enter fields which would be credentials or fields which required to retrieve credentials in case role based authentication. Plugins or any component which will be using these API, need to take care of validations as well for those fields. Core data_source_management plugin will be in charge to mount those UI component in parent form and calls sever side API to test connection and save the data source.
```
export interface AuthenticationMethod {
  name: string;
  credentialForm: React.JSX.Element;
  credentialSourceOption: EuiSuperSelectOption<string>;
}

# Map storing UI elements for custom authentication method
authMethods = new Map<string, AuthenticationMethod>();
```

## Tracking issue
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5838
## Follow up Issue
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5694

bandinib-amzn:
### Description

This PR adds the interfaces to register authentication method into the registry and get authentication registry. These interfaces are accessible via the setup and start hook in `plugin.ts`. Any core components or plugins that list dataSource and dataSourceManagement as a dependency will have access to these interfaces.

### Issues Resolved

Partially resolves #5692 

### Check List

- [X] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source/server/auth_registry/authentication_methods_registry.test.ts
+++ b/src/plugins/data_source/server/auth_registry/authentication_methods_registry.test.ts
@@ -1,0 +1,106 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { AuthenticationMethodRegistery } from './authentication_methods_registry';
+import { AuthMethodType } from '../../server/types';
+import { AuthType } from '../../common/data_sources';
+
+const createAuthenticationMethod = (authMethod: Partial<AuthMethodType>): AuthMethodType => ({
+  name: 'unknown',
+  authType: AuthType.NoAuth,
+  credentialProvider: jest.fn(),
+  ...authMethod,
+});
+
+describe('AuthenticationMethodRegistery', () => {
+  let registry: AuthenticationMethodRegistery;
+
+  beforeEach(() => {
+    registry = new AuthenticationMethodRegistery();
+  });
+
+  it('allows to register authentication method', () => {
+    registry.registerAuthenticationMethod(createAuthenticationMethod({ name: 'typeA' }));
+    registry.registerAuthenticationMethod(createAuthenticationMethod({ name: 'typeB' }));
+    registry.registerAuthenticationMethod(createAuthenticationMethod({ name: 'typeC' }));
+
+    expect(
+      registry
+        .getAllAuthenticationMethods()
+        .map((type) => type.name)
+        .sort()
+    ).toEqual(['typeA', 'typeB', 'typeC']);
+  });
+
+  it('throws when trying to register the same authentication method twice', () => {
+    registry.registerAuthenticationMethod(createAuthenticationMethod({ name: 'typeA' }));
+    registry.registerAuthenticationMethod(createAuthenticationMethod({ name: 'typeB' }));
+    expect(() => {
+      registry.registerAuthenticationMethod(createAuthenticationMethod({ name: 'typeA' }));
+    }).toThrowErrorMatchingInlineSnapshot(`"Authentication method 'typeA' is already registered"`);
+  });
+
+  describe('#getAuthenticationMethod', () => {
+    it(`retrieve a type by it's name`, () => {
+      const typeA = createAuthenticationMethod({ name: 'typeA' });
+      const typeB = createAuthenticationMethod({ name: 'typeB' });
+      registry.registerAuthenticationMethod(typeA);
+      registry.registerAuthenticationMethod(typeB);
+      registry.registerAuthenticationMethod(createAuthenticationMethod({ name: 'typeC' }));
+
+      expect(registry.getAuthenticationMethod('typeA')).toEqual(typeA);
+      expect(registry.getAuthenticationMethod('typeB')).toEqual(typeB);
+      expect(registry.getAuthenticationMethod('unknownType')).toBeUndefined();
+    });
+
+    it('forbids to mutate the registered types', () => {
+      registry.registerAuthenticationMethod(
+        createAuthenticationMethod({
+          name: 'typeA',
+          authType: AuthType.NoAuth,
+        })
+      );
+
+      const typeA = registry.getAuthenticationMethod('typeA')!;
+
+      expect(() => {
+        typeA.authType = AuthType.SigV4;
+      }).toThrow();
+      expect(() => {
+        typeA.name = 'foo';
+      }).toThrow();
+      expect(() => {
+        typeA.credentialProvider = jest.fn();
+      }).toThrow();
+    });
+  });
+
+  describe('#getAllTypes', () => {
+    it('returns all registered types', () => {
+      const typeA = createAuthenticationMethod({ name: 'typeA' });
+      const typeB = createAuthenticationMethod({ name: 'typeB' });
+      const typeC = createAuthenticationMethod({ name: 'typeC' });
+      registry.registerAuthenticationMethod(typeA);
+      registry.registerAuthenticationMethod(typeB);
+
+      const registered = registry.getAllAuthenticationMethods();
+      expect(registered.length).toEqual(2);
+      expect(registered).toContainEqual(typeA);
+      expect(registered).toContainEqual(typeB);
+      expect(registered).not.toContainEqual(typeC);
+    });
+
+    it('does not mutate the registered types when altering the list', () => {
+      registry.registerAuthenticationMethod(createAuthenticationMethod({ name: 'typeA' }));
+      registry.registerAuthenticationMethod(createAuthenticationMethod({ name: 'typeB' }));
+      registry.registerAuthenticationMethod(createAuthenticationMethod({ name: 'typeC' }));
+
+      const types = registry.getAllAuthenticationMethods();
+      types.splice(0, 3);
+
+      expect(registry.getAllAuthenticationMethods().length).toEqual(3);
+    });
+  });
+});
--- a/src/plugins/data_source/server/auth_registry/authentication_methods_registry.ts
+++ b/src/plugins/data_source/server/auth_registry/authentication_methods_registry.ts
@@ -1,0 +1,33 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { AuthMethodValues } from '../../server/types';
+
+export type IAuthenticationMethodRegistery = Omit<
+  AuthenticationMethodRegistery,
+  'registerAuthenticationMethod'
+>;
+
+export class AuthenticationMethodRegistery {
+  private readonly authMethods = new Map<string, AuthMethodValues>();
+  /**
+   * Register a authMethods with function to return credentials inside the registry.
+   * Authentication Method can only be registered once. subsequent calls with the same method name will throw an error.
+   */
+  public registerAuthenticationMethod(name: string, authMethodValues: AuthMethodValues) {
+    if (this.authMethods.has(name)) {
+      throw new Error(`Authentication method '${name}' is already registered`);
+    }
+    this.authMethods.set(name, authMethodValues);
+  }
+
+  public getAllAuthenticationMethods() {
+    return [...this.authMethods.values()];
+  }
+
+  public getAuthenticationMethod(name: string) {
+    return this.authMethods.get(name);
+  }
+}
--- a/src/plugins/data_source/server/auth_registry/authentication_methods_registry.ts
+++ b/src/plugins/data_source/server/auth_registry/authentication_methods_registry.ts
@@ -4,7 +4,7 @@
  */
 
 import { deepFreeze } from '@osd/std';
-import { AuthMethodType } from '../../server/types';
+import { AuthenticationMethod } from '../../server/types';
 
 export type IAuthenticationMethodRegistery = Omit<
   AuthenticationMethodRegistery,
@@ -12,16 +12,16 @@ export type IAuthenticationMethodRegistery = Omit<
 >;
 
 export class AuthenticationMethodRegistery {
-  private readonly authMethods = new Map<string, AuthMethodType>();
+  private readonly authMethods = new Map<string, AuthenticationMethod>();
   /**
    * Register a authMethods with function to return credentials inside the registry.
    * Authentication Method can only be registered once. subsequent calls with the same method name will throw an error.
    */
-  public registerAuthenticationMethod(method: AuthMethodType) {
+  public registerAuthenticationMethod(method: AuthenticationMethod) {
     if (this.authMethods.has(method.name)) {
       throw new Error(`Authentication method '${method.name}' is already registered`);
     }
-    this.authMethods.set(method.name, deepFreeze(method) as AuthMethodType);
+    this.authMethods.set(method.name, deepFreeze(method) as AuthenticationMethod);
   }
 
   public getAllAuthenticationMethods() {
--- a/src/plugins/data_source/server/auth_registry/authentication_methods_registry.ts
+++ b/src/plugins/data_source/server/auth_registry/authentication_methods_registry.ts
@@ -3,7 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { AuthMethodValues } from '../../server/types';
+import { deepFreeze } from '@osd/std';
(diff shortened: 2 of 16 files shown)


Flyingliuhub:
@bandinib-amzn I saw ciGroup2 was still failed, could you please take a look if related to your changes. Thanks

src/plugins/data_source/server/auth_registry/authentication_methods_registry.test.ts
```diff
@@ -0,0 +1,108 @@
...
+});
+
+describe('AuthenticationMethodRegistery', () => {
+  let registry: AuthenticationMethodRegistery;
+
+  beforeEach(() => {
+    registry = new AuthenticationMethodRegistery();
+  });
+
+  it('allows to register authentication method', () => {
```
Flyingliuhub, line 26: Is it allows to register authentication `methods`? Thanks

src/plugins/data_source/server/auth_registry/authentication_methods_registry.test.ts
```diff
@@ -0,0 +1,108 @@
...
+        .sort()
+    ).toEqual(['typeA', 'typeB', 'typeC']);
+  });
+
+  it('throws when trying to register the same authentication method twice', () => {
+    registry.registerAuthenticationMethod(createAuthenticationMethod({ name: 'typeA' }));
+    registry.registerAuthenticationMethod(createAuthenticationMethod({ name: 'typeB' }));
+    expect(() => {
+      registry.registerAuthenticationMethod(createAuthenticationMethod({ name: 'typeA' }));
+    }).toThrowErrorMatchingInlineSnapshot(`"Authentication method 'typeA' is already registered"`);
```
Flyingliuhub, line 44: do we cover the test cases for `typea` and limit how many method can be registered? Thanks

bandinib-amzn: There is no limit on authentication methods. It just restricts to add authentication method with same name. Because name is key.

src/plugins/data_source/server/routes/test_connection.ts
```diff
@@ -9,11 +9,13 @@ import { AuthType, DataSourceAttributes, SigV4ServiceName } from '../../common/d
...
 import { DataSourceServiceSetup } from '../data_source_service';
 import { CryptographyServiceSetup } from '../cryptography_service';
+import { IAuthenticationMethodRegistery } from '../auth_registry';
 
 export const registerTestConnectionRoute = (
   router: IRouter,
   dataSourceServiceSetup: DataSourceServiceSetup,
-  cryptography: CryptographyServiceSetup
+  cryptography: CryptographyServiceSetup,
+  authRegistryPromise: Promise<IAuthenticationMethodRegistery>
```
xinruiba, line 18: Since we update "dataSourceAttr" to support "AuthTypeContent". Do we plan to also update the "dataSourceAttr" here in validate object?

bandinib-amzn: Yes. This PR only covers adding interfaces. I will cover usage of auth registry and refactoring of existing code in following PR.

Flyingliuhub (approved):

xinruiba (approved):

ZilongX (approved):

src/plugins/data_source/server/auth_registry/authentication_methods_registry.ts
```diff
@@ -0,0 +1,34 @@
...
+   * Authentication Method can only be registered once. subsequent calls with the same method name will throw an error.
+   */
+  public registerAuthenticationMethod(method: AuthenticationMethod) {
+    if (this.authMethods.has(method.name)) {
+      throw new Error(`Authentication method '${method.name}' is already registered`);
+    }
+    this.authMethods.set(method.name, deepFreeze(method) as AuthenticationMethod);
+  }
+
+  public getAllAuthenticationMethods() {
```
zengyan-amazon, line 27: we can consider to make these get function to return readonly values

src/plugins/data_source_management/public/auth_registry/authentication_methods_registry.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export class AuthenticationMethodRegistery {
+  private readonly authMethods = new Map<string, AuthenticationMethod>();
+  /**
+   * Register a authMethods with function to return credentials inside the registry.
+   * Authentication Method can only be registered once. subsequent calls with the same method name will throw an error.
+   */
+  public registerAuthenticationMethod(method: AuthenticationMethod) {
+    if (this.authMethods.has(method.name)) {
+      throw new Error(`Authentication method '${method.name}' is already registered`);
```
zengyan-amazon, line 28: just to confirm this will fail the OSD bootstrap right? we should ensure OSD won't start if we see duplicated auth methods

src/plugins/data_source/server/auth_registry/authentication_methods_registry.ts
bandinib-amzn: While adding authentication method in map, we do deepFreeze object and then store. 

[deepFreeze](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/packages/osd-std/README.md#deepfreeze) converts the return type to Readonly variant recursively.

so by default getAllAuthenticationMethods returns readonly values. I have added [UT](https://github.com/bandinib-amzn/OpenSearch-Dashboards/blob/main/src/plugins/data_source/server/auth_registry/authentication_methods_registry.test.ts#L97-L108) to validate that as well.

src/plugins/data_source_management/public/auth_registry/authentication_methods_registry.ts
bandinib-amzn: Just to confirm here you are not talking about `yarn osd bootstrap`, right? 

`yarn osd bootstrap` is successful. 
`yarn start` fails. But it failing with server crash with following error. That needs to be handled properly. I will handle in following PR.
```
 FATAL  Error: Authentication method 'token_exchange' is already registered

 server crashed  with status code 1
```

Issue #5796, opened 2024-02-03
[RFC] Introduce Application Configuration Service to support hot reloading


tianleh:
# Problem Statement <a href="#problem-statement" id="problem-statement">#</a>

OpenSearch Dashboards (OSD) currently provides two ways to make configuration changes. The first one is to use OSD YML file `opensearch_dashboards.yml` (one example is to configure the OSD index name). The second is to use advanced settings page (one example is to configure dark mode). Recently there is an emerging requirement to have an application level configuration mechanism. These configurations may need to take effect without system restart and are not user level configuration. Besides supporting use cases from OSD core, we also plan to support OSD plugins use cases. The last factor to consider is the storage database. While current OSD assumes a default OpenSearch as database, we plan to have an extensible solution where other databases can be used.

We will use this issue to explain why the existing features won't meet the requirements and proposal a new solution which will abstract an interface and web APIs in order to decouple with storage, while we need a implementation of this interface for opensearch. The solution is not a direct extension of existing configurations in YML. Instead each use case developer will decide whether they want to pick YML configurations or dynamic configuration (or have both) and implement logic to use these configurations.

# Desired State <a href="#desired-state" id="desired-state">#</a>

- The configuration updates shall be "hot reload" which means that they will take effect without waiting for OSD to restart.
- Both OSD core and plugins could use it to read and write their application level configurations.
- It shall be feasible to use a different storage for the configurations other than the default OpenSearch.
- When FGAC is enabled, there shall be permission control over updates of the configurations.

# Out of Scope <a href="#out-of-scope" id="out-of-scope">#</a>
- We do not plan to provide new UX for reading and writing the configurations
- We do not plan to move all existing configurations in YML or advanced settings to the new application configuration

# Existing Limitations <a href="#existing-limitations" id="existing-limitations">#</a>
We will take a close look at the current limitations.

## OSD YML <a href="#osd-yml" id="osd-yml">#</a>
Only system administrators who have access to OSD YML could update the values. The changes need a server restart to take effect. It will cause the risk of service availability interruption. Use a cluster with 10 hosts running OSD as an example. The system administrator will need to connect to each of the host, update the YML with new value and then restart OSD. Note that regular OSD customers only need to access OSD through endpoint without even knowing the OSD's running environment under the hood.

Also OSD YML controls the static application level configurations and doesn't support user specific configurations.

## Advanced Settings <a href="#advanced-settings" id="advanced-settings">#</a>

[...]

PR #5855, opened 2024-02-09
Add support for dynamic application configurations

tianleh:
### Description
The PR is to add application configuration service. See related issue for context. The code is mostly extracted from my other PR https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5641 Once the current PR is merged, I will update the related PR to use the function provided by current PR.

**We plan to release this feature in `2.13.0` and may need a backport to `2.x` branch.**

**Note to the maintainers: please help add the backport label for `2.x`**

### Issues Resolved
closes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5796
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [Multiple Datasource] Refactor client and legacy client to use authentication registry ([#5881](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5881))
 - [Multiple Datasource] Improved error handling for the search API when a null value is passed for the dataSourceId ([#5882](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5882))
 - [Multiple Datasource] Hide/Show authentication method in multi data source plugin based on configuration ([#5916](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5916))
-- Add support for dynamic application configurations ([#5855](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855))
+- [[Dynamic Configurations] Add support for dynamic application configurations ([#5855](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855))
 
 ### 🐛 Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [Multiple Datasource] Refactor client and legacy client to use authentication registry ([#5881](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5881))
 - [Multiple Datasource] Improved error handling for the search API when a null value is passed for the dataSourceId ([#5882](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5882))
 - [Multiple Datasource] Hide/Show authentication method in multi data source plugin based on configuration ([#5916](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5916))
+- Add support for dynamic application configurations ([#5855](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855))
 
 ### 🐛 Bug Fixes
 
--- a/config/opensearch_dashboards.yml
+++ b/config/opensearch_dashboards.yml
@@ -29,6 +29,13 @@
 # dashboards. OpenSearch Dashboards creates a new index if the index doesn't already exist.
 #opensearchDashboards.index: ".opensearch_dashboards"
 
+# OpenSearch Dashboards uses an index in OpenSearch to store dynamic configurations.
+# This shall be a different index from opensearchDashboards.index.
+# opensearchDashboards.configIndex: ".opensearch_dashboards_config"
+
+# Set the value of this setting to true to enable plugin application config. By default it is disabled.
+# application_config.enabled: false
+
 # The default application to load.
 #opensearchDashboards.defaultAppId: "home"
 
--- a/config/opensearch_dashboards.yml
+++ b/config/opensearch_dashboards.yml
@@ -31,7 +31,7 @@
 
 # OpenSearch Dashboards uses an index in OpenSearch to store dynamic configurations.
 # This shall be a different index from opensearchDashboards.index.
-# opensearchDashboards.config_index: ".opensearch_dashboards_config"
+# opensearchDashboards.configIndex: ".opensearch_dashboards_config"
 
 # The default application to load.
 #opensearchDashboards.defaultAppId: "home"
--- a/config/opensearch_dashboards.yml
+++ b/config/opensearch_dashboards.yml
@@ -33,6 +33,9 @@
 # This shall be a different index from opensearchDashboards.index.
 # opensearchDashboards.configIndex: ".opensearch_dashboards_config"
 
+# Set the value of this setting to true to enable plugin application config. By default it is disabled.
+# application_config.enabled: false
+
 # The default application to load.
 #opensearchDashboards.defaultAppId: "home"
 
@@ -290,6 +293,3 @@
 
 # Set the value of this setting to true to enable plugin augmentation on Dashboard
 # vis_augmenter.pluginAugmentationEnabled: true
-
-# Set the value of this setting to true to enable plugin application config. By default it is disabled.
-# application_config.enabled: false
--- a/config/opensearch_dashboards.yml
+++ b/config/opensearch_dashboards.yml
@@ -29,6 +29,10 @@
 # dashboards. OpenSearch Dashboards creates a new index if the index doesn't already exist.
 #opensearchDashboards.index: ".opensearch_dashboards"
 
+# OpenSearch Dashboards uses an index in OpenSearch to store dynamic configurations.
+# This shall be a different index from opensearchDashboards.index.
+# opensearchDashboards.config_index: ".opensearch_dashboards_config"
+
 # The default application to load.
 #opensearchDashboards.defaultAppId: "home"
 
--- a/config/opensearch_dashboards.yml
+++ b/config/opensearch_dashboards.yml
@@ -290,3 +290,6 @@
 
 # Set the value of this setting to true to enable plugin augmentation on Dashboard
 # vis_augmenter.pluginAugmentationEnabled: true
+
+# Set the value of this setting to true to enable plugin application config. By default it is disabled.
+# application_config.enabled: false
--- a/config/opensearch_dashboards.yml
+++ b/config/opensearch_dashboards.yml
@@ -29,13 +29,6 @@
 # dashboards. OpenSearch Dashboards creates a new index if the index doesn't already exist.
 #opensearchDashboards.index: ".opensearch_dashboards"
 
-# OpenSearch Dashboards uses an index in OpenSearch to store dynamic configurations.
-# This shall be a different index from opensearchDashboards.index.
-# opensearchDashboards.configIndex: ".opensearch_dashboards_config"
-
-# Set the value of this setting to true to enable plugin application config. By default it is disabled.
-# application_config.enabled: false
-
 # The default application to load.
 #opensearchDashboards.defaultAppId: "home"
 
--- a/src/plugins/application_config/.eslintrc.js
+++ b/src/plugins/application_config/.eslintrc.js
@@ -1,0 +1,7 @@
+module.exports = {
+  root: true,
+  extends: ['@elastic/eslint-config-kibana', 'plugin:@elastic/eui/recommended'],
+  rules: {
+    '@osd/eslint/require-license-header': 'off',
+  },
+};
--- a/src/plugins/application_config/.eslintrc.js
+++ b/src/plugins/application_config/.eslintrc.js
@@ -1,7 +1,0 @@
-module.exports = {
-  root: true,
-  extends: ['@elastic/eslint-config-kibana', 'plugin:@elastic/eui/recommended'],
-  rules: {
-    '@osd/eslint/require-license-header': 'off',
-  },
-};
--- a/src/plugins/application_config/.eslintrc.js
+++ b/src/plugins/application_config/.eslintrc.js
@@ -1,0 +1,6 @@
+module.exports = {
+  root: true,
+  extends: ['@elastic/eslint-config-kibana', 'plugin:@elastic/eui/recommended'],
+  rules: {
+  },
+};
--- a/src/plugins/application_config/.eslintrc.js
+++ b/src/plugins/application_config/.eslintrc.js
@@ -1,6 +1,0 @@
-module.exports = {
-  root: true,
-  extends: ['@elastic/eslint-config-kibana', 'plugin:@elastic/eui/recommended'],
-  rules: {
-  },
-};
--- a/src/plugins/application_config/.i18nrc.json
+++ b/src/plugins/application_config/.i18nrc.json
@@ -1,7 +1,0 @@
-{
-  "prefix": "applicationConfig",
-  "paths": {
-    "applicationConfig": "."
-  },
-  "translations": ["translations/ja-JP.json"]
-}
--- a/src/plugins/application_config/.i18nrc.json
+++ b/src/plugins/application_config/.i18nrc.json
@@ -1,0 +1,7 @@
+{
+  "prefix": "applicationConfig",
+  "paths": {
+    "applicationConfig": "."
+  },
+  "translations": ["translations/ja-JP.json"]
+}
--- a/src/plugins/application_config/README.md
+++ b/src/plugins/application_config/README.md
@@ -1,12 +1,14 @@
-# applicationConfig
+# ApplicationConfig Plugin
 
-A OpenSearch Dashboards plugin
+An OpenSearch Dashboards plugin for application configuration service and OpenSearch implementation.
+
+---
+
+## Introduction
 
 This plugin introduces the support of dynamic application configurations as opposed to the existing static configuration in OSD YAML file `opensearch_dashboards.yml`. It stores the configuration in an index whose default name is `.opensearch_dashboards_config` and could be customized through the key `opensearchDashboards.config_index` in OSD YAML file. Initially the new index does not exist. Only OSD users who need dynamic configurations will create it.
 
 It also provides an interface `ConfigurationClient` for future extensions of external configuration clients. A default implementation based on OpenSearch as database is used.
-
----
 
 ## Configuration
 
--- a/src/plugins/application_config/README.md
+++ b/src/plugins/application_config/README.md
@@ -15,6 +15,7 @@ This plugin is disabled by default.
 ## Configuration
 
 OSD users who want to set up application configurations will first need to enable this plugin by the following line in OSD YML.
+
 ```
 application_config.enabled: true
 
--- a/src/plugins/application_config/README.md
+++ b/src/plugins/application_config/README.md
@@ -1,0 +1,11 @@
+# applicationConfig
+
+A OpenSearch Dashboards plugin
+
+---
+
+## Development
+
+See the [OpenSearch Dashboards contributing
+guide](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/CONTRIBUTING.md) for instructions
+setting up your development environment.
--- a/src/plugins/application_config/README.md
+++ b/src/plugins/application_config/README.md
@@ -6,7 +6,7 @@ An OpenSearch Dashboards plugin for application configuration service and OpenSe
 
 ## Introduction
 
-This plugin introduces the support of dynamic application configurations as opposed to the existing static configuration in OSD YAML file `opensearch_dashboards.yml`. It stores the configuration in an index whose default name is `.opensearch_dashboards_config` and could be customized through the key `opensearchDashboards.config_index` in OSD YAML file. Initially the new index does not exist. Only OSD users who need dynamic configurations will create it.
+This plugin introduces the support of dynamic application configurations as opposed to the existing static configuration in OSD YAML file `opensearch_dashboards.yml`. It stores the configuration in an index whose default name is `.opensearch_dashboards_config` and could be customized through the key `opensearchDashboards.configIndex` in OSD YAML file. Initially the new index does not exist. Only OSD users who need dynamic configurations will create it.
 
 It also provides an interface `ConfigurationClient` for future extensions of external configuration clients. A default implementation based on OpenSearch as database is used.
 
--- a/src/plugins/application_config/README.md
+++ b/src/plugins/application_config/README.md
@@ -1,6 +1,6 @@
 # ApplicationConfig Plugin
 
-An OpenSearch Dashboards plugin for application configuration service and OpenSearch implementation.
+An OpenSearch Dashboards plugin for application configuration and a default implementation based on OpenSearch as storage.
 
 ---
 
--- a/src/plugins/application_config/README.md
+++ b/src/plugins/application_config/README.md
@@ -47,7 +47,17 @@ While a default OpenSearch based client is implemented, OSD users can use extern
 
 Let's call this plugin `MyConfigurationClientPlugin`.
 
-First, this plugin will need to implement a class `MyConfigurationClient` based on interface `ConfigurationClient` defined in the `types.ts` under directory `src/plugins/application_config/server/types.ts`.
+First, this plugin will need to implement a class `MyConfigurationClient` based on interface `ConfigurationClient` defined in the `types.ts` under directory `src/plugins/application_config/server/types.ts`. Below are the functions inside the interface.
+
+```
+  getConfig(): Promise<Map<string, string>>;
+
+  getEntityConfig(entity: string): Promise<string>;
+
+  updateEntityConfig(entity: string, newValue: string): Promise<string>;
+
+  deleteEntityConfig(entity: string): Promise<string>;
+```
 
 Second, this plugin needs to declare `applicationConfig` as its dependency by adding it to `requiredPlugins` in its own `opensearch_dashboards.json`.
 
@@ -77,6 +87,12 @@ Then the plugin will import the new type `AppPluginSetupDependencies` and add to
   }
 
 ```
+
+## Onboarding Configurations
+
+Since the APIs and interfaces can take an entity, a new use case to this plugin could just pass their entity into the parameters. There is no need to implement new APIs or interfaces. To programmatically call the functions in `ConfigurationClient` from a plugin (the caller plugin), below is the code example.
+
+Similar to [section](#external-configuration-clients), a new type `AppPluginSetupDependencies` which encapsulates `ApplicationConfigPluginSetup` is needed. Then it can be imported into the `setup` function of the caller plugin. Then the caller plugin will have access to the `getConfigurationClient` and `registerConfigurationClient` exposed by `ApplicationConfigPluginSetup`.
 
 ## Development
 
--- a/src/plugins/application_config/README.md
+++ b/src/plugins/application_config/README.md
@@ -1,6 +1,6 @@
 # ApplicationConfig Plugin
 
-An OpenSearch Dashboards plugin for application configuration service and OpenSearch implementation.
+An experimental OpenSearch Dashboards plugin for application configuration service and OpenSearch implementation.
 
 ---
 
@@ -10,9 +10,17 @@ This plugin introduces the support of dynamic application configurations as oppo
 
 It also provides an interface `ConfigurationClient` for future extensions of external configuration clients. A default implementation based on OpenSearch as database is used.
 
+This plugin is disabled by default.
+
 ## Configuration
 
-OSD users who want to set up application configurations can perform such through CURL the OSD APIs.
+OSD users who want to set up application configurations will first need to enable this plugin by the following line in OSD YML.
+```
+application_config.enabled: true
+
+```
+
+Then they can perform configuration operations through CURL the OSD APIs.
 
 (Note that the commands following could be first obtained from a copy as curl option from the network tab of a browser development tool and then replaced with the API names)
 
--- a/src/plugins/application_config/README.md
+++ b/src/plugins/application_config/README.md
@@ -2,7 +2,81 @@
 
 A OpenSearch Dashboards plugin
 
+This plugin introduces the support of dynamic application configurations as opposed to the existing static configuration in OSD YAML file `opensearch_dashboards.yml`. It stores the configuration in an index whose default name is `.opensearch_dashboards_config` and could be customized through the key `opensearchDashboards.config_index` in OSD YAML file. Initially the new index does not exist. Only OSD users who need dynamic configurations will create it.
+
+It also provides an interface `ConfigurationClient` for future extensions of external configuration clients. A default implementation based on OpenSearch as database is used.
+
 ---
+
+## Configuration
+
+OSD users who want to set up application configurations can perform such through CURL the OSD APIs.
+
+(Note that the commands following could be first obtained from a copy as curl option from the network tab of a browser development tool and then replaced with the API names)
+
+Below is the CURL command to view all configurations.
+
+```
+curl '{osd endpoint}/api/appconfig' -X GET
+```
+
+Below is the CURL command to view the configuration of an entity.
+
+```
+curl '{osd endpoint}/api/appconfig/{entity}' -X GET
+
+```
+
+Below is the CURL command to update the configuration of an entity.
+
+```
+curl '{osd endpoint}/api/appconfig/{entity}' -X POST -H 'Accept: application/json' -H 'Content-Type: application/json' -H 'osd-xsrf: osd-fetch' -H 'Sec-Fetch-Dest: empty' --data-raw '{"newValue":"{new value}"}'
+```
+
+Below is the CURL command to delete the configuration of an entity.
+
+```
+curl '{osd endpoint}/api/appconfig/{entity}' -X DELETE -H 'osd-xsrf: osd-fetch' -H 'Sec-Fetch-Dest: empty'
+
+```
+
+
+## External Configuration Clients
+
+While a default OpenSearch based client is implemented, OSD users can use external configuration clients through an OSD plugin (outside OSD).
+
+Let's call this plugin `MyConfigurationClientPlugin`.
+
+First, this plugin will need to implement a class `MyConfigurationClient` based on interface `ConfigurationClient` defined in the `types.ts` under directory `src/plugins/application_config/server/types.ts`.
+
+Second, this plugin needs to declare `applicationConfig` as its dependency by adding it to `requiredPlugins` in its own `opensearch_dashboards.json`.
+
+Third, the plugin will define a new type called `AppPluginSetupDependencies` as follows in its own `types.ts`.
+
+```
+export interface AppPluginSetupDependencies {
+  applicationConfig: ApplicationConfigPluginSetup;
+}
+
+```
+
+Then the plugin will import the new type `AppPluginSetupDependencies` and add to its own setup input. Below is the skeleton of the class `MyConfigurationClientPlugin`.
+
+```
+// MyConfigurationClientPlugin
+  public setup(core: CoreSetup, { applicationConfig }: AppPluginSetupDependencies) {
+
+    ...
+    // The function createClient provides an instance of ConfigurationClient which
+    // could have a underlying DynamoDB or Postgres implementation.
+    const myConfigurationClient: ConfigurationClient = this.createClient();
+
+    applicationConfig.registerConfigurationClient(myConfigurationClient);
+     ...
+    return {};
+  }
+
+```
 
 ## Development
 
--- a/src/plugins/application_config/README.md
+++ b/src/plugins/application_config/README.md
@@ -1,6 +1,6 @@
 # ApplicationConfig Plugin
 
-An experimental OpenSearch Dashboards plugin for application configuration service and OpenSearch implementation.
+An OpenSearch Dashboards plugin for application configuration service and OpenSearch implementation.
 
 ---
 
--- a/src/plugins/application_config/common/index.ts
+++ b/src/plugins/application_config/common/index.ts
@@ -1,2 +1,7 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 export const PLUGIN_ID = 'applicationConfig';
 export const PLUGIN_NAME = 'applicationConfig';
--- a/src/plugins/application_config/common/index.ts
+++ b/src/plugins/application_config/common/index.ts
@@ -4,4 +4,4 @@
  */
 
 export const PLUGIN_ID = 'applicationConfig';
-export const PLUGIN_NAME = 'applicationConfig';
+export const PLUGIN_NAME = 'application_config';
--- a/src/plugins/application_config/common/index.ts
+++ b/src/plugins/application_config/common/index.ts
@@ -1,0 +1,2 @@
+export const PLUGIN_ID = 'applicationConfig';
+export const PLUGIN_NAME = 'applicationConfig';
--- a/src/plugins/application_config/opensearch_dashboards.json
+++ b/src/plugins/application_config/opensearch_dashboards.json
@@ -1,6 +1,6 @@
 {
   "id": "applicationConfig",
-  "version": "1.0.0",
+  "version": "opensearchDashboards",
   "opensearchDashboardsVersion": "opensearchDashboards",
   "server": true,
   "ui": false,
--- a/src/plugins/application_config/opensearch_dashboards.json
+++ b/src/plugins/application_config/opensearch_dashboards.json
@@ -4,6 +4,6 @@
   "opensearchDashboardsVersion": "opensearchDashboards",
   "server": true,
   "ui": false,
-  "requiredPlugins": ["navigation"],
+  "requiredPlugins": [],
   "optionalPlugins": []
 }
--- a/src/plugins/application_config/opensearch_dashboards.json
+++ b/src/plugins/application_config/opensearch_dashboards.json
@@ -1,0 +1,9 @@
+{
+  "id": "applicationConfig",
+  "version": "1.0.0",
+  "opensearchDashboardsVersion": "opensearchDashboards",
+  "server": true,
+  "ui": false,
+  "requiredPlugins": ["navigation"],
+  "optionalPlugins": []
+}
--- a/src/plugins/application_config/server/index.ts
+++ b/src/plugins/application_config/server/index.ts
@@ -1,0 +1,11 @@
+import { PluginInitializerContext } from '../../../core/server';
+import { ApplicationConfigPlugin } from './plugin';
+
+// This exports static code and TypeScript types,
+// as well as, OpenSearch Dashboards Platform `plugin()` initializer.
+
+export function plugin(initializerContext: PluginInitializerContext) {
+  return new ApplicationConfigPlugin(initializerContext);
+}
+
+export { ApplicationConfigPluginSetup, ApplicationConfigPluginStart } from './types';
--- a/src/plugins/application_config/server/index.ts
+++ b/src/plugins/application_config/server/index.ts
@@ -7,8 +7,10 @@ import { PluginConfigDescriptor, PluginInitializerContext } from '../../../core/
 import { ApplicationConfigSchema, configSchema } from '../config';
 import { ApplicationConfigPlugin } from './plugin';
 
-// This exports static code and TypeScript types,
-// as well as, OpenSearch Dashboards Platform `plugin()` initializer.
+/*
+This exports static code and TypeScript types,
+as well as, OpenSearch Dashboards Platform `plugin()` initializer.
+*/
 
 export const config: PluginConfigDescriptor<ApplicationConfigSchema> = {
   schema: configSchema,
--- a/src/plugins/application_config/server/index.ts
+++ b/src/plugins/application_config/server/index.ts
@@ -1,3 +1,8 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 import { PluginInitializerContext } from '../../../core/server';
 import { ApplicationConfigPlugin } from './plugin';
 
--- a/src/plugins/application_config/server/index.ts
+++ b/src/plugins/application_config/server/index.ts
@@ -3,11 +3,16 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { PluginInitializerContext } from '../../../core/server';
+import { PluginConfigDescriptor, PluginInitializerContext } from '../../../core/server';
+import { ApplicationConfigSchema, configSchema } from '../config';
 import { ApplicationConfigPlugin } from './plugin';
 
 // This exports static code and TypeScript types,
 // as well as, OpenSearch Dashboards Platform `plugin()` initializer.
+
+export const config: PluginConfigDescriptor<ApplicationConfigSchema> = {
+  schema: configSchema,
+};
 
 export function plugin(initializerContext: PluginInitializerContext) {
   return new ApplicationConfigPlugin(initializerContext);
--- a/src/plugins/application_config/server/opensearch_config_client.test.ts
+++ b/src/plugins/application_config/server/opensearch_config_client.test.ts
@@ -159,7 +159,7 @@ describe('OpenSearch Configuration Client', () => {
       expect(value).toBe('config1');
     });
 
-    it('throws exception when opensearch throws exception', async () => {
+    it('throws error when opensearch throws error', async () => {
       const error = new Error(ERROR_MESSAGE);
 
       opensearchClient.asCurrentUser.delete.mockImplementation(() => {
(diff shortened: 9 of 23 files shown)


src/plugins/application_config/server/opensearch_config_client.ts
```diff
@@ -0,0 +1,40 @@
...
+    this.logger = inputLogger;
+  }
+
+  async existsConfig(): Promise<boolean> {
+    try {
+      const exists = await this.client.asInternalUser.indices.exists({
+        index: this.configurationIndexName,
+      });
+
+      return exists.body;
```
kavilla, line 31: does this have a body? looks like a HEAD request

src/plugins/application_config/server/routes/index.ts
```diff
@@ -0,0 +1,50 @@
...
+  OpenSearchDashboardsResponseFactory,
+} from '../../../../core/server';
+import { ConfigurationClient } from '../types';
+
+export function defineRoutes(
+  router: IRouter,
+  getConfigurationClient: (inputOpenSearchClient: IScopedClusterClient) => ConfigurationClient,
+  logger: Logger
+) {
+  router.get(
```
kavilla, line 14: side note: i wonder why we dont have a router.head? Might have to look into this.

seraphjiang:
could we take a look configuration api in security plugin and see if we want to align the convention, and could we?

https://github.com/search?q=repo%3Aopensearch-project%2Fsecurity-dashboards-plugin+%22%2Fapi%2Fv1%2Fconfiguration%22&type=code

```
/api/v1/configuration/{entity}
```

@zengyan-amazon @Flyingliuhub @bandinib-amzn @ZilongX

src/plugins/application_config/README.md
```diff
@@ -0,0 +1,11 @@
+# applicationConfig
+
+A OpenSearch Dashboards plugin
```
seraphjiang, line 3:
```suggestion
OpenSearch Dashboards plugin for application config service and OpenSearch implementation
```

src/plugins/application_config/opensearch_dashboards.json
```diff
@@ -0,0 +1,9 @@
+{
+  "id": "applicationConfig",
+  "version": "1.0.0",
+  "opensearchDashboardsVersion": "opensearchDashboards",
+  "server": true,
+  "ui": false,
+  "requiredPlugins": ["navigation"],
```
seraphjiang, line 7: do we really need this dependency

src/plugins/application_config/server/opensearch_config_client.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { ConfigurationClient } from './types';
+
+export class OpenSearchConfigurationClient implements ConfigurationClient {
+  private client: IScopedClusterClient;
+  private configurationIndexName: string;
+  private readonly logger: Logger;
+
+  constructor(
+    inputOpenSearchClient: IScopedClusterClient,
```
seraphjiang, line 16: the input prefix of method parameters seems redundant

src/plugins/application_config/server/types.ts
```diff
@@ -0,0 +1,9 @@
+export interface ApplicationConfigPluginSetup {
+  setConfigurationClient: (inputConfigurationClient: ConfigurationClient) => void;
+}
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface ApplicationConfigPluginStart {}
+
+export interface ConfigurationClient {
```
seraphjiang, line 7: could we add jsdoc to explain what does this interface expect?

src/plugins/application_config/server/plugin.ts
```diff
@@ -0,0 +1,82 @@
...
+  implements Plugin<ApplicationConfigPluginSetup, ApplicationConfigPluginStart> {
+  private readonly logger: Logger;
+  private readonly config$: Observable<SharedGlobalConfig>;
+
+  private configurationClient: ConfigurationClient;
+  private configurationIndexName: string;
+
+  constructor(initializerContext: PluginInitializerContext) {
+    this.logger = initializerContext.logger.get();
+    this.config$ = initializerContext.config.legacy.globalConfig$;
```
seraphjiang, line 37: that's purpose of this private property

src/plugins/application_config/server/routes/index.ts
```diff
@@ -0,0 +1,50 @@
...
+  client: ConfigurationClient,
+  response: OpenSearchDashboardsResponseFactory,
+  logger: Logger
+) {
+  try {
+    const result = await client.existsConfig();
+    return response.ok({
+      body: {
+        exists: result,
+      },
```
seraphjiang, line 37: any reason we wrap output within body?

src/plugins/application_config/server/routes/index.ts
```diff
@@ -0,0 +1,50 @@
...
+import { ConfigurationClient } from '../types';
+
+export function defineRoutes(
+  router: IRouter,
+  getConfigurationClient: (inputOpenSearchClient: IScopedClusterClient) => ConfigurationClient,
+  logger: Logger
+) {
+  router.get(
+    {
+      path: '/api/config/exists',
```
seraphjiang, line 16: why do we need specific` exists` api

config/opensearch_dashboards.yml
```diff
@@ -29,6 +29,10 @@
 # dashboards. OpenSearch Dashboards creates a new index if the index doesn't already exist.
 #opensearchDashboards.index: ".opensearch_dashboards"
 
+# OpenSearch Dashboards uses an index in OpenSearch to store dynamic configurations.
+# This shall be a different index from opensearchDashboards.index.
+# opensearchDashboards.config_index: ".opensearch_dashboards_config"
```
BionIT, line 34: Looks like we use snake case instead of camel case which is used in all other configs for the config name/key, is it expected?

src/plugins/application_config/.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  root: true,
+  extends: ['@elastic/eslint-config-kibana', 'plugin:@elastic/eui/recommended'],
+  rules: {
+    '@osd/eslint/require-license-header': 'off',
```
BionIT, line 5: Why do we want to turn off require license header?

src/plugins/application_config/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+---
+
+## Development
```
BionIT, line 7: Can we add more details about this plugin? Here we can add instructions how to turn it on and off.

src/plugins/application_config/common/index.ts
```diff
@@ -0,0 +1,2 @@
+export const PLUGIN_ID = 'applicationConfig';
+export const PLUGIN_NAME = 'applicationConfig';
```
BionIT, line 2: If the value stored is the same, do we need 2 different variable names?

src/plugins/application_config/opensearch_dashboards.json
```diff
@@ -0,0 +1,9 @@
+{
+  "id": "applicationConfig",
+  "version": "1.0.0",
```
BionIT, line 3: How do we use version if this is a core plugin?

src/plugins/application_config/server/index.ts
```diff
@@ -0,0 +1,11 @@
+import { PluginInitializerContext } from '../../../core/server';
```
BionIT, line 1: Looks like we are missing the license header here

src/plugins/application_config/server/opensearch_config_client.ts
BionIT, line 31: Also wonder if the exists.body is a boolean value, I thought exists itself is a boolean

src/plugins/application_config/server/plugin.ts
```diff
@@ -0,0 +1,82 @@
...
+      this.configurationIndexName,
+      this.logger
+    );
+
+    this.setConfigurationClient(openSearchConfigurationClient);
+    return this.configurationClient;
+  }
+
+  public async setup(core: CoreSetup) {
+    this.logger.debug('applicationConfig: Setup');
```
BionIT, line 61: Do we need this log?

src/plugins/application_config/server/routes/index.ts
```diff
@@ -0,0 +1,50 @@
...
+        exists: result,
+      },
+    });
+  } catch (e) {
+    logger.error(e);
+    return errorResponse(response, e);
+  }
+}
+
+export function errorResponse(response: OpenSearchDashboardsResponseFactory, error: any) {
```
BionIT, line 45: This looks like a private function, do we need to export it? Also wonder if we can have response object which defines the structure similar how other Osd errors, possibly by extending OsdError.

src/plugins/application_config/server/plugin.ts
```diff
@@ -0,0 +1,82 @@
...
+    const router = core.http.createRouter();
+
+    const config = await this.config$.pipe(first()).toPromise();
+
+    this.configurationIndexName = config.opensearchDashboards.config_index;
+
+    // Register server side APIs
+    defineRoutes(router, this.getConfigurationClient.bind(this), this.logger);
```
BionIT, line 69: I thought the client is initiated already and we should be able to use the object. Are we not able to get the open search client object when the plugin is initiated?

BionIT (commented):
Left a few comments and please let me know if any questions

src/plugins/application_config/server/routes/index.ts
tianleh: Answered here https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5641#discussion_r1439933024

src/plugins/application_config/server/types.ts
tianleh: Sure. Will do.

src/plugins/application_config/opensearch_dashboards.json
tianleh: Good call out. Let me check other core plugins

tianleh:
> > could we take a look configuration api in security plugin and see if we want to align the convention, and could we?
> > https://github.com/search?q=repo%3Aopensearch-project%2Fsecurity-dashboards-plugin+%22%2Fapi%2Fv1%2Fconfiguration%22&type=code
> > ```
> > /api/v1/configuration/{entity}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @zengyan-amazon @Flyingliuhub @bandinib-amzn @ZilongX @BionIT
> 
> I will take a look the related links.

From the public doc https://opensearch.org/docs/latest/security/access-control/api/ , the APIs have the feature name (e.g `role`, `internalusers`) in the path. This is similar to what I am proposing. E.g for csp rules, we will have APIs like this

`curl -X GET 'http://localhost:5601/api/config/csp/rules'`

I also mention the alternative approach in the issue.
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5796#alternative-approach

seraphjiang:
> From the public doc https://opensearch.org/docs/latest/security/access-control/api/ , the APIs have the feature name (e.g `role`, `internalusers`) in the path. This is similar to what I am proposing. E.g for csp rules, we will have APIs like this
> 
> `curl -X GET 'http://localhost:5601/api/config/csp/rules'`
> 
> I also mention the alternative approach in the issue.

Didn't see the version in endpoint of new proposal, since we already has configuration api, introduce config seem will cause confusion.

tianleh:
> > From the public doc https://opensearch.org/docs/latest/security/access-control/api/ , the APIs have the feature name (e.g `role`, `internalusers`) in the path. This is similar to what I am proposing. E.g for csp rules, we will have APIs like this
> > `curl -X GET 'http://localhost:5601/api/config/csp/rules'`
> > I also mention the alternative approach in the issue.
> 
> Didn't see the version in endpoint of new proposal, since we already has configuration api, introduce config seem will cause confusion.

@zengyan-amazon Do you have context about adding `v1` to the security OSD APIs?

@seraphjiang I don't think it is common to have version in the API. E.g this is the ISM OSD plugin APIs https://github.com/opensearch-project/index-management-dashboards-plugin/blob/76b85ea6795f2b5a9b15df660dff3a191f520544/utils/constants.ts#L6

To avoid confusion with security OSD APIs, I am proposing the new APIs to look like this

```
/api/appconfig/csp/rules
```

tianleh:
Find a strange problem that the plugin registration is not working but my POC code is working https://github.com/tianleh/OpenSearch-Dashboards/blob/cj-dev/plugins/csp_storage_provider/server/plugin.ts#L63

This is for an external plugin to call the set function exposed from the configuration service plugin. I am still debugging it.

tianleh:
Gain some new ideas after syncing up with @kavilla 

Currently I am also experimenting an idea to add a new plugin which is responsible for implementing an OpenSearch based client and register to the configuration service. This can simplify the logic at configuration service plugin.

tianleh:
> > > From the public doc https://opensearch.org/docs/latest/security/access-control/api/ , the APIs have the feature name (e.g `role`, `internalusers`) in the path. This is similar to what I am proposing. E.g for csp rules, we will have APIs like this
> > > `curl -X GET 'http://localhost:5601/api/config/csp/rules'`
> > > I also mention the alternative approach in the issue.
> > 
> > 
> > Didn't see the version in endpoint of new proposal, since we already has configuration api, introduce config seem will cause confusion.
> 
> @zengyan-amazon Do you have context about adding `v1` to the security OSD APIs?
> 
> @seraphjiang I don't think it is common to have version in the API. E.g this is the ISM OSD plugin APIs https://github.com/opensearch-project/index-management-dashboards-plugin/blob/76b85ea6795f2b5a9b15df660dff3a191f520544/utils/constants.ts#L6
> 
> To avoid confusion with security OSD APIs, I am proposing the new APIs to look like this
> 
> ```
> /api/appconfig/csp/rules
> ```

Updated both PR and RFC to reflect the new API path.

tianleh:
> Find a strange problem that the plugin registration is not working but my POC code is working https://github.com/tianleh/OpenSearch-Dashboards/blob/cj-dev/plugins/csp_storage_provider/server/plugin.ts#L63
> 
> This is for an external plugin to call the set function exposed from the configuration service plugin. I am still debugging it.

Find the root cause. For the plugin which tries to register its client, it needs to use the exact variable name which is the configuration plugin id `applicationConfig` when importing into its `setup` function.

seraphjiang:
> Gain some new ideas after syncing up with @kavilla

Thanks @kavilla and @tianleh for explore and find different way to abstract

tianleh:
Updated the preferred approach to use a generic style API/interface https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5796#proposed-approach

and in progress to update the PR to reflect the changes.

config/opensearch_dashboards.yml
xinruiba, line 34: Since this index is to store app settings and used to do hot reload. Not sounds like configuration usage. 

To make it differentiate from a yml configuation,  maybe rename it to `opensearchDashboards.setting_index`?

src/plugins/application_config/server/opensearch_config_client.ts
xinruiba: +1

src/plugins/application_config/server/types.ts
tianleh: Added

src/plugins/application_config/server/routes/index.ts
tianleh: I have updated the code to use `route.get/delete/post` to avoid putting the head names into API path. Let me know if this addressed your comment.

src/plugins/application_config/server/routes/index.ts
tianleh: Removed the `exists` API. It was originally motivated by OpenSearch implementation detail.

src/plugins/application_config/server/opensearch_config_client.ts
tianleh: Updated the variable name to look clean

src/plugins/application_config/server/routes/index.ts
tianleh: Removed the API. However, I originally added the `body` due to existing examples in OSD code base

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/eff7cb5536a552692c43fedd60fb8a9f87f1e801/examples/routing_example/server/routes/message_routes.ts#L97

src/plugins/application_config/opensearch_dashboards.json
tianleh: Done.

src/plugins/application_config/README.md
tianleh: Updated.

src/plugins/application_config/server/opensearch_config_client.ts
tianleh: API is removed and there is no more `exists.body` logic in this PR.

src/plugins/application_config/server/index.ts
tianleh: Added.

src/plugins/application_config/opensearch_dashboards.json
tianleh: Updated to be `opensearchDashboards`

src/plugins/application_config/server/opensearch_config_client.ts
tianleh: API is removed.

src/plugins/application_config/README.md
tianleh: Updated readme.

config/opensearch_dashboards.yml
seraphjiang: The difference between config and settings is config is for application level, settings is for user level

src/plugins/application_config/.eslintrc.js
tianleh: The file was automatically added by the plugin generator helper. Just check that other plugins do not have this file. Let me just remove it. Thanks for the scrutiny!

src/plugins/application_config/server/plugin.ts
tianleh: Automatically created by plugin generator. Just removed.

src/plugins/application_config/common/index.ts
tianleh: Updated the value to be different.

config/opensearch_dashboards.yml
tianleh: Thanks @seraphjiang for the explanation!

tianleh: Updated to be camel case. Thanks for the call out.

src/plugins/application_config/server/routes/index.ts
tianleh: Originally it was for ease of unit testing to boost the test coverage to pass in Github workflow. I could actually indirectly test this function as part of other functions. I will update to not export it.

Looks like the `response` has a built in function `customError` for creating error response. I will switch to use it.

src/plugins/application_config/server/plugin.ts
tianleh: We need a scoped opensearch client which is available from a request context. This is for user permission against the index through calling `asCurrentUser` for update and delete. As a contrast, we are calling `asInternalUser` for read. Thus it is not yet there when plugin is initialized. 

It will need the caller to pass in their scoped client.

tianleh:
Got a strange lint failure.


```

ERROR 
      /home/ubuntu/workspace/clickjacking/OpenSearch-Dashboards/src/plugins/application_config/server/opensearch_config_client.test.ts
        9:8  error  Unexpected path "../../../core/server/opensearch/client/mocks" imported in restricted zone. Plugins may only import from top-level public and server modules in core  @osd/eslint/no-restricted-paths
      
      ✖ 1 problem (1 error, 0 warnings)
      
ERROR [eslint] errors
```

It is complaining about the following imports. However, when running jest test, the test itself is passing.

```

import {
  ScopedClusterClientMock,
  opensearchClientMock,
} from '../../../core/server/opensearch/client/mocks';
```

tianleh:
Disable this new plugin by default and also mentioned that this is an experimental feature.

tianleh:
Thanks all for all the feedback! I have addressed them and marked this PR ready for review. Also I have mentioned that this is an experimental feature and is disabled by default. Please help take a look!

@kavilla @BionIT @bandinib-amzn @xinruiba 

cc @seraphjiang

tianleh:
Need help to rerun the two failed functional tests which look like flaky tests.
https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/8027929095/job/21932500628?pr=5855
https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/8027929097/job/21932503003?pr=5855

seraphjiang:
Could we submit a dummy commit to trigger if we don't want to wait for maintainer's help since it is weekend

tianleh:
> Could we submit a dummy commit to trigger if we don't want to wait for maintainer's help since it is weekend

Good suggestion! Just submitted a dummy commit. https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855/commits/83cbc9ab6e78b8dfd2154bf06892f99692b0765b

tianleh:
After syncing up with @seraphjiang , we will still keep the plugin disabled but will not call it an experimental feature.

tianleh:
take a note of the failed workflows

Build and test / Build and Verify on Linux (ciGroup1) https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/8055655402/job/22002990289?pr=5855

Build and test / Build min release artifacts on macOS ARM64 https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/8055655402/job/22002999384?pr=5855

Note that these two are new ones which are different from failed ones in this run https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855#issuecomment-1962248915

tianleh:
Below is how the API call looks like.

Initially the entity `testdoc2` doesn't exist.

```
curl 'http://localhost:5601/api/appconfig/testdoc2'
```

Next update it with values.


```
curl 'http://localhost:5601/api/appconfig/testdoc2' -X POST -H 'Accept: application/json' -H 'Content-Type: application/json' -H 'osd-xsrf: osd-fetch' -H 'Sec-Fetch-Dest: empty' --data-raw '{"newValue":"frame-ancestors"}'

{"newValue":"frame-ancestors"}%

```

Get again.

```
curl 'http://localhost:5601/api/appconfig/testdoc2'

{"value":"frame-ancestors"}%

```

Check all existing entities.

```
curl 'http://localhost:5601/api/appconfig'

{"value":{"csp":"frame-ancestors","csp.rules":"frame-ancestors 'self' file://* filesystem:","testdoc2":"frame-ancestors"}}%
```

Delete the entity

```
curl 'http://localhost:5601/api/appconfig/testdoc2' -X DELETE -H 'osd-xsrf: osd-fetch' -H 'Sec-Fetch-Dest: empty'

{"deletedEntity":"testdoc2"}%
```

Check the entity again.

```
curl 'http://localhost:5601/api/appconfig/testdoc2'
```

src/plugins/application_config/.i18nrc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "prefix": "applicationConfig",
+  "paths": {
+    "applicationConfig": "."
+  },
+  "translations": ["translations/ja-JP.json"]
+}
```
bandinib-amzn, line 7: Same as above. Can we use central file - https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/.i18nrc.json?

src/plugins/application_config/README.md
```diff
@@ -0,0 +1,112 @@
...
+It also provides an interface `ConfigurationClient` for future extensions of external configuration clients. A default implementation based on OpenSearch as database is used.
+
+This plugin is disabled by default.
+
+## Configuration
+
+OSD users who want to set up application configurations will first need to enable this plugin by the following line in OSD YML.
+
+```
+application_config.enabled: true
```
bandinib-amzn, line 20: I didn't see entry for this configuration in YML file in this PR. Is it covered in another PR?

tianleh: Added.

src/plugins/application_config/server/index.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { PluginConfigDescriptor, PluginInitializerContext } from '../../../core/server';
+import { ApplicationConfigSchema, configSchema } from '../config';
+import { ApplicationConfigPlugin } from './plugin';
+
+// This exports static code and TypeScript types,
+// as well as, OpenSearch Dashboards Platform `plugin()` initializer.
```
bandinib-amzn, line 11: nit: can we use block comments here for better readability? `/*....*/`

src/plugins/application_config/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  root: true,
+  extends: ['@elastic/eslint-config-kibana', 'plugin:@elastic/eui/recommended'],
+  rules: {
+  },
+};
```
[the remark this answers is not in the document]
tianleh: They are introduced by the plugin generator by default. Let me check whether they are needed.

src/plugins/application_config/server/opensearch_config_client.test.ts
```diff
@@ -0,0 +1,202 @@
...
+
+      opensearchClient.asCurrentUser.delete.mockImplementation(() => {
+        return opensearchClientMock.createErrorTransportRequestPromise(error);
+      });
+
+      const client = new OpenSearchConfigurationClient(opensearchClient, INDEX_NAME, logger);
+
+      const value = await client.deleteEntityConfig('config1');
+
+      expect(value).toBe('config1');
```
bandinib-amzn, line 135: why it is returning entity when delete fails? Doesn't it throw error or reason to fail in response?

I see in test case just above this test case, return value from `deleteEntityConfig` is same. Then how would we differentiate between successful and failed delete operation?

src/plugins/application_config/server/opensearch_config_client.test.ts
```diff
@@ -0,0 +1,202 @@
...
```
bandinib-amzn, line 159: same as above.

src/plugins/application_config/server/opensearch_config_client.test.ts
```diff
@@ -0,0 +1,202 @@
...
+        return opensearchClientMock.createErrorTransportRequestPromise(error);
+      });
+
+      const client = new OpenSearchConfigurationClient(opensearchClient, INDEX_NAME, logger);
+
+      await expect(client.deleteEntityConfig('config1')).rejects.toThrowError(ERROR_MESSAGE);
+    });
+  });
+
+  describe('updateEntityConfig', () => {
```
bandinib-amzn, line 175: can we add one more test case - when entity is not present? 

Also question: Do we allow entity values such as undefined, '' empty to be passed to these APIs?

src/plugins/application_config/server/opensearch_config_client.ts
```diff
@@ -0,0 +1,115 @@
...
+    openSearchClient: IScopedClusterClient,
+    configurationIndexName: string,
+    logger: Logger
+  ) {
+    this.client = openSearchClient;
+    this.configurationIndexName = configurationIndexName;
+    this.logger = logger;
+  }
+
+  async getEntityConfig(entity: string) {
```
bandinib-amzn, line 25: _This doesn't need to in the scope of this PR._

Can we add validation for entity to not allow bad entity string?

src/plugins/application_config/server/opensearch_config_client.ts
```diff
@@ -0,0 +1,115 @@
...
+    } catch (e) {
+      const errorMessage = `Failed to get entity ${entity} due to error ${e}`;
+
+      this.logger.error(errorMessage);
+
+      throw e;
+    }
+  }
+
+  async updateEntityConfig(entity: string, newValue: string) {
```
bandinib-amzn, line 42: same as above.

src/plugins/application_config/server/opensearch_config_client.ts
```diff
@@ -0,0 +1,115 @@
...
+      return entity;
+    } catch (e) {
+      if (e?.body?.error?.type === 'index_not_found_exception') {
+        this.logger.info('Attemp to delete a not found index.');
+        return entity;
+      }
+
+      if (e?.body?.result === 'not_found') {
+        this.logger.info('Attemp to delete a not found document.');
+        return entity;
```
bandinib-amzn, line 78: instead of returning entity, can we pass response with meaningful information why delete operation failed? What would be the response status code if fails?

src/plugins/application_config/server/opensearch_config_client.ts
```diff
@@ -0,0 +1,115 @@
...
+
+      const errorMessage = `Failed to delete entity ${entity} due to error ${e}`;
+
+      this.logger.error(errorMessage);
+
+      throw e;
+    }
+  }
+
+  async getConfig(): Promise<any> {
```
bandinib-amzn, line 89: From the above readMe return type should be `Promise<Map<string, string>>`, right? Can you update return type here as well?

src/plugins/application_config/server/opensearch_config_client.ts
```diff
@@ -0,0 +1,115 @@
...
+    } catch (e) {
+      const errorMessage = `Failed to call getConfig due to error ${e}`;
+
+      this.logger.error(errorMessage);
+
+      throw e;
+    }
+  }
+
+  transformIndexSearchResponse(hits) {
```
bandinib-amzn, line 105: nit: add return type to  this function

src/plugins/application_config/server/plugin.ts
```diff
@@ -0,0 +1,81 @@
...
+  private configurationClient: ConfigurationClient;
+  private configurationIndexName: string;
+
+  constructor(initializerContext: PluginInitializerContext) {
+    this.logger = initializerContext.logger.get();
+    this.config$ = initializerContext.config.legacy.globalConfig$;
+    this.configurationIndexName = '';
+  }
+
+  private registerConfigurationClient(configurationClient: ConfigurationClient) {
```
bandinib-amzn, line 41: Can we create interface for this class and use that as a type?

src/plugins/application_config/server/plugin.ts
```diff
@@ -0,0 +1,81 @@
...
+    this.logger.info('Register a configuration client.');
+    this.configurationClient = configurationClient;
+  }
+
+  private getConfigurationClient(configurationClient: IScopedClusterClient): ConfigurationClient {
```
bandinib-amzn, line 46: same as above. Can we use Interface as return type rather class?

src/plugins/application_config/server/plugin.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    const config = await this.config$.pipe(first()).toPromise();
+
+    this.configurationIndexName = config.opensearchDashboards.configIndex;
+
+    // Register server side APIs
+    defineRoutes(router, this.getConfigurationClient.bind(this), this.logger);
+
+    return {
+      getConfigurationClient: this.getConfigurationClient.bind(this),
```
bandinib-amzn, line 71: Is this method needs to part of start? Have we tested this from plugin?

src/plugins/application_config/server/routes/index.test.ts
```diff
@@ -0,0 +1,353 @@
...
+
+      expect(response.customError).toBeCalledWith({
+        body: error,
+        statusCode: 500,
+      });
+
+      expect(logger.error).toBeCalledWith(error);
+    });
+  });
+});
```
bandinib-amzn, line 353: can we add UT for custom configure client registered via registerConfigurationClient?

src/plugins/application_config/server/routes/index.ts
```diff
@@ -0,0 +1,162 @@
...
+    });
+  } catch (e) {
+    logger.error(e);
+    return errorResponse(response, e);
+  }
+}
+
+function errorResponse(response: OpenSearchDashboardsResponseFactory, error: any) {
+  return response.customError({
+    statusCode: error?.statusCode || 500,
```
bandinib-amzn, line 159: I think appsec would not be happy about 500 error code. But it is just call out, not a blocker. you might want to check this when you do app sec review

src/plugins/application_config/server/types.ts
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { IScopedClusterClient } from 'src/core/server';
+
+export interface ApplicationConfigPluginSetup {
+  getConfigurationClient: (inputOpenSearchClient: IScopedClusterClient) => ConfigurationClient;
```
bandinib-amzn, line 9: I think this should be part of ApplicationConfigPluginStart interface.

src/plugins/application_config/server/types.ts
```diff
@@ -0,0 +1,52 @@
...
+  registerConfigurationClient: (inputConfigurationClient: ConfigurationClient) => void;
+}
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface ApplicationConfigPluginStart {}
+
+/**
+ * The interface defines the operations against the application configurations at both entity level and whole level.
+ *
+ */
+export interface ConfigurationClient {
```
bandinib-amzn, line 19: nit: In order to follow naming convention let's change it to `IConfigurationClient`.

src/plugins/application_config/server/plugin.ts
tianleh: 1. 

Initially I thought about putting it in a `start`. However, for the Clickjacking use case, it needs this function to register a request handler during `setup`. See the draft PR for reference 

https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5641/files#diff-f72753e435f681570e82eb6935f6a95941e57c22b4bf6a42234c8f9370e80099R70

(Note that the content of the referred PR will updated greatly but the piece of logic to register request handler is still valid).

With this said, for other plugins who want to use the get function, they could still retrieve it during `setup` but could actually call it to perform real work during `start`. So we are not limiting the users.


2.

In terms of testing, I have tested it in the related PR. Also I have written an example storage plugin in current workspace with my current PR to just verify that a get function is retrieved from a plugin. The example code is checked in here https://github.com/tianleh/StorageProviderPOC/blob/main/server/plugin.ts#L34

src/plugins/application_config/server/types.ts
tianleh: Answered here https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855/files#r1505186060

src/plugins/application_config/server/routes/index.ts
tianleh: I will follow up on this during review. Thanks for the call out!

src/plugins/application_config/server/plugin.ts
```diff
@@ -0,0 +1,81 @@
...
+} from './types';
+import { defineRoutes } from './routes';
+import { OpenSearchConfigurationClient } from './opensearch_config_client';
+
+export class ApplicationConfigPlugin
+  implements Plugin<ApplicationConfigPluginSetup, ApplicationConfigPluginStart> {
+  private readonly logger: Logger;
+  private readonly config$: Observable<SharedGlobalConfig>;
+
+  private configurationClient: ConfigurationClient;
```
tianleh, line 32: Only one is allowed. I will add some checks to prevent multiple registrations. Good call out! @bandinib-amzn

src/plugins/application_config/server/types.ts
tianleh: I did a quick search in the OSD code base. https://github.com/search?q=repo%3Aopensearch-project%2FOpenSearch-Dashboards+%22export+interface%22&type=code&p=1

Looks like most of the interfaces just have the regular name without "I" prefix. Let me know if there is a particular reason for this. Otherwise, I will leave the name as it is.

src/plugins/application_config/server/plugin.ts
tianleh: `ConfigurationClient` is already an interface.

src/plugins/application_config/server/plugin.ts
tianleh: replied here https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855/files#r1505207584

src/plugins/application_config/README.md
```diff
@@ -0,0 +1,112 @@
...
+First, this plugin will need to implement a class `MyConfigurationClient` based on interface `ConfigurationClient` defined in the `types.ts` under directory `src/plugins/application_config/server/types.ts`. Below are the functions inside the interface.
+
+```
+  getConfig(): Promise<Map<string, string>>;
+
+  getEntityConfig(entity: string): Promise<string>;
+
+  updateEntityConfig(entity: string, newValue: string): Promise<string>;
+
+  deleteEntityConfig(entity: string): Promise<string>;
```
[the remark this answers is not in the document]
tianleh: The responses are explained in the js docs https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855/files#diff-baec1ea8672dbf8991634d4cc7827fe9548ddf707cf602009b8c7e78a9997d21R43

Let me know if this is enough.

src/plugins/application_config/server/index.ts
tianleh: Updated.

src/plugins/application_config/server/opensearch_config_client.test.ts
tianleh: When the client throws error, we perform two additional checks. The first check is whether the error is due to `index` not found. The second check is whether the error is due to `document` not found. In both cases, since the system already has no those resources to delete, we treat it as successful. 

If the error is not caused by two cases above, we will throw the error.

src/plugins/application_config/server/opensearch_config_client.test.ts
tianleh: as this https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855#discussion_r1505246093

src/plugins/application_config/server/opensearch_config_client.ts
tianleh: added input validation as well as related unit tests.

src/plugins/application_config/server/opensearch_config_client.test.ts
tianleh: add input validation as well as unit tests

src/plugins/application_config/server/opensearch_config_client.ts
tianleh: Addressed here. https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855/files#r1505073929 Both are happy cases per my current definition.

src/plugins/application_config/server/opensearch_config_client.ts
tianleh: added input validation.

src/plugins/application_config/server/plugin.ts
tianleh: added a check in the register function.

src/plugins/application_config/server/opensearch_config_client.ts
tianleh: Good catch! just updated.

src/plugins/application_config/server/opensearch_config_client.ts
tianleh: added.

src/plugins/application_config/README.md
```diff
@@ -0,0 +1,112 @@
+# ApplicationConfig Plugin
+
+An OpenSearch Dashboards plugin for application configuration service and OpenSearch implementation.
```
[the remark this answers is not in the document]
tianleh: added more info.

src/plugins/application_config/server/routes/index.test.ts
tianleh: The current test file [src/plugins/application_config/server/routes/index.test.ts](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855/files#diff-5059f92734e8006e1bdd9dd51c074825d427ba7b9c12e505f5a2267905f06d3d) is about the `routers`.

The function `registerConfigurationClient` exists in the plugin class itself. Added a new test file `plugin.test.ts` to cover the `registerConfigurationClient` logic to verify error when double registering.

tianleh:
@bandinib-amzn  Thanks for all the detailed comments! I have responded/addressed all of your comments. Please help take a look again when you are available.

tianleh:
The workflow `WhiteSource Security Check` failed. https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855/checks?check_run_id=22145434828

This is not caused by this PR. Noticed that other PRs have this failure as well. See below for example.
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5989/checks?check_run_id=22141040727

config/opensearch_dashboards.yml
```diff
@@ -285,4 +289,7 @@
 # opensearchDashboards.survey.url: "https://survey.opensearch.org"
 
 # Set the value of this setting to true to enable plugin augmentation on Dashboard
-# vis_augmenter.pluginAugmentationEnabled: true
\ No newline at end of file
+# vis_augmenter.pluginAugmentationEnabled: true
+
+# Set the value of this setting to true to enable plugin application config. By default it is disabled.
+# application_config.enabled: false
```
bandinib-amzn, line 295: nit: Can we put `opensearchDashboards.configIndex` and `application_config.enabled` together as they are related to same feature?

src/plugins/application_config/server/opensearch_config_client.ts
```diff
@@ -0,0 +1,131 @@
...
+    this.client = openSearchClient;
+    this.configurationIndexName = configurationIndexName;
+    this.logger = logger;
+  }
+
+  async getEntityConfig(entity: string) {
+    if (isEmpty(entity)) {
+      this.logger.error(ERROR_MESSSAGE_FOR_EMPTY_INPUT);
+      throw ERROR_FOR_EMPTY_INPUT;
+    }
```
bandinib-amzn, line 30: nit: Can we create separate function for validation, maybe in separate file which only consist validation logic?

bandinib-amzn (approved):

config/opensearch_dashboards.yml
tianleh: Sure. Moved.

src/plugins/application_config/server/opensearch_config_client.ts
tianleh: Moved to a separate function. Indeed looks cleaner!

bandinib-amzn (approved):

tianleh:
The whitesource check failure https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855/checks?check_run_id=22149217297 will be fixed once this PR is merged. https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5966

src/plugins/application_config/server/plugin.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (this.configurationClient) {
+      const errorMessage = 'Configuration client is already registered! Cannot register again!';
+      this.logger.error(errorMessage);
+      throw new Error(errorMessage);
+    }
+
+    this.configurationClient = configurationClient;
+  }
+
+  private getConfigurationClient(configurationClient: IScopedClusterClient): ConfigurationClient {
```
BionIT, line 53: Can we change the argument name to clusterClient instead of configurationClient? I was confused since we are trying to get configuration client, but had to pass a configuration client but its type is cluster client

BionIT, line 53: can we extract the getConfigurationClient and register to its own class? Expose private members as interfaces for consumers to use actually makes it public, right?

src/plugins/application_config/server/plugin.ts
```diff
@@ -0,0 +1,88 @@
...
+      return this.configurationClient;
+    }
+
+    const openSearchConfigurationClient = new OpenSearchConfigurationClient(
+      configurationClient,
+      this.configurationIndexName,
+      this.logger
+    );
+
+    return openSearchConfigurationClient;
```
BionIT, line 64: Looks like if we use opensearch configuration client as default, but we never set this.configurationclient = opensearchconfigurationclient, the result is if we don't register a client, then we have to initialize the client everytime we use get client, is it efficient? Also the function name didn't say we can get a default client, consumers of the plugin would not know that they didn't register a client since they can get one all the time.

CHANGELOG.md
```diff
@@ -73,6 +73,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [Multiple Datasource] Refactor client and legacy client to use authentication registry ([#5881](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5881))
 - [Multiple Datasource] Improved error handling for the search API when a null value is passed for the dataSourceId ([#5882](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5882))
 - [Multiple Datasource] Hide/Show authentication method in multi data source plugin based on configuration ([#5916](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5916))
+- Add support for dynamic application configurations ([#5855](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855))
```
ZilongX, line 76: maybe add a cool prefix title like [Dynamic Configurations] ? :)

src/plugins/application_config/server/plugin.ts
ZilongX: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855/files#diff-56089cfe3899c248a5c74be540575dece8ee092e3100fd13e12d79d92c39058aR50 -> seems the default setup is actually here, so per the fist time consumption `getConfigurationClient` would return `openSearchConfigurationClient` and then `registerConfigurationClient` would bind it up.

ZilongX: But yeah how about we just bind a default client which seems more efficient, not two hops say get the client first then bind it up.

src/plugins/application_config/server/string_utils.ts
```diff
@@ -0,0 +1,24 @@
...
+
+function isEmpty(input: string): boolean {
+  if (!input) {
+    return true;
+  }
+
+  return !input.trim();
+}
+
+export function validate(input: string, logger: Logger) {
```
ZilongX, line 19: Other than empty checker, could a potential evil regex or unsupported entries (like a made up one) cause trouble on the server side. Or more generally, could a bad config entry mess up the index say` ".opensearch_dashboards_config"` ?

config/opensearch_dashboards.yml
ZilongX, line 1: Just a quick one off the top of my head - what's the upgrade story for this newly introduced in index ? Migrated as is ? Any concerns regarding the version discrepancy or this won't hurt (hopefully).

src/plugins/application_config/server/string_utils.ts
tianleh: Good call out! The APIs I call with the opensearch client are matching document ID exactly. Thus there is no concern for the attack. The `trim` is sufficient. But your comment does remind me to use the trimmed strings when passing to the client. I just made a change accordingly.

In your example of `a bad config entry`, it looks like an empty string which will trigger my current validation logic to throw error. 

I do agree that in the future we could enhance the checks, e.g max length.

config/opensearch_dashboards.yml
tianleh: It is just a key value pair structure and thus no concern for upgrade story.

src/plugins/application_config/server/string_utils.ts
tianleh: Added a follow up issue https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6000

CHANGELOG.md
tianleh: Nice. Looks more clear to customers. Added.

src/plugins/application_config/server/plugin.ts
tianleh: It is exposed for valid reasons. This is a common strategy in OSD.

E.g DevTools plugin exposes a register function https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/dev_tools/public/plugin.ts#L113C7-L113C15

Then Console plugin can consume it https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/console/public/plugin.ts#L58

tianleh: Good catch. Updated for better readability

src/plugins/application_config/server/plugin.ts
tianleh: The reason why a scoped opensearch client is passed in every time is that we need permission check through `asCurrentUser` from the scoped client. This client is only available from request context when a request happens. 

See this usage as reference.
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855/files#diff-b48e2be6347480ab021e8026d996f555ebb4f69abb5f820d00e53dc55c52b9a5R72

My new class `OpenSearchConfigurationClient` just builds a light weight wrapper around the scoped client. Its lifecycle is same as its parent request lifecycle after which will be garbage collected.

kavilla (approved):
I think it's good initial iteration. Seems pretty silo'd and I like the abstraction. 

I think we can have improvements to hook into some of the code that transforms the data and make it easier to create the config index.

I think we can also update the limits.yml file but just glanced over this PR for one more time. Will look again closely later but not blocking it.

ZilongX (approved):

Issue #5900, opened 2024-02-20
[MD] Index pattern selector which was used in maps dashboard plugin and visualization should show data source name as prefix


BionIT:
**Is your feature request related to a problem? Please describe.**

I feel confused when multiple data source is enabled, while using visualization and need to select an index pattern created from connected data sources, it is almost impossible to distinguish which data source the index pattern belongs to
![Screenshot 2024-02-19 at 3 47 33 PM](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/109f6513-7eab-4cdb-9b09-8e200563035b)

**Describe the solution you'd like**

I would want to be able to have the data source connection name as a prefix to the index pattern name, similar to what we have in saved objects page and discover data picker
![Screenshot 2024-02-19 at 4 01 47 PM](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/3c0105c4-de37-415a-9c28-f558c7214b77)

saved objects page
![Screenshot 2024-02-19 at 4 01 17 PM](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/0cfef428-8d4f-4763-9f9e-264b1700b0f5)

**Describe alternatives you've considered**
There is no alternative other than guessing which data source the index pattern belongs to

**Additional context**

BionIT:
Synced with @kgcreative , and please correct me if it is wrong, the prefix should be like `<connection_name>::<index_pattern_name>`

BionIT:
Happy to work on this

BionIT:
@kgcreative This is what I have for now and should be able to raise PR by end of day if this looks good
![Screenshot 2024-02-20 at 10 15 48 AM](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/d0a4ff36-971d-420d-936d-402b5e338e4e)

kgcreative:
Thank you @BionIT! This LGTM!

BionIT:
Thanks @seraphjiang ! Updated the term in my comment

PR #5907, opened 2024-02-21
[MD] Concatenate data source name with index pattern name and change delimiter to double colon

BionIT:
### Description
This change adds data source name as a prefix to the index pattern name for the index pattern selector. This change also changed the delimiter for the datasource name and index pattern name to be double colons by changing getIndexPatternTitle function which is used by index pattern service(https://github.com/opensearch-project/OpenSearch-Dashboards/blob/eff7cb5536a552692c43fedd60fb8a9f87f1e801/src/plugins/data/common/index_patterns/index_patterns/index_patterns.ts#L127) and saved object management page(https://github.com/opensearch-project/OpenSearch-Dashboards/blob/eff7cb5536a552692c43fedd60fb8a9f87f1e801/src/plugins/saved_objects_management/server/routes/find.ts#L99). 

### Issues Resolved

Fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5900

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/fc52f753-dbde-4a5f-974d-b2e5818f6779

## Testing the changes
The following steps were performed to test the change in the recording:
1. disable data source, goes to maps plugin page
2. click `add layer`, then see the index pattern selector with options populated
3. go to visualization and select `Controls` , then click `Add`, will see index pattern selector with options populated correctly
4. enable data source, goes to the same visualization page, and will see the options that have data source name prefixed before index pattern name, the delimiter is double colon
5. go to maps plugin, click `add layer`, then see the index pattern selector with options populated, data source name is added as prefix before the index pattern name and the delimiter is double colon

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


bandinib-amzn:
Can we add local cluster prefix for index pattern created from local cluster index for consistent look? cc: @kgcreative

kgcreative:
> Can we add local cluster prefix for index pattern created from local cluster index for consistent look? cc: @kgcreative 

I think that makes sense. What do you think, @BionIT?

kgcreative:
With concatenated names, would that also resolve the current bug of not being able to create an index pattern with the same name against two clusters?

Today, if I create an index pattern named `sample_logs*` on my local cluster, I am forced to get creative and create an index pattern named `sample_log*` against cluster 2.

We can update the unique names requirement to only check for unique index pattern names against the same data source

BionIT:
> > Can we add local cluster prefix for index pattern created from local cluster index for consistent look? cc: @kgcreative
> 
> I think that makes sense. What do you think, @BionIT?

I think that would make it confusing if user doesn't know what does `local cluster` represents and they have no control of it since it is hard coded there and they might wonder where does this local cluster name came from. The reason we add data source name as prefix is to make it easy to distinguish between index patterns created using different data sources and the data source name was created by user. To keep consistency, I didn't see local cluster being added in the dashboards today to emphasize the index pattern or saved object is from local cluster.

src/plugins/data/public/ui/index_pattern_select/index_pattern_select.tsx
```diff
@@ -161,16 +171,66 @@ export default class IndexPatternSelect extends Component<IndexPatternSelectProp
...
+      if (dataSourcesToFetch.length > 0) {
+        const resp = await savedObjectsClient.bulkGet(dataSourcesToFetch);
+        resp.savedObjects.map((dataSourceSavedObject: SimpleSavedObject<any>) => {
+          dataSourceIdToTitleToUpdate.set(
+            dataSourceSavedObject.id,
+            dataSourceSavedObject.attributes.title
+          );
+        });
+      }
+
```
bandinib-amzn, line 193: Can we move this code to getTitle? We have index pattern id, saved object client to find reference, right? We don't have to pass third parameter. Otherwise consumer of getTitle always needs to pass map with data source id and title? Instead can getTitle function handle that?

bandinib-amzn:
> > > Can we add local cluster prefix for index pattern created from local cluster index for consistent look? cc: @kgcreative
> > 
> > 
> > I think that makes sense. What do you think, @BionIT?
> 
> I think that would make it confusing if user doesn't know what does `local cluster` represents and they have no control of it since it is hard coded there and they might wonder where does this local cluster name came from. The reason we add data source name as prefix is to make it easy to distinguish between index patterns created using different data sources and the data source name was created by user. To keep consistency, I didn't see local cluster being added in the dashboards today to emphasize the index pattern or saved object is from local cluster.

But we do have Local cluster option in cluster selector. Do you mean when multi data source feature is disabled?

BionIT:
> > > > Can we add local cluster prefix for index pattern created from local cluster index for consistent look? cc: @kgcreative
> > > 
> > > 
> > > I think that makes sense. What do you think, @BionIT?
> > 
> > 
> > I think that would make it confusing if user doesn't know what does `local cluster` represents and they have no control of it since it is hard coded there and they might wonder where does this local cluster name came from. The reason we add data source name as prefix is to make it easy to distinguish between index patterns created using different data sources and the data source name was created by user. To keep consistency, I didn't see local cluster being added in the dashboards today to emphasize the index pattern or saved object is from local cluster.
> 
> But we do have Local cluster option in cluster selector. Do you mean when multi data source feature is disabled?

That's the place for user to select the source and not for adding to the existing title, if we look at 1) data picker in discover, 2) in index pattern management page, 3) in saved objects management, none of them show the `local cluster` in the title. If we plan to be explicit about which resource belong to which cluster or local cluster, then I would suggest to make the change in all the pages. If we just need it to distinguish from resources from local cluster, then the current practice of handling is to not have `local cluster` in the title

src/plugins/data/public/ui/index_pattern_select/index_pattern_select.tsx
BionIT: This part of code would perform bulk get index patterns(existing behavior), the added behavior is if the referred data source doesn't exist in the current state, it will fetch the data sources in bulk. Note there is only 2 requests here. The getTitle function is only used for selected index pattern which is a single index pattern, the existing behavior is it use an index pattern id to make individual call to saved object API to get the index pattern title, and if we use this function to handle all index patterns, this would multiple by number of index patterns, then to add data source title would double the requests.
If we want to refactor this, I can extract the logic here to a function to perform bulk operation and to produce the idToTitle map

kgcreative:
Let's keep this issue scoped. I would prefer we don't change the way we display `local cluster` just yet, especially if that will require a lot of additional work and scope. We can have a sepparate conversation about enhancements and ergonomics of index patterns, including potentially a `display name` and other concepts in order to make this more fully featured in the future.

src/plugins/data/public/ui/index_pattern_select/index_pattern_select.tsx
bandinib-amzn: Got it. No need to extract for now. We need to unify the pickers from all screen and that would be another effort. Thanks for the change.

bandinib-amzn (approved):

bandinib-amzn:
Approved the change. But can you check why several checks are failing?

ZilongX (approved):

PR #5946, opened 2024-02-24
[BUG] Support duplicate index pattern name

mengweieric:
### Description

Before
Screenshot 2024-02-23 at 11 02 21 PM

After
Screenshot 2024-02-23 at 11 03 01 PM

After adding the id key, the duplicate index pattern is able to be selected, and seeing the index pattern id is reflected in search url.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/data_sources/datasource_selector/datasource_selectable.test.tsx
+++ b/src/plugins/data/public/data_sources/datasource_selector/datasource_selectable.test.tsx
@@ -181,4 +181,47 @@ describe('DataSourceSelectable', () => {
     ];
     expect(optionTexts).toEqual(expectedIndexPatternSortedOrder);
   });
+
+  it('should allow display and selection of duplicated index patterns based on unique key', async () => {
+    const mockDataSourceOptionListWithDuplicates = [
+      {
+        label: 'Index patterns',
+        options: [
+          { label: 'duplicate-index-pattern', key: 'unique-key-1' },
+          { label: 'unique-index-pattern-1', key: 'unique-key-2' },
+          { label: 'duplicate-index-pattern', key: 'unique-key-3' },
+          { label: 'unique-index-pattern-2', key: 'unique-key-4' },
+        ],
+      },
+    ] as any;
+
+    const handleSelect = jest.fn();
+
+    render(
+      <DataSourceSelectable
+        dataSources={[
+          ({
+            getDataSet: jest.fn().mockResolvedValue([]),
+            getType: jest.fn().mockReturnValue('DEFAULT_INDEX_PATTERNS'),
+            getName: jest.fn().mockReturnValue('Index patterns'),
+          } as unknown) as DataSourceType,
+        ]}
+        dataSourceOptionList={mockDataSourceOptionListWithDuplicates}
+        selectedSources={selectedSourcesMock}
+        onDataSourceSelect={handleSelect}
+        setDataSourceOptionList={setDataSourceOptionListMock}
+        onGetDataSetError={onFetchDataSetErrorMock}
+      />
+    );
+
+    const button = screen.getByLabelText('Open list of options');
+    fireEvent.click(button);
+
+    const optionsToSelect = screen.getAllByText('duplicate-index-pattern');
+    fireEvent.click(optionsToSelect[1]);
+
+    expect(handleSelect).toHaveBeenCalledWith(
+      expect.objectContaining([{ key: 'unique-key-3', label: 'duplicate-index-pattern' }])
+    );
+  });
 });
--- a/src/plugins/data/public/data_sources/datasource_selector/datasource_selectable.tsx
+++ b/src/plugins/data/public/data_sources/datasource_selector/datasource_selectable.tsx
@@ -52,6 +52,7 @@ export const getSourceOptions = (dataSource: DataSourceType, dataSet: DataSetTyp
       ...optionContent,
       label: dataSet.title,
       value: dataSet.id,
+      key: dataSet.id,
     };
   }
   return {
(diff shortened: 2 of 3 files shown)


seraphjiang (approved):

src/plugins/data/public/data_sources/datasource_selector/datasource_selectable.tsx
```diff
@@ -52,6 +52,7 @@ export const getSourceOptions = (dataSource: DataSourceType, dataSet: DataSetTyp
       ...optionContent,
       label: dataSet.title,
       value: dataSet.id,
+      key: dataSet.id,
```
ashwin-pc, line 55: Why are you only passing key for Index patterns? cant we have a key for every option?

mengweieric: I actually thought about this to also add key to non index pattern datasource, to do it generally where have 2 ways due to there's no such id in 'dataSet' for non-index-patterns datasources: 
- use a some kind of id generator to add key few lines after this line
- define the id or leverage id like data in datasource fetched from backend in datasource class

The second approach is the one that I prefer as we should define it as part of the datasource itself either by generating the id there or leverage similar id like key in datasource data we fetched from backend. This approach, will require us to first align on what should be the unified and unique id other than index pattern id. This will require more work to modify interfaces and classes. Therefore, I'm not sure even if we proceed with this change, does it still qualify for a patch release?

The first approach, or other similar approaches I feel would require generating an id on the fly, which is more of a workaround way due to the current limitation, i don't feel we should go ahead with it in this case.

Therefore I've tested with key only for index patterns, it works with duplicated names across non-index-patterns and index patterns sections (see attached screenshot), without displaying and selecting issues, and also s3 datasource does not allow name duplications, plus the datasource registrations currently using the datasource name as key to register with datasource service which seems like a key for non-index-patterns datasource is not required for 2.11. 

Screenshot 2024-02-24 at 8 55 11 PM

ashwin-pc: This isnt going in a patch. This should be going in the next minor, so we dont have to rush in a solution. Can the user have more than one s3 connection? or a similar name in spark? It looks like its a simple update to the datasource service to add an ID for every dataset. When creating the datasets, if it does not have a unique id, you can reuse the name as the ID if its unique, or make a combination such as `connection.dataset_name` as an id

mengweieric: This PR is created under the assumption that it could be considered as part of the patch release as users may raise issue in the case they import index pattern saved objects with the same name, this could mitigate the issue. But we may leave it if this is not the case. 

> It looks like its a simple update to the datasource service to add an ID for every dataset. 

yes but at the meantime, the id has to be passed to the redirection url for log explorer, then the corresponding code there need to also be changed accordingly for duplicate name use cases and retest the entire flow. But if like you said we will handle it in next minor then for sure a long-term, id solution will be added anyway for all datasources without depending on if duplication is allowed or not to cover border scope of use cases.

ashwin-pc: Hmm, kayi see the value in this going in the patch. Since this component is due for a complete rewrite anyways, lets merge this in while you work on a proper rewrite so that the effort isnt duplicated.

ashwin-pc (approved):

Flyingliuhub (approved):

src/plugins/data/public/data_sources/datasource_selector/datasource_selectable.test.tsx
```diff
@@ -181,4 +181,47 @@ describe('DataSourceSelectable', () => {
...
+        setDataSourceOptionList={setDataSourceOptionListMock}
+        onGetDataSetError={onFetchDataSetErrorMock}
+      />
+    );
+
+    const button = screen.getByLabelText('Open list of options');
+    fireEvent.click(button);
+
+    const optionsToSelect = screen.getAllByText('duplicate-index-pattern');
+    fireEvent.click(optionsToSelect[1]);
```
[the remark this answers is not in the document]
mengweieric: The user will be able to see the data loaded for that index pattern selected with corresponding id.

src/plugins/data/public/data_sources/datasource_selector/datasource_selectable.tsx
mengweieric: sure sounds good.

kavilla:
Sorry about the delay, @mengweieric could you provide context on this one?

In terms of index patterns, the current structure is that the name is the key. I don't think it's impossible to get in a state because there is only client side validation when creating index patterns, but it was implemented assuming that you couldn't get easily create a duplicate index pattern through the UI. We currently have a different data source (if not local) being prepended to the index pattern so ideally that would solve the case where a non-local cluster's index pattern.

So if you could provide context on how we expect to see a duplicate index pattern in the way that we currently see in the playground and screen shot?

mengweieric:
> Sorry about the delay, @mengweieric could you provide context on this one?
> 
> In terms of index patterns, the current structure is that the name is the id for index patterns. I don't think it's impossible to get in a state because there is only client side validation when creating index patterns, but it was implemented assuming that you couldn't get easily create a duplicate index pattern through the UI. We currently have a different data source (if not local) being prepended to the index pattern so ideally that would solve the case where a non-local cluster's index pattern.
> 
> So if you could provide context on how we expect to see a duplicate index pattern in the way that we currently see in the playground and screen shot when it is created the expected way?

Like the issue currently exists in playground, when users import index pattern only from outside of the current cluster which contains any index pattern name same as the ones in current cluster, they will face this issue. These index patterns currently, after imported, are not prefixed with any additional datasource/cluster info therefore conflicts with the ones in the cluster.

Issue #4944, opened 2023-09-06
[Feature][META]Workspace


xluo-aws:
Workspaces help organize your work, library items/saved objects, and tools, and is a convenient way to enable sharing and collaboration between users. It’s also a way to customize the tools that are available, so that when users switch between workspace, they have a focused view of the features and tools that are relevant for the selected workspace. 

- [ ] https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4615

yoicomment:
when can we expect this wonderful feature. utterly needed :D

xluo-aws:
Glad to know that you like it, we tentatively plan to deliver it in Q2 next year. Feel free to let us know your suggestions on what feature you want to see in workspace if it's not already in our design. 

> when can we expect this wonderful feature. utterly needed :D

xluo-aws:
> @xluo-aws I think, I will be able to give better feedback, after I tried it out. Happy to see OSD evolving :)

Thanks, we have started working on this feature.

xluo-aws:
Will deliver the complete feature in 2.17

getsaurabh02:
Thanks @xluo-aws. Will update the label

PR #5084, opened 2023-09-21
[Workspace][Feature] Add ACL related functions

gaobinlong:
### Description
The main change of this PR are:

- Add `permissions` field to the default mapping of the OSD index(.kibana) to support ACL for workspace
- Update the test snapshot after we change the default mapping of the OSD index
- Add ACL class and basic functions

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5083

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`
  - [x] `yarn test:jest_integration`
  - [x] `yarn test:ftr`
- [x] New functionality includes testing.
- [x] New functionality has been documented.
- [x] Update [CHANGELOG.md](./../CHANGELOG.md)
- [x] Commits are signed per the DCO using --signoff

PR #5949, opened 2024-02-26
[Workspace]Optional workspaces params in repository

SuZhou-Joe:
### Description

As workspace will be a new concept to organize saved objects, this PR add a new field `workspaces` to indicate which workspaces the objects belong to and consume the field in related methods.

- [x] Add workspaces field in saved objects mappings
- [x] Consume worksapces when create object
- [x] Consume workspaces when bulkCreate objects
- [x] Consume workspaces when checkConflicts

### Issues Resolved

Partial #4944 

## Screenshot

As all of the changes happens in server side, currently do not have screenshot for this PR. 

## Testing the changes

As all of the changes happens in server side, can not test from browser, so we can run the integration_test in `src/plugins/workspace/server/saved_objects/integration_tests/saved_objects_wrapper_for_check_workspace_conflict.test.ts`. The test covers the cases:

- [x] Create object by calling /api/saved_objects/${type} with workspaces field, and the response of the created body contains workspaces field.
- [x] Overwrite a existing workspace(worksapce id: foo) with workspace id: bar, the API gives 409 error and the overwrite doesn't happen.
- [x] Call bulkCreate API to create two objects in different workspaces successfully.
- [x] Call bulkCreate API to overwrite a object existing in workspace foo with params workspace bar, and it returns the failed error.
- [x] Call import API to overwrite a object existing in workspace foo with params workspace bar, and it returns the failed error.
- [x] Call find API to get all the objects under a specific workspace successfully.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`
  - [x] `yarn test:ftr`

### DIFF
--- a/config/opensearch_dashboards.yml
+++ b/config/opensearch_dashboards.yml
@@ -294,5 +294,5 @@
 # Set the value of this setting to true to enable plugin augmentation on Dashboard
 # vis_augmenter.pluginAugmentationEnabled: true
 
-# Set the value to true enable workspace feature
+# Set the value to true to enable workspace feature
 # workspace.enabled: false
--- a/config/opensearch_dashboards.yml
+++ b/config/opensearch_dashboards.yml
@@ -295,4 +295,4 @@
 # vis_augmenter.pluginAugmentationEnabled: true
 
 # Set the value to true enable workspace feature
-# workspace.enabled: true
+# workspace.enabled: false
--- a/config/opensearch_dashboards.yml
+++ b/config/opensearch_dashboards.yml
@@ -293,3 +293,6 @@
 
 # Set the value of this setting to true to enable plugin augmentation on Dashboard
 # vis_augmenter.pluginAugmentationEnabled: true
+
+# Set the value to true enable workspace feature
+# workspace.enabled: true
--- a/src/core/server/saved_objects/import/validate_references.ts
+++ b/src/core/server/saved_objects/import/validate_references.ts
@@ -48,8 +48,7 @@ export async function getNonExistingReferenceAsKeys(
   savedObjects: SavedObject[],
   savedObjectsClient: SavedObjectsClientContract,
   namespace?: string,
-  retries?: SavedObjectsImportRetry[],
-  workspaces?: string[]
+  retries?: SavedObjectsImportRetry[]
 ) {
   const objectsToSkip = getObjectsToSkip(retries);
   const collector = new Map();
@@ -73,15 +72,10 @@ export async function getNonExistingReferenceAsKeys(
     return [];
   }
 
-  const fields = ['id'];
-  if (workspaces?.length) {
-    fields.push('workspaces');
-  }
-
   // Fetch references to see if they exist
   const bulkGetOpts = Array.from(collector.values()).map((obj) => ({
     ...obj,
-    fields,
+    fields: ['id'],
   }));
   const bulkGetResponse = await savedObjectsClient.bulkGet(bulkGetOpts, { namespace });
 
--- a/src/core/server/saved_objects/import/validate_references.ts
+++ b/src/core/server/saved_objects/import/validate_references.ts
@@ -48,7 +48,8 @@ export async function getNonExistingReferenceAsKeys(
   savedObjects: SavedObject[],
   savedObjectsClient: SavedObjectsClientContract,
   namespace?: string,
-  retries?: SavedObjectsImportRetry[]
+  retries?: SavedObjectsImportRetry[],
+  workspaces?: string[]
 ) {
   const objectsToSkip = getObjectsToSkip(retries);
   const collector = new Map();
@@ -72,8 +73,16 @@ export async function getNonExistingReferenceAsKeys(
     return [];
   }
 
+  const fields = ['id'];
+  if (workspaces?.length) {
+    fields.push('workspaces');
+  }
+
   // Fetch references to see if they exist
-  const bulkGetOpts = Array.from(collector.values()).map((obj) => ({ ...obj, fields: ['id'] }));
+  const bulkGetOpts = Array.from(collector.values()).map((obj) => ({
+    ...obj,
+    fields,
+  }));
   const bulkGetResponse = await savedObjectsClient.bulkGet(bulkGetOpts, { namespace });
 
   // Error handling
--- a/src/core/server/saved_objects/import/validate_references.ts
+++ b/src/core/server/saved_objects/import/validate_references.ts
@@ -73,10 +73,7 @@ export async function getNonExistingReferenceAsKeys(
   }
 
   // Fetch references to see if they exist
-  const bulkGetOpts = Array.from(collector.values()).map((obj) => ({
-    ...obj,
-    fields: ['id'],
-  }));
+  const bulkGetOpts = Array.from(collector.values()).map((obj) => ({ ...obj, fields: ['id'] }));
   const bulkGetResponse = await savedObjectsClient.bulkGet(bulkGetOpts, { namespace });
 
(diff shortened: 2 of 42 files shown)


SuZhou-Joe:
@kavilla @Flyingliuhub @ashwin-pc , could you please take a look on this PR? This PR introduces many fundamental features for Workspace and are totally controlled behind a `workspace.enabled` feature flag.

seraphjiang (commented):
@SuZhou-Joe would you add the new config to OSD.yaml file 

With disable by default and comment out.

Also add a few words to describe the feature flag

SuZhou-Joe:
> @SuZhou-Joe would you add the new config to OSD.yaml file
> 
> With disable by default and comment out.
> 
> Also add a few words to describe the feature flag

Thanks and done for that.

config/opensearch_dashboards.yml
```diff
@@ -285,4 +285,7 @@
 # opensearchDashboards.survey.url: "https://survey.opensearch.org"
 
 # Set the value of this setting to true to enable plugin augmentation on Dashboard
-# vis_augmenter.pluginAugmentationEnabled: true
\ No newline at end of file
+# vis_augmenter.pluginAugmentationEnabled: true
+
+# Set the value to true enable workspace feature
```
ZilongX, line 290: nit, just add `to` to the comment :) - Set the value to true **to** enable workspace feature

SuZhou-Joe: Thanks and done.

src/core/server/saved_objects/import/validate_references.ts
```diff
@@ -73,7 +73,10 @@ export async function getNonExistingReferenceAsKeys(
   }
 
   // Fetch references to see if they exist
-  const bulkGetOpts = Array.from(collector.values()).map((obj) => ({ ...obj, fields: ['id'] }));
+  const bulkGetOpts = Array.from(collector.values()).map((obj) => ({
```
Flyingliuhub, line 76: If there is no change for this line, can we keep the original, as it is likely your VS code formatting style.

src/core/server/saved_objects/migrations/core/build_active_mappings.test.ts
```diff
@@ -91,6 +92,12 @@ describe('buildActiveMappings', () => {
     expect(hashes.aaa).toEqual(hashes.bbb);
     expect(hashes.aaa).not.toEqual(hashes.ccc);
   });
+
+  test('workspaces field is added when workspace feature flag is enabled', () => {
```
Flyingliuhub, line 96: Do we need to add a test for when the workspace feature flag is not enabled, and the workspace field does not exist?

src/core/server/saved_objects/saved_objects_config.ts
```diff
@@ -49,6 +49,9 @@ export const savedObjectsConfig = {
   schema: schema.object({
     maxImportPayloadBytes: schema.byteSize({ defaultValue: 26214400 }),
     maxImportExportSize: schema.byteSize({ defaultValue: 10000 }),
+    permission: schema.object({
+      enabled: schema.boolean({ defaultValue: false }),
```
Flyingliuhub, line 53: this is conflict with #5084, please make sure that who will be merged first and then rebase your branch.

src/core/server/saved_objects/saved_objects_service.test.ts
```diff
@@ -70,6 +70,13 @@ describe('SavedObjectsService', () => {
         maxImportExportSize: new ByteSizeValue(0),
       });
     });
+    const config$ = new BehaviorSubject<Config>(
+      new ObjectToConfigAdapter({
```
Flyingliuhub, line 74: the same as above comment, it is the same change with #5084.

src/plugins/workspace/server/saved_objects/saved_objects_wrapper_for_check_workspace_conflict.ts
```diff
@@ -0,0 +1,317 @@
...
+
+      /**
+       * Check if overwrite with id
+       * If so, need to reserve the workspace params
+       */
+      if (id && overwrite) {
+        let currentItem;
+        try {
+          currentItem = await wrapperOptions.client.get(type, id);
+        } catch (e) {
```
Flyingliuhub, line 55: is it possible to handle specific error here? for example, e.response && e.response.status === 404, so we can avoid unintentionally swallowing other types of errors.

SuZhou-Joe: Sure, let me optimize this part.

src/core/server/saved_objects/saved_objects_config.ts
SuZhou-Joe: Thanks, will remove this part.

src/core/server/saved_objects/migrations/core/build_active_mappings.test.ts
SuZhou-Joe: In previous test cases I think they've already cover this case but will take a look to confirm.

SuZhou-Joe: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/saved_objects/migrations/core/__snapshots__/build_active_mappings.test.ts.snap#L4

This snapshot should cover the case when feature flag is not enabled.

Flyingliuhub (approved):

src/core/server/saved_objects/saved_objects_config.ts
Flyingliuhub: > Thanks, will remove this part.

I assume that #5084 will merge first, then yours, right? Just want to make sure it builds and functions properly.

ZilongX:
LGTM for the initial iteration, would this feature need backporting to 2.x ?

ZilongX (approved):

SuZhou-Joe:
> LGTM for the initial iteration, would this feature need backporting to 2.x ?

We'd like to backport to 2.x until other down stream PRs get merged, thanks for the call out.

src/core/server/saved_objects/migrations/opensearch_dashboards/opensearch_dashboards_migrator.ts
```diff
@@ -36,6 +36,7 @@
 import { OpenSearchDashboardsConfigType } from 'src/core/server/opensearch_dashboards_config';
 import { BehaviorSubject } from 'rxjs';
 
+import { Config } from 'packages/osd-config/target';
```
ruanyl, line 39: Should we import from `@osd/config`?

BionIT (approved):

src/core/server/saved_objects/saved_objects_service.ts
```diff
@@ -557,6 +564,7 @@ export class SavedObjectsService
         this.logger,
         migrationsRetryDelay
       ),
+      opensearchDashboardsRawConfig: this.opensearchDashboardsRawConfig as Config,
```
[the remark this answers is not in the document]
SuZhou-Joe: I'd prefer to keep as it is because:
1. If we do some filter here, once we introduce more config to control the mappings dynamically, we need to change two places, filter and the `build_active_mapping`
2. For now the configService.atPath does not provide functionality to subscribe to multiple paths(even though it accepts string[], it is more like a lodash.get(obj, path)).

ruanyl (approved):

SuZhou-Joe:
@ruanyl Will raise a following PR to address your comment.

Issue #5952, opened 2024-02-26
[BUG] Dashboards fails to come up on main 


derek-ho:
**Describe the bug**
OSD doesn't come up healthy when running the built artifact:

[...]

PR #5968, opened 2024-02-27
Add commands to verify built binary doesn't contian errors

derek-ho:
### Description

Related to #5952 - Dashboards verifies that it builds, but not that the built artifact can run. 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/.github/workflows/build_and_test_workflow.yml
+++ b/.github/workflows/build_and_test_workflow.yml
@@ -6,7 +6,7 @@ name: Build and test
 # trigger on every commit push and PR for all branches except pushes for backport branches
 on:
   push:
-    branches: ['main', '[0-9].x', '[0-9].[0=9]+'] # Run the functional test on push for only release branches
+    branches: ['**', '!backport/**'] # Run the functional test on push for only release branches
     paths-ignore:
       - '**/*.md'
       - 'docs/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test_workflow.yml
```diff
@@ -422,6 +422,15 @@ jobs:
...
 
+      - name: Run `${{ matrix.name }}`
+        run: ./artifacts/target/${{ env.ARTIFACT_BUILD_NAME }}/bin/opensearch-dashboards &
+        shell: bash
+
+      - name: Health check 
+        run: |
+          timeout 300 bash -c 'while [[ "$(curl http://localhost:5601/api/status | jq -r '.status.overall.state')" != "green" ]]; do sleep 5; done'
+        shell: bash
+
```
derek-ho, line 433: Should catch issues like: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5952

derek-ho:
@bandinib-amzn I think this PR should catch issues in the CI process in core

bandinib-amzn (approved):

.github/workflows/build_and_test_workflow.yml
derek-ho: It seems like this workflow is not being run currently, so can't verify it works. @bandinib-amzn can you help me run it/can you enable it from your end just so we can verify?

manasvinibs (approved):

.github/workflows/build_and_test_workflow.yml
AMoo-Miki: The linux artifact is used in BWC tests and I don't think OSD will come up "green" without an OS instance to connect to.

AMoo-Miki (changes requested):
The linux artifact is used in BWC tests and I don't think OSD will come up "green" without an OS instance to connect to.

derek-ho:
> The linux artifact is used in BWC tests and I don't think OSD will come up "green" without an OS instance to connect to.

Ack. Would you accept a PR in which I spin up a simple opensearch instance and verify that the artifacts then respond to a simple health check?

AMoo-Miki:
BWC already covers this. If a plugin needs to verify that their binaries are being build correctly, that test should be on that repo's workflows.

PR #5975, opened 2024-02-28
[MDS] Support Vega Visualizations

### LINKED ISSUE #5927: [MDS] Vega support with MDS
## Proposal
Currently, Multiple DataSource (MDS) does not support Vega visualizations. Thus, we propose to create a new optional field from within the Vega-spec `data_source_name` (tentatively named) that is passed within the `url` body of the `data` body. This field will take in a datasource name and under the hood, the visualization will be able to retrieve data from the index located in that datasource. This enables a user to retrieve data from one or more indices from one or more datasources to create custom Vega visualizations.

Here is an example Vega spec that should be supported with this feature. Note that `data` field can be a singular `url` object or an array that contains multiple `url` objects:
```
data: {
    url: {
      %context%: true
      %timefield%: @timestamp
      index: opensearch_dashboards_sample_data_logs
      datasource: some_datasource_id
      body: {
        aggs: {
          ...
      }
    }
    format: {
      property: aggregations.5
    }
  }
```

## Background
[Vega](https://vega.github.io/vega/) is a declarative visualization grammar that can be used to create and share custom interactive visualizations. It is important to note that [Vega-lite](https://vega.github.io/vega-lite/) is a similar but very much different lightweight visualization grammar. Both of these grammars are supported within Dashboards but with the caveat that the data is retrieved from the index BEFORE any rendering happens. This means that data is NOT dynamically loaded. Additionally, Vega is not supported with MDS since the local cluster is the assumed datasource. This proposal will fix the latter (with the former being out of scope).

It is also important to clarify what is meant by Vega support for MDS. Vega support for MDS can be interpreted in two different ways
1. **Option 1:** Vega can support visualizations which fetch data from multiple datasources
MDS drawio
    In this example, the Vega visualization is stored in Datasource A but references data from Datasources A, B, and C

2. **Option 2:** Vega can support visualizations which reference data from the local cluster or any remote datasource (but not both) 
MDS drawio(1)
    In this example, the Vega visualization is stored in Datasource A but references data only from Datasource B and not any other datasource

The proposal seeks to support **Option 1**. While there are more limitations with option 1 (see the `Limitations` section), having the option to fetch data from any index from any datasource (provided the user has the permissions) provides a robust visualization experience.

## Approach
When Dashboards parses the Vega-spec to render the visualization, it [parses the URL object](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/vis_type_vega/public/data_model/opensearch_query_parser.ts) and passes the object into the [search API](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/vis_type_vega/public/data_model/search_api.ts), which uses `IOpenSearchSearchRequest` as a parameter. This interface provides a field `dataSourceId` that will tell dashboards to use the [data source client](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source/README.md). All that the Vega plugin would need to do is check if MDS is enabled and if so, retrieve the associated datasource id from the `data_source_name` field and pass it into the search query.

Add a `datasource` field to the `UrlObject`
```Typescript
export interface UrlObject {
  [index: string]: any;
  [CONSTANTS.TIMEFILTER]?: string;
  [CONSTANTS.CONTEXT]?: boolean;
  [CONSTANTS.LEGACY_CONTEXT]?: string;
  [CONSTANTS.TYPE]?: string;
  name?: string;
  index?: string;
  data_source_name?: string;
  body?: Body;
  size?: number;
  timeout?: string;
}
```

Have the saved objects client get the associated `dataSourceId` from the `data_source_name` via `find`

Then, in the `_searchAPI.search()` method, we can pass in the `dataSourceId` as a parameter
```Typescript
return search({params, dataSourceId} , { abortSignal: this.abortSignal }).pipe(
	tap((data) => this.inspectSearchResult(data, requestResponders[requestId])),
	map((data) => ({
		name: requestId,
		rawResponse: data.rawResponse,
	}))
);
```

Thus, when the user wants to write a visualization with data from another datasource, they can do something like the following (mockup)
image

## Limitations
This approach provides greater flexibility in enabling users to make visualizations. However with great power comes great responsibility. 
1. Importing/exporting Vega saved objects will inevitably break. Because `data_source_name` is specific to an OpenSearch cluster, if users were to export/import into another cluster and the same datasource names are not configured, the visualizations cannot find the data and thus return errors. This is a tradeoff that can be made but in the future, this would ideally need to be mitigated.
2. As a sort of add-on to point 1, editing `data_source_names` to each `url` that contains one is cumbersome, especially when multiple `url` objects were involved.

## Alternatives
Initially the decision was made to use the `data_source_id` vs the `data_source_name`. This was due to the fact that `data_source_id` enforces a unique datasource to query from and does not make an extra find query to find the datasource. However, other plugins referred to datasource by name, not by id, and having name be the identifier is more user friendly.

## Open Question(s)
1. Since `datasource` can be a bit ambiguous here, what are some alternative field names here? Im thinking `data_source_id` would help disambiguate this field

huyaboo:
### Description

This PR adds Vega support for MDS. If `data_source.enabled: true` and a `data_source_id` field is provided in the Vega spec, then the Vega visualization will render with data from the data source (assuming the data source has the specified index).

### Issues Resolved
closes #5927 

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/73027756/2f36633b-075a-4e0b-accd-a141e4ae28a7

### Pre-requisites
1. Enable MDS (set`data_source.enabled: true` in `opensearch_dashboards.yml`)
2. Add data source (in this example, it is `vegaDataSource`)
3. Add the newly created data source name to the `data_source_name` field under the `url` body

### Testing
1. Go to Visualize -> Create visualization -> Vega
2. Copy-paste existing Vega visualization from the data source (or create a new one)
3. Add a field `data_source_id` to the vega spec under `url` and set the value to the data source id.
5. Vega should render, update when timefield is updated, and save.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
-- [Multiple Datasource] Add Vega support to MDS by specifying a data source name in the Vega spec ([#5975](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5975))
 
 ### 💥 Breaking Changes
 
@@ -12,7 +11,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### 🛡 Security
 
 ### 📈 Features/Enhancements
-- [MD]Change cluster selector component name to data source selector ([#6042](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6042)) 
+- [MD]Change cluster selector component name to data source selector ([#6042](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6042))
 - [Multiple Datasource] Add interfaces to register add-on authentication method from plug-in module ([#5851](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5851))
 - [Multiple Datasource] Able to Hide "Local Cluster" option from datasource DropDown ([#5827](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5827))
 - [Multiple Datasource] Add api registry and allow it to be added into client config in data source plugin ([#5895](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5895))
@@ -23,6 +22,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [Dynamic Configurations] Add support for dynamic application configurations ([#5855](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855))
 - [Workspace] Optional workspaces params in repository ([#5949](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5949))
 - [Multiple Datasource] Refactoring create and edit form to use authentication registry ([#6002](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6002))
+- [Multiple Datasource] Add Vega support to MDS by specifying a data source name in the Vega spec ([#5975](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5975))
 
 ### 🐛 Bug Fixes
 
@@ -36,7 +36,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [osd/std] Add additional recovery from false-positives in handling of long numerals ([#5956](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5956))
 - [BUG][Multiple Datasource] Fix missing customApiRegistryPromise param for test connection ([#5944](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5944))
 - [BUG][Multiple Datasource] Add a migration function for datasource to add migrationVersion field ([#6025](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6025))
-- [BUG][MD]Expose picker using function in data source management plugin setup([#6030](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6030)) 
+- [BUG][MD]Expose picker using function in data source management plugin setup([#6030](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6030))
 
 ### 🚞 Infrastructure
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [Multiple Datasource] Concatenate data source name with index pattern name and change delimiter to double colon ([#5907](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5907))
 - [Multiple Datasource] Refactor client and legacy client to use authentication registry ([#5881](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5881))
 - [Multiple Datasource] Improved error handling for the search API when a null value is passed for the dataSourceId ([#5882](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5882))
+- [Multiple Datasource] Add Vega support to MDS by specifying a data source id in the Vega spec ([#5975](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5975))
 
 ### 🐛 Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+- [Multiple Datasource] Add Vega support to MDS by specifying a data source name in the Vega spec ([#5975](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5975))
 
 ### 💥 Breaking Changes
 
@@ -70,7 +71,6 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [Multiple Datasource] Concatenate data source name with index pattern name and change delimiter to double colon ([#5907](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5907))
 - [Multiple Datasource] Refactor client and legacy client to use authentication registry ([#5881](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5881))
 - [Multiple Datasource] Improved error handling for the search API when a null value is passed for the dataSourceId ([#5882](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5882))
-- [Multiple Datasource] Add Vega support to MDS by specifying a data source id in the Vega spec ([#5975](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5975))
 
 ### 🐛 Bug Fixes
 
--- a/src/plugins/vis_type_vega/public/data_model/search_api.ts
+++ b/src/plugins/vis_type_vega/public/data_model/search_api.ts
@@ -31,6 +31,8 @@
 import { combineLatest } from 'rxjs';
 import { map, tap } from 'rxjs/operators';
 import { CoreStart, IUiSettingsClient } from 'opensearch-dashboards/public';
+import { SavedObjectsClientContract } from 'src/core/public';
+import { DataSourceAttributes } from 'src/plugins/data_source/common/data_sources';
 import {
   getSearchParamsFromRequest,
   SearchRequest,
@@ -45,6 +47,8 @@ export interface SearchAPIDependencies {
   uiSettings: IUiSettingsClient;
   injectedMetadata: CoreStart['injectedMetadata'];
   search: DataPublicPluginStart['search'];
+  dataSourceEnabled: boolean;
+  savedObjectsClient: SavedObjectsClientContract;
 }
 
 export class SearchAPI {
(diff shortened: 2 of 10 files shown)


src/plugins/vis_type_vega/public/data_model/search_api.ts
```diff
@@ -70,7 +71,9 @@ export class SearchAPI {
           requestResponders[requestId].json(params.body);
         }
 
-        return search({ params }, { abortSignal: this.abortSignal }).pipe(
+        const searchApiParams = dataSourceId ? { params, dataSourceId } : { params };
+
+        return search(searchApiParams, { abortSignal: this.abortSignal }).pipe(
```
BionIT, line 76: How do we support imported visualization for vega?

src/plugins/vis_type_vega/public/data_model/search_api.ts
```diff
@@ -61,6 +61,7 @@ export class SearchAPI {
     return combineLatest(
       searchRequests.map((request) => {
         const requestId = request.name;
+        const dataSourceId = request.data_source_id;
```
Flyingliuhub, line 64: Can we add test coverage? thanks

src/plugins/vis_type_vega/public/data_model/search_api.ts
bandinib-amzn: +1. Also I am curious about how do we support Sample Vega when we add sample data from remote data source?

bandinib-amzn:
_This need not to be scope in this PR_

Instead of adding a field `data_source_id` to the vega spec under url, can we show a screen to choose data source? cc: @kgcreative 

Once we have selected datasource, we will have the data source id and we can inject that data source id into vega spec(under url) under the hood. I think this will improve user experience overall. 

On top of my mind, flow will remain same as existing vizualization. Instead of showing index pattern, searches, we will show data sources.

Screenshot 2024-03-01 at 4 03 40 PM

CHANGELOG.md
```diff
@@ -70,6 +70,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [Multiple Datasource] Concatenate data source name with index pattern name and change delimiter to double colon ([#5907](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5907))
 - [Multiple Datasource] Refactor client and legacy client to use authentication registry ([#5881](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5881))
 - [Multiple Datasource] Improved error handling for the search API when a null value is passed for the dataSourceId ([#5882](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5882))
+- [Multiple Datasource] Add Vega support to MDS by specifying a data source id in the Vega spec ([#5975](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5975))
```
bandinib-amzn, line 73: can we move this change log to unreleased?

kgcreative:
I think the challenge here is Vega allows for customers to select both index patterns, but also raw indexes

src/plugins/vis_type_vega/public/data_model/search_api.ts
```diff
@@ -70,7 +71,9 @@ export class SearchAPI {
           requestResponders[requestId].json(params.body);
         }
 
-        return search({ params }, { abortSignal: this.abortSignal }).pipe(
+        const searchApiParams = dataSourceId ? { params, dataSourceId } : { params };
```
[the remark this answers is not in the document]
bandinib-amzn: Is this API public? I mean without UI, can I make API call? 

What is dataSourceId is empty and there is no OpenSearch running locally (or `opensearch.hosts:[]` in OSD YML?

seraphjiang:
vega is for advanced user who could write scripting.

I'm leaning for a simple user experience in the initial release if possible.

In the future we might considering  some auto completion or code intelligence feature, or integrate with AI assistant based on requirements 

BTW
@bandinib-amzn @Flyingliuhub @ZilongX @BionIT appreciate if you could help to review the current implementation

seraphjiang:
@BionIT @yujin-emma 

Do we need to change saved object export code in order to support vega in import/export scenarios for MDS. If so, could we create a follow up issue to track?

src/plugins/vis_type_vega/public/data_model/search_api.ts
seraphjiang: +1 for import/export

src/plugins/vis_type_vega/public/data_model/search_api.ts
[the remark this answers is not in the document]
huyaboo: Here are the four cases and expected behaviors:
1. MDS is disabled and no datasource name is provided: visualization will fetch data from local cluster as usual
2. MDS is disabled but a datasource name is provided:  visualization will throw error that MDS is disabled
3. MDS is enabled and no datasource name is provided: visualization will fetch data from local cluster as usual
4. MDS is enabled and a datasource name is provided: visualization will (attempt) to fetch data from remote

src/plugins/vis_type_vega/public/data_model/search_api.ts
huyaboo: Added some test cases to parse the id from the `data_source_name`

src/plugins/vis_type_vega/public/data_model/search_api.ts
huyaboo: Import visualization will be limited in support. 

If Vega visualization were to be imported from a non-MDS enabled cluster -> MDS enabled cluster, then if a datasource is specified, the associated datasource name can be added to each opensearch `url` body (which would require editing the Vega spec directly). This is what can happen when importing sample data. 

However, if the visualization were to be imported from an MDS enabled cluster -> MDS enabled cluster, then there is no current method to map new datasource names to `url` bodies. This is because the import saved objects can only determine 1 datasource. Since the Vega spec supports multiple `url` queries to multiple datasources, the import would inevitably break.

BionIT: @huyaboo import and export can be addressed in separate issue, we can check the vega script inside the saved object to see if datasource name is specified, if not, then we can add the data source name, if yes, then we don't add but checks if the data source exists. Does that simplify it?

src/plugins/vis_type_vega/public/data_model/search_api.ts
```diff
@@ -54,33 +58,72 @@ export class SearchAPI {
...
+    }
+
+    return dataSources.savedObjects.pop()?.id;
+  }
+
+  async dataSourceFindQuery(dataSourceName: string) {
+    return await this.dependencies.savedObjectsClient.find<DataSourceAttributes>({
+      type: 'data-source',
+      perPage: 10,
+      search: `"${dataSourceName}"`,
```
BionIT, line 121: do we need the double quotes?

seraphjiang:
@huyaboo 

Assume we are on track to deliver this in 2.13

Would you work with doc team to update below page. We could remove Vega from limitations

https://opensearch.org/docs/latest/dashboards/management/multi-data-sources/

cc: + @yujin-emma to help if need for process

src/plugins/vis_type_vega/public/data_model/search_api.ts
huyaboo: Yes, the quotes ensure that all the words in the `data_source_name` are in the results. For example, if `Data Source A` and `Data Source B` were data source names, running the `find` query for `Data Source A` would return both names. Logic was also added to ensure the data source name matches the title.

BionIT (approved):

src/plugins/vis_type_vega/public/data_model/search_api.ts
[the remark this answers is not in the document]
bandinib-amzn: What if - MDS is enabled and no datasource name is provided and opensearch.hosts:[]

bandinib-amzn (approved):

BionIT (approved):

bandinib-amzn (dismissed):

src/plugins/vis_type_vega/public/data_model/search_api.ts
```diff
@@ -54,31 +58,75 @@ export class SearchAPI {
...
+    // In the case that data_source_name is a prefix of another name, match exact data_source_name
+    const possibleDataSourceIds = dataSources.savedObjects.filter(
+      (obj) => obj.attributes.title === dataSourceName
     );
+
+    if (possibleDataSourceIds.length !== 1) {
+      throw new Error(
+        `Expected exactly 1 result for data_source_name "${dataSourceName}" but got ${possibleDataSourceIds.length} results`
+      );
+    }
```
xinruiba, line 117: This logic is to check no duplicate datasource names between all datasources.

But we set the paging size 10 here: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5975/files#diff-f1c265abe9e3b3c11a64d4110cf3e6a010dfe369cf6eed9a20ea532763489fbdR125

In that case, will the duplication check only applies to first 10 datasoruces?

xinruiba: This comment is not a blocker of this PR.
I don't think we are able to create datasource with duplication name.

xinruiba (approved):

BionIT (approved):

bandinib-amzn (approved):

PR #6013, opened 2024-03-04
[Workspace] Delete saved objects by workspace

Hailong-am:
### Description

Add delete saved objects by workspace functionality. When delete workspace we need delete all saved objects under this workspace.

### Issues Resolved

Partial of #4944

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


SuZhou-Joe:
A quick question that it seems all the changes are happening on server side, will we add a new public method like deleteByWorkspace in public/workspace.client?

Hailong-am:
> A quick question that it seems all the changes are happening on server side, will we add a new public method like deleteByWorkspace in public/workspace.client?

No, this's part of delete workspace logic at server side. There will have a public method `delete` in public/workspaceClient which will call delete method at server side.

src/core/server/saved_objects/service/lib/repository.ts
```diff
@@ -714,6 +715,55 @@ export class SavedObjectsRepository {
     return body;
   }
 
+  /**
+   * Deletes all objects from the provided workspace. It used when deleting a workspace.
```
ZilongX, line 719: bit confused by this comment - `It used when deleting a workspace.` ?

src/core/server/saved_objects/service/lib/repository.ts
```diff
@@ -714,6 +715,55 @@ export class SavedObjectsRepository {
...
+   *
+   * @param {string} workspace
+   * @param options SavedObjectsDeleteByWorkspaceOptions
+   * @returns {promise} - { took, timed_out, total, deleted, batches, version_conflicts, noops, retries, failures }
+   */
+  async deleteByWorkspace(
+    workspace: string,
```
ZilongX, line 726: Is it taking the workspace name or id as the param ?

ZilongX: if it's actually taking the id maybe let's try to name like `workspaceId` or add the detail in the comment

xluo-aws: It's the id, we use workspace everywhere. workspace name is named differently. If we change it to workspaceId, it may cause more confusion unless we change all to workspaceId.

src/core/server/saved_objects/service/lib/repository.ts
Hailong-am: For now we don't have an action on UI to only delete saved objects belongs to a workspace and not delete workspace itself. This may used in other case in future, let me remove it to avoid confusing.

src/core/server/saved_objects/service/lib/repository.ts
Hailong-am: make sense. As we are using workspace everywhere to reference as its id, to make the name consistent, let me add this to parameter comment.

ZilongX (approved):

Hailong-am:
link check failed due to http://fakenews.co/opensearch-dashboards-test/v6.8.2.json is not reachable in this file [`src/plugins/newsfeed/public/lib/api.test.ts`](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/newsfeed/public/lib/api.test.ts#L476)



```
Errors in src/plugins/newsfeed/public/lib/api.test.ts
+ echo
+ echo ::set-output name=exit_code::2
✗ http://fakenews.co/opensearch-dashboards-test/v6.8.2.json (error sending request for url (http://fakenews.co/opensearch-dashboards-test/v6.8.2.json): connection closed before message completed)
Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/```

File: CHANGELOG.md [hunk 3]
```diff
@@ -30,6 +30,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [Multiple Datasource] Add Vega support to MDS by specifying a data source name in the Vega spec ([#5975](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5975))
 - [Workspace] Consume workspace id in saved object client ([#6014](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6014))
```
SuZhou-Joe, line 32: Nit: useless empty line

SuZhou-Joe (approved):

PR #6014, opened 2024-03-04
[Workspace] Consume workspace id in saved object client

### LINKED ISSUE #6027: [Workspace] Consume workspace in saved objects client.

Workspace will be a concept to manage saved objects, therefore we need to make saved objects client(public side) be aware of what current workspace is and bypass the worksapce params to server side.

Add current workspace a private property of savedObjectsClient and expose a interface to change the property if user switches the workspace. And there may be some cases that user may want to overwrite the default workspace behavior so the client should honor the workspaces params in the input of each method call.

N/A

**Additional context**

SuZhou-Joe:
### Description

This PR adds a new private property in saved objects client to make the client be aware of current workspace. Furthermore when user tries to do Create/Update/Find operations on saved objects, the property will take effect if set and append itself to request payload to indicate it is doing the operation under the specific workspace.

### Issues Resolved

closes #6027 

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


SuZhou-Joe:
#6027

src/core/public/saved_objects/saved_objects_client.ts
```diff
@@ -227,6 +234,11 @@ export class SavedObjectsClient {
     this.batchQueue = [];
   }
 
+  public setCurrentWorkspace(workspaceId: string): boolean {
```
BionIT, line 237: why do we need to have a separate set workspace id when using saved object client? Can workspaces field be within the options? If each workspace should have one saved object client instance, we can add workspace id in the constructor

src/core/public/saved_objects/saved_objects_client.ts
```diff
@@ -249,13 +261,26 @@ export class SavedObjectsClient {
       overwrite: options.overwrite,
     };
 
+    const currentWorkspaceId = this.currentWorkspaceId;
+    let finalWorkspaces;
+    if (options.hasOwnProperty('workspaces')) {
+      finalWorkspaces = options.workspaces;
+    } else if (typeof currentWorkspaceId === 'string') {
+      finalWorkspaces = [currentWorkspaceId];
```
BionIT, line 269: what's the difference of the currentworkspace id vs the one from the payload/options?

src/core/server/saved_objects/import/resolve_import_errors.ts
```diff
@@ -131,6 +131,7 @@ export async function resolveSavedObjectsImportErrors({
     retries,
     createNewCopies,
     dataSourceId,
+    workspaces,
   };
   const checkConflictsResult = await checkConflicts(checkConflictsParams);
```
BionIT, line 136: how would workspaces checked in checkConflicts?

src/plugins/workspace/public/plugin.ts
```diff
@@ -3,16 +3,33 @@
...
+  }
   public async setup() {
     return {};
   }
 
-  public start() {
+  public start(core: CoreStart) {
+    this.coreStart = core;
+
+    this.currentWorkspaceSubscription = this._changeSavedObjectCurrentWorkspace();
```
BionIT, line 28: how is this currentWorkspaceSubscription used? should it be observable instead?

src/plugins/workspace/public/plugin.ts
```diff
@@ -3,16 +3,33 @@
...
  */
 
-import { Plugin } from '../../../core/public';
+import type { Subscription } from 'rxjs';
+import { Plugin, CoreStart } from '../../../core/public';
 
 export class WorkspacePlugin implements Plugin<{}, {}, {}> {
+  private coreStart?: CoreStart;
+  private currentWorkspaceSubscription?: Subscription;
+  private _changeSavedObjectCurrentWorkspace() {
```
BionIT, line 12: what's the benefit of making this as a function? does it affect the lifecycle of corestart object by storing it as a field for the plugin?

BionIT (commented):
left some comments

src/core/public/saved_objects/saved_objects_client.ts
SuZhou-Joe: This method is used in workspace.start lifeCycle, which indicates current workspace id. We support workspaces field within options, but it is import that we add a internal `currentWorkspaceId` so that no matter where calls savedObjectsClient methods, the client will append workspaces fields when doing the real request.

src/core/public/saved_objects/saved_objects_client.ts
SuZhou-Joe: The workspaces fields inside payload/options has a higher priority than currentWorkspaceId.

src/core/server/saved_objects/import/resolve_import_errors.ts
SuZhou-Joe: checkConflicts logic: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/workspace/server/saved_objects/saved_objects_wrapper_for_check_workspace_conflict.ts#L219

src/plugins/workspace/public/plugin.ts
SuZhou-Joe: The benefit is for code readability because we extract all the listener handler codes into a handler function.

SuZhou-Joe:
@BionIT As the PR will contain more than 30 files changes if including both saved objects client changes and saved objects management page changes. So I will reply to your comment on saved objects management page in #6026 .

CHANGELOG.md
```diff
@@ -81,6 +81,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [[Dynamic Configurations] Add support for dynamic application configurations ([#5855](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5855))
 - [Workspace] Optional workspaces params in repository ([#5949](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5949))
 - [Multiple Datasource] Refactoring create and edit form to use authentication registry ([#6002](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6002))
+- [Workspace] Consume workspace id in saved object client ([#6014](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6014))
```
BionIT, line 84: The changelog is under 2.12 right now, can you move to unreleased section above?

src/core/public/saved_objects/saved_objects_client.ts
BionIT: What do you mean to make the workspaces transparent? this is to set the workspace id not get, right? plus, why would plugins need to get current workspace id from the saved objects client. Can we have the workspace id in the global context instead of setting it specific for saved object client, is it the only place where it needs to know the current workspace?

src/plugins/workspace/public/plugin.ts
BionIT: okay, how is this observable currentWorkspaceId$ initialized?

src/core/public/saved_objects/saved_objects_client.ts
BionIT: Can't we just set workspace id into options when using the saved object client?

src/core/public/saved_objects/saved_objects_client.ts
```diff
@@ -348,7 +384,25 @@ export class SavedObjectsClient {
...
-    const renamedQuery = renameKeys<SavedObjectsFindOptions, any>(renameMap, options);
+    const currentWorkspaceId = this.currentWorkspaceId;
+    let finalWorkspaces;
+    if (options.hasOwnProperty('workspaces')) {
+      finalWorkspaces = options.workspaces;
+    } else if (typeof currentWorkspaceId === 'string') {
+      finalWorkspaces = Array.from(new Set([currentWorkspaceId]));
+    }
+
+    const renamedQuery = renameKeys<Omit<SavedObjectsFindOptions, 'ACLSearchParams'>, any>(
```
ruanyl, line 395: seems `ACLSearchParams` doesn't exist at the moment

src/core/public/saved_objects/saved_objects_client.ts
```diff
@@ -183,6 +185,11 @@ const getObjectsToFetch = (queue: BatchQueueEntry[]): ObjectTypeAndId[] => {
 export class SavedObjectsClient {
   private http: HttpSetup;
   private batchQueue: BatchQueueEntry[];
+  /**
+   * if currentWorkspaceId is undefined, it means
+   * we should not carry out workspace info when doing any operation.
```
ruanyl, line 190: I feel this sounds a bit inaccurate, event `currentWorkspaceId` is not set, the `workspaces` in `options` will be used.

src/core/public/saved_objects/saved_objects_client.ts
```diff
@@ -227,6 +234,11 @@ export class SavedObjectsClient {
     this.batchQueue = [];
   }
 
+  public setCurrentWorkspace(workspaceId: string): boolean {
+    this.currentWorkspaceId = workspaceId;
+    return true;
```
ruanyl, line 239: Is `return true` necessary?

src/core/public/saved_objects/saved_objects_client.ts
```diff
@@ -271,11 +296,22 @@ export class SavedObjectsClient {
    * @returns The result of the create operation containing created saved objects.
    */
   public bulkCreate = (
-    objects: SavedObjectsBulkCreateObject[] = [],
+    objects: Array<Omit<SavedObjectsBulkCreateObject, 'workspaces'>> = [],
```
ruanyl, line 299: Could you leave a comment on why `workspaces` should be omit from the passed `objects` when bulk create?

src/core/public/saved_objects/saved_objects_client.ts
SuZhou-Joe: You are right, adjust the comment.

src/core/public/saved_objects/saved_objects_client.ts
SuZhou-Joe: You are right, remove that.

src/core/public/saved_objects/saved_objects_client.ts
SuZhou-Joe: Actually it should not, change back to un-omit.

src/core/public/saved_objects/saved_objects_client.ts
SuZhou-Joe: For a standard `set` method, `return true` is not neeeded. Remove that.

src/core/public/saved_objects/saved_objects_client.ts
```diff
@@ -271,11 +296,22 @@ export class SavedObjectsClient {
...
     const path = this.getPath(['_bulk_create']);
-    const query = { overwrite: options.overwrite };
+    const query: HttpFetchOptions['query'] = { overwrite: options.overwrite };
+    const currentWorkspaceId = this.currentWorkspaceId;
+    let finalWorkspaces;
+    if (options.hasOwnProperty('workspaces')) {
+      finalWorkspaces = options.workspaces;
+    } else if (typeof currentWorkspaceId === 'string') {
+      finalWorkspaces = [currentWorkspaceId];
+    }
```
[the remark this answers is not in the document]
SuZhou-Joe: Thanks for call it out, extracted.

src/core/public/saved_objects/saved_objects_client.ts
SuZhou-Joe: The `setCurrentWorkspace` will only be called by workspace plugin when setup, and once the `currentWorkspaceId` is set, the workspaces field will be automatically append into the final request payload accordingly.

src/core/public/saved_objects/saved_objects_client.ts
SuZhou-Joe: There are 2 sources of workspaces:
- workspaces inside options.
- currentWorkspaceId in the instance.

the `workspaces inside options` has higher priority than the current workspaceId, if somewhere calls a method by specifying `workspaces inside options`, we will honor that and use exactly what they pass to the request payload.

We add `currentWorkspaceId in the instance` because actually there is few case that plugins want to overwrite the `workspaces inside options`, by using the `currentWorkspaceId in the instance`, dashboards page / visualizations page / data source management page will support workspace automatically without a single line of code changes because currentWorkspaceId is already set by workspace plugin when they call any CRUD operations through saved objects client.

src/plugins/workspace/public/plugin.ts
SuZhou-Joe: The `currentWorkspaceId$` will be initialized by a workspace client. You can find the usage here: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6094

CHANGELOG.md
SuZhou-Joe: Sure, done for that.

src/core/public/saved_objects/saved_objects_client.ts
```diff
@@ -183,6 +185,29 @@ const getObjectsToFetch = (queue: BatchQueueEntry[]): ObjectTypeAndId[] => {
 export class SavedObjectsClient {
   private http: HttpSetup;
   private batchQueue: BatchQueueEntry[];
+  /**
+   * The currentWorkspaceId may be undefiend when workspace plugin is not enabled.
```
ruanyl, line 189: typo `undefiend`

src/core/public/saved_objects/saved_objects_client.ts
```diff
@@ -183,6 +185,29 @@ const getObjectsToFetch = (queue: BatchQueueEntry[]): ObjectTypeAndId[] => {
...
+    const currentWorkspaceId = this.currentWorkspaceId;
+    let finalWorkspaces;
+    if (options.hasOwnProperty('workspaces')) {
+      finalWorkspaces = options.workspaces;
+    } else if (typeof currentWorkspaceId === 'string') {
+      finalWorkspaces = [currentWorkspaceId];
+    }
+
+    return finalWorkspaces;
+  }
```
ruanyl, line 210: Nit,  not a big deal, but would this look a bit cleaner?
```suggestion
  private formatWorkspacesParams(options: {
    workspaces?: SavedObjectsCreateOptions['workspaces'];
  }): {workspaces: string[]} | {} {
    const currentWorkspaceId = this.currentWorkspaceId;
    let finalWorkspaces;
    if (options.hasOwnProperty('workspaces')) {
      finalWorkspaces = options.workspaces;
    } else if (typeof currentWorkspaceId === 'string') {
      finalWorkspaces = [currentWorkspaceId];
    }
    
    if (finalWorkspaces) {
      return {workspaces: finalWorkspaces}
    }

    return {};
  }
  
  // then you could do
  const workspaceParams = this.formatWorkspacesParams(options);
```

ruanyl (commented):
Thanks for the quick updating, it looks good to me, just left two more minor comments

src/core/public/saved_objects/saved_objects_client.ts
SuZhou-Joe: Yes, and reduce more duplicate code. Done for that.

src/plugins/saved_objects_management/public/lib/import_file.ts
```diff
@@ -40,12 +40,12 @@ interface ImportResponse {
...
   http: HttpStart,
   file: File,
-  { createNewCopies, overwrite }: ImportMode,
+  { createNewCopies, overwrite, workspaces }: ImportMode,
   selectedDataSourceId?: string
 ) {
   const formData = new FormData();
   formData.append('file', file);
-  const query = createNewCopies ? { createNewCopies } : { overwrite };
+  const query = createNewCopies ? { createNewCopies, workspaces } : { overwrite, workspaces };
```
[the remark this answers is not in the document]
SuZhou-Joe: Move to PR #6026

src/plugins/saved_objects_management/public/lib/resolve_import_errors.test.ts
```diff
@@ -303,4 +303,69 @@ describe('resolveImportErrors', () => {
...
+          },
+        ],
+        importMode: { createNewCopies: false, overwrite: false },
+      },
+      workspaces: ['foo'],
+    });
+    expect(httpMock.post.mock.calls).toMatchInlineSnapshot(`
+      Array [
+        Array [
+          "/api/saved_objects/_resolve_import_errors",
```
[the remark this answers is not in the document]
SuZhou-Joe: Move to PR #6026

src/plugins/saved_objects_management/public/management_section/objects_table/saved_objects_table.tsx
```diff
@@ -167,9 +171,35 @@ export class SavedObjectsTable extends Component<SavedObjectsTableProps, SavedOb
       exportAllOptions: [],
       exportAllSelectedOptions: {},
       isIncludeReferencesDeepChecked: true,
+      currentWorkspaceId: this.props.workspaces.currentWorkspaceId$.getValue(),
+      workspaceEnabled: this.props.applications.capabilities.workspaces.enabled,
     };
   }
 
+  private get workspaceIdQuery() {
+    const { currentWorkspaceId, workspaceEnabled } = this.state;
```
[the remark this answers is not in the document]
SuZhou-Joe: Move to PR #6026

src/plugins/saved_objects_management/public/management_section/objects_table/saved_objects_table.tsx
```diff
@@ -167,9 +171,35 @@ export class SavedObjectsTable extends Component<SavedObjectsTableProps, SavedOb
...
+    } else {
+      if (!currentWorkspaceId) {
+        return undefined;
+      } else {
+        return [currentWorkspaceId];
+      }
+    }
+  }
+
+  private formatWorkspaceIdParams<T extends { workspaces?: string[] }>(
```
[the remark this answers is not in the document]
SuZhou-Joe: Move to PR #6026

src/plugins/saved_objects_management/public/management_section/objects_table/saved_objects_table.tsx
```diff
@@ -167,9 +171,35 @@ export class SavedObjectsTable extends Component<SavedObjectsTableProps, SavedOb
...
+  private get workspaceIdQuery() {
+    const { currentWorkspaceId, workspaceEnabled } = this.state;
+    // workspace is turned off
+    if (!workspaceEnabled) {
+      return undefined;
+    } else {
+      if (!currentWorkspaceId) {
+        return undefined;
+      } else {
+        return [currentWorkspaceId];
```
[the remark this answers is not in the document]
SuZhou-Joe: Move to PR #6026

SuZhou-Joe:
@ruanyl @BionIT Could you please give approval if no more concerns on this PR?

ruanyl (approved):
Thanks for addressing the comments, looks good to me :)

BionIT (approved):

ruanyl (approved):

PR #6032, opened 2024-03-05
adds use of sha256 instead of md5 for active mappings

### LINKED ISSUE #6031: To use more secure sha256sums instead of md5 for active mappings

Currently, opensearch-dashboard builds active mappings with md5 hash string, I suggest using more secure sha256 for the same. MD5 is considered cryptographically broken and unsuitable for further use in any security-sensitive capacity due to vulnerabilities to collision attacks (two different inputs producing the same output) while sha256 is very much more secure.

I have worked on a solution and will link the PR too for that soon

I think this should work itself 

**Additional context**

mritunjaysharma394:
### Description

This change aims to use better and more secure sha256 instead of current md5 for active mappings.

### Issues Resolved

Closes #6031 
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
- [x ] Commits are signed per the DCO using --signoff


mritunjaysharma394:
Hi @bandinib-amzn, just wanted to check if any suggestions on this one, thanks for doing it!

src/core/server/saved_objects/migrations/core/build_active_mappings.ts
```diff
@@ -153,9 +153,9 @@ function canonicalStringify(obj: any): string {
   return `{${sortedObj}}`;
 }
 
-// Convert an object's values to md5 hash strings
+// Convert an object's values to sha256 hash strings
 function md5Values(obj: any) {
```
bandinib-amzn, line 157: nit: can we change name of the function from `md5Values` to `sha256Values`?

bandinib-amzn:
Hi @mritunjaysharma394,

First of all, thank you for opening an issue and raising a PR. I completely agree that SHA-256 is definitely the better choice over MD5.

Having said that, this change necessitate various types of testing.
1. Unit testing: We need to fix existing UT. Because of this change existing UT which compares fields generated by MD5, are failing. 

I am mostly worried about integration, regression, data migration and compatibility.

2. Integration Testing and Regression Testing: We need to make sure it works correctly within the context of the application and does not introduce new bugs or issues in existing functionality.  Can you share a plan for integration and regression testing?

3. Data Migration Testing: Existing data stored using MD5. It will need to migrated to SHA-256. In order to bring this change on board, we need to have data migration plan. And once data migration is in place, we need to ensure that the migration process is accurate and data integrity is maintained through testing.

4. Compatibility Testing: As you might already know, there are many plugins which interacts with OSD. We need make sure it 
does not break compatibility with other systems or components (plugins) that interact with OpenSearch-Dashboards.

There are other testing as well which we can do such as performance testing, security testing, error handling and edge case testing.

@mritunjaysharma394, This is a great start, and I really appreciate you bringing up this topic. However, the scope and impact of this change are significant. If you're interested in working on this, could you please share your plan regarding the previously mentioned points? Thanks!

ananzh:
Convert to draft due to no progress.

Issue #5243, opened 2023-10-07
[RFC][Workspace]Add workspace id into url to indicate current visiting workspace


SuZhou-Joe:
**Is your feature request related to a problem? Please describe.**

In Trinity project, we are gonna to introduce a new concept: Workspace as a place to organize saved objects and features. And `workspace` will be a new type of saved object which has its own id and attributes. As workspace will be a new place to separate saved objects and features, we need to store `current workspaceId` in a place. And this issue will used mainly to discuss the options we can take and their pros and cons.

**Describe the solution you'd like**

## Option 1: store current workspaceId into cookie or sessionStorage.
- description:
  - We can store a key-value pair inside cookie or sessionStorage to indicate current workspaceId, in cookie it looks like document.cookie = '&CURRENT_WORKSPACE_ID={currentWorkspaceId}&' while in sessionStorage in looks like `{ CURRENT_WORKSPACE_ID: {currentWorkspaceId} }`.
- pros
  - 1. Easy to implement.
  - 2. Has little effect on current system.
- cons
  - 1. The url is not shareable. Imaging that a user is visiting a dashboard page under workspaceA, when user tries to copy the url and paste the url to another user, `current workspaceId` information will be missed as the workspace information is stored in a url-non-related place.
  - 2. If we storage the current workspaceId info into cookie, a user can not visit multiple workspaces in a single browser, it is not user-friendly.

## ~~Option 2: store current workspaceId into query or hash.~~ (Use Option 4 if we choose option 2)
- description:
  - We can store a key-value pair inside hash to indicate current workspaceId. Let's use dashboard for example: /app/dashboards/#/view/{dashboardId}?CURRENT_WORKSPACE_ID={currentWorkspaceId}.
- pros
  - 1. Has little effect on current system.
  - 2. The url is shareable as current workspaceId is a part of url.
- cons
  - 1. If we storage the current workspaceId info into query, when user jump from a page with workspace information to a page, the redirect will refresh the page as the query inside target url(without current workspace id) is not the same as the query inside original url(with current workspace), it will totally change the redirect behavior on OSD.
[...]

PR #6060, opened 2024-03-07
[Workspace] Add workspace id in basePath

### LINKED ISSUE #6015: [Workspace] Make url stateful in url with workspace id

In order to let users switch to specific workspace when paste a deep link from other users, we need to make url stateful with workspace info. This PR mainly introduce the capability to keep the workspace Id in url consistent with the workspace id in memory so that user can feel free to copy the url and share to others.

Use the hash to store current workspace id

#5243 

**Additional context**

SuZhou-Joe:
### Description

In order to let users switch to specific workspace when paste a deep link from other users, we need to make url stateful with workspace info. This PR mainly introduce the capability to keep the workspace Id in url consistent with the workspace id in memory so that user can feel free to copy the url and share to others.

You can find all the options and discussions from the RFC issue: #5243 .

### Issues Resolved

#6015 

## Testing the changes

- Enable workspace by adding `workspace.enabled` to opensearch_dashboards.yml file 
- Start OSD by using `yarn start --no-base-path` to make sure no random base path will be appended
- Visit the OpenSearch Dashboards under `foo` workspace: `http://localhost:5601/w/foo/app/dev_tools`
- You will find `/w/foo` params will always be there when you navigate between left navigation, making the url shareable.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


src/core/utils/index.ts
```diff
@@ -37,4 +37,5 @@ export {
   IContextProvider,
 } from './context';
 export { DEFAULT_APP_CATEGORIES } from './default_app_categories';
-export { WORKSPACE_TYPE } from './constants';
+export { WORKSPACE_PATH_PREFIX, WORKSPACE_TYPE } from './constants';
+export { getWorkspaceIdFromUrl, formatUrlWithWorkspaceId, cleanWorkspaceId } from './workspace';
```
ashwin-pc, line 41: Not something we need to decide now, but do you think we should make workspaces a part of core? Right now this PR has a lot of changes in core that are workspace specific. Maybe we just want to keep the UI layer of workspaces in the plugin while the core functionality stays in core. I havent reviewed a lot of the other workspace PR's so this might be what you are already doing so ignore me if thats the case :)

src/plugins/workspace/public/types.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { CoreStart } from '../../../core/public';
+import { WorkspaceClient } from './workspace_client';
+
+export type Services = CoreStart & { workspaceClient: WorkspaceClient };
```
ashwin-pc, line 9: Where is this used?

src/plugins/workspace/server/plugin.ts
```diff
@@ -15,12 +15,29 @@ import { WorkspaceClient } from './workspace_client';
...
 import { WORKSPACE_CONFLICT_CONTROL_SAVED_OBJECTS_CLIENT_WRAPPER_ID } from '../common/constants';
 import { WorkspaceConflictSavedObjectsClientWrapper } from './saved_objects/saved_objects_wrapper_for_check_workspace_conflict';
+import { cleanWorkspaceId, getWorkspaceIdFromUrl } from '../../../core/server/utils';
 
 export class WorkspacePlugin implements Plugin<{}, {}> {
   private readonly logger: Logger;
   private client?: IWorkspaceClientImpl;
   private workspaceConflictControl?: WorkspaceConflictSavedObjectsClientWrapper;
 
+  private proxyWorkspaceTrafficToRealHandler(setupDeps: CoreSetup) {
```
ashwin-pc, line 25: Wouldnt this be better as a part of core and not a plugin?

ashwin-pc (commented):
I like the overall approach. Also learnt a lot about core during this review :) just a few questions about the separation of the logic between core and the workspace plugin but not really blockers for the PR. Will pull down the changes and test as per the test instructions and approve soon after

src/core/public/http/http_service.ts
```diff
@@ -50,9 +52,23 @@ export class HttpService implements CoreService<HttpSetup, HttpStart> {
 
   public setup({ injectedMetadata, fatalErrors }: HttpDeps): HttpSetup {
     const opensearchDashboardsVersion = injectedMetadata.getOpenSearchDashboardsVersion();
+    let workspaceBasePath = '';
+    const plugins = injectedMetadata.getPlugins();
```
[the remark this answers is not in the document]
SuZhou-Joe: Yes but actually what I want to do here is to make the logic here exactly the same as before when workspace is not enabled. But it turns out that it will introduce another indirect dependency on the workspace plugin. I will revert to previous version, in which it will try to construct workspaceBasePath no matter workspace is enabled or not, and let server side to gives a `404` when workspace is disabled, do you think that would be better?

src/core/utils/index.ts
SuZhou-Joe: Actually we have a workspace module in core (https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/workspace/workspaces_service.ts), which simply returns some observables.

I guess you want to make these constants and functions into the core module, but unluckily these utils will be used in both public side and server side, so we had to put it under directory `core/utils`.

src/plugins/workspace/public/types.ts
SuZhou-Joe: Remove this file as it should be in other PRs.

src/plugins/workspace/server/plugin.ts
SuZhou-Joe: We need to put this code here because we want it to be controlled by our feature flag `workspace.enabled`. So only when workspace is enabled will the function take effect.

ashwin-pc:
I was following your test instructions. Inserting a test workspace did not do anything. If i inserted a workspace called `foo` but added `/w/bar` to the URL. `/w/bar` persisted. Deleting the `foo` workspace from the `.kibana` index did not change this behaviour either.

SuZhou-Joe:
> Deleting the foo workspace from the .kibana index did not change this behaviour either.

Correct because in this PR we do not introduce the validation part, so even if there is no specific workspace id in `.kibana` index, the persistent behavior will still be there.

Validation part will be introduced by another PR, which depends on this PR and #6093 .

src/core/public/http/base_path.test.ts
```diff
@@ -110,4 +110,36 @@ describe('BasePath', () => {
       expect(new BasePath('/foo/bar', '/foo').serverBasePath).toEqual('/foo');
     });
   });
+
+  describe('workspaceBasePath', () => {
+    it('get path with workspace', () => {
```
Flyingliuhub, line 115: should we add one more get path without workspace with `withoutWorkspace: true`?

SuZhou-Joe: Actually the `construct BasePath without workspace` should be covered by previous existing test case. But it reminds me that we need to add one on `remove` as it support `withoutWorkspace: true`. Thanks for catching.

src/core/utils/workspace.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  return '';
+};
+
+export const cleanWorkspaceId = (path: string) => {
+  return path.replace(/^\/w\/([^\/]*)/, '');
+};
+
+export const formatUrlWithWorkspaceId = (url: string, workspaceId: string, basePath: IBasePath) => {
```
Flyingliuhub, line 24: Do we handle the edge cases here? like as `workspaceId` is an empty string or `undefined` when `basePath` is not provided or is empty string?

Flyingliuhub, line 24: Should we includes a try-catch block to handle potential errors during URL parsing and formatting? Thanks

SuZhou-Joe: I think that's fine because we should handle the edge cases as what function declares and the basePath has to be present. And the test cases for formatUrlWithWorkspaceId has covered the case when basePath is present but basePath.serverBasePath is empty.

SuZhou-Joe: I think so, the `new URL` may gives some unexpected error.

Flyingliuhub (approved):

src/core/public/http/base_path.ts
```diff
@@ -29,37 +29,47 @@
...
 
 import { modifyUrl } from '@osd/std';
+import type { PrependOptions } from './types';
 
 export class BasePath {
   constructor(
     private readonly basePath: string = '',
-    public readonly serverBasePath: string = basePath
+    public readonly serverBasePath: string = basePath,
+    private readonly workspaceBasePath: string = ''
```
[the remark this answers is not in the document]
SuZhou-Joe: Totally agreed, will change that.

SuZhou-Joe:
@ashwin-pc I think the major concern left is the one that we may introduce an in-direct dependency in core to plugin (https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6060#discussion_r1517106496) , For now there are two options:
1. Leave it as it is and add a TODO comment to optimize that part in the future with a following issue.
2. Revert to the version I mentioned in the response, which will get `clientBasePath` no matter workspace is enabled or not. You can find the diff here: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6060/commits/64b364545164ac3da2e43189ec060ad72f4a559b#diff-f1218ada97a1395de9cad0d115290e84da30f1666d5092e86a12d92a486813b3 .

Which option do you perfer?

src/plugins/workspace/server/plugin.ts
ashwin-pc: Doesnt core have access to config values?

ashwin-pc:
> @ashwin-pc I think the major concern left is the one that we may introduce an in-direct dependency in core to plugin ([#6060 (comment)](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6060#discussion_r1517106496)) , For now there are two options:
> 
> 1. Leave it as it is and add a TODO comment to optimize that part in the future with a following issue.
> 2. Revert to the version I mentioned in the response, which will get `clientBasePath` no matter workspace is enabled or not. You can find the diff here: [64b3645#diff-f1218ada97a1395de9cad0d115290e84da30f1666d5092e86a12d92a486813b3](https://github.com/opensearch-project/OpenSearch-Dashboards/commit/64b364545164ac3da2e43189ec060ad72f4a559b#diff-f1218ada97a1395de9cad0d115290e84da30f1666d5092e86a12d92a486813b3) .
> 
> Which option do you perfer?

Think this goes back to my initial question, dont we already have access to the config in core? Here is where we inject the metadata from the config into the injected metadata ([Link](https://github.com/suzhou-joe/opensearch-dashboards/blob/d01be238e3e5bde0a5652a504fcfa914b04d469c/src/core/server/rendering/rendering_service.tsx#L114)). Branding uses this to decide how to apply branding to the app. We then use this data in core to do things similar to what you are doing here. The difference is that this looks for the config value instead of the existence of a plugin and hence does not have a dependency on the plugin. This should solve both, the need for the front end workspaces component in this PR for rerouting and the selective workspace ID parsing

SuZhou-Joe:
But there are some difference, config for `branding` is a core configuration while the config for `workspace` is a plugin configuration. Plugin configurations are transformed to `uiPlugins`, so `workspace.enabled = true` equals: workspace plugin can be found in `uiPlugins` list. If we try to find the real config `workspace.enabled`, we need to expose the raw config to browser side in server/render_service and I afraid that may expose too much info.

ruanyl:
> @ashwin-pc I think the major concern left is the one that we may introduce an in-direct dependency in core to plugin ([#6060 (comment)](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6060#discussion_r1517106496)) , For now there are two options:
> 
> 1. Leave it as it is and add a TODO comment to optimize that part in the future with a following issue.
> 2. Revert to the version I mentioned in the response, which will get `clientBasePath` no matter workspace is enabled or not. You can find the diff here: [64b3645#diff-f1218ada97a1395de9cad0d115290e84da30f1666d5092e86a12d92a486813b3](https://github.com/opensearch-project/OpenSearch-Dashboards/commit/64b364545164ac3da2e43189ec060ad72f4a559b#diff-f1218ada97a1395de9cad0d115290e84da30f1666d5092e86a12d92a486813b3) .
> 
> Which option do you perfer?

I tend to choose the 2nd approach. If someone tries to visit a workspace URL when workspace is disabled, it makes sense to me as long as it returns 404.

src/core/public/http/base_path.ts
```diff
@@ -29,37 +29,47 @@
...
+        parts.pathname = `${basePath}${parts.pathname}`;
       }
     });
   };
 
-  public remove = (path: string): string => {
-    if (!this.basePath) {
+  public remove = (path: string, prependOptions?: PrependOptions): string => {
+    const { withoutWorkspace } = prependOptions || {};
+    const basePath = withoutWorkspace ? this.basePath : this.get();
```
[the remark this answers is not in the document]
SuZhou-Joe: I was intending to rename the withoutWorkspace to withoutClientBasePath but I was thinking in the future, the clientBasePath may consist of workspace/dataSourceId/other stuff and prependOptions will have withoutWorkspace/withoutDataSourceId/withoutClientBasePath options accordingly so I kept them.

But for now, there is no such case so it has no harm to have only one `withoutClientBasePath` config in prependOptions, but will add some comment here.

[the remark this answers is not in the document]
ruanyl: I see, thanks for the clarification^^

src/plugins/workspace/server/plugin.ts
SuZhou-Joe: Yes, core has access to config values but it is a plugin's behavior to proxy the traffic or not. Core has provided the capability to let plugins to `registerOnPreRouting` and core should not depend on a plugin's config to enable/disable any functionality.

ruanyl (approved):

SuZhou-Joe:
@ashwin-pc Reverted to option 2, could you please help to review here?

src/plugins/workspace/public/plugin.test.ts
```diff
@@ -30,4 +30,22 @@ describe('Workspace plugin', () => {
     coreStart.workspaces.currentWorkspaceId$.next('foo');
     expect(coreStart.savedObjects.client.setCurrentWorkspace).toHaveBeenCalledWith('foo');
   });
+
+  it('#setup when workspace id is in url and enterWorkspace return error', async () => {
```
ashwin-pc, line 34:
```suggestion
  it('should correctly set up WorkspacePlugin when workspace ID is present in the URL', async () => {
```

The current test title does not match whats being tested for. You dont seem to be testing for an error. This is just a suggestion based on what the test seems to be doing

src/core/public/http/types.ts
```diff
@@ -87,25 +87,43 @@ export interface HttpSetup {
...
 
+/**
+ * prepend options
+ *
+ * withoutClientBasePath option will prepend a relative url with serverBasePath only.
+ * For now, clientBasePath is consist of:
+ * workspacePath, which has the pattern of /w/{workspaceId}.
+ *
+ * In the future, clientBasePath may have other parts but keep `withoutClientBasePath` for now to not over-design the interface,
+ */
```
ashwin-pc, line 98: Thanks for adding context here

ashwin-pc (approved):
LGTM.

ruanyl (approved):

src/plugins/workspace/public/plugin.test.ts
SuZhou-Joe: Thanks, will address it in #6154 .

PR #6057, opened 2024-03-06
[MD] Expose a few properties for customize the appearance of the data source selector component

BionIT:
### Description
This change exposes a few properties like 
- defaultOption: the defaulted selected option when the component renders
- placeholderText: the place holder text when the selected option is cleared, default is `Select a data source`
- removePrepend: whether there will be a prepend
- filterFn: the function that will be used to filter the data source options
- compressed

### Issues Resolved

Fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5992

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/beaa10c3-4905-4805-8ff6-0c307d5ff7d1

## Testing the changes
The following steps are performed in the recording above:
1. enable data source plugin
2. go to add sample data and data source picker should render correctly and import sample data from a data source should be successful, going to the dashboard, it should render correctly
3. go to dev tools and data source picker should render correctly and query against a data source should return results
4. go to add sample data to check the options without filtering, it should show the data source named noauth2
5. go to query workbench plugin to check the options with filter to remove noauth2, and it should not show data source named noauth2, query using the selected data source is successful
6. go to search relevance plugin to check the removePrepend which is true and means no prepend should show, and defaultOption which is empty thus placeholder text should show

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #6113, opened 2024-03-11
[Multiple Datasource] Expose filterfn in datasource menu component

BionIT:
### Description

In https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6057, we exposed a few properties in data source selector component for plugins to consume, and in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6082, we created a new data source menu component to be mounted into navigation bar. Based on requirements, we should also expose this filter function to allow plugins to exclude data sources that don't make sense to render in options

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/9a60ceb0-dc3d-4bfe-9b8a-ba8cfd98ae2f

## Testing the changes
1. create a data source with no authentication
2. go to add sample data page, and see that the data source picker shows all available data sources including the one configured with no authentication
3. go to query workbench plugin, and see that the data source menu doesn't show the data source with no auth
4. go to the query workbench implementation, and see the filterfn passed in will exclude the data sources with no auth

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source_management/public/components/data_source_menu/data_source_selectable.test.tsx
+++ b/src/plugins/data_source_management/public/components/data_source_menu/data_source_selectable.test.tsx
@@ -8,17 +8,21 @@ import { SavedObjectsClientContract } from '../../../../../core/public';
 import { notificationServiceMock } from '../../../../../core/public/mocks';
 import React from 'react';
 import { DataSourceSelectable } from './data_source_selectable';
+import { AuthType } from '../../types';
+import { getDataSourcesWithFieldsResponse, mockResponseForSavedObjectsCalls } from '../../mocks';
 
 describe('DataSourceSelectable', () => {
   let component: ShallowWrapper<any, Readonly<{}>, React.Component<{}, {}, any>>;
 
   let client: SavedObjectsClientContract;
   const { toasts } = notificationServiceMock.createStartContract();
+  const nextTick = () => new Promise((res) => process.nextTick(res));
 
   beforeEach(() => {
     client = {
       find: jest.fn().mockResolvedValue([]),
     } as any;
+    mockResponseForSavedObjectsCalls(client, 'find', getDataSourcesWithFieldsResponse);
   });
 
   it('should render normally with local cluster not hidden', () => {
@@ -34,7 +38,7 @@ describe('DataSourceSelectable', () => {
     );
     expect(component).toMatchSnapshot();
     expect(client.find).toBeCalledWith({
-      fields: ['id', 'description', 'title'],
+      fields: ['id', 'title', 'auth.type'],
       perPage: 10000,
       type: 'data-source',
     });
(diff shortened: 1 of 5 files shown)


yujin-emma (approved):

src/plugins/data_source_management/public/components/data_source_menu/data_source_selectable.tsx
```diff
@@ -69,18 +70,24 @@ export class DataSourceSelectable extends React.Component<
 
   async componentDidMount() {
     this._isMounted = true;
-    getDataSources(this.props.savedObjectsClient)
+    getDataSourcesWithFields(this.props.savedObjectsClient, ['id', 'title', 'auth.type'])
```
Flyingliuhub, line 73: how we extend this, which not only can filter auth.type?

src/plugins/data_source_management/public/components/data_source_menu/data_source_selectable.test.tsx
```diff
@@ -54,10 +58,28 @@ describe('DataSourceSelectable', () => {
...
     expect(client.find).toBeCalledWith({
-      fields: ['id', 'description', 'title'],
+      fields: ['id', 'title', 'auth.type'],
       perPage: 10000,
       type: 'data-source',
     });
     expect(toasts.addWarning).toBeCalledTimes(0);
   });
+
+  it('should filter options if configured', async () => {
```
Flyingliuhub, line 68: Where we configure this filterfn, is configurable in the yml file or somewhere or must change the code to configure the filter? thanks

BionIT: This filterFn is passed by plugins when they consume the data source menu component, it is used then in the component when it is mounted to remove unwanted data sources before rendering

src/plugins/data_source_management/public/components/data_source_menu/data_source_selectable.tsx
```diff
@@ -69,18 +70,24 @@ export class DataSourceSelectable extends React.Component<
...
+          let filteredDataSources = [];
+          if (this.props.filterFn) {
+            filteredDataSources = fetchedDataSources.filter((ds) => this.props.filterFn!(ds));
+          }
 
-          dataSourceOptions = dataSourceOptions.sort((a, b) =>
-            a.label.toLowerCase().localeCompare(b.label.toLowerCase())
-          );
+          if (filteredDataSources.length === 0) {
+            filteredDataSources = fetchedDataSources;
```
xinruiba, line 82: If no filteredDataSources exist after apply filter, we will show all dataSource instead. And those datasource won't meet the filter condition.

Will it introduce in some confusions?

Thanks

BionIT: This filter is for plugins to decide how they want to render the options, there is possibility the filter would remove all options then there will be no options displayed. If plugins decide not to add filter options, then it should show all data sources, why it is confusing?

Flyingliuhub (approved):

ZilongX (approved):

src/plugins/data_source_management/public/components/data_source_menu/data_source_selectable.tsx
xinruiba: Suppose we have following dataSources: 
`[{datasource1, NoAuth}, {datasource2, NoAuth}, {datasource3, NoAuth}]`

And filter out all "NoAuth" dataSources.

What is the expected output?

xinruiba (approved):

PR #6123, opened 2024-03-12
[MDS] Add Vega support for importing saved objects

### LINKED ISSUE #5927: [MDS] Vega support with MDS

huyaboo:
### Description
This PR adds (limited) support for importing Vega saved objects when `data_source.enabled: true` and when specifying a datasource. Support in this case means:
- A user can import Vega Visualizations from non-MDS enabled clusters to MDS enabled clusters in `create` mode or in `overwrite` mode
- A user can import Vega Visualizations from other MDS enabled clusters to MDS enabled clusters in `create` mode or in `overwrite` mode

Limitations:
- Only one `data_source_name` can be changed; if users have multiple `data_source_names` specified in their Vega spec, the origin datasource name will be changed but the others will stay the same
- Possibility of duplicate Data-source saved objects being created. The user may potentially have to handle that

### Issues Resolved

Closes #5927 

## Screenshot

When updating the `data_source_name`s, the datasource references also change. In this recording, `Data Source A` is updated to `Data Source B`

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/73027756/bda9745e-8cea-4ad9-b132-4577938eb1d8

Importing a similar visualization that references `Data Source B` and has a local query will update the spec to include `Data Source A` in the old local queries. `Data Source A` is also shown to be added as a reference

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/73027756/ba050fc1-61ab-46c7-bbb2-68d68636c983

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/73027756/58f006d4-e076-47bd-b519-291e3fcac160

## Testing the changes
The following scenarios were tested:

- [x] Import from non-MDS cluster -> non-MDS cluster (existing functionality should not be impacted)
- [x] Import from non-MDS cluster -> MDS cluster (new datasource name should be attached to each OpenSearch query)
- [x] Import from MDS cluster -> MDS cluster (all local queries will be replaced with the new `data_source_name` and the references will append this new datasource id)

### Check List

- [x] All tests pass
  - [ ] `yarn test:jest`


src/core/server/saved_objects/import/check_conflict_for_data_source.ts
```diff
@@ -74,12 +92,37 @@ export async function checkConflictsForDataSource({
...
          */
+
+        // Some visualization types will need special modifications, like Vega visualizations
+        if (object.type === 'visualization') {
+          const vegaSpec = extractVegaSpecFromSavedObject(object);
+
+          if (!!vegaSpec) {
+            const previousDataSourceTitle =
+              previoudDataSourceId && savedObjectsClient
+                ? await getDataSourceTitleFromId(previoudDataSourceId, savedObjectsClient)
```
BionIT, line 103: in this logic, previousDataSourceId is something we get from object id which is not the case for vega right, since we don't save the data source id into vega vis object id?

src/core/server/saved_objects/import/check_conflict_for_data_source.ts
```diff
@@ -74,12 +92,37 @@ export async function checkConflictsForDataSource({
...
+                : undefined;
```
BionIT, line 104: Can we avoid making individual calls if we can make a bulk find request?

src/core/server/saved_objects/import/create_saved_objects.ts
```diff
@@ -83,88 +88,117 @@ export const createSavedObjects = async <T>({
...
-        // @ts-expect-error
-        const searchSourceString = object.attributes?.kibanaSavedObjectMeta?.searchSourceJSON;
-        // @ts-expect-error
-        const visStateString = object.attributes?.visState;
+        // Some visualization types will need special modifications, like Vega visualizations
+        if (object.type === 'visualization') {
+          const vegaSpec = extractVegaSpecFromSavedObject(object);
+
+          if (!!vegaSpec && !!dataSourceTitle) {
+            const idComponents = object.id.split('_');
```
BionIT, line 111: Do we save the data source id as part of vega vis object id?

BionIT:
Can we add the manual test details for this PR?

src/core/server/saved_objects/import/utils.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    if (jsonSpec && typeof jsonSpec === 'object') {
+      return jsonSpec;
+    }
+  } catch (e) {
+    return false;
+  }
+
+  return false;
+};
```
xinruiba, line 107: Instead of returning a json when condition meets and return a boolean when failed.
What if directly return undefined when condition not meet?

```
const parseJSONSpec = (spec: string) => {
  try {
    const jsonSpec = JSON.parse(spec);

    if (jsonSpec && typeof jsonSpec === 'object') {
      return jsonSpec;
    }
  } catch (e) {
    return;
  }

  return;
};
```

src/core/server/saved_objects/import/utils.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export const extractVegaSpecFromSavedObject = (savedObject: SavedObject) => {
+  if (isVegaVisualization(savedObject)) {
+    // @ts-expect-error
+    const visStateObject = JSON.parse(savedObject.attributes?.visState);
+    return visStateObject.params.spec;
+  }
+
+  return false;
```
xinruiba, line 62: Nit:
Since the function named as `extractVegaSpecFromSavedObject`, the called of this func will be expect an object.
What if we return undefined here instead of a boolean. What do you think?

src/core/server/saved_objects/import/check_conflict_for_data_source.ts
```diff
@@ -43,7 +55,13 @@ export async function checkConflictsForDataSource({
...
     new Map<string, SavedObjectsImportRetry>()
   );
-  objects.forEach((object) => {
+
+  const dataSourceTitle =
+    !!dataSourceId && !!savedObjectsClient
+      ? await getDataSourceTitleFromId(dataSourceId, savedObjectsClient)
+      : undefined;
+
+  for await (const object of objects) {
```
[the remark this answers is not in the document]
huyaboo: No I have since made the loop synchronous

src/core/server/saved_objects/import/check_conflict_for_data_source.ts
huyaboo: Yes. Now since we attach datasource references, we do not need a previous datasource id. When a vega saved object is imported from to a datasource, any local queries will be modified to append a `data_source_name` and the new datasourceid reference attached to the references list.

src/core/server/saved_objects/import/check_conflict_for_data_source.ts
huyaboo: Removed repeated individual `get` requests

src/core/server/saved_objects/import/create_saved_objects.ts
huyaboo: Yes. The new datasourceid is still part of the vega vis object id and the reference to that datasourceid is also created.

src/core/server/saved_objects/import/utils.ts
huyaboo: Changed to `undefined`

src/plugins/vis_type_vega/server/vega_visualization_client_wrapper.ts
```diff
@@ -0,0 +1,110 @@
...
+  findDataSourceIdbyName,
+} from './utils';
+import { getDataSourceEnabled } from './services';
+
+export const VEGA_VISUALIZATION_CLIENT_WRAPPER_ID = 'vega-visualization-client-wrapper';
+
+export const vegaVisualizationClientWrapper: SavedObjectsClientWrapperFactory = (
+  wrapperOptions: SavedObjectsClientWrapperOptions
+) => {
+  const createForVega = async <T = unknown>(
```
BionIT, line 24: for create, I think there are a few validations we can do here when mds is enabled:
1. there must be reference data source fields and data source name fields in the data object
2. the data source references must exist - using bulk get
3. the data source references names fetched from 2 should match with all data source names

that means when we call this create API, we need to be able to get the data source references

src/plugins/vis_type_vega/server/utils.ts
```diff
@@ -0,0 +1,98 @@
...
+  if (possibleDataSourceIds.length !== 1) {
+    throw new Error(
+      `Expected exactly 1 result for data_source_name "${dataSourceName}" but got ${possibleDataSourceIds.length} results`
+    );
+  }
+
+  return possibleDataSourceIds.pop()?.id;
+};
+
+export const extractVegaSpecFromAttributes = (attributes: unknown) => {
```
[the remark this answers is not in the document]
huyaboo: Visualization types (Vega, area, timeline) are determined by parsing the stringified `visState` of the object, which is in the `attributes` of a saved object. What's being passed here is the `attributes` field of a visualization saved object, which may or may not be a Vega visualization.

src/plugins/vis_type_vega/server/utils.ts
```diff
@@ -0,0 +1,98 @@
...
+      dataSourceNameSet.add(dataSourceName);
+    }
+  } else {
+    throw new Error(`"data" field should be an object or an array of objects`);
+  }
+
+  return dataSourceNameSet;
+};
+
+const getDataSourceNameFromObject = (dataObject: any) => {
```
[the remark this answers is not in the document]
huyaboo: `dataObject` is parsed from stringified JSON/HJSON, which doesn't have a type at compile time. Additionally, there are multiple types of `dataObjects` as per the Vega [Data](https://vega.github.io/vega/docs/data/) field.

src/plugins/vis_type_vega/server/plugin.ts
```diff
@@ -44,10 +49,19 @@ export class VisTypeVegaPlugin implements Plugin<VisTypeVegaPluginSetup, VisType
...
+  public setup(
+    core: CoreSetup,
+    { home, usageCollection, dataSource }: VisTypeVegaPluginSetupDependencies
+  ) {
     if (usageCollection) {
       registerVegaUsageCollector(usageCollection, this.config, { home });
     }
+    setDataSourceEnabled({ enabled: dataSource?.dataSourceEnabled() || false });
+    core.savedObjects.addClientWrapper(
+      10,
```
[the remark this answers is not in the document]
huyaboo: `10` is the priority number (required) for adding client wrapper. This may need setting up

src/plugins/vis_type_vega/server/vega_visualization_client_wrapper.ts
huyaboo: 1. The only way for a Vega spec to be updated through UI is from the save button. The save button, whether overwriting the existing saved object or creating a new visualization, will call `create` to the saved objects client. Thus, from UI perspective, a saved object cannot have a `data_source_name` field but no associated references. The only way a saved object does not have references is if the Vega spec does not specify a `data_source_name` (local cluster queries only).
2. If a non-existent datasource reference (like a datasource that has been deleted) will be filtered out when updating the `references` field because calling bulk get on the object will return the `id` field and an `error` field instead of the `attributes` field.
```Typescript
(await wrapperOptions.client.bulkGet(existingDataSourceReferences)).saved_objects.forEach(
  (object) => {
    // @ts-expect-error
    if (!!object.attributes && !!object.attributes.title) {
      // @ts-expect-error
      existingDataSourceIdToNameMap.set(object.id, object.attributes.title);
    }
  }
);
```
3. This has also been checked. The datasource reference name must be in the `dataSourceNamesSet` generated from the Vega spec.
```Typescript
const newReferences = options?.references?.filter((reference) => {
  if (reference.type !== 'data-source') {
    return true;
  }
  const dataSourceName = existingDataSourceIdToNameMap.get(reference.id);
  if (dataSourceNamesSet.has(dataSourceName)) {
    dataSourceNamesSet.delete(dataSourceName);
    return true;
  }

  return false;
});
```

**Note:** 
Before saving a Vega visualization, the user must press `update` on the visualization renderer. If a user were to type a nonexistent/outdated `data_source_name`, the renderer should catch the error and return the message to the user before saving. This means that a non-existent datasource name cannot be in the Vega spec (unless the user ignores the error message and saves anyway).

There is a chance that the user will edit the data source names in the data sources page but not the Vega spec and export, which will cause a mismatch in the `data_source_name` and the actual reference. However, this edge case is hard to check because if the visualization has more than one data source reference and the names of these references were all changed, we cannot map the old names -> updated names.

src/plugins/vis_type_vega/server/utils.ts
```diff
@@ -0,0 +1,98 @@
...
+  dataSourceName: string;
+  savedObjectsClient: SavedObjectsClientContract;
+}
+
+export const findDataSourceIdbyName = async (props: FindDataSourceByTitleQueryProps) => {
+  const { dataSourceName } = props;
+  const dataSources = await dataSourceFindQuery(props);
+
+  // In the case that data_source_name is a prefix of another name, match exact data_source_name
+  const possibleDataSourceIds = dataSources.saved_objects.filter(
```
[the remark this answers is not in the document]
huyaboo: changed to `possibleDataSourceObjects`

src/core/server/saved_objects/import/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+export interface UpdateDataSourceNameInVegaSpecProps {
+  spec: string;
+  newDataSourceName: string;
+}
+
+export const updateDataSourceNameInVegaSpec = (
+  props: UpdateDataSourceNameInVegaSpecProps
+): string => {
+  // Exit early if import data cannot determine new datasource name
+  if (!!props.newDataSourceName === false) {
```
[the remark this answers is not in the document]
huyaboo: Removed the check since all calls to this function checks if `dataSourceTitle` exists.

src/core/server/saved_objects/import/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+        keepWsc: true,
+      });
+};
+
+export const getDataSourceTitleFromId = async (
+  dataSourceId: string,
+  savedObjectsClient: SavedObjectsClientContract
+) => {
+  return await savedObjectsClient.get('data-source', dataSourceId).then((response) => {
+    const attributes: any = response?.attributes || undefined;
```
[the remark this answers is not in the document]
huyaboo: Refactored

src/plugins/vis_type_vega/server/vega_visualization_client_wrapper.ts
```diff
@@ -0,0 +1,110 @@
...
+    attributes: T,
+    options?: SavedObjectsCreateOptions
+  ) => {
+    const vegaSpec = extractVegaSpecFromAttributes(attributes);
+    if (type !== 'visualization' || vegaSpec === undefined || !getDataSourceEnabled().enabled) {
+      return await wrapperOptions.client.create(type, attributes, options);
+    }
+    const dataSourceNamesSet = extractDataSourceNamesInVegaSpec(vegaSpec);
+
+    const existingDataSourceIds = options?.references
```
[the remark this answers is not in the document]
huyaboo: Changed to `existingDataSourceReferences`

src/plugins/vis_type_vega/server/vega_visualization_client_wrapper.ts
```diff
@@ -0,0 +1,110 @@
...
+      });
+
+    const existingDataSourceNamesMap = new Map();
+    if (!!existingDataSourceIds && existingDataSourceIds.length > 0) {
+      (await wrapperOptions.client.bulkGet(existingDataSourceIds)).saved_objects.forEach(
+        (object) => {
+          // @ts-expect-error
+          if (!!object.attributes && !!object.attributes.title) {
+            // @ts-expect-error
+            existingDataSourceNamesMap.set(object.id, object.attributes.title);
```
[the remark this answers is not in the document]
huyaboo: Changed to `existingDataSourceIdToNameMap`

src/core/server/saved_objects/import/check_conflict_for_data_source.test.ts
```diff
@@ -140,4 +180,82 @@ describe('#checkConflictsForDataSource', () => {
...
+        errors: [],
+        importIdMap: new Map([
+          [
+            `visualization:some-object-id`,
+            { id: 'some-datasource-id_some-object-id', omitOriginId: true },
+          ],
+        ]),
+      })
+    );
+  });
```
bandinib-amzn, line 226: minor: for better coverage, maybe we want to test that savedobjectclient.get have been called? And also test called with expected attributes?

src/core/server/saved_objects/import/check_conflict_for_data_source.test.ts
```diff
@@ -140,4 +180,82 @@ describe('#checkConflictsForDataSource', () => {
...
```
bandinib-amzn, line 260: same as above

bandinib-amzn: also can we add test for duplicate and undefined data source title?

BionIT (approved):

src/core/server/saved_objects/import/check_conflict_for_data_source.test.ts
huyaboo: Done, added a test to ensure `get()` was called.

src/core/server/saved_objects/import/check_conflict_for_data_source.test.ts
huyaboo: Since this makes a `get` by id, I don't think we need to check for duplicate titles here since there is no ambiguity. The duplicate title check will be done when the visualization is saved and when the visualization is rendered.

BionIT (approved):

ZilongX (dismissed):

BionIT (approved):

BionIT (approved):

bandinib-amzn (approved):

Issue #6115, opened 2024-03-12
[BUG]MDS - Transfer data from MDS enabled data source to MDS disabled data source, hit auth failure


yujin-emma:
**Describe the bug**

I have some saved objects exported with data source, when imported to the default cluster, hit the auth failure

**To Reproduce**
1. Go to Add sample data to some data source ( not local cluster )
2. Go to saved object
3. Select the saved object just added and export
4. Click import button, import the saved objects just exported to local cluster, choose "Checking for existing objects and automatically override "
5. Click the newly imported saved objects
6. Dashboards or viz failed to load the data
7. Go back to the datasource, click on the data source at step1, click "Test connection", hit failing to connect error

**Expected behavior**
Import from MDS to non MDS should work as normal
**OpenSearch Version**
2.12
**Dashboards Version**
2.12
**Plugins**

N/A
**Screenshots**

<img width="1400" alt="Screenshot 2024-03-11 at 16 56 36" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/112784385/df8ee3e0-eea4-474a-8c35-51afd9a697a1">

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

yujin-emma:
Data source object maintain credentials, it should not be exportable as saved object. Therefore, asking user to update the password to reset the auth method is the expected behavior. There is no need to fix the import api

yujin-emma:
Right now, we can import/export data source saved objects. From user friendly perspective, we need to give warning to user that data source object cannot work after export/import. @kgcreative  Could we add warning on the UX

kgcreative:
I think data source objects should not be exportable. Instead, when exporting, we should remove the reference to the data source from all the saved objects. This will also make MDS enabled exports backward compatible with MDS disabled clusters. We might need an additional option on import to associate saved index patterns with existing data sources.

seraphjiang:
> data source objects should not be exportable.

+1 

We might need to clarify this in official doc. 🙏

yujin-emma:
+1 on excluding the data source object when exporting saved object

yujin-emma:
To resolve this issue, we might need to do the following:
1. Add warning on UX to let customer know the data source saved object is not exportable
2. In parallel with 1, in export api, the data source object should be filtered out
3. In the same time, update official doc, if customer hit the issue above, explain the reason and let them update the auth info

[...]

PR #6124, opened 2024-03-12
filter out data source object when export

yujin-emma:
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/server/saved_objects/export/get_sorted_objects_for_export.ts
+++ b/src/core/server/saved_objects/export/get_sorted_objects_for_export.ts
@@ -180,6 +180,7 @@ export async function exportSavedObjectsToStream({
     exportedObjects = sortObjects(rootObjects);
   }
 
+  exportedObjects = exportedObjects.filter((obj) => obj.type !== 'data-source');
   // redact attributes that should not be exported
   const redactedObjects = exportedObjects.map<SavedObject<unknown>>(
     ({ namespaces, ...object }) => object
--- a/config/opensearch_dashboards.yml
+++ b/config/opensearch_dashboards.yml
@@ -247,7 +247,7 @@
 # vis_builder.enabled: false
 
 # Set the value of this setting to true to enable multiple data source feature.
-#data_source.enabled: false
+data_source.enabled: true
 # Set the value of this setting to true to hide local cluster in data source feature.
 #data_source.hideLocalCluster: false
 # Set the value of these settings to customize crypto materials to encryption saved credentials
--- a/config/opensearch_dashboards.yml
+++ b/config/opensearch_dashboards.yml
@@ -247,7 +247,7 @@
 # vis_builder.enabled: false
 
 # Set the value of this setting to true to enable multiple data source feature.
-data_source.enabled: true
+#data_source.enabled: false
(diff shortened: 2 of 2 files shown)


src/core/server/saved_objects/export/get_sorted_objects_for_export.ts
```diff
@@ -180,6 +180,7 @@ export async function exportSavedObjectsToStream({
     exportedObjects = sortObjects(rootObjects);
   }
 
+  exportedObjects = exportedObjects.filter((obj) => obj.type !== 'data-source');
```
ruanyl, line 183: Instead of filtering the final results, shall we filter the `types` passed in before sending it to `savedObjectsClient.find`? Or maybe even throw error if it tries to export data-source object, like in `fetchObjectsToExport` we do validations for the export parameters.

BionIT: +1, I assume we are using the saved object find API and we can specify the object types there

src/core/server/saved_objects/export/get_sorted_objects_for_export.ts
```diff
@@ -180,6 +180,7 @@ export async function exportSavedObjectsToStream({
...
   // redact attributes that should not be exported
```
BionIT, line 184: Can we please add tests for this change as well?

SuZhou-Joe:
I'm thinking why don't we make data-source as a hidden type? Hidden type won't be export as allowed types in saved objects management and can not be performed CRUD operations on by a normal saved objects client. That meets what data source need I suppose.

src/core/server/saved_objects/export/get_sorted_objects_for_export.ts
yujin-emma: Sure, + 1 on using saved object find API and specifying the types; as for 

> Or maybe even throw error if export data-source is not allowed
+ 0.5, error can be some interruption on UX, not quite sure about the user experience, but we can have warning first, more discussion is [here](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6115)

yujin-emma:
> Maybe it is not a good time to ask such question but I'm thinking why don't we make data-source as a hidden type? Hidden type won't be export as allowed types in saved objects management and can not be performed CRUD operations on by a normal saved objects client. That meets what data source needs I suppose.

We still need to list the data source object, but since it contains the auth information, we cannot import/export, more discussion is [here](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6115)

Hailong-am:
If an index pattern depends on datasource, and we don't export datasource will that break the reference? If yes, the user experience is same, all dashboards/visualization will be broken.

we might need remove the reference in index pattern as well when export.

kavilla:
@yujin-emma will remove the 2.14 label as this in draft still.

BionIT:
@yujin-emma Is this PR still being worked on, are we plan to fix in 2.15?

yujin-emma:
Close this pull request since we do not plan to work on filter data source object when export, we are going to stop importing data source objects, and the code change already merged in main :  https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6395

PR #6125, opened 2024-03-12
[Multiple Datasource] Add component to show single selected data source in read only mode

BionIT:
### Description

This change adds component to show single selected data source in read only mode

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/e88fb8e3-076c-49aa-b9fc-7b476d14cd99

## Testing the changes
The following were performed in the recording:
1. enable data source plugin, and see the mounted single selected data source component
2. when clicking on the info icon, a drop down shows which displays the currently used data source
3. disable data source plugin, and we should not see the mounted single selected data source component

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source_management/public/components/data_source_menu/data_source_menu.tsx
+++ b/src/plugins/data_source_management/public/components/data_source_menu/data_source_menu.tsx
@@ -15,15 +15,17 @@ import {
 import { MountPointPortal } from '../../../../opensearch_dashboards_react/public';
 import { DataSourceSelectable } from './data_source_selectable';
 import { DataSourceOption } from '../data_source_selector/data_source_selector';
+import { DataSourceView } from '../data_source_view';
 
 export interface DataSourceMenuProps {
-  showDataSourceSelectable: boolean;
+  showDataSourceSelectable?: boolean;
+  showDataSourceView?: boolean;
   appName: string;
-  savedObjects: SavedObjectsClientContract;
-  notifications: NotificationsStart;
+  savedObjects?: SavedObjectsClientContract;
+  notifications?: NotificationsStart;
   fullWidth: boolean;
-  hideLocalCluster: boolean;
-  dataSourceCallBackFunc: (dataSource: DataSourceOption) => void;
+  hideLocalCluster?: boolean;
+  dataSourceCallBackFunc?: (dataSource: DataSourceOption) => void;
   disableDataSourceSelectable?: boolean;
   className?: string;
   selectedOption?: DataSourceOption[];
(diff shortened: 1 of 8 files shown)


yujin-emma (approved):

Flyingliuhub:
@BionIT can we add the changelog? Thanks

BionIT:
> @BionIT can we add the changelog? Thanks

yeah, the change log has been added!

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -0,0 +1,109 @@
...
+          iconSide="left"
+          size="s"
+          disabled={true}
+        >
+          {this.props.selectedOption && this.props.selectedOption.length > 0
+            ? this.props.selectedOption[0].label
+            : ''}
+        </EuiButtonEmpty>
+        <EuiPopover
+          id={'dataSourceSViewContextMenuPopover'}
```
Flyingliuhub, line 97: typo? dataSource**S**ViewContextMenuPopover or dataSourceViewContextMenuPopover

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  onClick() {
+    this.setState({ ...this.state, isPopoverOpen: !this.state.isPopoverOpen });
+  }
+
+  closePopover() {
+    this.setState({ ...this.state, isPopoverOpen: false });
+  }
+
+  async componentDidMount() {
```
Flyingliuhub, line 46: componentDidMount is empty method, and no logic intend to execute when the component mounts, I'm assuming that if we remove this, it should be the same behavior, right? Thanks

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -0,0 +1,109 @@
...
+        <EuiButtonEmpty
+          className="euiHeaderLink"
+          data-test-subj="dataSourceViewContextMenuHeaderLink"
+          aria-label={i18n.translate('dataSourceView.dataSourceOptionsButtonAriaLabel', {
+            defaultMessage: 'dataSourceViewMenuButton',
+          })}
+          iconType="database"
+          iconSide="left"
+          size="s"
+          disabled={true}
```
Flyingliuhub, line 90: should we have some condition here to dynamically compute the value of `disabled`? like as something`this.props.selectedOption` ? Thanks

src/plugins/data_source_management/public/components/data_source_menu/data_source_menu.tsx
```diff
@@ -15,15 +15,17 @@ import {
...
   appName: string;
-  savedObjects: SavedObjectsClientContract;
-  notifications: NotificationsStart;
+  savedObjects?: SavedObjectsClientContract;
+  notifications?: NotificationsStart;
   fullWidth: boolean;
-  hideLocalCluster: boolean;
-  dataSourceCallBackFunc: (dataSource: DataSourceOption) => void;
+  hideLocalCluster?: boolean;
+  dataSourceCallBackFunc?: (dataSource: DataSourceOption) => void;
```
xinruiba, line 28: Any reasons we set these props as optional?

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
BionIT: This is always disabled based on requirements, since this component is only used in readyonly mode

src/plugins/data_source_management/public/components/data_source_menu/data_source_menu.tsx
BionIT: the callback function doesn't make sense for read only component, since there is no change to the selection, but only makes sense for selectables, hope it makes sense

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
BionIT: right, will fix in new commit

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
BionIT: good catch, will fix

ZilongX (approved):

src/plugins/data_source_management/public/components/data_source_menu/data_source_menu.tsx
xinruiba: Thanks for the response. In that case I assume we don't need this line?:
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6125/files#diff-ca572b8e5224968b4b5cfd2a810b1280195b20f6eff17b93657fefa89bb50ed1R91

xinruiba (approved):

Flyingliuhub (approved):

ZilongX (approved):

PR #6126, opened 2024-03-12
[ignoreFilterIfFieldNotInIndex] Parse query string filters to determine if fields match an index

### LINKED ISSUE #6036: Improve ignoreFilterIfFieldNotInIndex functionality for multi-field query string filters

Our users need to view data from multiple, disparate index patterns on a dashboard. They also need to filter the data they are viewing, and they need only the applicable filters for each index to be applied. As such, we have ignoreFilterIfFieldNotInIndex turned on. This works fine for simple filters (specifically, the ones where `filter.meta.key` points to a single field being filtered on), but it does not work at all for QueryStringFilters. We have a plugin visualization that dynamically constructs a QueryStringFilter, and the filters it creates are incorrectly ignored when the setting is enabled. The filter also adds a "Warning" to the filter's pill in the filter bar.

Currently we are running a fork of Kibana that parses the filter's query string to determine if the filter matches an index. We are using the [lucene npm package](https://www.npmjs.com/package/lucene) to parse the query string. The filterMatchesIndex() function is updated to use this list of fields when checking a QueryStringFilter against an index pattern. To address the filter pill warning, we could either turn it off entirely for QueryStringFilters, or we could call filterMatchesIndex() to determine whether to show the warning.

We have considered creating a different type of filter, but no filters seem to work correctly with multiple fields and ignoreFilterIfFieldNotInIndex enabled. We also briefly considered disabling ignoreFilterIfFieldNotInIndex, but it is not a viable solution for us given the data we want to display.

**Additional context**

The lucene npm package mentioned would be a new dependency.

neodescis:
### Description

For "query_string" filters, parse the query string as lucene to determine if the query's fields matches any given index. This is pertinent when the `ignoreFilterIfFieldNotInIndex` advanced setting is enabled, as filters are only applied to applicable indexes.

Also, for filter bar items, the "Warning" when a filter does not match any indexes has been updated to use the same logic.

### Issues Resolved

closes #6036 

## Changelog
- feat: Parse query string filters to determine if fields match an index when `ignoreFilterIfFieldNotInIndex` is enabled

## Testing the changes

1. Enable `ignoreFilterIfFieldNotInIndex`
2. Go to a dashboard
3. Click Add Filter
4. Click Edit as Query DSL
5. Enter a "query string" filter, such as: `{ "query": { "query_string": { "query": "foo: bar" } } }`
6. Click Save
7. Notice that the filter should now be applied to matching indexes, whereas before it was not
8. Also notice that if the filter does not have an index pattern specifically associated with it, "Warning" will not appear in the filter item's label, whereas before it would

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


kavilla:
Hello @neodescis, apologies about the delay. Going to go through this. I understand the issue, but could you provide some quick insight as that setting has caused some trouble in the past. Do we see any issue with bringing up that helper function and using it where you have it. Any impact to performance that is noticeable?

Thank you!

neodescis:
> Hello @neodescis, apologies about the delay. Going to go through this. I understand the issue, but could you provide some quick insight as that setting has caused some trouble in the past. Do we see any issue with bringing up that helper function and using it where you have it. Any impact to performance that is noticeable?
> 
> Thank you!

Do you mean using `filterMatchesIndex` in `FilterItem`? I would not expect it to have much of a performance impact. Aside from the parsing I added for "query_string" filters, the logic is largely the same anyway. Conceptually, it should be using the same logic: the goal is to warn users of filters that are not applicable, and `filterMatchesIndex` is what determines if they are.

kavilla:
@neodescis, I think perhaps updates to the PR is disabled for this PR?

There's some workflows that are failing and I believe the branch just needs to be updated. Could you enable it or just rebase from main?

Otherwise, all good for approval upon passing workflow. Let me know if you need any assistance. Thank you so much!

neodescis:
Hmm, it seems that GitHub pull requests from another "organization" cannot be set to allow edits by you. That is unfortunate! At any rate, I did just update the branch, so we'll see if that gets the workflow to pass. Do you need a CHANGELOG entry, or anything else?

kavilla:
> Hmm, it seems that GitHub pull requests from another "organization" cannot be set to allow edits by you. That is unfortunate! At any rate, I did just update the branch, so we'll see if that gets the workflow to pass. It seems you have to approve running them every time though, which is also unfortunate. 

That is good to know. I was curious why I couldn't update the branch via GitHub. Thank you for updating it! Re-triggered the workflows.

> Do you need a CHANGELOG entry, or anything else? I was holding off on that one just because merging it is never fun.

I have updated the PR with the template. So after updating the branch you should be good and don't need to change any file. If you don't mind checking the Changelog section I have added to the PR to make sure it's accurate that would be great.

neodescis:
Workflow passed. Changelog entry in the PR description looks good to me. Thanks!

kavilla (approved):
Awesome thank you. Going to try to get this into 2.14.

ananzh (approved):
LGTM

PR #6129, opened 2024-03-13
[Multiple Datasource] Add data source aggregated view to show all compatible data sources or only show used data sources

BionIT:
### Description
This change is based on requirement to show readonly view for all compatible connected data sources in the navigation bar, and also is capable of showing only the data sources passed by the plugins 

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/226c0348-d83e-4264-bc88-808f067a5ea3

## Testing the changes
The following were performed in the recording above:
1. enable data source plugin
2. change prop to not display all compatible data sources and pass in active data source ids should show the info icon and clicking on it, will show drop down with title called `SELECTED DATA SOURCES`, the button should show the number of active/used data sources
3. change prop to display all compatible data sources and remove active data source ids should show the info icon and clicking on it, will show drop down with title called `DATA SOURCES` followed by number of compatible data sources and display them.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
```diff
@@ -0,0 +1,171 @@
...
+import { DataSourceOption } from '../data_source_selector/data_source_selector';
+import { getDataSourcesWithFields } from '../utils';
+
+interface DataSourceAggregatedViewProps {
+  savedObjectsClient: SavedObjectsClientContract;
+  notifications: ToastsStart;
+  hideLocalCluster: boolean;
+  fullWidth: boolean;
+  activeDatasourceIds?: string[];
+  filterFn?: (dataSource: any) => boolean;
```
ruanyl, line 25: Would be nice to avoid `any`

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
```diff
@@ -0,0 +1,171 @@
...
+export class DataSourceAggregatedView extends React.Component<
+  DataSourceAggregatedViewProps,
+  DataSourceAggregatedViewState
+> {
+  private _isMounted: boolean = false;
+
+  constructor(props: DataSourceAggregatedViewProps) {
+    super(props);
+
+    this.state = {
```
ruanyl, line 45: I noticed a typescript error when checking the code locally, not sure if it's just my editor's issue:
```
Type '{ isPopoverOpen: false; allDataSourcesIdMap: Map<any, any>; }' is missing the following properties from type 'Readonly<DataSourceAggregatedViewState>': dataSourceOptions, selectedOption (tsserver 2739)
```

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
```diff
@@ -0,0 +1,171 @@
...
+        if (fetchedDataSources?.length) {
+          let filteredDataSources = fetchedDataSources;
+          if (this.props.filterFn) {
+            filteredDataSources = fetchedDataSources.filter((ds) => this.props.filterFn!(ds));
+          }
+
+          const allDataSourcesIdMap = new Map();
+
+          filteredDataSources.forEach((ds) => {
+            allDataSourcesIdMap.set(ds.id, ds.attributes!.title || '');
```
ruanyl, line 76:
```
Property 'title' does not exist on type '{}'. (tsserver 2339)
```

Here is another TS error I see, may not harm, but maybe we can type `getDataSourcesWithFields` as generic. Something like
```typescript
export async function getDataSourcesWithFields<T = unknown>(
  fields: string[]
) {
  const response = await savedObjectsClient.find<T>({
    type: 'data-source',
    fields,
    perPage: 10000,
  });

  return response?.savedObjects;
}


// then you could call getDataSourcesWithFields with
getDataSourcesWithFields<{title: string, id: string, 'auth.type': string}>()
```

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
```diff
@@ -0,0 +1,171 @@
...
+          });
+
+          if (!this.props.hideLocalCluster) {
+            allDataSourcesIdMap.set('', 'Local cluster');
```
ruanyl, line 80: Q: what's a "local cluster"?

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
```diff
@@ -0,0 +1,171 @@
...
+          let filteredDataSources = fetchedDataSources;
+          if (this.props.filterFn) {
+            filteredDataSources = fetchedDataSources.filter((ds) => this.props.filterFn!(ds));
+          }
+
+          const allDataSourcesIdMap = new Map();
+
+          filteredDataSources.forEach((ds) => {
+            allDataSourcesIdMap.set(ds.id, ds.attributes!.title || '');
+          });
```
xinruiba, line 77: Nit: Maybe we can do it like this:

```
const allDataSourcesIdMap = new Map(filteredDataSources.map(dataSource => [dataSource.id, dataSource.attributes?.title ?? '']));
```

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
```diff
@@ -0,0 +1,171 @@
...
+        return {
+          name: this.state.allDataSourcesIdMap.get(id),
+          disabled: true,
+        };
+      });
+    } else {
+      items = [...this.state.allDataSourcesIdMap.values()].map((v) => {
+        return {
+          name: v,
+          disabled: true,
```
ruanyl, line 123: Just curious, why it's always disabled?

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
```diff
@@ -0,0 +1,171 @@
...
+    // only display active data sources
+    if (this.props.activeDatasourceIds && this.props.activeDatasourceIds.length > 0) {
+      items = this.props.activeDatasourceIds.map((id) => {
+        return {
+          name: this.state.allDataSourcesIdMap.get(id),
+          disabled: true,
+        };
+      });
+    } else {
+      items = [...this.state.allDataSourcesIdMap.values()].map((v) => {
```
xinruiba, line 120: nit:
instead of `'v'`, what if using `'dataSourceTitle'` or some other names which is more meaningful?

yujin-emma (approved):

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
BionIT: I think if we are looping through the values, it will be value thus I used v.

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
```diff
@@ -0,0 +1,171 @@
...
+        );
+      });
+  }
+
+  render() {
+    const button = (
+      <EuiButtonIcon
+        iconType="iInCircle"
+        display="empty"
+        aria-label="Next"
```
ruanyl, line 104: The `aria-label="Next"` doesn't sound applicable, should it be something like `Show data source list`?

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
BionIT: This is a readonly view, thus always disabled, this is based on UX requirements for now

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
BionIT: @xinruiba might be at better position about this local cluster, basically, it refers to the open search cluster running at local host:9200, and for multiple data source, since it enables us to connect to remote data source, thus we need to be able to distinguish local vs remote

ruanyl: That makes sense, thanks for the clarification!

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
BionIT: Changed map name to allDataSourcesIdToTiltleMap and changed the variable name from v to title as well in new commit

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
```diff
@@ -0,0 +1,171 @@
...
+import { SavedObjectsClientContract, ToastsStart } from 'opensearch-dashboards/public';
+import { DataSourceOption } from '../data_source_selector/data_source_selector';
+import { getDataSourcesWithFields } from '../utils';
+
+interface DataSourceAggregatedViewProps {
+  savedObjectsClient: SavedObjectsClientContract;
+  notifications: ToastsStart;
+  hideLocalCluster: boolean;
+  fullWidth: boolean;
+  activeDatasourceIds?: string[];
```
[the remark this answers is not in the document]
BionIT: Thanks for pointing out! Fixed in new commit

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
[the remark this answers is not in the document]
BionIT: Right, I changed the name of the function prop to dataSourceFilter in new commit

BionIT: Thanks, and I change the type to SavedObject type in new commit

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
BionIT: Thanks! The 2 fields are not needed anymore, and cleaned up in new commit

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
BionIT: After adding the type to the function and data source object, the compiler doesn't show unknown prop anymore in new commit

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
BionIT: I think wrapping the function this way would make it a bit hard to read

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
BionIT: Right, changed the aria-label to show data source instead in new commit

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
xinruiba: It's fine, thanks for the response.

ZilongX (approved):

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+          if (!this._isMounted) return;
+          this.setState({
+            ...this.state,
+            allDataSourcesIdToTitleMap,
+          });
+        }
+      })
+      .catch(() => {
+        this.props.notifications.addWarning(
```
Flyingliuhub, line 92: We catch warning notifications here, but we didn't change the component state, do we need to set state to reflect the error condition?

Flyingliuhub (approved):

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
BionIT: If there is error, this will show a toast about the error

src/plugins/data_source_management/public/components/data_source_menu/data_source_menu.tsx
```diff
@@ -40,14 +46,17 @@ export function DataSourceMenu(props: DataSourceMenuProps): ReactElement | null
...
-    filterFn,
+    dataSourceFilter,
+    activeDataSourceIds,
+    displayAllCompatibleDataSources,
   } = props;
 
-  if (!showDataSourceSelectable && !showDataSourceView) {
+  if (!showDataSourceSelectable && !showDataSourceView && !showDataSourceAggregatedView) {
     return null;
   }
```
AMoo-Miki, line 61: These 3 boolean values are used to control the appearance of this component. If it was just one variable, it would have been acceptable but anything more than one variable controlling how _exclusively_ something is rendered is wrong. This has to be changed to a single ENUM variable.

BionIT: Thanks for the comment! Will address this in https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6239 if the proposed change looks good @AMoo-Miki

Issue #5932, opened 2024-02-23
[BUG] Data source plugin authentication registry should not allow name of type AuthType.


bandinib-amzn:
**Describe the bug**

PR #5881 assigns `type` to name property if name is not defined. This might introduce problem when external component or plugin which want to register authentication method and uses `AuthType` as key. We should make clear separation between default auth method and additional auth methods.

PR #6135, opened 2024-03-13
 Improves connection pooling support for AWSSigV4 clients in data sources.

### LINKED ISSUE #5838: [Meta] Refactoring data source plugin to support add-on authentication method with plug-in module
This is meta tracking issue for [Refactoring data source plugin to support add-on authentication method with plug-in module](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5692).

**Tasks:**

- [X] Modify data_source and data_source_management plugin to register authentication method from plugin. 
       PR for above task: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5851/

- [X] Refactoring [client](https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/src/plugins/data_source/server/client) and [legacy client](https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/src/plugins/data_source/server/legacy) to use authentication registry.  
      PR for above task: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5881

- [X] Refactoring [create](https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/src/plugins/data_source_management/public/components/create_data_source_wizard) and [edit](https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/src/plugins/data_source_management/public/components/edit_data_source) form to use authentication registry. 
- [X] https://github.com/opensearch-project/opensearch-js/issues/696
- [ ] #6114

### LINKED ISSUE #6114: Connection pooling support for AWSSigV4 clients/connection in data source
### Overview

There are some use cases where you may need multiple instances of the client. You can easily do that by calling new Client() as many times as you need, but you will lose all the benefits of using one single client, such as the long living connections and the connection pool handling. Connection pooling is an important concept when working with OpenSearch to efficiently manage and reuse network connections. 

[Multi data source client management](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/2.11/docs/multi-datasource/client_management_design.md) provides efficient way to manage multiple clients/connection efficiently, and not consume all the memory.

* For data sources with different endpoint, use client Pooling (E.g. LRU cache)
* For data sources with same endpoint, but different user, use connection pooling strategy (child client) provided by [opensearch-js](https://github.com/opensearch-project/opensearch-js)

### Problem Statement

Currently, we are using [opensearch-js](https://github.com/opensearch-project/opensearch-js/tree/2.3.1) version `2.3.1` to sign requests using sigv4. For legacy datasource client, we are using legacy client package `elasticsearch.js`. As `elasticsearch.js` doesn't provide option to sign requests using sigv4, we used used a third party Connection handler for Amazon ES, [HttpAmazonESConnector](https://github.com/TheDeveloper/http-aws-es). [HttpAmazonESConnector](https://github.com/TheDeveloper/http-aws-es) is no longer being maintained. Therefore, for 2.8.0, while adding support for serverless we used a [drop-in replacement package](https://www.npmjs.com/package/@zhongnansu/http-aws-es?activeTab=readme) of http-aws-es, to add support of configurable service name.

While both libraries provides AWS SigV4 support, it currently lacks an efficient connection pooling mechanism for SigV4 (Signature Version 4) authentication method. 

SigV4 requires clients to include a cryptographic signature in each HTTP request to authenticate and authorize the request. This signature is computed using a combination of the client's AWS access key, secret key, region and other request-specific information. Current architecture does not allow updating credentials for existing client in both [opensearch-js](https://github.com/opensearch-project/opensearch-js) client and legacy client. As a result, every HTTP request to AWS services entails the costly process of creating new client. This leads to resource inefficiency, and performance bottlenecks, particularly in scenarios with high request rates.

#### For opensearch-js:

We have recently released opensearch-js 2.6.0 client to [npm](https://www.npmjs.com/package/@opensearch-project/opensearch/v/2.6.0). 2.6.0 client has [inherited AwsSigV4 in .child](https://github.com/opensearch-project/opensearch-js/pull/725). I propose to upgrade the opensearch-js 2.6.0 and refactor data source plugin to create child client if endpoint/node is same.

Tasks:
[ ] Upgrade `@opensearch/opensearch@2.6.0` which supports AwsSigV4 in .child
[ ] Add support to create child client for AWS sigv4 auth type, similar to basic auth
[ ] Modify client caching mechanism
[ ] For 2.x backports, update dependency version for alias `@opensearch-project/opensearch-next`

#### For Legacy client:

We are using `HttpAmazonESConnector` from `http-aws-es`. As it is no longer being maintained in upstream, I propose to import the HttpAmazonESConnector class and it's UT in OpenSearch-Dashboards repo and maintain ourselves. 

[ ] Copy HttpAmazonESConnector
[ ] Add child support in HttpAmazonESConnector
[ ] Add support to create child client for AWS sigv4 auth type, similar to basic auth
[ ] Modify client caching mechanism

bandinib-amzn:
### Description

As of today, both [opensearch-js](https://github.com/opensearch-project/opensearch-js) client and legacy client does not allow updating credentials for existing client.  As a result, every HTTP request to AWS services entails the costly process of creating new client.

This PR solves exactly that problem. We have recently released opensearch-js 2.6.0 client to [npm](https://www.npmjs.com/package/@opensearch-project/opensearch/v/2.6.0). 2.6.0 client has [inherited AwsSigV4 in .child](https://github.com/opensearch-project/opensearch-js/pull/725). This PR upgrades the opensearch-js 2.6.0 and refactor data source plugin to create child client if endpoint/node is same. Also this PR imports the HttpAmazonESConnector class and it's UT in OpenSearch-Dashboards repo. This will help us to maintain `HttpAmazonESConnector` proactively. 

This PR 
-  Upgrade `@opensearch/opensearch@2.6.0` which supports AwsSigV4 in .child
- Add support to create child client for AWS sigv4 auth type, similar to basic auth
- Modify client caching mechanism
- Copy HttpAmazonESConnector
- Add child support in HttpAmazonESConnector
- Add support to create child client for AWS sigv4 auth type, similar to basic auth
- Does not allow authentication to be registered with the same name as no_auth, username_password, sigv4.

### Issues Resolved

#6114,  #5932, partially resolves #5838

### Check List

- [X] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_source/server/auth_registry/authentication_methods_registry.ts
```diff
@@ -18,6 +19,15 @@ export class AuthenticationMethodRegistery {
    * Authentication Method can only be registered once. subsequent calls with the same method name will throw an error.
    */
   public registerAuthenticationMethod(method: AuthenticationMethod) {
+    if (
+      method.name === AuthType.NoAuth ||
+      method.name === AuthType.UsernamePasswordType ||
+      method.name === AuthType.SigV4
+    ) {
+      throw new Error(
```
zengyan-amazon, line 27: will this block auth types being registered by other plugins?

src/plugins/data_source/server/client/configure_client.ts
```diff
@@ -69,24 +66,25 @@ export const configureClient = async (
...
-    const rootClient = getRootClient(
-      dataSource,
-      openSearchClientPoolSetup.getClientFromPool,
-      dataSourceId
-    ) as Client;
+    let clientPool = openSearchClientPoolSetup;
+    const authenticationMethod = getAuthenticationMethod(dataSource, authRegistry);
+    if (authenticationMethod !== undefined) {
+      clientPool = authenticationMethod.clientPoolSetup;
+    }
```
zengyan-amazon, line 73: what is the type of clientPool? why do we need to override it from the `authenticationMethod`

src/plugins/data_source/server/client/configure_client.ts
```diff
@@ -172,10 +172,12 @@ const getQueryClient = async (
           ? await getAWSCredential(dataSourceAttr, cryptography!)
           : (dataSourceAttr.auth.credentials as SigV4Content));
 
-      const awsClient = rootClient ? rootClient : getAWSClient(credential, clientOptions);
-      addClientToPool(cacheKey, type, awsClient);
+      if (!rootClient) {
+        rootClient = getAWSClient(credential, clientOptions);
```
zengyan-amazon, line 176: for root client, let's not provide any credentials

src/plugins/data_source/server/auth_registry/authentication_methods_registry.ts
bandinib-amzn: This condition does not allow auth types to be registered with the reserved auth type names. So that it won't accidentally interfere with our existing auth types.

src/plugins/data_source/server/client/configure_client.ts
bandinib-amzn: I agree. This is how `getAWSClient` function looks like. I'm just using region from `credential` as it required parameter while creating client. Should we refactor to only pass `region: string` parameter instead of `credential`. I might need to fix UT.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/d273c57081dd3e4652d8d40e613b8f25c72bd25f/src/plugins/data_source/server/client/configure_client.ts#L204-L213

src/plugins/data_source/server/types.ts
```diff
@@ -57,6 +58,8 @@ export interface AuthenticationMethod {
   name: string;
   authType: AuthType;
   credentialProvider: DataSourceCredentialsProvider;
+  clientPoolSetup: OpenSearchClientPoolSetup;
+  legacyClientPoolSetup: OpenSearchClientPoolSetup;
```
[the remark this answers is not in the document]
bandinib-amzn: The intention here was to provide flexibility to auth type developer to build their own client pool. Auth type developer can decide what type of cache they can used, they can define size, build their unique cache key.

src/plugins/data_source/server/client/configure_client.ts
bandinib-amzn: Type is [`OpenSearchClientPoolSetup`](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source/server/client/client_pool.ts#L13).

src/plugins/data_source/server/client/configure_client.ts
```diff
@@ -144,12 +138,18 @@ const getQueryClient = async (
     });
     credential = credentialProvider.credential;
     type = credentialProvider.type;
+
+    if (credential.service === undefined) {
+      credential = { ...credential, service: dataSourceAttr.auth.credentials?.service };
+    }
   }
 
+  const cacheKey = endpoint;
```
xinruiba, line 147: For same endpoint, are there any possibilities that we may send request with different AuthTypes?

For Example:
Test endpoint connection with UsernamePassword and SigV4.

In that case, UsernamePassword client is likely to be reused by SigV4 request?

bandinib-amzn: That's a really good point to bring up. As you can see below, we have two caches - one for regular clients (no auth and basic auth)  and one for aws clients.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/31e8481fd2c0bf185d34c1140724e9320f6a676c/src/plugins/data_source/server/client/client_pool.ts#L24-L32

And while adding and fetching clients from cache, based on type it looks at correct cache.
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/31e8481fd2c0bf185d34c1140724e9320f6a676c/src/plugins/data_source/server/client/client_pool.ts#L76-L87

So answer to your question, UsernamePassword client and SigV4 client will not remain in same cache.

src/plugins/data_source/server/types.ts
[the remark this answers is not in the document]
bandinib-amzn: Removed client pool from auth method. We will use existing client pool. Modified generateCacheKey to take suffix, so that if any auth method want to generate unique key they can just pass suffix.

src/plugins/data_source/server/client/configure_client.ts
bandinib-amzn: As clientpool from auth method removed, this code has been changes in latest revision.

xinruiba (approved):

src/plugins/data_source/server/legacy/http_aws_es/connector.js
```diff
@@ -0,0 +1,134 @@
...
+ * A connection handler for Amazon ES.
+ *
+ * Uses the aws-sdk to make signed requests to an Amazon ES endpoint.
+ *
+ * @param client {Client} - The Client that this class belongs to
+ * @param config {Object} - Configuration options
+ * @param [config.protocol=http:] {String} - The HTTP protocol that this connection will use, can be set to https:
+ * @class HttpConnector
+ */
+import { Config, Credentials } from 'aws-sdk';
```
Flyingliuhub, line 16: It seem like that `Config` and `Credentials` were not used directly in the code. We may remove this. it is not block, we can do it later.

Flyingliuhub (approved):

PR #6154, opened 2024-03-15
[Workspace] Validate if workspace exists when setup inside a workspace

### LINKED ISSUE #6155: [Workspace] Validate the workspace id in the url

For now, workspace plugin will take whatever in url path if the path matches the pattern of workspace, there should be a validation to check if the workspace in url is valid.

Check when workspace start to setup and gives an friendly error page when user can not get access to the specific workspace.

**Additional context**

#4944

SuZhou-Joe:
### Description

When user enters a url with workspaceId pattern inside url, workspace plugin should validate if the id exists, or should show an friendly error page to ask user to go back to home page without a pattern in url.

### Issues Resolved

closes #6155 

## Screenshot

image

## Testing the changes

- Enable workspace feature flag in yml file by set `workspace.enabled: true`
- Start OSD by using `yarn start --no-base-path` to make sure no random base path will be appended
- Navigate to devTools
- Insert test workspaces by calling
```
PUT .kibana/_doc/workspace:foo
{
  "type": "workspace",
  "workspace": {
    "name": "foo"
  }
}
```
- Visit devTools under the test workspace: `http://localhost:5601/w/foo/app/dev_tools`, the page should render successfully.
- Visit devTools under a unexisting workspace: `http://localhost:5601/w/not_existing/app/dev_tools`, should see a page with error just like the screenshot above.
- Visit the error page under a existing worksapce: `http://localhost:5601/w/foo/app/workspace_fatal_error`, user should be redirect to `http://localhost:5601/w/foo/app/workspace_overview` page, though the page has not been implement yet.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/public/application.tsx
+++ b/src/plugins/workspace/public/application.tsx
@@ -1,0 +1,26 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { AppMountParameters, ScopedHistory } from '../../../core/public';
+import { OpenSearchDashboardsContextProvider } from '../../opensearch_dashboards_react/public';
+import { WorkspaceFatalError } from './components/workspace_fatal_error';
+import { Services } from './types';
+
+export const renderFatalErrorApp = (params: AppMountParameters, services: Services) => {
+  const { element } = params;
+  const history = params.history as ScopedHistory<{ error?: string }>;
+  ReactDOM.render(
+    <OpenSearchDashboardsContextProvider services={services}>
+      <WorkspaceFatalError error={history.location.state.error} />
+    </OpenSearchDashboardsContextProvider>,
+    element
+  );
+
+  return () => {
+    ReactDOM.unmountComponentAtNode(element);
+  };
+};
--- a/src/plugins/workspace/public/application.tsx
+++ b/src/plugins/workspace/public/application.tsx
@@ -15,7 +15,7 @@ export const renderFatalErrorApp = (params: AppMountParameters, services: Servic
   const history = params.history as ScopedHistory<{ error?: string }>;
   ReactDOM.render(
     <OpenSearchDashboardsContextProvider services={services}>
-      <WorkspaceFatalError error={history.location.state.error} />
+      <WorkspaceFatalError error={history.location.state?.error} />
     </OpenSearchDashboardsContextProvider>,
     element
   );
--- a/src/plugins/workspace/public/components/workspace_fatal_error/workspace_fatal_error.tsx
+++ b/src/plugins/workspace/public/components/workspace_fatal_error/workspace_fatal_error.tsx
@@ -1,0 +1,68 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+  EuiButton,
+  EuiEmptyPrompt,
+  EuiPage,
+  EuiPageBody,
+  EuiPageContent,
+  EuiCallOut,
+} from '@elastic/eui';
+import React from 'react';
+import { FormattedMessage } from '@osd/i18n/react';
+import { IBasePath } from 'opensearch-dashboards/public';
+import { useOpenSearchDashboards } from '../../../../opensearch_dashboards_react/public';
+import { formatUrlWithWorkspaceId } from '../../../../../core/public/utils';
+
+export function WorkspaceFatalError(props: { error?: string }) {
+  const {
+    services: { application, http },
+  } = useOpenSearchDashboards();
+  const goBackToHome = () => {
+    window.location.href = formatUrlWithWorkspaceId(
+      application?.getUrlForApp('home') || '',
+      '',
+      http?.basePath as IBasePath
+    );
+  };
+  return (
+    <EuiPage style={{ minHeight: '100vh' }}>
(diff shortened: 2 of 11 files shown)


src/plugins/workspace/public/application.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { OpenSearchDashboardsContextProvider } from '../../opensearch_dashboards_react/public';
+import { WorkspaceFatalError } from './components/workspace_fatal_error';
+import { Services } from './types';
+
+export const renderFatalErrorApp = (params: AppMountParameters, services: Services) => {
+  const { element } = params;
+  const history = params.history as ScopedHistory<{ error?: string }>;
+  ReactDOM.render(
+    <OpenSearchDashboardsContextProvider services={services}>
+      <WorkspaceFatalError error={history.location.state.error} />
```
Flyingliuhub, line 18: if error is null, should we consider `const error = history.location.state?.error;` to ensure error exists before accessing it, then it won't throw an error when rendering the component.

src/plugins/workspace/public/plugin.ts
```diff
@@ -33,9 +43,60 @@ export class WorkspacePlugin implements Plugin<{}, {}, {}> {
...
+         * Fatal error service does not support customized actions
+         * So we have to use a self-hosted page to show the errors and redirect.
+         */
+        (async () => {
+          const [{ application, chrome }] = await core.getStartServices();
+          chrome.setIsVisible(false);
+          application.navigateToApp(WORKSPACE_FATAL_ERROR_APP_ID, {
+            replace: true,
+            state: {
+              error: result.error,
```
Flyingliuhub, line 58: do we consider to to check result.error exists before access it. like as `const error = result.error || "Unknown error";` so ensures that `error` always have value

Flyingliuhub (approved):

src/plugins/workspace/public/plugin.ts
SuZhou-Joe: If the error is empty or null, we will hide the callout component to not display the detailed error message, but update to `result?.error` to protect the code.

src/plugins/workspace/public/application.tsx
SuZhou-Joe: Sure, done for that.

src/plugins/workspace/public/components/workspace_fatal_error/workspace_fatal_error.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { FormattedMessage } from '@osd/i18n/react';
+import { IBasePath } from 'opensearch-dashboards/public';
+import { useOpenSearchDashboards } from '../../../../opensearch_dashboards_react/public';
+import { formatUrlWithWorkspaceId } from '../../../../../core/public/utils';
+
+export function WorkspaceFatalError(props: { error?: string }) {
+  const {
+    services: { application, http },
+  } = useOpenSearchDashboards();
+  const goBackToHome = () => {
```
ruanyl, line 24: Should it simply go back to `/` which might  be cleaner?

src/plugins/workspace/public/components/workspace_fatal_error/workspace_fatal_error.tsx
```diff
@@ -0,0 +1,68 @@
...
+                  id="core.fatalErrors.somethingWentWrongTitle"
+                  defaultMessage="Something went wrong"
+                />
+              </h2>
+            }
+            body={
+              <p>
+                <FormattedMessage
+                  id="core.fatalErrors.tryGoBackToDefaultWorkspaceDescription"
+                  defaultMessage="The workspace you want to go can not be found, try go back to home."
```
ruanyl, line 50: Nit:
```suggestion
                  defaultMessage="The workspace you are trying to access cannot be found. Please return to the homepage and try again."
```

src/plugins/workspace/public/components/workspace_fatal_error/workspace_fatal_error.tsx
```diff
@@ -0,0 +1,68 @@
...
+                />
+              </p>
+            }
+            actions={[
+              <EuiButton color="primary" fill onClick={goBackToHome}>
+                <FormattedMessage
+                  id="core.fatalErrors.goBackToHome"
+                  defaultMessage="Go back to home"
```
ruanyl, line 58:
```suggestion
                  defaultMessage="Go back to homepage"
```

src/plugins/workspace/public/components/workspace_fatal_error/workspace_fatal_error.tsx
SuZhou-Joe: In order to go back to homepage of default workspace, we need to:
1. Remove the workspace id inside the url path. (which will be done by second parameter of `formatUrlWithWorkspaceId`)
2. construct the path with potential basePath defined in yml file.

So I think the call to formatUrlWithWorkspaceId is necessary here.

ruanyl: If we let it navigate to `/<basePath>`, then it should be able to automatically redirect to landing page, which by default is the `home` app, what do you think?

SuZhou-Joe: Make sense, we should take `default route` into consideration.Thanks for catching this and updated accordingly.

ruanyl (approved):

PR #6164, opened 2024-03-15
Pass request headers when making application config calls

tianleh:
### Description

Application config plugin allows external plugin to register a different storage other than the default OpenSearch. The external plugin may need some additional information about the request headers when handling the requests.

## Testing the changes

Update the csp rules.

call
```
curl 'http://localhost:5601/api/appconfig/csp.rules' -X POST -H 'Accept: application/json' -H 'Content-Type: application/json' -H 'osd-xsrf: osd-fetch' -H 'Sec-Fetch-Dest: empty' --data-raw '{"newValue":"frame-ancestors https://example-site.org"}'
```
response

```
{"newValue":"frame-ancestors https://example-site.org"}%
```

Confirm on UI that the new value is used.

Screenshot 2024-03-15 at 2 51 39 PM

Check API calls to all configs

`curl 'http://localhost:5601/api/appconfig'`

response

```
{"value":{"csp.rules":"frame-ancestors https://example-site.org"}}%
```

Check API to get only csp rules.

```
curl 'http://localhost:5601/api/appconfig/csp.rules'
```

response

```
{"value":"frame-ancestors https://example-site.org"}%
```

Delete
```
curl 'http://localhost:5601/api/appconfig/csp.rules' -X DELETE -H 'osd-xsrf: osd-fetch' -H 'Sec-Fetch-Dest: empty'
```

Response

```
{"deletedEntity":"csp.rules"}%
```

Get all again

```
curl 'http://localhost:5601/api/appconfig'
```

response
```
{"value":{}}%
```

verify that default csp with frame ancestors is used.

Screenshot 2024-03-15 at 2 52 04 PM

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/application_config/README.md
+++ b/src/plugins/application_config/README.md
@@ -61,13 +61,13 @@ Let's call this plugin `MyConfigurationClientPlugin`.
 First, this plugin will need to implement a class `MyConfigurationClient` based on interface `ConfigurationClient` defined in the `types.ts` under directory `src/plugins/application_config/server/types.ts`. Below are the functions inside the interface.
 
 ```
-  getConfig(): Promise<Map<string, string>>;
+  getConfig(options?: ConfigurationClientOptions): Promise<Map<string, string>>;
 
-  getEntityConfig(entity: string): Promise<string>;
+  getEntityConfig(entity: string, options?: ConfigurationClientOptions): Promise<string>;
 
-  updateEntityConfig(entity: string, newValue: string): Promise<string>;
+  updateEntityConfig(entity: string, newValue: string, options?: ConfigurationClientOptions): Promise<string>;
 
-  deleteEntityConfig(entity: string): Promise<string>;
+  deleteEntityConfig(entity: string, options?: ConfigurationClientOptions): Promise<string>;
 ```
 
 Second, this plugin needs to declare `applicationConfig` as its dependency by adding it to `requiredPlugins` in its own `opensearch_dashboards.json`.
--- a/src/plugins/application_config/README.md
+++ b/src/plugins/application_config/README.md
@@ -70,6 +70,8 @@ First, this plugin will need to implement a class `MyConfigurationClient` based 
   deleteEntityConfig(entity: string, options?: ConfigurationClientOptions): Promise<string>;
 ```
 
+`ConfigurationClientOptions` wraps some additional parameters including request headers.
+
 Second, this plugin needs to declare `applicationConfig` as its dependency by adding it to `requiredPlugins` in its own `opensearch_dashboards.json`.
 
 Third, the plugin will define a new type called `AppPluginSetupDependencies` as follows in its own `types.ts`.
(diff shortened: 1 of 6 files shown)


src/plugins/application_config/server/routes/index.ts
```diff
@@ -85,8 +85,12 @@ export async function handleGetEntityConfig(
   response: OpenSearchDashboardsResponseFactory,
   logger: Logger
 ) {
+  logger.info(`Received a request to get entity config for ${request.params.entity}.`);
```
yujin-emma, line 88: should we keep this logging?

tianleh: I intentionally added it for audit purposes.

src/plugins/application_config/README.md
```diff
@@ -61,13 +61,13 @@ Let's call this plugin `MyConfigurationClientPlugin`.
 First, this plugin will need to implement a class `MyConfigurationClient` based on interface `ConfigurationClient` defined in the `types.ts` under directory `src/plugins/application_config/server/types.ts`. Below are the functions inside the interface.
 
 ```
-  getConfig(): Promise<Map<string, string>>;
+  getConfig(options?: ConfigurationClientOptions): Promise<Map<string, string>>;
 
-  getEntityConfig(entity: string): Promise<string>;
+  getEntityConfig(entity: string, options?: ConfigurationClientOptions): Promise<string>;
```
xinruiba, line 66: Nit:
Thanks for this change, this comment is not a blocker of this PR~

Since `entity` also get passed from `Request`. What if we also put `entity` as an attribute of `ConfigurationClientOptions` and only pass `ConfigurationClientOptions` into these functions ?

tianleh: There is a subtle difference here. The parameter `options?:` has a `?` which indicates an optional parameter. As a contrast, `entity` is a mandatory parameter.

tianleh:
The failed test https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/8331507596/job/22798565980?pr=6164 is not related to my changes.

src/plugins/application_config/server/routes/index.ts
kavilla: I know we have telemetry stuff in the works, where you thinking of audit purposes in terms of usage collection or just how people utilize it?

kavilla (dismissed):

src/plugins/application_config/server/routes/index.ts
tianleh: One case is for ease of troubleshooting why the configuration values are changed.

kavilla (approved):
Cypress test likely to fail unrelated

bandinib-amzn (approved):

bandinib-amzn:
Re-running failed CI

ZilongX (approved):

Issue #4524, opened 2023-07-07
[MD] Issue connecting to data source cluster with self-signed certificates


zhongnansu:
I had problems creating a data source connection to an OpenSearch with HTTPS protocol (created with a self-signed certificate). I received this problem:

<img src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/94870017/968e886e-185f-4583-ba0c-a3f1778b3ec4" alt="drawing" width="350"/>

Response:

```json
{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "Data Source Error: unable to verify the first certificate",
    "attributes": {
        "error": "Data Source Error: unable to verify the first certificate"
    }
}
```

> **observation**: 
> I tested the communication in the `opensearch-dashboard` terminal. 
> I executed `$ curl -k https://opensearch-b:9200` and worked it.

---

When I configure a new data source connection to an OpenSearch with HTTPS protocol with a valid certificate, it works as expected. It also works when I configure OpenSearch with HTTP protocol.

<img src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/94870017/6ae56e4a-cb64-40ea-9845-ef4f6ca5b338" alt="drawing" width="350"/>

---

Do you have some tips or suggestions? Is it a requirement that the data source connection with HTTPS must have a valid certificate? I did not find this information in the [documentation](https://opensearch.org/docs/latest/dashboards/discover/multi-data-sources/).

> - OpenSearch: 2.8.0
> - OpenSearch-Dashboards: 2.8.0
> - _Tests made on docker-compose_

_Originally posted by @carlos-neto-trustly in https://github.com/opensearch-project/OpenSearch-Dashboards/issues/2829#issuecomment-1611585079_

zhongnansu:
Even with `opensearch.ssl.certificateAuthorities: ['path/to/root-ca.pem']` configured in `opensearch-dashboards.yml`. It will still fail. The issues are, 
1. Datasource clients is configured to always validate certificate.
2. The ssl config in yml is not being passed to data source client

Those issues can also be part of https://github.com/opensearch-project/OpenSearch-Dashboards/issues/2009, that we should have a way to support configuration of datasource clients, including SSL configuration

augustoliks:
Hello! @zhongnansu 

I created a docker-compose lab to reproduce the Issue description.

- https://github.com/augustoliks/lab-opensearch/tree/main/multi-data-source-bug

[...]

PR #6171, opened 2024-03-16
[Multiple Datasources] Add TLS configuration for multiple data sources

cwperks:
### Description

This PR adds 2 settings to multiple datasources:

```
data_source.ssl.verificationMode
data_source.ssl.certificateAuthorities
```

Example multi-datasource config:

```
data_source.enabled: true
data_source.ssl.verificationMode: full
data_source.ssl.certificateAuthorities: [ "/Users/cwperx/Projects/opensearch/OpenSearch-Dashboards/config/root-ca.pem" ]
```

These settings are common across all configured datasources so setting `verificationMode` to `none` will disable ssl verification for all datasources. Similarly, the list of all certificateAuthorities will be used to verify across all configured datasources. 

These configuration values are separate then their equivalents when OSD is configured with one OpenSearch cluster where these are used to configure the legacy client to connect with the default OpenSearch cluster:

```
opensearch.ssl.verificationMode
opensearch.ssl.certificateAuthorities
```

### Issues Resolved

- https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4524

## Testing the changes

I have been testing by running 2 clusters locally, both configured with the demo certificates. One cluster is the default OpenSearch and the other cluster I add as another data source.

Tested following scenarios:

1. Both clusters have the security plugin installed and https enabled
     - Able to connect to Cluster 2 when `verificationMode` is set to `none`
     - Able to connect to Cluster 2 when `verificationMode` is set to `certificate` and `certificateAuthorities` are set with correct CAs
     - Able to connect to Cluster 2 when `verificationMode` is set to `full` and `certificateAuthorities` are set with correct CAs
     - Not able to connect to Cluster 2 when `verificationMode` is set to `certificate` and `certificateAuthorities` are set with incorrect CAs
     - Not able to connect to Cluster 2 when `verificationMode` is set to `full` and `certificateAuthorities` are set with incorrect CAs

2. Disable HTTPS on cluster 2
    - Able to connect to cluster 2 with https disabled. When entering hostname make sure the protocol is http instead of https

I am setting this PR to draft while tests are added in the security plugin to verify all of these scenarios.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


seraphjiang:
We might add the new config into default opensearch_dashboards.yml and comment out with recommended default value

cwperks:
> We might add the new config into default opensearch_dashboards.yml and comment out with recommended default value

@seraphjiang Added commented out examples with reasonable defaults to opensearch_dashboards.yml

cwperks:
@seraphjiang In what circumstances is the [legacy_client](https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/src/plugins/data_source/server/legacy) used? I'm looking through the data-source plugin now and it looks like this PR may also need to be extended to the legacy client. Is the legacy client used to connect to ES clusters?

cwperks:
> Is the legacy client used to connect to ES clusters?

I see an open issue around [version decoupling](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6177). I added the legacy client changes into this PR as well.

@BionIT @ZilongX Could you take a look at this PR when you have a chance?

config/opensearch_dashboards.yml
```diff
@@ -281,6 +281,13 @@
 #    'ff00::/8',
 # ]
 
+# Optional setting that enables you to specify a path to PEM files for the certificate
+# authority for your connected datasources.
+#data_source.ssl.certificateAuthorities: [ "/path/to/your/CA.pem" ]
+
+# To disregard the validity of SSL certificates for connected data sources, change this setting's value to 'none'.
+#data_source.ssl.verificationMode: full
```
ZilongX, line 289: do we support only two verificationMode [full, none] or more ?

ZilongX: kk find three support modes [none, full, certificate], maybe let's add more description in the yml entry comment

src/plugins/data_source/server/client/client_config.test.ts
```diff
@@ -26,4 +29,54 @@ describe('parseClientOptions', () => {
       })
     );
   });
+
+  test('test ssl config with verification mode set to none', () => {
```
ZilongX, line 33: looks like we need another test case `test ssl config with verification mode set to certificate`, also curious what's the difference between mode `full` and `certificate`

src/plugins/data_source/server/client/client_config.ts
```diff
@@ -18,14 +28,70 @@ export function parseClientOptions(
...
+    const verificationMode = config.ssl.verificationMode;
+    switch (verificationMode) {
+      case 'none':
+        sslConfig.rejectUnauthorized = false;
+        break;
+      case 'certificate':
+        sslConfig.rejectUnauthorized = true;
+
+        // by default, NodeJS is checking the server identify
+        sslConfig.checkServerIdentity = () => undefined;
```
ZilongX, line 46: ok, so the certificate mode would skip the cert validation whether the certificate cert is issued to hostname - https://nodejs.org/api/tls.html#tlscheckserveridentityhostname-cert

ZilongX:
also triggered rerunning of failed checks, there were some reverted changes in the FT repo yesterday and now all CI should pass

src/plugins/data_source/server/client/client_config.test.ts
cwperks: I will add a test for this case

src/plugins/data_source/server/client/client_config.ts
```diff
@@ -18,14 +28,70 @@ export function parseClientOptions(
...
     plugins: registeredSchema,
   };
 
   return clientOptions;
 }
+
+export const readCertificateAuthorities = (rawConfig: any) => {
+  let certificateAuthorities: string[] | undefined;
+
+  const addCAs = (ca: string[] | undefined) => {
```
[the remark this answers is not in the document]
cwperks: This code is based on a similar method in [src/core/server/opensearch/opensearch_config.ts](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/opensearch/opensearch_config.ts#L364-L425) but only supports `certificateAuthorities` as a setting. 

The lines in this function are hit through the tests for `verificationMode = full`, but I can also add tests specifically for this method too. The code in `opensearch_config.ts` is [tested similarly](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/opensearch/opensearch_config.test.ts) as this PR is.

src/plugins/data_source/server/client/client_config.test.ts
cwperks: Added test for `certificate` mode that asserts that it overwrites `checkServerIdentity`

src/plugins/data_source/server/client/client_config.ts
```diff
@@ -18,14 +28,70 @@ export function parseClientOptions(
...
-      rejectUnauthorized: true,
-    },
+    ssl: sslConfig,
     plugins: registeredSchema,
   };
 
   return clientOptions;
 }
+
+export const readCertificateAuthorities = (rawConfig: any) => {
```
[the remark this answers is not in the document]
cwperks: Any suggestions on where this can be extracted to?

config/opensearch_dashboards.yml
cwperks: Added possible values as a comment in this file.

config/opensearch_dashboards.yml
```diff
@@ -281,6 +281,14 @@
 #    'ff00::/8',
 # ]
 
+# Optional setting that enables you to specify a path to PEM files for the certificate
+# authority for your connected datasources.
+#data_source.ssl.certificateAuthorities: [ "/path/to/your/CA.pem" ]
```
peternied, line 286: @BionIT @ZilongX This configuration introduces a potential security risk by implying a global setting for certificate authorities across all data sources. This approach might not meet sysadmins' expectations, who typically require the ability to specify distinct certificate authorities for each data source for enhanced security. Could we consider revising this design to allow each data source to be securely and separately configured, ensuring attributes such as the certificate authority are isolated?

cwperks: @peternied I have called that out on the corresponding issue here: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4524#issuecomment-1989323830

These settings would explicitly be for a common config if TLS is common between the attached data sources. TLS per data source would require UX when configuring a data source.

src/plugins/data_source/server/client/client_config.ts
xinruiba, line 69: Looks like we only need `rawConfig.ssl.certificateAuthorities` here.
Maybe we can only pass that value as the parameter of this function instead of pass the whole config with an any type. What do you think?

src/plugins/data_source/server/client/client_config.ts
xinruiba, line 72: Nit:
For me, it's more readable to not using abbreviation.
addCAs ----> addCertificateAuthorities
Any concerns?

cwperks: I chose the abbreviation since its consistent with how the opensearch client can be configured [here](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/opensearch/opensearch_config.ts#L364-L425), but I can update this to not use an abbreviation. CA is the standard abbr for certificate authority

src/plugins/data_source/server/client/client_config.ts
cwperks: In the future, this method could be updated to add more TLS configuration like is done with the client for [default opensearch](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/opensearch/opensearch_config.ts#L364-L425). Since only certificateAuthorities is supported for the datasource opensearch client, I will update this to only consume certificateAuthorities as a parameter.

src/plugins/data_source/server/client/client_config.ts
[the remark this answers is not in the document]
cwperks: Removed undefined from types this method consumes since it will never consume undefined.

src/plugins/data_source/server/client/client_config.ts
```diff
@@ -18,14 +28,70 @@ export function parseClientOptions(
...
+    if (ca && ca.length) {
+      certificateAuthorities = [...(certificateAuthorities || []), ...ca];
+    }
+  };
+
+  const ca = rawConfig.ssl.certificateAuthorities;
+  if (ca) {
+    const parsed: string[] = [];
+    const paths = Array.isArray(ca) ? ca : [ca];
+    if (paths.length > 0) {
```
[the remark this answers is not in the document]
cwperks: You're right, it will always be a list and its not necessary to check if its empty or not since the forEach will not execute if its an empty list. Removed this check.

src/plugins/data_source/server/client/client_config.ts
cwperks: Changed this to consume only the certificateAuthorities param.

src/plugins/data_source/server/client/client_config.ts
cwperks: Changed all abbreviations to the long form - certificateAuthorities.

src/plugins/data_source/server/client/client_config.ts
cwperks: Marking this as resolved.

config/opensearch_dashboards.yml
cwperks: @peternied Do you think that this should block the PR or can we add this shared setting to unblock testing with HTTPS in the security-dashboards-plugin? I can file an issue around configured TLS per data-source to track the work. That would either require UX where TLS is configured when adding a data source or possibly a separate config file from `opensearch_dashboards.yml` where you can config your data sources from yml instead of through OSD pages.

cwperks: @BionIT @ZilongX Is there a plan to configure multiple data-sources through a yml file?

cwperks: Filed an issue to track datasource-specific TLS configuration: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6226

src/plugins/data_source/server/client/client_config.ts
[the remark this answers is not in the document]
BionIT: we can use the https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/src/plugins/data_source/server/util to store commonly used utils then add test there.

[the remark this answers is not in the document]
cwperks: Done. I extracted this to a util and added tests specifically for the util to cover 4 scenarios:

1. List of paths
2. Empty list
3. Undefined
4. Single entry

Let me know what you think.

cwperks:
@BionIT @ZilongX I have replied to all PR feedback, can you take another look when you have a chance?

BionIT (approved):

cwperks:
I will open a manual backport

cwperks:
@ZilongX @BionIT Created a manual backport to 2.x for this here: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6244

Issue #6174, opened 2024-03-18
[Multiple Datasource] Filter data source options more dynamically using the data source filter for the data source selector


BionIT:
**Is your feature request related to a problem? Please describe.**
In https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5992, we exposed filter function to allow consumers of the component to specify a filter function which would hide certain data sources before rendering. This function works at initial load but doesn't work dynamically when the filter condition changes.

**Describe the solution you'd like**
Based on requirement for Search Relevance plugin which consumes this component, we need to be able to filter out options based on options passed by parent filter component more dynamically, thus, to resolve, we should store all data sources fetched during initial load, then apply the filter before rendering so that it can filter based on changes in the filter function.

**Describe alternatives you've considered**
An alternative is to expose another prop to only render options passed into the component, then the current component would have more complex logic and it doesn't work well with the existing filter function

Another alternative is to make another component which only renders options passed into it
**Additional context**

PR #6175, opened 2024-03-18
[Multiple Datasource] Use data source filter function before rendering

BionIT:
### Description
This change will store all data sources fetched during initial load, then apply the datasource filter function before rendering so that it can filter based on changes in the filter function.

### Issues Resolved

Fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6174

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/27b5e077-3a90-4094-a311-19c402de1dd3

## Testing the changes
1. go to add sample data, and data source selector should render correctly with currently available data sources
2. add sample data from a data source is successful, and dashboard renders successfully
3. go to dev tool, and data source selector should render correctly with currently available data sources
4. make a query and is successfully returning response
5. go to search relevance and the data source selector should render correctly, with local cluster as default option and no other options available to choose
6. select a data source from the data source menu in the nav bar which acts as filter, then check the data source selector, which should have the selected data source as another option in addition to local cluster

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source_management/public/components/data_source_selector/data_source_selector.tsx
+++ b/src/plugins/data_source_management/public/components/data_source_selector/data_source_selector.tsx
@@ -32,8 +32,8 @@ export interface DataSourceSelectorProps {
 }
 
 interface DataSourceSelectorState {
-  dataSourceOptions: DataSourceOption[];
   selectedOption: DataSourceOption[];
+  allDataSources: Array<SavedObject<DataSourceAttributes>>;
 }
 
 export interface DataSourceOption {
@@ -52,11 +52,7 @@ export class DataSourceSelector extends React.Component<
     super(props);
 
     this.state = {
-      dataSourceOptions: this.props.defaultOption
-        ? this.props.defaultOption
-        : this.props.hideLocalCluster
-        ? []
-        : [LocalCluster],
+      allDataSources: [],
       selectedOption: this.props.defaultOption
         ? this.props.defaultOption
         : this.props.hideLocalCluster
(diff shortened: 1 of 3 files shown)


ruanyl (approved):

xinruiba (approved):

ZilongX (approved):

ZilongX:
LGTM, picked up latest main branch and now rerun all checks

src/plugins/data_source_management/public/components/data_source_selector/data_source_selector.tsx
```diff
@@ -119,6 +99,16 @@ export class DataSourceSelector extends React.Component<
       this.props.placeholderText === undefined
         ? 'Select a data source'
         : this.props.placeholderText;
+
+    const dataSources = this.props.dataSourceFilter
```
Flyingliuhub, line 103: Do we have UT to cover this changes? Thanks

bandinib-amzn (approved):

Flyingliuhub (approved):

src/plugins/data_source_management/public/components/data_source_selector/data_source_selector.tsx
bandinib-amzn: +1. Approved but I'm wondering if we need UT for this change.

BionIT: yes, we use the same unit test for the filtering which was added when the datasource filter was introduced https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source_management/public/components/data_source_selector/data_source_selector.test.tsx#L137

Issue #6177, opened 2024-03-18
[Multiple Datasource] [Plugins Version Decoupling] Fetch and include DataSource version in the data-source saved object when creating a new data source in MDS


ZilongX:
**Is your feature request related to a problem? Please describe.**

This feature is the first solid step per discussion and feedback collected from RFC https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5877. 

In order to decouple the plugin versions with OSD core, first step we're making is to fetch the data source version when creating a new data source which would be consumed by plugins for supporting check

**Describe the solution you'd like**

When we create a new data source in MDS, an info() api call would be fired towards the targeted data source to fetch the data source version, then this version number will be included as part of the data-soruce saved object.

**Describe alternatives you've considered**

We may need to switch to different api for example leveraging the nodes ones since the compatibility mode might be turned on on the data source domain, yet even that the version number would be locked to `7.10.2` which is supposed to be supported by plugins. We still need to further check with each plugin owners to finalized the use cases and this feature for now is more focusing on getting the version in place first.

**Additional context**

[Plugins Version Decoupling] work will be spanning over both `2.13.0` and `2.14.0` releases as planned for now

ZilongX:
Per active RFC https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5877

Issue #5877, opened 2024-02-15
 [RFC] Plugins Version Decoupling


Flyingliuhub:
# Overview
OpenSearch plugins currently tightly couple their versions with the underlying OpenSearch/OpenSearch Dashboards version. This tight coupling can lead to challenges in maintaining and upgrading plugins independently. This RFC proposes a solution to decouple the plugin versions from the OpenSearch versions, allowing for more flexibility in managing and updating plugins.

# Motivation
The motivation behind decoupling OpenSearch plugin versions is to provide a more independent and flexible approach to plugin development and maintenance. This will enable plugin developers to release updates independently of OpenSearch releases, allowing for quicker bug fixes, feature enhancements, and overall better plugin management.

# Proposed Solution

### Plugin Compatibility Matrix
Establish a compatibility matrix that clearly outlines which versions of the plugin are compatible with different versions of OpenSearch. By collecting plugin versions for selected OpenSearch versions and analyzing the compatibility matrix, we can identify potential issues and gaps. Based on the output of the matrix, we can implement a systematic approach to list the possible contents of each plugin, ensuring comprehensive coverage and documentation of compatibility scenarios. This process will not only aid in understanding the compatibility landscape but also guide developers in making informed decisions when selecting and updating plugins for specific OpenSearch versions.

# Scope
In the scope of this proposal, we will focus on selected OpenSearch versions, specifically covering OpenSearch versions ('1.0.0,' '1.1.0,' '1.2.0,' '1.3.0,' '2.3.0,' '2.5.0,' '2.7.0,' '2.9.0', '2.11.0') and Elasticsearch ('7.10'). The plugins included in this analysis are 'alerting,' 'anomaly-detection,' 'index_management,' 'sql,' and 'Machinelearning.'

This focused approach will allow us to thoroughly assess the compatibility matrix for the specified OpenSearch versions and plugins. By concentrating on these specific OpenSearch versions and plugins, we aim to deeply understand how our proposed changes will affect and improve compatibility. The insights gained from this focused analysis will be the basis for developing broader strategies and documentation to ensure plugin compatibility in upcoming OpenSearch versions.

# Execution

[...]

PR #6178, opened 2024-03-18
add version number to newly created datasource object

ZilongX:
### Description

* This change is the first step per RFC https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5877 aiming at decouple the strict version constrains between OSD Core and Plugins

* To make it happen this first step change is to include the data source version number in the newly created `data-source` object which would be consumed by plugins as part of the proof for determining version support

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6177

## Screenshot

* Unit Test passed

image

* Demonstration Recording

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/99905560/c0b82a86-2eb9-4f2b-a34d-1a46433146a0

In the recording we demoed two scenarios:

1. When data source connection checks out, the data source version number is added to the data-source saved object
2. When data source connection doesn't work, the data source version number will be empty but will not block the creation of data-source saved object

## Testing the changes

R1 - https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6178/commits/028ae3cc9af7538fe3678c02a1bd6ef625339512 - `update attribute key from version to dataSourceVersion`

image

R2 - https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6178/commits/9beabfb7734480b4558656cde1d81c475c0f5d5b - `add support and test coverage for custom auth type`

image

### Check List

- [x] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source/server/routes/data_source_connection_validator.ts
+++ b/src/plugins/data_source/server/routes/data_source_connection_validator.ts
@@ -35,4 +35,27 @@ export class DataSourceConnectionValidator {
       throw createDataSourceError(e);
     }
   }
+
+  async fetchDataSourceVersion() {
+    let dataSourceVersion = '';
+    try {
+      // OpenSearch Serverless does not have version concept
+      if (
+        this.dataSourceAttr.auth?.credentials?.service === SigV4ServiceName.OpenSearchServerless
+      ) {
+        return dataSourceVersion;
+      }
+      await this.callDataCluster
+        .info()
+        .then((response) => response.body)
+        .then((body) => {
+          dataSourceVersion = body.version.number;
+        });
+
+      return dataSourceVersion;
+    } catch (e) {
+      // return empty dataSoyrce version instead of throwing exception in case info() api call fails
+      return dataSourceVersion;
+    }
+  }
 }
--- a/src/plugins/data_source/server/routes/fetch_data_source_version.test.ts
+++ b/src/plugins/data_source/server/routes/fetch_data_source_version.test.ts
@@ -1,0 +1,218 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import supertest from 'supertest';
+import { UnwrapPromise } from '@osd/utility-types';
+import { setupServer } from '../../../../../src/core/server/test_utils';
+
+import { IAuthenticationMethodRegistery } from '../auth_registry';
+import { authenticationMethodRegisteryMock } from '../auth_registry/authentication_methods_registry.mock';
+import { CustomApiSchemaRegistry } from '../schema_registry';
+import { DataSourceServiceSetup } from '../../server/data_source_service';
+import { CryptographyServiceSetup } from '../cryptography_service';
+import { registerFetchDataSourceVersionRoute } from './fetch_data_source_version';
+import { AuthType } from '../../common/data_sources';
+// eslint-disable-next-line @osd/eslint/no-restricted-paths
+import { opensearchClientMock } from '../../../../../src/core/server/opensearch/client/mocks';
+
+type SetupServerReturn = UnwrapPromise<ReturnType<typeof setupServer>>;
+
+const URL = '/internal/data-source-management/fetchDataSourceVersion';
+
+describe(`Fetch DataSource Version ${URL}`, () => {
+  let server: SetupServerReturn['server'];
+  let httpSetup: SetupServerReturn['httpSetup'];
+  let handlerContext: SetupServerReturn['handlerContext'];
+  let cryptographyMock: jest.Mocked<CryptographyServiceSetup>;
+  const customApiSchemaRegistry = new CustomApiSchemaRegistry();
+  let customApiSchemaRegistryPromise: Promise<CustomApiSchemaRegistry>;
+  let dataSourceClient: ReturnType<typeof opensearchClientMock.createInternalClient>;
+  let dataSourceServiceSetupMock: DataSourceServiceSetup;
+  let authRegistryPromiseMock: Promise<IAuthenticationMethodRegistery>;
+  const dataSourceAttr = {
+    endpoint: 'https://test.com',
+    auth: {
+      type: AuthType.UsernamePasswordType,
+      credentials: {
+        username: 'testUser',
+        password: 'testPassword',
+      },
+    },
+  };
+
+  const dataSourceAttrMissingCredentialForNoAuth = {
+    endpoint: 'https://test.com',
+    auth: {
+      type: AuthType.NoAuth,
+      credentials: {},
+    },
+  };
+
+  const dataSourceAttrMissingCredentialForBasicAuth = {
+    endpoint: 'https://test.com',
+    auth: {
+      type: AuthType.UsernamePasswordType,
+      credentials: {},
+    },
+  };
+
+  const dataSourceAttrMissingCredentialForSigV4Auth = {
+    endpoint: 'https://test.com',
+    auth: {
+      type: AuthType.SigV4,
+      credentials: {},
+    },
+  };
+
+  const dataSourceAttrPartialCredentialForSigV4Auth = {
+    endpoint: 'https://test.com',
+    auth: {
+      type: AuthType.SigV4,
+      credentials: {
+        accessKey: 'testKey',
+        service: 'service',
+      },
+    },
+  };
+
+  const dataSourceAttrPartialCredentialForBasicAuth = {
+    endpoint: 'https://test.com',
+    auth: {
+      type: AuthType.UsernamePasswordType,
+      credentials: {
+        username: 'testName',
+      },
+    },
+  };
+
+  const dataSourceAttrForSigV4Auth = {
+    endpoint: 'https://test.com',
+    auth: {
+      type: AuthType.SigV4,
+      credentials: {
+        accessKey: 'testKey',
+        service: 'es',
+        secretKey: 'testSecret',
+        region: 'testRegion',
+      },
+    },
+  };
+
+  beforeEach(async () => {
+    ({ server, httpSetup, handlerContext } = await setupServer());
+    customApiSchemaRegistryPromise = Promise.resolve(customApiSchemaRegistry);
+    authRegistryPromiseMock = Promise.resolve(authenticationMethodRegisteryMock.create());
+    dataSourceClient = opensearchClientMock.createInternalClient();
+
+    dataSourceServiceSetupMock = {
+      getDataSourceClient: jest.fn(() => Promise.resolve(dataSourceClient)),
+      getDataSourceLegacyClient: jest.fn(),
+    };
+
+    const router = httpSetup.createRouter('');
+    dataSourceClient.info.mockImplementationOnce(() =>
+      opensearchClientMock.createSuccessTransportRequestPromise({ version: { number: '2.11.0' } })
+    );
+    registerFetchDataSourceVersionRoute(
+      router,
+      dataSourceServiceSetupMock,
+      cryptographyMock,
+      authRegistryPromiseMock,
+      customApiSchemaRegistryPromise
+    );
+
+    await server.start();
+  });
+
+  afterEach(async () => {
+    await server.stop();
+  });
+
+  it('shows successful response', async () => {
+    const result = await supertest(httpSetup.server.listener)
+      .post(URL)
+      .send({
+        id: 'testId',
+        dataSourceAttr,
+      })
+      .expect(200);
+    expect(result.body).toEqual({ dataSourceVersion: '2.11.0' });
+    expect(dataSourceServiceSetupMock.getDataSourceClient).toHaveBeenCalledWith(
+      expect.objectContaining({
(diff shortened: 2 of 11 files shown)


ZilongX:
Will make a video record tomorrow morning ~

src/plugins/data_source/server/routes/fetch_data_source_version.ts
```diff
@@ -0,0 +1,104 @@
...
+        const dataSourceValidator = new DataSourceConnectionValidator(
+          dataSourceClient,
+          dataSourceAttr
+        );
+
+        dataSourceVersion = await dataSourceValidator.fetchDataSourceVersion();
+
+        return response.ok({
+          body: {
+            dataSourceVersion,
```
Flyingliuhub, line 88: should we use `dataSourceversion` in the attr? Thanks

yujin-emma (approved):
LGTM

src/plugins/data_source/server/routes/fetch_data_source_version.ts
ZilongX: yes, let me update the entry key in the attributes

src/plugins/data_source/server/routes/fetch_data_source_version.test.ts
```diff
@@ -0,0 +1,218 @@
...
+import { setupServer } from '../../../../../src/core/server/test_utils';
+
+import { IAuthenticationMethodRegistery } from '../auth_registry';
+import { authenticationMethodRegisteryMock } from '../auth_registry/authentication_methods_registry.mock';
+import { CustomApiSchemaRegistry } from '../schema_registry';
+import { DataSourceServiceSetup } from '../../server/data_source_service';
+import { CryptographyServiceSetup } from '../cryptography_service';
+import { registerFetchDataSourceVersionRoute } from './fetch_data_source_version';
+import { AuthType } from '../../common/data_sources';
+// eslint-disable-next-line @osd/eslint/no-restricted-paths
```
Flyingliuhub, line 17: Should we remove this line 17 if we not comment out here. Thanks

src/plugins/data_source/server/routes/fetch_data_source_version.ts
ZilongX: updated entry key from `version` to `dataSourceVersion`

src/plugins/data_source/server/routes/fetch_data_source_version.ts
```diff
@@ -0,0 +1,104 @@
...
+  router: IRouter,
+  dataSourceServiceSetup: DataSourceServiceSetup,
+  cryptography: CryptographyServiceSetup,
+  authRegistryPromise: Promise<IAuthenticationMethodRegistery>,
+  customApiSchemaRegistryPromise: Promise<CustomApiSchemaRegistry>
+) => {
+  const authRegistry = await authRegistryPromise;
+  router.post(
+    {
+      path: '/internal/data-source-management/fetchDataSourceVersion',
```
Flyingliuhub, line 25: If the user has enabled "compatibility mode" in their [cluster](https://repost.aws/questions/QUwL9qwhP3QGCB7VAx2mxH8A/how-to-enable-compatibility-mode-on-existing-opensearch-domain), it will always be set to version 7.10.2. While this isn't a blocking issue, it's something we should probably take into consideration. Thank you.

src/plugins/data_source/server/routes/fetch_data_source_version.test.ts
ZilongX: need this line to pass the lint check for line 18, similar to https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source/server/routes/test_connection.test.ts#L17, we may need to find a better way for organizing the imports through

Flyingliuhub (dismissed):

src/plugins/data_source/server/routes/fetch_data_source_version.ts
ZilongX: Nice call out ! We've already included this concern in https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6177 specifically as 
```Describe alternatives you've considered

We may need to switch to different api for example leveraging the nodes ones since the compatibility mode might be turned on on the data source domain, yet even that the version number would be locked to 7.10.2 which is supposed to be supported by plugins. We still need to further check with each plugin owners to finalized the use cases and this feature for now is more focusing on getting the version in place first.``` , while it would need more work including coordination with plugin owners we'll for sure keep it in mind and make the api switch if needed

**Review by xinruiba — APPROVED** [review]

File: src/plugins/data_source/server/routes/data_source_connection_validator.ts [hunk 4]
```diff
@@ -35,4 +35,27 @@ export class DataSourceConnectionValidator {
       throw createDataSourceError(e);
     }
   }
+
+  async fetchDataSourceVersion() {
```
BionIT, line 39: Should this be called fetchDataSourceMetadata or do we want to create another API to fetch enabled plugins?

src/plugins/data_source_management/public/types.ts
```diff
@@ -140,6 +140,7 @@ export interface DataSourceAttributes extends SavedObjectAttributes {
   title: string;
   description?: string;
   endpoint?: string;
+  dataSourceVersion?: string;
```
BionIT, line 143: There is another DataSourceAttributes interface in data source plugin, should we update there as well?

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -119,6 +119,27 @@ export async function testConnection(
   });
 }
 
+export async function fetchDataSourceVersion(
+  http: HttpStart,
+  { endpoint, auth: { type, credentials } }: DataSourceAttributes,
+  dataSourceID?: string
```
BionIT, line 125: should data source id be required?

src/plugins/data_source/server/routes/fetch_data_source_version.ts
```diff
@@ -0,0 +1,104 @@
...
+      validate: {
+        body: schema.object({
+          id: schema.maybe(schema.string()),
+          dataSourceAttr: schema.object({
+            endpoint: schema.string(),
+            auth: schema.maybe(
```
BionIT, line 31: for auth, I think we are introducing custom auth types, thus need to be similar to https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source/server/routes/test_connection.ts#L56, @xinruiba might have more context

src/plugins/data_source/server/routes/fetch_data_source_version.ts
```diff
@@ -0,0 +1,104 @@
...
+                  }),
+                }),
+              ])
+            ),
+          }),
+        }),
+      },
+    },
+    async (context, request, response) => {
+      const { dataSourceAttr, id: dataSourceId } = request.body;
```
BionIT, line 63: since data source id is optional, how do we handle undefined data source id here?

src/plugins/data_source/server/routes/data_source_connection_validator.ts
```diff
@@ -35,4 +35,27 @@ export class DataSourceConnectionValidator {
...
+    let dataSourceVersion = '';
+    try {
+      // OpenSearch Serverless does not have version concept
+      if (
+        this.dataSourceAttr.auth?.credentials?.service === SigV4ServiceName.OpenSearchServerless
```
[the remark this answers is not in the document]
ZilongX: Yeah would like to learn more on the serverless/premise side, actually we may need to add another attri entry say dataSource type maybe (OS | OS Serverless | On Premise)

src/plugins/data_source/server/routes/data_source_connection_validator.ts
ZilongX: yes, this is just the first step to fetch the version number, next one is to fetch the installed plugins which would be consolidated into one method say fetchDataSourceMetadata with multiple api calls behind the scene

src/plugins/data_source_management/public/types.ts
ZilongX: yes but maybe later, that one seems to be leveraged for data source display page, this one is focusing on the saving part, we may need to display the version number on the datasource list page, with some UI updates as well, will need to check further.

src/plugins/data_source/server/routes/fetch_data_source_version.ts
ZilongX: same as the test connection call , dataSourceId is actually needed for the OS API call but more on the MD internal side for validation to make sure the call is actually triggered by data source creation.

src/plugins/data_source_management/public/components/utils.ts
bandinib-amzn: I think it is optional because of test connection. Because if we do test connection before saving data source, we won't have data source id. But what will happen if it empty or undefined? Does it throw error?

src/plugins/data_source/server/routes/data_source_connection_validator.ts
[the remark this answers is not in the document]
xinruiba: We use `attributes.auth.credentials.service` to determine the service type for sigV4 dataSources. 

This attribute, as far as I know, only has effect on how to deal with credentials in server side. But I don't think we add this server type as an attribute of our datasource object.

src/plugins/data_source/server/routes/data_source_connection_validator.ts
```diff
@@ -35,4 +35,27 @@ export class DataSourceConnectionValidator {
...
+      ) {
+        return dataSourceVersion;
+      }
+      await this.callDataCluster
+        .info()
+        .then((response) => response.body)
+        .then((body) => {
```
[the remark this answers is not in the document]
BionIT: It's redundant to use both await and then in the same function.

src/plugins/data_source/server/routes/fetch_data_source_version.ts
bandinib-amzn: Yes. @ZilongX can you add schema validation for custom auth type?

src/plugins/data_source/server/routes/data_source_connection_validator.ts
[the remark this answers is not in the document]
xinruiba: I'm thinking if the server type is a must have, are we able to decide the service type based on the open-search endpoint format?

Like server endpoint (have **es**):
https://someurl.region.es.amazonaws.com/

Serverless endpoint (have **aoss**):
https://someurl.region.aoss.amazonaws.com/

src/plugins/data_source/server/routes/fetch_data_source_version.test.ts
```diff
@@ -0,0 +1,218 @@
...
+    const result = await supertest(httpSetup.server.listener)
+      .post(URL)
+      .send({
+        id: 'testId',
+        dataSourceAttr: dataSourceAttrForSigV4Auth,
+      })
+      .expect(200);
+    expect(result.body).toEqual({ dataSourceVersion: '2.11.0' });
+  });
+});
```
bandinib-amzn, line 218: Can you please add test case for cutsom auth as well?

src/plugins/data_source/server/routes/data_source_connection_validator.ts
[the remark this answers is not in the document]
ZilongX: Good call out ! Let me check further on above when adding the dataSourceType to data-source (in case it's needed)

src/plugins/data_source_management/public/components/utils.ts
ZilongX: just tested and it won't cause any exception :)  further digging indicates a maybe check`id: schema.maybe(schema.string()),` so an empty one won't cause trouble for now

src/plugins/data_source/server/routes/fetch_data_source_version.ts
ZilongX: Nice call out, let me add the validation for custom auth tpe

src/plugins/data_source/server/routes/fetch_data_source_version.test.ts
ZilongX: Yes, let me add the validation and test for custom auth

xinruiba:
@ZilongX Thanks for the change.

It will be good if we can also add a video to test registered Auth Type.
Feel free to sync up offline~

Flyingliuhub (dismissed):

xinruiba (approved):

xinruiba:
> @ZilongX Thanks for the change.
> 
> It will be good if we can also add a video to test registered Auth Type. Feel free to sync up offline~

Synced up with @ZilongX offline and test registered Auth Type together, it works.

Flyingliuhub (approved):

bandinib-amzn (approved):

bandinib-amzn:
All checks are passing.

PR #6190, opened 2024-03-19
[ignoreFilterIfFieldNotInIndex] Build separate Timeline filters by index

### LINKED ISSUE #6184: [BUG] Timeline visualization does not work with ignoreFilterIfFieldNotInIndex enabled
**Describe the bug**

The timeline visualization applies all filters to all indexes from which it is retrieving data, regardless of the ignoreFilterIfFieldNotInIndex setting.

**To Reproduce**
1. Create dashboard with visualizations for more than one index
2. Create timeline visualization for one of those indexes and add it to the dashboard
3. Create filter that applies only to the index *not* referenced by the timeline visualization
4. See the timeline visualization flatline, because the filter is applied to it

**Expected behavior**
The timeline visualization should apply filters only to applicable indexes if ignoreFilterIfFieldNotInIndex is enabled.

**OpenSearch Version**
2.12.0

**Dashboards Version**
2.12.0

 - Windows 11
 - Chrome 122.0.6261.129

**Additional context**

We have a fix for this that we are currently testing. The change involves parsing the timeline expression client-side and building the query for each index listed in the expression.

neodescis:
### Description

When ignoreFilterIfFieldNotInIndex is enabled:
1. Parse the Timeline visualization expression for indexes
2. For each index in the expression, build a separate OpenSearch query from the current filters
3. Pass the filters for each index to the Timeline request handler
4. In the request handler, use the index-specific query instead of the single index-agnostic query

### Issues Resolved

closes #6184

## Testing the changes

1. Enable ignoreFilterIfFieldNotInIndex 
2. Create dashboard with visualizations for more than one index
3. Create timeline visualization for one of those indexes and add it to the dashboard
4. Create filter that applies only to the index not referenced by the timeline visualization
5. The timeline visualization should no longer flatline, because the filter is no longer being applied to it

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ananzh:
@neodescis Found an issue when I play with in. So in OSD, we can typically omit the explicit `index=` when specifying the index directly after the .es()/.os()/.opensearch() function. So the regex function `const indexRegEx = /index=['"]?([\sA-z0-9_\-\*]*?)[^\sA-z0-9_\-\*]/g;` will fail to capture the index.

For example, in the following expression, for opensearch_dashboards_sample_data_logs we have `index=` but for opensearch_dashboards_sample_data_ecommerce we omit
```
.es(index=opensearch_dashboards_sample_data_logs,metric=max:bytes).holt(0.8, 0.2, 0.8, 2h).color(#c6c6c6).lines(10).label('Prediction Value'), .es(index=opensearch_dashboards_sample_data_logs,metric=max:bytes).color(red).lines(1).label('Actual Value'), .es(index=opensearch_dashboards_sample_data_logs,metric=max:bytes).lines(1).holt(0.8, 0.2, 0.8, 2h).subtract(.es(index=opensearch_dashboards_sample_data_logs,metric=max:bytes)).abs().if(lt, 500, null, .es(index=opensearch_dashboards_sample_data_logs,metric=max:bytes)).points(5,2,1.5).color(green).label('Anomaly Value').title('max types anomaly'),
.es(opensearch_dashboards_sample_data_ecommerce, metric=max:taxful_total_price).label('Maximum Price'),
.es(opensearch_dashboards_sample_data_ecommerce, metric=avg:taxful_total_price).movingaverage(window=7).label('7-day Avg Price Moving Average').color(green)

```

Screenshot 2024-04-18 at 7 29 20 AM


If we add a filter to `opensearch_dashboards_sample_data_ecommerce`, then in your handler, because the regex function will only capture one index pattern 
Screenshot 2024-04-18 at 7 12 27 AM

and this will cause the `filterByIndex` not apply the filter to the `opensearch_dashboards_sample_data_ecommerce`
Screenshot 2024-04-18 at 7 10 44 AM

ananzh (changes requested):
@neodescis could you check my comment and fix the expression? Feel without `index=` is hard to capture. Maybe we could add `/\.(es|elasticsearch|os|opensearch)\(\s*['"]?([\w\-\*]+)['"]?/` since`.es, .elasticsearch, .os, .opensearch` can help to capture and index name is always the first one. But I am worried about potential edge cases, like could index name not be the first string in the expression or if there is any special characters in index names.

neodescis:
Thanks for looking at this, @ananzh. I appreciate the analysis. I didn't know that `index=` was optional. I understand the issue, and I will see what I can do.

neodescis:
> @neodescis Found an issue when I play with in. So in OSD, we can typically omit the explicit `index=` when specifying the index directly after the .es()/.os()/.opensearch() function. So the regex function `const indexRegEx = /index=['"]?([\sA-z0-9_\-\*]*?)[^\sA-z0-9_\-\*]/g;` will fail to capture the index.
> 
> For example, in the following expression, for opensearch_dashboards_sample_data_logs we have `index=` but for opensearch_dashboards_sample_data_ecommerce we omit
> 
> ```
> .es(index=opensearch_dashboards_sample_data_logs,metric=max:bytes).holt(0.8, 0.2, 0.8, 2h).color(#c6c6c6).lines(10).label('Prediction Value'), .es(index=opensearch_dashboards_sample_data_logs,metric=max:bytes).color(red).lines(1).label('Actual Value'), .es(index=opensearch_dashboards_sample_data_logs,metric=max:bytes).lines(1).holt(0.8, 0.2, 0.8, 2h).subtract(.es(index=opensearch_dashboards_sample_data_logs,metric=max:bytes)).abs().if(lt, 500, null, .es(index=opensearch_dashboards_sample_data_logs,metric=max:bytes)).points(5,2,1.5).color(green).label('Anomaly Value').title('max types anomaly'),
> .es(opensearch_dashboards_sample_data_ecommerce, metric=max:taxful_total_price).label('Maximum Price'),
> .es(opensearch_dashboards_sample_data_ecommerce, metric=avg:taxful_total_price).movingaverage(window=7).label('7-day Avg Price Moving Average').color(green)
> ```
> 
> Screenshot 2024-04-18 at 7 29 20 AM
> If we add a filter to `opensearch_dashboards_sample_data_ecommerce`, then in your handler, because the regex function will only capture one index pattern Screenshot 2024-04-18 at 7 12 27 AM
> 
> and this will cause the `filterByIndex` not apply the filter to the `opensearch_dashboards_sample_data_ecommerce` Screenshot 2024-04-18 at 7 10 44 AM

I've done some digging, and I've determined that the `.os()` function does not function as you stated. When omitting an argument's name, it seems the argument must be at a particular place in the list of arguments. In the case of the index argument, it must be supplied as the 4th argument to `.os()`. The first argument is for "query". I have verified this is how it works, as I get completely different results when using `.os(indexname)` and `.os(index=indexname)`. Furthermore, I found that it is [documented in Kibana](https://www.elastic.co/guide/en/kibana/current/timelion.html#_function_parameters), though I cannot find the same in the documentation for OpenSearch Dashboards.

That said, I do have a solution that supports the unnamed argument as the 4th parameter. I did end up changing the logic a fair amount to do so. I will get it pushed to this pull request soon.

neodescis:
Any chance anyone can look at this? @ananzh ?

angle943:
This is Kiro posting.

This has been awaiting requested changes for 18+ months with no author activity. Recommend closing; please reopen or open a fresh PR if you'd like to continue the work.

neodescis:
I'm afraid that is false. Changes were made after the initial review, but never re-reviewed. We've given up on this getting merged long ago, however!

If you want to close the PR, feel free.

PR #6233, opened 2024-03-21
[Workspace] Add base path when parsing url in http service

### LINKED ISSUE #6015: [Workspace] Make url stateful in url with workspace id

SuZhou-Joe:
### Description

Make basePath required in `getWorkspaceIdFromUrl` method.

### Issues Resolved

fixes #6015

## Testing the changes

- Enable workspace by adding `workspace.enabled` to opensearch_dashboards.yml file 
- Start OSD by using `yarn start` to enable `basePath` feature.
- Navigate to devTools
- Insert test workspaces by calling
```
PUT .kibana/_doc/workspace:foo
{
  "type": "workspace",
  "workspace": {
    "name": "foo"
  }
}
```
- Visit the OpenSearch Dashboards under `foo` workspace: `http://localhost:5603/{random_basePath}/w/foo/app/dev_tools`
- The page should be able to load.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/core/utils/workspace.test.ts
+++ b/src/core/utils/workspace.test.ts
@@ -8,11 +8,11 @@ import { httpServiceMock } from '../public/mocks';
 
 describe('#getWorkspaceIdFromUrl', () => {
   it('return workspace when there is a match', () => {
-    expect(getWorkspaceIdFromUrl('http://localhost/w/foo')).toEqual('foo');
+    expect(getWorkspaceIdFromUrl('http://localhost/w/foo', '')).toEqual('foo');
   });
 
   it('return empty when there is not a match', () => {
-    expect(getWorkspaceIdFromUrl('http://localhost/w2/foo')).toEqual('');
+    expect(getWorkspaceIdFromUrl('http://localhost/w2/foo', '')).toEqual('');
   });
 
   it('return workspace when there is a match with basePath provided', () => {
--- a/src/plugins/workspace/public/plugin.ts
+++ b/src/plugins/workspace/public/plugin.ts
@@ -32,6 +32,7 @@ export class WorkspacePlugin implements Plugin<{}, {}, {}> {
       });
     }
   }
+
   private getWorkspaceIdFromURL(basePath: string): string | null {
     return getWorkspaceIdFromUrl(window.location.href, basePath);
   }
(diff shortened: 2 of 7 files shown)


src/plugins/workspace/public/plugin.ts
```diff
@@ -33,7 +33,7 @@ export class WorkspacePlugin implements Plugin<{}, {}, {}> {
     }
   }
 
-  private getWorkspaceIdFromURL(basePath?: string): string | null {
+  private getWorkspaceIdFromURL(basePath: string): string | null {
```
xinruiba, line 36: Thanks for this change~

Nit:
Since `getWorkspaceIdFromURL` is a private function and have only one line to call another function.

What if we get rid of this `getWorkspaceIdFromURL(basePath: string)` function, and directly call `getWorkspaceIdFromUrl(window.location.href, basePath)` instead?

Thanks

SuZhou-Joe: Sure, will do that.

xinruiba:
LGTM, ship

xinruiba (approved):

src/core/utils/workspace.test.ts
```diff
@@ -8,11 +8,11 @@ import { httpServiceMock } from '../public/mocks';
 
 describe('#getWorkspaceIdFromUrl', () => {
   it('return workspace when there is a match', () => {
-    expect(getWorkspaceIdFromUrl('http://localhost/w/foo')).toEqual('foo');
+    expect(getWorkspaceIdFromUrl('http://localhost/w/foo', '')).toEqual('foo');
```
ruanyl, line 11: Q: why it need to specify `''` as the second parameter explicitly?

ruanyl: Maybe my question is more about why we need to change the `getWorkspaceIdFromUrl` function signature?

SuZhou-Joe: 1. From the implementation perspective, the `basePath` parameter is required when constructing the regular expression to parse the workspace id from url.
2. The bug comes from the loose function signature, I change it to a more strict signature so that others won't step into similar bug.

ruanyl (approved):

Issue #6153, opened 2024-03-15
[Workspace] Filter left nav menu according to the current workspace


ruanyl:
**Is your feature request related to a problem? Please describe.**

With the introduction of workspaces, users can now configure different sets of features/plugins to be displayed within each workspace. However, the current implementation of the left navigation menu in OpenSearch Dashboards presents all available features, which doesn't align with the workspace-specific configurations.

**Describe the solution you'd like**

~~To address this issue, we propose leverage the existing `linkUpdaters$` observable in the `NavLinksService`. It would require a small refactor so that it can support custom link filters.~~

Use `application.registerAppUpdater` to update the apps to `inaccessible` if the workspace doesn't have the apps configured.

**Describe alternatives you've considered**

None

**Additional context**

https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5282
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5092
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4944

PR #6234, opened 2024-03-21
[Workspace] Filter left nav menu items according to the current workspace

ruanyl:
### Description
In this PR, left nav bar will be filtered by the feature configuration of current workspace.

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6153

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/486382/03c7a1c3-6a5e-4557-bd13-ef3d927ecea9

Accessing app which is NOT configured by the workspace will display "Application Not Found" page
image

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/core/public/chrome/nav_links/nav_links_service.ts
```diff
@@ -53,6 +53,13 @@ export interface ChromeNavLinks {
    */
   getNavLinks$(): Observable<Array<Readonly<ChromeNavLink>>>;
 
+  /**
+   * Get an observable for the current link updaters. Link updater is used to modify the
+   * nav links, for example, filter the nav links or update a specific nav link's properties.
+   * {@link LinksUpdater}
+   */
+  getLinkUpdaters$(): BehaviorSubject<LinksUpdater[]>;
```
Hailong-am, line 61: should we update the doc https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/chrome/README.md 
and https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/public.api.md as well?

ruanyl: good point, I will update [README.md](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/chrome/README.md), `public.api.md` is automatically generated.

SuZhou-Joe (approved):

src/plugins/workspace/public/utils.ts
```diff
@@ -0,0 +1,57 @@
...
+export const featureMatchesConfig = (featureConfigs: string[]) => ({
+  id,
+  category,
+}: {
+  id: string;
+  category?: AppCategory;
+}) => {
+  let matched = false;
+
+  for (const featureConfig of featureConfigs) {
```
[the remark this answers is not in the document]
ruanyl: Sure, updated.

seraphjiang:
@xluo-aws  @ruanyl @BionIT @Flyingliuhub  @bandinib-amzn   @kgcreative @xeniatup @zengyan-amazon 

Looks like we are going to have to global contexts: Workspace at left nav, DataSource at top navi if both are enabled

have we discussed or made the decision between Workspace and DataSource?

I assume DataSource always belongs to Workspace. if we all agree about it,  from information hierarchy's perspective, should workspace selector being put on top nav bar. Once cx select workspace, then they could select datasource?

I feel we need to build both experience into future playground to get board feedback asap. 

what do you think?

image

ruanyl:
@seraphjiang 
> have we discussed or made the decision between Workspace and DataSource?
> I feel we need to build both experience into future playground to get board feedback asap.

Agree on that, and yes, the current behavior of setting workspace and data source are more like setting two global contexts. However, base on my understanding so far, they may have correlations. For example, data source available in one workspace may not be available in another workspace.

I think we need to take this into consideration from UX design's perspective as early as possible. @kgcreative @lauralexis @xeniatup

SuZhou-Joe:
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/application/application_service.tsx#L231 
Happen to find a filter by using capability, not sure if we can leverage capability service to filter the navLinks.

ruanyl:
> https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/application/application_service.tsx#L231 Happen to find a filter by using capability, not sure if we can leverage capability service to filter the navLinks.

Good point! Let me try it out

SuZhou-Joe (dismissed):
LGTM

ruanyl:
@SuZhou-Joe I changed the implementation to use `application.registerAppUpdater` to update the app to `inaccessible`, with this change, if accessing the apps which are not configured into a workspace, the "Application Not Found" page will be displayed.

src/plugins/workspace/public/utils.ts
```diff
@@ -55,3 +60,42 @@ export const featureMatchesConfig = (featureConfigs: string[]) => ({
...
+   * When workspace has no features configured, all apps are considered to be accessible
+   */
+  if (!workspace.features) {
+    return true;
+  }
+
+  /**
+   * The app is configured into a workspace, it is accessible after entering the workspace
+   */
+  const featureMatch = featureMatchesConfig(workspace.features);
```
Flyingliuhub, line 78: `featureMatch` or `featureMatcher`, I feel that "matcher" more accurately refers to the mechanism responsible for performing the evaluation. Just my two cents. :)

ruanyl: @Flyingliuhub Updated :)

SuZhou-Joe:
@ruanyl sounds great, thanks for the change!

SuZhou-Joe (dismissed):

SuZhou-Joe:
Some lines are missing unit test coverage, could you please write some to cover that so that all checks can pass for this PR. :)

SuZhou-Joe (approved):

Flyingliuhub (approved):

Hailong-am:
I have concern about mark those links as `inaccessible` for now, before we move index pattern/ saved objects page out, user have no ways to import data and create index pattern. That means user can't start create visualization and dashboards in a fresh workspace

ruanyl:
> I have concern about mark those links as `inaccessible` for now, before we move index pattern/ saved objects page out, user have no ways to import data and create index pattern. That means user can't start create visualization and dashboards in a fresh workspace

You'r right, maybe we should wait for index pattern to be moved out before merging this PR. For saved objects page, you mean saved object management page? Do we have upcoming PR/task about that?

ruanyl:
@xluo-aws In this PR, it will forbid user to access the applications which are not configured in a workspace, even if user access directly via url. I believe this is the desired behavior, but this is different from what we implemented earlier which we just hide the nav links from ui in left nav menu. Do you have any concern with this?

SuZhou-Joe:
Perhaps we can hide it from nav in this PR, and adjust the logic to make these inaccessible in the PR that moves index pattern out of data management.

xluo-aws:
> @xluo-aws In this PR, it will forbid user to access the applications which are not configured in a workspace, even if user access directly via url. I believe this is the desired behavior, but this is different from what we implemented earlier which we just hide the nav links from ui in left nav menu. Do you have any concern with this?

No concern,  it's the right behavior.

ruanyl:
> I have concern about mark those links as `inaccessible` for now, before we move index pattern/ saved objects page out, user have no ways to import data and create index pattern. That means user can't start create visualization and dashboards in a fresh workspace

Since Index pattern will remain in "Dashboards Management" and  "Dashboards Management" is available in workspace, so this is no longer a concern, is that correct?

Hailong-am (approved):

PR #6237, opened 2024-03-21
Make sure customer always have a default datasource

zhyuanqi:
### Description

Make sure customer always have a default datasource

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/53279298/52b94146-5b26-4c74-80d1-5a3357973af6

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


yujin-emma (approved):
LGTM

tianleh:
One question about the flow: after you create the first data source, it goes back to the list page. But why would we require user to refresh page for the default icon to show up?

zhyuanqi:
> One question about the flow: after you create the first data source, it goes back to the list page. But why would we require user to refresh page for the default icon to show up?

Hi Tianle. Thanks for pointing this out. I have fixed the issue. Now it will show the default. You can check it on the video

tianleh:
1. There is one more case to consider: If my OSD is in previous version and has 3 data sources, there will not be default data source after upgrading to the version with your current change. Customers have to either delete or add a data source for your logic to trigger. Is this expected?

2. Customers cannot use UI to update their default data source. The data sources are sorted by title. https://github.com/zhyuanqi/OpenSearch-Dashboards/blob/19d4ba69c0ffde755022693da0d992f4bd8c1a0e/src/plugins/data_source_management/public/components/utils.ts#L38C22-L38C27 If my preferred data source starts with letter `z`, I may have to delete all other data sources to make it default which could be annoying.

 We may need confirmation from product manager about this behavior.

src/plugins/data_source_management/public/components/utils.test.ts
```diff
@@ -274,7 +278,26 @@ describe('DataSourceManagement: Utils.ts', () => {
       expect(getDefaultAuthMethod(authenticationMethodRegistery)?.name).toBe(AuthType.NoAuth);
     });
   });
+  describe('handleSetDefaultDatasourceAfterDeletion', () => {
+    test('should remove defaultDataSource setting and set new defaultDataSource if data sources exist', async () => {
+      mockResponseForSavedObjectsCalls(savedObjects.client, 'find', getDataSourcesResponse);
+      mockUiSettingsCalls(uiSettings, 'get', 'test');
 
+      await handleSetDefaultDatasourceAfterDeletion(savedObjects.client, uiSettings);
+      expect(uiSettings.set).toHaveBeenCalled();
```
ruanyl, line 287: Nit: according to the behavior which set the first data source in the list as default, maybe it's nice to assert that `uiSettings.set` is called with the first data source id. 
```
expect(uiSettings.set).toHaveBeenCalledWith('defaultDataSource', 'test');
```

src/plugins/data_source_management/public/components/utils.test.ts
```diff
@@ -274,7 +278,26 @@ describe('DataSourceManagement: Utils.ts', () => {
...
+    });
+  });
+  describe('handleSetDefaultDatasourceDuringCreation', () => {
+    test('should set defaultDataSource if only one data source exists', async () => {
```
ruanyl, line 291: For unit test of this function, I think it's worth to add a test to cover the case that when number of data source > 1 that `uiSettings.set` should not be called.

src/plugins/data_source_management/public/components/utils.test.ts
zhyuanqi: Would do

src/plugins/data_source_management/public/components/utils.test.ts
zhyuanqi: Added!

zhyuanqi:
> 1. There is one more case to consider: If my OSD is in previous version and has 3 data sources, there will not be default data source after upgrading to the version with your current change. Customers have to either delete or add a data source for your logic to trigger. Is this expected?
> 2. Customers cannot use UI to update their default data source. The data sources are sorted by title. https://github.com/zhyuanqi/OpenSearch-Dashboards/blob/19d4ba69c0ffde755022693da0d992f4bd8c1a0e/src/plugins/data_source_management/public/components/utils.ts#L38C22-L38C27 If my preferred data source starts with letter `z`, I may have to delete all other data sources to make it default which could be annoying.
> 
> We may need confirmation from product manager about this behavior.

Thanks for pointing this out. For first question, after customer enables data_source, then he or she can adding datasource and the first created datasource would be default datasouce. This is the default behavior. I don't see a scenario customer can create the first datasource without setting it be default. I might miss something. Please point to me if I do.

For second one, I wil check with product manager for this.Thank you for pointing this out.

ruanyl:
@zhyuanqi 
One more question:  it seems the current implementation only works when deleting default data source from UI(the data source table), perhaps there are other cases that the default data source get deleted?

@seraphjiang @xluo-aws @kgcreative
And also I'm curious how will the "default data source" behave after introducing workspace? Maybe this is something we need to discuss

seraphjiang:
Workspace is new to everyone. Let's list what are workspace type we plan to support and see what's experience for each case.

It is going to be a big topic.

xluo-aws:
In a multi tenant env, each tenant will have it's own default data source, right?

seraphjiang:
> In a multi tenancy env, each tenant will have it's own default data source, right?

Each tenant can has only one.

Each workspace can has one, but not sure what does it mean to home  workspace concept

xluo-aws:
Home workspace can have it's default data source as well. If a user create a saved object in home workspace,  the default data source will be used unless they choose a different data source, there is no difference than other workspace.

seraphjiang:
> Home workspace can have it's default data source as well. If a user create a saved object in home workspace, the default data source will be used unless they choose a different data source, there is no difference than other workspace.

Isn't home workspace an aggregate concept for all workspaces ?

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -49,6 +54,29 @@ export async function getDataSourcesWithFields(
   return response?.savedObjects;
 }
 
+export async function handleSetDefaultDatasourceAfterDeletion(
+  savedObjects: SavedObjectsClientContract,
```
xinruiba, line 58: nit: 
Based on the type "SavedObjectsClientContract", Looks like here we want to provide a savedObjectClient here.
What if we rename ```savedObjects``` to ```savedObjectsClient``` like we we did here: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6237/files#diff-fde383d0fdaaed3bb243ed36abd631c9c311596b2057f737253fc5ab11c4d5edR45?
What do you think?

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -49,6 +54,29 @@ export async function getDataSourcesWithFields(
...
+  uiSettings: IUiSettingsClient
+) {
+  const listOfDataSources: DataSourceTableItem[] = await getDataSources(savedObjects);
+  if (Array.isArray(listOfDataSources) && listOfDataSources.length === 1) {
+    const datasourceId = listOfDataSources[0].id;
+    await uiSettings.set('defaultDataSource', datasourceId);
+  }
+}
+
```
xinruiba, line 79: Thanks for this change, 

Maybe following scenario is out of scope: 

Suppose we already have dataSource created before before we support default dataSource, this line will failed to set the first dataSource as default: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6237/files#diff-fde383d0fdaaed3bb243ed36abd631c9c311596b2057f737253fc5ab11c4d5edR74

Thus I'm thinking if we can handle that scenario and also keep one function to handle set default dataSource by updating it to:

```suggestion
export async function handleSetDefaultDatasource(
  savedObjects: SavedObjectsClientContract,
  uiSettings: IUiSettingsClient
) {
  const currentDefaultDataSourceId = uiSettings.get('defaultDataSource');
  if (currentDefaultDataSourceId) {
     const targetDataSource = await getDataSourceById(currentDefaultDataSourceId, savedObjectsClient);
     if (targetDataSource.id === currentDefaultDataSourceId) {
        return;
     }
  }
  setFirstDataSourceAsDefault(savedObjectsClient, uiSettings)
}

async function setFirstDataSourceAsDefault(  
  savedObjectsClient: SavedObjectsClientContract,
  uiSettings: IUiSettingsClient
) {
  uiSettings.remove('defaultDataSource');
  const listOfDataSources: DataSourceTableItem[] = await getDataSources(savedObjects);
  if (Array.isArray(listOfDataSources) && listOfDataSources.length >= 1) {
    const datasourceId = listOfDataSources[0].id;
    await uiSettings.set('defaultDataSource', datasourceId);
  }
}
```

Just suggestions, what do you think?

zhyuanqi: I think this is a better approach. I will have a new CR on default with picker. Do you mind if i make the relative change in the new CR.

src/plugins/data_source_management/public/components/utils.ts
zhyuanqi: Fixed it!

xinruiba (approved):

xinruiba (dismissed):

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
```diff
@@ -245,6 +252,22 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
...
 
+  const setDefaultDataSource = async () => {
+    try {
+      for (const dataSource of selectedDataSources) {
+        if (uiSettings.get('defaultDataSource') === dataSource.id) {
+          await handleSetDefaultDatasourceAfterDeletion(savedObjects.client, uiSettings);
+        }
+      }
+    } catch (e) {
+      setIsLoading(false);
```
BionIT, line 263: what happens if error happens and we set isloading as false?

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
```diff
@@ -245,6 +252,22 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
...
+      handleDisplayToastMessage({
+        id: 'dataSourcesManagement.editDataSource.setDefaultDataSourceFailMsg',
+        defaultMessage: 'Unable to set the Data Source to default. Please try it again.',
```
BionIT, line 266: How does index pattern handle the error? How would user be able to retry if the data source is already deleted?

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
zhyuanqi: It will stop loading and show error as "Unable to set the Data Source to default. Please try it again."

zhyuanqi: Should use setIsDeleting(false), let me rephrase this part

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
zhyuanqi: Let me chanke the default message to "Unable to find a default datasource. Please set a new default datasource."

BionIT (approved):

ZilongX (approved):

Issue #6226, opened 2024-03-20
[Multiple Datasource] Support datasource-specific TLS configuration


cwperks:
**Is your feature request related to a problem? Please describe.**

Related PR that introduces common TLS configuration across _all_ configured datasource: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6171

This feature request is to extend the settings introduced in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6171 and make them datasource-specific. 

The settings to add are:

```
ssl.verificationMode
ssl.certificateAuthorities
```

and have these settings utilized when configuring the opensearch client for that particular datasource. For instance, with configuration per datasource you can configure one datasource with no SSL verification while enforcing SSL verification on another datasource.

**Describe the solution you'd like**

The TLS could either be configuredL

- When adding a datasource through OSD
- Through a yml file that contains all datasources and their configuration

**Describe alternatives you've considered**

- Without datasource-specific TLS configuration, all configured datasources must share the same TLS settings.

PR #6244, opened 2024-03-22
[Backport 2.x] [Multiple Datasources] Add TLS configuration for multiple data sources (#6171)

cwperks:
Manual backport of #6171 to 2.x.

### DIFF
--- a/config/opensearch_dashboards.yml
+++ b/config/opensearch_dashboards.yml
@@ -281,6 +281,14 @@
 #    'ff00::/8',
 # ]
 
+# Optional setting that enables you to specify a path to PEM files for the certificate
+# authority for your connected datasources.
+#data_source.ssl.certificateAuthorities: [ "/path/to/your/CA.pem" ]
+
+# To disregard the validity of SSL certificates for connected data sources, change this setting's value to 'none'.
+# Possible values include full, certificate and none
+#data_source.ssl.verificationMode: full
+
 # Set enabled false to hide authentication method in OpenSearch Dashboards.
 # If this setting is commented then all 3 options will be available in OpenSearch Dashboards.
 # Default value will be considered to True.
--- a/src/plugins/data_source/config.ts
+++ b/src/plugins/data_source/config.ts
@@ -31,6 +31,15 @@ export const configSchema = schema.object({
       defaultValue: new Array(32).fill(0),
     }),
   }),
+  ssl: schema.object({
+    verificationMode: schema.oneOf(
+      [schema.literal('none'), schema.literal('certificate'), schema.literal('full')],
+      { defaultValue: 'full' }
+    ),
+    certificateAuthorities: schema.maybe(
+      schema.oneOf([schema.string(), schema.arrayOf(schema.string(), { minSize: 1 })])
+    ),
+  }),
   clientPool: schema.object({
     size: schema.number({ defaultValue: 5 }),
   }),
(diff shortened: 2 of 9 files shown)


BionIT (approved):

zhongnansu (approved):
lgtm

config/opensearch_dashboards.yml
```diff
@@ -281,6 +281,14 @@
 #    'ff00::/8',
 # ]
 
+# Optional setting that enables you to specify a path to PEM files for the certificate
+# authority for your connected datasources.
+#data_source.ssl.certificateAuthorities: [ "/path/to/your/CA.pem" ]
+
+# To disregard the validity of SSL certificates for connected data sources, change this setting's value to 'none'.
+# Possible values include full, certificate and none
+#data_source.ssl.verificationMode: full
```
AMoo-Miki, line 290: Are these explained in a README doc?

src/plugins/data_source/config.ts
```diff
@@ -31,6 +31,15 @@ export const configSchema = schema.object({
...
     }),
   }),
+  ssl: schema.object({
+    verificationMode: schema.oneOf(
+      [schema.literal('none'), schema.literal('certificate'), schema.literal('full')],
+      { defaultValue: 'full' }
+    ),
+    certificateAuthorities: schema.maybe(
+      schema.oneOf([schema.string(), schema.arrayOf(schema.string(), { minSize: 1 })])
+    ),
```
AMoo-Miki, line 41: The `minSize` unnecessarily complicates this; why should we not consider a blank array as undefined?

config/opensearch_dashboards.yml
cwperks: @AMoo-Miki Where do you think would be best to document this? These settings are already defined for the default opensearch via the settings [`opensearch.ssl.verificationMode` and `opensearch.ssl.certificationAuthorities`](https://opensearch.org/docs/latest/install-and-configure/install-dashboards/tls/) and this PR extends those to datasources. These settings would be commonly used for all configured datasources, although I filed another [issue](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6226) on per datasource configuration.

Here are the [analogous entries](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/config/opensearch_dashboards.yml#L84-L89) for default OpenSearch.

src/plugins/data_source/config.ts
cwperks: This is the same as the client for [default opensearch](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/opensearch/opensearch_config.ts#L103-L105). The ways its implemented in this PR handles empty as you describe so it can be removed but I had this here to keep it consistent with the other place its defined for default opensearch.

cwperks:
@AMoo-Miki Is there anything you'd like me to change in this backport or on main first and then into this backport?

src/plugins/data_source/config.ts
AMoo-Miki: That is great. i think we can progressively enhance these. I am glad to hear that the code can handle that case.

AMoo-Miki:
> @AMoo-Miki Is there anything you'd like me to change in this backport or on main first and then into this backport?

@cwperks let's deal with it in a separate PR and backport.

config/opensearch_dashboards.yml
AMoo-Miki: Adding some more info to `src/plugins/data_source/README` would be great. As a user, I find OSD's documentation very inadequate and this small step would help users find answers sooner.

PR #6248, opened 2024-03-22
Fix for checkForFunctionProperty so that order does not matter

cwperks:
### Description

Fixes an issue with the implementation of a utility function used in vis_type_vega plugin where order matters when determining if an object has a property that is a function. This PR adds test to assert the desired behavior in multiple scenarios. 

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


src/plugins/vis_type_vega/public/data_model/utils.test.js
```diff
@@ -91,4 +91,36 @@ describe('Utils.handleInvalidQuery', () => {
...
+        alert('Hello!');
+      },
+    };
+
+    expect(Utils.checkForFunctionProperty(input1)).toBe(true);
+
+    const input2 = {
+      key2: () => {
+        alert('Hello!');
+      },
```
seraphjiang, line 108: could we use other method other than alert() in unit test. looks weird UT code has test dependencies on UI function.

just we just be?
```suggestion
      key2: () => {},
```

cwperks: Replaced with jest.fn() which is much better. I was copying the same pattern as other tests in this fixture.

src/plugins/vis_type_vega/public/data_model/utils.test.js
```diff
@@ -91,4 +87,30 @@ describe('Utils.handleInvalidQuery', () => {
       }).toThrowError();
     });
   });
+
+  test('should identify object contains function properties', async () => {
```
tianleh, line 91: Good catch! For test completeness, could we have a test case whose result is `false`?

cwperks: Added additional tests to cover negative cases.

tianleh (approved):

cwperks:
@ashwin-pc @joshuarrrr Could I get some reviews on this PR?

SuZhou-Joe (approved):
LGTM.

bandinib-amzn (approved):
LGTM!

src/plugins/vis_type_vega/public/data_model/utils.ts
```diff
@@ -79,14 +79,14 @@ export class Utils {
...
-    });
-    return result;
+    for (const value of Object.values(object)) {
+      if (typeof value === 'function') {
+        return true;
+      } else if (Utils.isObject(value) && Utils.checkForFunctionProperty(value)) {
+        return true;
+      }
+    }
+    return false;
```
seraphjiang, line 89: @cwperks  could we refactory code a little

```

    for (const value of Object.values(object)) {
      if (typeof value === 'function') return true;
      if (Utils.isObject(value) && Utils.checkForFunctionProperty(value)) return true;
    }
    return false;
```

Issue #6281, opened 2024-03-27
[BUG] Add datasource with test connect failed


zhyuanqi:
**Describe the bug**

Connect failed to remote datasource. After change data_source.ssl.verificationMode: none  in config file, connection succeed

**To Reproduce**
1. Go to Datasource
2. Click on Create datasource connection
4. Add a datasource and test connection
5. Datasource connection failed. 
6. In the config/opensearch_dashboards.yml file, after change data_source.ssl.verificationMode: none. Connection would work

**Expected behavior**
Expect test connection to work

**Dashboards Version**
Main branch

**Plugins**

Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

BionIT:
The bug is introduced in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6171/files
The reason is because when passing the sslConfig object, ca field is set to empty array

cwperks:
Looks like the default of empty list should be removed [here](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6171/files#diff-b08115fb7b3fe6a92ddc034122496543e8cdaf025fd8e4c9100a1b5bb8a9709dR59). 

This would match the behavior of the existing [opensearch_client_config](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/opensearch/legacy/opensearch_client_config.ts#L183)

cwperks:
@zhyuanqi @BionIT Thank you for filing this issue. I opened a Draft PR to address the issue here: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6282

If the `datasource.ssl.certificateAuthorities` setting is not present in `opensearch_dashboards.yml` then the sslConfig portion of the client config should contain `undefined` instead of empty array.

kavilla:
@cwperks , looks like this didn't get into 2.13 cut so I have added the label 2.14 and will resolve this issue. If this is incorrect please, update it as needed.

Thank you.

cwperks:
This is good to be closed. The backport to 2.x was merged yesterday: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6296

PR #6282, opened 2024-03-27
Fix sslConfig for multiple datasource to handle when certificateAuthorities is unset

cwperks:
### Description

Fixes the logic in TLS configuration for multiple datasources to match the behavior of [opensearch_client_config.ts](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/opensearch/legacy/opensearch_client_config.ts) which sets the `ca` portion of the ssl config to `undefined` (instead of empty list) when the `datasource.ssl.certificateAuthorities` setting is not set in `opensearch_dashboards.yml`.

### Issues Resolved

- https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6281

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source/server/client/client_config.ts
+++ b/src/plugins/data_source/server/client/client_config.ts
@@ -56,7 +56,7 @@ export function parseClientOptions(
       config.ssl?.certificateAuthorities
     );
 
-    sslConfig.ca = certificateAuthorities || [];
+    sslConfig.ca = certificateAuthorities;
   }
 
   const clientOptions: ClientOptions = {
--- a/src/plugins/data_source/server/legacy/client_config.ts
+++ b/src/plugins/data_source/server/legacy/client_config.ts
@@ -55,7 +55,7 @@ export function parseClientOptions(
       config.ssl?.certificateAuthorities
     );
 
-    sslConfig.ca = certificateAuthorities || [];
+    sslConfig.ca = certificateAuthorities;
   }
 
   const configOptions: ConfigOptions = {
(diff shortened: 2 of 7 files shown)


src/plugins/data_source/server/legacy/client_config.ts
```diff
@@ -55,7 +55,7 @@ export function parseClientOptions(
       config.ssl?.certificateAuthorities
     );
 
-    sslConfig.ca = certificateAuthorities || [];
+    sslConfig.ca = certificateAuthorities;
```
BionIT, line 58: we also need to fix the interface to make ca as optional I think, at line 16

src/plugins/data_source/server/client/client_config.ts
```diff
@@ -56,7 +56,7 @@ export function parseClientOptions(
       config.ssl?.certificateAuthorities
     );
 
-    sslConfig.ca = certificateAuthorities || [];
+    sslConfig.ca = certificateAuthorities;
```
BionIT, line 59: we also need to fix the interface to make ca as optional I think, at line 16

src/plugins/data_source/server/legacy/client_config.ts
cwperks: That should already be covered because DataSourceSSLConfigOptions is a [Partial](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype)

src/plugins/data_source/server/client/client_config.ts
cwperks: That should already be covered because DataSourceSSLConfigOptions is a [Partial](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype)

BionIT (approved):

src/plugins/data_source/server/client/client_config.ts
BionIT: Oh I see, thanks!

ZilongX (approved):

BionIT:
@cwperks Can you fix the test which is failing https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/8459709350/job/23211770139?pr=6282

cwperks:
> @cwperks Can you fix the test which is failing https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/8459709350/job/23211770139?pr=6282

@BionIT Pushed a fix for the tests.

BionIT (approved):

ZilongX (approved):

cwperks:
Creating a manual backport. The backport bot still has not run after the label was added.

Issue #6286, opened 2024-03-28
[Workspace] Allow copy/duplicate saved objects among workspaces


gaobinlong:
**Is your feature request related to a problem? Please describe.**

With workspace feature enabled, in the saved object management page or the visualization page, we should allow users to copy or duplicate the saved objects in one workspace to another, the copied saved objects are new saved objects in the target workspace, not reference.

Here are some mockups about this functionality:
![image](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/15811131/14914638-3c32-432b-9204-e2f2c71e2e71)

![image](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/15811131/9770936c-f375-492f-9527-d7cabc53f3f6)

**Describe the solution you'd like**

1. Firstly, we need to implement a new copy/duplicate API which can be called by the UI.
2. Secondly, add all the UI change of this functionality.

PR #6288, opened 2024-03-28
[Workspace] Add duplicate saved objects API

gaobinlong:
### Description

This PR adds a new server api named `duplicate_saved_objects` which can be used to duplicate saved objects among workspaces, the implementation is simple, we leverage the code of the existing export api and import api to implement this functionality. In addition, same to export api and import api, the duplicate api will also be under the control of the setting `savedObjects.maxImportExportSize`(defaults to 10000) and `savedObjects.maxImportPayloadBytes` (defaults to 26214400).

For this new api, the usage is:
Request:
```
curl -X POST "localhost:5601/api/workspaces/_duplicate_saved_objects" -H 'osd-xsrf: true' -H 'Content-Type: application/json' -d '{"objects":[{"type":"index-pattern","id":"619cc200-ecd0-11ee-95b1-e7363f9e289d"}],"targetWorkspace":"9gt4lB"}'
```
Response:
```
{"successCount":1,"success":true,"successResults":[{"type":"index-pattern","id":"619cc200-ecd0-11ee-95b1-e7363f9e289d","meta":{"title":"test*","icon":"indexPatternApp"},"destinationId":"2ffee5da-55b3-49b4-b9e1-c3af5d1adbd3"}]}
```

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6286

## Screenshot

No UI change in this PR, the UI change will be in the following PR, here is the video about it:

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/15811131/6fc165a9-db20-4097-8115-4d1a9a90fbd4

## Testing the changes

1. Clone the latest code and run yarn osd bootstrap
2. Modify `config/opensearch_dashboards.yml`, add `workspace.enabled: true`
3. Run `yarn start --no-base-path`
4. Create a workspace and create some saved objects like index patterns, visualizations, etc.
5. Call the copy api following the above usage.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


ashwin-pc (commented):
Can we add docs for the new API changes? Als can we document the existing API as a part of this change?

gaobinlong:
> Can we add docs for the new API changes? Als can we document the existing API as a part of this change?

I'm glad to do that, and how could we add documents for the server apis?

ashwin-pc:
Here are all the Readme's in the repo that document various features. https://opensearch-project.github.io/OpenSearch-Dashboards/docs/index.html#/. I think adding it to the main saved object plugin readme should be good enough. https://opensearch-project.github.io/OpenSearch-Dashboards/docs/index.html#/../src/plugins/saved_objects/README

src/core/server/saved_objects/routes/copy.ts
```diff
@@ -0,0 +1,72 @@
...
+import { SavedObjectConfig } from '../saved_objects_config';
+import { exportSavedObjectsToStream } from '../export';
+import { importSavedObjectsFromStream } from '../import';
+
+export const registerCopyRoute = (router: IRouter, config: SavedObjectConfig) => {
+  const { maxImportExportSize } = config;
+
+  router.post(
+    {
+      path: '/_copy',
```
bandinib-amzn, line 17: 1. Is there any plan to make this API generic? For example today we have only workspace. So this copy API will copy saved objects to workspace. But say in future we have entity A, entity B something like that. So my question is based on parameter are we planning to support cop saved object to entity A, copy saved objects to entity B? If not, I would suggest to make it more self explanatory something like `_copyToWorkspace`? 

2. Is there plan to add api to delete saved objects from workspace?

3. Is there impact on saved object client wrapper? As you are using exiting import export, so I am guessing no. Can you confirm that?

4. Also how the permission plays role for this API? Does any user can call this API and copy saved objects to any workspace?
Thanks!

gaobinlong:
> Here are all the Readme's in the repo that document various features. https://opensearch-project.github.io/OpenSearch-Dashboards/docs/index.html#/. I think adding it to the main saved object plugin readme should be good enough. https://opensearch-project.github.io/OpenSearch-Dashboards/docs/index.html#/../src/plugins/saved_objects/README

I've added the documentation for all saved objects API, the format refers to OpenSearch API's document,  please help to take a look, thanks!

src/core/server/saved_objects/routes/copy.ts
gaobinlong: Thanks for your comment!
1. This new API is only used to copy to workspace, so I think it makes sense that we change the name to `_copyToWorkspace`.
2. We'll add a new function `deleteByWorkspace` to the saved objects client, it will only be used in the server side, so no api needs to add.
3. No, for this API, nothing change in the saved object client wrapper.
4. The permission control performs on the `WorkspaceSavedObjectsClientWrapper` which will be added in the following PR, for this API, when calling `savedObjectClient.bulkGet()` and `savedObjectsClient.bulkCreate()`, we'll check if the user has required permissions based on the `permissions` field and `workspaces` field of the saved objects.

config/opensearch_dashboards.yml
wanglam, line 1: It seems this file doesn't need to be changed. Shall we remove the updates of this file?

gaobinlong: Thanks, I've done that.

xluo-aws:
@bandinib-amzn @ashwin-pc , could you help review Binlong's latest change according to your suggestion?

src/core/server/saved_objects/routes/copy.ts
bandinib-amzn: So if I understand correctly this API will copy a saved object to destination workspace, right? And Id and all the fields will remain same of the same object in both workspaces. Is that right? Will it create a new document or update one field (say workspace) for existing saved object?

Can you please attach the video of demonstration?

CHANGELOG.md
```diff
@@ -71,6 +71,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [Workspace] Add create workspace page ([#6179](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6179))
 - [Multiple Datasource] Make sure customer always have a default datasource ([#6237](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6237))
 - [Workspace] Add workspace list page ([#6182](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6182))
+- [Workspace] Add duplicate saved objects API ([#6288](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6288))
```
bandinib-amzn, line 74: This is little bit confusing to me. How about "Add API to copy saved objects between workspaces"? Also why we are using duplicate keyword and not copy? any specific reason?

src/plugins/saved_objects/README.md
```diff
@@ -175,3 +175,656 @@ The migraton version will be saved as a `migrationVersion` attribute in the save
 ```
 
 For a more detailed explanation on the migration, refer to [`saved objects management`](src/core/server/saved_objects/migrations/README.md).
+
+## Server APIs
```
bandinib-amzn, line 179: Thank you for adding documentation for just not only for your changes but also all other APIs. Much appreciated!

src/plugins/saved_objects/README.md
```diff
@@ -175,3 +175,656 @@ The migraton version will be saved as a `migrationVersion` attribute in the save
...
+```
+
+### Duplicate saved objects API
+
+Duplicate saved objects among workspaces.
+
+* Path and HTTP methods
+
+```json
+POST <osd host>:<port>/api/workspaces/_duplicate_saved_objects
```
bandinib-amzn, line 780: This is more related to workspace. Should we add this in workspace docs?

bandinib-amzn:
Can you please rebase your branch and fix codecov/patch CI by adding more UT? Thanks!

src/core/server/saved_objects/routes/copy.ts
gaobinlong: No, we create a new ID for the duplicated saved object in the target workspace,  correspondingly a new document is created in the `.kibana` index. I've made a video about the duplicate UI, please take a look.

CHANGELOG.md
gaobinlong: Because all of the UI mockups use `duplicate`, so I use this wording to make it consistent. I've modified the change log, thank you!

src/plugins/saved_objects/README.md
gaobinlong: Sure, I've moved this API's doc to the workspace plugin's doc, even though currently we don't have a workspace doc, but same to this doc file, I've created a new markdown file in workspace plugin and added the docs of all the server APIs in workspace plugin to the file.

gaobinlong:
> Can you please rebase your branch and fix codecov/patch CI by adding more UT? Thanks!

I've added more unit tests for this change, please help to take a look, thank you!

src/core/server/saved_objects/routes/copy.ts
bandinib-amzn: Thanks for adding video.

src/plugins/workspace/server/integration_tests/duplicate.test.ts
```diff
@@ -0,0 +1,306 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// eslint-disable-next-line @osd/eslint/no-restricted-paths
+import * as exportMock from '../../../../core/server/saved_objects/export';
+import supertest from 'supertest';
+// eslint-disable-next-line @osd/eslint/no-restricted-paths
+import { SavedObjectsErrorHelpers } from '../../../../core/server/saved_objects';
```
bandinib-amzn, line 10: nit: remove no-restricted-paths patch check.

bandinib-amzn, line 10:
```suggestion
// eslint-disable-next-line @osd/eslint/no-restricted-paths
import { SavedObjectsErrorHelpers } from '../../../../core/server';
```

src/plugins/workspace/server/integration_tests/duplicate.test.ts
```diff
@@ -0,0 +1,306 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// eslint-disable-next-line @osd/eslint/no-restricted-paths
```
bandinib-amzn, line 6: nit: remove no-restricted-paths patch check. All methods from '../../../../core/server/saved_objects/export' are vaiable at [server level](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/index.ts)

src/plugins/workspace/server/integration_tests/duplicate.test.ts
```diff
@@ -0,0 +1,306 @@
...
+import * as exportMock from '../../../../core/server/saved_objects/export';
```
bandinib-amzn, line 7:
```suggestion
import * as exportMock from '../../../../core/server';
```

src/plugins/workspace/server/integration_tests/duplicate.test.ts
```diff
@@ -0,0 +1,306 @@
...
+import supertest from 'supertest';
+// eslint-disable-next-line @osd/eslint/no-restricted-paths
+import { SavedObjectsErrorHelpers } from '../../../../core/server/saved_objects';
+import { UnwrapPromise } from '@osd/utility-types';
+import { loggingSystemMock, savedObjectsClientMock } from '../../../../core/server/mocks';
+import { setupServer } from '../../../../core/server/test_utils';
+// eslint-disable-next-line @osd/eslint/no-restricted-paths
+import { workspaceClientMock } from '../../public/workspace_client.mock';
```
bandinib-amzn, line 15: nit: no-restricted-paths

use a common directory. move workspaceClientMock to common directory and update all the import statement

src/plugins/workspace/server/integration_tests/duplicate.test.ts
```diff
@@ -0,0 +1,306 @@
...
+import { registerDuplicateRoute } from '../routes/duplicate';
+import { createListStream } from '../../../../core/server/utils/streams';
+// eslint-disable-next-line @osd/eslint/no-restricted-paths
+import { mockUuidv4 } from '../../../../core/server/saved_objects/import/__mocks__';
+// eslint-disable-next-line @osd/eslint/no-restricted-paths
+import { createExportableType } from '../../../../core/server/saved_objects/routes/test_utils';
```
bandinib-amzn, line 21: not sure if we can export this at server level. Can you try and see if possible to remove `no-restricted-paths`

bandinib-amzn:
Hi @gaobinlong,

I am taking one step back and thinking logic again after watching video and going over PR again - Why we need to introduce new API? You already retrieve and load data on page. You just need to create new saved object with new workspace. Why you are exporting and importing again?

SuZhou-Joe (dismissed):

gaobinlong:
> Hi @gaobinlong,
> 
> I am taking one step back and thinking logic again after watching video and going over PR again - Why we need to introduce new API? You already retrieve and load data on page. You just need to create new saved object with new workspace. Why you are exporting and importing again?
> 
> And if we still want to use export import workflow, instead of adding new api, we can call directly existing export and import api, right? With new approach I observed you need to add validation logic which we already have in export api.

Thanks @bandinib-amzn , I think all of your comments above have been addressed yet, please take a look at the latest change, thank you!

For why we add this new API, firstly, the loaded data on the page are not complete, we need to fetch all the details of the references before creating them in the target workspace, secondly, instead of call export API and then import API, this new API makes it easier for users to duplicate saved objects when they just call the server APIs to complete their work, I saw many users asked about how to call the OSD's API in OpenSearch's forum, so I think this new API may reduce the complexity when they need.

SuZhou-Joe (approved):
LGTM

src/plugins/workspace/server/integration_tests/duplicate.test.ts
```diff
@@ -0,0 +1,318 @@
...
+            id: 'my-dashboard',
+          },
+        ],
+        includeReferencesDeep: true,
+        targetWorkspace: 'test_workspace',
+      })
+      .expect(200);
+
+    expect(result.body).toEqual({ success: true, successCount: 0 });
+    expect(savedObjectsClient.bulkCreate).not.toHaveBeenCalled(); // no objects were created
```
bandinib-amzn, line 136: What is the purpose of this test case? 

Is it testing following scenario:
objects in request body not available in system index (.kibana index). That's why exportSavedObjectsToStream returns empty [].

So for such case user will receive 200 success response with `{ success: true, successCount: 0 }` and no saved object client api will get called. Am I right? If yes, can we change test description to make more clear what it is testing? Thanks!

bandinib-amzn: Except above comment, overall looks good to me. Can you please address this comment and rebase your PR?

gaobinlong: Correct, I've modified the test description, thank you!

src/plugins/workspace/README.md
ashwin-pc, line 1: Can you run `yarn docs:generateDevDocs` to add this to our dev docs?

Nit: lets keep PR's limited to one feature and not club multiple things in one.

bandinib-amzn (approved):

bandinib-amzn:
`Build and test / Build and Verify on Linux (ciGroup3)` CI is failing due to #6360 . PR is open https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6411 to fix issue.

src/plugins/workspace/README.md
gaobinlong: Sure, I'll add that in another PR soon.

SuZhou-Joe (approved):

PR #6291, opened 2024-03-29
[BUG] Fix bug in data source aggregated view to change it to depend on displayAllCompatibleDataSources property to show the badge value

BionIT:
### Description
This change fixes bug in data source aggregated view to change it to depend on displayAllCompatibleDataSources property to show the badge value

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/1b1d17b3-ecf6-471a-99f9-e0a9b0c045a0

## Testing the changes
The following was performed in the recording above:
1. enable data source
2. go to maps plugin list page, and should see the data source component mounted correctly and display all data sources
3. go to maps detail page, and should see the badge number which is the number of data source being used in all layers, add a layer, should see the badge number is updated correctly, remove a layer should reduce 1 in the badge number

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
```diff
@@ -111,17 +111,17 @@ export class DataSourceAggregatedView extends React.Component<
     let items = [];
 
     // only display active data sources
```
Flyingliuhub, line 113: do we need to update the comment here?

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
```diff
@@ -111,17 +111,17 @@ export class DataSourceAggregatedView extends React.Component<
...
+      items = [...this.state.allDataSourcesIdToTitleMap.values()].map((title) => {
         return {
-          name: this.state.allDataSourcesIdToTitleMap.get(id),
+          name: title,
           disabled: true,
         };
       });
     } else {
-      items = [...this.state.allDataSourcesIdToTitleMap.values()].map((title) => {
+      items = this.props.activeDataSourceIds!.map((id) => {
```
xinruiba, line 122: Force unwrap looks a little risky here. Are there any possibilities that activeDataSourceIds will be undefined?
Maybe we can update it in following way to avoid force unwrap:
```
const activeDataSoureceIds = this.props.activeDataSoruceIds ?? [];
items = activeDataSoureceIds.map() .....
```
What do you think?

xinruiba (approved):

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
BionIT: Right, sine the two states start to diverge, I think we can make it simpler by having a separate component for display active. The reason I am hesitate to give a default value is the activeDataSourceIds are required when displayAllCompatibleDataSources is false

PR #6303, opened 2024-04-02
[Workspace] Add APIs to support plugin state in request

### LINKED ISSUE #6130: [Workspace] support workspace in saved objects client

We are introducing a new concept `workspace` to manage saved objects, and for now in OpenSearch Dashboard, there are two saved objects clients(browser/public side and server side) that plugins can use to perform CRUD operations on saved objects. And for now there exists two flows to use the clients.

image

And as `workspace` will be used to separate saved objects within specific workspace, CRUD operations on saved objects need to tell the client **which workspace it sits**.

There are cases that plugins use server side saved object client directly, in such cases, it won't know the current workspace if we only support current workspace in public side.

# Describe the solution you'd like

### Support workspace in saved objects client in server side.
We can see that no matter which approach plugins use to manage their objects, eventually it goes to saved objects client in server side as this is the only entry that can send request to .kibana index to fetch the data. So changes are only required in saved objects client in server side.

# Describe alternatives you've considered

### Support workspace in code of plugins and saved objects client on public side 
- Requires requires numerous changes in plugins and not elegant.
Here is a list of places that use directly the saved object client in server side.

    dashboards-maps
    - server/common/stats/stats_helper.ts
    - server/routes/stats_router.ts
    
    dashboards-observability
    - server/adaptors/integrations/integrations_manager.ts
    - server/plugin.ts
    
    
    OpenSearch-Dashboards/src/plugins/vis_type_timeseries
    - server/lib/get_fields.ts
    
    OpenSearch-Dashboards/src/plugins/share
    - server/routes/get.ts
    - server/routes/goto.ts
    
    OpenSearch-Dashboards/src/plugins/telemetry
    - server/fetcher.ts
    
    OpenSearch-Dashboards/src/plugins/vis_augmenter
    - server/routes/stats_helpers.ts
    - server/routes/stats.ts
    
    OpenSearch-Dashboards/src/plugins/data/
    - server/search/search_service.ts
    
    // may not an issue, it read/update with id
    OpenSearch-Dashboards/src/core/server/ui_settings/ui_settings_client.ts
    
    - src/plugins/legacy_export/server/routes/import.ts
    - src/plugins/data_source/server/client/configure_client_utils.ts
    - src/plugins/data/server/index_patterns/index_patterns_service.ts

- Much more maintenance effort in the future.

**Additional context**

#4944

SuZhou-Joe:
### Description

Workspace need to store its workspace id throughout the whole request lifecycle but for now `OpensearchDashboardsRequest` only provides readonly interfaces. In order to add some states into the request without any impact on the response header or the connection between OSD and opensearch cluster, workspace will leverage the `app` interface in the raw request.

#### How will the functions be used in workspace:
- `updateWorkspaceState`: https://github.com/ruanyl/OpenSearch-Dashboards/pull/293/files#diff-5462d700cd760de03fd73de704ceaf51a9f6a519df36429a3b384a21d4f9b68bR56 . workspace will parse the workspaceId from the real request and add the workspace id into workspaceState, which under the hood add `requestWorkspaceId` into rawRequest.app.
- `getWorkspaceState`: https://github.com/ruanyl/OpenSearch-Dashboards/pull/293/files#diff-4ca4e52d5f1dbb92bcdb3694cd11e6f902fa3acadda1ab0f2e7045a0cd9e573dR25 In a wrapper of saved objects client, workspace will detect if there is requestWorkspaceId inside workspace state and try to append the workspace into the params accordingly.

### Issues Resolved

partial resolve #6130 

## Testing the changes

This PR only introduces a function to mutate the `request.app.requestWorkspaceId` field and has already been covered by unit test. Will have a following PR to consume such function and we can test in that PR.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


src/core/server/utils/workspace.ts
ruanyl, line 1: I may suggest to implement it in this way:
1. add `workspaceId` to `request.app` by adding it to `setupRequestStateAssignment` in `http_server.ts`
```diff
   private setupRequestStateAssignment(config: HttpConfig) {
     this.server!.ext('onRequest', (request, responseToolkit) => {
+      const workspaceId = getWorkspaceIdFromUrl(request.url.toString(), '');
       request.app = {
         ...(request.app ?? {}),
         requestId: getRequestId(request, config.requestId),
         requestUuid: uuid.v4(),
+        requestWorkspaceId: workspaceId,
       } as OpenSearchDashboardsRequestState;
       return responseToolkit.continue;
     });
```

2. I may just add `requestWorkspaceId` to `OpenSearchDashboardsRequest` class like this:

```diff
+  public readonly requestWorkspaceId: string;

  constructor(
    request: Request,
    public readonly params: Params,
    public readonly query: Query,
    public readonly body: Body,
    // @ts-expect-error we will use this flag as soon as http request proxy is supported in the core
    // until that time we have to expose all the headers
    private readonly withoutSecretHeaders: boolean
  ) {
+    this.requestWorkspaceId =
+      (request.app as OpenSearchDashboardsRequestState | undefined)?.requestWorkspaceId ?? '';
  }
```

What do you think?

SuZhou-Joe: Yeah, that works. But my concern is that
1. RequestWorkspaceId is an **internal** state that should only be honored by workspace plugin. I do not want other plugins to be aware of the workspace, especially given the fact that workspace is a experimental feature controlled by a feature flag in the near future.
2. The workspace state util will be used in another place: `isDashboardAdmin`, which will be raised later.

And as workspace is a plugin, I do not want an internal state to pollute the OpenSearchDashboardsRequest until it get GA or enabled by default.

src/core/server/utils/workspace.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @returns void
+ */
+export const updateWorkspaceState = (
+  request: OpenSearchDashboardsRequest,
+  payload: Partial<PluginsStates['workspace']>
+) => {
+  const rawRequest = ensureRawRequest(request);
+
+  if (!rawRequest.plugins) {
+    rawRequest.plugins = {};
```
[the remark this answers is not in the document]
SuZhou-Joe: Sure, I just thought the `plugins` here somehow stands for the level of workspace plugin. But as the `rawRequest.plugins` has not been used yet, I am open to change to store it into `rawRequest.app`.

src/core/server/utils/workspace.ts
ruanyl: > 1. RequestWorkspaceId is an internal state that should only be honored by workspace plugin. I do not want other plugins to be aware of the workspace, especially given the fact that workspace is a experimental feature controlled by a feature flag in the near future.

Right, you have a good point, actually I have a bit on that as well, I'm not insist on that :) btw, if this is workspace plugin specific, does it make sense to move it to workspace plugin?

> 2. The workspace state util will be used in another place: isDashboardAdmin, which will be raised later.

You mean `isDashboardAdmin ` will be put in `WorkspaceState`?

SuZhou-Joe: > does it make sense to move it to workspace plugin?
We can not because `ensureRawRequest` is only importable from core folder.

> You mean isDashboardAdmin  will be put in WorkspaceState?
Yes.

ruanyl: > > does it make sense to move it to workspace plugin?
> 
> We can not because `ensureRawRequest` is only importable from core folder.

I see console plugin and data_source plugin import it with disabling the eslint rule, this maybe not a good practice in general. But if this function is required by plugins, maybe we can export it? Pinging @ashwin-pc @AMoo-Miki @kavilla  for more insights :) 

> > You mean isDashboardAdmin  will be put in WorkspaceState?
> 
> Yes.

I tend to separate isDashboardAdmin and all related logic from workspace. Essentially, dashboard admin has nothing to do with workspace(if I'm not mistaken :))

ruanyl:
@SuZhou-Joe would you mind to update the PR description a bit to include how will `updateWorkspaceState` and `getWorkspaceState` be used? Just for more context for the other reviewers, I believe the workspace saved object wrapper will use it as we discussed offline.

src/core/server/utils/workspace.ts
SuZhou-Joe: > I tend to separate isDashboardAdmin and all related logic from workspace.
Agree, but should separate together with permission control. As dashboard admin is a role that can bypass the permission check. But for now, I do not come up with a better place that we can separate the logic to, do you have any suggestion?

SuZhou-Joe:
> update the PR description a bit to include how will `updateWorkspaceState` and `getWorkspaceState` be used

Sure, updated.

src/core/server/utils/workspace.ts
ruanyl: > do you have any suggestion?

I think the consume of this flag `isDashboardAdmin` can be in permission control wrapper(in workspace plugin at the moment). But the setup of this flag shouldn't be in workspace plugin. 

Honestly, I think security dashboards plugin is the right place. But this is not related to this PR, let's move this discussion to the other PR which setup `isDashboardAdmin` flag.

SuZhou-Joe: Sure, let's discuss in another PR.

ruanyl (dismissed):

src/core/server/utils/workspace.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * This function will be used as a proxy
+ * because `ensureRequest` is only importable from core module.
+ *
+ * @param workspaceId string
+ * @returns void
+ */
+export const updateWorkspaceState = (
+  request: OpenSearchDashboardsRequest,
+  payload: Partial<WorkspaceState>
```
BionIT, line 21: looks like workspacestate only has one field which is optional, is it necessary to specify partial?

src/core/server/utils/workspace.ts
```diff
@@ -0,0 +1,40 @@
...
+) => {
+  const rawRequest = ensureRawRequest(request);
+
+  if (!rawRequest.app) {
+    rawRequest.app = {};
```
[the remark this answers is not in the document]
BionIT: I see, this is to use spread later

BionIT (dismissed):

src/core/server/utils/workspace.ts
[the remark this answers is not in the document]
SuZhou-Joe: Yes, in case the request.app is undefined, but it can be safely removed, updated that.

src/core/server/utils/workspace.ts
SuZhou-Joe: I think yes because actually the function just spread all the property into the request.app, which means user can pass in partial of the `WorkspaceState` to update the state.

BionIT (approved):

ZilongX:
LGTM, looking forward to the actual consumption of the updated API

ZilongX (approved):

Issue #6336, opened 2024-04-03
[BUG] Hide local cluster should be set using yaml config


BionIT:
**Describe the bug**
In https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5754, we introduced a new yaml config to allow hide local cluster from the data source component. This property is needed for the component to decide if local cluster should be shown as an option. 

Note that this property should be set from the yaml config, note that we can get the prop from the dataSource object https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source/public/types.ts#L8 and should set it in https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source_management/public/plugin.ts#L105

**Additional context**

zhongnansu:
Hi Lu @BionIT,  I'd like to work on this issue

BionIT:
Awesome, thanks @zhongnansu !

PR #6361, opened 2024-04-07
[MD]Read hideLocalCluster setting from yml and set in data source selector and data source menu

zhongnansu:
### Description

- Read `data_source.hideLocalCluster` setting from `opensearch_dashboards.yml`, and always use this value to decide if to hide local cluster in `DataSourceSelector` and `DataSourceMenu`. This is expected behaviour because hide/show local cluster in data source visual components should be consistent across anywhere that consumes those components.
- make `hideLocalHost` an optional property of `DataSourceSelector` and `DataSourceMenu`
- Update related unit tests

### Issues Resolved

#6336 

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/32652829/2c224539-8bfa-44f1-b165-c96c999275e9

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source_management/public/components/data_source_menu/create_data_source_menu.tsx
+++ b/src/plugins/data_source_management/public/components/data_source_menu/create_data_source_menu.tsx
@@ -13,10 +13,10 @@ import { MountPointPortal } from '../../../../opensearch_dashboards_react/public
 
 export function createDataSourceMenu<T>(
   uiSettings: IUiSettingsClient,
-  dataSource: DataSourcePluginSetup
+  dataSourcePluginSetup: DataSourcePluginSetup
 ) {
   return (props: DataSourceMenuProps<T>) => {
-    const { hideLocalCluster } = dataSource;
+    const { hideLocalCluster } = dataSourcePluginSetup;
     if (props.setMenuMountPoint) {
       return (
         <MountPointPortal setMountPoint={props.setMenuMountPoint}>
--- a/src/plugins/data_source_management/public/components/data_source_menu/create_data_source_menu.tsx
+++ b/src/plugins/data_source_management/public/components/data_source_menu/create_data_source_menu.tsx
@@ -6,21 +6,32 @@
 import React from 'react';
 import { EuiHeaderLinks } from '@elastic/eui';
 import { IUiSettingsClient } from 'src/core/public';
+import { DataSourcePluginSetup } from 'src/plugins/data_source/public';
 import { DataSourceMenu } from './data_source_menu';
 import { DataSourceMenuProps } from './types';
 import { MountPointPortal } from '../../../../opensearch_dashboards_react/public';
 
-export function createDataSourceMenu<T>(uiSettings: IUiSettingsClient) {
+export function createDataSourceMenu<T>(
+  uiSettings: IUiSettingsClient,
+  dataSource: DataSourcePluginSetup
+) {
   return (props: DataSourceMenuProps<T>) => {
+    const { hideLocalCluster } = dataSource;
     if (props.setMenuMountPoint) {
       return (
         <MountPointPortal setMountPoint={props.setMenuMountPoint}>
           <EuiHeaderLinks data-test-subj="top-nav" gutterSize="xs">
-            <DataSourceMenu {...props} uiSettings={uiSettings} />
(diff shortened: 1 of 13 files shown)


src/plugins/data_source_management/public/components/data_source_menu/create_data_source_menu.tsx
```diff
@@ -6,21 +6,30 @@
...
+              {...props}
+              uiSettings={uiSettings}
+              hideLocalCluster={hideLocalCluster}
+            />
           </EuiHeaderLinks>
         </MountPointPortal>
       );
     }
-    return <DataSourceMenu {...props} />;
+    return <DataSourceMenu {...props} hideLocalCluster={hideLocalCluster} />;
```
BionIT, line 33: Looks like we missed adding the uiSettings option here in https://github.com/opensearch-project/OpenSearch-Dashboards/commit/726fb0e34704dbd54c71f9e716654323d67a302e, could you help to add uiSettings here, to test, please use the example plugin for top nav menu https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/examples/multiple_data_source_examples/public/components/data_source_via_top_nav_menu.tsx

BionIT (commented):
Added comments, in the recording, we can also show the local cluster should not be an option any more in the drop down when hidelocalcluster prop is set true

src/plugins/data_source_management/public/components/data_source_menu/create_data_source_menu.tsx
zhongnansu: Added. However when I tried to test it by using `DataSourceSelectable` for top nav menu, it doesn't show "default" label for any options, Here's the recording.

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/32652829/4756cfa1-6836-46cb-9542-619355e88462


I think the issue is with that in top_nav_menu.tsx,  `DataSourceMenu` is not obtained by calling the `dataSourceManagement.ui.getDataSourceMenu<DataSourceSelectableConfig>();`, but through a direct import as below code suggests. 
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a0eaf8483cbc29daf95d559c4b98965ad168052d/src/plugins/navigation/public/top_nav_menu/top_nav_menu.tsx#L44

Therefore, even I added the `uiSetting` to `createDataSourceMenu()`, it makes no difference for top nav menu use case, because `ui.getDataSourceMenu` API is never called, and `uiSetting` is never being passed through.
@BionIT What do you think? I can file an issue if it makes sense

zhongnansu:
> Added comments, in the recording, we can also show the local cluster should not be an option any more in the drop down when hidelocalcluster prop is set true

sure, updated the recording

src/plugins/data_source_management/public/components/data_source_menu/create_data_source_menu.tsx
BionIT: Right, yeah, could you help to cut an issue for that?

BionIT (dismissed):

yujin-emma (approved):
LGTM

src/plugins/data_source_management/public/components/data_source_selector/create_data_source_selector.tsx
```diff
@@ -5,10 +5,15 @@
 
 import React from 'react';
 import { IUiSettingsClient } from 'src/core/public';
+import { DataSourcePluginSetup } from 'src/plugins/data_source/public';
 import { DataSourceSelector, DataSourceSelectorProps } from './data_source_selector';
 
-export function createDataSourceSelector(uiSettings: IUiSettingsClient) {
+export function createDataSourceSelector(
+  uiSettings: IUiSettingsClient,
+  dataSource: DataSourcePluginSetup
```
xinruiba, line 13:
```suggestion
  dataSourcePluginSetup: DataSourcePluginSetup
```

zhongnansu: @xinruiba Fixed the naming, could you take another look?

xinruiba (approved):

BionIT (approved):

PR #6366, opened 2024-04-08
[OE] Support compression for requests

### LINKED ISSUE #5296: [Resource Management] Compress traffic

OpenSearch Dashboards Server supports compressed traffic. OpenSearch Core APIs support compression. OpenSearch JS Client supports the compression header but OpenSearch Dashboards does not currently hook into the available compression algorithm.

Hook into the the clients compression algorithm (currently only gzip) and allow the feature to be configurable at an cluster admin level (`opensearch_dashboards.yml` config). Ideally the benefit would help request-response time overall with the compression on the client side. It should improve resource management if too much physical limitations of machines prevent successful response (one example being test runners).

Requirements:
* Benchmark before enabling compression
* Implementation
* Benchmark after enabling compression
* Update relevant workflows and CI 

n/a

**Additional context**

Compression on sample data indices can be improvement. Better documentation on node API compression support (saved objects supports compression).

Hook into existing benchmark project can assist with this.

Audit requests and response are correctly compressed and decompressed. For example, concurrent search feature tends to not support the total max hits header

Example of implementation the solution above: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5223/files

kavilla:
### Description

Support compressing traffic with `opensearch.compression: true` or `opensearch.compress: 'gzip'`.

Also, set compression in the Node server and OpenSearch traffic for CI.

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5296

## Changelog
- feat: Support compressing traffic with `opensearch.compression: true` or `opensearch.compress: 'gzip'`

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/server/opensearch/client/client_config.ts
+++ b/src/core/server/opensearch/client/client_config.ts
@@ -53,6 +53,7 @@ export type OpenSearchClientConfig = Pick<
   | 'username'
   | 'password'
   | 'disablePrototypePoisoningProtection'
+  | 'compression'
 > & {
   memoryCircuitBreaker?:
     | OpenSearchConfig['memoryCircuitBreaker']
@@ -118,6 +119,10 @@ export function parseClientOptions(config: OpenSearchClientConfig, scoped: boole
 
   if (config.disablePrototypePoisoningProtection != null) {
     clientOptions.disablePrototypePoisoningProtection = config.disablePrototypePoisoningProtection;
+  }
+
+  if (config.compression) {
+    clientOptions.compression = 'gzip';
   }
 
   return clientOptions;
--- a/src/core/server/opensearch/client/client_config.ts
+++ b/src/core/server/opensearch/client/client_config.ts
@@ -122,7 +122,7 @@ export function parseClientOptions(config: OpenSearchClientConfig, scoped: boole
   }
 
   if (config.compression) {
-    clientOptions.compression = 'gzip';
+    clientOptions.compression = config.compression;
   }
 
   return clientOptions;
--- a/src/core/server/opensearch/opensearch_config.ts
+++ b/src/core/server/opensearch/opensearch_config.ts
@@ -32,12 +32,14 @@ import { schema, TypeOf } from '@osd/config-schema';
 import { Duration } from 'moment';
 import { readFileSync } from 'fs';
 import { ConfigDeprecationProvider } from 'src/core/server';
+import { ClientOptions } from '@opensearch-project/opensearch';
 import { readPkcs12Keystore, readPkcs12Truststore } from '../utils';
 import { ServiceConfigDescriptor } from '../internal_types';
 
 const hostURISchema = schema.uri({ scheme: ['http', 'https'] });
 
 export const DEFAULT_API_VERSION = '7.x';
+export const DEFAULT_COMPRESSION = 'gzip';
 
 export type OpenSearchConfigType = TypeOf<typeof configSchema>;
 type SslConfigSchema = OpenSearchConfigType['ssl'];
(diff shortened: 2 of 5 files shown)


src/core/server/opensearch/opensearch_config.ts
```diff
@@ -131,6 +131,7 @@ export const configSchema = schema.object({
   healthCheck: schema.object({ delay: schema.duration({ defaultValue: 2500 }) }),
   ignoreVersionMismatch: schema.boolean({ defaultValue: false }),
   disablePrototypePoisoningProtection: schema.maybe(schema.boolean({ defaultValue: false })),
+  compression: schema.maybe(schema.boolean({ defaultValue: false })),
```
ashwin-pc, line 134: Why do we default this to false? Why would i not want this by default?

ashwin-pc (dismissed):
LGTM

src/core/server/opensearch/opensearch_config.ts
kavilla: There could be a little bit of overhead. Without much insights on performance, likely this is more safe set to false.

SuZhou-Joe, line 134: I am thinking why don't we make it a string? Though `gzip` is the most common method to compress request.

AMoo-Miki, line 134:
```suggestion
  compression: schema.boolean({ defaultValue: false }),
```
When we have a default value, `maybe` is not needed. However, I like the idea @SuZhou-Joe proposed. I would prefer for this to be `string | boolean`. Then we validate that string is a supported compression.

I understand that probably `gzip` is the only option right now but it would be nice to have this be future proof.

AMoo-Miki (dismissed):

ashwin-pc (dismissed):

src/core/server/opensearch/opensearch_config.ts
seraphjiang: +1 to @kavilla we should be more conservative before we have performance and load test as part of CI

kavilla: this is actually due to the OpenSearch JS client:
```
https://github.com/opensearch-project/opensearch-js/blob/main/lib/Transport.d.ts#L73
```

kavilla: updated!

kavilla: updated!

src/core/server/opensearch/opensearch_config.ts
```diff
@@ -334,6 +345,12 @@ export class OpenSearchConfig {
     this.password = rawConfig.password;
     this.customHeaders = rawConfig.customHeaders;
     this.disablePrototypePoisoningProtection = rawConfig.disablePrototypePoisoningProtection;
+    this.compression =
+      typeof rawConfig.compression === 'boolean' && rawConfig.compression
+        ? DEFAULT_COMPRESSION
+        : typeof rawConfig.compression === 'string'
+        ? rawConfig.compression
+        : undefined;
```
AMoo-Miki, line 353: An expression of `rawConfig.compression = false` gets out of this as `undefined` which is the wrong message out of this function (because it was defined). For the sanity of future devs, we should:
```suggestion
        : false;
```

BionIT:
@AMoo-Miki @ashwin-pc could you help take another look at this PR which is targeting 2.15?

AMoo-Miki:
@ashwin-pc, I think Rocky is tired up. If you agree with my suggestion, we can move this along.

src/core/server/opensearch/client/client_config.ts
```diff
@@ -120,6 +121,10 @@ export function parseClientOptions(config: OpenSearchClientConfig, scoped: boole
     clientOptions.disablePrototypePoisoningProtection = config.disablePrototypePoisoningProtection;
   }
 
+  if (config.compression) {
+    clientOptions.compression = config.compression;
```
virajsanghvi, line 125: do we want to add a test for this given we updated the test to not do this by default?

src/core/server/opensearch/opensearch_config.ts
virajsanghvi: Also, for the sanity of future devs, can we not use nested ternaries, especially without parens? calling a function with readable if statements isn't the worst thing imo :)

ananzh:
move to 2.16

angle943:
This is Kiro posting.

This PR has merge conflicts. The last substantive author activity was some time ago. @kavilla — is this still active? If yes, please rebase against main and request re-review. If not, please close.

PR #6421, opened 2024-04-12
fix bug: management section id not match the key defined in capabilities

ruanyl:
### Description

The `Dashboards management` section id(`opensearch-dashboards`) of `Management` category doesn't match the id(`opensearchDashboards`) defined in `capabilities`, the consequence is the following code will never work for apps of `Dashboards management`. Even set the capability of the app to false, the nav link still displayed on the UI(it does redirect to home page when accessing the link).

```typescript
    this.getAllSections().forEach((section) => {
      if (capabilities.management.hasOwnProperty(section.id)) {
        const sectionCapabilities = capabilities.management[section.id];
        section.apps.forEach((app) => {
          if (sectionCapabilities.hasOwnProperty(app.id) && sectionCapabilities[app.id] !== true) {
            app.disable();
          }
        });
      }
    });
```

The original PR introduced the `id` change is https://github.com/opensearch-project/OpenSearch-Dashboards/pull/260
@kavilla do you have more context of changing id from `opensearchDashboards` to `opensearch-dashboards`?

## Changelog

- fix: management section id not match the key defined in capabilities

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/management/public/management_sections_service.test.ts
+++ b/src/plugins/management/public/management_sections_service.test.ts
@@ -28,6 +28,7 @@
  * under the License.
  */
 
+import { DEFAULT_MANAGEMENT_CAPABILITIES } from '../common/contants';
 import {
   ManagementSectionsService,
   getSectionsServiceStartPrivate,
@@ -104,5 +105,30 @@ describe('ManagementService', () => {
         "test-app-3",
       ]
     `);
+  });
+
+  it('should disable apps register in opensearchDashboards section', () => {
+    const originalDataSourcesCapability =
+      DEFAULT_MANAGEMENT_CAPABILITIES.management.opensearchDashboards.dataSources;
+
+    const setup = managementService.setup();
+
+    // Register app with id dataSources, the app id will be capability id
+    setup.section.opensearchDashboards.registerApp({
+      id: 'dataSources',
+      title: 'Data source',
+      mount: jest.fn(),
+    });
+
+    // Now set dataSources to capability to false should disable
+    // the dataSources app registered in opensearchDashboards section
+    DEFAULT_MANAGEMENT_CAPABILITIES.management.opensearchDashboards.dataSources = false;
+
+    managementService.start({ capabilities: DEFAULT_MANAGEMENT_CAPABILITIES });
+    expect(
+      setup.section.opensearchDashboards.apps.find((app) => app.id === 'dataSources')?.enabled
+    ).toBe(false);
+
+    DEFAULT_MANAGEMENT_CAPABILITIES.management.opensearchDashboards.dataSources = originalDataSourcesCapability;
   });
 });
--- a/src/plugins/management/public/management_sections_service.test.ts
+++ b/src/plugins/management/public/management_sections_service.test.ts
@@ -107,13 +107,15 @@ describe('ManagementService', () => {
     `);
   });
 
-  it('should disable apps register in opensearchDashboards section', () => {
+  it('should disable apps register in predefined opensearchDashboards section', () => {
+    // The management capabilities has `opensearchDashboards` as the key
     const originalDataSourcesCapability =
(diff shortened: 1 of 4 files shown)


src/plugins/management/public/management_sections_service.ts
```diff
@@ -94,8 +98,9 @@ export class ManagementSectionsService {
 
   start({ capabilities }: SectionsServiceStartDeps) {
     this.getAllSections().forEach((section) => {
-      if (capabilities.management.hasOwnProperty(section.id)) {
-        const sectionCapabilities = capabilities.management[section.id];
+      const capabilityId = MANAGEMENT_ID_TO_CAPABILITIES[section.id] || section.id;
```
manasvinibs, line 101: @ruanyl Just curious to know why don't we instead update the Id in the Capabilities to `opensearch-dashboards` instead of mapping it here?

ruanyl: @manasvinibs Good question, that's for compatibility, `opensearch-dashboards` maybe deep linked, for example, in function test repo.

virajsanghvi, line 101: Is there an associated test to validate this is working correctly?

manasvinibs: +1

BionIT:
Hi @ruanyl , are we targeting this PR for 2.15 or 2.16?

ruanyl:
> Hi @ruanyl , are we targeting this PR for 2.15 or 2.16?

Maybe 2.16, haven't got a chance to add unit tests

src/plugins/management/public/management_sections_service.ts
ruanyl: @kavilla Is there a way to estimate the impact of changing capabilities key from `opensearchDashboards` to `opensearch-dashboards` in `src/plugins/management/server/capabilities_provider.ts`?

ruanyl: I've seen some places in OSD core uses `capabilities.management.opensearchDashboards`, also there might other plugins are already using this capability key, I think changing this key would be a breaking change, so I think it would be better to do key mapping. What do you think? @manasvinibs

ruanyl: @virajsanghvi unit test added

ruanyl:
@virajsanghvi @manasvinibs Could you please take another look at the PR?

src/plugins/management/public/management_sections_service.test.ts
```diff
@@ -105,4 +106,29 @@ describe('ManagementService', () => {
       ]
     `);
   });
+
+  it('should disable apps register in opensearchDashboards section', () => {
+    const originalDataSourcesCapability =
+      DEFAULT_MANAGEMENT_CAPABILITIES.management.opensearchDashboards.dataSources;
```
virajsanghvi, line 112: nit: ideally we could mock this over setting it globally, but you do clean it up so not blocking.

src/plugins/management/public/management_sections_service.test.ts
```diff
@@ -105,4 +106,29 @@ describe('ManagementService', () => {
       ]
     `);
   });
+
+  it('should disable apps register in opensearchDashboards section', () => {
```
virajsanghvi, line 110: Could you help explain how this is testing the "opensearch-dashboards" translation code?

ruanyl: In the capability provider, the capability key is `opensearchDashboards`, when registering the default section `opensearchDashboards: this.registerSection(OpenSearchDashboardsSection)`, the section id is defined as `opensearch-dashboards` and this id was used to retrieve the app status register with capability key `opensearchDashboards`.

So we need to do the mapping from `opensearch-dashboards` -> `opensearchDashboards`

virajsanghvi (approved):

Issue #6242, opened 2024-03-22
Suggestion to update thresholds for first-time-contributor,repeat-contributor


seraphjiang:
**Is your feature request related to a problem? Please describe.**

the question related to thresholds of first-time-contributor,repeat-contributor

### current setup
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/0dce00a806bf8815ca84802541010767b6b3a8bd/.github/workflows/github-workflow-badger.yml#L19

```
          badges: '[first-time-contributor,repeat-contributor,valued-contributor,seasoned-contributor,all-star-contributor,distinguished-contributor]'
          thresholds: '[0,3,6,13,25,50]'
```

### recommended setup
to be accurate, should we change thresholds of repeat-contributor to 1? currently people contributed 1~2 labeled as first-time contributor

some references

https://github.com/cdklabs/github-merit-badger

```
steps:
  - uses: aws-github-ops/github-merit-badger@main
    id: github-merit-badger
    with:
      github-token: ${{ secrets.GITHUB_TOKEN }}
      badges: '[first-time-contributor,repeat-contributor,allstar-contributor]'
      thresholds: '[0,1,5]'
      days: 365
```

```
steps: 
  - uses: aws-github-ops/github-merit-badger@main
    id: github-merit-badger
    with:
      github-token: ${{ secrets.GITHUB_TOKEN }}
      badges: '[first-time-contributor,repeat-contributor,allstar-contributor]'
      thresholds: '[0,1,5]'
      title-prefixes: '[feat,fix]'
```

seraphjiang:
@bbarani @prudhvigodithi
I realize the original idea was from https://github.com/opensearch-project/opensearch-build/pull/3305#issuecomment-1468910113

Are there any concern to make threshold of repeat-contributor to 1, to be accuracy and consistent with recommended value from its original repo?

cc: @opensearch-project/admin  could we move this issue to root repo since result will impact across 4 repos

https://github.com/search?q=org%3Aopensearch-project%20github-merit-badger&type=code

prudhvigodithi:
Hey @seraphjiang the workflow exists here https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/.github/workflows/github-workflow-badger.yml#L19, its upto the the OpenSearch-Dashboards repo maintainers to adjust the thresholds, each repo can maintain its own thresholds, as of now AFAIK there are no fixed thresholds.
Thank you

[...]

PR #6441, opened 2024-04-13
fix thresholds for repeat-contributor to avoid logical confusion on f…

seraphjiang:
### Description
current threshold setup for repeat-contributor cause mis-leading to first-time-contributor. even contributor made PR more than once, it still labelled as first-time-contributor

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6242

### Check List


xinruiba (approved):

zhongnansu (approved):

zhongnansu:
I don't think we need to backport this change to 2.x branch. This file only exists in main branch. Original commit to create this fie is not being backported.

Issue #1040, opened 2021-12-22
[BUG] Telemetry docs migration issue


kavilla:
**Describe the bug**

There is a problem with indices that came from a legacy version with telemetry added and migrating to OpenSearch Dashboards that prevents the clean up of the previous system index and prevents it from starting automatically.

**To Reproduce**
1. Have a doc like the following:
```
 {
    "_index" : ".kibana_1",
    "_type" : "_doc",
    "_id" : "ui-metric:console:GET_cluster.state",
    "_score" : 1.0,
    "_source" : {
      "ui-metric" : {
        "count" : 1
      },
      "type" : "ui-metric",
      "references" : [ ],
      "updated_at" : "2021-03-02T14:08:50.823Z"
    }
  },
```
2. Start OpenSearch Dashboards
3. See if failure in migration
4. Check `_cat/aliases` to see if the same alias for a system index points to two system indices.

**Expected behavior**
No exception thrown

**OpenSearch Version**
1.0+

**Dashboards Version**
1.0+

**Plugins**
Security

**Screenshots**
n/a

**Additional context**
A temporary workaround can be to manually cleanup the previous alias with 
`curl -X DELETE localhost:9200/.kibana_1/_alias/.kibana` if you have a `.kibana_2`. To prevent future issues until this is patched you can delete those telemetry docs. Using the app Dev Tools with `DELETE /.kibana/_doc/{ID of TELEMETRY OBJECT`.  [[REF]](https://opensearch.org/docs/latest/opensearch/rest-api/document-apis/delete-document/)

AMoo-Miki:
A similar problem was also observed with `Detected mapping change in "properties.originId"` on OSD 1.3.2.

bandinib-amzn:
Is there any plan to fix this?

bandinib-amzn:
Is this issue only present in 1.x version? What about 2.x?

wbeckler:
I propose that we skip this type during the migration. Thoughts? I know this will lead to data loss, but it's better than completely dying.

wbeckler:
Also, what about a startup flag that includes "ignore this thing" which can be enabled optionally, so that some users can have a hard fail that's more transparent?

kavilla:
Looking into there is a `migrations.skip` that can be configured. However, the logic in how it's being used seems to still then migrate if plugins aren't not initialized at the time of startup. Which seems to be always, then in goes into the plugin migration part. Still looking into.

PR #6443, opened 2024-04-13
[saved objects] enable deletion of saved objects by type if configured

kavilla:
### Description

Adds the following settings:
```
migrations.delete.enabled
migrations.delete.types
```

`unknown` types already exist but the purpose of this type is for plugins that are disabled. OpenSearch Dashboards gets confused when a plugin is not defining a saved object type but the saved object exists. This can occur when migrating from a non-OSD version and there exists non-compatiable saved objects.

If OSD is failing to migrate an index because of a document, I can now configure OSD to delete types of saved objects that I specified because I know that these types should not be carried over.

### Issues Resolved

resolves: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/1040

## Changelog
- feat: Adds `migrations.delete` to delete saved objects by type during a migration

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


seraphjiang:
Thanks @kavilla for the improvement 


@BionIT  @CCongWang to take a look as related to migration and upgration scenario.

src/core/server/saved_objects/migrations/core/index_migrator.ts
```diff
@@ -171,6 +172,31 @@ async function deleteIndexTemplates({ client, log, obsoleteIndexTemplatePattern
...
+      query: {
+        bool: {
+          should: [...typesToDelete.map((type) => ({ term: { type } }))],
+        },
+      },
+    },
+    conflicts: 'proceed',
+    refresh: true,
+  });
+}
```
seraphjiang, line 198: could we print out query when log level is debug

src/core/server/saved_objects/migrations/core/migration_context.ts
```diff
@@ -199,6 +208,37 @@ export function disableUnknownTypeMappingFields(
...
+ * - For each type, it deletes the corresponding property from the target mappings object.
+ *
+ * The purpose of this function is to allow for dynamic modification of the target mappings object
+ * based on the application's configuration. This can be useful in scenarios where certain type
+ * mappings are no longer needed and should be removed from the target mappings.
+ *
+ * Note: The function does not return a value. It directly modifies the targetMappings object passed to it.
+ *
+ * @param {Object} targetMappings - The target mappings object to be modified.
+ * @param {Object} opensearchDashboardsRawConfig - The application's configuration object.
```
seraphjiang, line 226: this is object parsed from yaml config file?

src/core/server/saved_objects/migrations/core/migration_context.ts
```diff
@@ -199,6 +208,37 @@ export function disableUnknownTypeMappingFields(
...
+ * The function operates under the following conditions:
+ * - It checks if the 'migrations.delete.enabled' configuration is set to true.
+ * - If true, it retrieves the 'migrations.delete.types' configuration
+ * - For each type, it deletes the corresponding property from the target mappings object.
+ *
+ * The purpose of this function is to allow for dynamic modification of the target mappings object
+ * based on the application's configuration. This can be useful in scenarios where certain type
+ * mappings are no longer needed and should be removed from the target mappings.
+ *
+ * Note: The function does not return a value. It directly modifies the targetMappings object passed to it.
```
seraphjiang, line 223: i really like the clear code and jsdoc for the method. One question about mutable input param, it is generally safer to return a new value from a function instead of modifying its inputs in functional programming, what's the rational behind to modify the targetMappings?

seraphjiang: refer to https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/DEVELOPER_GUIDE.md#avoid-mutability-and-state

bandinib-amzn:
Can we add specific use case or attach a video or screenshot (if possible) in this PR details to make it more clear to the people who does not have context? Details like what problem we are trying to solve, what type of users would receive help from this change etc.  would be helpful. Thanks!

src/core/server/saved_objects/migrations/core/index_migrator.ts
```diff
@@ -171,6 +172,31 @@ async function deleteIndexTemplates({ client, log, obsoleteIndexTemplatePattern
   return Promise.all(templateNames.map((name) => client.indices.deleteTemplate({ name: name! })));
 }
 
+/**
+ * Delete saved objects by type. This is used to remove saved object types that
+ * are not compatible with the current version of OpenSearch Dashboards.
```
bandinib-amzn, line 177: We are not actually adding logic to determine which saved object types are not compatible, right? We are deleting configured `typesToDelete`.

can we change the description to avoid confusion? maybe - `If migrations.delete.types is specified, this will delete any saved object type that match it`

src/core/server/saved_objects/migrations/core/migration_context.ts
kavilla: yes. i would need dig deeper into what they opted to use it this way as they have the config but mimicked what they did before.

src/core/server/saved_objects/migrations/core/migration_context.ts
kavilla: > generally safer to return a new value from a function instead of modifying its inputs in functional programming

yeah. i originally had the spread operation as well and noticed the code became even harder to understand what is happening here.

i can update to mimic the other behavior when marking types unknown

src/core/server/saved_objects/migrations/core/index_migrator.ts
kavilla: will do

config/opensearch_dashboards.yml
wbeckler, line 1: What if, instead of deleting them, we could configure it to move those documents to an index that is subsequently ignored by the dashboard migration process

src/core/server/saved_objects/migrations/core/index_migrator.ts
kavilla: updated

src/core/server/saved_objects/migrations/core/migration_context.ts
kavilla: updated

seraphjiang (approved):

seraphjiang:
@bandinib-amzn @Flyingliuhub @BionIT @xinruiba would you help to take a look and approve if no need. this is need for 2.14

bandinib-amzn:
This change overall looks good to me. Can you add more UT to pass codecov CI? For link check failure, it looks like it is not related to your change, can you confirm?

Flyingliuhub (approved):

Flyingliuhub:
LGTM.

ZilongX (approved):

ZilongX:
re-run some failed checks

ananzh:
ciGroup3 failed.

PR #6459, opened 2024-04-15
[Workspace] fix apps are missing when updating a workspace

ruanyl:
This is caused by #6234 which marked apps as inaccessible when the apps are not configured into the current workspace. However, inaccessible apps are filter out on workspace update page, this is unexpected as such apps can be configured into the current workspace again though they are not configured into the workspace at the moment.

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6481

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


SuZhou-Joe (approved):
No much concern but maybe we should consider using capabilities service to filter out apps in workspace in the future.

src/plugins/workspace/public/components/workspace_form/utils.ts
```diff
@@ -44,18 +44,28 @@ export const getNumberOfErrors = (formErrors: WorkspaceFormErrors) => {
...
 
   // Filter out all hidden applications and management applications and default selected features
-  const visibleApplications = applications.filter(
-    ({ navLinkStatus, chromeless, category, id }) =>
+  const visibleApplications = applications.filter(({ navLinkStatus, chromeless, category, id }) => {
+    /**
+     * Restrict apps are apps that can be configured into a workspace, but restrict to access
+     * because the current workspace didn't have the apps configured, such apps should NOT filter out
+     */
+    if (restrictedApps && restrictedApps.has(id)) {
```
wanglam, line 58: In order to keep the same source in workspace create, update and list page, can we use `restrictedApps` for all customized features for workspaces. We can remove the check below.
```ts
  return (
        navLinkStatus !== AppNavLinkStatus.hidden &&
        !chromeless &&
        !DEFAULT_SELECTED_FEATURES_IDS.includes(id) &&
        category?.id !== DEFAULT_APP_CATEGORIES.management.id
    );
```

SuZhou-Joe (dismissed):
LGTM

ZilongX:
picked up latest main and re-running all checks now

ZilongX (approved):

src/plugins/workspace/public/components/workspace_form/utils.ts
```diff
@@ -65,23 +61,14 @@ export const convertApplicationsToFeaturesOrGroups = (
 ) => {
```
wanglam, line 61: It seems the applications only need id, title and category. Can we remove navLinkStatus and chromeless?

src/plugins/workspace/public/components/workspace_form/workspace_feature_selector.tsx
```diff
@@ -19,21 +19,20 @@ import { PublicAppInfo } from '../../../../../core/public';
 import { isWorkspaceFeatureGroup, convertApplicationsToFeaturesOrGroups } from './utils';
 
 export interface WorkspaceFeatureSelectorProps {
-  applications: Array<
-    Pick<PublicAppInfo, 'id' | 'title' | 'category' | 'chromeless' | 'navLinkStatus'>
-  >;
   selectedFeatures: string[];
   onChange: (newFeatures: string[]) => void;
+  workspaceConfigurableApps?: PublicAppInfo[];
```
wanglam, line 24: Seems we only need id, title and category for workspaceConfigurableApps here. Could we change to `Pick<PublicAppInfo, 'id' | 'title' | 'category'>`?

src/plugins/workspace/public/utils.ts
```diff
@@ -99,3 +107,16 @@ export function isAppAccessibleInWorkspace(app: App, workspace: WorkspaceObject)
   }
   return false;
 }
+
+export const filterWorkspaceConfigurableApps = (applications: PublicAppInfo[]) => {
+  const visibleApplications = applications.filter(({ navLinkStatus, chromeless, category, id }) => {
+    return (
```
wanglam, line 113: Shall we add some unit tests here to cover this line?

ruanyl: Yes, there will be a followup PR from @yubonluo to cover this.

src/plugins/workspace/public/components/workspace_form/workspace_feature_selector.tsx
ruanyl: Actually I don't see a strong need to turn it into `Pick<PublicAppInfo, 'id' | 'title' | 'category'>` in this case. I understood that passing minimum set of data to child components is a best practice, but the `application ` which type is `PublicAppInfo` is provided by the core module, the components expect `PublicAppInfo` type of data sounds reasonable to me as well.

I'm not holding a strong opinion, what else concerns do you have?

SuZhou-Joe (approved):

src/plugins/workspace/public/components/workspace_form/workspace_feature_selector.tsx
wanglam: Not a big issue. We can keep current implementation.

ZilongX:
pick up latest main and re-running all checks now

ZilongX (approved):

PR #6364, opened 2024-04-08
Improve dynamic configurations by adding cache and simplifying client fetch

tianleh:
### Description

Address the comments from this issue https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6300

This PR updates the parameters to the ` getConfigurationClient` function and adds a cache.

### Issues Resolved

closes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6300
closes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6341

## Testing the changes

call the get API

```
curl -X GET 'http://localhost:5601/api/appconfig/csp.rules'

{"value":"frame-ancestors https://example-site.org"}%

```

Then call the get API again.

```
curl -X GET 'http://localhost:5601/api/appconfig/csp.rules'

{"value":"frame-ancestors https://example-site.org"}%
```

We find two entries in the logs where the first one doesn't show cache entry but the second one shows cache entry.

```
server    log   [02:13:58.836] [info][server][OpenSearchDashboards][http] http server running at http://localhost:5601
server    log   [02:14:04.440] [info][applicationConfig][plugins] Received a request to get entity config for csp.rules.
server    log   [02:14:04.441] [info][applicationConfig][plugins] Key csp.rules is not found from cache.
server    log   [02:14:07.936] [info][applicationConfig][plugins] Received a request to get entity config for csp.rules.
```

Verified that the CSP rules are set properly.
Screenshot 2024-04-09 at 19 00 10

call update API

```
curl 'http://localhost:5601/api/appconfig/csp.rules' -X POST -H 'Accept: application/json' -H 'Content-Type: application/json' -H 'osd-xsrf: osd-fetch' -H 'Sec-Fetch-Dest: empty' --data-raw '{"newValue":"script-src 'unsafe-eval' 'self'; worker-src blob: 'self'; style-src 'unsafe-inline' 'self'; frame-ancestors 'self' file://* filesystem:"}'

{"newValue":"script-src unsafe-eval self; worker-src blob: self; style-src unsafe-inline self; frame-ancestors self file://* 
```

call the get API

```
curl -X GET 'http://localhost:5601/api/appconfig/csp.rules'

{"value":"script-src unsafe-eval self; worker-src blob: self; style-src unsafe-inline self; frame-ancestors self file://* filesystem:"}%
```

From the logs, we can see that cache value is used.

```
[...]

PR #6474, opened 2024-04-16
fix dynamic config API calls to pass correct input

tianleh:
### Description

During https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6364/files, we simplify the input to `getConfigurationClient` from a client to a request. We were focusing on the use case of other plugins to onboard onto it. However, we also have API use case which we missed during the related PR. So, this PR is to fix the issue by passing request into the function instead of a client. 

Currently it is not failing because the `asScoped` function doesn't fail. Check the code reference here https://github.com/opensearch-project/OpenSearch-Dashboards/blob/eef417c03e7c4d34e85c70edb6c2079528ebc683/src/core/server/opensearch/client/cluster_client.ts#L137 It just treats it as using default headers and then doesn't pass user information to get a real scoped client.

## Testing the changes

Enable the following in YML.
```
# Set the value of this setting to true to enable plugin application config. By default it is disabled.
application_config.enabled: true

# Set the value of this setting to true to enable plugin CSP handler. By default it is disabled.
# It requires the application config plugin as its dependency.
csp_handler.enabled: true


```

call Get API

```
curl -X GET 'http://localhost:5601/api/appconfig/csp.rules.frame-ancestors'
```
```
```

call update API

```
curl 'http://localhost:5601/api/appconfig/csp.rules.frame-ancestors' -X POST -H 'Accept: application/json' -H 'Content-Type: application/json' -H 'osd-xsrf: osd-fetch' -H 'Sec-Fetch-Dest: empty' --data-raw '{"newValue":"file://* filesystem:"}'
```

```
{"newValue":"file://* filesystem:"}%
```

call get API again

```
curl -X GET 'http://localhost:5601/api/appconfig/csp.rules.frame-ancestors'
```

```
{"value":"file://* filesystem:"}%
```

call delete API

```
curl 'http://localhost:5601/api/appconfig/csp.rules.frame-ancestors' -X DELETE -H 'osd-xsrf: osd-fetch' -H 'Sec-Fetch-Dest: empty'
```

```
{"deletedEntity":"csp.rules.frame-ancestors"}%
```

call get API again.
```
curl -X GET 'http://localhost:5601/api/appconfig/csp.rules.frame-ancestors'
```

```
```

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [BUG][Multiple Datasource]Read hideLocalCluster setting from yml and set in data source selector and data source menu ([#6361](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6361))
 - [BUG][Multiple Datasource] Refactor read-only component to cover more edge cases ([#6416](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6416))
 - [BUG] Fix for checkForFunctionProperty so that order does not matter ([#6248](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6248))
+- [Dynamic Configurations] Fix dynamic config API calls to pass correct input ([#6474](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6474))
 - [BUG][Multiple Datasource] Validation succeed as long as status code in response is 200 ([#6399](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6399))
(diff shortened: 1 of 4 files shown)


tianleh:
@SuZhou-Joe Since you have some prior context about https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6364/files could you please help review this one? Thanks.

SuZhou-Joe (approved):
LGTM

ZilongX (approved):

ZilongX:
All checks passed and ready for merging

Issue #6420, opened 2024-04-12
[Multiple Datasource] Add new component to handle no available data sources and add to all data source components


yujin-emma:
**Is your feature request related to a problem? Please describe.**

There is no clear state or component when there is no data source found

**Describe the solution you'd like**

Add a "No data sources" display for the empty state
**Describe alternatives you've considered**

**Additional context**

PR #6499, opened 2024-04-16
add empty state and test except the aggregated view

yujin-emma:
### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6420

## Screenshot

when there is no remote data source but do not hide local cluster

empty-state-lc-new

when there is no remote data source but do hide local cluster

empty-state-hide-lc-new

when there is data source but filter out, still show empty-state
Screenshot 2024-04-16 at 22 55 33

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_source_management/public/components/data_source_selectable/data_source_selectable.tsx
```diff
@@ -180,6 +183,9 @@ export class DataSourceSelectable extends React.Component<
         'title',
         'auth.type',
       ]);
+      if (fetchedDataSources?.length === 0 && this.props.hideLocalCluster) {
```
BionIT, line 186: instead of evaluate fetechedDataSources, we should evaluate the options after filter, right?

yujin-emma: yeah, updated

BionIT (approved):

zhongnansu (approved):

PR #6052, opened 2024-03-06
[Workspace]Add permission control logic for workspace

wanglam:
### Description

This PR is for adding permission control logic for workspace. It's includes below changes:
1. Add `savedObjects.permission` to global config object
2. Add permissions field in workspace create and update API
3. Consume workspaces in repository get method
4. Add `ACLSearchParams` and `workspaceSearchOperator` to repository find method
5. Add permission control client and workspace saved object client wrapper

### Issues Resolved

closes #6051 

## Testing the changes

Write unit tests and integration tests for workspace saved object client wrapper. 

### Test instructions

Since all these changes are in the server side. We need to call these APIs manual to verify if permission control work fine.
There are two types saved objects in permission control. The first one is saved object with `workspaces` property, another one is saved object with `permissions` property. In this test instruction, we will use `workspace` type saved object to verify permission control when has `permissions` property. Use `dashboard` type saved object to verify permission control when has `workspaces` property. If one saved object has `workspaces` property, the permission control logic will check if has related permissions to the workspaces. Then if it has `permissions` property, it will do the permission validation on the `permissions` property.

```yml
savedObjects.permission.enabled: true
workspace.enabled: true
```

To run all below tests, need to add above flags in `opensearch_dashboards.yml`, and install security-dashboards-plugin. There are two internal users will be used in following tests. Here are the user details:
- user1:
    username: admin, password: myStrongPassword123! backend-roles: admin
- user2:
    username: another-user, password: myStrongPassword123! backend-roles: kibanauser

The `admin` user is not a specific user here, all the permission control process is the same as a normal user. 
We will add `authorization: Basic YWRtaW46bXlTdHJvbmdQYXNzd29yZDEyMyE=` to `curl` command to simulate `admin` user and add `authorization: Basic YW5vdGhlci11c2VyOm15U3Ryb25nUGFzc3dvcmQxMjMh` to simulate `another-user`. 
After all the environments ready and user created, we can start to test permission control feature.

##### Create workspace
This steps is for creating test workspace for future test cases. Run below workspace create API.
```sh
curl 'http://localhost:5601/api/workspaces' \
  -X 'POST' \
  -H 'authorization: Basic YWRtaW46bXlTdHJvbmdQYXNzd29yZDEyMyE=' \
  -H 'Content-Type: application/json' \
  -H 'osd-xsrf: osd-fetch' \
  -H 'osd-version: 3.0.0' \
  --data-raw '{"attributes":{"name":"admin-only-workspace"}}' \
  --compressed
[...]

PR #6531, opened 2024-04-18
[Backport 2.x][Workspace]Add permission control logic for workspace

### LINKED ISSUE #6051: [Workspace] Add permission control logic for saved objects

This issue is a feature request. Workspace will be a new concept to organize saved objects. The saved objects under specific workspaces should be isolate. There are three key features need to be implement for the permission control logic.

- Create / update workspace with permission setting

Each workspace should be independent of each other. When calling the create workspace API, it should be able to set which users have permissions on the workspace. The user who calls the creation API should have the permission to modify the saved objects in the workspace and update the workspace by default. In addition, for the created workspace, it should be able to update the permission settings through the workspace update API.

- Find saved objects by ACL params

In Opensearch Dashboards, the find method in the saved objects client is generally used to search and filter saved objects. After adding permission control logic, the find method can only return data that the user has permission to access. Therefore, the find method should support querying saved objects through ACL to limit query returns in permission control logic.

- Add permission control for saved object client CRUD

In Opensearch Dashboards, saved objects client is generally used to operate saved objects. Operations on saved objects should follow the corresponding permission settings if permission control is turned on. For example, when creating a saved object, if the user specifies that it should be created in a specific workspace. Then when performing the creation operation, it should check whether the user has write permissions for the workspace. The above steps should be followed for all CRUD operations, and it is necessary to check whether the user has permissions for the corresponding workspace or saved objects.

For the above three main features, we need to modify the code in OSD to implement them. Most of the code will be located in the workspaces plug-in, and only a small part that needs to modify the search dsl will be placed in the saved objects service. There are four kinds of permission modes. Here is a [fully introduction](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5083) about these permission modes.
All the permission control related features has been protected by `savedObjects.permission.enabled` and `workspace.enabled` flag. Need to set these two flags to `true` to enable them. 

- Create / update workspace with permission setting

To add permission settings to the workspace, you need to add the corresponding permission setting fields to the workspace creation and update methods. Then put the incoming permission settings into the corresponding saved objects client method call. This completes the creation and update of the workspace's permission settings.

- Find saved objects by ACL params

For the find method in the saved objects client, new `ACLSearchParams` and `workspaceSearchOperator` fields will be added. `ACLSearchParams` is used to specify the user name, user group and corresponding permission information to be searched.
```ts
interface Principals {
  users?: string[];
  groups?: string[];
}

interface SavedObjectsFindOptions {
  //...other fields
  /** If specified, will only retrieve objects that are in the workspaces */
  workspaces?: string[];
  /** By default the operator will be 'AND' */
  workspacesSearchOperator?: 'AND' | 'OR';
  /**
   * The params here will be combined with bool clause and is used for filtering with ACL structure.
   */
  ACLSearchParams?: {
    principals?: Principals;
    permissionModes?: string[];
  };
}
```
The `workspaceSearchOperator` is used to specify the search method for the incoming workspaces. The value will be `AND` by default, which means that the result data returned by the query will belong to the incoming workspaces. If it is `OR`, which means that in addition to other ACL conditions, the data in the incoming workspaces will also be returned. 
Here are some examples:
```json
{
    "workspaces": ["workspace-1"]
}
```
For this case, the result will only return saved objects belongs to workspace-1.
```json
{
    "workspaces": ["workspace-1"],
    "workspacesSearchOperator": "OR"
    "ACLSearchParams": [
        {
            "principals": {
                "users": ["user-1"]
            },
            "permissionModes": ["read", "write"]
        }
    ]
}
```
For this case, the result will return saved objects belongs to workspace-1 and all saved objects that user-1 has read and write permission.

- Add permission control for saved object client CRUD

The main solution of add permission validation logic is using saved objects client wrapper just like the [workspace conflicts validation logic](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/workspace/server/saved_objects/saved_objects_wrapper_for_check_workspace_conflict.ts). The saved objects client related methods will be overwrite after adding client wrapper to the saved objects service. This client wrapper depends on permission control service. The permission control service provide methods to validate saved objects based `permissions` attribute. It's using [`hasPermission`](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5083) in `ACL` to validate if saved object matched passed permission modes and principles. For CRUD methods in client wrapper, there are different validation rules for different methods.
* create / bulkCreate
For `create` method, it will check if has `library_write` permission to passed workspaces when `options.workspaces` was provided. It will throw permission error if doesn't have permission. Since the `create` method can overwrite existing saved objects, it will validate permission if `options.overwrite` and `options.id` passed. It will try to get existing saved objects, and validate `library_write` permission to its workspaces or `write` permission to itself. It will throw permission error if not permitted. The `bulkCreate` permission validation logic was the very similar as `create`. The `bulkCreate` will do these permission validation logic for all passed saved objects. 
* get / bulkGet
For `get` method, it will get saved object first. And then will check if has `library_read` or `library_write` permission to **one of** the saved object's workspaces. If it doesn't have permission, it will check the `read` or `write` permission on saved object. Then throw permission error if not permitted. The `bulkGet` method do the same thing for all passed saved objects.
* find
For `find` method, there are three different cases. 
The `find` method will check the `options.type` first. This is the **case 1**, for finding `workspace` type saved objects. It will assign the `principals` from request to the `ACLSearchParams.principals` and assign `read` or `write` to the `ACLSearchParams.permissionModes` too if not provided. This means find all workspace saved objects that user has read and write permission. 
If `options.type` was not provided, these are case 2 and case 3. For these two cases, need to get user permitted workspaces first. 
If `options.workspaces` provided, this is **case 2**. It will filter out not permitted workspaces in `options.workspaces`. This means find all permitted saved objects under permitted workspaces. 
If `options.workspaces` was not provided, it's **case 3**. It will assign request principals to `options.ACLSearchParams.principals`. Which means find all permitted saved objects based user passed permission modes.
* update / bulkUpdate
For `update` method, it will get passed saved objects first. Then it will check if has **one of** the `library_write` permission to saved objects's workspace or if has `write` permission to saved object itself. It will throw permission error if not permitted. The `bulkUpdate` methods do the same things. It will check the passed saved objects one by one.
* delete
For delete method, it will get the delete saved object first. If the delete saved object has `workspaces` attribute, it will check if have `library_write` permissions to delete saved object's workspaces. If the user doesn't have `library_write` permission to these workspaces, it will check the `write` permission of delete saved object. It will throw permission error if not permitted.
* deleteByWorkspace
For deleteByWorkspace method, it will validate if user has `library_write` permission to passed workspace. If the user doesn't have this permission, it will throw permission error. If user has related permission, all the saved objects belongs to the workspace will be deleted.

N/A
**Additional context**
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5083
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5163
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4615

wanglam:
### Description

backport fb31b2def6a6200425492d772a8d0b1bdfcbe132 from #6052 

This PR is for manual backport #6052 to 2.x branch.

### Issues Resolved

#6051 

## Testing the changes

See #6052 

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/public/saved_objects/saved_objects_client.ts
+++ b/src/core/public/saved_objects/saved_objects_client.ts
@@ -45,7 +45,11 @@ import { HttpFetchOptions, HttpSetup } from '../http';
 
 type SavedObjectsFindOptions = Omit<
   SavedObjectFindOptionsServer,
-  'sortOrder' | 'rootSearchFields' | 'typeToNamespacesMap'
+  | 'sortOrder'
+  | 'rootSearchFields'
+  | 'typeToNamespacesMap'
+  | 'ACLSearchParams'
+  | 'workspacesSearchOperator'
 >;
 
 type PromiseType<T extends Promise<any>> = T extends Promise<infer U> ? U : never;
--- a/src/core/server/index.ts
+++ b/src/core/server/index.ts
@@ -322,6 +322,10 @@ export {
   exportSavedObjectsToStream,
   importSavedObjectsFromStream,
   resolveSavedObjectsImportErrors,
+  ACL,
+  Principals,
+  PrincipalType,
+  Permissions,
   updateDataSourceNameInVegaSpec,
   extractVegaSpecFromSavedObject,
 } from './saved_objects';
--- a/src/core/server/mocks.ts
+++ b/src/core/server/mocks.ts
@@ -89,6 +89,9 @@ export function pluginInitializerContextConfigMock<T>(config: T) {
     path: { data: '/tmp' },
     savedObjects: {
       maxImportPayloadBytes: new ByteSizeValue(26214400),
+      permission: {
+        enabled: true,
+      },
     },
   };
 
--- a/src/core/server/plugins/plugin_context.test.ts
+++ b/src/core/server/plugins/plugin_context.test.ts
@@ -108,7 +108,12 @@ describe('createPluginInitializerContext', () => {
         pingTimeout: duration(30, 's'),
       },
       path: { data: fromRoot('data') },
-      savedObjects: { maxImportPayloadBytes: new ByteSizeValue(26214400) },
+      savedObjects: {
+        maxImportPayloadBytes: new ByteSizeValue(26214400),
+        permission: {
+          enabled: false,
+        },
+      },
     });
   });
 
--- a/src/core/server/plugins/types.ts
+++ b/src/core/server/plugins/types.ts
@@ -295,7 +295,7 @@ export const SharedGlobalConfigKeys = {
   ] as const,
   opensearch: ['shardTimeout', 'requestTimeout', 'pingTimeout'] as const,
   path: ['data'] as const,
-  savedObjects: ['maxImportPayloadBytes'] as const,
+  savedObjects: ['maxImportPayloadBytes', 'permission'] as const,
 };
 
 /**
--- a/src/core/server/saved_objects/index.ts
+++ b/src/core/server/saved_objects/index.ts
@@ -84,3 +84,5 @@ export {
 
 export { savedObjectsConfig, savedObjectsMigrationConfig } from './saved_objects_config';
 export { SavedObjectTypeRegistry, ISavedObjectTypeRegistry } from './saved_objects_type_registry';
+
+export { Permissions, ACL, Principals, PrincipalType } from './permission_control/acl';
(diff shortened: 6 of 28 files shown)


ruanyl (approved):

ZilongX:
picking up latest 2.x and re-running all checks

SuZhou-Joe (approved):

PR #6543, opened 2024-04-18
[MDS] TSVB Add import logic

### LINKED ISSUE #6290: [MDS] Full TSVB Support
## Proposal
As of today, [Time Series Visual Builder (TSVB)](https://opensearch.org/docs/2.12/dashboards/visualize/viz-index/#tsvb)  does not support remote datasource connections (also referred to as multiple datasource or MDS). [#6005](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6005) provides most of the building blocks to support TSVB. However, there are a few features to consider to fully integrate MDS into TSVB:
- **Export/import logic:** With [#5712](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5712), users can specify which datasource to import from. TSVB should seamlessly import from different datasources.
- **Timeseries annotations:** One feature of the timeseries graph visualization is the ability to place annotations or markers on the graph. Each annotation has its own index pattern to fetch data from.

In this issue, these features will be the main focus. For details on how TSVB can support MDS, refer to [#6005](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6005).

## High-Level Approach
TSVB, like other visualizations, store all configurations within the `visState`. For TSVB, an example `visState` looks like the following:
```
{
  title: 'Test',
  type: 'metrics',
  aggs: [],
  params: {
    annotations: [{
      fields: '',
      template: 'field',
      index_pattern: 'open*',
      query_string: {
        query: 'some-query',
        language: 'kuery'
      },
      color: '#F00',
      icon: 'fa-asterisk',
      id: 'c1693f50-e7ae-11ee-a975-3b6cd68aa468',
      ignore_global_filters: 1,
      ignore_panel_filters: 1,
      time_field: '@timestamp'
    }, {
      fields: 'some-field',
      template: 'field',
      index_pattern: 'open*',
      query_string: {
        query: 'some-query',
        language: 'kuery'
      },
      color: '#F00',
      icon: 'fa-tag',
      id: 'c278da90-e7ae-11ee-a975-3b6cd68aa468',
      ignore_global_filters: 1,
      ignore_panel_filters: 1,
      time_field: '@timestamp'
    }],
    axis_formatter: 'number',
    axis_position: 'left',
    axis_scale: 'normal',
    background_color_rules: [
      [Object]
    ],
    bar_color_rules: [
      [Object]
    ],
    default_index_pattern: 'test*',
    drop_last_bucket: 0,
    filter: {
      language: 'kuery',
      query: ''
    },
    gauge_color_rules: [
      [Object]
    ],
    gauge_inner_width: 10,
    gauge_style: 'half',
    gauge_width: 10,
    id: '61ca57f0-469d-11e7-af02-69e470af7417',
    index_pattern: 'opensearch_*',
	// New field to be added in params
	data_source_id: 'some-datasource-id',
    interval: '',
    isModelInvalid: false,
    series: [
      [Object]
    ],
    show_grid: 1,
    show_legend: 1,
    time_field: null,
    tooltip_mode: 'show_all',
    type: 'timeseries'
  }
}
```

[#5717](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5717) introduced new datasource picker components exposed from core for plugins to consume. This provides the plugin with `dataSourceId` information that can be used to pass into the [`/api/metrics/fields`](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/vis_type_timeseries/server/routes/fields.ts#L39) and [`/api/metrics/vis/data`](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/vis_type_timeseries/server/routes/vis.ts#L47) API calls. When a user selects the datasource, the `params` object will be updated to add a new field `data_source_id: <id>`. 

Below is an example of the new component added to the Panel options tab:
TSVB Panel options edit box

As for export/import logic, there first needs to be a datasource reference since the TSVB visualization will make use of a datasource. This can be configured at save time where the reference will be appended. To do this, a custom saved objects client wrapper can be defined within the `server/plugin` and the `create()` call can be overridden to update the datasource reference. Then, in the saved objects import logic (`src/core/server/saved_objects/import/`), the `visState` object will be updated to add/remove/update the datasource id and the associated datasource reference.

## Alternatives 
Timeseries, one of the visualization types of TSVB, supports annotations as a feature of the timeseries graph. **Annotations** are a way to place markers at specific points in time on the timeseries graph. A common use case for annotations is to mark an event on the graph that would be associated with the timeseries data. For example, OpenSearch can ingest some CPU utilization from a VM in an index pattern `cpu-utilization-*` and in a separate index `deployment-info-*`, store any instance of restarting/starting a VM job. If the user wants to monitor which VM job caused the spike in CPU utilization, the user could make the `cpu-utilization-*` the timeseries graph with the `deployment-info-*` pattern as the annotation.

Below is an example of annotations rendered alongside the timeseries graph.
TSVB with annotations

The current proposal will use the datasource taken from the Panel options tab meaning annotations can only pull data from the same datasource as the panel. However, since TSVB makes separate `search` queries to obtain annotation data, there is a use case to enable a user to select which datasource to pull the annotation from via a datasource picker component added under the Annotations tab. This means that a user can create annotations from indices of different datasources. 

Below is a mockup of the new component:
Annotation edit box

However, there are caveats with this added feature:
- **Complex import/export logic:** import was designed with one datasource in mind. Import will be complex and capturing user intent can be a challenge. However, another issue ([#5927](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5927)) faced a similar issue so this visualization can follow a similar approach. It is not a blocking challenge but is listed here nonetheless as there does not exist a perfect solution to address this limitation.
- **Increased `search()` API calls:** the current annotations make one `bulk search()` call to fetch annotation data. This was only possible because there was no consideration for the datasource used. With this proposal, there is a chance that each annotation will call a different datasource, therefore increasing the number of `search()` calls and incurring additional performance costs.

huyaboo:
### Description

This PR adds datasource references to TSVB visualizations, which are needed to successfully export/import the object in MDS world. This PR also adds import logic to select a datasource.

### Issues Resolved

Closes #6290

### Adding datasource reference

**Recording:**

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/73027756/96f1711f-bee7-44b6-8034-6dd385bfc1a0

**Test case steps:**
1. Add sample data
2. Navigate to the TSVB edit page and change the datasource 
3. Go to saved objects management page and navigate to the visualization
4. In references, this should be modified to add the correct datasource
5. If the datasource is updated in the TSVB edit page, so too will the datasource reference
6. If the local cluster is used, the datasource reference will be removed.

### For import logic

**Test cases:**
1. Non-MDS TSVB visualization import to `Source B` and create new object
    - This should append `Source B` to references and should use `Source B` as the datasource to query from

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/73027756/eb249935-13ca-4d99-8a5a-20e6189fbf8c

2. Non-MDS TSVB visualization import to `Source B` and overwrite old object
    - This should append `Source B` to references and should use `Source B` as the datasource to query from

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/73027756/06d65180-90b3-4434-8244-3472d8230d98

3. MDS TSVB visualization import to `Source A` and create new object
    - This should update `Source B` to `Source A` in references and should use `Source A` as the datasource to query from

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/73027756/b7f1e663-f461-42a6-8206-662b14b5129c

4. MDS TSVB visualization import to `Source A` and override old object
    - This should update `Source B` to `Source A` in references and should use `Source A` as the datasource to query from

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/73027756/0bae86bf-12b7-4fad-a654-7ae455b0a23a

## Changelog
- feat: [MDS] Add import logic to TSVB visualizations

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/server/saved_objects/import/check_conflict_for_data_source.test.ts
+++ b/src/core/server/saved_objects/import/check_conflict_for_data_source.test.ts
@@ -330,9 +330,9 @@ describe('#checkConflictsForDataSource', () => {
   ])('will update datasource reference + visState of TSVB visualization', async ({ id }) => {
     const tsvbSavedObject = createTSVBVisualizationObject(id);
     // @ts-expect-error
-    const newVisState = JSON.parse(tsvbSavedObject.attributes.visState);
-    newVisState.params.data_source_id = 'some-datasource-id';
-    const expectedVisState = JSON.stringify(newVisState);
+    const expectedVisState = JSON.parse(tsvbSavedObject.attributes.visState);
+    expectedVisState.params.data_source_id = 'some-datasource-id';
+    const newVisState = JSON.stringify(expectedVisState);
     const params = setupParams({
       objects: [tsvbSavedObject],
       ignoreRegularConflicts: true,
@@ -348,7 +348,7 @@ describe('#checkConflictsForDataSource', () => {
             ...tsvbSavedObject,
             attributes: {
               title: 'some-title',
-              visState: expectedVisState,
+              visState: newVisState,
             },
             id: 'some-datasource-id_some-object-id',
             references: [
--- a/src/core/server/saved_objects/import/check_conflict_for_data_source.test.ts
+++ b/src/core/server/saved_objects/import/check_conflict_for_data_source.test.ts
@@ -40,6 +40,23 @@ const createVegaVisualizationObject = (id: string): SavedObjectType => {
   } as SavedObjectType;
 };
 
+const createTSVBVisualizationObject = (id: string): SavedObjectType => {
+  const idParse = id.split('_');
+  const params = idParse.length > 1 ? { data_source_id: idParse[1] } : {};
+  const visState = {
+    type: 'metrics',
+    params,
+  };
+
+  return {
+    type: 'visualization',
+    id,
+    attributes: { title: 'some-title', visState: JSON.stringify(visState) },
+    references:
+      idParse.length > 1 ? [{ id: idParse[1], type: 'data-source', name: 'dataSource' }] : [],
+  } as SavedObjectType;
+};
+
 const getSavedObjectClient = (): SavedObjectsClientContract => {
   const savedObject = {} as SavedObjectsClientContract;
   savedObject.get = jest.fn().mockImplementation((type, id) => {
@@ -294,6 +311,60 @@ describe('#checkConflictsForDataSource', () => {
           [
             `visualization:some-object-id`,
             { id: 'nonexistent-datasource-title-id_some-object-id', omitOriginId: true },
+          ],
+        ]),
+      })
+    );
+  });
+
+  /**
+   * TSVB test cases
+   */
+  it.each([
+    {
+      id: 'some-object-id',
+    },
+    {
+      id: 'old-datasource-id_some-object-id',
+    },
+  ])('will update datasource reference + visState of TSVB visualization', async ({ id }) => {
+    const tsvbSavedObject = createTSVBVisualizationObject(id);
+    // @ts-expect-error
+    const newVisState = JSON.parse(tsvbSavedObject.attributes.visState);
+    newVisState.params.data_source_id = 'some-datasource-id';
+    const expectedVisState = JSON.stringify(newVisState);
+    const params = setupParams({
+      objects: [tsvbSavedObject],
+      ignoreRegularConflicts: true,
+      dataSourceId: 'some-datasource-id',
+      savedObjectsClient: getSavedObjectClient(),
+    });
+    const checkConflictsForDataSourceResult = await checkConflictsForDataSource(params);
+
+    expect(checkConflictsForDataSourceResult).toEqual(
+      expect.objectContaining({
+        filteredObjects: [
+          {
+            ...tsvbSavedObject,
+            attributes: {
+              title: 'some-title',
+              visState: expectedVisState,
+            },
+            id: 'some-datasource-id_some-object-id',
+            references: [
+              {
+                id: 'some-datasource-id',
+                name: 'dataSource',
+                type: 'data-source',
+              },
+            ],
+          },
+        ],
+        errors: [],
+        importIdMap: new Map([
+          [
+            `visualization:some-object-id`,
+            { id: 'some-datasource-id_some-object-id', omitOriginId: true },
           ],
         ]),
       })
--- a/src/core/server/saved_objects/import/check_conflict_for_data_source.test.ts
+++ b/src/core/server/saved_objects/import/check_conflict_for_data_source.test.ts
@@ -11,6 +11,7 @@ import {
   checkConflictsForDataSource,
   ConflictsForDataSourceParams,
 } from './check_conflict_for_data_source';
+import { VisualizationObject } from './types';
 
 type SavedObjectType = SavedObject<{ title?: string }>;
 
@@ -40,7 +41,7 @@ const createVegaVisualizationObject = (id: string): SavedObjectType => {
   } as SavedObjectType;
 };
 
-const createTSVBVisualizationObject = (id: string): SavedObjectType => {
+const createTSVBVisualizationObject = (id: string): VisualizationObject => {
   const idParse = id.split('_');
   const params = idParse.length > 1 ? { data_source_id: idParse[1] } : {};
   const visState = {
@@ -54,7 +55,7 @@ const createTSVBVisualizationObject = (id: string): SavedObjectType => {
     attributes: { title: 'some-title', visState: JSON.stringify(visState) },
     references:
       idParse.length > 1 ? [{ id: idParse[1], type: 'data-source', name: 'dataSource' }] : [],
-  } as SavedObjectType;
+  } as VisualizationObject;
 };
 
 const getSavedObjectClient = (): SavedObjectsClientContract => {
@@ -329,7 +330,6 @@ describe('#checkConflictsForDataSource', () => {
     },
   ])('will update datasource reference + visState of TSVB visualization', async ({ id }) => {
     const tsvbSavedObject = createTSVBVisualizationObject(id);
-    // @ts-expect-error
     const expectedVisState = JSON.parse(tsvbSavedObject.attributes.visState);
     expectedVisState.params.data_source_id = 'some-datasource-id';
     const newVisState = JSON.stringify(expectedVisState);
--- a/src/core/server/saved_objects/import/check_conflict_for_data_source.ts
+++ b/src/core/server/saved_objects/import/check_conflict_for_data_source.ts
@@ -96,6 +96,8 @@ export async function checkConflictsForDataSource({
         // Some visualization types will need special modifications, like Vega visualizations
         if (object.type === 'visualization') {
           const vegaSpec = extractVegaSpecFromSavedObject(object);
+          // @ts-expect-error
+          const visStateObject = JSON.parse(object.attributes?.visState);
 
           if (!!vegaSpec && !!dataSourceTitle) {
             const updatedVegaSpec = updateDataSourceNameInVegaSpec({
@@ -103,8 +105,6 @@ export async function checkConflictsForDataSource({
               newDataSourceName: dataSourceTitle,
             });
 
-            // @ts-expect-error
-            const visStateObject = JSON.parse(object.attributes?.visState);
             visStateObject.params.spec = updatedVegaSpec;
 
             // @ts-expect-error
@@ -116,6 +116,25 @@ export async function checkConflictsForDataSource({
                 type: 'data-source',
               });
             }
+          }
+
+          if (visStateObject.type === 'metrics' && !!dataSourceId) {
+            const oldDataSourceId = visStateObject.params.data_source_id;
+            const newReferences = object.references.filter((reference) => {
+              return reference.id !== oldDataSourceId && reference.type === 'data-source';
+            });
+
+            visStateObject.params.data_source_id = dataSourceId;
+
+            newReferences.push({
+              id: dataSourceId,
+              name: 'dataSource',
+              type: 'data-source',
+            });
+
+            // @ts-expect-error
+            object.attributes.visState = JSON.stringify(visStateObject);
+            object.references = newReferences;
           }
         }
 
--- a/src/core/server/saved_objects/import/check_conflict_for_data_source.ts
+++ b/src/core/server/saved_objects/import/check_conflict_for_data_source.ts
@@ -9,6 +9,7 @@ import {
   SavedObjectsImportError,
   SavedObjectsImportRetry,
 } from '../types';
+import { VisualizationObject } from './types';
 import {
   extractVegaSpecFromSavedObject,
   getDataSourceTitleFromId,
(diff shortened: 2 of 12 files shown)


src/core/server/saved_objects/import/create_saved_objects.ts
```diff
@@ -125,6 +125,24 @@ export const createSavedObjects = async <T>({
...
+
+            object.references.push({
+              id: dataSourceId,
+              name: 'dataSource',
+              type: 'data-source',
+            });
+
+            // @ts-expect-error
+            object.attributes.visState = JSON.stringify(visStateObject);
+          }
```
zhyuanqi, line 145: Hi Huy. Looks like the code here is very similar to the check_conflict except check_conflict makes sure datasourceId exists. Can we combine these two and adds  it in the util?

src/plugins/vis_type_timeseries/server/plugin.ts
```diff
@@ -115,6 +120,13 @@ export class VisTypeTimeseriesPlugin implements Plugin<VisTypeTimeseriesSetup> {
       searchStrategyRegistry,
     };
 
+    setDataSourceEnabled({ enabled: !!plugins.dataSource });
+    core.savedObjects.addClientWrapper(
+      11,
```
zhyuanqi, line 125: Hi Huy, how we we decide this priority here.

src/plugins/vis_type_timeseries/server/lib/timeseries_visualization_client_wrapper.ts
```diff
@@ -0,0 +1,82 @@
...
+      return await wrapperOptions.client.create(type, attributes, options);
+    }
+
+    // @ts-expect-error
+    const visState = JSON.parse(attributes.visState);
+
+    if (visState.type !== 'metrics' || !visState.params) {
+      return await wrapperOptions.client.create(type, attributes, options);
+    }
+
```
zhyuanqi, line 34: Hi Huy I am really new to fix TSVB, could you help me to understand the logic here?  
I see the first condition `(type !== 'visualization' || !getDataSourceEnabled().enabled)` and ` (visState.type !== 'metrics' || !visState.params) ` both returns  `return await wrapperOptions.client.create(type, attributes, options); `  Can we combine it here?

zhyuanqi: So only when type is metrics and visualization, then we consider adding datasource?

huyaboo: So this client wrapper can receive any type of saved object and only Visualization objects will have the `visState` property. Thus, the first condition checks if the object is a visualization object (or if MDS is disabled) before proceeding. The second condition then checks that the visualization object has `params` and is a `metrics` (TSVB) type.

src/plugins/vis_type_timeseries/server/plugin.ts
huyaboo: The lower the number, the higher the priority a client wrapper is. After some investigation, these should be unique and since `10` was used for Vega, 11 was chosen.

src/core/server/saved_objects/import/create_saved_objects.ts
huyaboo: Refactored

src/plugins/vis_type_timeseries/server/lib/timeseries_visualization_client_wrapper.ts
```diff
@@ -0,0 +1,82 @@
...
+    const newReferences = options?.references?.filter(
+      (reference) => reference.type !== 'data-source'
+    );
+
+    if (visState.params.data_source_id) {
+      if (await checkIfDataSourceExists(visState.params.data_source_id, wrapperOptions.client)) {
```
BionIT, line 40: how do we handle network failure when check data source exists? Will creation be successful?

src/plugins/vis_type_timeseries/server/plugin.ts
BionIT, line 125: Can we give the number a more meanful name?

BionIT:
One thing noticed in the recording that for TSVB visualization created with MDS, we don't seem to add data source name as suffix, but from import, looks like we do

src/plugins/vis_type_timeseries/server/lib/timeseries_visualization_client_wrapper.ts
huyaboo: Added `try catch` block to throw an error in the case that the promise rejects.

src/plugins/vis_type_timeseries/server/plugin.ts
huyaboo: Refactored to `TIMESERIES_VISUALIZATION_CLIENT_WRAPPER_PRIORITY`

huyaboo:
> One thing noticed in the recording that for TSVB visualization created with MDS, we don't seem to add data source name as suffix, but from import, looks like we do

I'm not sure if this is a good idea to append the datasource name to the visualization on create. I believe the import logic appends a datasource name because there could be potential duplicate title conflicts when importing. For example, a user can import a saved object twice: one to `Source A` and one to `Source B` so the way to distinguish the two is to append the datasource name.

BionIT (dismissed):

src/core/server/saved_objects/import/check_conflict_for_data_source.test.ts
```diff
@@ -299,4 +316,58 @@ describe('#checkConflictsForDataSource', () => {
...
+  it.each([
+    {
+      id: 'some-object-id',
+    },
+    {
+      id: 'old-datasource-id_some-object-id',
+    },
+  ])('will update datasource reference + visState of TSVB visualization', async ({ id }) => {
+    const tsvbSavedObject = createTSVBVisualizationObject(id);
+    // @ts-expect-error
```
bandinib-amzn, line 332: nit: Let's remove `@ts-expect-error` and handle Type error.

src/core/server/saved_objects/import/check_conflict_for_data_source.ts
```diff
@@ -117,6 +118,14 @@ export async function checkConflictsForDataSource({
               });
             }
           }
+
+          if (!!dataSourceId) {
+            const { visState, references } = getUpdatedTSVBVisState(object, dataSourceId);
+
+            // @ts-expect-error
```
bandinib-amzn, line 125: same as above.

src/plugins/vis_type_timeseries/server/lib/timeseries_visualization_client_wrapper.ts
```diff
@@ -0,0 +1,89 @@
...
+  SavedObjectsClientContract,
+  SavedObjectsClientWrapperFactory,
+  SavedObjectsClientWrapperOptions,
+  SavedObjectsCreateOptions,
+  SavedObjectsErrorHelpers,
+} from '../../../../core/server';
+import { getDataSourceEnabled } from './services';
+
+export const TIMESERIES_VISUALIZATION_CLIENT_WRAPPER_ID = 'timeseries-visualization-client-wrapper';
+export const TIMESERIES_VISUALIZATION_CLIENT_WRAPPER_PRIORITY = 11;
```
bandinib-amzn, line 16: Can you add comment to describe why you are using 11 priority?

bandinib-amzn: Did you confirm that we are not using this priority in any other plugin?

src/core/server/saved_objects/import/check_conflict_for_data_source.test.ts
```diff
@@ -299,4 +316,58 @@ describe('#checkConflictsForDataSource', () => {
...
+    const newVisState = JSON.parse(tsvbSavedObject.attributes.visState);
+    newVisState.params.data_source_id = 'some-datasource-id';
+    const expectedVisState = JSON.stringify(newVisState);
```
bandinib-amzn, line 335: nit:
```
const expectedVisState = tsvbSavedObject.attributes.visState;
const newVisState = JSON.parse(expectedVisState);
```

src/core/server/saved_objects/import/utils.ts
```diff
@@ -19,6 +19,43 @@ export interface UpdateDataSourceNameInVegaSpecProps {
   spacing?: number;
 }
 
+/**
+ * Given a visualization saved object and datasource id, return the updated visState and references if the visualization was a TSVB visualization
+ * @param {SavedObject} object
+ * @param {string} dataSourceId
+ * @returns {{visState: string, references: SavedObjectReference[]}} - the updated stringified visState and references
+ */
+export const getUpdatedTSVBVisState = (object: SavedObject, dataSourceId: string) => {
```
bandinib-amzn, line 28: Let's define type of object to SavedObject<T> and then object.attributes as T.

src/plugins/vis_type_timeseries/server/lib/timeseries_visualization_client_wrapper.test.ts
```diff
@@ -0,0 +1,233 @@
...
+      title: 'some-dashboard',
+    };
+    const wrapper = timeSeriesVisualizationClientWrapper(mockedWrapperOptions);
+    await wrapper.create('dashboard', attributes, { references: [] });
+
+    expect(client.create).toBeCalledWith(
+      'dashboard',
+      attributes,
+      expect.objectContaining({ references: [] })
+    );
```
bandinib-amzn, line 77: This block seems to be repeated. Can we extract this to function for better maintenance?

src/core/server/saved_objects/import/check_conflict_for_data_source.test.ts
huyaboo: Added an internal type `VisualizationObject` to `saved_objects/import`

src/plugins/vis_type_timeseries/server/lib/timeseries_visualization_client_wrapper.ts
huyaboo: Added comment regarding using priority. I checked how other plugins were using priority and they used `-4,-3,-2,-1,-,1`

src/core/server/saved_objects/import/check_conflict_for_data_source.test.ts
huyaboo: Refactored

src/core/server/saved_objects/import/utils.ts
huyaboo: I defined a type `VisualizationType`

bandinib-amzn (approved):

BionIT (approved):

Issue #6535, opened 2024-04-18
[BUG] Data source selectable and readonly component are not consistent


BionIT:
**Describe the bug**

- The height of the popover header should be consistent between the read-only and selectable
- The stroke around the header looks better in the read-only popover, the floating line between the search field and the header looks a bit weird. This might be a part of next item.
- It looks like additional padding is added to the selectable popover content

**To Reproduce**
1. Go to future playground
2. Go to security plugin or index management plugin
3. Go to main page vs details page
4. See problem

**Expected behavior**

**OpenSearch Version**
Please list the version of OpenSearch being used.

**Dashboards Version**
Please list the version of OpenSearch Dashboards being used.

**Plugins**

Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

zhyuanqi:
Will work on it.

BionIT:
Awesome, thanks @zhyuanqi

BionIT:
I believe this is fixed, please reopen if not @zhyuanqi

PR #6545, opened 2024-04-18
Fix on data source selectable and readonly component are not consistent

zhyuanqi:
### Description

fix on data source selectable and readonly component are not consistent
Related Issue: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6535

Describe the bug

* The height of the popover header should be consistent between the read-only and selectable
* The stroke around the header looks better in the read-only popover, the floating line between the search field and the header looks a bit weird. This might be a part of next item.
* It looks like additional padding is added to the selectable popover content
### screenshot

Screenshot 2024-04-18 at 2 39 07 PM

Screenshot 2024-04-18 at 2 39 32 PM

## Changelog
- fix: fix on data source selectable and readonly component are not consistent

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
+++ b/src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
@@ -193,21 +193,9 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
         panelPaddingSize="none"
         anchorPosition="downLeft"
       >
-        <EuiContextMenuPanel>
-          <EuiPanel
-            className={'dataSourceViewOuiPanel'}
-            borderRadius="none"
-            color="transparent"
-            paddingSize="s"
-          >
-            {
-              <DataSourceDropDownHeader
-                totalDataSourceCount={1}
-                application={this.props.application}
-              />
-            }
-          </EuiPanel>
-          <EuiPanel color="subdued" paddingSize="xs" borderRadius="none">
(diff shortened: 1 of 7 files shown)


src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -193,21 +193,9 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
-          >
-            {
-              <DataSourceDropDownHeader
-                totalDataSourceCount={1}
-                application={this.props.application}
-              />
-            }
-          </EuiPanel>
-          <EuiPanel color="subdued" paddingSize="xs" borderRadius="none">
+        <DataSourceDropDownHeader totalDataSourceCount={1} application={this.props.application} />
```
zhongnansu, line 196: I saw in the `DataSourceSelectable`, `DataSourceDropDownHeader ` is wrapped by `EuiContextMenuPanel `, do we need the same for data source view?

BionIT (dismissed):

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
zhyuanqi: Thank you for bringing this up. I did a close look at it and i think the Drop down header should be outside of Panel as it is using the EuiPopoverTitle in the back

zhongnansu (approved):

BionIT (approved):

Issue #6330, opened 2024-04-03
[Workspace] Add dev tool to top right navigation


raintygao:
**Is your feature request related to a problem? Please describe.**

As one of new features of workspace, navigation will be refactored. As a progressive update, we will first add the dev tool entrance to the navigation in the top right header.

**Describe the solution you'd like**
1. Update chrome service and define functions and components internally to register top right navigation for applications to call.
2. Register top right navigation in dev tool app.

**Additional context**

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4944

raintygao:
We create this issue to let core team know this change in advance, hoping to reduce gap before code review and discuss if needed. cc @ashwin-pc @kavilla @lauralexis @kgcreative .

SuZhou-Joe:
cc @joshuarrrr as well.

PR #6553, opened 2024-04-19
feat: Add `rightNavigationButton` component in chrome service for applications to register and add dev tool to top right navigation

raintygao:
### Description

1. Update chrome service and define functions and components internally to register top right navigation for applications to call.
2. Register top right navigation in dev tool app.

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6330

## Screenshot

image
image

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/core/public/chrome/nav_controls/nav_controls_service.ts
```diff
@@ -104,7 +116,17 @@ export class NavControlsService {
...
           new Set([...navControlsExpandedCenter$.value.values(), navControl])
         ),
-
+      registerRightNavigation: (props: RightNavigationProps) => {
+        const nav = {
+          order: props.order,
+          mount: mountReactNode(
+            React.createElement(RightNavigationButton, {
+              ...props,
+            })
```
SuZhou-Joe, line 125:
```suggestion
            React.createElement(RightNavigationButton, props)
```
Seems the spread is not needed here.

src/core/public/chrome/nav_controls/nav_controls_service.ts
```diff
@@ -104,7 +116,17 @@ export class NavControlsService {
         navControlsExpandedCenter$.next(
           new Set([...navControlsExpandedCenter$.value.values(), navControl])
         ),
-
+      registerRightNavigation: (props: RightNavigationProps) => {
+        const nav = {
+          order: props.order,
```
SuZhou-Joe, line 121: Do we need to give a warning here if user specify an existing order?

src/core/public/chrome/ui/header/right_navigation_button.tsx
```diff
@@ -0,0 +1,46 @@
...
+}: RightNavigationButtonProps) => {
+  const navigateToApp = () => {
+    const appUrl = application.getUrlForApp(appId, {
+      path: '/',
+      absolute: false,
+    });
+    // Remove prefix in Url including workspace and other prefix
+    const targetUrl = http.basePath.prepend(http.basePath.remove(appUrl), {
+      withoutClientBasePath: true,
+    });
```
SuZhou-Joe, line 33: I am thinking if we just use `navigateToApp` here, the method will detect if the app is accessible inside workspace and use the correct prepend params when navigate.

src/plugins/dev_tools/public/plugin.ts
```diff
@@ -125,10 +127,19 @@ export class DevToolsPlugin implements Plugin<DevToolsSetup> {
...
   }
 
-  public start() {
+  public start(core: CoreStart) {
     if (this.getSortedDevTools().length === 0) {
       this.appStateUpdater.next(() => ({ navLinkStatus: AppNavLinkStatus.hidden }));
     }
+    core.chrome.navControls.registerRightNavigation({
+      // order of dev tool should be after advance settings
+      order: 2,
```
SuZhou-Joe, line 136: Maybe we should put the orders into a common constant file so that other maintainers are easy to tell who sits before who.

src/plugins/dev_tools/public/plugin.ts
```diff
@@ -125,10 +127,19 @@ export class DevToolsPlugin implements Plugin<DevToolsSetup> {
...
+      appId: this.id,
+      http: core.http,
+      application: core.application,
```
SuZhou-Joe, line 139: A little bit weird why should plugins explicitly pass the props: `http` and `application` here. Should be accessible within RightNavigation component I suppose.

CHANGELOG.md
```diff
@@ -105,6 +105,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [Workspace] Hide datasource and advanced settings menu in dashboard management when in workspace. ([#6455](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6455))
 - [Multiple Datasource] Modify selectable picker to remove group label and close popover after selection ([#6515](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6515))
 - [Workspace] Add workspaces filter to saved objects page. ([#6458](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6458))
+- Add dev tool to top right navigation. ([#6553](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6553))
```
SuZhou-Joe, line 108:
```suggestion
- Introduce `registerRightNavigation` in chrome service to reigster standard top right navigation and add dev tool to top right navigation. ([#6553](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6553))
```
Better to include the changes inside chrome_service.

src/plugins/dev_tools/public/plugin.ts
raintygao: Would it be easy for component to get core dependency?

SuZhou-Joe: I am thinking what if RightNavigation requires more props from core, should all the registeration adjust accordingly?

raintygao: But this need to set osdReact as dependency? I think it may not be good to set core module depend on this.

SuZhou-Joe: What about passing the coreStart to the component inside chrome_service instead of asking plugins to pass in such params?

src/core/public/chrome/ui/header/right_navigation_button.tsx
raintygao: Here we use `application.navigateToUrl`. Does this also be covered?

src/plugins/dev_tools/public/plugin.ts
raintygao: Sure. That makes sense.

src/core/public/chrome/ui/header/right_navigation_button.tsx
SuZhou-Joe: Nope, navigateToUrl won't detect app related info.

ashwin-pc:
How is this different from `registerRight`?

raintygao:
> How is this different from `registerRight`?

Top right navigation has common component style and navigate function. This function will maintain this in chrome.

ashwin-pc:
A few comments:

1. Can you add docs for this?
2. Can you add a screenshot for how this will work in mobile? We want collapse this like the others in mobile view too
3. Each app can chose to register this nav option right? E.g. not every app may want dev tools available 
4. The API looks more complex than it needs to be. All the props I expect this function to take are : label, type, priority/position, onClick, icon. Then on click the app is free to whatever it wants.

raintygao:
> A few comments:
> 
> 1. Can you add docs for this?
> 2. Can you add a screenshot for how this will work in mobile? We want collapse this like the others in mobile view too
> 3. Each app can chose to register this nav option right? E.g. not every app may want dev tools available
> 4. The API looks more complex than it needs to be. All the props I expect this function to take are : label, type, priority/position, onClick, icon. Then on click the app is free to whatever it wants.

1. Sure.
2. Sure. Normally this will look like the same as `registerRight`.
3. Each app can call this function to register top right nav in its hook. This is only determined by the app itself not other apps. 
4. What's the type of `icon`? I think one of reasons that we add this function is to maintain this common component in chrome.  If we don't do this, we will maintain a separate circle icon component in each app with the same style.

src/core/public/chrome/ui/header/right_navigation_button.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+      },
+    };
+    const { getByTestId } = render(<RightNavigationButton {...props} />);
+    const icon = getByTestId('rightNavigationButton');
+    fireEvent.click(icon);
+    expect(getUrlForApp).toHaveBeenCalledWith('app_id', {
+      path: '/',
+      absolute: false,
+    });
+    expect(navigateToUrl).toHaveBeenCalled();
```
wanglam, line 43: Shall we expect called with specific URL instead of getUrllForApp?

src/core/public/chrome/nav_controls/nav_controls_service.ts
```diff
@@ -104,7 +123,19 @@ export class NavControlsService {
...
+          mount: mountReactNode(
+            React.createElement(RightNavigationButton, {
+              ...props,
+              http,
+              application,
+            })
+          ),
+        };
+        return navControlsRight$.next(new Set([...navControlsRight$.value.values(), nav]));
```
SuZhou-Joe, line 137: Nit: why don't we just call the `registerRight` function? The registerRight logic may change in the future.

src/core/public/chrome/ui/header/right_navigation_button.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export const RightNavigationButton = ({
+  application,
+  http,
+  appId,
+  iconType,
+  title,
+}: RightNavigationButtonProps) => {
+  const navigateToApp = () => {
```
SuZhou-Joe, line 26: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/chrome/ui/header/nav_link.tsx#L79

I think we keep the same logic here as left navigation.

1. When clicking, it will use SPA routing to jump to the target App.
2. When clicking combined with `metaKey`(command in Mac), it will open a new browser tab to load the app.

src/core/public/chrome/ui/header/right_navigation_button.test.tsx
raintygao: Logic of generating has been moved to useMemo.

raintygao: Yes, logic of generating has been moved to useMemo.

src/core/public/chrome/nav_controls/nav_controls_service.ts
raintygao: Updated.

src/core/public/chrome/ui/header/right_navigation_button.tsx
raintygao: Updated.

ashwin-pc:
Ah thanks for the context. Yeah Icon was for simplifying the API. My main suggestion though was to extend the registerRight API rather than introduce a new one. How you wish to do that I'm open to suggestions :)

src/core/public/chrome/ui/header/right_navigation_button.tsx
SuZhou-Joe: Oh, yes. navigateToUrl will parse the app info from the url. Forget it.

SuZhou-Joe:
> extend the registerRight API.

1. Actually for now `registerRight` takes only two required fields: order and mount, which is clean and generic for all kinds of needs to render a customized React Component into top right. It may pollute the clean registerRight API if we extend the API to add more navigation-specific fields.
2. The changes to add top right navigations may be controlled by a `feature flag`, only by using a new API can we enable top right navigation in a centralized place.

SuZhou-Joe:
@joshuarrrr Would you like to raise some concerns as you have more knowledge on this API?

ashwin-pc:
>  It may pollute the concise registerRight API if we extend the API to add more navigation-specific fields.

If the existing API to register an element to the top right does not do what it should, doesnt it make sense to extend it? If it does what you want it to do, then why do we even need this change? e.g. if i have a button component, but that button component does not let me disable it, do i extend the button to also be able to disable it, or do i create a new disableButton component?

SuZhou-Joe:
> do i extend the button to also be able to disable it, or do i create a new disableButton component?

That's a very interesting topic and for me in this case I prefer to extend the props to add a `disabled` field because it sounds more like a generic props. But actually there is a `OuiIconButton` component in OUI, instead of introducing an icon props in `OuiButton`. I suppose the reason is that Icon is not a generic props for Button so OUI creates a new component for that.

Anyway could you please take a look on the second reason that by using a new API we can introduce potential feature flag control over it?

Overall, I think it also makes sense that we make `RightNavigationButton` a common Component from chrome_service and let plugins to use registerRight to register the top right navigation by leveraging the `RightNavigationButton` component, but in this case we may not be able to give feature flag control on the all new-added top right navs.

cc @raintygao

raintygao:
Extending registerRight will make this API unclean because there are too many navigation related props, we decide to export rightNavigationButton from chrome.

src/plugins/dev_tools/public/plugin.ts
```diff
@@ -125,10 +134,22 @@ export class DevToolsPlugin implements Plugin<DevToolsSetup> {
...
     }
+    core.chrome.navControls.registerRight({
+      order: RightNavigationOrder.DevTool,
+      mount: toMountPoint(
+        React.createElement(RightNavigationButton, {
+          appId: this.id,
+          iconType: 'consoleApp',
+          title: this.title,
+          application: core.application,
+          http: core.http,
```
SuZhou-Joe, line 149: Nit: perhaps `CoreStart` will be better for backward-capability.

SuZhou-Joe (approved):

src/core/public/chrome/constants.ts
```diff
@@ -31,3 +31,9 @@
 export const OPENSEARCH_DASHBOARDS_ASK_OPENSEARCH_LINK = 'https://forum.opensearch.org/';
 export const GITHUB_CREATE_ISSUE_LINK =
   'https://github.com/opensearch-project/OpenSearch-Dashboards/issues/new/choose';
+
+export enum RightNavigationOrder {
+  // order of dev tool should be after advance settings
+  Settings = 1,
+  DevTool = 2,
```
Hailong-am, line 38: nit: the value could be 10, 20 and leave some gaps between them so that we can add more between settings and devtool without change the value.

raintygao: That makes sense. Updated.

src/core/public/chrome/ui/header/right_navigation_button.tsx
```diff
@@ -0,0 +1,59 @@
...
+    // Remove prefix in Url including workspace and other prefix
+    return http.basePath.prepend(http.basePath.remove(appUrl), {
+      withoutClientBasePath: true,
+    });
+  }, [application, http.basePath, appId]);
+
+  const navigateToApp = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+    /* Use href and onClick to support "open in new tab" and SPA navigation in the same link */
+    if (
+      event.button === 0 && // ignore everything but left clicks
```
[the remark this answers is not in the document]
raintygao: Updated.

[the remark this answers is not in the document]
raintygao: Seems your guys approvals were dismissed because of commits to address comments. Could you help add 2.x label and approve again?

ruanyl (approved):

SuZhou-Joe (approved):

Issue #5211, opened 2023-10-04
[Datasources] Address Remaining Items in Datasource Selector


mengweieric:
**Is your feature request related to a problem? Please describe.**

Due to time constraints, some review comments have not been addressed. This issue is for tracking those remaining tasks need to be addressed in the PR #5167 .

**Describe the solution you'd like**

Addressing all reviews after the merge, including: 
<s> - add / improve interfaces and types </s>
<s>- code movement of datasource registration from discover to data plugin </s>
- Adjustments for datasource API definitions
  - removing multiple selectable API
  - add type to datasource and remove hardcoding type name mapping
<s>  - using i18n for user facing text </s>
- Unit / E2E test coverage

anirudha:
lets target this at 2.12 or sooner

anirudha:
In the next release we should move this to the view-switcher asap

mengweieric:
Spoke with @ashwin-pc on this issue that we will:

- Remove this issue from 2.12 release
- Add tests to datasource code lines that still need coverages
- Make change to datasource interface addressing item 'Adjustments for datasource API definitions' after 2.12 when we'd settled new datasource interface designs

Issue #4991, opened 2023-09-12
[Design] Data Explorer : Discover 2.0 + Event Explorer MDS cohesion


mengweieric:
## Overview

Flint and several other projects are broadening OpenSearch's capabilities, enabling it to not only support consuming OpenSearch's internal data but also data from external data sources like S3, CloudWatch and Prometheus. Once these external data sources are appropriately configured, they will be made available for selection from both data explorer and other plugins including Dashboards Observability. This document aims to discuss the datasource selector as a new user interface for shared datasource selection experience in OpenSearch Dashboards.

### Functional

* **Datasource selector user interface:** A newly designed datasource selector UI is required. 
    * It should not only be centered around providing a smooth interaction for selecting datasources but also an intuitive transition experience for user who are accustomed to use index-pattern selector in Discover for selections of index patterns. 
    * It should retrieve all datasources from different datasource instances, and consolidate them in datasource selector
    * It should allow enabling/disabling of multi-selection capabilities of datasources
* **Dynamic context switching:** The new datasource selector should have the ability and flexibility to dynamically switch to the context of the selected data source, including
    * **Load relevant datasource service:** Once a datasource is selected, the system should load the corresponding datasource service associated with this datasource.
    * **Load relevant datasource instance:** In addition to the service, the system should also load the relevant datasource instance. A datasource instance refers to an initialized instance that encapsulates metadata, along with a suite of APIs and services supported by the current datasource.
    * **Load relevant UI components:** It should also dynamically load the UI components corresponding to the chosen data source. 
        * Observability
            * **Query Editor**: A tool for editing and executing ad-hoc queries. This component should be tailored to the selected data source.
            * **View Switcher**: Not all views are available for all datasources, switching datasources will also switch available views. Note that when Observability specific datasource is selected, available view switcher will list only observability even and visualizations views. 
            * **Visualization configuration panel**: Visualization setting options change according to different datasources.

### Non-functional

* **Reusability and Extensibility of Datasource Service:** The design and implementation of the datasource selector, datasource service and their related components should emphasize flexibility, reusability and extensibility, enabling it to be seamlessly integrated with any other plugins or modules of OpenSearch Dashboards.
* **Self Contained:** The datasource selection components should be structured as a standalone, self-sufficient plugin within the OpenSearch Dashboards whenever it can to allow them to function independently, without reliance on any other plugins or features outside of the OSD scope.

### Prototype Overview - Datasource selector

[...]

PR #6571, opened 2024-04-22
[Discover] Data selector enhancement

mengweieric:
### Description

Discover data selector enhancements includes:

- Add a few missing i18n
- Add id, improve data source metadata with fields addition for data sources
- Improve data source filtering
- Add in-memory cache for data selector list with refresh button
- Improve data selector rendering logics
- Removed hard coded data source type-to-display mapping from selector and read from metadata
- Increased test coverage
- Types and interfaces improvements

### Issues Resolved

- https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4991
- https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5211

## Screenshot
Screenshot 2024-04-29 at 12 03 46 PM

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


src/plugins/data/public/data_sources/register_default_datasource.ts
```diff
@@ -4,23 +4,32 @@
...
     dataSourceFactory.getDataSourceInstance(DEFAULT_DATASOURCE_TYPE, {
+      id: htmlIdGenerator('local-cluster')('indices'),
       name: DEFAULT_DATASOURCE_NAME,
       type: DEFAULT_DATASOURCE_TYPE,
-      metadata: null,
+      metadata: {
+        ui: {
+          label: 'Index patterns', // display name of your data source,
+          typeLabel: 'Index patterns', // display name of your data source type,
+          typeGroup: 'DEFAULT_INDEX_PATTERNS',
```
kavilla, line 27: i think this can be an enum

mengweieric: Thanks for early review, thinking about the same, will update it.

kavilla (dismissed):
Really like these changes. Just have some nits but I don't see any reason to block this unless CI fails.

changelogs/fragments/6571.yml
```diff
@@ -0,0 +1,2 @@
+refactor:
+- discover data selector enhancement and refactoring ([#6571](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6571))
```
kavilla, line 2: Can we verify it's just Discover? I think it's any thing that uses it the data source selector right but just doesn't pass it.

mengweieric: For now it's just discover, but since we have already aligned on having data selector as its name to distinguish from MDS selector I think we can remove 'discover' as it can be used outside of discover.

src/plugins/data/public/data_sources/constants.tsx
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { DataSourceUIGroupType } from './datasource/types';
+import { DEFAULT_DATA_SOURCE_DISPLAY_NAME } from './register_default_datasource';
+
+export const defaultDataSourceMetadata = {
```
kavilla, line 9: qq: did we settle on this code wise being internally referenced as data source?

changelogs/fragments/6571.yml
kavilla: gotcha.

but also if some 3rd party plugin referenced this selector technically it would be modified as well right?

kavilla:
one more question, is this configurable with a ui setting? or config? could be an easy win to let it be a ui setting

src/plugins/data/public/data_sources/constants.tsx
mengweieric: In last Thursday's meeting @kgcreative mentioned we should for now only aligns on data connector in replace of previous data source selector naming, but not for renaming other existing datasource names through out the code for this. Soon will have a push for renaming stuff only for datasource selector to be data selector.

src/plugins/data/public/data_sources/datasource_selector/datasource_selectable.tsx
```diff
@@ -162,6 +165,24 @@ export const DataSourceSelectable = ({
       onChange={handleSourceChange}
       singleSelection={singleSelection}
       isClearable={false}
+      append={
+        <EuiText size="s">
```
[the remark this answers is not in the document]
mengweieric: Will add this with a push soon.

changelogs/fragments/6571.yml
mengweieric: yea that is correct.

mengweieric:
> one more question, is this configurable with a ui setting? or config? could be an easy win to let it be a ui setting

for now is not configurable through ui setting, but I believe what you are pointing to is the ideal way we want to go with in near releases.

src/plugins/data/public/data_sources/datasource/types.ts
```diff
@@ -3,49 +3,79 @@
...
-export interface IDataSourceMetaData {
+export interface IDataSourceGroup {
   name: string;
 }
 
-export interface IDataSourceGroup {
+export interface DataSetWithDataSource<T = []> {
+  ds: DataSource;
+  list: T;
+}
```
joshuali925, line 24: maybe
```suggestion
export interface DataSetWithDataSource<T = unknown> {
  ds: DataSource;
  list: T[];
}
```

src/plugins/data/public/data_sources/datasource/types.ts
```diff
@@ -3,49 +3,79 @@
...
+
+export interface IDataSetParams<T = {}> {
+  query: T;
+}
+
+export interface IDataSourceQueryParams<T = {}> {
+  query: T;
+}
+
+export interface IDataSourceQueryResult<T = {}> {
```
joshuali925, line 34: why are the generics defaulting to `{}`?

src/plugins/data/public/data_sources/datasource/types.ts
```diff
@@ -3,49 +3,79 @@
...
+  data: T;
+}
+
+export interface DataSourceConnectionStatus {
+  status: string;
```
joshuali925, line 39: can the status be an arbitrary string or can the type be more narrowed down?

src/plugins/data/public/data_sources/datasource/types.ts
```diff
@@ -3,49 +3,79 @@
...
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface IDataSourceQueryParams {}
+export interface IDataSourceUISettings {
+  selector: IDataSourceUISelector;
+  label: string; // the display name of data source
+  groupType: DataSourceUIGroupType; // the group to which the data source belongs
+  typeLabel: string; // the display name of data source type
+  displayOrder?: number; // the order in which the data source should be displayed in selector
+  description?: string; // short description of your database
+  icon?: string; // uri of the icon
```
joshuali925, line 66: i think tsdoc/jsdoc has `@member` or `/** xxx */` to document fields of interface, might be better so typescript language server would be aware of it

i'm not sure if `// xxx` works, if it does then ignore

src/plugins/data/public/data_sources/datasource/types.ts
mengweieric: For these interfaces basically there's no concrete implementations needed at this point so just providing a default object so elsewhere it doesn't need to specify a type which it doesn't really need.

changelogs/fragments/6571.yml
kavilla: sounds good to me! thanks!

kavilla (approved):
Link is not failing because of this PR. All good to me

CHANGELOG.md
```diff
@@ -108,6 +108,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - [Workspace] Add workspaces filter to saved objects page. ([#6458](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6458))
 - [Multiple Datasource] Support multi data source in Region map ([#6654](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6654))
 - Add `rightNavigationButton` component in chrome service for applications to register and add dev tool to top right navigation. ([#6553](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6553))
+- [Discover Data Selector] - Data selector enhancement and refactoring ([#6571](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6571))
```
ashwin-pc, line 111: We dont need this in two places. You can delete this.

src/plugins/data/public/data_sources/constants.tsx
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { i18n } from '@osd/i18n';
+import { DataSourceUIGroupType } from './datasource/types';
+
+export const S3_GLUE_DATA_SOURCE_DISPLAY_NAME = 'Amazon S3';
```
ashwin-pc, line 9: Are we still hard coding these? Is there a plan to remove them?

kavilla (dismissed):

src/plugins/data/public/data_sources/datasource_services/datasource_service.ts
```diff
@@ -50,21 +56,24 @@ export class DataSourceService {
    * @returns A registration result indicating success or failure.
    * @throws {DataSourceRegistrationError} Throws an error if a data source with the same name already exists.
    */
-  async registerDataSource(ds: GenericDataSource): Promise<IDataSourceRegistrationResult> {
-    const dsName = ds.getName();
-    if (dsName in this.dataSources) {
+  async registerDataSource(ds: DataSource): Promise<IDataSourceRegistrationResult> {
+    const dsId = ds.getId();
```
ashwin-pc, line 60: Nit: dont use abbreviated variable names. Always expand them.

src/plugins/data/public/data_sources/datasource/types.ts
```diff
@@ -3,49 +3,112 @@
...
+
+export interface IDataSourceDataSet<T = {}> {
+  dataSets: T;
+}
+
+export interface IDataSourceQueryResponse<T = {}> {
+  data: T;
+}
+
+export enum DataSourceUIGroupType {
```
ashwin-pc, line 110: Same as above, do we have a plan on getting away from hardcoding datasource types in this service? These should be generic enough to allow different groups of datasources.

ashwin-pc (commented):
Can you outline in the PR description what exactly this change is about? Currently it looks like this addresses all the feedback from the previous PR and closes the Data Sources issue. However i see in here that S3 and Spark data sources are still hard coded and the registration function is still not generic enough to handle future data sources.

mengweieric:
> Can you outline in the PR description what exactly this change is about? Currently it looks like this addresses all the feedback from the previous PR and closes the Data Sources issue. However i see in here that S3 and Spark data sources are still hard coded and the registration function is still not generic enough to handle future data sources.

Thanks for the review.

The removal of hardcoding in this PR is to remove previous hardcoded type-to-display-name mapping and instead reading from data source metadata. The s3 data sources are backend data sources where it doesn't have any UI specific definitions. Ideally these front end specific definitions should be retrieving from settings or front end data sources separated from backend data sources where user can configure them. Does this align with the direction you are pointing to? 

As for registration function, are you point to the registerDataSources in data source service or the way data source owner register the data source?

kavilla (approved):

kavilla:
> > Can you outline in the PR description what exactly this change is about? Currently it looks like this addresses all the feedback from the previous PR and closes the Data Sources issue. However i see in here that S3 and Spark data sources are still hard coded and the registration function is still not generic enough to handle future data sources.
> 
> Thanks for the review.
> 
> The removal of hardcoding in this PR is to remove previous hardcoded type-to-display-name mapping and instead reading from data source metadata. The s3 data sources are backend data sources where it doesn't have any UI specific definitions. Ideally these front end specific definitions should be retrieving from settings or front end data sources separated from backend data sources where user can configure them. Does this align with the direction you are pointing to?
> 
> As for registration function, are you point to the registerDataSources in data source service or the way data source owner register the data source?

I think @ashwin-pc is saying the next iteration of this where if I am a cluster admin, I can install a plugin that adds the new data source. For example, if I want a S3 connection type I install a S3 connection plugin that wires that up. Or better yet, some type of catalogue.

Flyingliuhub (approved):

mengweieric:
> > > Can you outline in the PR description what exactly this change is about? Currently it looks like this addresses all the feedback from the previous PR and closes the Data Sources issue. However i see in here that S3 and Spark data sources are still hard coded and the registration function is still not generic enough to handle future data sources.
> > 
> > 
> > Thanks for the review.
> > The removal of hardcoding in this PR is to remove previous hardcoded type-to-display-name mapping and instead reading from data source metadata. The s3 data sources are backend data sources where it doesn't have any UI specific definitions. Ideally these front end specific definitions should be retrieving from settings or front end data sources separated from backend data sources where user can configure them. Does this align with the direction you are pointing to?
> > As for registration function, are you point to the registerDataSources in data source service or the way data source owner register the data source?
> 
> I think @ashwin-pc is saying the next iteration of this where if I am a cluster admin, I can install a plugin that adds the new data source. For example, if I want a S3 connection type I install a S3 connection plugin that wires that up. Or better yet, some type of catalogue.

Yea understood that we will want this in future iterations, this PR addressed the step one mapping (updated the description), eventually we will not want those to be defined in code.

ashwin-pc:
@mengweieric yes my goal is to have no service specific names hardcoded in the framework. If any other company wanted to extend the datasources to support their version of the service like Azure for example. They should not need to make a code change in OSD to show up as a new Datasource as long as they adhere to the schema.

e.g. 

```
registerDataSource({
  name: 'Postgress`
  schema: any
  ...
})
```

today, if i wanted to add postgress or any other datasource type to this list, i will need to make a code change in OSD to show up as an option in the DataSources menu. That should not be necessary. That should all be abstracted away in the API.

PR #6358, opened 2024-04-05
[Multiple DataSource] DataSourceSelectable support to render label by getting it from dataSourceOptions

yujin-emma:
### Description

Today the datasource menu does not populate the label of the datasource based on the datasourceId. This can be confusing to users, who may not know what human-readable datasource they are looking at.

This PR is going to resolve: when only given the dataSourceId to `DataSourceSelectable`, the component can get the label from existing `dataSourceIdToLabel` map

## Testing the changes

test cases: 

1. this is invalid case, when there is no activeOption pass in and hide local cluster, there should be a toast error

activeOption={[{id: ''}]}
hideLocalCluster={true} // need to set in the `opensearc_dashboards.yml`

empty-activeOption-hide-lc-warning

2. there are ds options, even not pass in activeOptions, display the default options

        hideLocalCluster={true}

dsOpts-hidelc-display-default

please also refer to case 9

3. do not need to take care empty array, since we do not have placeholder for DataSourceSelectable

activeOption={[]}
hideLocalCluster={true} //does not matter
// not filter out default

we do not need this case for DataSourceSelectable

4. show toast, when pass invalid dataSourceId in activeOption

  selectedOption={[{id: 'invalid'}]}
        hideLocalCluster={false}

invalid-activeOption-toast

5. show local cluster when pass empty string for activeOption and not hide local cluster

selectedOption={[{id: ''}]}
hideLocalCluster={false}

not-hide-lc-activeOption-empty-string-show-lc

6. filter out default, display local cluster
```
 dataSourceFilter={(ds) => ds.attributes.title !== 'test1'}
        hideLocalCluster={false}
```
filter-default-show-lc

7. check when all data sources are deleted, uisetting throw error and no any datasource 

hideLocalCluster={true}

8. when there is no activeOption pass in, and there is default dataSourceOption, show default dataSourceOption and show checked status on the memu

no-active-option-show-default

9. when there is activeOption pass in, this is the valid dataSourceOption, display it with the label it fetch from dataSourceOptions, and show checked status in the menu list
valid-active-option-show-itself

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #6594, opened 2024-04-22
[Multiple Datasource] Pass data source menu to top nav via function instead of using component

BionIT:
### Description
In this change, we changed to pass data source menu to top nav via function instead of using the component, this way we avoid coupling the plugin level props into the top nav menu component

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/b0090572-06ec-4580-bef9-ca599ee02019

## Testing the changes
1. mount data source list all view along with the topnav menu, will see the component mounted, default data source is shown, no local cluster is showing since hide local cluster setting is true
2. mount data source list active view along with the topnav menu, will see the component mounted, default data source is shown , no local cluster is showing since hide local cluster setting is true
3. mount data source single readonly along with the topnav menu, will see the component mounted, default data source is shown , no local cluster is showing since hide local cluster setting is true
4. mount data source single selectable along with the topnav menu, will see the component mounted, default data source is shown , no local cluster is showing since hide local cluster setting is true

Screenshot when hide local cluster is true:
Screenshot 2024-04-22 at 11 53 53 AM

Screenshot when hide local cluster is false:
Screenshot 2024-04-22 at 11 53 30 AM

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source_management/public/components/data_source_menu/create_data_source_menu.test.tsx
+++ b/src/plugins/data_source_management/public/components/data_source_menu/create_data_source_menu.test.tsx
@@ -14,7 +14,6 @@ import React from 'react';
 import { act, render } from '@testing-library/react';
 import { DataSourceComponentType, DataSourceSelectableConfig } from './types';
 import { ReactWrapper } from 'enzyme';
-import { mockDataSourcePluginSetupWithShowLocalCluster } from '../../mocks';
 import * as utils from '../utils';
 
 describe('create data source menu', () => {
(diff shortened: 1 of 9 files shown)


Flyingliuhub (approved):

src/plugins/data_source_management/public/components/data_source_menu/create_data_source_menu.test.tsx
```diff
@@ -71,10 +71,10 @@ describe('create data source menu', () => {
         notifications,
       },
     };
-    const TestComponent = createDataSourceMenu<DataSourceSelectableConfig>(
-      uiSettings,
-      mockDataSourcePluginSetupWithShowLocalCluster
-    );
+    spyOn(utils, 'getApplication').and.returnValue({ id: 'test2' });
+    spyOn(utils, 'getUiSettings').and.returnValue(uiSettings);
+    spyOn(utils, 'getHideLocalCluster').and.returnValue({ enabled: true });
```
zhongnansu, line 76: should we set this to `enabled: false` based on the test case description?

BionIT: @zhongnansu This test itself needs to be fixed since it doesn't change disregards of local cluster is set since it doesn't check the options from popover. We can cut another meta issue for all the tests that we are missing testing, or do you think this needs to block the change?

zhongnansu: @BionIT agreed that we should fix the text, this is non blocking issue
I think in this PR, the test was modified to remove the check for local cluster existence
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6358/files#diff-ff98db435bf1f4c1bfe7ea1869b5d25b7e731148ce6060333bf37b0dee33219fL78

zhongnansu (approved):

PR #6602, opened 2024-04-22
[MDS] Expose common url key for consumers

### LINKED ISSUE #6534: Export a common URL key for plugins to check for datasource
Currently, the selected datasource is not a global concept. This means that each plugin has to manage its own selected datasource, which can be done either via session/local storage, internally within their react app, or via URL, which many plugins seem to be doing. This issue is to propose datasource or datasource management plugin to export a common URL key to use, such as: https://github.com/opensearch-project/security-dashboards-plugin/blob/feature/supporting-multiple-datasources/public/utils/datasource-utils.ts#L22. This key could then be used by plugins to set/read the datasource via the URL, and could help with cohesion and maintaining datasource context when navigating between plugins. 

Alternatives: key for session/local storage, but this would not be able to be link shared between users/bookmarked, so I think URL is better.

huyaboo:
### Description

This PR adds a common URL key for plugins/other consumers to consume.

### Issues Resolved

Closes #6534 

## Changelog
- fix: [MDS] Expose common url key to query for datasource 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source_management/public/plugin.ts
+++ b/src/plugins/data_source_management/public/plugin.ts
@@ -17,7 +17,12 @@ import {
   IAuthenticationMethodRegistry,
   AuthenticationMethodRegistry,
 } from './auth_registry';
-import { noAuthCredentialAuthMethod, sigV4AuthMethod, usernamePasswordAuthMethod } from './types';
+import {
+  DATA_SOURCE_URL_KEY,
+  noAuthCredentialAuthMethod,
+  sigV4AuthMethod,
+  usernamePasswordAuthMethod,
+} from './types';
 import { DataSourceSelectorProps } from './components/data_source_selector/data_source_selector';
 import { createDataSourceMenu } from './components/data_source_menu/create_data_source_menu';
 import { DataSourceMenuProps } from './components/data_source_menu';
@@ -42,6 +47,7 @@ export interface DataSourceManagementPluginSetup {
   ui: {
     DataSourceSelector: React.ComponentType<DataSourceSelectorProps>;
     getDataSourceMenu: <T>() => React.ComponentType<DataSourceMenuProps<T>>;
+    dataSourceURLKey: string;
   };
   dataSourceSelection: DataSourceSelectionService;
   getDefaultDataSourceId: typeof getDefaultDataSourceId;
@@ -126,6 +132,7 @@ export class DataSourceManagementPlugin
       ui: {
         DataSourceSelector: createDataSourceSelector(uiSettings, dataSource),
         getDataSourceMenu: <T>() => createDataSourceMenu<T>(),
+        dataSourceURLKey: DATA_SOURCE_URL_KEY,
       },
       getDefaultDataSourceId,
       getDefaultDataSourceId$,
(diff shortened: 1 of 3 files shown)


jackiehanyang:
Can we hold off on this one and include it in 2.15, given that we're close to the 2.14 code freeze date? It might look like a small change, but it requires the plugin team to go through the development cycle again to adopt this change

huyaboo:
> Can we hold off on this one and include it in 2.15, given that we're close to the 2.14 code freeze date? It might look like a small change, but it requires the plugin team to go through the development cycle again to adopt this change

Acknowledged. Moved to `2.15`

src/plugins/data_source_management/public/plugin.ts
```diff
@@ -110,6 +116,7 @@ export class DataSourceManagementPlugin
       ui: {
         DataSourceSelector: createDataSourceSelector(uiSettings, dataSource),
         getDataSourceMenu: <T>() => createDataSourceMenu<T>(),
+        dataSourceURLKey: DATA_SOURCE_URL_KEY,
```
zhyuanqi, line 119: Hi Huy. Could you help me to understand what does this  dataSourceURLKey do and how does it help plugin team?

huyaboo: Before, plugins had to manage how they wanted to persist the selected datasource across their pages. Most opted for the datasource to be a part of the url since it makes the page bookmark-able and easier than drilling down props. However, this urlkey `&dataSource=` is not global, meaning different plugins can have different urlkeys. This PR will expose a common key `dataSource` for them to consume to increase consistency among plugin urls.

zhyuanqi: Got it. Thanks for explaining it.

zhyuanqi (approved):

src/plugins/data_source_management/public/plugin.ts
```diff
@@ -34,6 +39,7 @@ export interface DataSourceManagementPluginSetup {
   ui: {
     DataSourceSelector: React.ComponentType<DataSourceSelectorProps>;
     getDataSourceMenu: <T>() => React.ComponentType<DataSourceMenuProps<T>>;
+    dataSourceURLKey: string;
```
wanglam, line 42: Shall we use `typeof DATA_SOURCE_URL_KEY` here?

zhongnansu (commented):
Nit: do we have developer documentation for plugins/consumers about how to use this key?

src/plugins/data_source_management/public/plugin.ts
BionIT: @huyaboo +1, can we use type instead of exposing it like this?

LDrago27:
@huyaboo Is this still being targeted for 2.16.

huyaboo:
> @huyaboo Is this still being targeted for 2.16.

@LDrago27 No this is pushed back to 2.17

sejli:
Hey @huyaboo, is this still targeted for `2.17`?

Issue #6581, opened 2024-04-22
[BUG] data source connection link in Index pattern table breaks


zhongnansu:
**Describe the bug**

data source connection link in Index pattern table is not working as expected. It fails to redirect to the corresponding data source detail page, instead it goes to the dashboards management landing page

**Screenshots**

![iShot_2024-04-21_23 18 23](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/32652829/6320cdba-8f4b-472a-907b-e8c3ad57c6e2)

zhongnansu:
I'll work on this

The issue is with the link url. Currently it's
```
/app/management/opensearch-dashboards/dataSourceManagement/<data source id>
```
it should be 

```
/app/management/opensearch-dashboards/dataSourceManagement/<data source id>
```
`dataSourceManagement` is the data source management plugin id

cc: @BionIT

seraphjiang:
> The issue is with the link url. Currently it's
> 
> ```
> /app/management/opensearch-dashboards/dataSourceManagement/<data source id>
> ```
> 
> it should be
> 
> ```
> /app/management/opensearch-dashboards/dataSourceManagement/<data source id>
> ```
> 
> `dataSourceManagement` is the data source management plugin id

Hi @zhongnansu the actual and expected Url seems identical, would you check

zhongnansu:
@seraphjiang sorry for typo, currently it's
```
/app/management/opensearch-dashboards/dataSources/<data source id>
```
E.g: https://future.playground.opensearch.org/app/management/opensearch-dashboards/dataSources/93d6c680-dcbd-11ee-987c-f557473dd2d5

zhongnansu:
fixed by PR ##6604, closing

Issue #6582, opened 2024-04-22
[MD]data source link from index pattern page should redirect without reloading


zhongnansu:
Since OpenSearch Dashboards is a SPA(single page application), data source links from index pattern page should redirect without triggering a reload.

To reproduce:
1. Go to index pattern management page in OpenSearch playground: https://playground.opensearch.org/app/management/opensearch-dashboards/indexPatterns
2. click any data source connections 

![iShot_2024-04-22_00 53 34](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/32652829/bc77fa86-6748-4f57-95ea-5d5ef632dd82)

PR #6604, opened 2024-04-23
[MD]Fix data source link to point to correct page, and use in page redirect instead of reloading

zhongnansu:
### Description

[MD]Fix data source link to point to correct page, and use in page redirect instead of reloading

### Issues Resolved

#6581 
#6582 

## Screenshot
iShot_2024-04-22_23 45 41

## Testing the changes

1. create a data source and create index pattern with this data source
2. go to index pattern page, and click the data source name link under column "Data source connection", it should redirect to the corresponding data source detail page, without triggering a page reload.
3. Test the change works for dropdown header, and no data source component, since global var `DSM_APP_ID` is being used at those places. See below gif

iShot_2024-04-23_15 57 54

## Changelog

- fix: Fix data source link to point to correct page, and use in page redirect instead of reloading

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source_management/public/components/data_source_column/data_source_column.tsx
+++ b/src/plugins/data_source_management/public/components/data_source_column/data_source_column.tsx
@@ -4,18 +4,15 @@
  */
 
 import { i18n } from '@osd/i18n';
-import { HttpStart, SavedObjectsStart } from 'opensearch-dashboards/public';
+import { SavedObjectsStart } from 'opensearch-dashboards/public';
 import { EuiBadge, EuiLink } from '@elastic/eui';
 import React from 'react';
-import {
-  IndexPatternTableColumn,
-  IndexPatternTableRecord,
-} from '../../../../index_pattern_management/public';
-import { getDataSources } from '../utils';
+import { IndexPatternTableColumn } from '../../../../index_pattern_management/public';
+import { getApplication, getDataSources } from '../utils';
 import { DataSourceTableItem } from '../../types';
+import { DSM_APP_ID } from '../../plugin';
 
-type DataSourceColumnItem = DataSourceTableItem & { relativeUrl: string };
-type DataSourceMap = Map<string, DataSourceColumnItem> | undefined;
+type DataSourceMap = Map<string, DataSourceTableItem> | undefined;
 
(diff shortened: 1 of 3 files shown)


src/plugins/data_source_management/public/components/data_source_column/data_source_column.tsx
```diff
@@ -4,18 +4,15 @@
...
+import { IndexPatternTableColumn } from '../../../../index_pattern_management/public';
+import { getApplication, getDataSources } from '../utils';
 import { DataSourceTableItem } from '../../types';
+import { DSM_APP_ID } from '../../plugin';
 
-type DataSourceColumnItem = DataSourceTableItem & { relativeUrl: string };
-type DataSourceMap = Map<string, DataSourceColumnItem> | undefined;
+type DataSourceMap = Map<string, DataSourceTableItem> | undefined;
 
 export class DataSourceColumn implements IndexPatternTableColumn<DataSourceMap> {
```
BionIT, line 17: do we have test for this class?

src/plugins/data_source_management/public/plugin.ts
```diff
@@ -41,7 +41,7 @@ export interface DataSourceManagementPluginStart {
   getAuthenticationMethodRegistry: () => IAuthenticationMethodRegistry;
 }
 
-export const DSM_APP_ID = 'dataSourceManagement';
+export const DSM_APP_ID = 'dataSources';
```
BionIT, line 44: Looks like we are changing the global constant which is used in multiple places, can we tests if all the places using the constant are working as expected?

zhongnansu: yes, it's also used in dropdown header, and no data source component. Added video to the PR description -> testing section

src/plugins/data_source_management/public/components/data_source_column/data_source_column.tsx
zhongnansu: Currently we don't, let me add it

zhongnansu:
Failed CI are not related to the change of this PR

bandinib-amzn (approved):

PR #6610, opened 2024-04-23
[Multiple Datasource]DataSourceView add switch to default data source button 

yujin-emma:
### Description

Add switch to default data source when pass in invalid data source id for DataSourceView 
We expect to see
When pass in invalid data source id
| |  default data source is available | no default data source |
|------|-------|-------|
| hide local cluster | show switch button and provide the switch to default data source functionality on both popover and toast | no switch button display either on popover or toast |
| not hide local cluster | show switch button and provide the switch to default data source functionality on both popover and toast | no switch button display either on popover or toast |

## Testing the changes

Happy case:
when pass in valid id without label, the DataSourceView can display the correct label
happy-case

Happy case but the result got filter out, we treat it as invalid id error:
```
 activeOption: [{ id: '829c0ee0-065f-11ef-ad27-4377edb9ad8f' }],
          dataSourceFilter: (ds) => {
            return ds.id !== '829c0ee0-065f-11ef-ad27-4377edb9ad8f'
          },
```
happy-case-but-filterout

Overall Test Result:

When pass in invalid data source id
| |  default data source is available | no default data source |
|------|-------|-------|
| hide local cluster |hide-lc-default-show-button |Screenshot 2024-04-23 at 14 07 48Screenshot 2024-04-23 at 14 07 41 |
| not hide local cluster | both-show-button | Screenshot 2024-04-23 at 14 07 41Screenshot 2024-04-23 at 14 07 48 |

test input:

```
<DataSourceMenu
        setMenuMountPoint={setActionMenu}
        componentType={'DataSourceView'}
        componentConfig={{
          notifications,
          savedObjects: savedObjects.client,
          fullWidth: false,
          activeOption: [{ id: 'invalid-id' }],
          dataSourceFilter: (ds) => {
            return true;
          },
          onSelectedDataSources: (ds) => {
            setSelectedDataSources(ds);
          },
        }}
      />
    );

```

#### 1. When hide local cluster, and there is no default data source

Screenshot 2024-04-23 at 14 07 48

Screenshot 2024-04-23 at 14 07 41

#### 2. When hide local cluster, and there is default data source

hide-lc-default-show-button

#### 3. Test mobile view 
Screenshot 2024-04-23 at 11 43 21

#### 4. Test iphone SE View 
Screenshot 2024-04-23 at 12 00 46

##### 5.1.  When no default data source - both the popover and toast should not display the button to switch the default data source
Screenshot 2024-04-23 at 14 07 41

Screenshot 2024-04-23 at 14 07 48

##### 5.2. When there is default data source - both the popover and toast should display the switch button to switch to the default data source

both-show-button

#### 6. If only pass in valid data source id, the DataSourceView will render the correct data source title, if click the header button, the popover will show up and only checked with the activeOption, there is no error toast, no switch to default button

Test input from examples:

```
<DataSourceMenu
        setMenuMountPoint={setActionMenu}
        componentType={'DataSourceView'}
        componentConfig={{
          notifications,
          savedObjects: savedObjects.client,
          fullWidth: false,
          activeOption: [{ id: '98652370-01b8-11ef-9df2-d1bcc041cdef' }], // this is basically the test2 data source id
          dataSourceFilter: (ds) => {
            return true;
          },
          onSelectedDataSources: (ds) => {
            setSelectedDataSources(ds);
          },
        }}
      />

```

Screenshot 2024-04-23 at 14 30 36

## Changelog
 - skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ZilongX:
5 ci group failed somehow, re-running now

BionIT:
> When hide local cluster, and there is no default data source
What's the input for the data source component? Also can we show the popover?

BionIT:
> When hide local cluster, and there is default data source

The toast message should change too and can you double check the requirement?

BionIT:
Can we add test for when invalid datasource id gets passed in and there is default data source, we should the right popover with toast and be able to switch to use default data source?

yujin-emma:
> Can we add test for when invalid datasource id gets passed in and there is default data source, we should the right popover with toast and be able to switch to use default data source?

Test 5.2 cover this case

yujin-emma:
> > When hide local cluster, and there is default data source
> 
> The toast message should change too and can you double check the requirement?



> > When hide local cluster, and there is default data source
> 
> The toast message should change too and can you double check the requirement?

Test 2 cover this

yujin-emma:
> > When hide local cluster, and there is no default data source
> 
> What's the input for the data source component? Also can we show the popover?

Thanks for pointing it out, already updated in the descriptions with test set up

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -110,10 +116,15 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
           this.props.onSelectedDataSources([{ id: optionId, label: selectedDataSource.title }]);
         }
       } catch (error) {
-        handleDataSourceFetchError(
+        if (defaultDataSource) {
```
BionIT, line 119: we also need to check if the default data source has been filtered out or not

src/plugins/data_source_management/public/components/data_source_view/data_source_view_error.tsx
```diff
@@ -0,0 +1,98 @@
...
+      })}
+    </EuiButton>
+  );
+
+  const iconButton = (
+    <EuiButtonIcon
+      className="euiHeaderLink"
+      data-test-subj="dataSourceViewErrorHeaderLink"
+      aria-label={i18n.translate('dataSourceError.dataSourceErrorHeaderLink', {
+        defaultMessage: 'dataSourceErrorMenuHeaderLink',
```
BionIT, line 56: can we have more consistent names, className, data-test-subj, aria-label are all different

src/plugins/data_source_management/public/components/data_source_view/data_source_view_error.tsx
```diff
@@ -0,0 +1,98 @@
...
+      iconType={() => <ErrorIcon />}
+      size="s"
+      onClick={() => setShowPopover(!showPopover)}
+    />
+  );
+
+  return (
+    <>
+      <EuiPopover
+        id={'dataSourceErrorPopover'}
```
BionIT, line 67: can we have more specific id?

src/plugins/data_source_management/public/components/data_source_view/data_source_view_error.tsx
```diff
@@ -0,0 +1,98 @@
...
+        panelPaddingSize="none"
+        anchorPosition="downLeft"
+        data-test-subj={'dataSourceViewErrorPopover'}
+      >
+        <DataSourceDropDownHeader totalDataSourceCount={0} application={application} />
+        <EuiPanel
+          hasBorder={false}
+          hasShadow={false}
+          className="dataSourceEmptyStatePanel"
+          data-test-subj="datasourceTableEmptyState"
```
BionIT, line 80: same here

src/plugins/data_source_management/public/components/data_source_view/data_source_view_error.tsx
```diff
@@ -0,0 +1,98 @@
...
+        >
+          <EuiText size="s" textAlign="center">
+            {i18n.translate('dataSourcesManagement.dataSourceErrorMenu.text', {
+              defaultMessage: `Data source ${dataSourceId.toString()} is not available`,
```
BionIT, line 84: do we need to call toString() if data source id is already a string type

src/plugins/data_source_management/public/components/toast_button/switch_default_button.tsx
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { EuiButton } from '@elastic/eui';
+import React from 'react';
+import { i18n } from '@osd/i18n';
+
+export const switchDefaultButton = (handleSwitchDefaultDatasource: () => Promise<void>) => {
```
BionIT, line 10: switchToDefaultButton is better

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -303,6 +304,23 @@ export const handleDataSourceFetchError = (
...
+  failedDataSourceId: string,
+  defaultDataSource: string | null,
+  callback?: () => Promise<void>
+) => {
+  changeState({ showError: true });
+  notifications.add({
+    title: i18n.translate('dataSource.dataSourceUnavailableError', {
+      defaultMessage: `Data source ${failedDataSourceId} is not available `,
+    }),
+    text: defaultDataSource && callback ? toMountPoint(switchDefaultButton(callback)) : '',
```
BionIT, line 319: is it possible that callback is undefined or null?

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
yujin-emma: Sure, will test this case

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -110,10 +116,15 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
+          this.setState({ defaultDataSource });
+        }
```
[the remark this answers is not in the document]
yujin-emma: Refer to `handleDataSourceFetchError`, this function pass `this.props.onSelectedDataSources` as callback, and set [] when there is no selectedOption and hide local cluster

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -136,10 +147,39 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
   }
+  /**
+   * when call handleSwitchDefaultDatasource, the default data source must exist,
+   * since the button only display when defaultDataSource exist
+   */
+  async handleSwitchDefaultDatasource() {
+    /**
+     * default dataSource label must exist when defaultDataSource exist
+     */
+    const defaultDataSourceId = this.state.defaultDataSource;
```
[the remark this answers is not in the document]
yujin-emma: Well, I am fine wether put the defaultDataSource as a parameter to handleSwitchDefaultDatasource or extract here,  the defaultDataSource here only used to construct the selectedOption

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -136,10 +147,39 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
   closePopover() {
     this.setState({ ...this.state, isPopoverOpen: false });
   }
+  /**
+   * when call handleSwitchDefaultDatasource, the default data source must exist,
+   * since the button only display when defaultDataSource exist
+   */
+  async handleSwitchDefaultDatasource() {
```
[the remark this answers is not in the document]
yujin-emma: Sure, will add test about this

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -136,10 +147,39 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
+    const defaultDataSourceId = this.state.defaultDataSource;
+    const defaultDataSourceLabel = (
+      await getDataSourceById(defaultDataSourceId!, this.props.savedObjectsClient!)
+    ).title;
+
+    const defaultDataSourceOption = { id: defaultDataSourceId!, label: defaultDataSourceLabel };
+    // reset the state to close popover and error, selectedOption will be replaced by default option
+    this.setState({
+      selectedOption: [defaultDataSourceOption],
+      showError: false,
```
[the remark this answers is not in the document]
yujin-emma: The error popover display since `showError` is turn to true, if not turn it back, it will keep true and popover keep open

src/plugins/data_source_management/public/components/data_source_view/data_source_view_error.tsx
yujin-emma: Sure, will update

src/plugins/data_source_management/public/components/data_source_view/data_source_view_error.tsx
yujin-emma: Sure, will update

src/plugins/data_source_management/public/components/data_source_view/data_source_view_error.tsx
yujin-emma: Sure, will remove it

src/plugins/data_source_management/public/components/toast_button/switch_default_button.tsx
yujin-emma: Sure, will update

src/plugins/data_source_management/public/components/utils.ts
yujin-emma: Possible to have the callback undefined or null when not pass it. The callback here is set to optional referring to `handleDataSourceFetchError`, defaultDataSource can be null since we might not have it set up from uiSettings. The case here is: when default data source exist, the callback has to exist, when no default data source, there is no need for the callback. 
I agree to set it mandatory and only use defaultDataSource as condition check

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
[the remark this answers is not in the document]
yujin-emma: Synced offline, the `handleSwitchDefaultDatasource` is defined out of the `componentDidMount`, passing the defaultDataSource can be a option, it also make sense to get it from state directly

src/plugins/data_source_management/public/components/utils.ts
yujin-emma: synced offline, the defaultDataSource should be the only condition check

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -110,10 +111,28 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
           this.props.onSelectedDataSources([{ id: optionId, label: selectedDataSource.title }]);
         }
       } catch (error) {
-        handleDataSourceFetchError(
+        // since we do not set the defaultDataSource when hit error, we need to set it before moving forward
+        if (defaultDataSource) {
+          const { dataSourceFilter } = this.props;
```
zhyuanqi, line 116: Is it possible to write 
```
   if (checkIfDefaultDataSourceStillExistsAfterFiltered) {
         this.setState({
               defaultDataSource: filteredDefaultDataSource,
          });
   }
```

yujin-emma: sure,I will update it

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -136,10 +155,40 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
+  /**
+   * when call handleSwitchDefaultDatasource, the default data source must exist,
+   * since the button only display when defaultDataSource exist
+   */
+  async handleSwitchDefaultDatasource(defaultDataSourceId: string | null) {
+    /**
+     * default dataSource label must exist when defaultDataSource exist
+     */
+    const defaultDataSourceLabel = (
+      await getDataSourceById(defaultDataSourceId!, this.props.savedObjectsClient!)
```
zhyuanqi, line 167: What if this getDataSourceById failed?

yujin-emma: nice catch, while basically, we assume the defaultDataSource is a valid data source object, we can always get it. If exception happens, we are already in `error` state, the popover and toast will keep open until we manually close it or the switch work with valid defaultDataSource option.

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -303,6 +304,23 @@ export const handleDataSourceFetchError = (
   });
 };
 
+export const handleDataSourceViewError = (
```
huyaboo, line 307: So the naming can be a bit confusing: `handleDataSourceViewError()` creates the notification, the `DataSourceViewError` component renders the popover, and `this.handleSwitchDefaultDatasource()` sets the state of the `DataSourceViewError`. Could `handleDataSourceViewError()` at least be renamed to something like `renderDataSourceViewErrorNotification()` to make it distinct from the other two?

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -303,6 +304,23 @@ export const handleDataSourceFetchError = (
...
+  changeState: (state: { showError: boolean }) => void,
+  notifications: ToastsStart,
+  failedDataSourceId: string,
+  defaultDataSource: string | null,
+  callback: () => Promise<void>
```
huyaboo, line 312: Since we are defining a lot of props here, could we extract it into its own props class?

yujin-emma: Good idea, I will update, thanks for the suggestion

src/plugins/data_source_management/public/components/utils.ts
yujin-emma: You are right, let me rethink the naming here, DataSourceView got something different from other component, naming like this might add the complexity to use it, thanks for pointing it out

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
zhyuanqi: Okay. So if there is a internet connection issue, there still is a possibilities that getDataSourcById would fail. Maybe we should have this getDataSourcById  in the componentDidMount try block to catch the error if it fails. Not a blocking issue.

yujin-emma: I thinking wrapping another api call in catch block will be very, we can either set the default data source label when we set it in the uiSettings, or, call getDataSourceById in the first try to get it label, then we can handle either error when getDataSource for the activeOption or defaultDataSource, to have this, I  also recommend maintaining another state call defaultDataSourceOption as DataSourceOption `{id: defaultDataSourceId, label: getDataSourceById(defaultDataSourceId).title}`

yujin-emma: For refactoring uiSettings to have defaultDataSource as `DataSourceOption`, I think better to have a separate issue to track, and we need to make sure all the other components' behavior will not be impacted by this change. 
The benefit of this change can add the request call for defaultDataSource out of any dataSource component since it is in uiSettings
While at another side, not all component/plugin use defaultDataSource, and not all of them will need to use default data source as the data source option

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -136,10 +152,49 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
+    const filteredDataSource = dataSourceFilter
+      ? [{ id: defaultDataSource }].filter(dataSourceFilter)
+      : [{ id: defaultDataSource }];
+    return filteredDataSource.length === 0 ? null : filteredDataSource[0].id;
+  }
+  /**
+   * when call handleSwitchDefaultDatasource, the default data source must exist,
+   * since the button only display when defaultDataSource exist
+   */
+  async handleSwitchDefaultDatasource(defaultDataSourceId: string | null) {
```
yujin-emma, line 168: This function can take null and might cause error when calling getDataSourceById when the input id is null, however, when we use this function, the input dataSourceId always exist as string

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -111,11 +116,22 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
-          this.onError.bind(this),
-          this.props.notifications!,
-          this.props.onSelectedDataSources
-        );
+        // return [] to customer when no default data source
+        if (!defaultDataSource) {
+          this.props.onSelectedDataSources?.([]);
+        }
+        this.setState({
+          defaultDataSource: this.getFilteredDefaultDataSource(defaultDataSource),
```
BionIT, line 124: why do we need to set default data source here?

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -303,6 +304,42 @@ export const handleDataSourceFetchError = (
...
+  dataSourceViewErrorWithDefaultParams: DataSourceViewErrorWithDefaultParams
+) => {
+  const {
+    changeState,
+    notifications,
+    failedDataSourceId,
+    defaultDataSource,
+    callback,
+  } = dataSourceViewErrorWithDefaultParams;
+  changeState({ showError: true });
```
BionIT, line 333: the input {showError: true } does nothing, can we remove? @zhyuanqi found this before

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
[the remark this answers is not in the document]
BionIT: popover has nothing to do with this state, it is controlled by isPopoveropen prop

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -110,10 +111,28 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
+        if (defaultDataSource) {
+          const { dataSourceFilter } = this.props;
+          // check if the defaultDataSouce can be filtered out
+          const filteredDataSource = dataSourceFilter
+            ? [{ id: defaultDataSource }].filter(dataSourceFilter)
+            : [{ id: defaultDataSource }];
+          const filteredDefaultDataSource =
+            filteredDataSource.length === 0 ? null : filteredDataSource[0].id;
+          this.setState({
+            defaultDataSource: filteredDefaultDataSource,
```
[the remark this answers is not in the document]
yujin-emma: When hit error, we get the defaultDataSource from uiSettings, but never set it to local state, if we want to use it out of the componentDidMount, we cannot use it due to the block scope, therefore, I check if it exist and set it to the local state, so that we can use it in other handlers

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
yujin-emma: Because when we hit the error state, we never save the defaultDataSource to some local state, functions out side of componentDidMount cannot use it

src/plugins/data_source_management/public/components/utils.ts
yujin-emma: I tried to remove showError, the Button to open popover has a wired behavior, the button is first no error state, after one click, the database icon is with the red dot, then click, the popover display
remove-showError-default

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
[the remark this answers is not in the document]
yujin-emma: Oh, the reason have the showError here is when we choose to use defaultDataSource to replace the activeOption, there is no more error, we need to turn it back from `error: true`

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -111,11 +116,22 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
+          this.props.onSelectedDataSources?.([]);
+        }
+        this.setState({
+          defaultDataSource: this.getFilteredDefaultDataSource(defaultDataSource),
+        });
+        const handleDataSourceViewErrorParams = {
+          changeState: this.onError.bind(this),
+          notifications: this.props.notifications!,
+          failedDataSourceId: optionId,
+          defaultDataSource: this.state.defaultDataSource,
```
BionIT, line 130: so here we are setting the default data source at line 124, then has to wait until it is set then get the value? Can we extract the this.getFilteredDefaultDataSource(defaultDataSource) as const?

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -111,11 +116,22 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
         }
       } catch (error) {
-        handleDataSourceFetchError(
-          this.onError.bind(this),
-          this.props.notifications!,
-          this.props.onSelectedDataSources
-        );
+        // return [] to customer when no default data source
+        if (!defaultDataSource) {
+          this.props.onSelectedDataSources?.([]);
```
BionIT, line 121: when there is error to get the data source, do we return empty array and when do we return something else?

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -137,10 +153,49 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
+    ).title;
+
+    const defaultDataSourceOption = { id: defaultDataSourceId!, label: defaultDataSourceLabel };
+    // reset the state to close popover and error, selectedOption will be replaced by default option
+    this.setState({
+      selectedOption: [defaultDataSourceOption],
+      showError: false,
+      isPopoverOpen: false,
+    });
+    this.props.onSelectedDataSources!([defaultDataSourceOption]);
```
BionIT, line 184: looks like we are enforcing an optional prop here, is it expected? can we add a test that when onSelectedDataSources is not provided, it will not break?

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
yujin-emma: Updated

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
yujin-emma: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6610#discussion_r1576973918, should also check if the local cluster is hide or not

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
yujin-emma: Thanks for pointing it out, when tested, the UI not broken, while the console throw error when onSelectedDataSources is not provided, already updated, will add this in test plan

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -137,10 +154,51 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
   closePopover() {
     this.setState({ ...this.state, isPopoverOpen: false });
   }
+  getFilteredDefaultDataSource(defaultDataSource: string | null): string | null {
```
zhyuanqi, line 157: nit: Need extra line here between closePopover and getFilteredDefaultDataSource.

zhyuanqi (approved):

huyaboo (approved):

zhongnansu (dismissed):

zhongnansu:
@yujin-emma Great job, the PR description is very detailed

BionIT:
Failed codecov, do we have enough tests? Follow https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6610/checks?check_run_id=24317184661 to add tests

zhongnansu (approved):

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -111,11 +116,23 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
+        const filteredDefaultDataSource = this.getFilteredDefaultDataSource(defaultDataSource);
+        this.setState({
+          defaultDataSource: filteredDefaultDataSource,
+        });
+        const handleDataSourceViewErrorParams = {
+          changeState: this.onError.bind(this),
+          notifications: this.props.notifications!,
+          failedDataSourceId: optionId,
+          defaultDataSource: filteredDefaultDataSource,
+          callback: this.handleSwitchDefaultDatasource.bind(this, this.state.defaultDataSource),
```
BionIT, line 132: can we use filteredDefaultDataSource here?

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -111,11 +116,23 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
-          this.props.notifications!,
-          this.props.onSelectedDataSources
-        );
+        // return [] to customer when no default data source
+        if (!defaultDataSource && this.props.hideLocalCluster) {
+          this.props.onSelectedDataSources?.([]);
+        }
+        const filteredDefaultDataSource = this.getFilteredDefaultDataSource(defaultDataSource);
+        this.setState({
+          defaultDataSource: filteredDefaultDataSource,
```
BionIT, line 125: can we avoid setstate multiple times in one block?

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -137,10 +154,51 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
   closePopover() {
     this.setState({ ...this.state, isPopoverOpen: false });
   }
+  getFilteredDefaultDataSource(defaultDataSource: string | null): string | null {
+    if (!defaultDataSource) return defaultDataSource;
```
BionIT, line 158:
```suggestion
    if (!defaultDataSource) return null;
    
```

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -137,10 +154,51 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
+    const { dataSourceFilter } = this.props;
+    // check if the defaultDataSouce can be filtered out
+    const filteredDataSource = dataSourceFilter
+      ? [{ id: defaultDataSource }].filter(dataSourceFilter)
```
BionIT, line 162: data source filter received a data source object, we should pass in the complete data source object fetched from saved object   `dataSourceFilter?: (dataSource: any) => boolean;`

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
yujin-emma: already updated

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
yujin-emma: updated

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
yujin-emma: updated

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -57,77 +63,147 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
+    this.setState({
+      ...this.state,
+      defaultDataSource,
+    });
+    this.props.onSelectedDataSources?.([dataSourceOption]);
+  }
+
+  handleLegitLocalCluster(defaultDataSource: string | null) {
+    this.setState({
+      selectedOption: [LocalCluster],
```
xinruiba, line 86: Do we need to also inherit other state?

```
    this.setState({
      ...this.state,
      selectedOption: [LocalCluster],
      defaultDataSource,
    });
```

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -57,77 +63,140 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
+    if (filteredSelectedDataSourceOption) {
+      this.setState({
+        selectedOption: [filteredSelectedDataSourceOption],
+        defaultDataSource,
+      });
+      return;
+    }
+
+    await this.getAndSetDefaultDataSourceOption(dataSourceId, defaultDataSource);
+    this.handleInvalidDataSourceError(dataSourceId);
```
BionIT, line 132: are we calling this.handleInvalidDataSourceError twice if there is exception in this.getAndSetDefaultDataSourceOption

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
```diff
@@ -57,77 +63,140 @@ export class DataSourceView extends React.Component<DataSourceViewProps, DataSou
...
-      if (this.props.onSelectedDataSources) {
-        this.props.onSelectedDataSources([]);
+    // 2. handle local cluster
+    if (option && optionId === '') {
+      // 2.1 handle legit
+      if (!this.props.hideLocalCluster) {
+        this.updateSelectedOptionState(defaultDataSource, LocalCluster);
+        return;
       }
+      this.clearSelectedOptionState();
```
BionIT, line 168: should this be same as handle invalid data source?

yujin-emma: what clearSelectedOptionState does:
1. clear the state for selectedOption
2. pass empty array to `onSelectedDataSources` if it exist
what handle invalid data source does:
1. show error state
2. set defaultDataSourceOption
3. show notification
4. pass empty array to `onSelectedDataSources` if it exist
5. show switch to default option if the defaultOption exist
the only overlap is `pass empty array to `onSelectedDataSources` if it exist`, therefore, I think they are different

src/plugins/data_source_management/public/components/data_source_view/data_source_view.tsx
yujin-emma: Woops, it did call twice then, let me update the getAndSetDefaultDataSourceOption to only return error not call handleInvalidDataSourceError

LDrago27:
@yujin-emma Is this still being targeted for 2.16.

Issue #6517, opened 2024-04-17
Validation for Endpoint URL when creating a dataSource


riysaxen-amzn:
**DataSource Endpoint Validation**
Presently, there is no validation for the Endpoint URL during the creation of a dataSource in Dashboards Management. Consequently, it is possible to create a dataSource with an invalid Endpoint URL, as shown below.

![Image 4-17-24 at 3 16 PM](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/69919272/fdcb64fa-56c6-4e82-8d98-406857a4ad11)

Despite repeated attempts to test the connection, error messages related to authentication issues persist, masking the actual problem associated with the Endpoint URL. Also, I was able to create a dataSource with an invalid Endpoint

![Image 4-17-24 at 3 33 PM](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/69919272/26896190-01c4-4dc7-bd35-9b97a887f874)

**Describe the solution you'd like**

Implementing DataSource Endpoint Validation would significantly improve the user experience. By incorporating this validation, error messages can accurately reflect the cause of failure during dataSource creation, leading to more efficient issue resolution.

**Describe alternatives you've considered**

**Additional context**

BionIT:
Thanks @riysaxen-amzn to report this issue! @zhongnansu could you help to work on this and this issue in functional test repo is somewhat related https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/1202

zhongnansu:
> Thanks @riysaxen-amzn to report this issue! @zhongnansu could you help to work on this and this issue in functional test repo is somewhat related [opensearch-project/opensearch-dashboards-functional-test#1202](https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/1202)

Sure, I'll work on this. What I plan to do is to add the same logic of "test connection", to the create data source validation.

PR #6631, opened 2024-04-24
[MD] Add endpoint validation to create data source API

zhongnansu:
### Description

 Add endpoint validation to create data source API, by testing if endpoint is reachable, using the same logic as what's behind "test connection" feature

### Issues Resolved

#6517 

## Screenshot

iShot_2024-04-25_15 02 52

## Testing the changes

1. go to create datasource page, and create one with invalid OpenSearch endpoint, e.g. google.com
2. click test connection, it should fail as expected
3. click create data source, it should fail too, as the endpoint is not a valid endpoint
4. switch to use a valid opensearch domain
5. both test connection and create data source should pass

## Changelog

- feat: Add endpoint validation to create data source API

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source/server/plugin.ts
+++ b/src/plugins/data_source/server/plugin.ts
@@ -61,16 +61,26 @@ export class DataSourcePlugin implements Plugin<DataSourcePluginSetup, DataSourc
     const cryptographyServiceSetup: CryptographyServiceSetup = this.cryptographyService.setup(
       config
     );
+    const dataSourceServiceSetup: DataSourceServiceSetup = await this.dataSourceService.setup(
+      config
+    );
 
     const authRegistryPromise = core.getStartServices().then(([, , selfStart]) => {
       const dataSourcePluginStart = selfStart as DataSourcePluginStart;
       return dataSourcePluginStart.getAuthenticationMethodRegistry();
     });
+    const auditTrailPromise = core.getStartServices().then(([coreStart]) => coreStart.auditTrail);
+    const customApiSchemaRegistryPromise = core.getStartServices().then(([, , selfStart]) => {
+      const dataSourcePluginStart = selfStart as DataSourcePluginStart;
+      return dataSourcePluginStart.getCustomApiSchemaRegistry();
+    });
 
     const dataSourceSavedObjectsClientWrapper = new DataSourceSavedObjectsClientWrapper(
+      dataSourceServiceSetup,
       cryptographyServiceSetup,
       this.logger.get('data-source-saved-objects-client-wrapper-factory'),
       authRegistryPromise,
+      customApiSchemaRegistryPromise,
       config.endpointDeniedIPs
     );
 
@@ -104,20 +114,12 @@ export class DataSourcePlugin implements Plugin<DataSourcePluginSetup, DataSourc
       },
     };
     core.auditTrail.register(auditorFactory);
-    const auditTrailPromise = core.getStartServices().then(([coreStart]) => coreStart.auditTrail);
-
-    const dataSourceService: DataSourceServiceSetup = await this.dataSourceService.setup(config);
-
-    const customApiSchemaRegistryPromise = core.getStartServices().then(([, , selfStart]) => {
-      const dataSourcePluginStart = selfStart as DataSourcePluginStart;
-      return dataSourcePluginStart.getCustomApiSchemaRegistry();
-    });
 
     // Register data source plugin context to route handler context
     core.http.registerRouteHandlerContext(
       'dataSource',
       this.createDataSourceRouteHandlerContext(
-        dataSourceService,
+        dataSourceServiceSetup,
         cryptographyServiceSetup,
         this.logger,
         auditTrailPromise,
@@ -129,14 +131,14 @@ export class DataSourcePlugin implements Plugin<DataSourcePluginSetup, DataSourc
     const router = core.http.createRouter();
     registerTestConnectionRoute(
       router,
-      dataSourceService,
+      dataSourceServiceSetup,
       cryptographyServiceSetup,
       authRegistryPromise,
       customApiSchemaRegistryPromise
     );
     registerFetchDataSourceMetaDataRoute(
       router,
-      dataSourceService,
+      dataSourceServiceSetup,
       cryptographyServiceSetup,
       authRegistryPromise,
       customApiSchemaRegistryPromise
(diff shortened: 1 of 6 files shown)


BionIT (dismissed):

BionIT:
Probably better to remove the credentials or modify the recording to hide it

src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
```diff
@@ -252,26 +258,56 @@ export class DataSourceSavedObjectsClientWrapper {
...
+  }
+
+  private validateTitle(title: string) {
+    if (!title.trim().length) {
       throw SavedObjectsErrorHelpers.createBadRequestError(
         '"title" attribute must be a non-empty string'
       );
     }
 
+    if (title.length > 32) {
```
huyaboo, line 274: nit: could we make this a constant with a meaningful name?

src/plugins/data_source/server/plugin.ts
```diff
@@ -61,16 +61,24 @@ export class DataSourcePlugin implements Plugin<DataSourcePluginSetup, DataSourc
     const cryptographyServiceSetup: CryptographyServiceSetup = this.cryptographyService.setup(
       config
     );
+    const dataSourceService: DataSourceServiceSetup = await this.dataSourceService.setup(config);
```
zhyuanqi, line 64: Should we name it dataSourceServiceSetUp as cryptographyServiceSetup

huyaboo (approved):
Left nonblocking comment

src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
```diff
@@ -252,26 +258,56 @@ export class DataSourceSavedObjectsClientWrapper {
...
+        `endpoint is not valid OpenSearch endpoint`
+      );
+    }
   }
 
   private async validateAuth<T = unknown>(auth: T) {
+    if (!auth) {
+      throw SavedObjectsErrorHelpers.createBadRequestError('"auth" attribute is required');
+    }
+
```
zhyuanqi, line 310: So from now on, we cannot choose no auth?

zhyuanqi:
Just a general thought. I see the error is 'Creation of the Data Source failed with some errors.'. Not sure if we need to give more specifc error in different error cases such as `no auth found, please provide authentication`

src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
zhongnansu: creating data source with `no auth ` auth type still has an an attribute in the request

image

src/plugins/data_source/server/plugin.ts
zhongnansu: fixed

src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
zhongnansu: updated to use constant

zhongnansu:
> Just a general thought. I see the error is 'Creation of the Data Source failed with some errors.'. Not sure if we need to give more specifc error in different error cases such as `no auth found, please provide authentication`

The error message is the existing behavior when failing to create data source , with/without my change,  I agree we should improve by displaying the error with details to user. It could be a separate task to work on, and it should take care of both test connection and create data source

xinruiba:
I remember datasource is able to be created even test connection failed used to be the expected behavior.
So we decide to change it?

zhongnansu:
> I remember datasource is able to be created even test connection failed used to be the expected behavior. We decide to change it?

AFAIK, yes, is there any concern? @xinruiba

BionIT (approved):

xinruiba:
> > I remember datasource is able to be created even test connection failed used to be the expected behavior. We decide to change it?
> 
> AFAIK, yes, is there any concern? @xinruiba

No concerns if that's by design~

src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
```diff
@@ -252,26 +259,56 @@ export class DataSourceSavedObjectsClientWrapper {
...
       );
     }
+    try {
+      const dataSourceClient: OpenSearchClient = await this.dataSourcesService.getDataSourceClient({
+        savedObjects: {} as any,
+        cryptography: this.cryptography,
+        testClientDataSourceAttr: attributes as DataSourceAttributes,
+        authRegistry: await this.authRegistryPromise,
+        customApiSchemaRegistryPromise: this.customApiSchemaRegistryPromise,
+      });
```
xinruiba, line 295: Thanks for this change.

Are there any chances we can reuse this function: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source_management/public/components/utils.ts#L206?

Since datasource and datasource manager plugin will both exist in the same time, so the validation path in datasource manager plugin should always been routed. 

Then we depend on sending request to `/internal/data-source-management/validate` to test connection instead of creating a new datasource client.

What do you think?

zhongnansu: Do you mean when user click create data source button, we first call `/internal/data-source-management/validate` to validate and then call create data source API? @xinruiba

xinruiba: Yes

Pros:

- We rely on the URL for data source validation.

Cons:

- If customers directly create a data source by calling the URL, the data source validation will be bypassed.

I'll leave the decision to you~

xinruiba: Sounds good to me. Approving~

xinruiba (approved):

src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
xinruiba: Non blocking comment.

Maybe a question to designer. Since we already do validation while datasource creation. Do we sill need a separate test connection?

zhongnansu: good point, @kgcreative @xeniatup, I wonder what's your thought on this

kgcreative: Given the updated validation, i don't think we need it in the create page.

PR #6668, opened 2024-04-29
[Workspace] Fix: keep disallowed types when importing with overwrite

SuZhou-Joe:
### Description

When importing data sources and configuration into a workspace with overwrite: true, we will prevent the creation of such objects. However, to ensure that objects dependent on the disallowed ones continue to function properly, we must leave their references untouched.

#### How the changes can keep the references untouched.
When user tries to import with `overwrite: true`, the logic will go into https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/saved_objects/import/check_conflicts.ts#L90, and in order to keep the references untouched, we just need to make sure the logic won't go into the branch of https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/saved_objects/import/check_conflicts.ts#L105 , which sets a new id reference to the current item. And the result it is consumed comes from the `savedObjectsClient.checkConflicts()`, which can be wrapped by savedObjectsClientWrapper mechanism to intercept special logic on specific objects.

## Testing the changes

There is a integration test case, named `checkConflicts when importing disallowed types` inside `src/plugins/workspace/server/saved_objects/integration_tests/saved_objects_wrapper_for_check_workspace_conflict.test.ts` to cover this case.

## Changelog

- fix: keep disallowed types when importing with overwrite

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/server/saved_objects/saved_objects_wrapper_for_check_workspace_conflict.ts
+++ b/src/plugins/workspace/server/saved_objects/saved_objects_wrapper_for_check_workspace_conflict.ts
@@ -289,11 +289,33 @@ export class WorkspaceConflictSavedObjectsClientWrapper {
         return { errors: [] };
       }
 
+      const { workspaces } = options;
+
+      const disallowedSavedObjects: SavedObjectsCheckConflictsObject[] = [];
+      const allowedSavedObjects: SavedObjectsCheckConflictsObject[] = [];
+      objects.forEach((item) => {
+        const isImportIntoWorkspace = workspaces?.length;
+        // config can not be created inside a workspace
+        if (this.isConfigType(item.type) && isImportIntoWorkspace) {
+          disallowedSavedObjects.push(item);
+          return;
+        }
+
+        // For 2.14, data source can only be created without workspace info
+        if (this.isDataSourceType(item.type) && isImportIntoWorkspace) {
+          disallowedSavedObjects.push(item);
+          return;
+        }
+
+        allowedSavedObjects.push(item);
+        return;
+      });
+
       /**
        * Workspace conflict only happens when target workspaces params present.
        */
       if (options.workspaces) {
-        const bulkGetDocs: any[] = objects.map((object) => {
+        const bulkGetDocs: any[] = allowedSavedObjects.map((object) => {
           const { type, id } = object;
 
           return {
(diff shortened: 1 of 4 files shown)


ruanyl:
> we must leave their references untouched.

@SuZhou-Joe could you help me to understand how this PR addressed the above issue?

src/plugins/workspace/server/saved_objects/saved_objects_wrapper_for_check_workspace_conflict.ts
```diff
@@ -289,11 +289,33 @@ export class WorkspaceConflictSavedObjectsClientWrapper {
         return { errors: [] };
       }
 
+      const { workspaces } = options;
+
+      const disallowedSavedObjects: SavedObjectsCheckConflictsObject[] = [];
+      const allowedSavedObjects: SavedObjectsCheckConflictsObject[] = [];
+      objects.forEach((item) => {
+        const isImportIntoWorkspace = workspaces?.length;
```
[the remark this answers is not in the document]
SuZhou-Joe: Sure, updated.

wanglam (approved):

ruanyl (approved):

src/plugins/workspace/server/saved_objects/saved_objects_wrapper_for_check_workspace_conflict.ts
```diff
@@ -289,11 +289,33 @@ export class WorkspaceConflictSavedObjectsClientWrapper {
         return { errors: [] };
       }
 
+      const { workspaces } = options;
+
+      const disallowedSavedObjects: SavedObjectsCheckConflictsObject[] = [];
+      const allowedSavedObjects: SavedObjectsCheckConflictsObject[] = [];
+      objects.forEach((item) => {
+        const isImportIntoWorkspace = !!workspaces?.length;
```
Hailong-am, line 297: nit: this line could move out the forEach loop

src/plugins/workspace/server/saved_objects/saved_objects_wrapper_for_check_workspace_conflict.ts
```diff
@@ -289,11 +289,33 @@ export class WorkspaceConflictSavedObjectsClientWrapper {
...
+        }
+
+        // For 2.14, data source can only be created without workspace info
+        if (this.isDataSourceType(item.type) && isImportIntoWorkspace) {
+          disallowedSavedObjects.push(item);
+          return;
+        }
+
+        allowedSavedObjects.push(item);
+        return;
```
Hailong-am, line 311: nit: return is not need here

PR #6669, opened 2024-04-29
[Workspace] Fix: optimization on handling invalid workspace id in workspace_ui_settings wrapper

SuZhou-Joe:
### Description

A new wrapper, workspace_ui_settings, has been introduced by workspace plugin. When a user attempts to access an invalid workspace, OSD will display plain JSON instead of a user-friendly web page. This PR primarily addresses the issue by adding a `try catch` wrapper to the `client.get` method.

### Before fix
image

### After fix
image

As well as an error log in OSD:

image

## Testing the changes

- Use the branch to bootstrap and enable workspace feature flag
- Type an invalid url like `http://localhost:5601/w/123/app/workspace_overview`
- An error page with error message saying: `Saved object [workspace/123] not found` should be found.

## Changelog

- fix: optimization on handling invalid workspace id in workspace_ui_settings wrapper

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
+++ b/src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
@@ -17,12 +17,14 @@ import {
   SavedObjectsClientContract,
 } from '../../../../core/server';
 import { WORKSPACE_UI_SETTINGS_CLIENT_WRAPPER_ID } from '../../common/constants';
+import { Logger } from '../../../../core/server';
 
 /**
  * This saved object client wrapper offers methods to get and update UI settings considering
  * the context of the current workspace.
  */
 export class WorkspaceUiSettingsClientWrapper {
+  constructor(private readonly logger: Logger) {}
   private getScopedClient?: SavedObjectsServiceStart['getScopedClient'];
 
   /**
@@ -61,9 +63,23 @@ export class WorkspaceUiSettingsClientWrapper {
           options
         );
 
-        const workspaceObject = await this.getWorkspaceTypeEnabledClient(
-          wrapperOptions.request
-        ).get<WorkspaceAttribute>(WORKSPACE_TYPE, requestWorkspaceId);
+        let workspaceObject: SavedObject<WorkspaceAttribute> = {
+          id: requestWorkspaceId,
+          type: 'workspace',
+          attributes: {
+            id: requestWorkspaceId,
+            name: '',
+          },
+          references: [],
+        };
+
+        try {
+          workspaceObject = await this.getWorkspaceTypeEnabledClient(wrapperOptions.request).get<
+            WorkspaceAttribute
+          >(WORKSPACE_TYPE, requestWorkspaceId);
+        } catch (e) {
+          this.logger.error(`Unable to get workspaceObject on id: ${requestWorkspaceId}`);
+        }
 
         configObject.attributes = {
           ...configObject.attributes,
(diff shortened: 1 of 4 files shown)


src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -61,9 +63,23 @@ export class WorkspaceUiSettingsClientWrapper {
...
+          references: [],
+        };
+
+        try {
+          workspaceObject = await this.getWorkspaceTypeEnabledClient(wrapperOptions.request).get<
+            WorkspaceAttribute
+          >(WORKSPACE_TYPE, requestWorkspaceId);
+        } catch (e) {
+          this.logger.error(`Unable to get workspaceObject on id: ${requestWorkspaceId}`);
+        }
```
ruanyl, line 82: Instead of giving `workspaceObject` a dummy empty value, what about just initialize it as `null`?

```suggestion
        let workspaceObject: SavedObject<WorkspaceAttribute> | null = null

        try {
          workspaceObject = await this.getWorkspaceTypeEnabledClient(wrapperOptions.request).get<
            WorkspaceAttribute
          >(WORKSPACE_TYPE, requestWorkspaceId);
        } catch (e) {
          this.logger.error(`Unable to get workspaceObject on id: ${requestWorkspaceId}`);
        }
        
        configObject.attributes = {
          ...configObject.attributes,
          ...(workspaceObject ? workspaceObject.attributes.uiSettings : {}),
        };
```

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -111,7 +127,9 @@ export class WorkspaceUiSettingsClientWrapper {
           options
         );
 
-        configObject.attributes = workspaceUpdateResult.attributes.uiSettings;
+        if (workspaceUpdateResult.attributes.uiSettings) {
```
ruanyl, line 130: Is it necessary to check if `workspaceUpdateResult.attributes.uiSettings` is null? Because for `update` case, the uiSettings seems cannot be `null`.

SuZhou-Joe: There is a type error saying that configObject.attributes can not be null so that I added a `if`, type assertion should work as well but I feel it is more safe to check.

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
SuZhou-Joe: Sure, it looks more concise.

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -61,13 +63,24 @@ export class WorkspaceUiSettingsClientWrapper {
...
+        } catch (e) {
+          this.logger.error(`Unable to get workspaceObject with id: ${requestWorkspaceId}`);
+        }
+
+        configObject.attributes = {
+          ...configObject.attributes,
+          ...(workspaceObject ? workspaceObject.attributes.uiSettings : {}),
+        };
 
         configObject.attributes = {
```
wanglam, line 81: Seems the `configObject.attributes` has already been set above. Why we need to assign it again?

SuZhou-Joe: You are right, remove the useless assign.

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
ruanyl: I see, that makes sense

wanglam (approved):

ruanyl (approved):

Hailong-am (approved):

kavilla (approved):

PR #6683, opened 2024-04-30
Optimize scrolling behavior of Discover table

AMoo-Miki:
### Description

This change addresses three problems by lazy-loading additional rows when infinitely scrolling the table:
1. While a page size of 10 helped speed up rendering, scrolling users encountered the bottom of the table too often. By continuing to load a page but lazy-loading 4 additional pages (for a total of 5), that frequency is reduced.
2. As new rows were mounted, their content pushed cell boundaries and caused column widths to change; this was unpleasant and distracting. After the first page is loaded, the columns are allowed to adjust to the content and then the dimensions are fixed to prevent that.
3. Allow the paginated experience have the 50 pages that were previously allocated to it.

### Issues Resolved

TBA

#### Before
https://github.com/opensearch-project/OpenSearch-Dashboards/assets/3527403/ec727395-8b4c-4632-a2e7-76ecee53765f

#### After
https://github.com/opensearch-project/OpenSearch-Dashboards/assets/3527403/1abc0d62-961c-40c4-9af8-2a4fe0f24444

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/components/default_discover_table/table_header.tsx
+++ b/src/plugins/discover/public/application/components/default_discover_table/table_header.tsx
@@ -39,7 +39,7 @@ export function TableHeader({
 }: Props) {
   return (
     <tr data-test-subj="docTableHeader" className="osdDocTableHeader">
-      <th style={{ width: '24px' }} />
+      <th style={{ width: '28px' }} />
       {displayedColumns.map((col) => {
         return (
           <TableHeaderColumn
--- a/changelogs/fragments/6683.yml
+++ b/changelogs/fragments/6683.yml
@@ -1,0 +1,2 @@
+feat:
+- Optimize scrolling behavior of Discover table ([#6683](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6683))
--- a/src/plugins/discover/public/application/components/data_grid/data_grid_table.tsx
+++ b/src/plugins/discover/public/application/components/data_grid/data_grid_table.tsx
@@ -3,8 +3,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import './_data_grid_table.scss';
-
 import React, { useState } from 'react';
 import { EuiPanel } from '@elastic/eui';
 import { IndexPattern, getServices } from '../../../opensearch_dashboards_services';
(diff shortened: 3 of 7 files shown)


src/plugins/discover/public/application/components/default_discover_table/table_header.tsx
```diff
@@ -39,7 +39,7 @@ export function TableHeader({
 }: Props) {
   return (
     <tr data-test-subj="docTableHeader" className="osdDocTableHeader">
-      <th style={{ width: '24px' }} />
+      <th style={{ width: '28px' }} />
```
AMoo-Miki, line 42: 24px for the button + 4px for a right padding = 28px

kavilla:
@AMoo-Miki for the test failure, I think it might be worth to reduce the number instead of having the test build in the non-optimized version. Versus modifying your logic for the test to pass.

AMoo-Miki:
> @AMoo-Miki for the test failure, I think it might be worth to reduce the number instead of having the test build in the non-optimized version. Versus modifying your logic for the test to pass.

Since `retry.try` has a huge timeout of [2 mins](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/2de11ff2a277ef9dc4ec50fd54f3cf4f3dca3f6f/packages/osd-test/src/functional_test_runner/lib/config/schema.ts#L119), it is safe to assume that the lazy-loading would have run in the given time to render the 50 rows that we would expect.

kavilla:
> > @AMoo-Miki for the test failure, I think it might be worth to reduce the number instead of having the test build in the non-optimized version. Versus modifying your logic for the test to pass.
> 
> Since `retry.try` has a huge timeout of [2 mins](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/2de11ff2a277ef9dc4ec50fd54f3cf4f3dca3f6f/packages/osd-test/src/functional_test_runner/lib/config/schema.ts#L119), it is safe to assume that the lazy-loading would have run in the given time to render the 50 rows that we would expect.

oh so you are saying it's an actual failure? i thought maybe the test might have just been actually expecting 50 rows that aren't in fully view but since we optimized it wasn't able to find those rows unless it was scrolled with

kavilla (approved):
one test is flaky. and the link checker has been fixed. lgtm!!

ananzh (approved):

PR #6701, opened 2024-05-02
fix: register right navigation for dev tool only when console is enabled

raintygao:
### Description

register right navigation for dev tool only when console is enabled

## Screenshot

image
console is disabled

image
console is enabled

## Testing the changes

1. use `yarn start --console.enabled=false`, dev tool and top right navigation won't be registered.
2. use  `yarn start`, dev tool and top right navigation will both be registered.

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


kavilla (approved):

ananzh:
nice! @raintygao thanks for the quick fix.

ananzh (approved):

ananzh:
@kavilla I think the latest feedback is when we start the dashboard `yarn start`, the default behavior is that the `top right navigation` will not be registered. Then `yarn start --console.enabled=true`, both dev tool and top right navigation should be registered. Am I right?

Issue #6704, opened 2024-05-02
[BUG] Startup fails with theme overrides


BionIT:
**Describe the bug**

![Screenshot 2024-05-02 at 12 15 07 PM](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/f9258ee4-761e-4317-ab65-f0de322af1ac)
![Screenshot 2024-05-02 at 12 15 15 PM](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/2b488be6-72bf-43eb-8370-7169bc8d3b71)
![Screenshot 2024-05-02 at 12 15 25 PM](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/17714150/45768161-2512-435e-9fad-bc37f0c49ed8)

**To Reproduce**

**Expected behavior**

**OpenSearch Version**
Please list the version of OpenSearch being used.

**Dashboards Version**
Please list the version of OpenSearch Dashboards being used.

**Plugins**

Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

BionIT:
checkout a commit older than this PR would resolve thus narrow down to this PR https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5652

BionIT:
Different from the screen recording in the PR https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6703

The dashboard in future playground won't display anything

BionIT:
Found issue still happening, thus re-opened

BionIT:
Yep, @joshuarrrr this is after switching to an older commit than the PR which introduced the bug, otherwise, we see empty page. Do we prefer an empty page? There is another bug in observability, already pinged owners.

joshuarrrr:
I just need more info to be able to debug it, so if there was a version that included https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6703 and https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5652, that would be really helpful, because it doesn't repro running locally. I can also provide some additional context:

[...]

PR #4475, opened 2023-07-01
Enable theme-switching via Advanced Settings to preview the Next theme

AMoo-Miki:
### Description

Enabled theme-switching via "Advanced Settings" to preview the Next theme

Also:
* Removed the unused "v8 (beta)" theme
* Removed the overrides that locked in the light default theme
* Corrected theme version selection logic in the legacy UI renderer

### Demo

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/3527403/727dbc24-e142-42ac-a3c7-387fdc39033f

### Check List

- [X] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
  - [ ] `yarn test:ftr`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [X] Update [CHANGELOG.md](./../CHANGELOG.md)
- [X] Commits are signed per the DCO using --signoff

PR #6730, opened 2024-05-07
Fix/UI settings overrides

joshuarrrr:
### Description

There are a couple different fixes here, which can be viewed separately by looking at the separate commits. But the root cause of https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6704 was primarily the fact that the implementation of `getOverridesOrDefault` assumed the wrong object shape for the `overrides` property, when it tried to access a `value` property instead of just the value itself. Note that the existing unit test also had this same error.

Because the `ui_render_mixin` is an old legacy `.js` file, there's not much help from typescript, so I also added some enforcement to make sure we're always sending boolean values to the template.

Additionally, I realized that the validation for the `theme:version` setting was not correct, and needed updating to match our current themes.

Finally, as requested, I updated the new user theme controls to be deactivated by default.

No changelog necessary, as the [previous PR this fixes](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5652) has not yet been released.

### Future work

@AMoo-Miki @manasvinibs - The [past discussion](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/4475/files#r1251126091) and this [TODO comment](https://github.com/AMoo-Miki/OpenSearch-Dashboards/blob/31e8481fd2c0bf185d34c1140724e9320f6a676c/src/core/server/ui_settings/ui_settings_config.ts#L50-L53) seem to indicate that we don't actually want to maintain support for these overrides long term. But even if we do, I don't think the schema definition for overrides makes much sense:

```
```
We probably want users to be aware of if they specify invalid overrides and not silently ignore (which is what was happening with an incorrectly set value in @BionIT 's configuration.

### Issues Resolved

Fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6704

## Testing the changes

The linked issue only surfaced when deploying https://future.playground.opensearch.org/ , and only when not yet authenticated, because the configuration file specified overrides such as:

```yaml
uiSettings:
  overrides:
    'theme:darkMode': false
```

To validate these fixes, test with authentication enabled, a non-authenticated user, and theme UI Settings overrides.

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ashwin-pc (approved):
Thanks for these changes Josh!

ashwin-pc:
@joshuarrrr Dont we want a changelog entry for this change?

joshuarrrr:
> @joshuarrrr Dont we want a changelog entry for this change?

Well... originally I was thinking no, in terms of fixing the new theme controls, which haven't yet been released. But now I'm remembering that this was actually a latent bug in the UI settings client, so I'll add a changelog entry.

ananzh (approved):

ashwin-pc (approved):

AMoo-Miki (changes requested):
The failures indicate substantial visual diff. Putting a hold so we don't accidentally merge before examining the diff.

ananzh:
Remove 2.15 label since the https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5652 has not yet been released in 2.14.

ananzh:
### ciGroup9 Failure Research

Currently there are 9 failed tests. From debug, I found there are two issues.

#### default dark mode 
From debug I found the change of default mode is caused by !! order.  For `await !!uiSettings.get('theme:darkMode’)`, the !! operator is applied before await, causing the following steps:
1. uiSettings.get('theme:darkMode') returns a Promise.
2. !! tries to coerce the Promise object to a boolean.
3. Since any non-null object in JavaScript is truthy, !! converts the Promise to true.
4. await then resolves the true value, but this doesn't make sense in the usual await context.

So if we move !! before await `!!(await uiSettings.get('theme:darkMode'));`, the operations are correctly ordered:
1. await uiSettings.get('theme:darkMode') resolves the Promise and gets the actual boolean value (false in your case).
2. !! then converts this resolved boolean value to true or false.

But why do we need !!? Do we see any cases that we will receive truthy or falsy value? If not, maybe go back to `await uiSettings.get('theme:darkMode')`?

#### can't find `advancedSetting-saveButton`

after resolve the first one, there are 4 failed tests left, all due to
```
Error: retry.try timeout: TimeoutError: Waiting for element to be located By(css selector, [data-test-subj="advancedSetting-saveButton"])
         │ Wait timed out after 10024ms
```
This is because when we set up like below:
```
await PageObjects.settings.toggleAdvancedSettingCheckbox('theme:enableUserControl');
await PageObjects.settings.toggleAdvancedSettingCheckbox('theme:darkMode');
```
The `enableUserControl` will block admin to set up darkMode.

Screenshot 2024-06-06 at 8 47 31 AM

**This is a bit wired cuz this means admin now can't control admin's own mode.**  If this is expected, then we could test

```
it('admin's customized dark mode logo for opensearch overview header is applied', async () => {
        await PageObjects.common.navigateToApp('management/opensearch-dashboards/settings');
        await PageObjects.settings.toggleAdvancedSettingCheckbox('theme:darkMode');
        await PageObjects.common.navigateToApp('opensearch_dashboards_overview');
        await testSubjects.existOrFail('osdOverviewPageHeaderLogo');
        const actualLabel = await testSubjects.getAttribute(
          'osdOverviewPageHeaderLogo',
          'data-test-logo'
        );
        expect(actualLabel.toUpperCase()).to.equal(expectedMarkLogoDarkMode.toUpperCase());
      });

it('if enable user control, admin's customized dark mode logo for opensearch overview header is not applied', async () => {
        await PageObjects.common.navigateToApp('management/opensearch-dashboards/settings');
        await PageObjects.settings.toggleAdvancedSettingCheckbox('theme:enableUserControl');
        const button = await testSubjects.find('advancedSetting-editField-theme:darkMode');
        const isDisabled = await button.getAttribute('disabled') !== null;
        expect(isDisabled).equal(true);
        await PageObjects.common.navigateToApp('opensearch_dashboards_overview');
        await testSubjects.existOrFail('osdOverviewPageHeaderLogo');
        const actualLabel = await testSubjects.getAttribute(
          'osdOverviewPageHeaderLogo',
          'data-test-logo'
        );
        expect(actualLabel.toUpperCase()).to.equal(expectedMarkLogo.toUpperCase());
      });
```

ananzh:
Resolving ciGroup9 to run correct light mode will automatically resolve ciGroup3.

AMoo-Miki:
After talking to Ashwin, I would also like to understand why the original `override` mechanism is not working as expected because that bit is ancient code and we might need to make other changes accordingly.

ananzh:
Close this as tests are fixed in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7776

PR #6733, opened 2024-05-07
[OSD Availability] Prevent OSD process crashes when disk is full

### LINKED ISSUE #6607: Feature request: Allow admin to customize OSD logging exception handling behavior

OpenSearch Dashboards has its own logger implementation inspired by log4j 2. 

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/logging/README.md

The current implementation doesn't offer advanced capability like async logger that is useful when admin want to customize the their logger exception handle behavior

e.g. Handle the fatal exception when OSD fail to write to log to avoid crash

Provide a feature, which admin could configure logger exception handling 

Evaluate other open source library to see if they could be leveraged and replace OSD's implementation for same purpose
e.g.

https://github.com/log4js-node/log4js-node

Flyingliuhub:
### Description

Introducing the new feature flag `logging.ignoreEnospcError` to handle the corner case when the disk is full and causes the OSD process to crash. This configuration option controls the handling of error messages in the logging stream. It is disabled by default. When set to true, the 'ENOSPC' error message will not cause the OpenSearch Dashboards process to crash; otherwise, the original behavior will be maintained.

```
logging.ignoreEnospcError: true
```

### Issues Resolved

closes #6607 

## Screenshot or video 

Before the changes, the video is as follows:

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/33105471/a35b59d1-0803-459e-a48f-c20f4a27f6c6

## Testing the changes

Adding the feature flag and verifying the changes to handle the no disk space issue

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/33105471/1a51505c-b216-4b12-953c-23d8e67ebbac

## Changelog
- fix: [OSD Availability] Prevent OSD process crashes when disk is full

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ZilongX:
updated base and re-running all checks

ZilongX (approved):

src/legacy/server/logging/configuration.js
```diff
@@ -64,6 +64,7 @@ export default function loggingConfiguration(config) {
       json: config.get('logging.json'),
       dest: config.get('logging.dest'),
       timezone: config.get('logging.timezone'),
+      streamErrorHandling: config.get('logging.streamErrorHandling'),
```
kavilla, line 67: NIT: totally cool as is.

do we want to make this even further?

like it accept a array of strings? and if the array.includes(error.code) that it can bypass it?

if the only plan for ENOSPC, then probably can switch the name to something like`logging.ignoreEnospcError` or something no space.

kavilla:
not sure if folks still use the local file here but do you think we should add this new config here:

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/c09285c24703f9015d811f76bcd026061b705398/src/dev/build/tasks/os_packages/docker_generator/resources/bin/opensearch-dashboards-docker#L17

but prolly should def added to build repo too upon merge.

example: https://github.com/opensearch-project/opensearch-build/commit/a213aadad95e102df9d6a5620648df6242acd79f

src/legacy/server/logging/log_reporter.js
```diff
@@ -51,18 +52,33 @@ export function getLoggerStream({ events, config }) {
...
     dest = writeStr(config.dest, {
       flags: 'a',
       encoding: 'utf8',
     });
 
-    logInterceptor.on('end', () => {
-      dest.end();
-    });
+    if (config.streamErrorHandling) {
+      pipeline(logInterceptor, squeeze, format, dest, onFinished);
```
kavilla, line 63: will look into later. but qq:

does pipeline call dest end automatically? or does onfinished get invoked and then we should check if error.code === 'end' and then call dest.end()

Flyingliuhub: Yes, pipeline() automatically handles closing the streams once the pipeline is finished or if an error occurs during processing. This automatic cleanup ensures that resources associated with the streams are properly released when they are no longer needed. This behavior helps prevent memory leaks and ensures efficient resource usage in Node.js applications

kavilla: dang. could do you mind opening an issue? i feel like this legacy stuff needs to go and we updated to use the non legacy, or at least be updated to utilize the non-legacy to modern Node APIs.

kavilla (approved):

src/legacy/server/logging/configuration.js
Flyingliuhub: good call. I modified the feature flag name to ignoreEnospcError

Flyingliuhub:
> not sure if folks still use the local file here but do you think we should add this new config here:
> 
> https://github.com/opensearch-project/OpenSearch-Dashboards/blob/c09285c24703f9015d811f76bcd026061b705398/src/dev/build/tasks/os_packages/docker_generator/resources/bin/opensearch-dashboards-docker#L17
> 
> but prolly should def added to build repo too upon merge.
> 
> example: [opensearch-project/opensearch-build@a213aad](https://github.com/opensearch-project/opensearch-build/commit/a213aadad95e102df9d6a5620648df6242acd79f)

Good catch. I added new feature flag to opensearch-dashboards-docker file. And will raise another PR for opensearch-build side.

src/legacy/server/logging/log_reporter.js
Flyingliuhub: I will open an new issue for it.

kavilla (approved):

ZilongX (approved):

Issue #6766, opened 2024-05-10
[BUG] Timefield (global state) is not captured when migrate url for Discover from 2.9 to 2.11


ananzh:
**Describe the bug**

When clicks a legacy discover url, it redirect to new URL with `data_explorer` parameter. But timefield is not set , it only shows the last 15minutes in the timefield filter section. Here is a url from 2.9:
```
/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2024-04-03T03:30:00.000Z',to:'2024-05-03T04:30:00.000Z'))&_a=(columns:!(_source),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:ff959d40-b880-11e8-a6d9-e546fe2bba5f,key:category.keyword,negate:!t,params:(query:'Men!'s%20Clothing'),type:phrase),query:(match_phrase:(category.keyword:'Men!'s%20Clothing'))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:ff959d40-b880-11e8-a6d9-e546fe2bba5f,key:currency,negate:!f,params:(query:EUR),type:phrase),query:(match_phrase:(currency:EUR)))),index:ff959d40-b880-11e8-a6d9-e546fe2bba5f,interval:auto,query:(language:kuery,query:''),sort:!()) 
```
It will become
```
[...]

PR #6780, opened 2024-05-13
[Discover][Bug] Migrate global state from legacy URL

ananzh:
### Description
Allow migrate global state. I also thought about capture all other states besides `_g`, `_a, and `_q` like customized states, but our current helper func requiring us to know the key in advance. Given customized keys might have various formats, it is easy to make new bugs. 
```
export function getStatesFromOsdUrl(
  url: string = window.location.href,
  keys?: string[]
): Record<string, unknown> {
  const query = parseUrlHash(url)?.query;

  if (!query) return {};
  const decoded: Record<string, unknown> = {};
  Object.entries(query)
    .filter(([key]) => (keys ? keys.includes(key) : true))
    .forEach(([q, value]) => {
      decoded[q] = decodeState(value as string);
    });

  return decoded;
}
```

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6766

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/79961084/8da52c7d-58d4-4c0e-9083-8876d91f8bc0

## Changelog

- fix: [Discover][Bug] Migrate global state from legacy URL

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/migrate_state.ts
+++ b/src/plugins/discover/public/migrate_state.ts
@@ -133,7 +133,9 @@ export function migrateUrlState(oldPath: string, newPath = '/'): string {
           indexPattern: index,
         },
       };
+      const _g = getStateFromOsdUrl<any>('_g', oldPath);
 
+      path = setStateToOsdUrl('_g', _g, { useHash: false }, path);
       path = setStateToOsdUrl('_a', _a, { useHash: false }, path);
       path = setStateToOsdUrl('_q', _q, { useHash: false }, path);
 
--- a/changelogs/fragments/6780.yml
+++ b/changelogs/fragments/6780.yml
@@ -1,0 +1,2 @@
+fix:
+- [Discover][Bug] Migrate global state from legacy URL ([#6780](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6780))
--- a/src/plugins/discover/public/migrate_states.test.ts
+++ b/src/plugins/discover/public/migrate_states.test.ts
@@ -1,0 +1,196 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { migrateUrlState } from './migrate_state';
+import { setStateToOsdUrl, getStateFromOsdUrl } from '../../opensearch_dashboards_utils/public';
+
+jest.mock('../../opensearch_dashboards_utils/public', () => ({
+  setStateToOsdUrl: jest.fn(),
+  getStateFromOsdUrl: jest.fn(),
+}));
+
+describe('migrateUrlState', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should return the new path if no matching pattern', () => {
+    const result = migrateUrlState('#/unknown', '/newPath');
+    expect(result).toBe('/newPath');
+  });
+
+  it('should migrate doc view', () => {
+    const result = migrateUrlState('#/doc/indexPattern/id', '/newPath');
+    expect(result).toBe('#/doc/indexPattern/id');
+  });
+
+  it('should migrate context view', () => {
+    const result = migrateUrlState('#/context/indexPattern/id', '/newPath');
+    expect(result).toBe('#/context/indexPattern/id');
+  });
+
+  it('should migrate discover view with saved search id and with global state', () => {
+    (getStateFromOsdUrl as jest.Mock).mockImplementation((key) => {
+      if (key === '_a') {
+        return {
+          columns: ['column1'],
+          filters: [],
+          index: 'indexPattern',
+          interval: 'auto',
+          query: { language: 'kuery', query: 'test' },
+          sort: [['field', 'desc']],
+          savedQuery: 'savedQueryId',
+        };
+      }
+      if (key === '_g') {
+        return {
+          time: { from: 'now-15m', to: 'now' },
+          filters: [],
+          refreshInterval: { pause: true, value: 0 },
+        };
+      }
+      return null;
+    });
+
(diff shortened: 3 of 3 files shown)


ananzh:
Will add a ftr test in the ftr repo.

src/plugins/discover/public/migrate_state.ts
```diff
@@ -133,7 +133,9 @@ export function migrateUrlState(oldPath: string, newPath = '/'): string {
           indexPattern: index,
         },
       };
+      const _g = getStateFromOsdUrl<any>('_g', oldPath);
```
virajsanghvi, line 136: Is this path only entered if the url uses url fragments? I ask because it looks like `parseUrlHash` would just parse the url hash, so if it wasn't using fragments, you'd potentially override a validly set `_g` query parameter.

Also, does your comment about ftr tests address the lack of test coverage here?

abbyhu2000, line 136: we should also define the type here like the previous LegacyDiscoverState from line #114. From my memory, i think global state should include time filter and global filter, and then we can do something like the previous code: 

```
const _g = {
    timefilter,
    filters
}

**Comment by abbyhu2000** [comment]
> Will add a ftr test in the ftr repo.

I think unit test is also needed for function `migrateUrlState` https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6780/files#diff-1738b52af72f333e3c7154730455af2dc46debf51bd704ebc5324db3030b1c1aR59

File: src/plugins/discover/public/migrate_state.ts (hunk 1 above)
Comment (ananzh) [reply src/plugins/discover/public/migrate_state.ts:136]: My current understanding is that the legacy urls are all like http://localhost:5601/app/discover#/, there should be no issue with parsing and handling _g.

Comment (ananzh) [reply src/plugins/discover/public/migrate_state.ts:136]: Will add the unit test.

File: src/plugins/discover/public/migrate_state.ts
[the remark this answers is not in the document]
Comment (ananzh) [reply src/plugins/discover/public/migrate_state.ts:1]: I think we would like to proceed with the migration regardless of the presence of the global state (`_g`).

First of all, checking for both appState and globalState together and returning the path if either is null could miss valid migrations. 

Here are some 2.9 urls with/withoug `_g`:

* with `_g`, index pattern has a timestamp

If we remove `_g` from the url
```
http://localhost:5601/app/discover#/view/2e72c2f0-21a8-11ef-b7a3-09997417d2e2?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15w,to:now))&_a=(columns:!(_source),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:ff959d40-b880-11e8-a6d9-e546fe2bba5f,key:category.keyword,negate:!f,params:(query:'Men!'s%20Clothing'),type:phrase),query:(match_phrase:(category.keyword:'Men!'s%20Clothing')))),index:ff959d40-b880-11e8-a6d9-e546fe2bba5f,interval:auto,query:(language:kuery,query:Thursday),sort:!())
```
then refresh the url, it will turn to
```
http://localhost:5601/app/discover#/view/2e72c2f0-21a8-11ef-b7a3-09997417d2e2?_a=(columns:!(_source),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:ff959d40-b880-11e8-a6d9-e546fe2bba5f,key:category.keyword,negate:!f,params:(query:'Men!'s%20Clothing'),type:phrase),query:(match_phrase:(category.keyword:'Men!'s%20Clothing')))),index:ff959d40-b880-11e8-a6d9-e546fe2bba5f,interval:auto,query:(language:kuery,query:Thursday),sort:!())&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))
```
, **where a default `_g` is added**

* without `_g`: index pattern without time series data

By default, for idx without time series, there is a default  `_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))` but if remove `_g` from the url 

```
http://localhost:5601/app/discover#/view/2e72c2f0-21a8-11ef-b7a3-09997417d2e2?_a=(columns:!(_source),filters:!(),index:c915dbd0-21a8-11ef-b7a3-09997417d2e2,interval:auto,query:(language:kuery,query:''),sort:!())&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))
```
it will turn to
```
http://localhost:5601/app/discover#/view/2e72c2f0-21a8-11ef-b7a3-09997417d2e2?_a=(columns:!(_source),filters:!(),index:c915dbd0-21a8-11ef-b7a3-09997417d2e2,interval:auto,query:(language:kuery,query:''),sort:!())
``` 
**without adding back `_g` with default state.**

**Therefore, it is possible for a URL to be without global state, either by mistake or non-time-series idx and still require proper migration.**

Second, if the globalState is null (_g=!n), migration can still proceed. Just like the previous example, it will either add a default one, which is the same as 2.9 experience. 

Therefore, it is more robust if we don't check global state in the migration function and just let opensearch_dashboards_utils handles the logic.

src/plugins/discover/public/migrate_state.ts
ananzh: It is beneficial to define the type for globalState to ensure type safety and clarity. However, since the globalState can include various parameters (time, filters, freshInterval and etc), we should ensure that the type definition encompasses all possible states. So I put `any` for now to make sure any states can be migrated.

virajsanghvi (approved):

abbyhu2000 (approved):

manasvinibs (approved):

Issue #6825, opened 2024-05-24
[Multiple Datasource] Add data source selection service to support storing and getting selected data source updates


raintygao:
**Is your feature request related to a problem? Please describe.**

Some plugins may need to get data source selected info through data source selector components, there isn't a centralized way to record or manage this. 

**Describe the solution you'd like**

We plan to add a service named `DataSourceSelection` in DSM to support this. In setup stage of DSM, a instance of `DataSourceSelectionService` would be generated and this instance will be passed to each data source selector or menu component and returned to other plugin.  Selector component can stored selected data source through this instance and other plugin can subscribe or get selected data source through this instance.

```
 class DataSourceSelection {

 // This is an observable map to record all selected data sources in OSD.
 // This data is stored in memory and will be cleared with OSD context and reset if page reloads.
 // The key of map is the id of each selector component.Since DSM provides several components and there can be many components on the page at the same time,the component needs to generate an unique ID to specify itself when mounted.
 // The value would be the detail info of selected data source.
  private selectedDataSource$ = new BehaviorSubject(new Map<string, DataSourceOption[]>());
  
  // When users select data source through component, component will call this function to set selected data source.
  public selectDataSource = (componentId: string, dataSource: DataSourceOption[]) => {
  };

  // When component unmounts, it will call this function to delete stored selected data source.
  public remove = (componentId: string) => {
  };

  // Other plugins can call this function to get all selected data source value.
  public getSelectionValue = () => {
  };

  // Other plugins can call this function to get observable subject, component can use this value to subscribe update.
  public getSelection$ = () => {
  };
}
```

BionIT:
@raintygao Do we have all PRs merged for this feature?

raintygao:
Sure, will close as completed/

PR #6827, opened 2024-05-24
[Multiple Datasource] Add data source selection service to support storing and getting selected data source updates

raintygao:
### Description

1. Add data source selection service in DSM.
2. Add dataSourceSelection instance as new props to each data source selector component and call selectDataSource when selecting data source.
3. Return this instance as setup dependency for other plugins.

### Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/42465692/f43a2de9-1745-4a3f-8946-0cf369c5b624

Set selected value and other plugins can get selected value.

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/42465692/d1feadb1-bc38-491a-8931-9782793d728e

A component lifecycle from mount and select data source to unmount and remove.

image
Multiple selector components in page and data stored.

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6825 
https://github.com/opensearch-project/dashboards-assistant/issues/192

## Changelog
- feat: Add data source selection service to support storing and getting selected data source updates

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source_management/public/components/data_source_menu/types.ts
+++ b/src/plugins/data_source_management/public/components/data_source_menu/types.ts
@@ -10,6 +10,7 @@ import {
   IUiSettingsClient,
   ApplicationStart,
 } from '../../../../../core/public';
+import { DataSourceSelection } from '../../service/data_source_selection_service';
 import { DataSourceAttributes } from '../../types';
 
 export interface DataSourceOption {
@@ -40,6 +41,7 @@ export interface DataSourceMenuProps<T = any> {
   uiSettings?: IUiSettingsClient;
   application?: ApplicationStart;
   setMenuMountPoint?: (menuMount: MountPoint | undefined) => void;
+  dataSourceSelection: DataSourceSelection;
 }
 
 export const DataSourceComponentType = {
--- a/src/plugins/data_source_management/public/components/data_source_menu/types.ts
+++ b/src/plugins/data_source_management/public/components/data_source_menu/types.ts
@@ -10,7 +10,7 @@ import {
   IUiSettingsClient,
   ApplicationStart,
 } from '../../../../../core/public';
-import { DataSourceSelection } from '../../service/data_source_selection_service';
+import { DataSourceSelectionService } from '../../service/data_source_selection_service';
 import { DataSourceAttributes } from '../../types';
 
 export interface DataSourceOption {
@@ -41,7 +41,7 @@ export interface DataSourceMenuProps<T = any> {
   uiSettings?: IUiSettingsClient;
   application?: ApplicationStart;
   setMenuMountPoint?: (menuMount: MountPoint | undefined) => void;
-  dataSourceSelection: DataSourceSelection;
+  dataSourceSelection: DataSourceSelectionService;
 }
 
 export const DataSourceComponentType = {
--- a/src/plugins/data_source_management/public/components/data_source_menu/types.ts
+++ b/src/plugins/data_source_management/public/components/data_source_menu/types.ts
@@ -10,7 +10,6 @@ import {
   IUiSettingsClient,
   ApplicationStart,
 } from '../../../../../core/public';
-import { DataSourceSelectionService } from '../../service/data_source_selection_service';
 import { DataSourceAttributes } from '../../types';
 
 export interface DataSourceOption {
@@ -41,7 +40,6 @@ export interface DataSourceMenuProps<T = any> {
   uiSettings?: IUiSettingsClient;
   application?: ApplicationStart;
   setMenuMountPoint?: (menuMount: MountPoint | undefined) => void;
-  dataSourceSelection: DataSourceSelectionService;
 }
 
 export const DataSourceComponentType = {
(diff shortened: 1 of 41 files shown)


src/plugins/navigation/public/top_nav_menu/top_nav_menu.test.tsx
```diff
@@ -133,6 +136,7 @@ describe('TopNavMenu', () => {
             fullWidth: true,
             activeOption: [{ label: 'what', id: '1' }],
           },
+          dataSourceSelection,
```
SuZhou-Joe, line 139: Does it mean all the places that are using data source pickers need to pass the dataSourceSelection explicitly?  I thought it is an internal state and plugins do not need to be aware of this context.

raintygao: There is no need to pass this instance manually, because usually all components are created by functions provided by DSM and passed within the function.  Except for some test cases when manually rendering components that need to pass.

raintygao:
> Hi Tianyu. Could you help me to better understanding what we are trying to do here? I see that we have add the pops of DataSourceSelectable and states of ComponenId. What are we trying to achieve here? Thanks

Hi Ella, unlike other plugins that need to select their own data sources, chatbot needs to get the data source selected by other applications from the OSD page. As these selector components don't store selected value and there isn't a centralized way in DSM to manage this, so this PR mainly does two things: 
1. Add a new service to the DSM which supports storing and getting data source selected by each component. 
2. After each selector component completes the selection, the value will be stored in the instance of the service, which can be got by other plugins.

raintygao:
> @raintygao looks like there are lots of UI change, would you add a few screenshot and video to help the review.

Sure. Although this PR makes updates in each UI component, it only adds the logic to store selected value in service after selecting the data source, and doesn't not make any changes to the UI.  The changes to assistant plugin are still in progress and depending on this PR, I have added a sample video to the PR content to demonstrate the purpose.

ruanyl:
@raintygao could you update the PR the description as well to includes the backgrounds of why you need to introduce this change? I'm afraid many people don't have enough context to review the changes.

raintygao:
> @raintygao could you update the PR description as well to includes the backgrounds of why you need to introduce these changes? I'm afraid many people don't have enough context to review the changes.

Sure. Already shared background and design offline. Will create a new issue in plugin repo for more context.

src/plugins/data_source_management/public/components/data_source_menu/types.ts
```diff
@@ -40,6 +41,7 @@ export interface DataSourceMenuProps<T = any> {
   uiSettings?: IUiSettingsClient;
   application?: ApplicationStart;
   setMenuMountPoint?: (menuMount: MountPoint | undefined) => void;
+  dataSourceSelection: DataSourceSelection;
```
BionIT, line 44: I think this should be part of componentConfig since not all components need to provide this as prop. Also, enforcing this prop as required would break all consumers, since it is provided by the DSM plugin, we can make it optional?

src/plugins/data_source_management/public/service/data_source_selection_service.ts
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { BehaviorSubject } from 'rxjs';
+import { DataSourceOption } from '../components/data_source_menu/types';
+
+export class DataSourceSelection {
```
BionIT, line 9: Can the class name match the module name like `DataSourceSelectionService`?

src/plugins/data_source_management/public/components/data_source_menu/types.ts
raintygao: I prefer to let it out of componentConfig because this instance should be passed from DSM not other plugins. As this instance is shared and limited in DSM, so we can set it as strict required.

I initially thought that other plugins only can render DS selector components through the two functions `ui.DataSourceSelector` and `ui.getDataSourceMenu` (like this way https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/dev_tools/public/application.tsx#L118), so I only need to pass the dataSourceSelection instance to the component through this two functions of this DSM.

Until I saw home plugin directly importing component class for rendering without using two functions above(https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/home/public/application/components/tutorial_directory.js#L236), in this way the dataSourceSelection would be missing in the component.  

So I'm thinking that in order to safely share  dataSourceSelection instance among all components and be compatible with all usage ways, what about using a getter in each component to get instance directly (such as https://github.com/opensearch-project/OpenSearch -Dashboards/pull/6827/files#diff-ec5b6ef64504177e50cad326dcc07ada2135faafe924d25b0f4df088bc7b4823R80), so that I no longer have to manually pass dataSourceSelection through `ui.DataSourceSelector` and `ui.getDataSourceMenu`.

BionIT: Using getter sounds good

BionIT (approved):

zhyuanqi (approved):

Issue #6830, opened 2024-05-24
[Workspace] Only OSD admin can create workspace


yubonluo:
**Is your feature request related to a problem? Please describe.**
With workspace feature enabled, saved objects permission control enabled and security plugin installed, only OSD admin(https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6389) can create workspace. If permission control is closed, anyone has the permission to create workspace.

**Describe the solution you'd like**
Before creating a workspace, saved objects wrapper will verify if the user/group is the OSD admin.

**Additional context**
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4944

yubonluo:
> Hi @yubonluo, is this issue for 2.15 or 2.16?

Hi @BionIT rhis issue is for 2.15.

PR #6831, opened 2024-05-24
[Workspace] Only OSD admin can create workspace

yubonluo:
### Description

Only OSD admin can create workspace

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6830

## Screenshot
image

## Testing the changes
1. Clone the latest osd/security-dashboard plugin code and run yarn osd bootstrap
2. Modify config/opensearch_dashboards.yml
```yml
opensearchDashboards.dashboardAdmin.groups: ["dashboard_admin"]
opensearchDashboards.dashboardAdmin.users: ["username"]
workspace.enabled: true 
opensearch.username: "username"
opensearch.password: "password" 
opensearch.ssl.verificationMode: none
```
3. Run yarn start --no-base-path
4. Create a workspace.

## Changelog
- feat: [Workspace] Only OSD admin can create workspace

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/workspace/server/workspace_client.ts
```diff
@@ -77,6 +84,14 @@ export class WorkspaceClient implements IWorkspaceClientImpl {
   private formatError(error: Error | any): string {
     return error.message || error.error || 'Error';
   }
+  private hasOSDAdminPermission(requestDetail: IRequestDetail) {
+    if (
```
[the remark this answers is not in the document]
Hailong-am: +1

src/plugins/workspace/server/workspace_client.ts
```diff
@@ -89,6 +104,7 @@ export class WorkspaceClient implements IWorkspaceClientImpl {
     payload: Omit<WorkspaceAttributeWithPermission, 'id'>
   ): ReturnType<IWorkspaceClientImpl['create']> {
     try {
+      this.hasOSDAdminPermission(requestDetail);
```
ruanyl, line 107: Having this check in workspace client may not enough, as people may call saved object api to create workspace directly. Maybe better to have it in the saved object client wrapper. What do you think? cc @SuZhou-Joe

SuZhou-Joe: Agree, let's do it in the permission wrapper so that we can ensure the check will be performed only when permission control feature flag is enabled.

yubonluo: Thanks, I will move the related logic to saved object client.

src/plugins/workspace/server/workspace_client.ts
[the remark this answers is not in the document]
yubonluo: Thanks, the tests has been added.

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
```diff
@@ -330,6 +349,9 @@ export class WorkspaceSavedObjectsClientWrapper {
       attributes: T,
       options?: SavedObjectsCreateOptions
     ) => {
+      // Only OSD admin can create workspace
+      if (isTypeContained(WORKSPACE_TYPE, type)) throw generateOSDAdminPermissionError();
```
ruanyl, line 353: Nit: I'd prefer simply do `if (WORKSPACE_TYPE === type)`

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
```diff
@@ -272,6 +282,15 @@ export class WorkspaceSavedObjectsClientWrapper {
       objects: Array<SavedObjectsBulkCreateObject<T>>,
       options: SavedObjectsCreateOptions = {}
     ): Promise<SavedObjectsBulkResponse<T>> => {
+      // Only OSD admin can bulkCreate workspace
+      if (
+        isTypeContained(
+          WORKSPACE_TYPE,
+          objects.map((object) => object.type)
+        )
+      )
```
ruanyl, line 291: Nit: will this look a bit cleaner?
```suggestion
      // Only OSD admin can bulkCreate workspace
      if (objects.some(obj => obj.type === WORKSPACE_TYPE))
```

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
yubonluo: Sure, I will simply the code.

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
yubonluo: Sure, I will simply the code.

src/plugins/workspace/server/saved_objects/integration_tests/workspace_saved_objects_client_wrapper.test.ts
```diff
@@ -316,6 +316,17 @@ describe('WorkspaceSavedObjectsClientWrapper', () => {
 
       expect(createResult.error).toBeUndefined();
     });
+
+    it('should throw forbidden error when user create a workspce and is not OSD admin', async () => {
+      let error;
+      try {
+        await notPermittedSavedObjectedClient.create('workspace', {}, {});
```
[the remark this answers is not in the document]
yubonluo: Your understanding is correct. I will change it to `permittedSavedObjectedClient `.

ruanyl (approved):

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
```diff
@@ -330,6 +343,9 @@ export class WorkspaceSavedObjectsClientWrapper {
       attributes: T,
       options?: SavedObjectsCreateOptions
     ) => {
+      // Only OSD admin can create workspace
+      if (type === WORKSPACE_TYPE) throw generateOSDAdminPermissionError();
```
SuZhou-Joe, line 347: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/d911fa7144fec19b9409a8dea6b7d4d06af9a2b8/src/plugins/workspace/server/workspace_client.ts#L195

It seems we are using `create with overwrite` to `update` the workspace object.

Hailong-am: Yes, it's by intention. Here is how update api behaves, what we want to update is to fully replace existing document. 

> The update API also supports passing a partial document, which is merged into the existing document. To fully replace an existing document, use the [index API].

yubonluo: Make sense, when users update a workspace, `createWithWorkspacePermissionControl` with `options.overwrite=true` will be called. 
The code of checking overwrite has been added.

wanglam: Call update API with permissions object won't remove existing permissions attributes. For example, there is a workspace saved object with permissions attribute like:
```
{
    read: {users:['foo']}
}
```
If I call update API with 
```
{
    library_read: {users:['foo']}
}
```
The read permission won't be updated by savedObjects.update method. That's why we use create method here.

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
```diff
@@ -272,6 +281,10 @@ export class WorkspaceSavedObjectsClientWrapper {
       objects: Array<SavedObjectsBulkCreateObject<T>>,
       options: SavedObjectsCreateOptions = {}
     ): Promise<SavedObjectsBulkResponse<T>> => {
+      // Only OSD admin can bulkCreate workspace
+      if (objects.some((obj) => obj.type === WORKSPACE_TYPE) && !options?.overwrite)
```
wanglam, line 285: Why we need add `options.overwrite` flag check here? As my understanding, user can create new saved objects when `options.overwrite` equal true. See https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/saved_objects/service/lib/repository.ts#L361 Seems `options.overwrite` only worked when saved object id exists. Shall we handle this case?

yubonluo: In create method, I will add the check `options.id` and `options.overwrite` to make users with admin permission update workspace successfully. In bulkCreate method, I will not add the check as the bulkCreate method is not called in the workspaceClient. Is it feasible to do this?

wanglam: I'd recommend to do these. I think object.id will be stored in objects. I think you can have a test in your local. You can try to pass a workspace saved objects array without id and `options.overwrite` equal `true`. Then the  workspace can be created even not an OSD admin.  I think it will bypass the OSD admin permission check.

yubonluo: In my latest commit, if users without OSD admin permission want to bulkCreate workspace, they will fail. 
```
wrapper.bulkCreate([{ type: 'workspace', attributes: {} }])
```
If users without OSD admin permission want to create a workspace, they will fail.
```
wrapper.create('workspace', {name: 'test'}, {})
```
If users with admin permission want to update a workspace, they will success.
```
wrapper.create('workspace', {name: 'test'}, {id: 'ws', overwrite: true})
```

SuZhou-Joe: I think we can safely say user is trying to `update a workspace` only if:
1. Using overwrite mode and request with object id.
3. The workspace user wants to overwrite indeed exists.

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
```diff
@@ -330,6 +346,12 @@ export class WorkspaceSavedObjectsClientWrapper {
       attributes: T,
       options?: SavedObjectsCreateOptions
     ) => {
+      // Only OSD admin can create workspace.
+      // If options contains id and overwrite, it is an update action.
+      if (type === WORKSPACE_TYPE && !(options?.id && options?.overwrite)) {
```
SuZhou-Joe, line 351: Do we need to check if the workspace exists with `options.id`?

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
```diff
@@ -272,6 +281,13 @@ export class WorkspaceSavedObjectsClientWrapper {
       objects: Array<SavedObjectsBulkCreateObject<T>>,
       options: SavedObjectsCreateOptions = {}
     ): Promise<SavedObjectsBulkResponse<T>> => {
+      // If objects contain workspace id and options.overwrite is true, it is a bulkCreate action.
+      const isUpdateWorkspaceMode =
+        objects.some((obj) => obj.id && obj.type === WORKSPACE_TYPE) && options.overwrite;
+      // Only OSD admin can bulkCreate workspace.
+      if (objects.some((obj) => obj.type === WORKSPACE_TYPE) && !isUpdateWorkspaceMode)
+        throw generateOSDAdminPermissionError();
```
SuZhou-Joe, line 289:
```suggestion
      // Objects with id in overwrite mode will be regarded as update
      const objectsToCreate: Array<SavedObjectsBulkCreateObject<T>> = options.overwrite ? objects.filter((obj) => !obj.id) : objects;
      // Only OSD admin can bulkCreate workspace.
      if (objectsWaitToCreate.some((obj) => obj.type === WORKSPACE_TYPE))
        throw generateOSDAdminPermissionError();
```
How do you like the implementation like this?

yubonluo: Sure, the logic has been updated. Thanks.

yubonluo:
CC @SuZhou-Joe @ruanyl If you don‘t have any concerns about this CR, could you please help me approve it?

SuZhou-Joe (dismissed):

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
```diff
@@ -272,6 +281,12 @@ export class WorkspaceSavedObjectsClientWrapper {
       objects: Array<SavedObjectsBulkCreateObject<T>>,
       options: SavedObjectsCreateOptions = {}
     ): Promise<SavedObjectsBulkResponse<T>> => {
+      // Objects with id in overwrite mode will be regarded as update
+      const objectsToCreate = options.overwrite ? objects.filter((obj) => !obj.id) : objects;
+      // Only OSD admin can bulkCreate workspace.
+      if (objectsToCreate.some((obj) => obj.type === WORKSPACE_TYPE))
```
virajsanghvi, line 287: nit: not sure what code style is, but definitely recommend using braces, even for single line if statements

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
```diff
@@ -272,6 +281,12 @@ export class WorkspaceSavedObjectsClientWrapper {
       objects: Array<SavedObjectsBulkCreateObject<T>>,
       options: SavedObjectsCreateOptions = {}
     ): Promise<SavedObjectsBulkResponse<T>> => {
+      // Objects with id in overwrite mode will be regarded as update
+      const objectsToCreate = options.overwrite ? objects.filter((obj) => !obj.id) : objects;
+      // Only OSD admin can bulkCreate workspace.
```
virajsanghvi, line 286: Sorry, I may be missing something here - how is this allowing only admins to create workspaces?

src/plugins/workspace/server/saved_objects/integration_tests/workspace_saved_objects_client_wrapper.test.ts
```diff
@@ -316,6 +332,62 @@ describe('WorkspaceSavedObjectsClientWrapper', () => {
 
       expect(createResult.error).toBeUndefined();
     });
+
+    it('should able to create workspace with override', async () => {
```
virajsanghvi, line 336: nit: overwrite?

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
yubonluo: This is a new requirement regarding workspaces in version 2.15.

src/plugins/workspace/server/saved_objects/integration_tests/workspace_saved_objects_client_wrapper.test.ts
yubonluo: Consistent with the previous test name.

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
yubonluo: Thank you for your suggestion. I have added the braces.

ruanyl (approved):

SuZhou-Joe (approved):

PR #6833, opened 2024-05-24
Discover Flint datasource selector implementation

paulstn:
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/discover/public/application/helpers/get_data_source_fetch_strategy.ts
```diff
@@ -0,0 +1,39 @@
...
+ * under the License.
+ */
+
+/**
+ * Returns a polling strategy based off of the data source type, which is meant to be given as
+ * an option during fetching.
+ */
+
+export function dataSourceFetchingStrategy(dataSource: string): string {
+  if (dataSource === 's3glue') return 'polling';
```
kavilla, line 37: data sources aren't unique to discover. so shouldn't any querying to this data source type be polling?

would it make sense to make the strategy apart of the data source type?

src/plugins/data/common/constants.ts
```diff
@@ -30,6 +30,10 @@
 
 export const DEFAULT_QUERY_LANGUAGE = 'kuery';
 
+// to resolve: copies constants from src/plugins/data/public/data_sources/constants.tsx, should it be removed from there?
+export const S3_GLUE_DATA_SOURCE_TYPE = 's3glue';
```
kavilla, line 34: nit: i understand unrelated to this PR is s3glue the only planned polling type of data source?

perhaps we can iteratively improve the existing hardcode resources that has registration which can registration one time within a secondary plugin.

similar to other patterns: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/search/search_service.ts#L112

like certain patterns will utilize a record and if the data source type exists we don't re-register.

src/plugins/data/common/search/search_source/types.ts
```diff
@@ -29,6 +29,7 @@
  */
 
 import { NameList } from 'elasticsearch';
+import { DataSource } from 'src/plugins/data/public';
```
kavilla, line 32: does OSD complain about this? common folder should prohibit stuff from public and server folders.

however, I think this raises a good point that this data source should be within the common folder. if it is in conflicts with the other data_sources perhaps a different name for it temporarily and aliasing it so we don't have to change in many locations.

src/plugins/data/public/data_sources/datasource/types.ts
```diff
@@ -51,6 +51,7 @@ export interface IDataSourceSettings<T extends IDataSourceMetadata = IDataSource
   id: string;
   type: string;
   name: string;
+  connectionType: 'os' | 'flint';
```
kavilla, line 54: could this be a registration point of connection types and then in the query enhancers we can register the conneciton type.

src/plugins/data/public/data_sources/datasource_selector/types.ts
```diff
@@ -23,6 +23,7 @@ export interface DataSourceOption {
   label: string;
   value: string;
   type: string;
+  connectionType: 'os' | 'flint';
```
kavilla, line 26: i'd watch out with 'os' it might be confusing. within this repo we use `oss` as open-source software and the implication can get blurred since 'os' could be OpenSearch the service. It could also be "open source" since Flint is a product if i'm not wrong. so if the connectiontype is opensearch we can probably be more specific. or even call it 'default' and default is OpenSearch

src/plugins/data_explorer/public/components/sidebar/index.tsx
```diff
@@ -77,31 +77,24 @@ export const Sidebar: FC = ({ children }) => {
...
   const handleSourceSelection = useCallback(
     (selectedDataSources: DataSourceOption[]) => {
+      setSelectedSources(selectedDataSources);
       if (selectedDataSources.length === 0) {
-        setSelectedSources(selectedDataSources);
         return;
       }
-      // Temporary redirection solution for 2.11, where clicking non-index-pattern data sources
-      // will prompt users with modal explaining they are being redirected to Observability log explorer
-      if (selectedDataSources[0]?.ds?.getType() !== 'DEFAULT_INDEX_PATTERNS') {
```
kavilla, line 97: nice

src/plugins/data/common/search/search_source/fetch/get_search_params.ts
```diff
@@ -60,9 +60,11 @@ export function getExternalSearchParamsFromRequest(
   const searchParams = getSearchParams(getConfig);
   const dataFrame = getDataFrame();
   const indexTitle = searchRequest.index.title || searchRequest.index;
+  const dataSource = searchRequest.dataSource;
 
   return {
     index: indexTitle,
+    dataSource,
```
kavilla, line 67: within the dataframe, dataframe.name should be the datasource. at the time it was implemented it was the index or index pattern.

does the name still work here? if not do you think we can add the data source information to the dataframe.meta field? 

for the async query @sejli is adding query meta infromation like the session id to the df.meta. I think that fits. It also will help with future work with stacking queries as we would want to get the data source attached to query which we will get from the meta field

kavilla: this we wouldn't need to update these parts.

src/plugins/data/public/data_sources/datasource/types.ts
```diff
@@ -51,6 +51,7 @@ export interface IDataSourceSettings<T extends IDataSourceMetadata = IDataSource
...
   metadata: T;
```
kavilla, line 55: do we know what is apart of the metadata field?

src/plugins/data/public/data_sources/datasource_selector/types.ts
paulstn: agreed, i put these names here as temporary placeholders. currently what i'm planning on implementing is a system where we have 'default' for OS clusters (mds for example), and other types that get set by plugins are treated as foreign. This connectionType field would be removed in that case

src/plugins/data/public/data_sources/datasource/types.ts
paulstn: yeah that's how it would be done in this implementation, in query enhancers we would specify the connection type during registration. i think that this should be changed though, it would be a redundant field since we could figure this out by seeing if its 'default' or not

src/plugins/data/common/search/search_source/types.ts
paulstn: it definitely didn't conflict or raise an issue at the time, i'll follow back on all of these import type things at the end since im iterating pretty quickly

src/plugins/data/common/search/search_source/fetch/get_search_params.ts
paulstn: agreed but since the dataframe isn't generated by the data plugin initially, it would have to go through the search request for now. might need to be changed when stacked queries are designed and implemented

BionIT:
Hi @paulstn , is this PR targeting 2.15 or 2.16?

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -72,7 +72,7 @@ export const useSearch = (services: DiscoverViewServices) => {
   const [savedSearch, setSavedSearch] = useState<SavedSearch | undefined>(undefined);
   const { savedSearch: savedSearchId, sort, interval } = useSelector((state) => state.discover);
   const { data, filterManager, getSavedSearchById, core, toastNotifications, chrome } = services;
-  const indexPattern = useIndexPattern(services);
+  const { indexPattern, dataSource } = useIndexPattern(services);
```
paulstn, line 75: @kavilla I was using the index pattern hook to pass along the data source information as a temporary solution, should I create another hook for datasource or generalize this index pattern hook?

BionIT:
@paulstn is the PR ready for review?

src/plugins/data/public/data_sources/constants.tsx
```diff
@@ -6,9 +6,7 @@
 import { i18n } from '@osd/i18n';
 import { DataSourceUIGroupType } from './datasource/types';
 
-export const S3_GLUE_DATA_SOURCE_DISPLAY_NAME = 'Amazon S3';
-export const S3_GLUE_DATA_SOURCE_TYPE = 's3glue';
-export const DEFAULT_DATA_SOURCE_TYPE = 'DEFAULT_INDEX_PATTERNS';
+export const DEFAULT_DATA_SOURCE_TYPE = 'default';
```
kavilla, line 9: does this have functional purposes outside of redirecting?

src/plugins/data/common/search/search_source/fetch/get_search_params.ts
kavilla: i think we got something working with sean's stuff for this: https://github.com/sejli/OpenSearch-Dashboards/tree/async-polling-3 so we can unite the logic post merge.

kavilla (approved):
merging as is! working for sean's stuff. ty!

Issue #6835, opened 2024-05-24
[BUG] Set default data source unstable issue


zhyuanqi:
**Describe the bug**
The first created datasource should be automatically set as the default datasource. However, this feature is unstable. Sometimes the first created datasource has not been selected as default datasource. It is flaky

**To Reproduce**
1. Go to management page
2. Click on Datasources ---> create datasource
3. Create datasource and save datasource
4. Datasource has been saved, but it has not been selected as default datasource

**Expected behavior**
The first created datasource should be automatically set as the default datasource. 

**OpenSearch Version**
OS_2.14

**Dashboards Version**
OSD_2.14

BionIT:
@zhyuanqi @yujin-emma Is the tag correct? We are targeting to fix in 2.15, right?

PR #6837, opened 2024-05-24
[MDS]Update the default data source set logic to make it stable

yujin-emma:
### Description

Update the logic to set default data source when create data source and multi delete data source on data source table page.
The reason for being unstable to set default data source from UI is that the uiSetting maintain a global map, while it can cache for some reason, when the cache is not cleaning, even the first data source can be created successfully and should be set to default data source, the cache will block it.
The change in this PR will set default data source when there is only one item from the getDataSource list ( another explaining of setting the first data source to be the default data source )

The multi delete behave correctly with a refresh, the change in this PR does not need a refresh

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6835

## Screenshot
Create the first data source, it should be the default data source
create
When delete the default data source, the next one will be the default data source
delete

## Changelog
- fix: Make default data source set stable

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
```diff
@@ -250,10 +250,13 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
...
   const setDefaultDataSource = async () => {
     try {
-      for (const dataSource of selectedDataSources) {
-        if (uiSettings.get('defaultDataSource') === dataSource.id) {
-          await setFirstDataSourceAsDefault(savedObjects.client, uiSettings, true);
-        }
+      const restDataSources = dataSources.filter(
+        (dataSource) => !selectedDataSources.includes(dataSource)
+      );
+      const restDataSourceIds = restDataSources.map((item) => item.id);
```
zhongnansu, line 256: nit: can we chain `filter()` and `map` commands, since I don't see `restDataSources ` being used elsewhere.

BionIT:
> should be set to default data source, the cache will block it.
 just want to make sure I understand it correctly:
1. can we refer to the line of code where the cache would block the setting?
2. if the cache is the problem, can we fix it?

yujin-emma:
> > should be set to default data source, the cache will block it.
> 
> just want to make sure I understand it correctly:
> 
> 1. can we refer to the line of code where the cache would block the setting?
> 2. does the block happen only when the first data source is added?
> 3. if the cache is the problem, can we fix it?

The cache issue cause the behavior is not stable, sometimes can work and sometimes not, the code change here is to using data source list as the source of truth, not depends on how uiSettings cache it.
The cache can be a issue when create and delete ( not yet found issue when editing )

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
yujin-emma: Updated

BionIT:
Hi @yujin-emma, are we targeting to fix this in 2.15?

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
```diff
@@ -255,10 +255,12 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
...
-      for (const dataSource of selectedDataSources) {
-        if (getDefaultDataSourceId(uiSettings) === dataSource.id) {
-          await setFirstDataSourceAsDefault(savedObjects.client, uiSettings, true);
-        }
+      const restDataSourceIds = dataSources
+        .filter((dataSource) => !selectedDataSources.includes(dataSource))
+        .map((item) => item.id);
+      if (restDataSourceIds.length === 0) uiSettings.remove('defaultDataSource');
+      if (!restDataSourceIds.includes(uiSettings.get('defaultDataSource'))) {
+        uiSettings.set('defaultDataSource', restDataSourceIds[0]);
```
zhyuanqi, line 263: Hi Emma. in the onClickDelete(), we have setSelectedDataSources([]); I understand the state changes are asynchronous and react tends to batch multiple state updates. But there is still an possibility the selectedDataSource has been updated to []. Also, if we remove the setFirstDataSourceAsDefault, we can remove the async. 
```
    deleteMultipleDataSources(savedObjects.client, selectedDataSources)
      .then(() => {
        setSelectedDataSources([]);
        // Fetch data sources
        fetchDataSources();
        setConfirmDeleteVisible(false);
        // Check if default data source is deleted or not.
        // if yes, then set the first existing datasource as default datasource.
        setDefaultDataSource();
      })
```

zhyuanqi: Let's sync this offline maybe?

yujin-emma:
@BionIT will have another PR to fix it, close this one

PR #6852, opened 2024-05-29
[WIP] experimental placeholder

RyanL1997:
Signed-off-by: Ryan Liang <jiallian@amazon.com>
### Description

Create a centralized datasource panel for consuming both Zero-ETL datasources and OpenSearch datasources

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
Signed-off-by: Ryan Liang <jiallian@amazon.com>


kavilla:
> Do we have a design for this? Im concerned that we are hardcoding datasources logic here. Ideally the code in OSD should be limited to features that are natively available in OS core and all additional functionality should be moved to a plugin and brought in using core OSD apis. See how the Multi Query language feature brings in additional query language support using such API's

+1.

Will look at this PR tmrw. One general feedback would be it seems like very specific components that can be genericized and registered to

src/plugins/data_source_management/public/components/flint_data_sources_components/icons/query_assistant_logo.svg
kavilla, line 1: this is one that i kind of more of less stand by, in a draft pr, im gunna raise a temp plugin into the feature branch that should be a 3rd party. but one thing i noticed is that there is a central plugin for icons that dont exist in OUI so then i see multiple plugins add this same svg but not identical. this gives the feeling of disjointed when navigating to different parts of the applications where the svg is slightly different.

there needs a central location, a potential second-party type plugin folder could be a candidate. but like i said will raise a draft pr explaining

src/plugins/data_source_management/public/types.ts
```diff
@@ -140,3 +140,8 @@ export {
   SigV4Content,
   DataSourceAttributes,
 } from '../../data_source/common/data_sources';
+
+// flint datasources types
+export type DatasourceType = 'S3GLUE' | 'PROMETHEUS';
+
+export type DatasourceStatus = 'ACTIVE' | 'DISABLED';
```
kavilla, line 147: are these the only status? and i don't know if these status are unique to datasources

src/plugins/data_source_management/public/components/data_source_home_panel/data_source_home_panel.tsx
```diff
@@ -0,0 +1,73 @@
...
+  EuiPageHeader,
+  EuiPanel,
+} from '@elastic/eui';
+import { DataSourceHeader } from './data_source_page_header';
+import { DataSourceTableWithRouter } from '../data_source_table/data_source_table';
+import { ManageFlintDataConnectionsTable } from '../flint_data_sources_components/flint_data_connection/manage_flint_data_connections_table';
+
+const tabs = [
+  {
+    id: 'manageFlintDataSources',
```
kavilla, line 23: nit: within this repo do we have a clear explanation of Flint? searching on google I find some explanation: https://gist.github.com/rupeshtiwari/7d40764617d97adc78ba39654fdb0827

But it could be confusing for contributors who fork this repo as there isn't a clear definition of what Flint is housed in this repo.

Flint appears to be a service that community members that don't have immediate access to without going through a provider.

Can a consideration here being, auditing what is not currently available with the data sources registration. like if i needs a categorization property that as a provider can add to a plugin to register a category `flint` that has specific data sources?

if Flint is technically accessible for open source, can we be more descriptive in the functional components behind it. if it's for query acceleration then although lengthy, it wouldn't require a deep dive in what components is what in the future.

src/plugins/data_source_management/public/components/constants.tsx
```diff
@@ -17,3 +18,29 @@ export const NO_DATASOURCES_CONNECTED_MESSAGE = 'No data sources connected yet.'
 export const CONNECT_DATASOURCES_MESSAGE = 'Connect your data sources to get started.';
 export const NO_COMPATIBLE_DATASOURCES_MESSAGE = 'No compatible data sources are available.';
 export const ADD_COMPATIBLE_DATASOURCES_MESSAGE = 'Add a compatible data source.';
+
+export const OPENSEARCH_DOCUMENTATION_URL =
+  'https://opensearch.org/docs/latest/dashboards/management/data-sources/';
```
kavilla, line 23: don't do this. use doc link service: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/doc_links/doc_links_service.ts

it's versioned

src/plugins/data_source_management/public/components/constants.tsx
```diff
@@ -17,3 +18,29 @@ export const NO_DATASOURCES_CONNECTED_MESSAGE = 'No data sources connected yet.'
...
+
+export const PrometheusURL = 'Prometheus';
+export const AmazonS3URL = 'AmazonS3AWSGlue';
+
+export const UrlToDatasourceType: { [key: string]: DatasourceType } = {
+  [PrometheusURL]: 'PROMETHEUS',
+  [AmazonS3URL]: 'S3GLUE',
+};
+
+export type AuthMethod = 'noauth' | 'basicauth' | 'awssigv4';
```
kavilla, line 46: not seeing this being used yet. but is this to be used later when integrating with the existing auth work done by MDS? https://github.com/opensearch-project/OpenSearch-Dashboards/blob/cf4e496aa03d12b18244f271185863f5fbf1ff6d/src/plugins/data_source/public/plugin.ts#L27.

if not, we should get a design proposal, as this might be doing a rehash of heavy lifting done by that work which also has saved objects integration.

src/plugins/data_source_management/public/components/flint_data_sources_components/flint_data_connection/delete_modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+  EuiFormRow,
+  EuiModalBody,
+  EuiModalFooter,
+  EuiModalHeader,
+  EuiModalHeaderTitle,
+  EuiSpacer,
+  EuiText,
+} from '@elastic/eui';
+
+export const DeleteModal = ({
```
kavilla, line 23: regardless of previous ux mocks, some of these components do not seem unique to flint.

data sources can delete already and this should wire in that.

further research should be done to verify if they are using EuiConfirmModal like we are in other parts of the application for a cohesive experience. to avoid the issue of having identical functionalities in the application but are a little different.

BionIT:
Hi @RyanL1997 , is this PR targeting 2.15 or 2.16?

RyanL1997:
> Hi @RyanL1997 , is this PR targeting 2.15 or 2.16?

Hi @BionIT, this is PR will not go into 2.15 release branch.

PR #6894, opened 2024-06-03
[Discover-next] Add search bar extensions

joshuali925:
### Description

A search bar extension can display a UI component above the query bar. The component has the ability to read and write discover search bar states to enhance the search experience for users. The configuration is part of Query Enhancements.

```ts
export interface SearchBarExtensionDependencies {
  /**
   * Currently selected index patterns.
   */
  indexPatterns?: IIndexPattern[];
}

export interface SearchBarExtensionConfig {
  /**
   * The id for the search bar extension.
   */
  id: string;
  /**
   * Lower order indicates higher position on UI.
   */
  order: number;
  /**
   * A function that determines if the search bar extension is enabled and should be rendered on UI.
   * @returns whether the extension is enabled.
   */
  isEnabled: () => Promise<boolean>;
  /**
   * A function that returns the mount point for the search bar extension.
   * @param dependencies - The dependencies required for the extension.
   * @returns The mount point for the search bar extension.
   */
  getComponent: (dependencies: SearchBarExtensionDependencies) => React.ReactElement;
}

export interface QueryEnhancement {
  ...
  searchBar?: {
    ...
    extensions?: SearchBarExtensionConfig[];
  };
}
```

### Issues Resolved

Developers can utilize search bar extensions to add additional features to the search bar, such as query assist.

Part of #6820 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #6895, opened 2024-06-03
[Discover-next] add query assist to query enhancements plugin

### LINKED ISSUE #6820: Support query assist in discover

In log explorer there's a feature to generate PPL queries through natural language. But in discover user has to manually write the queries, this feature does not exist there.

Add query assist to discover.

**Additional context**

TBD

joshuali925:
### Description

This PR is working in progress.

This PR contains https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6894. It adds query assist specific logic in query enhancements plugin to show a UI above the PPL search bar if user has configured PPL agent.

### Issues Resolved

#6820 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


plugins-extra/query_enhancements/public/query_assist/components/query_assist_bar.tsx
```diff
@@ -0,0 +1,95 @@
...
+      setCallOutType('empty_index');
+      return;
+    }
+    dismissCallout();
+    previousQuestionRef.current = inputRef.current.value;
+    persistedLog.add(inputRef.current.value);
+    const params = {
+      question: inputRef.current.value,
+      index: selectedIndex,
+      language: 'PPL',
```
kavilla, line 55: Future proofing, could we make this more generic even if it only works for PPL right now?

The last query in the query service should have the language attached to it assuming that is the language. And even more straight forward approach and preferred could be pulling from localstorage:
Screenshot 2024-06-04 at 14-12-41 (11) Discover UI Updates - Quip

which you can access from the ui service in the data plugin: 
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/ui/ui_service.ts#L65

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/ui/settings/settings.ts#L28

plugins-extra/query_enhancements/public/query_assist/utils/create_extension.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { HttpSetup } from 'opensearch-dashboards/public';
+import { QueryAssistBar } from '../components';
+import { SearchBarExtensionConfig } from '../../../../../src/plugins/data/public/ui/search_bar_extensions';
+
+export const createQueryAssistExtension = (http: HttpSetup): SearchBarExtensionConfig => {
+  return {
+    id: 'query-assist-ppl',
+    order: 1000,
+    isEnabled: (() => {
+      let agentConfigured: boolean;
```
kavilla, line 11: do we want to pull from advanced settings to be able to catch all disable extensions?

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/server/ui_settings.ts

plugins-extra/query_enhancements/public/query_assist/utils/get_persisted_log.ts
```diff
@@ -0,0 +1,19 @@
...
+import { UI_SETTINGS } from '../../../../../src/plugins/data/common';
+import { PersistedLog } from '../../../../../src/plugins/data/public';
+import { IStorageWrapper } from '../../../../../src/plugins/opensearch_dashboards_utils/public';
+
+export function getPersistedLog(
+  uiSettings: IUiSettingsClient,
+  storage: IStorageWrapper,
+  language: string
+) {
+  return new PersistedLog(
```
kavilla, line 11: nice

plugins-extra/query_enhancements/public/query_assist/components/query_assist_input.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface QueryAssistInputProps {
+  inputRef: React.RefObject<HTMLInputElement>;
+  persistedLog: PersistedLog;
+  initialValue?: string;
+  selectedIndex?: string;
+  previousQuestion?: string;
+}
+
+export const QueryAssistInput: React.FC<QueryAssistInputProps> = (props) => {
```
kavilla, line 18: we are just using recent suggestions right now right?

i'd imagine its similar to assistant dashboards the initial will keep track of the suggestions as they get updated for reference? https://github.com/opensearch-project/dashboards-assistant/blob/main/public/services/incontext_insight/incontext_insight_registry.ts#L63

kavilla: or is the plan that we will just persist it

plugins-extra/query_enhancements/public/assets/query_assist_logo.svg
kavilla, line 1: completely unrelated to this PR and not a blocker for it but assistance in tracking would be great since a couple plugins define their own svg for this and breaks the cohesion. 

imo, i'm on the side of making this logo: https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/src/core/server/core_app/assets/logos if it's not apart of oui. and add it apart of this service: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/common/logos/get_logos.ts#L52

i dont think it's so maybe one of the action items for us to add it to the opensearch logos: https://oui.opensearch.org/1.6/#/display/icons#opensearch-logos

as i wonder if we have a dark mode logo of that as well. or if it is expected to use both

plugins-extra/query_enhancements/public/query_assist/components/submit_button.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const QueryAssistSubmitButton: React.FC<SubmitButtonProps> = (props) => {
+  return (
+    <EuiButtonIcon
+      iconType="returnKey"
+      display="base"
+      isDisabled={props.isDisabled}
+      size="s"
+      type="submit"
+      aria-label="submit-question"
```
kavilla, line 16: do we want to make this more detailed? imo i think it's fine as is for [accessibility](https://oui.opensearch.org/1.6/#/guidelines/accessibility) purposes but i'm not expert. perhaps we can get u/x feedback on what the aria label should be. like should be "Submit question to query assistant".

plugins-extra/query_enhancements/public/query_assist/components/query_assist_input.tsx
```diff
@@ -0,0 +1,75 @@
...
+      .map((recentSearch) => ({
+        type: QuerySuggestionTypes.RecentSearch,
+        text: recentSearch,
+        start: 0,
+        end: value.length,
+      }));
+  }, [props.persistedLog, value]);
+
+  return (
+    <EuiOutsideClickDetector onOutsideClick={() => setIsSuggestionsVisible(false)}>
```
kavilla, line 37: should we check with ux to see if they want us to also submit the question if set on an outside click?

plugins-extra/query_enhancements/server/routes/query_assist/agents.ts
```diff
@@ -0,0 +1,62 @@
+import { ApiResponse } from '@opensearch-project/opensearch';
+import { RequestBody, TransportRequestPromise } from '@opensearch-project/opensearch/lib/Transport';
+import { RequestHandlerContext } from 'src/core/server';
+
+const ML_COMMONS_API_PREFIX = '/_plugins/_ml';
```
kavilla, line 5: should we place this here: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/feature/discover-next/plugins-extra/query_enhancements/common/index.ts

plugins-extra/query_enhancements/server/routes/query_assist/agents.ts
```diff
@@ -0,0 +1,62 @@
...
+const AGENT_REQUEST_OPTIONS = {
```
kavilla, line 6: this should be an advanced setting and pull from that: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/server/ui_settings.ts

i will work on adding the base one and people can add to it

plugins-extra/query_enhancements/public/query_assist/components/call_outs.tsx
```diff
@@ -0,0 +1,79 @@
...
+    iconType="alert"
+    dismissible
+    onDismiss={props.onDismiss}
+  />
+);
+
+const EmptyQueryCallOut: React.FC<CalloutDismiss> = (props) => (
+  <EuiCallOut
+    data-test-subj="query-assist-empty-query-callout"
+    title="Enter a natural language question to automatically generate a query to view results."
```
kavilla, line 44: do these need i18n and aria-labels.

kavilla (approved):

plugins-extra/query_enhancements/public/query_assist/components/query_assist_bar.tsx
joshuali925: good point i missed moving this to a variable

query assist can run before before any query runs, would localStorage be set after changing language or executing query? I tested on my side but didn't see this key in localStorage

i think i can still pass in language when creating the query assist search bar extension, since currently they are language specific

plugins-extra/query_enhancements/public/query_assist/components/query_assist_input.tsx
joshuali925: currently just recent suggestions, maybe plus some hard-coded suggestions for sample data.

i didn't really understand the difference,

> initial will keep track of the suggestions as they get updated for reference

is this for dedup purpose, or when user edits a question and submit, the recent suggestions gets edited instead of adding a new one?

plugins-extra/query_enhancements/public/assets/query_assist_logo.svg
joshuali925: i agree it would be good to put logo assets together to enforce a standard. right now i only have this one svg that's used for both themes

plugins-extra/query_enhancements/public/query_assist/components/query_assist_input.tsx
joshuali925: then if user clicks the data-grid to copy a field, since it's an outside click, their incomplete question will be submitted. i think that can be weird?

plugins-extra/query_enhancements/public/query_assist/components/call_outs.tsx
joshuali925: good point, yes

PR #6906, opened 2024-06-04
[Discover] Async Queries

sejli:
### Description

Initial commits for direct queries.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/plugins-extra/query_enhancements/public/search/sql_async_search_interceptor.ts
+++ b/plugins-extra/query_enhancements/public/search/sql_async_search_interceptor.ts
@@ -1,0 +1,91 @@
+import { trimEnd } from 'lodash';
+import { Observable, from } from 'rxjs';
+import { stringify } from '@osd/std';
+import { i18n } from '@osd/i18n';
+import {
+  DataPublicPluginStart,
+  IOpenSearchDashboardsSearchRequest,
+  IOpenSearchDashboardsSearchResponse,
+  ISearchOptions,
+  SearchInterceptor,
+  SearchInterceptorDeps,
+} from '../../../../src/plugins/data/public';
+import { SQL_SEARCH_STRATEGY } from '../../common';
+import { QueryEnhancementsPluginStartDependencies } from '../types';
+
+export class SQLAsyncQlSearchInterceptor extends SearchInterceptor {
+  protected queryService!: DataPublicPluginStart['query'];
+  protected aggsService!: DataPublicPluginStart['search']['aggs'];
+
+  constructor(deps: SearchInterceptorDeps) {
+    super(deps);
+
+    deps.startServices.then(([coreStart, depsStart]) => {
+      this.queryService = (depsStart as QueryEnhancementsPluginStartDependencies).data.query;
+      this.aggsService = (depsStart as QueryEnhancementsPluginStartDependencies).data.search.aggs;
+    });
+  }
+
+  protected runSearch(
+    request: IOpenSearchDashboardsSearchRequest,
+    signal?: AbortSignal,
+    strategy?: string
+  ): Observable<IOpenSearchDashboardsSearchResponse> {
+    const { id, ...searchRequest } = request;
+    const path = trimEnd('/api/sqlasyncql/jobs');
+
+    const fetchDataFrame = (queryString: string, df = null, session = undefined) => {
+      const body = stringify({
+        query: { qs: queryString, format: 'jdbc' },
+        df,
+        sessionId: session,
+      });
+      return from(
+        this.deps.http.fetch({
+          method: 'POST',
+          path,
+          body,
+          signal,
+        })
+      );
+    };
+
+    const fetchJobStatusDataFrame = (queryId: string) => {
+      return from(
+        this.deps.http.fetch({
+          method: 'GET',
+          path: `${path}/${queryId}`,
+        })
+      );
+    };
+
+    let dataFrame;
+    if (searchRequest.params.body.query.queries[0]?.queryId) {
+      dataFrame = fetchJobStatusDataFrame(searchRequest.params.body.query.queries[0]?.queryId);
+    } else {
+      dataFrame = fetchDataFrame(
+        searchRequest.params.body.query.queries[0].query,
+        searchRequest.params.body.df,
+        searchRequest.params.body.query.queries[0].sessionId
+      );
+    }
+
+    // subscribe to dataFrame to see if an error is returned, display a toast message if so
+    dataFrame.subscribe((df) => {
+      if (!df.body.error) return;
+      const jsError = new Error(df.body.error.response);
+      this.deps.toasts.addError(jsError, {
+        title: i18n.translate('dqlPlugin.sqlQueryError', {
+          defaultMessage: 'Could not complete the SQL async query',
+        }),
+        toastMessage: df.body.error.msg,
+      });
+    });
+
+    return dataFrame;
+  }
+
+  public search(request: IOpenSearchDashboardsSearchRequest, options: ISearchOptions) {
+    return this.runSearch(request, options.abortSignal, SQL_SEARCH_STRATEGY);
+  }
+}
--- a/plugins-extra/query_enhancements/public/search/sql_async_search_interceptor.ts
+++ b/plugins-extra/query_enhancements/public/search/sql_async_search_interceptor.ts
@@ -10,7 +10,12 @@ import {
   SearchInterceptor,
   SearchInterceptorDeps,
 } from '../../../../src/plugins/data/public';
-import { SQL_SEARCH_STRATEGY } from '../../common';
+import {
+  getRawDataFrame,
+  getRawQueryId,
+  getRawQueryString,
+} from '../../../../src/plugins/data/common';
+import { SQL_ASYNC_SEARCH_STRATEGY } from '../../common';
 import { QueryEnhancementsPluginStartDependencies } from '../types';
 
 export class SQLAsyncQlSearchInterceptor extends SearchInterceptor {
@@ -34,11 +39,10 @@ export class SQLAsyncQlSearchInterceptor extends SearchInterceptor {
     const { id, ...searchRequest } = request;
     const path = trimEnd('/api/sqlasyncql/jobs');
 
-    const fetchDataFrame = (queryString: string, df = null, session = undefined) => {
+    const fetchDataFrame = (queryString: string | undefined, df = null) => {
       const body = stringify({
         query: { qs: queryString, format: 'jdbc' },
         df,
-        sessionId: session,
       });
       return from(
         this.deps.http.fetch({
@@ -60,14 +64,12 @@ export class SQLAsyncQlSearchInterceptor extends SearchInterceptor {
     };
 
     let dataFrame;
-    if (searchRequest.params.body.query.queries[0]?.queryId) {
-      dataFrame = fetchJobStatusDataFrame(searchRequest.params.body.query.queries[0]?.queryId);
+    const rawQueryId = getRawQueryId(searchRequest);
+    const rawDataFrame = getRawDataFrame(searchRequest);
+    if (rawQueryId) {
+      dataFrame = fetchJobStatusDataFrame(rawQueryId);
     } else {
-      dataFrame = fetchDataFrame(
-        searchRequest.params.body.query.queries[0].query,
-        searchRequest.params.body.df,
-        searchRequest.params.body.query.queries[0].sessionId
-      );
+      dataFrame = fetchDataFrame(getRawQueryString(searchRequest), rawDataFrame);
     }
 
     // subscribe to dataFrame to see if an error is returned, display a toast message if so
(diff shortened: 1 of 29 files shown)


plugins-extra/query_enhancements/public/search/sql_async_search_interceptor.ts
```diff
@@ -0,0 +1,91 @@
...
+        this.deps.http.fetch({
+          method: 'GET',
+          path: `${path}/${queryId}`,
+        })
+      );
+    };
+
+    let dataFrame;
+    if (searchRequest.params.body.query.queries[0]?.queryId) {
+      dataFrame = fetchJobStatusDataFrame(searchRequest.params.body.query.queries[0]?.queryId);
```
kavilla, line 64: can use this helper function? https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/common/data_frames/utils.ts#L39

plugins-extra/query_enhancements/public/plugin.ts
```diff
@@ -72,6 +80,26 @@ export class QueryEnhancementsPlugin
...
 
+    data.__enhance({
+      ui: {
+        query: {
+          language: 'SQL Async',
+          search: sqlAsyncSearchInterceptor,
+          searchBar: {
+            showDatePicker: false,
+            showFilterBar: false,
+            queryStringInput: { initialValue: 'SHOW DATABASES IN mys3' },
```
kavilla, line 91: Could this be `<data_source>`?

plugins-extra/query_enhancements/public/search/sql_async_search_interceptor.ts
```diff
@@ -0,0 +1,91 @@
...
+        }),
+        toastMessage: df.body.error.msg,
+      });
+    });
+
+    return dataFrame;
+  }
+
+  public search(request: IOpenSearchDashboardsSearchRequest, options: ISearchOptions) {
+    return this.runSearch(request, options.abortSignal, SQL_SEARCH_STRATEGY);
```
kavilla, line 89: is the right search strategy?

plugins-extra/query_enhancements/server/search/sql/sql_async_search_strategy.ts
```diff
@@ -0,0 +1,76 @@
...
+        let rawResponse: any;
+        if (request.params?.queryId) {
+          rawResponse = await sqlAsyncJobsFacet.describeQuery(request);
+        } else {
+          request.body.query = request.body.query.qs;
+          rawResponse = await sqlAsyncFacet.describeQuery(request);
+        }
+        if (!rawResponse.success) {
+          return {
+            type: 'data_frame',
```
kavilla, line 34: should this be data_frame_polling?

src/plugins/data/common/data_frames/utils.ts
```diff
@@ -472,3 +472,77 @@ export const dataFrameToSpec = (dataFrame: IDataFrame, id?: string): IndexPatter
     fields: fields.reduce(flattenFields, {} as IndexPatternFieldMap),
   };
 };
+
+type FetchFunction<T, P = void> = (params?: P) => Promise<T>;
```
kavilla, line 476: can this be in types?

src/plugins/data/common/data_frames/utils.ts
```diff
@@ -472,3 +472,77 @@ export const dataFrameToSpec = (dataFrame: IDataFrame, id?: string): IndexPatter
...
+
+export interface PollingConfigurations {
+  tabId: string;
+}
+
+export class UsePolling<T, P = void> {
```
kavilla, line 482: i think this could extend the partial data frame if not then this could fits more of a function. 

or avoid the usage of `Use` to avoid confusion with https://react.dev/reference/react/use

src/plugins/data/common/data_frames/utils.ts
```diff
@@ -472,3 +472,77 @@ export const dataFrameToSpec = (dataFrame: IDataFrame, id?: string): IndexPatter
...
+  public data: T | null = null;
+  public error: Error | null = null;
+  public loading: boolean = true;
+  private shouldPoll: boolean = false;
+  private intervalRef?: NodeJS.Timeout;
+
+  constructor(
+    private fetchFunction: FetchFunction<T, P>,
+    private interval: number = 5000,
```
kavilla, line 491: can we make sure we add an advanced setting for this?

src/plugins/data/common/data_frames/utils.ts
```diff
@@ -472,3 +472,77 @@ export const dataFrameToSpec = (dataFrame: IDataFrame, id?: string): IndexPatter
...
+      this.loading = false;
+
+      if (this.onPollingError && this.onPollingError(this.error)) {
+        this.stopPolling();
+      }
+    }
+  }
+
+  async startPolling(params?: P) {
+    this.shouldPoll = true;
```
kavilla, line 518: how come we have a boolean here vs using the interval?

src/plugins/data/common/data_frames/utils.ts
```diff
@@ -472,3 +472,77 @@ export const dataFrameToSpec = (dataFrame: IDataFrame, id?: string): IndexPatter
...
+    }
+  }
+
+  async waitForPolling(): Promise<void> {
+    return new Promise<void>((resolve) => {
+      const checkLoading = () => {
+        if (!this.loading) {
+          resolve();
+        } else {
+          setTimeout(checkLoading, 5000);
```
kavilla, line 542: is this the timeout logic? can we get an advanced setting for this as well? but could you explain if when we are timing out? if it is loading and then 5 seconds?

src/plugins/data/common/search/search_source/search_source.ts
```diff
@@ -412,6 +455,50 @@ export class SearchSource {
           const dataFrameResponse = response as IDataFrameResponse;
           await this.setDataFrame(dataFrameResponse.body as IDataFrame);
           return onResponse(searchRequest, convertResult(response as IDataFrameResponse));
+        } else if ((response as IDataFrameResponse).type === DATA_FRAME_TYPES.POLLING) {
+          const dataFrameResponse = response as IDataFrameResponse;
+          const responseQueryId = (dataFrameResponse.body as IDataFrame).meta?.queryId;
+          const responseSessionId = (dataFrameResponse.body as IDataFrame).meta?.sessionId;
+          // if there wasn't a session before, set current datasource as session
+          // TODO: MQL update with selected datasource
```
kavilla, line 463: i know it probably isnt the best supported with the data source being specific for mys3 but i think we should be able to use dataframe.name as right now that is the datasource. and then for demo purposes we can modify the read only advanced setting

src/plugins/data/common/search/search_source/search_source.ts
```diff
@@ -412,6 +455,50 @@ export class SearchSource {
...
+          if (!sessionId || (sessionId && sessionId !== responseSessionId)) {
+            this.setSession('mys3', responseSessionId);
+          }
+          const queryIdRequest: ISearchRequestParams = {
```
kavilla, line 467: i think this logic should exist in a dataframe polling or in the search strategy can handle this logic since it will get the search request anyways.

plugins-extra/query_enhancements/server/routes/index.ts
```diff
@@ -91,4 +91,75 @@ export function defineRoutes(
...
+  router.post(
+    {
+      path: `/api/sqlasyncql/jobs`,
+      validate: {
+        body: schema.object({
+          query: schema.object({
+            qs: schema.string(),
+            format: schema.string(),
+          }),
+          df: schema.any(),
```
kavilla, line 105: i think we can let the df.meta.sessionId keep responsible for this.

if it has a session id then we just pull it and use it.

plugins-extra/query_enhancements/server/search/async/sql_async_facet.ts
```diff
@@ -0,0 +1,42 @@
...
+  private fetch = async (request: any, format: string, responseFormat: string) => {
+    const res = {
+      success: false,
+      data: {},
+    };
+    try {
+      const params = {
+        body: {
+          query: request.body.query,
+          datasource: 'mys3',
```
kavilla, line 20: we should only define this one spot. the dataframe.name can probably have this and pass here right?

plugins-extra/query_enhancements/public/search/sql_async_search_interceptor.ts
```diff
@@ -0,0 +1,91 @@
...
+        searchRequest.params.body.query.queries[0].sessionId
+      );
+    }
+
+    // subscribe to dataFrame to see if an error is returned, display a toast message if so
+    dataFrame.subscribe((df) => {
+      if (!df.body.error) return;
+      const jsError = new Error(df.body.error.response);
+      this.deps.toasts.addError(jsError, {
+        title: i18n.translate('dqlPlugin.sqlQueryError', {
```
kavilla, line 78: namespace is probably broken for the rest of the query enhancements work but for i18n the namespace has to match the plugin and the path dqlPlugin will probably fail the extraction script

plugins-extra/query_enhancements/public/plugin.ts
```diff
@@ -72,6 +80,26 @@ export class QueryEnhancementsPlugin
       },
     });
 
+    data.__enhance({
+      ui: {
+        query: {
+          language: 'SQL Async',
```
kavilla, line 86: did we verify the language with spaces has no issues?

plugins-extra/query_enhancements/public/plugin.ts
```diff
@@ -31,6 +31,14 @@ export class QueryEnhancementsPlugin
       usageCollector: data.search.usageCollector,
     });
 
+    const sqlAsyncSearchInterceptor = new SQLAsyncQlSearchInterceptor({
+      toasts: core.notifications.toasts,
+      http: core.http,
+      uiSettings: core.uiSettings,
+      startServices: core.getStartServices(),
+      usageCollector: data.search.usageCollector,
```
kavilla, line 39: we should use the usagecollector to see if they optin into using this async queries. ask me if you ened more details

kavilla:
Looks awesome! Thanks for this.

But we should take all the async sql specific stuff out of the search source. rely on the search strategy to handle that logic.

BionIT:
Hi @sejli , is this PR targeting 2.15 or 2.16?

sejli:
Hi @BionIT, this feature is targeting `2.15` release, but this PR is meant to go into the `feature/discover-next` branch.

plugins-extra/query_enhancements/public/plugin.ts
sejli: I do get errors when the language is `SQL Async` and `SQL_Async`, but I think they might be due to the index not being set correctly and other miscellaneous things related to the datasources type not being in this branch. Will verify

plugins-extra/query_enhancements/public/search/sql_async_search_interceptor.ts
sejli: Should be `SQL_ASYNC_SEARCH_STRATEGY`. Funny enough, I don't think any of the search interceptors use the strategy there, like [ppl_search_interceptor.ts](https://github.com/sejli/OpenSearch-Dashboards/blob/e2891fef9e738bee44884db2f908c09726dff8f3/plugins-extra/query_enhancements/public/search/ppl_search_interceptor.ts#L179) and [sql_search_interceptor](https://github.com/sejli/OpenSearch-Dashboards/blob/e2891fef9e738bee44884db2f908c09726dff8f3/plugins-extra/query_enhancements/public/search/sql_search_interceptor.ts#L70) (`strategy` parameter in `runSearch` function isn't used.

plugins-extra/query_enhancements/public/search/sql_async_search_interceptor.ts
sejli: didn't realize there were helper functions, thanks for the tip

src/plugins/data/common/data_frames/utils.ts
sejli: Should be lifted straight from [`use_polling.ts`](https://github.com/opensearch-project/dashboards-observability/blob/main/public/components/hooks/use_polling.ts) in `dashboards-observability`.

src/plugins/data/common/data_frames/utils.ts
sejli: This is the logic I used to wait for the polling response, was having issues before where the UI wouldn't display the data since it was polling. I can add a timeout though to cancel long queries

src/plugins/data/common/search/search_source/search_source.ts
sejli: Was having issues passing this in, since we need the `setSession` function from `SearchService`. Is there a way to call the session cache from `SearchService` somewhere in the search interceptor or strategy?

sejli:
Closing for #6949

PR #6950, opened 2024-06-06
Cleanup yarn.lock

AMoo-Miki:
### Description

Replaced version spam in the lockfile with the greatest one used for each pattern to combine entries.

`Do not backport; do not merge before 2.15.0 is released to prevent complications.`

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/package.json
+++ b/package.json
@@ -373,7 +373,7 @@
     "chance": "1.0.18",
     "cheerio": "1.0.0-rc.1",
     "chromedriver": "^121.0.1",
-    "classnames": "2.3.1",
+    "classnames": "^2.3.1",
     "compare-versions": "3.5.1",
     "cypress": "9.5.4",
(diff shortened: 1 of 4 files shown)


ananzh:
Why only 3.0.0 tag? what is the breaking change here?
The package is in devDependencies so will not release in artifact right? So should not be breaking changes for users.

AMoo-Miki:
> Why only 3.0.0 tag? what is the breaking change here? The package is in devDependencies so will not release in artifact right? So should not be breaking changes for users.

`main` and `2.x` deps have diverged. I am handling each individually.

SuZhou-Joe (approved):

ZilongX (approved):

Issue #6557, opened 2024-04-19
[Multiple Datasource] Do not pass dataSourceEnabled as parameter, determine it from saved object service itself


yujin-emma:
**Is your feature request related to a problem? Please describe.**

dataSourceEnabled should be determined by saved object itself
**Describe the solution you'd like**

Get the dataSourceEnabled value frm yml directly
**Describe alternatives you've considered**

**Additional context**

PR #6961, opened 2024-06-07
[MDS] remove datasourceEnabled in the request

yujin-emma:
### Description

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6557

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6557

## Screenshot

non MDS
non_mds_import

MDS

mds_import

## Changelog
- fix: Remove the dataSourceEnabled from request

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/core/server/saved_objects/import/import_saved_objects.ts
```diff
@@ -60,11 +60,12 @@ export async function importSavedObjectsFromStream({
   dataSourceId,
   dataSourceTitle,
   workspaces,
-  dataSourceEnabled,
 }: SavedObjectsImportOptions): Promise<SavedObjectsImportResponse> {
   let errorAccumulator: SavedObjectsImportError[] = [];
   const supportedTypes = typeRegistry.getImportableAndExportableTypes().map((type) => type.name);
 
+  const dataSourceEnabled = supportedTypes.includes('data-source');
```
BionIT, line 67: can we get it from config? Not sure if I understand what we try to achieve here

yujin-emma: typeRegistry is from the config, this is the savedObjectConfig: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/saved_objects/saved_objects_service.ts#L372

BionIT: We are checking a config not a supported type, right?

yujin-emma: yeah, the supported type is from config

BionIT: How is supported type configured? How is it related to if data source is enabled config?

yujin-emma: yeah, when constructing the saved object client, it will read config from the raw YAML file, and register it, if data source is enabled, it will support data-source type

virajsanghvi: So the existing meaning of `dataSourceEnabled` was that it's a supported type, and now we're just reading it directly?

LDrago27:
@yujin-emma Is this still being targeted for 2.16 ?

PR #6974, opened 2024-06-07
Add Scripted Field Support in Vis Builder

LDrago27:
### Description

Add Scripted Field Support in Vis Builder.    

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/38322563/332a8663-fa5c-4612-aeff-5c6c2564f633

## Testing the changes

1. Create an Index pattern with scripted fields
2. Navigate to Vis Builder and verify if the scripted field is available in the left side panel

## Changelog
- feat: Add Scripted Field Support in Vis Builder

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/vis_builder/public/application/components/data_tab/utils/get_available_fields.ts
+++ b/src/plugins/vis_builder/public/application/components/data_tab/utils/get_available_fields.ts
@@ -17,7 +17,7 @@ export const getAvailableFields = (
   filterFieldTypes: FieldTypes = '*'
 ) => {
   const filteredFields = fields.filter((field: IndexPatternField) => {
-    if (!field.aggregatable || isNestedField(field) || field.scripted) {
+    if (!field.aggregatable || isNestedField(field)) {
       return false;
     }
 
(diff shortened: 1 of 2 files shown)


src/plugins/vis_builder/public/application/components/data_tab/utils/get_available_fields.ts
```diff
@@ -17,7 +17,7 @@ export const getAvailableFields = (
   filterFieldTypes: FieldTypes = '*'
 ) => {
   const filteredFields = fields.filter((field: IndexPatternField) => {
-    if (!field.aggregatable || isNestedField(field) || field.scripted) {
+    if (!field.aggregatable || isNestedField(field)) {
```
virajsanghvi, line 20: Just a question: should this be behind a flag? 

I assume not as I think most customers would expect these to be there, but just checking if we need to manage this change for customers.

LDrago27: @virajsanghvi  In my opinion we don't need a flag here, since this experience is going to be common for all the users and not configurable via flags or settings. This behavior is consistent with the current Discover experience.

virajsanghvi (approved):

ashwin-pc:
@LDrago27 Can you add a test for this?

LDrago27:
> @LDrago27 Can you add a test for this?

@ashwin-pc I have added a unit test specific to this change.

ananzh (approved):

manasvinibs (approved):

ashwin-pc (approved):

Issue #6982, opened 2024-06-10
[BUG] Duplicate timestamp column in legacy table view surrounding document


ananzh:
**Describe the bug**

Duplicate time stamp in surrounding document in legacy table
<img width="1706" alt="Screenshot 2024-06-10 at 9 56 42 AM" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/79961084/adeb9907-191c-4d75-883d-ce56fe60e5d9">

Compared with 2.9
<img width="1742" alt="Screenshot 2024-06-09 at 2 01 48 PM" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/79961084/bc40b1c2-3c8b-4eee-8d82-84fbbecf1d46">

**To Reproduce**
1. Go to 'discover' and use legacy table
2. Expand one row 
3. Click 'View surrounding document'
4. See duplicate column

**Expected behavior**
2.9 behavior

**OpenSearch Version**
2.12+

**Dashboards Version**
2.12+

ananzh:
Resolved.

PR #6586, opened 2024-04-22
[Draft][OE] Adds TS check workflow

ashwin-pc:
### Description

Adds a TS check to the github workflow. This causes increases in TS Errors to be flagged and fail the CI

## Testing the changes

The Typescript Check in the PR's CI should pass

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [x] Commits are signed per the DCO using --signoff

Issue #1660, opened 2022-06-02
[BUG] Typescript compiler output disregarded as part of normal path build commands


CPTNB:
Standard build scripts run tsc but do not fail on or show typescript errors.  This has led to a confusing relationship with typescript and allowed 700+ typescript compilation errors to creep into the project.

### Expected:
You can demonstrate this by adding in the following erroneous code to a .ts file and trying any of the standard build commands (yarn start, yarn build, yarn osd bootstrap).
```
type X = true;
function returnX (): X { }
```
This should generate the error: ```error TS2355: A function whose declared type is neither 'void' nor 'any' must return a value.```

### Actual:
No error is surfaced in the console, and the build commands proceed as if the code compiled correctly.

## Research
It looks like we're effectively using typescript as a babel passthrough.  There are a few scripts in [src/dev/typescript](https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/src/dev/typescript) that look for local tsconfig.json files and execute tsc inside those directories.  I spent some time fiddling with these scripts yesterday and got to the point where I could run tsc against a new plugin with only 9 errors not originating in the new plugin, which feels much more tractable.

I did that by creating a local [tsconfig.json](https://pastebin.com/NTFNhPC2) inside the plugin and running:
```
node scripts/type_check.js --skip-lib-check --project src/plugins/wizard/tsconfig.json 
```
It's not totally 💯 though since I did have to hard-code my baseUrl.  More research needs to be done to see how we can run the typescript compiler targeted at plugins in an automated way.

## Ideal State
I came into the project with the assumption that if I wrote incorrect typescript the project wouldn't build and the CI system would reject my changes.  I think that's a very reasonable assumption, and we should try to meet it.

## Suggested path forward
Since we have quite a few errors it's probably not tractable to fix them all to get back to a completely healthy state.  Also, there may be other prohibitive factors that led previous maintainers to opt out of type checking.

I think we can quickly get to a place where we can opt plugins into type checking via config, then we can incrementally reintroduce type checking.

Another path would be to mark all errors today as known and prevent PRs from adding errors.  This is probably cleaner in aggregate but I imagine it'd take more code to integrate into builds.  It might be as easy as (pseudocode):

```typescript
const tsc = require('typescript');
const allowedErrors = new Set([908127128903, 09812309180923, 109238091823, ...]);
const { errors } = tsc.runOnTheProject();
const newErrors = errors
[...]

PR #6983, opened 2024-06-10
[BUG][Discover] Check if the timestamp is already included to remove duplicate col

ananzh:
### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6982

## Changelog
- fix: [Discover] Check if the timestamp is already included to remove duplicate col

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/components/default_discover_table/helper.tsx
+++ b/src/plugins/discover/public/application/components/default_discover_table/helper.tsx
@@ -94,7 +94,9 @@ export function getLegacyDisplayedColumns(
       colRightIdx: idx + 1 >= columns.length ? -1 : idx + 1,
     };
   });
-  return !hideTimeField && indexPattern.timeFieldName
+  return !hideTimeField &&
+    indexPattern.timeFieldName &&
+    !columns.includes(indexPattern.timeFieldName)
     ? [getTimeColumn(indexPattern.timeFieldName), ...columnProps]
     : columnProps;
 }
--- a/src/plugins/discover/public/application/components/default_discover_table/helper.tsx
+++ b/src/plugins/discover/public/application/components/default_discover_table/helper.tsx
@@ -94,9 +94,13 @@ export function getLegacyDisplayedColumns(
       colRightIdx: idx + 1 >= columns.length ? -1 : idx + 1,
     };
   });
-  return !hideTimeField &&
-    indexPattern.timeFieldName &&
-    !columns.includes(indexPattern.timeFieldName)
-    ? [getTimeColumn(indexPattern.timeFieldName), ...columnProps]
+
+  const shouldIncludeTimeField =
+    !hideTimeField &&
+    typeof indexPattern.timeFieldName === 'string' &&
+    !columns.includes(indexPattern.timeFieldName);
+
+  return shouldIncludeTimeField
+    ? [getTimeColumn(indexPattern.timeFieldName as string), ...columnProps]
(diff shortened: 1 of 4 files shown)


src/plugins/discover/public/application/components/doc_viewer/doc_viewer_render_tab.tsx
```diff
@@ -29,7 +29,7 @@
  */
 
 import React, { useRef, useEffect } from 'react';
-import { DocViewRenderFn, DocViewRenderProps } from '../../../doc_views/doc_views_types';
+import { DocViewRenderFn, DocViewRenderProps } from '../../doc_views/doc_views_types';
```
virajsanghvi, line 32: How did this build if this wasn't imported correctly?

src/plugins/discover/public/application/components/default_discover_table/helper.tsx
```diff
@@ -94,7 +94,9 @@ export function getLegacyDisplayedColumns(
       colRightIdx: idx + 1 >= columns.length ? -1 : idx + 1,
     };
   });
-  return !hideTimeField && indexPattern.timeFieldName
+  return !hideTimeField &&
+    indexPattern.timeFieldName &&
+    !columns.includes(indexPattern.timeFieldName)
```
virajsanghvi, line 99: nit: consider assigning this to a meaningfully named variable so the reader doesn't have to parse out what this is trying to do

src/plugins/discover/public/application/components/doc_viewer/doc_viewer_render_tab.tsx
ananzh: I do see a type error when I debug the issue
```
Cannot find module '../../../doc_views/doc_views_types' or its corresponding type declarations.ts(2307)
```

I believe this is related to an old discussion https://github.com/opensearch-project/OpenSearch-Dashboards/issues/1660, which highlights that the standard build scripts in the project do not fail on or show TypeScript errors. This has led to a situation where TypeScript compilation errors (like the one I encountered) do not cause the build to fail, allowing many such errors to creep into the project unnoticed.

src/plugins/discover/public/application/components/default_discover_table/helper.tsx
ananzh: Will think a good name and update

src/plugins/discover/public/application/components/doc_viewer/doc_viewer_render_tab.tsx
ananzh: I believe we don't have the bandwidth to address this issue right now, but we should consider implementing a way to prevent new type errors from being added. This will help ensure the problem doesn't grow larger. 

CC: @AMoo-Miki @ashwin-pc

ashwin-pc: Yeah typescript error right now rely on the IDE or us manually running tsc and validating the diff. I have a work in progress PR to help catch this in the CI but its not complete yet. https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6586

virajsanghvi (approved):

abbyhu2000 (approved):

zhyuanqi (approved):

PR #7022, opened 2024-06-13
Fix add filter for numeric scripted field

### LINKED ISSUE #6996: [BUG] is filter not working with numeric Scripted fields
**Describe the bug**
Adding filters for scripted numeric fields using `is` operator after clicking the **Add Filter** Button does not produce any result. However directly clicking on a visualization containing the required field adds the filter correctly. This behavior is observed across visualizations. (Tested the behavior in VisBuilder & Vertical bar Visualization)

**Payload sent while using the Add Filter button**
[PayloadAddFilter.json](https://github.com/user-attachments/files/15779561/PayloadAddFilter.json)
**Payload being sent while clicking on the visualization directly**
[PayloadVisualization.json](https://github.com/user-attachments/files/15779562/PayloadVisualization.json)

Upon analysis of Payloads, the following observation was found.
When using the Add Filter button, The value is being sent as string.
`"value": "66"`
While when we are using the visualization to filter it sends the 
`"value": 66`

This explains why we are not getting the expected results with the Add Filter option in case of numeric fields  since we are sending the value as string. 

**Expected behavior**
Add Filter should work with scripted numeric fields while using the is operator.

**OpenSearch Version**
3.0.0-snapshot 
**Dashboards Version**
3.0.0

**Plugins**

**Screenshots**

Using the Add filter 
image

Clicking on the Visualization 
image

**Additional context**

abbyhu2000:
### Description

Fix add filter for numeric scripted field. Previously, adding filters for scripted numeric fields using is operator after clicking the Add Filter Button does not produce any result.

### Issues Resolved
resolves #6996 

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/43937633/8a5e60ad-9639-4ea7-81c4-933d4514d57b

## Changelog

- fix: Fix add filter for numeric scripted field


src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.tsx
```diff
@@ -147,6 +147,17 @@ class ValueInputTypeUI extends Component<Props> {
...
   };
 
+  private onNumberChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const params = event.target.value;
+    this.props.onChange(
+      typeof params === 'string'
+        ? parseFloat(params)
+        : typeof params === 'bigint'
+        ? (params as BigInt).toString()
+        : params
```
virajsanghvi, line 157: nit: instead of using a nested ternary extract part of it out, use an if/else statement, or use parens in the nested ternary to make it easier to read.

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.tsx
```diff
@@ -147,6 +147,17 @@ class ValueInputTypeUI extends Component<Props> {
     this.props.onChange(boolValue);
   };
 
+  private onNumberChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
virajsanghvi, line 150: Are these cases already covered in unit tests?

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.tsx
abbyhu2000: Changed to if else statement

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.tsx
abbyhu2000: This file does not have any unit tests, added unit test for this file.

ananzh (dismissed):

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.tsx
virajsanghvi: Thanks for adding the unit tests! Do they cover the change handler you added?

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { mountWithIntl } from 'test_utils/enzyme_helpers';
+import { ValueInputType } from './value_input_type';
+
+let onChangeMock: any;
+
+describe('Value input type', () => {
+  beforeAll(() => {
+    onChangeMock = jest.fn();
+  });
+  it('is number', async () => {
```
AMoo-Miki, line 16: Do we have tests elsewhere for IPs, Dates, and anything else this input handles? If not, this would be good opportunity to add them.

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const valueInputProps = {
+      value: 1,
```
AMoo-Miki, line 18: For BigInt, use exactly the same test and only change value to `1n`.

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.tsx
```diff
@@ -147,6 +147,17 @@ class ValueInputTypeUI extends Component<Props> {
...
+    const params = event.target.value;
+    let numValue;
+    if (typeof params === 'string') {
+      numValue = parseFloat(params);
+    } else if (typeof params === 'bigint') {
+      numValue = (params as BigInt).toString();
```
AMoo-Miki, line 156: We shouldn't have this. If the value is a BigInt, we should leave it as it.
```suggestion

```

abbyhu2000: I added this logics according to previous logics on line 71, https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7022/files/6b7d63232ed9b7b0c8a152ad66c262d41d8b4629#diff-d1ea4f28ecf59eb6ec2702dd4f29cd9a42aeab29c6065fc80c73be1f3ad9825aR71

I see they called `toString()` if the value is Bigint

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.test.tsx
abbyhu2000: Will add date and ip unit test

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.test.tsx
abbyhu2000: added a unit test for bigint using 1n

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.tsx
virajsanghvi: We don't test the methods directly, but we still need to exercise the logic - otherwise, how are we validating it works? I'm mostly reacting to the lack of code coverage below.

abbyhu2000: I agree, but how do we test those private method logics? I thought if the function is not exported by the module, it cannot be called by test code outside the module. Could you provide an example to help me? Thanks!

ananzh (approved):

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+  it('is number bigint', async () => {
+    const valueInputProps = {
+      value: '1n',
+      type: 'number',
+      onChange: onChangeMock,
+      onBlur: () => {},
+      placeholder: '',
+    };
+    const component = mountWithIntl(<ValueInputType {...valueInputProps} />);
+    expect(component).toMatchSnapshot();
```
ashwin-pc, line 37: Can we not use snapshot tests here? Taking a look at the snapshot and they dont offer much value here

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.tsx
ashwin-pc, line 156: Where are we using `numValue`?

abbyhu2000: should be `this.props.onChange(numValue);`. It is a typo when addressing later comments

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.test.tsx
abbyhu2000: I think they do provide values, the unit tests are mainly testing if `switch (this.props.type)` are rendering correct components based on the input type. For example, string type should render <EuiFieldText> with correct placeholder, boolean type should render <EuiSelect> with correct selections.

```
switch (this.props.type) {
      case 'string':
        inputElement = (
          <EuiFieldText
           ...
          />
        );
        break;
      case 'number':
        inputElement = (
          <EuiFieldNumber
            ...
          />
        );
        break;
      case 'date':
        inputElement = (
          <EuiFieldText
            ...
          />
        );
        break;
      case 'ip':
        inputElement = (
          <EuiFieldText
           ...
          />
        );
        break;
      case 'boolean':
        inputElement = (
          <EuiSelect
            fullWidth={this.props.fullWidth}
            options={[...]}
            ...
          />
        );
```

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.tsx
abbyhu2000: will fix soon

src/plugins/data/public/ui/filter_bar/filter_editor/value_input_type.test.tsx
ashwin-pc: i meant can we just snapshot the value we expect instead of the whole thing. If this test breaks, it should be only because of the changes related to what you are testing. Right now, if the component chnages, these tests will break too. Keep unit tests scoped to just test for just the feature and not anything else. Otherwise when things change, a lot of these test will break.

ashwin-pc (approved):
Thanks for making the change. This looks good to me

ananzh (approved):

Issue #7021, opened 2024-06-13
[Multiple Datasource] [Plugins Version Decoupling] Fetch and include Data Source Engine Type in the data-source saved object when creating a new data source in MDS


ZilongX:
**Is your feature request related to a problem? Please describe.**

This feature is required per discussion and feedback collected from RFC https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5877.

In order to decouple the plugin versions with OS cluster versions, we will need to specify a supported data source version range for example `>= 2.1.0`, yet given we support both OpenSearch and ElasticSearch clusters, the version could be overlapped since we only record the numeric part, so for an ES 7.10.2 data source the version would be recorded as `7.10.2`, which would break the support range check as it would also satisfy the semver check of `>=2.1.0`

**Describe the solution you'd like**

Add a new layer of `dataSourceType`, to record the backend engine type like `OpenSearch` or `ElasticSearch`

**Describe alternatives you've considered**

Basically this approach would be the most straightforward and simplified one to resolve the issue.

**Additional context**

[Plugins Version Decoupling] Dev work is in progress and targeting release `2.16.0`

PR #7026, opened 2024-06-13
[Multiple Datasource][Version Decoupling] Add data source engine type to data source saved object

ZilongX:
### Description

* This one is to enrich the `data source` saved object by adding a new meta data `dataSourceEngineType`
* This feature is required per discussion and feedback collected from RFC https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5877.
* In order to decouple the plugin versions with OS cluster versions, we will need to specify a supported data source version range for example `>= 2.1.0`, yet given we support both OpenSearch and ElasticSearch clusters, the version range could be overlapped since only the numeric part is recorded for now, so for an ES 7.10.2 data source the version would be recorded as `7.10.2`, which would break the support range check as it would also satisfy the semver check of `>=2.1.0`. This extra piece of engine type information would get this issue resolved.

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/7021

## Screenshot

* Newly created OS datasource with `dataSourceEngineType` as `OpenSearch`
image

* Newly created ES datasource with `dataSourceEngineType` as `Elasticsearch`
image

* Newly created OSS datasource with `dataSourceEngineType` as `OpenSearch Serverless`
image

## Testing the changes

All related test cases passed as

image

## Changelog
- fix: [MDS] Add data source engine type to data source saved object

### Check List

- [x] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_source/common/data_sources/types.ts
```diff
@@ -52,3 +53,10 @@ export enum SigV4ServiceName {
 }
 
 export { DataSourceError } from './error';
+
+export enum DataSourceEngineType {
+  OpenSearch = 'OpenSearch',
```
kavilla, line 58: How come not `default`? 

If the data source then is undefined then it's assumed `default`. One thing I noticed related to the flint stuff passing around types with spaces causes some issues if the global url states.

vs using `default` which is the implied OpenSearch and is BWC if we assume default if no type explicitly defined.

I know in the repo it's a mixed casing with enums but not sure if using default will cause some issues lowercase so maybe

```
export enum DATA_SOURCE_ENGINE_TYPE {
  DEFAULT = 'default',
  SERVERLESS = 'serverless',
  LEGACY = 'legacy',
  NONE = 'none'
}
```

the serverless banking off that we assume it's OpenSearch. The legacy because internally we reference it as legacy internally [[example](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/opensearch/version_check/opensearch_opensearch_dashboards_version_compatability.ts#L115)]

and NONE because it's matches the pattern to other enums in the repo [[example](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a4aa682226b9d22d2c6e7a41a11a44225b6d7db1/src/plugins/vis_type_timeseries/public/application/visualizations/constants/chart.js#L49)]

src/plugins/data_source/common/data_sources/types.ts
```diff
@@ -10,6 +10,7 @@ export interface DataSourceAttributes extends SavedObjectAttributes {
   description?: string;
   endpoint: string;
   dataSourceVersion?: string;
+  dataSourceEngineType?: DataSourceEngineType;
```
kavilla, line 13: do we need a migration? i dont think so right 

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a4aa682226b9d22d2c6e7a41a11a44225b6d7db1/src/plugins/data_source/server/saved_objects/data_source.ts#L44

it's optional field. so think it doesn't need it.

src/plugins/data_source/common/data_sources/types.ts
```diff
@@ -10,6 +10,7 @@ export interface DataSourceAttributes extends SavedObjectAttributes {
   description?: string;
   endpoint: string;
   dataSourceVersion?: string;
```
kavilla, line 12: nit: prolly keeping in line with version which makes sense too.

but might be redundant to put `dataSource` in the field name since it's apart of the DataSource attributes. but i also understand the concept of data sources has different meanings even within this repo so might make sense to specify.

kavilla:
just an opinion on the enums: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7026#discussion_r1639496792

but this is great thank you!

src/plugins/data_source/common/data_sources/types.ts
ZilongX: Why we need a `default` ?  

This engine type is retrieved when the data source object is pulled together and assuming a default just makes no sense to the users where the default just mean OS cluster ? This info is not leveraged in global url states but do share the flint issues link if any so we can double check. (is flint even open sourced ?)

Per the `legacy` it's pretty vague where it could be meaning either `Elasticsearch` or `Legacy OpenSearch` (say when we have OpenSearch 3.0 out and OpenSearch 1.x retired which become Legacy OpenSearch). For supported data source types there are three now explicitly so given three exact type rather than giving general terms looks more accurate.

src/plugins/data_source/common/data_sources/types.ts
ZilongX: that's why this field is marked optional yes

src/plugins/data_source/common/data_sources/types.ts
ZilongX: yet you got the point, the `dataSource` prefix was added intentionally just to distinguish with other data source terms

ZilongX:
> just an opinion on the enums: [#7026 (comment)](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7026#discussion_r1639496792)
> 
> but this is great thank you!

Thanks @kavilla, all great feedbacks ~~~

yujin-emma (approved):
Great, hope it can resolve the version filter well

ZilongX:
> Great, hope it can resolve the version filter well

Thanks @yujin-emma , yeah first baby step ...

src/plugins/data_source/server/routes/data_source_connection_validator.ts
```diff
@@ -36,26 +38,42 @@ export class DataSourceConnectionValidator {
...
         .info()
         .then((response) => response.body)
         .then((body) => {
-          dataSourceVersion = body.version.number;
+          dataSourceInfo.dataSourceVersion = body.version.number;
+
+          if (
+            body.version.distribution !== null &&
+            body.version.distribution !== undefined &&
+            body.version.distribution === 'opensearch'
```
xinruiba, line 65: nit: 
`body.version.distribution === 'opensearch'` should be enough here. 
But no hurts to add `!== null` and `!== undefined` explicitly.

xinruiba (approved):

src/plugins/data_source/server/routes/fetch_data_source_metadata.test.ts
```diff
@@ -366,6 +373,7 @@ describe(`Fetch DataSource MetaData ${URL}`, () => {
       .expect(200);
     expect(result.body).toEqual({
       dataSourceVersion: '2.11.0',
+      dataSourceEngineType: 'OpenSearch',
       installedPlugins: ['opensearch-ml', 'opensearch-sql'],
```
bandinib-amzn, line 377: Is this is the only place where we are comparing result.body during testing? What if we are testing result.body at other places like functional, integration? It will break, right? We will need to include dataSourceEngineType everyehwhere.

bandinib-amzn (approved):

src/plugins/data_source/server/routes/data_source_connection_validator.ts
zhongnansu, line 65:
```suggestion
          if (
            body.version.distribution === 'opensearch'
```
This can be simplified

src/plugins/data_source/server/routes/fetch_data_source_metadata.test.ts
ZilongX: yes as `dataSourceEngineType` is marked as optional to ensure existing test won't fail

src/plugins/data_source/common/data_sources/types.ts
kavilla: > (is flint even open sourced ?)

but i dont think elasticsearch is either which is why internally codewise we refer to elasticsearch references as `legacy`

ZilongX: this info is not displayed to user, would like to follow up on this legal verification mentioned here, do you have any previous ref or poc from legal ? have no problem to switch all reference of elasticsearch to legacy

Issue #6320, opened 2024-04-03
[BUG][NewHomePage] Avoid Requiring Index Write Permission for the New Homepage Loading


Flyingliuhub:
**Describe the bug**

The new homepage was introduced in the 2.13 release, which required index write permission when loading. However, this should be avoided as dashboard users may not have index write permission. 

**To Reproduce**
1. Enable useNewHomePage in the advance setting 
2. Open the homepage with users who only have readonly permission users
3. The new homepage was broken with error message "There was an error loading the homepage."

Steps to reproduce the behavior in local cluster:
1. Run OS with security plugin `yarn opensearch snapshot --security`
2. Run osd `yarn start --no-base-path`
3. Login with osd with admin user and enable newhome feature from advanced setting for `Global tenant`
4. Copy the default role `readall` and add Global tenant read permission   
5. Create user `test` and map to the copied `readall_copy` role
6. Login with osd with `test` user which created in the step 5
7. Navigate to homepage

**Expected behavior**
The new home page loading should not required to index write permission

**Impact**
In case of admin enabled advance setting in the global tenant, all the readonly users under global tenant will impact when visit new home page. 

**OpenSearch Version**
8.13.0

**Dashboards Version**
2.13.0

**Plugins**

All the official release plugins

**Screenshots**

<img width="995" alt="image" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/33105471/5bd5bcd5-73f0-49a6-a5c0-703191609e68">

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/33105471/072027e1-7cd8-4b8e-8d46-67fbd5148a8c

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

````
{
    "statusCode": 403,
    "error": "Forbidden",
    "message": "no permissions for [indices:data/write/index] and User [name=xxx, backend_roles=[xxx], requestedTenant=null]: security_exception: [security_exception] Reason: no permissions for [indices:data/write/index] and User [name=xxxx, backend_roles=[xxxx], requestedTenant=null]"
}

````

seraphjiang:
Thanks @Flyingliuhub @ashwin-pc  

@bbarani  @wbeckler  do you think we should conduct a patch release to include fix for this?

cc: @kgcreative @elfisher

bbarani:
[...]

PR #7054, opened 2024-06-18
[BUG][NewHomePage] Temp Solution to avoid crash for anonymous user with no write permission

ananzh:
### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6320

## Changelog
- fix: [BUG][NewHomePage] Temp Solution to avoid crash for anonymous user with no write permission

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/home/public/services/section_type/section_type.ts
+++ b/src/plugins/home/public/services/section_type/section_type.ts
@@ -127,10 +127,7 @@ export class SectionTypeService {
             }
 
             if (heroes || sections) {
-              homepage
-                .save({})
-                .then(() => error$.next(undefined))
-                .catch((e) => error$.next(e));
+              homepage.save({});
             }
           })
         );


src/plugins/home/public/services/section_type/section_type.ts
```diff
@@ -127,10 +127,7 @@ export class SectionTypeService {
             }
 
             if (heroes || sections) {
-              homepage
-                .save({})
-                .then(() => error$.next(undefined))
-                .catch((e) => error$.next(e));
+              homepage.save({});
```
virajsanghvi, line 130: is there anything that tracks fixing this permanently?

ananzh: I could open an issue. Currently we are using this tmp one for unblocking purpose.

AMoo-Miki (approved):

src/plugins/home/public/services/section_type/section_type.ts
ananzh: actually we could keep the issue open and revisit to see if there is better solution

zhyuanqi: Hi Anan. Will this fix have some potential issue as it swallowed the error. Will it show information that not suppose to show to user?

ruanyl: Maybe we could leave a comment here to describe why it's crashed(I'm curious ;)

manasvinibs: +1

manasvinibs (approved):

PR #7057, opened 2024-06-19
Condence context menus

AMoo-Miki:
Hold for @kgcreative and Viraj's signoff.

condenced

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/filter_bar/filter_options.tsx
+++ b/src/plugins/data/public/ui/filter_bar/filter_options.tsx
@@ -179,13 +179,13 @@ class FilterOptionsUI extends Component<Props, State> {
         panelPaddingSize="none"
         repositionOnScroll
       >
-        <EuiPopoverTitle>
+        <EuiPopoverTitle paddingSize="s">
           <FormattedMessage
             id="data.filter.searchBar.changeAllFiltersTitle"
             defaultMessage="Change all filters"
           />
         </EuiPopoverTitle>
-        <EuiContextMenu initialPanelId={0} panels={[panelTree]} />
+        <EuiContextMenu initialPanelId={0} panels={[panelTree]} size="s" />
       </EuiPopover>
     );
   }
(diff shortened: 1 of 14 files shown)


kavilla:
is the proposal within OSD or are these examples for potentially having the default values in OUI be smaller? so it hits the entire app?

src/plugins/data/public/ui/filter_bar/filter_options.tsx
```diff
@@ -179,13 +179,13 @@ class FilterOptionsUI extends Component<Props, State> {
         panelPaddingSize="none"
         repositionOnScroll
       >
-        <EuiPopoverTitle>
+        <EuiPopoverTitle paddingSize="s">
```
virajsanghvi, line 182: Do you think its better to do these locally rather than set a default through OUI?

virajsanghvi (approved):
We probably need to follow up with 1) strategy on which changes will need to be behind a flag, 2) whether popover defaults should be made at OUI level and 3) how to address small context menu titles being a smaller size than popover titles even though body text size is the same, but changes lgtm and these could be addressed after.

src/plugins/data/public/ui/filter_bar/filter_options.tsx
AMoo-Miki: Due to the potential blast radius, I think we should localize these changes. I am also concerned about different fonts and font-sizes across themes and their impact on all of these.

AMoo-Miki:
> is the proposal within OSD or are these examples for potentially having the default values in OUI be smaller? so it hits the entire app?

There is a big collection of changes. Some are done globally within OUI (via a new theme) and some are done surgically. There will most likely be some that will go behind a config flag too.

zhongnansu (approved):

Hailong-am:
@AMoo-Miki seem test snapshot update was missed, that cause CI to fail https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/9966986613/job/27539949592?pr=7254

PR #7034, opened 2024-06-14
[Discover-next] Add query editor extensions

joshuali925:
### Description

see #6894 

This PR picks #6894, #6895, #6933, #6972 to main. Additionally,
- separates extensions from query enhancements
- adds banner support
- partially revert #6972 as it's pending on the data source commit to main
- renames search bar extension to query editor extension

A query editor extension can display a UI component above the query editor and/or a banner above the language selector. The component has the ability to read and write discover search bar states to enhance the search experience for users. The configuration is part of UI Enhancements.

```ts
export interface QueryEditorExtensionDependencies {
  /**
   * Currently selected index patterns.
   */
  indexPatterns?: Array<IIndexPattern | string>;
  /**
   * Currently selected data source.
   */
  dataSource?: DataSource;
  /**
   * Currently selected query language.
   */
  language: string;
}

export interface QueryEditorExtensionConfig {
  /**
   * The id for the search bar extension.
   */
  id: string;
  /**
   * Lower order indicates higher position on UI.
   */
  order: number;
  /**
   * A function that determines if the search bar extension is enabled and should be rendered on UI.
   * @returns whether the extension is enabled.
   */
  isEnabled: (dependencies: QueryEditorExtensionDependencies) => Promise<boolean>;
  /**
   * A function that returns the search bar extension component. The component
   * will be displayed on top of the query editor in the search bar.
   * @param dependencies - The dependencies required for the extension.
   * @returns The component the search bar extension.
   */
  getComponent?: (dependencies: QueryEditorExtensionDependencies) => React.ReactElement | null;
  /**
   * A function that returns the search bar extension banner. The banner is a
   * component that will be displayed on top of the search bar.
   * @param dependencies - The dependencies required for the extension.
   * @returns The component the search bar extension.
   */
  getBanner?: (dependencies: QueryEditorExtensionDependencies) => React.ReactElement | null;
}

export interface UiEnhancements {
  query?: QueryEnhancement;
+ queryEditorExtension?: QueryEditorExtensionConfig;
}
```

### Issues Resolved

Developers can utilize search bar extensions to add additional features to the search bar, such as query assist.

Part of #6077

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

Issue #6957, opened 2024-06-06
[Discover-Next] Meta Issue for Bugs, Updates, Fast-Follows


kavilla:
#### Issues list

- [ ] #7032
- [x] #7036
- [ ] #7037
- [x] #7038
- [ ] #7039
- [x] #7040
- [ ] #7041
- [x] #7042
- [x] #7046
- [ ] Add more tests
- [ ] `stats` display in table
   - [ ] is this log patterns?

#### Query Assist
- [ ] Clean up reference of `queryEditorHeaderRef` [[COMMENT](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7034/files#r1641529201)][[COMMENT](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7034/files#r1641567709)]
- [ ]  Pass dataSource to TopNavMenu once dataSource work is completed [[COMMENT](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7034/files#r1641532625)]

#### Async queries (not in review into main yet)
- [ ] Loading state, ensuring the loading is giving a proper indication for long standing queries

#### Pending items post second bug bash

- [ ] #7043
- [ ] Async queries, when navigating the request, the request is still polling
  - [ ]  Should we cancel or should we let it go and not fire again on landing 
- [ ] Better error handling
- [ ] Extend OSD field type mapper function registration points
- [ ] Remove temp logic for https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7106 if not needed
- [ ] Highlight fields if PPL where filter
- [ ] Time field width gets rendered before data populates causing time column to get cropped and displayed poorly
   - [ ] Validate this is still happening after all bug fixs
   - [ ] If it is still happening ensure we don't render until we got the data
- [ ] UI shifts when hovering over quick filter buttons
   - [ ] Validate this is still happening after all bug fixs
   - [ ] This could be fixed with the timefield render issue
   - [ ] If not this could be setting a min width in the sidebar to prevent pushing from canvas
- [ ] when toggling on and then off, we force the query to be `''` and language to `kuery` for DQL
   - [SOURCE](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/ui/settings/settings.ts#L52)
   - This one done to make sure users get back to a default state if the previous query was not natively supported like PPL or SQL.
[...]

PR #7077, opened 2024-06-20
Fix object empty check and minor perf issue in query editor extensions

joshuali925:
### Description

Follow up for #7034, this PR
- addresses https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7034#discussion_r1646749538
- addresses https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7034#discussion_r1646750783
- fixes render order
   - QueryEditorExtensions requires the container divs to be mounted first,
but in the previous implementation, extensions will be mounted first and
it relied on rerendering of queryEditorTopRow. Moving them into query
editor fixes the render order and ensures refs are set.

@AMoo-Miki I didn't use the object check `'[object Object]' !== Object.prototype.toString.call(configMap)`. I don't know what access user has, but if an attacker can arbitrarily alter `configMap`, the code will still break with something like
```tsx
        configMap={
          new Proxy(
            {},
            {
              ownKeys(target) {
                throw new Error('Accessing keys is not allowed.');
              },
            }
          )
        }
```

Given that our code creates the config map here, I think relying on static type check is enough, but feel free to comment if otherwise.
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/7f0e39eb9809c95b98069cc971611edc2cbbc62b/src/plugins/data/public/ui/ui_service.ts#L31

No test changes because this PR does not change code behavior.

## Changelog
- fix: Fix object empty check and minor perf issue in query editor extensions

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/query_editor/query_editor.tsx
+++ b/src/plugins/data/public/ui/query_editor/query_editor.tsx
@@ -58,6 +58,7 @@ interface State {
   index: number | null;
   suggestions: QuerySuggestion[];
   indexPatterns: IIndexPattern[];
+  queryEditorRect: DOMRect | undefined;
 }
 
 const KEY_CODES = {
@@ -82,6 +83,7 @@ export default class QueryEditorUI extends Component<Props, State> {
     index: null,
     suggestions: [],
     indexPatterns: [],
+    queryEditorRect: undefined,
   };
 
   public inputRef: HTMLTextAreaElement | null = null;
@@ -89,6 +91,8 @@ export default class QueryEditorUI extends Component<Props, State> {
   private persistedLog: PersistedLog | undefined;
   private abortController?: AbortController;
   private services = this.props.opensearchDashboards.services;
+  private componentIsUnmounting = false;
+  private queryEditorDivRefInstance: RefObject<HTMLDivElement> = createRef();
   private headerRef: RefObject<HTMLDivElement> = createRef();
   private bannerRef: RefObject<HTMLDivElement> = createRef();
   private extensionMap = this.props.settings?.getQueryEditorExtensionMap();
@@ -234,6 +238,12 @@ export default class QueryEditorUI extends Component<Props, State> {
 
     this.initPersistedLog();
     // this.fetchIndexPatterns().then(this.updateSuggestions);
+    this.handleListUpdate();
+
+    window.addEventListener('scroll', this.handleListUpdate, {
+      passive: true, // for better performance as we won't call preventDefault
+      capture: true, // scroll events don't bubble, they must be captured instead
+    });
   }
 
   public componentDidUpdate(prevProps: Props) {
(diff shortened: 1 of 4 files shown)


src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -92,6 +93,9 @@ export default class QueryEditorUI extends Component<Props, State> {
   private services = this.props.opensearchDashboards.services;
   private componentIsUnmounting = false;
   private queryEditorDivRefInstance: RefObject<HTMLDivElement> = createRef();
```
kavilla, line 95: tbh I think I carried this over from the query string input component and it was used for attaching suggestions. I wonder then if we should just delete this then since i think the header ref more or less solves it.

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -120,6 +124,30 @@ export default class QueryEditorUI extends Component<Props, State> {
     });
   };
 
+  private renderQueryEditorExtensions() {
```
kavilla, line 127: thanks! nice clean up!

src/plugins/data/public/ui/query_editor/query_editor_top_row.tsx
```diff
@@ -246,33 +243,12 @@ export default function QueryEditorTopRow(props: QueryEditorTopRowProps) {
...
           dataTestSubj={props.dataTestSubj}
-          queryEditorHeaderRef={queryEditorHeaderRef}
-          queryEditorBannerRef={queryEditorBannerRef}
+          queryLanguage={queryLanguage}
         />
       </EuiFlexItem>
     );
   }
 
-  function renderQueryEditorExtensions() {
```
kavilla, line 256: this is a good clean up. that i need to do after restructuring. I have a clean up task in https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6957 but some of the render functions i added are in the wrong location and that's cause i tried to not touch query string input. so thanks for starting the clean up!

kavilla (approved):

kavilla:
nit: can you update the changelog section to your PR? https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7077#issuecomment-2183359503

you can add `- skip` as well

src/plugins/data/public/ui/query_editor/query_editor.tsx
joshuali925: removed the handleListUpdate related stuff in 5b3e3589cc3

Swiddis (approved):

kavilla (dismissed):

kavilla (approved):

abbyhu2000 (approved):

kavilla (approved):

abbyhu2000 (approved):

Issue #7099, opened 2024-06-24
[FEATURE] Support Version Decoupling in Index Patterns Dashboards Plugin


ZilongX:
**Is your feature request related to a problem? Please describe.**

OpenSearch plugins currently tightly couple their versions with the underlying OpenSearch/OpenSearch Dashboards version. This tight coupling can lead to challenges in maintaining and upgrading plugins independently. To enable supporting of the decoupling of the plugin versions from the OpenSearch versions will allow for more flexibility in managing and updating plugins.

More details is available at RFC Plugins Version Decoupling - https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5877

**Describe the solution you'd like**

Support version decoupling in the index patterns dashboards plugin, where end users would be able to consume the latest version of index patterns dashboards plugin with a range of supported OS versions instead of just one version

**Describe alternatives you've considered**

More as detailed in [Plugins Version Decoupling](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5877)

**Additional context**

More as detailed in [Plugins Version Decoupling](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5877)

ZilongX:
When the default data source is not compatible we'll need further action to handle the case. Only for legacy data source though since Index Patterns actually support all OS versions ranging from 1.0.0

ZilongX:
We also need the capability to filter out incompatible data sources based on whether the required backend plugins are installed or not.

ZilongX:
Need to improve the code coverage for Index Pattern Plugin

PR #7100, opened 2024-06-24
[Multiple Datasource][Version Decoupling] Support Version Decoupling in Index Patterns Dashboards Plugin

ZilongX:
### Description

* This PR is to add support of `Version Decoupling` in `Index Patterns Dashboards Plugin`
* This is achieved mainly by add a new entry `supportedOSDataSourceVersions` in the plugin manifest file
*  For `Index Patterns Plugin` it would support all versions of OS clusters so the supported range would be `>=1.0.0`
* The current behavior is to filter out the incompatible data sources out of the data source list.

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/7099

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/99905560/883852c5-51a2-4b50-9226-f84d873721b0

## Changelog

- fix: [MDS][Version Decoupling] Add support of Version Decoupling in Index Patterns Dashboards Plugin

### Check List

- [x] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/index_pattern_management/public/components/create_index_pattern_wizard/components/step_data_source/components/header/header.tsx
+++ b/src/plugins/index_pattern_management/public/components/create_index_pattern_wizard/components/step_data_source/components/header/header.tsx
@@ -23,10 +23,12 @@ import {
   DataSourceRef,
   IndexPatternManagmentContext,
 } from 'src/plugins/index_pattern_management/public/types';
+import semver from 'semver';
 import { useOpenSearchDashboards } from '../../../../../../../../../plugins/opensearch_dashboards_react/public';
 import { getDataSources } from '../../../../../../components/utils';
 import { DataSourceTableItem, StepInfo } from '../../../../types';
 import { LoadingState } from '../../../loading_state';
+import * as pluginManifest from '../../../../../../../opensearch_dashboards.json';
 
 interface HeaderProps {
   onDataSourceSelected: (id: string, type: string, title: string) => void;
@@ -68,6 +70,12 @@ export const Header: React.FC<HeaderProps> = (props: HeaderProps) => {
       .then((fetchedDataSources: DataSourceTableItem[]) => {
         setIsLoading(false);
         if (fetchedDataSources?.length) {
+          fetchedDataSources = fetchedDataSources.filter((dataSource) =>
+            semver.satisfies(
+              dataSource.datasourceversion,
+              pluginManifest.supportedOSDataSourceVersions
+            )
+          );
           setDataSources(fetchedDataSources);
         }
       })
--- a/src/plugins/index_pattern_management/public/components/create_index_pattern_wizard/types.ts
+++ b/src/plugins/index_pattern_management/public/components/create_index_pattern_wizard/types.ts
@@ -92,4 +92,5 @@ export interface DataSourceTableItem {
   sort: string;
   checked?: 'on' | 'off';
   label: string;
+  datasourceversion: string;
 }
--- a/src/plugins/index_pattern_management/public/components/utils.ts
+++ b/src/plugins/index_pattern_management/public/components/utils.ts
@@ -90,7 +90,7 @@ export async function getDataSources(savedObjectsClient: SavedObjectsClientContr
     savedObjectsClient
       .find<DataSourceAttributes>({
         type: 'data-source',
-        fields: ['title', 'type'],
+        fields: ['title', 'type', 'dataSourceVersion'],
         perPage: 10000,
       })
       .then((response) =>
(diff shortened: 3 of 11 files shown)


xinruiba (approved):

yujin-emma (approved):
LGTM

bandinib-amzn:
Can you please add more UT for uncovered lines so that codecov CI would pass?

bandinib-amzn:
> For Index Patterns Plugin it would support all versions of OS clusters so the supported range would be >=1.0.0

Do we have any test result to support above sentence? Can we attach that here as source of truth?

src/plugins/index_pattern_management/public/components/create_index_pattern_wizard/components/step_data_source/components/header/header.tsx
```diff
@@ -23,10 +23,12 @@ import {
   DataSourceRef,
   IndexPatternManagmentContext,
 } from 'src/plugins/index_pattern_management/public/types';
+import semver from 'semver';
 import { useOpenSearchDashboards } from '../../../../../../../../../plugins/opensearch_dashboards_react/public';
 import { getDataSources } from '../../../../../../components/utils';
 import { DataSourceTableItem, StepInfo } from '../../../../types';
 import { LoadingState } from '../../../loading_state';
+import * as pluginManifest from '../../../../../../../opensearch_dashboards.json';
```
bandinib-amzn, line 31: nit: can we use absolute path here instead of relative path with directory traversal?

src/plugins/index_pattern_management/public/components/create_index_pattern_wizard/types.ts
```diff
@@ -92,4 +92,5 @@ export interface DataSourceTableItem {
   sort: string;
   checked?: 'on' | 'off';
   label: string;
+  datasourceversion: string;
```
bandinib-amzn, line 95: Is it expected that each plugin needs to have this change in order to onboard version decoupling?

src/plugins/index_pattern_management/public/components/utils.ts
```diff
@@ -99,13 +99,15 @@ export async function getDataSources(savedObjectsClient: SavedObjectsClientContr
             const id = dataSource.id;
             const type = dataSource.type;
             const title = dataSource.get('title');
+            const datasourceversion = dataSource.get('dataSourceVersion');
```
bandinib-amzn, line 102: same as above.

src/plugins/index_pattern_management/public/components/create_index_pattern_wizard/components/step_data_source/components/header/header.tsx
```diff
@@ -68,6 +70,12 @@ export const Header: React.FC<HeaderProps> = (props: HeaderProps) => {
       .then((fetchedDataSources: DataSourceTableItem[]) => {
         setIsLoading(false);
         if (fetchedDataSources?.length) {
+          fetchedDataSources = fetchedDataSources.filter((dataSource) =>
+            semver.satisfies(
+              dataSource.datasourceversion,
```
zhongnansu, line 75: what about for existing/old data sources that doesn't have this field? Does it break backwards compatibility?

ZilongX:
> Can you please add more UT for uncovered lines so that codecov CI would pass?

Yes I was planned yet later found the whole section was not covered in the first place, I'll try to add coverage for the whole section or maybe in separate PR

ZilongX:
> > For Index Patterns Plugin it would support all versions of OS clusters so the supported range would be >=1.0.0
> 
> Do we have any test result to support above sentence? Can we attach that here as source of truth?

Yes, we have test results indicating latest Index Pattern would only support ES 7.9 + which for OS it would be 1.0 +

src/plugins/index_pattern_management/public/components/create_index_pattern_wizard/components/step_data_source/components/header/header.tsx
ZilongX: yes, this one just to keep the save relative style as existing imports lol

src/plugins/index_pattern_management/public/components/create_index_pattern_wizard/types.ts
ZilongX: Yes, datasource would be the minimal fields to check in order to decide the version compatibility

src/plugins/index_pattern_management/public/components/create_index_pattern_wizard/components/step_data_source/components/header/header.tsx
ZilongX: for older data sources it won't work yest as they won't support version decoupling, MDS initially just list out all data sources created regardless so there would be a bunch of unexpected behaviors, version decoupling is aiming to improve that user experiences by only allowing consumption against compatible data sources.

besides there is a `refresh` flow of data sources planning in progress which would also resolve the concern here, which will refresh the current saved object to bring it up-to-date

bandinib-amzn (approved):

PR #7101, opened 2024-06-25
[Workspace][Data Source] Support data source assignment in workspace

raintygao:
### Description

1. Add data source assignment in workspace create/update page.
2. Add addToWorkspaces and deleteFromWorkspaces method in savedObject client.
3. Update savedObject update/bulkUpdate in savedObject client and repository to support update workspace field.

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/42465692/7e2a7b89-4d01-42d7-b6a3-dcac5e4adf18

image

## Testing the changes
1. Use find API to find any data sources and what workspaces they belong to. `http://localhost:6601/w/${OSD_ID}/api/saved_objects/_find?fields=id&fields=title&fields=workspaces&per_page=10000&type=data-source&workspaces=*`

2. Enter workspace create/update page to create workspaces and assign data sources or update and assign data sources.

4. Enter workspace update page or use find API to find that data source have been assigned to specific workspace.
`http://localhost:6601/w/${OSD_ID}/api/saved_objects/_find?fields=id&fields=title&fields=workspaces&per_page=10000&type=data-source&workspaces=${WORKSPACE_ID}`

## Changelog

- feat: Support data source assignment in workspace.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/common/types.ts
+++ b/src/plugins/workspace/common/types.ts
@@ -6,6 +6,6 @@
 import { DataSourceAttributes } from 'src/plugins/data_source/common/data_sources';
 
 export type DataSource = Pick<DataSourceAttributes, 'title'> & {
-  // SavedObjectAttribute could be single or array
+  // Id defined in SavedObjectAttribute could be single or array, here only should be single string.
   id: string;
 };
--- a/src/plugins/workspace/common/types.ts
+++ b/src/plugins/workspace/common/types.ts
@@ -1,0 +1,11 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { DataSourceAttributes } from 'src/plugins/data_source/common/data_sources';
+
+export type DataSource = Pick<DataSourceAttributes, 'title'> & {
+  // SavedObjectAttribute could be single or array
+  id: string;
+};
--- a/src/plugins/workspace/public/components/workspace_form/select_data_source_panel.tsx
+++ b/src/plugins/workspace/public/components/workspace_form/select_data_source_panel.tsx
@@ -1,0 +1,148 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React, { useCallback, useEffect, useState } from 'react';
+import {
+  EuiButton,
+  EuiFormRow,
+  EuiText,
+  EuiSpacer,
+  EuiFlexGroup,
+  EuiFlexItem,
+  EuiButtonIcon,
+  EuiComboBox,
+  EuiComboBoxOptionOption,
+} from '@elastic/eui';
+import { i18n } from '@osd/i18n';
+import { SavedObjectsStart } from '../../../../../core/public';
+import { getDataSourcesList } from '../../utils';
+import { DataSource } from '../../../common/types';
+
+export interface SelectDataSourcePanelProps {
+  errors?: { [key: number]: string };
+  savedObjects: SavedObjectsStart;
+  selectedDataSources: DataSource[];
+  onChange: (value: DataSource[]) => void;
+}
+
+export const SelectDataSourcePanel = ({
+  errors,
+  onChange,
+  selectedDataSources,
+  savedObjects,
+}: SelectDataSourcePanelProps) => {
+  const [dataSourcesOptions, setDataSourcesOptions] = useState<EuiComboBoxOptionOption[]>([]);
+  useEffect(() => {
+    if (!savedObjects) return;
+    getDataSourcesList(savedObjects.client, ['*']).then((result) => {
+      const options = result.map(({ title, id }) => ({
+        label: title,
+        value: id,
+      }));
+      setDataSourcesOptions(options);
+    });
+  }, [savedObjects, setDataSourcesOptions]);
+  const handleAddNewOne = useCallback(() => {
+    onChange?.([
+      ...selectedDataSources,
+      {
+        title: '',
+        id: '',
+      },
+    ]);
+  }, [onChange, selectedDataSources]);
+
+  const handleSelect = useCallback(
+    (selectedOptions, index) => {
+      const newOption = selectedOptions[0]
+        ? // Select new data source
+          {
+            title: selectedOptions[0].label,
+            id: selectedOptions[0].value,
+          }
+        : // Click reset button
+          {
+            title: '',
+            id: '',
+          };
+      const newSelectedOptions = [...selectedDataSources];
+      newSelectedOptions.splice(index, 1, newOption);
+
+      onChange(newSelectedOptions);
+    },
+    [onChange, selectedDataSources]
+  );
+
+  const handleDelete = useCallback(
+    (index) => {
+      const newSelectedOptions = [...selectedDataSources];
+      newSelectedOptions.splice(index, 1);
+
+      onChange(newSelectedOptions);
+    },
+    [onChange, selectedDataSources]
+  );
+
+  return (
+    <div>
+      <EuiText>
+        <strong>
+          {i18n.translate('workspace.form.selectDataSource.subTitle', {
+            defaultMessage: 'Data source',
+          })}
+        </strong>
+      </EuiText>
+      <EuiSpacer size="s" />
+      {selectedDataSources.map(({ id, title }, index) => (
+        <EuiFormRow key={index} isInvalid={!!errors?.[index]} error={errors?.[index]}>
+          <EuiFlexGroup gutterSize="l">
+            <EuiFlexItem grow={false}>
+              <EuiComboBox
+                data-test-subj="workspaceForm-select-dataSource-comboBox"
+                singleSelection
+                options={dataSourcesOptions}
+                selectedOptions={
+                  id
+                    ? [
+                        {
+                          label: title,
+                          value: id,
+                        },
+                      ]
+                    : []
+                }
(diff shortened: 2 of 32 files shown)


raintygao:
According to coverage there should be more test cases to be added, currently working on it.

src/plugins/workspace/public/components/workspace_form/select_data_source_panel.tsx
```diff
@@ -0,0 +1,148 @@
...
+          }
+        : // Click reset button
+          {
+            title: '',
+            id: '',
+          };
+      const newSelectedOptions = [...selectedDataSources];
+      newSelectedOptions.splice(index, 1, newOption);
+
+      onChange(newSelectedOptions);
```
SuZhou-Joe, line 73: I am thinking why don't we make the selectedOptions an array of id, the label can be parsed from dataSourceList.

src/plugins/workspace/public/components/workspace_updater/workspace_updater.tsx
```diff
@@ -111,8 +114,17 @@ export const WorkspaceUpdater = (props: WorkspaceUpdaterProps) => {
...
+    const rawFormData = getFormDataFromWorkspace(currentWorkspace);
+
+    if (rawFormData && savedObjects && currentWorkspace) {
+      getDataSourcesList(savedObjects.client, [currentWorkspace.id]).then((selectedDataSources) => {
+        setCurrentWorkspaceFormData({
+          ...rawFormData,
+          selectedDataSources,
+        });
+      });
+    }
```
SuZhou-Joe, line 126: This is the drawback that we add label into the selectedDataSources, you need to fulfill the selected data source id by looking in the data source list. If we only use id as selectedDataSource, then we do not need the getDataSourcesList call here.

src/plugins/workspace/server/integration_tests/routes.test.ts
```diff
@@ -577,15 +577,21 @@ describe('workspace service api integration test when savedObjects.permission.en
         .post(root, `/api/workspaces`)
         .send({
           attributes: omitId(testWorkspace),
-          permissions: { invalid_type: { users: ['foo'] } },
+          settings: {
+            permissions: { invalid_type: { users: ['foo'] } },
+            dataSources: [],
```
SuZhou-Joe, line 582: Nit: Though workspace is an experimental feature, I still not recommend we do the breaking change in the API side. Maybe next time we can add a parameter.

src/plugins/workspace/server/workspace_client.ts
```diff
@@ -110,6 +115,12 @@ export class WorkspaceClient implements IWorkspaceClientImpl {
           permissions,
         }
       );
+      if (dataSources) {
```
SuZhou-Joe, line 118: Shall we assign dataSources before `create` the workspace?

src/plugins/workspace/server/workspace_client.ts
```diff
@@ -192,6 +205,27 @@ export class WorkspaceClient implements IWorkspaceClientImpl {
           throw new Error(DUPLICATE_WORKSPACE_NAME_ERROR);
         }
       }
+
+      if (newDataSources) {
```
SuZhou-Joe, line 209: Can we write unit test to cover the diff check code below? It is kind of complex.

src/plugins/workspace/server/workspace_client.ts
raintygao: I think this should be added once we figure out the logic of default DS in workspace. It would be added in a separate PR.

src/plugins/workspace/public/components/workspace_updater/workspace_updater.tsx
raintygao: Do you mean we stored selected data sources ids in workspace attributes? I maybe confused about this if we store a new field in workspace because its value may not be reliable as in savedObject.

src/plugins/workspace/server/workspace_client.ts
raintygao: Yes we can, also i'm thinking whether we should add a logic that call deleteFromWorkspaces if creation is failed, what do you think?

src/plugins/workspace/server/workspace_client.ts
raintygao: Sure, currently there is not any tests for workspace client, maybe we need to do some work to fulfill workspace client test mock.

src/plugins/workspace/server/workspace_client.ts
SuZhou-Joe: That would be better, for now I think adjusting the order should be good enough.

src/plugins/workspace/public/components/workspace_updater/workspace_updater.tsx
SuZhou-Joe: Oh, my bad. Anyway we need to fetch the list of data source. nvm.

src/plugins/workspace/server/utils.ts
```diff
@@ -89,3 +90,26 @@ export const getOSDAdminConfigFromYMLConfig = async (
...
+
+export const getDataSourcesList = (client: SavedObjectsClientContract, workspaces: string[]) => {
+  return client
+    .find({
+      type: 'data-source',
+      fields: ['id', 'title'],
+      perPage: 10000,
+      workspaces,
+    })
+    .then((response) => {
```
Hailong-am, line 102: shall we handle exception here? if user don't have permission to list all data sources

src/plugins/workspace/public/components/workspace_form/select_data_source_panel.tsx
```diff
@@ -0,0 +1,148 @@
...
+export const SelectDataSourcePanel = ({
+  errors,
+  onChange,
+  selectedDataSources,
+  savedObjects,
+}: SelectDataSourcePanelProps) => {
+  const [dataSourcesOptions, setDataSourcesOptions] = useState<EuiComboBoxOptionOption[]>([]);
+  useEffect(() => {
+    if (!savedObjects) return;
+    getDataSourcesList(savedObjects.client, ['*']).then((result) => {
```
Hailong-am, line 39: As we use `*` here, what about workspace owner go to the edit page, will it throw error and prevent the owner to update other workspace attributes like name/description?

raintygao: There will be a new PR later to disable data source assignment.

src/plugins/workspace/server/utils.ts
raintygao: In new design only OSD admin will call this function and there will be a new PR to return empty array if user doesn't have permission. So i think we don't need to handle exception here.

src/plugins/workspace/server/workspace_client.ts
```diff
@@ -102,6 +106,13 @@ export class WorkspaceClient implements IWorkspaceClientImpl {
       if (existingWorkspaceRes && existingWorkspaceRes.total > 0) {
         throw new Error(DUPLICATE_WORKSPACE_NAME_ERROR);
       }
+
+      if (dataSources) {
+        for (const ds of dataSources) {
+          await client.addToWorkspaces(DATA_SOURCE_SAVED_OBJECT_TYPE, ds.id, [id]);
```
ruanyl, line 112: Instead of calling `client.addToWorkspaces` sequentially, what about doing something like `Promise.all`?

src/plugins/workspace/common/types.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { DataSourceAttributes } from 'src/plugins/data_source/common/data_sources';
+
+export type DataSource = Pick<DataSourceAttributes, 'title'> & {
+  // SavedObjectAttribute could be single or array
```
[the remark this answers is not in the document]
raintygao: Yes,  id defined in SavedObjectAttribute could be single string or array, here only could be single string, so I declared it additionally. I will update this comment to make it looks more clear.

src/plugins/workspace/public/workspace_client.ts
```diff
@@ -185,15 +186,18 @@ export class WorkspaceClient {
    */
   public async create(
     attributes: Omit<WorkspaceAttribute, 'id'>,
-    permissions?: SavedObjectPermissions
+    settings: {
+      dataSources?: DataSource[];
```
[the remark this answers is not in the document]
raintygao: Sure, I use omit to exclude title.

src/plugins/workspace/server/workspace_client.ts
```diff
@@ -86,10 +88,12 @@ export class WorkspaceClient implements IWorkspaceClientImpl {
   }
   public async create(
     requestDetail: IRequestDetail,
-    payload: Omit<WorkspaceAttributeWithPermission, 'id'>
+    payload: Omit<WorkspaceAttributeWithPermission, 'id'> & {
```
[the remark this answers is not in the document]
raintygao: Updated.

src/plugins/workspace/server/workspace_client.ts
raintygao: Updated.

ruanyl (approved):

ruanyl (approved):

SuZhou-Joe (approved):

PR #7104, opened 2024-06-25
[Discover-Next] Async Query Support

sejli:
### Description

* Adds async query support
* Currently using regex to identify datasource, will switch to using datasource selector once that implementation is in

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/common/search/search_source/search_source.ts
```diff
@@ -412,6 +441,12 @@ export class SearchSource {
           const dataFrameResponse = response as IDataFrameResponse;
           await this.setDataFrame(dataFrameResponse.body as IDataFrame);
           return onResponse(searchRequest, convertResult(response as IDataFrameResponse));
+        } else if ((response as IDataFrameResponse).type === DATA_FRAME_TYPES.POLLING) {
```
kavilla, line 444: tbh this might not be as needed since the value gets updated it from the search source's point of view. but lets keep it for now incase there is a mechanism that will end up getting that does something new if polling compared to the default

src/plugins/data/common/search/search_source/search_source.ts
```diff
@@ -412,6 +441,12 @@ export class SearchSource {
...
+          const dataFrameResponse = response as IDataFrameResponse;
+          if (!response.body?.error) {
+            await this.setDataFrame(dataFrameResponse.body as IDataFrame);
+          }
+          return onResponse(searchRequest, convertResult(response as IDataFrameResponse));
         }
         // TODO: MQL else if data_frame_polling then poll for the data frame updating the df fields only
```
kavilla, line 451: can delete.

src/plugins/data/common/data_frames/utils.ts
```diff
@@ -472,3 +497,70 @@ export const dataFrameToSpec = (dataFrame: IDataFrame, id?: string): IndexPatter
     fields: fields.reduce(flattenFields, {} as IndexPatternFieldMap),
   };
 };
+
+export class Polling<T, P = void> {
```
kavilla, line 501: nit: in my opinion it's a little bit weird to call it Polling if it just takes takes params and has functions.

unless we extend it it from DataFrame and then we could call like dataFramePolling.fetch

kavilla:
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7104#issuecomment-2189543868 i wouldn't do a manual changelog i would install the bot

kavilla:
i like it just some comments about the class

ashwin-pc:
@sejli Whats left here?

kavilla (changes requested):
I dont think we need this anymore.

For other folks, this was going to be temp workaround to not display `SQLAsync` as a language

PR #7106, opened 2024-06-25
[Discover-Next] Adding datasources support for dataframes

sejli:
### Description

Parses user string between `::datasource::` to get the data source Then appends the dataSourceId in the meta.

Also creates the data frame before the interceptor to keep the meta info passed around.

If this gets accepted follow ups need to happen:
* should be ensuring the virtual index pattern includes the data source name with the index pattern name to avoid conflicts and a weird state.
* some reason dataframe is being always wiped out the first call in the search source. need to fix so that the schema is persisted
* weird initial load state. likely just too many things at once so we should make sure the usual and fresh experience is respected
* better indicator on the datasource, or just autocomplete that displays all the datasource connections when user types initial `::`.

Rebase of #7092 

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6957

## Changelog

- feat: onboard dataframes support to MDS and create dataframe before request

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/7106.yml
+++ b/changelogs/fragments/7106.yml
@@ -1,0 +1,2 @@
+feat:
+- Onboard dataframes support to MDS and create dataframe before request ([#7106](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7106))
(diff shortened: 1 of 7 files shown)


ashwin-pc:
@sejli This is temporary right? Can we mark it inthe code and link back to an issue or this PR to indicate that if thats the case

kavilla (approved):

abbyhu2000 (approved):

kavilla:
> @sejli This is temporary right? Can we mark it inthe code and link back to an issue or this PR to indicate that if thats the case

I believe so. One thing though it could help to trigger some suggestions based on datasources

AMoo-Miki:
Manually backported in #7139

Issue #7129, opened 2024-06-29
[BUG] Default data source that isn't local breaks Discover and dashboards


kavilla:
**Describe the bug**

Discover plugin and Dashboards plugin is based on the current context. The current context has the data source being the local cluster (or the one it started up with). When Discover loads it will retrieve the index patterns from the local cluster.

These index patterns can have references to other data sources but the index patterns that were created and have no external data source ref will still be based on the local cluster.

Then switching the default data source so now the local cluster is not the default cluster. Then any index patterns, that were created with no reference, that are selected will eventually make to the [deciding the client](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/b85e1775c4c1b46f3a04b4e63ab60943fad318a5/src/plugins/data/server/search/opensearch_search/decide_client.ts#L18) with no data source id so it will go with the default client which is **NOT** the local cluster that has the saved index pattern.

Any search calls that are sourced from the local cluster and did not expect that the default client can be swapped out should break and crash.

Any thing added to a dashboard that doesn't include the data source will break if the default client is not the local client.

Users are stuck because there doesn't seem to be a clear way to set the default data source to the local cluster:
<img width="1193" alt="Screenshot 2024-06-29 at 4 12 54 AM" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/12060704/33ef481a-e96b-4108-b890-be154c1f5bf2">

So I would have to delete all my external data sources added.

The Discover plugin data source selector bases off a default Index in Advanced Settings:
<img width="1171" alt="Screenshot 2024-06-29 at 4 14 04 AM" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/12060704/b3fe463e-e223-4d6f-9a99-752061b7b959">

This again is based on the default cluster, if the user swaps the default cluster, Discover will do it's initial call to get that value so if the default data source does not have any clue about that ID because it was autopopulated based on a different data source then Discover will crash.

We can add data source selectors every where, but the problem is shared URLs have no context about the data source ref.

If you go into Discover and click "Share", we will save it with the index pattern which will have the data source ref. But if the default cluster is not the local cluster then there is no reference to that for any **EXISTING** shared URLs for example:

```
[...]

PR #7157, opened 2024-07-02
[Discover-next] add datasource container

kavilla:
### Description

Enables the editor to let plugins to mount their own data source component to the data source container.

Related to:
https://github.com/kavilla/queryEnhancements/pull/29

### Issues Partially Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/7129

## Screenshot

Screenshot 2024-07-02 at 12 21 55 PM
Screenshot 2024-07-02 at 12 22 11 PM

## Testing the changes

`yarn start:enhancements`

## Changelog
- feat: query editor and dataframes datasources container

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/query_editor/_query_editor.scss
+++ b/src/plugins/data/public/ui/query_editor/_query_editor.scss
@@ -24,6 +24,22 @@
 }
 
 .osdQueryEditor__dataSourceWrapper {
+  .dataSourceSelect {
+    border-bottom: $euiBorderThin !important;
+
+    :first-child {
+      box-shadow: none !important;
+      height: 100%;
+      border-radius: 0;
+    }
+
+    div:is([class$="--group"]) {
+      padding: 0 !important;
+    }
+  }
+}
+
+.osdQueryEditor__dataSetWrapper {
   .dataExplorerDSSelect {
     border-bottom: $euiBorderThin !important;
 
--- a/src/plugins/data/public/ui/query_editor/query_editor.tsx
+++ b/src/plugins/data/public/ui/query_editor/query_editor.tsx
@@ -25,6 +25,7 @@ export interface QueryEditorProps {
   indexPatterns: Array<IIndexPattern | string>;
   dataSource?: DataSource;
   query: Query;
+  dataSourceContainerRef?: React.RefCallback<HTMLDivElement>;
   containerRef?: React.RefCallback<HTMLDivElement>;
   settings: Settings;
   disableAutoFocus?: boolean;
@@ -54,6 +55,7 @@ interface Props extends QueryEditorProps {
 
 interface State {
   isDataSourcesVisible: boolean;
+  isDataSetsVisible: boolean;
   isSuggestionsVisible: boolean;
   index: number | null;
   suggestions: QuerySuggestion[];
@@ -77,7 +79,8 @@ const KEY_CODES = {
 // eslint-disable-next-line import/no-default-export
 export default class QueryEditorUI extends Component<Props, State> {
   public state: State = {
-    isDataSourcesVisible: true,
+    isDataSourcesVisible: false,
+    isDataSetsVisible: true,
     isSuggestionsVisible: false,
     index: null,
     suggestions: [],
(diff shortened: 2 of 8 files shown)


ashwin-pc (approved):

abbyhu2000 (approved):

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -212,7 +215,10 @@ export default class QueryEditorUI extends Component<Props, State> {
       : undefined;
     this.onChange(newQuery, dateRange);
     this.onSubmit(newQuery, dateRange);
-    this.setState({ isDataSourcesVisible: enhancement?.searchBar?.showDataSourceSelector ?? true });
+    this.setState({ isDataSetsVisible: enhancement?.searchBar?.showDataSetsSelector ?? true });
+    this.setState({
+      isDataSourcesVisible: enhancement?.searchBar?.showDataSourcesSelector ?? true,
+    });
```
joshuali925, line 221: is it better to combine the setStates into one call? I forgot if react optimizes this

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -280,6 +296,11 @@ export default class QueryEditorUI extends Component<Props, State> {
           <EuiFlexItem grow={false}>
             <EuiFlexGroup gutterSize="xs" alignItems="center" className={`${className}__wrapper`}>
               <EuiFlexItem grow={false}>{this.props.prepend}</EuiFlexItem>
+              {this.state.isDataSourcesVisible && (
+                <EuiFlexItem grow={false} className={`${className}__dataSourceWrapper`}>
+                  <div ref={this.props.dataSourceContainerRef} />
```
joshuali925, line 301: what's the consideration to expose a ref rather than directly rendering the component here?

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -288,8 +309,8 @@ export default class QueryEditorUI extends Component<Props, State> {
                   appName={this.services.appName}
                 />
               </EuiFlexItem>
-              {this.state.isDataSourcesVisible && (
-                <EuiFlexItem grow={false} className={`${className}__dataSourceWrapper`}>
+              {this.state.isDataSetsVisible && (
```
joshuali925, line 312: should this be `isDataSetVisible`? I see some are plural some are singular, do we plan to allow multiple dataSets/dataSources on UI or just one?

src/plugins/data/public/ui/query_editor/_query_editor.scss
```diff
@@ -24,6 +24,22 @@
...
+    border-bottom: $euiBorderThin !important;
+
+    :first-child {
+      box-shadow: none !important;
+      height: 100%;
+      border-radius: 0;
+    }
+
+    div:is([class$="--group"]) {
+      padding: 0 !important;
```
joshuali925, line 37: i see `!important` is used in many places already, is it avoidable?

joshuali925 (commented):
just some questions

src/plugins/data/public/ui/query_editor/query_editor.tsx
kavilla: you're right that would be the better approach i'll fast follow

src/plugins/data/public/ui/query_editor/query_editor.tsx
kavilla: thanks for asking this one:

i'm using the data source management plugin for that component. so to avoid a circular dependency and not introduce datasourcemanagements as a required bundle. i do it in the query enhancements which can take that optional depedency

src/plugins/data/public/ui/query_editor/_query_editor.scss
kavilla: it's actually the OUI combo box. I would like to avoid it but some of the classes everywhere have like added padded so this is to avoid the height of the flex box getting pushed around

src/plugins/data/public/ui/query_editor/query_editor.tsx
kavilla: scared me. had to check. yeah i modified it for this because plural makes sense for me  like its the selector which can select multiple datasets. But tbh, I was on the fence too. I just wanted to updated in this file so its both plural.

joshuali925: got it, it's minor but i wasn't sure if it's on purpose. if it's plural should it be `areDataSetsVisible` or `dataSetsVisible`?

PR #7188, opened 2024-07-08
[Workspace] [Data Source] feat: support workspace level default data source 

raintygao:
### Description

Support set default data source in workspace

1.  Recover data source management when in workspace
2. Restrict workspace level default data source ui settings value extend from global when getting
3. Set default data source when creating work space and add check when editing workspace

## Screenshot

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/42465692/ccf06f32-edc4-4a55-a69b-1b96f98af6ce

## Testing the changes

1. Go to home and set the default data source
2. Enter a workspace and. go to data source management, the default data source would not be extended.
3. Create workspace and assign some data sources, go to data source management and you could see a default data source would be set.
4. Edit workspace and assign/ unassign data sources, if default data source is unassigned, another default data source would be set.

## Changelog

- feat: support workspace level default data source 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/workspace/server/saved_objects/workspace_data_source_operation_wrapper.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  private throwOperationError() {
+    throw new Error(
+      'You could not create/update/delete data source in workspace, please exit workspace and do it.'
+    );
+  }
+
+  public wrapperFactory: SavedObjectsClientWrapperFactory = (wrapperOptions) => {
+    const isDashboardAdmin = getWorkspaceState(wrapperOptions.request)?.isDashboardAdmin;
+    const isInWorkspace = getWorkspaceState(wrapperOptions.request)?.requestWorkspaceId;
```
SuZhou-Joe, line 38: What if the `workspace` is inside the options? I think this wrapper should be put behind id consumer and only consume the workspaces inside options.

src/plugins/workspace/server/utils.ts
```diff
@@ -113,3 +115,23 @@ export const getDataSourcesList = (client: SavedObjectsClientContract, workspace
       }
     });
 };
+
+export const checkAndSetDefaultDataSources = async (
```
SuZhou-Joe, line 119:
```suggestion
export const checkAndSetDefaultDataSource = async (
```
defaultDataSource should be singular I suppose.

And I guess there are some similar logic in DSM that sets default data source, can we try reuse that?

src/plugins/workspace/server/workspace_client.ts
```diff
@@ -122,6 +125,15 @@ export class WorkspaceClient implements IWorkspaceClientImpl {
           permissions,
         }
       );
+      if (dataSources && this.uiSettings && client) {
+        // This is for setting in workspace environment, otherwise uiSettings can't set workspace level value.
+        updateWorkspaceState(requestDetail.request, {
```
SuZhou-Joe, line 130: I feel a little bit uncomfortable here as workspaceState should be set in the preRoute liftCycle. Can we as least reset the workspaceState once we finish `checkAndSetDefaultDataSources`?

src/plugins/workspace/server/saved_objects/workspace_data_source_operation_wrapper.ts
raintygao: I would prefer to use `InWorkspace`  because what we want to limit is that user operations in the DSM inside workspace, it would look more directly. And this has more priority because id consumer would have some parse logic but this wrapper would limit directly based on role and operations, what do you think?

src/plugins/workspace/common/constants.ts
```diff
@@ -32,10 +33,12 @@ export const WORKSPACE_ID_CONSUMER_WRAPPER_ID = 'workspace_id_consumer';
 
 /**
  * The priority for these wrappers matters:
- * 1. WORKSPACE_ID_CONSUMER wrapper should be the first wrapper to execute, as it will add the `workspaces` field
+ * 1. WORKSPACE_DATA_SOURCE_OPERATION_WRAPPER could be the first wrapper to execute as it is an independent wrapper and not related to other wrappers.
```
SuZhou-Joe, line 36: The wrapper should sit behind id_consumer wrapper I think.

src/plugins/workspace/server/saved_objects/workspace_data_source_operation_wrapper.ts
SuZhou-Joe: I would say: the logic to get `workspaces` should be unified. I do not see there is any special use case for this wrapper to get workspaceId directly from workspaceState.

Imaging that a user calls
```
/w/${workspaceId}/updateObject, { options: { workspaces: null } }
```, should we consider it isInWorkspace or outOfWorkspace?

File: src/plugins/workspace/server/utils.ts [hunk 5]
```diff
@@ -113,3 +115,23 @@ export const getDataSourcesList = (client: SavedObjectsClientContract, workspace
...
+) => {
+  if (dataSources?.length > 0) {
+    if (!isNeededCheck) {
+      // Create# Will set first data source as default data source.
+      await uiSettingsClient.set(DEFAULT_DATA_SOURCE_UI_SETTINGS_ID, dataSources[0]);
+    } else {
+      // Update will check if default DS still exists.
+      const defaultDSId = (await uiSettingsClient.get(DEFAULT_DATA_SOURCE_UI_SETTINGS_ID)) ?? '';
+      const isDefaultDSExist = dataSources.indexOf(defaultDSId) > -1;
+      if (!isDefaultDSExist) {
```
ruanyl, line 132: Nit:
```suggestion
      if (!dataSources.includes(defaultDSId)) {
```

src/plugins/workspace/server/saved_objects/workspace_data_source_operation_wrapper.ts
ruanyl: Agree with @SuZhou-Joe, `options.workspaces` should be the source of truth, that's actually the value we are using to call saved object client to create the object.

src/plugins/workspace/server/utils.ts
```diff
@@ -113,3 +115,23 @@ export const getDataSourcesList = (client: SavedObjectsClientContract, workspace
       }
     });
 };
+
+export const checkAndSetDefaultDataSources = async (
+  uiSettingsClient: IUiSettingsClient,
+  dataSources: string[],
+  isNeededCheck: boolean
```
wanglam, line 122: Shall we separate the `checkAndSetDefaultDataSources` to `checkAndSetDefaultDataSource` and `setDefaultDataSource`? The method name with a `isNeededCheck` flag looks a little bit confused to me.

raintygao: Im open to this, but I think we could combine this logic together with some comments.

raintygao: We'd better not call different methods in create and update.

src/plugins/workspace/server/saved_objects/workspace_data_source_operation_wrapper.ts
wanglam: @raintygao Shall we restrict these data source operations when out of workspace? These client wrapper is for API level right? User can execute CUD operations if this API call doesn't include workspace info.

src/plugins/workspace/server/utils.ts
wanglam: Call with different methods is make sense to me.
For new creating workspace, we don't need to check if default data source has been set. We can just call `setDefaultDataSource` directly.
For new updating workspace, we always check if the default data source has been set. We will call `setDefaultDataSources` if default data source not set.
BTW,  There are two cases about data source been deleted:
1. All data source has been unassigned during workspace update
2. The default data source has been deleted out of workspace.
Will current implementation handle these two cases? Do we need to clear the default data source after data source been deleted?

raintygao: We have judged the lengths of array,we will only set when length is more than 0.

src/plugins/workspace/server/saved_objects/workspace_data_source_operation_wrapper.ts
raintygao: Thanks @SuZhou-Joe @ruanyl , that makes sense, I would try to do some updates and take a look.

src/plugins/workspace/server/utils.ts
wanglam: Yeah. But we won't clear the old default data source if the length of array equal zero, right? For example, create workspace with 3 datasources. The default data source will be the data source 1. And then remove all the data sources in the workspace update page. Will the default data source be cleared? Is there any issues if the default data source still data source 1?

src/plugins/workspace/server/saved_objects/workspace_data_source_operation_wrapper.ts
raintygao: @wanglam This PR is only aiming to limit data source operations in workspace, you could take a look at description of this PR.

ruanyl: @raintygao What about creating a separate PR for the wrapper? It seems it's not related to default data source setting. This way, we can get default data source setting reviewed asap.

src/plugins/workspace/server/utils.ts
raintygao: Could you list any issues if existing?

raintygao: Thanks for your comments,that would be more safe if we do a clear operation if there is no data source left. I would add this.

src/plugins/workspace/server/saved_objects/workspace_data_source_operation_wrapper.ts
raintygao: That sounds good. I will do a separation.

src/plugins/workspace/server/workspace_client.ts
```diff
@@ -248,6 +265,12 @@ export class WorkspaceClient implements IWorkspaceClientImpl {
           version: workspaceInDB.version,
         }
       );
+
+      if (newDataSources && this.uiSettings && client) {
```
yubonluo, line 269: Nit: I think we can remove `newDataSources` from `if (newDataSources && this.uiSettings && client)` because this part of the code is already in `if (newDataSources)`

raintygao: Could you elaborate it more? I suppose this code is not in the block of `if (newDataSources)`.

src/plugins/workspace/server/saved_objects/workspace_data_source_operation_wrapper.ts
raintygao: @ruanyl @SuZhou-Joe I have separated wrapper, there is no unresolved comment about default DS, could we merge it?

src/plugins/workspace/server/utils.ts
```diff
@@ -113,3 +115,25 @@ export const getDataSourcesList = (client: SavedObjectsClientContract, workspace
       }
     });
 };
+
+export const checkAndSetDefaultDataSource = async (
+  uiSettingsClient: IUiSettingsClient,
+  dataSources: string[],
+  isNeededCheck: boolean
```
SuZhou-Joe, line 122:
```suggestion
  needCheck: boolean
```
Nit: `isNeededCheck` seems a little bit too much for me.

src/plugins/workspace/server/utils.ts
```diff
@@ -113,3 +115,25 @@ export const getDataSourcesList = (client: SavedObjectsClientContract, workspace
...
+) => {
+  if (dataSources?.length > 0) {
+    if (!isNeededCheck) {
+      // Create# Will set first data source as default data source.
+      await uiSettingsClient.set(DEFAULT_DATA_SOURCE_UI_SETTINGS_ID, dataSources[0]);
+    } else {
```
SuZhou-Joe, line 128: What if we always check if the default data source is set?

src/plugins/workspace/server/workspace_client.ts
```diff
@@ -122,6 +125,20 @@ export class WorkspaceClient implements IWorkspaceClientImpl {
...
       );
+      if (dataSources && this.uiSettings && client) {
+        const rawState = getWorkspaceState(requestDetail.request);
+        // This is for setting in workspace environment, otherwise uiSettings can't set workspace level value.
+        updateWorkspaceState(requestDetail.request, {
+          requestWorkspaceId: id,
+        });
+        // Set first data source as default after creating workspace
+        const uiSettingsClient = this.uiSettings.asScopedToClient(client);
+        await checkAndSetDefaultDataSource(uiSettingsClient, dataSources, false);
```
SuZhou-Joe, line 136: I'd prefer a try catch here, and move the reset code to `finally` block.

src/plugins/workspace/server/utils.ts
```diff
@@ -113,3 +115,25 @@ export const getDataSourcesList = (client: SavedObjectsClientContract, workspace
...
+      // Update will check if default DS still exists.
+      const defaultDSId = (await uiSettingsClient.get(DEFAULT_DATA_SOURCE_UI_SETTINGS_ID)) ?? '';
+      if (!dataSources.includes(defaultDSId)) {
+        await uiSettingsClient.set(DEFAULT_DATA_SOURCE_UI_SETTINGS_ID, dataSources[0]);
+      }
+    }
+  } else {
+    // If there is no data source left, clear workspace level default data source.
+    await uiSettingsClient.set(DEFAULT_DATA_SOURCE_UI_SETTINGS_ID, undefined);
```
SuZhou-Joe, line 137: Nit: A test to cover this line would be nice.

src/plugins/workspace/server/utils.ts
raintygao: I would be uninclined to do this, we don't need to send an additional request to check default data source for a new workspace even if this could make code less.

src/plugins/workspace/server/utils.ts
raintygao: Updated.

SuZhou-Joe (approved):

ruanyl (dismissed):

SuZhou-Joe (approved):

ruanyl (approved):

PR #7060, opened 2024-06-19
[navigation-next] feat: introduce new interface for group

SuZhou-Joe:
### Description

This PR mainly introduces new interface to register and get useCases from chrome service, and it is a fundamental change for the new navigation.

### Issues Resolved

closes #7061 

## Screenshot

No UI changes.

## Testing the changes

As there is no place using the new interfaces introduced in this PR. You can refer to the test cases here:
src/core/public/chrome/chrome_service.test.ts

## Changelog

- feat: Introduce new interface for group

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`
  - [x] `yarn test:jest_integration`
- [x] New functionality includes testing.
- [x] New functionality has been documented.
- [x] Update [CHANGELOG.md](./../CHANGELOG.md)
- [x] Commits are signed per the DCO using --signoff

PR #7221, opened 2024-07-11
[Workspace] use registered nav groups in workspace form

wanglam:
### Description

In 2.15, we add the workspace use case selector in the workspace form. It used a fixed use cases list. In #7060 , we added a new concept nav group. The workspace form should use these registered nav groups as use cases and displayed in the workspace use case selector. This PR mainly includes below changes:
1. Add available use cases property in workspace selector
2. Subscribe registered use cases in the workspace plugin
3. Pass registered use cases to workspace selector
4. Remove not used `workspaceConfigurableApps$` 

### Issues Resolved

#7222 

## Screenshot

No UI changes

## Testing the changes

The nav groups have not been displayed in the navigation, so we need to use unit tests to verify all changes.

- For  `getRegisteredUseCases$` inside use case service
Can refer src/plugins/workspace/public/services/use_case_service.test.ts
- For app updater and registeredUseCase subscription
Can refer src/plugins/workspace/public/plugin.test.ts
- For workspace creator / updater / overview functionality
Can refer below files:
src/plugins/workspace/public/components/workspace_creator/workspace_creator.test.tsx
src/plugins/workspace/public/components/workspace_updater/workspace_updater.test.tsx
src/plugins/workspace/public/components/workspace_overview/workspace_overview.test.tsx

## Changelog

- feat: Use registered nav group as workspace use case

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/workspace/public/utils.ts
```diff
@@ -202,3 +203,44 @@ export const getDataSourcesList = (client: SavedObjectsStart['client'], workspac
...
+}: NavGroupItemInMap): WorkspaceUseCase => ({
+  id,
+  title,
+  description,
+  features: navLinks.map((item) => item.id),
+  systematic: type === NavGroupType.SYSTEM,
+  order,
+});
+
+export const isEqualWorkspaceUseCase = (a: WorkspaceUseCase, b: WorkspaceUseCase) => {
```
SuZhou-Joe, line 223: Do we need to check if the use case from constant equal to the one from navGroup? I thought match on id should be good enough.

SuZhou-Joe (approved):

src/plugins/workspace/public/utils.ts
wanglam: For now, the `isEqualWorkspaceUseCase` only been used in `UseCaseService`. It's just for avoid duplicate changes fired. The `appUpdater$` in `filterNavLinks` will generate new nav links. Then the `getNavGroupsMap$` will emit new nav groups map. It will generate an infinite loop. So we add `distinctUntilChanged` after `getNavGroupsMap$` to avoid these.

SuZhou-Joe: I c, thanks.

SuZhou-Joe (approved):

ruanyl (approved):

PR #7247, opened 2024-07-15
[Navigation-next]Register workspace list card into home page 

Hailong-am:
### Description

Register workspace list card into home page 

image

**Empty state**

image

### Issues Resolved

#7294

## Testing the changes

1. start OSD with workspace and new navigation enabled, add this to opensearch_dashboards.yml file
```yml
workspace.enabled: true
uiSettings:
  overrides:
    "home:useNewHomePage": true
```
2. start OSD and navigates to home page

## Changelog

- feat: Register workspace list card into home page

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/public/components/service_card/workspace_list_card.tsx
+++ b/src/plugins/workspace/public/components/service_card/workspace_list_card.tsx
@@ -1,0 +1,239 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React, { Component } from 'react';
+import {
+  EuiPanel,
+  EuiLink,
+  EuiDescriptionList,
+  EuiIcon,
+  EuiFlexGroup,
+  EuiFlexItem,
+  EuiSelect,
+  EuiButtonIcon,
+  EuiSpacer,
+  EuiListGroup,
+  EuiText,
+  EuiTitle,
+  EuiToolTip,
+  EuiEmptyPrompt,
+} from '@elastic/eui';
+import { i18n } from '@osd/i18n';
+import { Subscription } from 'rxjs';
+import moment from 'moment';
+import _ from 'lodash';
+import { CoreStart, WorkspaceObject } from '../../../../../core/public';
+import { navigateToWorkspaceDetail } from '../utils/workspace';
+
+import { WORKSPACE_CREATE_APP_ID, WORKSPACE_LIST_APP_ID } from '../../../common/constants';
+import { WorkspaceEntry, recentWorkspaceManager } from '../../recent_workspace_manager';
+
+const WORKSPACE_LIST_CARD_DESCRIPTIOIN = i18n.translate('workspace.list.card.descriptionh', {
+  defaultMessage:
+    'Workspaces are dedicated environments for organizing and collaborating on your data, dashboards, and analytics workflows. Each Workspace acts as a self-contained space with its own set of saved objects and access controls.',
+});
+
+const MAX_ITEM_IN_LIST = 5;
+
+export interface WorkspaceListCardProps {
+  core: CoreStart;
+}
+
+export interface WorkspaceListItem {
+  id: string;
+  name: string;
+  time?: string | number;
+}
+
+export interface WorkspaceListCardState {
+  availiableWorkspaces: WorkspaceObject[];
+  filter: string;
+  workspaceList: WorkspaceListItem[];
+  recentWorkspaces: WorkspaceEntry[];
+}
+
+export class WorkspaceListCard extends Component<WorkspaceListCardProps, WorkspaceListCardState> {
+  private workspaceSub?: Subscription;
+  constructor(props: WorkspaceListCardProps) {
+    super(props);
+    this.state = {
+      availiableWorkspaces: [],
+      recentWorkspaces: [],
+      workspaceList: [],
+      filter: 'viewed',
+    };
+  }
+
+  componentDidMount() {
+    this.setState({
+      recentWorkspaces: recentWorkspaceManager.getRecentWorkspaces() || [],
+    });
+    this.workspaceSub = this.props.core.workspaces.workspaceList$.subscribe((list) => {
+      this.setState({
+        availiableWorkspaces: list || [],
+      });
+    });
+    this.loadWorkspaceListItems();
+  }
+
+  componentDidUpdate(
+    prevProps: Readonly<WorkspaceListCardProps>,
+    prevState: Readonly<WorkspaceListCardState>
+  ): void {
+    if (
+      !_.isEqual(prevState.filter, this.state.filter) ||
+      !_.isEqual(prevState.availiableWorkspaces, this.state.availiableWorkspaces) ||
+      !_.isEqual(prevState.recentWorkspaces, this.state.recentWorkspaces)
+    ) {
+      this.loadWorkspaceListItems();
+    }
+  }
+
+  private loadWorkspaceListItems() {
+    if (this.state.filter === 'viewed') {
+      this.setState({
+        workspaceList: _.orderBy(this.state.recentWorkspaces, ['timestamp'], ['desc'])
+          .filter((ws) => this.state.availiableWorkspaces.some((a) => a.id === ws.id))
+          .slice(0, MAX_ITEM_IN_LIST)
+          .map((item) => ({
+            id: item.id,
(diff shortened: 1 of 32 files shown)


virajsanghvi:
I haven't looked at this yet, but just wanted to call out that some tests were failing and code coverage for patch is slightly lower than target - have you taken a look at these?

Hailong-am:
> I haven't looked at this yet, but just wanted to call out that some tests were failing and code coverage for patch is slightly lower than target - have you taken a look at these?


Thanks for looking into this, i need to rebase from main and change the target branch to main. Mark this as dfart now.

ruanyl:
Looks like unnecessary commits are included, maybe you can do a cherry pick instead

src/plugins/workspace/public/components/service_card/workspace_list_card.tsx
```diff
@@ -0,0 +1,239 @@
...
+  EuiListGroup,
+  EuiText,
+  EuiTitle,
+  EuiToolTip,
+  EuiEmptyPrompt,
+} from '@elastic/eui';
+import { i18n } from '@osd/i18n';
+import { Subscription } from 'rxjs';
+import moment from 'moment';
+import _ from 'lodash';
```
ruanyl, line 26: Prefer to only import what's needed, for example:
```typescript
import isEqual from 'lodash/isEqual'
```

src/plugins/workspace/public/components/service_card/workspace_list_card.tsx
```diff
@@ -0,0 +1,239 @@
...
+      recentWorkspaces: [],
+      workspaceList: [],
+      filter: 'viewed',
+    };
+  }
+
+  componentDidMount() {
+    this.setState({
+      recentWorkspaces: recentWorkspaceManager.getRecentWorkspaces() || [],
+    });
```
ruanyl, line 72: Can we simply init the `recentWorkspaces` in constructor?
```typescript
this.state = {
  availiableWorkspaces: [],
  recentWorkspaces: recentWorkspaceManager.getRecentWorkspaces() || [],
  workspaceList: [],
  filter: 'viewed',
};
```

src/plugins/workspace/public/components/service_card/workspace_list_card.tsx
```diff
@@ -0,0 +1,239 @@
...
+      !_.isEqual(prevState.recentWorkspaces, this.state.recentWorkspaces)
+    ) {
+      this.loadWorkspaceListItems();
+    }
+  }
+
+  private loadWorkspaceListItems() {
+    if (this.state.filter === 'viewed') {
+      this.setState({
+        workspaceList: _.orderBy(this.state.recentWorkspaces, ['timestamp'], ['desc'])
```
ruanyl, line 97: It seems `workspaceList` can be derived from existing state, I think we can simply compute it in the `render()` function.

If you use functional react component, you can use `useMemo` to avoid unnecessary recompute. Btw, why not use functional component? I can see it will make it implementation much simple

Hailong-am: > Btw, why not use functional component? I can see it will make it implementation much simple

when i try using function component, dashboard embeddable not able to render it

src/plugins/workspace/public/components/service_card/workspace_list_card.tsx
Hailong-am: thanks, updated

src/plugins/workspace/public/components/service_card/workspace_list_card.tsx
Hailong-am: thanks, updated

src/plugins/workspace/public/components/service_card/workspace_list_card.tsx
ruanyl: Oh, that's odd, then I think you can keep the current implementation, I will investigate a bit on this, thanks ;)

ruanyl (approved):

src/plugins/workspace/public/components/service_card/workspace_list_card.tsx
```diff
@@ -0,0 +1,217 @@
...
+import { Subscription } from 'rxjs';
+import moment from 'moment';
+import { orderBy } from 'lodash';
+import { CoreStart, WorkspaceObject } from '../../../../../core/public';
+import { navigateToWorkspaceDetail } from '../utils/workspace';
+
+import { WORKSPACE_CREATE_APP_ID, WORKSPACE_LIST_APP_ID } from '../../../common/constants';
+import { WorkspaceEntry, recentWorkspaceManager } from '../../recent_workspace_manager';
+
+const WORKSPACE_LIST_CARD_DESCRIPTIOIN = i18n.translate('workspace.list.card.descriptionh', {
```
ruanyl, line 33: typo
```suggestion
const WORKSPACE_LIST_CARD_DESCRIPTION = i18n.translate('workspace.list.card.description', {
```

src/plugins/workspace/public/components/service_card/workspace_list_card.tsx
```diff
@@ -0,0 +1,217 @@
...
+}
+
+export interface WorkspaceListItem {
+  id: string;
+  name: string;
+  time?: string | number;
+}
+
+export interface WorkspaceListCardState {
+  availiableWorkspaces: WorkspaceObject[];
```
ruanyl, line 51:
```suggestion
  availableWorkspaces: WorkspaceObject[];
```

src/plugins/workspace/public/components/service_card/workspace_list_card.tsx
```diff
@@ -0,0 +1,217 @@
...
+    'Workspaces are dedicated environments for organizing and collaborating on your data, dashboards, and analytics workflows. Each Workspace acts as a self-contained space with its own set of saved objects and access controls.',
+});
+
+const MAX_ITEM_IN_LIST = 5;
+
+export interface WorkspaceListCardProps {
+  core: CoreStart;
+}
+
+export interface WorkspaceListItem {
```
ruanyl, line 44: Nit: no need to export, seems not used anywhere else

src/plugins/workspace/public/components/service_card/workspace_list_card.tsx
Hailong-am: thanks to @ruanyl help on debugging, seems functional component don't any issue, convert to functional component now.

src/plugins/workspace/public/components/service_card/workspace_list_card.tsx
Hailong-am: yes, removed

src/plugins/workspace/public/plugin.ts
```diff
@@ -407,9 +415,30 @@ export class WorkspacePlugin implements Plugin<{}, {}, WorkspacePluginSetupDeps>
...
     return {};
   }
 
+  private registerWorkspaceListToHome(
+    core: CoreStart,
+    contentManagement: ContentManagementPluginStart
+  ) {
+    if (contentManagement) {
+      contentManagement.registerContentProvider({
+        id: HOME_PAGE_ID,
```
SuZhou-Joe, line 430: I guess the id is used for the content, not the page.

SuZhou-Joe (approved):

src/plugins/workspace/public/components/service_card/workspace_list_card.tsx
```diff
@@ -0,0 +1,192 @@
...
+                  application.navigateToApp(WORKSPACE_CREATE_APP_ID);
+                }}
+              />
+            </EuiToolTip>
+          </EuiFlexItem>
+        )}
+      </EuiFlexGroup>
+
+      <EuiSpacer />
+      <EuiListGroup flush={true} bordered={false} style={{ minHeight: '300px' }}>
```
ruanyl, line 145: Nit: I'd suggest to have a future improvement task make this service card a flex container, so we don't need to specify a fix number 300px to stretch the list.

src/plugins/workspace/public/plugin.ts
ruanyl: Yes, this is the content provider id, you can give it any unique string, like `workspace_list_card`

Hailong-am: thanks for catching this, update to `workspace_list_card_home`

ruanyl (approved):

SuZhou-Joe (approved):

PR #7285, opened 2024-07-17
feat: support PPL in vega visualization

ruanyl:
This commit add PPL support in Vega visualization, it rely on query enhancement plugin to support search with PPL

Vega visualization UI changes is out of the scope of this PR, will add UI changes in separate commits

### Description
image

## Changelog

- feat: support PPL in vega visualization

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/server/search/routes/search.ts
```diff
@@ -62,7 +62,7 @@ export function registerSearchRoute(
       try {
         const { withLongNumeralsSupport, ...response } = await selfStart.search.search(
           context,
-          { ...searchRequest, id },
+          { ...searchRequest, id, rawRequest: request },
```
kavilla, line 65: i wouldn't update this. when it's ppl, the whole request should be passed as long as you just send it

ruanyl: You mean we check the `strategy` here, and when it's `ppl`, we send the whole request instead of request.body?

SuZhou-Joe (approved):

SuZhou-Joe (dismissed):

src/plugins/data/server/search/routes/search.ts
ruanyl: The `client.asScoped(request)` we call in search strategy requires the http request object, currently I see no way to get it unless pass it specifically.

src/plugins/vis_type_vega/public/data_model/ppl_parser.ts
kavilla, line 1: would it be possible to put like a parsers map in the search service? ideally. all non-out of the box experience is encapsulated within the query enhancements. so the parsers service can host the parser and take in a search strategy id. then in query enhancements we register the ppl

avoids having to re-write shared logic too

ananzh (approved):

SuZhou-Joe (approved):

src/plugins/vis_type_vega/public/data_model/ppl_parser.ts
ruanyl: Agree, let's explore the reusability, especially with considering the vis builder+vega which requires to make the requirement clear both from technical and design perspective, I'll leave it for future improvement, this PR is just the first step to add ppl support in vega

PR #7212, opened 2024-07-09
[Discover-next] add query enhancements plugin as a core plugin

kavilla:
***NOTE***: Will update this and ~~move some of the stuff that was added if it was a 3rd party plugin~~ and will add justifications based on this rubric: 

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/7081#issuecomment-2201575495

Also need to swap the default values around.

### Description

Introduces the addition of a new core plugin called `Query Enhancements`

This plugin seems to be focused on enhancing query capabilities, particularly in areas like PPL (Piped Processing Language), SQL, and natural language query assistance.

#### Plugin Configuration and Setup

The plugin is set up with the necessary configuration files:

- src/plugins/query_enhancements/opensearch_dashboards.json: Defines the plugin's metadata and dependencies.
- src/plugins/query_enhancements/common/config.ts: Contains the configuration schema for the plugin.
- src/plugins/query_enhancements/public/plugin.tsx: The main plugin file that sets up the public-facing part of the plugin.
- src/plugins/query_enhancements/server/plugin.ts: The server-side setup for the plugin.

#### Query Assist Feature

A significant part of the plugin is dedicated to a "Query Assist" feature:

- src/plugins/query_enhancements/public/query_assist/: This directory contains components, hooks, and utilities for the Query Assist feature.
- src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx: A React component for the Query Assist bar.
- src/plugins/query_enhancements/server/routes/query_assist/: Server-side routing for Query Assist functionality.

#### Search Interceptors

The plugin introduces several search interceptors:

- src/plugins/query_enhancements/public/search/ppl_search_interceptor.ts
- src/plugins/query_enhancements/public/search/sql_search_interceptor.ts
- src/plugins/query_enhancements/public/search/sql_async_search_interceptor.ts

These interceptors likely modify or enhance the search functionality for different query languages.

#### Data Source Connection

There's a new feature for data source connections:

- src/plugins/query_enhancements/public/data_source_connection/: This directory contains components and services for managing data source connections.

#### Server-side Search Strategies

The plugin implements server-side search strategies for different query types:

- src/plugins/query_enhancements/server/search/ppl_search_strategy.ts
- src/plugins/query_enhancements/server/search/sql_search_strategy.ts
- src/plugins/query_enhancements/server/search/sql_async_search_strategy.ts

#### Modifications to Existing Files

There are some modifications to existing files outside the new plugin directory:

[...]

PR #7291, opened 2024-07-17
[Discover-next] remove plugin config toggle for data enhancements

kavilla:
### Description

Remove the data config toggle for enhancements and rely on the UI settings for being enabled. Also removes
the readonly UI setting that was previously added to prevent confusion unless users knew what they were using.

### Issues Followup

https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7212

## Changelog

- deprecate: Remove data enhancements config and readonly flag. Removes dead url link,

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/settings/settings.ts
+++ b/src/plugins/data/public/ui/settings/settings.ts
@@ -35,7 +35,7 @@ export class Settings {
     private readonly queryEnhancements: Map<string, QueryEnhancement>,
     private readonly queryEditorExtensionMap: Record<string, QueryEditorExtensionConfig>
   ) {
-    this.isEnabled = this.config.enabled;
+    this.isEnabled = true;
     this.setUserQueryEnhancementsEnabled(this.isEnabled);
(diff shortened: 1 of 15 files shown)


ashwin-pc (dismissed):

src/plugins/data/public/ui/settings/settings.ts
```diff
@@ -35,7 +35,7 @@ export class Settings {
     private readonly queryEnhancements: Map<string, QueryEnhancement>,
     private readonly queryEditorExtensionMap: Record<string, QueryEditorExtensionConfig>
   ) {
-    this.isEnabled = this.config.enabled;
+    this.isEnabled = true;
```
ashwin-pc, line 38: Why is this always true?

kavilla: ah that actually makes sense not to be true good catch.

AMoo-Miki (approved):

ashwin-pc (approved):

PR #7214, opened 2024-07-10
[DataSource] Restrict to manage data source on the DSM UI

yubonluo:
### Description
Restrict to edit data source on the DSM UI.
manageableBy:
```
in workspace:
	anyone can not manage data source on the DSM
out of workspace:
	none: anyone can not manage data source on the DSM
	dashboard_admin: only OSD admin can manage source on the DSM
	all: anyone can manage data source on the DSM
in the saved object management page:
         all: anyone can delete/import data source
	dashboard_admin: only OSD admin can delete/import source
	none: anyone can not delete data source
```
### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6889

## Screenshot
manageableBy="all" or
manageableBy="dashboard_admin" and user is OSD admin
image
image
image

manageableBy="none" or
manageableBy="dashboard_admin" and user is not OSD admin or
user in a workspace
image
image
image

## Testing the changes
1. Clone the latest osd/security-dashboard plugin code and run yarn osd bootstrap
2. Modify config/opensearch_dashboards.yml
```yml
opensearchDashboards.dashboardAdmin.groups: ["dashboard_admin"]
opensearchDashboards.dashboardAdmin.users: ["username"]
workspace.enabled: true 
opensearch.username: "username"
opensearch.password: "password" 
opensearch.ssl.verificationMode: none
savedObjects.permission.enabled: true
data_source.enabled: true
data_source.manageableBy: "all|dashboard_admin|none"
```
3. Run yarn start --no-base-path
4. Operate data source.

## Changelog
- feat: [DataSource] Restrict to edit data source on the DSM UI.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #7296, opened 2024-07-18
[Workspace] Recover data source management in workspace

raintygao:
### Description

Data source is a global level saved object, so we disable data source management when in workspace. 
For now, data source is a workspace level object, so we should recover data source management in workspace and disable create/update/delete operation in frontend(https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7214 has implemented).

## Screenshot

image

## Testing the changes
Enter a workspace then you can visit data source management then will see workspace level data source objects and there is not entry for creating/updating/deleting.

## Changelog

- feat: Recover data source management in workspace

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


SuZhou-Joe:
Hi @raintygao I think we can close this PR because:
When new navigation is enabled, we will show data source in each of the use case.
When new navigation is disabled, we had a assumption that workspace should be disabled as well.

SuZhou-Joe (approved):

ZilongX:
Pick up the main and re-running all checks

ZilongX:
> Hi @raintygao I think we can close this PR because: When new navigation is enabled, we will show data source in each of the use case. When new navigation is disabled, we had a assumption that workspace should be disabled as well.

So do we still need this change to be merged as part of 2.16.0 ?

raintygao:
> > Hi @raintygao I think we can close this PR because: When new navigation is enabled, we will show data source in each of the use case. When new navigation is disabled, we had a assumption that workspace should be disabled as well.
> 
> So do we still need this change to be merged as part of 2.16.0 ?

Yes, after discussing we can merge it in 2.16 as this is a very tiny change. CC @SuZhou-Joe @ruanyl

ruanyl (approved):

PR #7305, opened 2024-07-18
[navigation-next] fix: redirect to standard index pattern applications while nav group is enabled

SuZhou-Joe:
### Description

When nav group is enabled and users want to visit `Dashboards` page if no index pattern is configured in their OSD, OSD will direct user to the index pattern page, which is a sub app under management.

This PR is mainly to solve this problem by redirecting users to the right application.

- Address the issue that in-app navigation is not hide when nav group is enabled.
- Address the issue that the app will still navigate to legacy index pattern management oage.
- Address the scss code concern from @AMoo-Miki 

### Before the fix
https://github.com/user-attachments/assets/34df4f15-9d2d-44ea-8875-bb51db117751

### After the fix
https://github.com/user-attachments/assets/a902d738-25a3-4a7a-a685-882531eb6a3c

## Changelog

- feat: [navigation-next] fix: redirect to standard index pattern applications while nav group is enabled

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


src/core/public/chrome/nav_group/nav_group_service.ts
```diff
@@ -213,7 +213,7 @@ export class ChromeNavGroupService {
     };
 
     // erase current nav group when switch app don't belongs to any nav group
-    application.currentAppId$.subscribe((appId) => {
+    application.currentAppId$.pipe(debounceTime(500)).subscribe((appId) => {
```
virajsanghvi, line 216: Why do you need to debounce?

virajsanghvi (commented):
Can you update after video (think its the same as before)?

src/core/public/chrome/nav_group/nav_group_service.ts
SuZhou-Joe: Sure, added some comment.

SuZhou-Joe:
> Can you update after video (think its the same as before)?

Thanks for catching this, updated with correct video.

src/core/public/chrome/ui/header/collapsible_nav_group_enabled.scss
```diff
@@ -2,8 +2,8 @@
   border: none !important;
 
   .nav-link-item {
-    padding: $ouiSize / 4 $ouiSize;
-    border-radius: $ouiSize;
+    padding: calc($euiSize / 4) $euiSize;
+    border-radius: $euiSize;
```
ruanyl, line 6: Are these changes expected? Seems unrelated

SuZhou-Joe: It is expected, as Miki told me to do so.

src/core/public/chrome/nav_group/nav_group_service.ts
ruanyl: I feel it would be better to have the debounceTime here, so that the consumer that subscribe to `currentNavGroupSorted$` to update UI wouldn't cause the UI flashing when the value changed so frequently.

```typescript
    const currentNavGroupSorted$ = combineLatest([
      this.getSortedNavGroupsMap$(),
      this.currentNavGroup$,
    ])
      .pipe(takeUntil(this.stop$))
      .pipe(debounceTime(500))
      .pipe(
        map(([navGroupsMapSorted, currentNavGroup]) => {
          if (currentNavGroup) {
            return navGroupsMapSorted[currentNavGroup.id];
          }
        })
      );

```

ruanyl: Btw, what's the consequence if no `debounceTime(500)`? Anything UI broken?

SuZhou-Joe: I just removed the reset logic because:
When redirecting to discover, the appId will get changed twice, first is discover the latter is data-explorer. The reset logic will always reset the currentNavGroup to undefined because data-explorer is a hidden application that won't be registered in left navigation in the left navigation.

Hailong-am: > The effect is that only when user click the `back` button in the new navigation will we reset the currentNavGroup to undefined.

make sense

PR #7318, opened 2024-07-19
[Workspace]Add "All use case" option to workspace form

wanglam:
### Description

The use case of workspace can only be one specific use case or all use case. This PR mainly includes below changes:
1. Modify multi use case selection to single and add "all use case" option
2. Avoid application been set to Inaccessible when workspace using all use case and application in any use case
3. Avoid nav group been hidden when workspace using all use case

### Issues Resolved

#7319 

## Screenshot

image

## Testing the changes

- Check out branch and Run `yarn osd boostrap`
- Add below config to the `opensearch_dashboards.yml`
```
workspace.enabled: true
uiSettings.overrides["home:useNewHomePage"]: true
```
- Run `yarn start --no-base-path` and login with admin user
- Go to [workspace create page](http://localhost:5601/app/workspace_create)
- Fill workspace unique name and select All use case
- Click create workspace and it will jump to workspace detail page like below image
image

(The "all use case" will be selected and the left navigation will changed to all style)

## Changelog

- feat: [Workspace]Add "All use case" option to workspace form

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/public/plugin.ts
+++ b/src/plugins/workspace/public/plugin.ts
@@ -121,6 +121,16 @@ export class WorkspacePlugin implements Plugin<{}, {}, WorkspacePluginSetupDeps>
     this.currentWorkspaceSubscription = currentWorkspace$.subscribe((currentWorkspace) => {
       if (currentWorkspace) {
         this.navGroupUpdater$.next((navGroup) => {
+          /**
+           * The following logic determines whether a navigation group should be hidden or not based on the workspace's feature configurations.
+           * It checks the following conditions:
+           * 1. The navigation group is not a system-level group (system groups are always visible).
+           * 2. The current workspace has feature configurations set up.
+           * 3. The current workspace's use case it not "All use case".
+           * 4. The current navigation group is not included in the feature configurations of the workspace.
+           *
+           * If all these conditions are true, it means that the navigation group should be hidden.
+           */
           if (
             navGroup.type !== NavGroupType.SYSTEM &&
             currentWorkspace.features &&
--- a/src/plugins/workspace/public/plugin.ts
+++ b/src/plugins/workspace/public/plugin.ts
@@ -21,6 +21,7 @@ import {
   NavGroupStatus,
   DEFAULT_NAV_GROUPS,
   NavGroupType,
+  ALL_USE_CASE_ID,
 } from '../../../core/public';
 import {
   WORKSPACE_FATAL_ERROR_APP_ID,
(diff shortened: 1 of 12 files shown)


src/plugins/workspace/public/utils.ts
```diff
@@ -32,14 +37,14 @@ export const getUseCaseFromFeatureConfig = (featureConfig: string) => {
...
-  const useCase = useCases.find(({ id }) => id === getUseCaseFromFeatureConfig(featureConfig));
-  if (useCase) {
-    return useCase.features.includes(featureId);
-  }
-  return false;
+  const availableFeatures =
+    useCaseId === ALL_USE_CASE_ID
+      ? useCases.reduce<string[]>((previous, { features }) => previous.concat(features ?? []), [])
+      : useCases.find(({ id }) => id === useCaseId)?.features ?? [];
+  return availableFeatures.includes(featureId);
```
Hailong-am, line 47: if an app have not register to any use case, will it be visible/accesable in all use case?

src/plugins/workspace/public/plugin.ts
```diff
@@ -122,6 +124,7 @@ export class WorkspacePlugin implements Plugin<{}, {}, WorkspacePluginSetupDeps>
           if (
             navGroup.type !== NavGroupType.SYSTEM &&
             currentWorkspace.features &&
+            getFirstUseCaseOfFeatureConfigs(currentWorkspace.features) !== ALL_USE_CASE_ID &&
```
ruanyl, line 127: Would be nice to have a comment on this logic

ruanyl (dismissed):

src/plugins/workspace/public/utils.ts
```diff
@@ -32,14 +37,14 @@ export const getUseCaseFromFeatureConfig = (featureConfig: string) => {
...
-  featureConfig: string,
+  useCaseId: string,
   useCases: WorkspaceUseCase[]
 ) => {
-  const useCase = useCases.find(({ id }) => id === getUseCaseFromFeatureConfig(featureConfig));
-  if (useCase) {
-    return useCase.features.includes(featureId);
-  }
-  return false;
+  const availableFeatures =
```
SuZhou-Joe, line 43: I thought we will just bypass all the features in all use case. Why do we change that?

src/plugins/workspace/public/utils.ts
wanglam: Based current implementation, the all use case will match all features inside registered use cases. If an app not register to any use case, it still can be accessible if it match below conditions:
1. `workspaceAvailability` is `outsideWorkspace` 
2. `navLinkStatus` equal `hidden`
3. `chromeless` equal `true`

SuZhou-Joe (approved):

src/plugins/workspace/public/utils.ts
wanglam: Do you mean we don't need to check whether the application id exists in registered use cases? We add skip the feature verification logic outside for all use case workspace.

SuZhou-Joe (approved):

Hailong-am (approved):

ruanyl (approved):

PR #7337, opened 2024-07-21
Track progress updates in async querying interceptors

Swiddis:
A backport of https://github.com/kavilla/queryEnhancements/pull/30, this PR adds progress tracking for async queries via UI Actions, so callers are able to implement features like progress bars when waiting on long-running queries.

### Issues Resolved

N/A

## Screenshot

N/A

## Testing the changes

The client events were received when developing a loading progress POC, but getting a full set of unit tests with the UI Actions bus might be nontrivial.

### Check List

- [ ] All tests pass
  - [X] `yarn test:jest`


Swiddis:
Failures seem unrelated? Tests pass if I run `yarn test:jest query_enhancements` locally, I get four failures from `LogRotator` which seems to be related to something involving temporary files. This is a non-functional change (creates ui actions in the background) so I wouldn't expect any visible failures based on this.

Swiddis:
Copying the usage instructions from when the same PR was raised on [the old repo](https://github.com/kavilla/queryEnhancements/pull/30)

---

Alright: got something actually working. Sends query data to the `ASYNC_TRIGGER_ID` trigger. From the caller side you need to make a Subject that accepts the async query type. (To avoid cross-plugin imports this type also has to be redefined). Here's a sample method to create one of those (given a constant with the same async trigger id, and assuming the trigger ids have been registered):

```ts
import { Subject } from 'rxjs';
import { UiActionsStart, createAction } from '../../../ui_actions/public';
import { ASYNC_TRIGGER_ID, ASYNC_ACTION_ID } from '../../common';

interface AsyncQueryContext {
  queryId: string;
  queryStatus: string;
}

export function createQueryProgressSubject(uiActions: UiActionsStart): Subject<AsyncQueryContext> {
  const subject: Subject<AsyncQueryContext> = new Subject();
  uiActions.addTriggerAction(
    ASYNC_TRIGGER_ID,
    createAction<typeof ASYNC_ACTION_ID>({
      id: ASYNC_ACTION_ID,
      execute: async (context: unknown) => subject.next(context as AsyncQueryContext),
    })
  );
  return subject;
}
```

Then from the plugin `start` method you can create a subject:

```ts
const progressSubject = createQueryProgressAction(uiActions);
```

And pass this to the front-end where you can subscribe in order to update the progress:

```ts
const [progress, setProgress] = useState('NONE');
data.progressSubject.subscribe((v) => setProgress(v.queryStatus));
```

Pretty roundabout and far from ideal (requires lots of boilerplate to onboard the progress subject + query types + registration + passing all the refs around), but it gets the job done.

Swiddis:
Pulling out of draft tentatively -- still need to figure out what's going on with this changelog action (it says `feat` is an acceptable prefix but seems to be forcing a manual changelog update?)

Swiddis:
How is the changelog action supposed to work?

src/plugins/data/public/search/search_interceptor.ts
```diff
@@ -52,6 +53,7 @@ export interface SearchInterceptorDeps {
   startServices: Promise<[CoreStart, any, unknown]>;
   toasts: ToastsSetup;
   usageCollector?: SearchUsageCollector;
+  uiActions: UiActionsStart;
```
kavilla, line 56: is uiactions not available in core start? since this one is the default interceptor did we want to consider making part of the query enhancements plugin's extension of this interceptor

Swiddis: How would I add it as an extension? I could make it an optional field

Swiddis: Made the dep optional

src/plugins/query_enhancements/common/types.ts
```diff
@@ -13,3 +14,45 @@ export interface FetchDataFrameContext {
...
+ * @returns The corresponding {@link JobState} if one exists, otherwise null.
+ */
+export const parseJobState = (maybeState: unknown): JobState | null => {
+  if (!maybeState || typeof maybeState !== 'string') {
+    return null;
+  }
+  maybeState = maybeState.toUpperCase();
+  const result = Object.values(JobState).find((state) => state === maybeState);
+  return result ?? null;
+};
```
AMoo-Miki, line 48: Let's move `Object.values(JobState)` outside so we don't produce it for every call.
```
const jobStatesMap = Object.values(x).reduce(
    (acc, value) => {
      acc.set(value.toUpperCase(), value); 
      return acc;
    }, 
    new Map()
  ); 
```

```suggestion
export const parseJobState = (maybeState: unknown): JobState | undefined =>
  return jobStatesMap.get(maybeState?.toUpperCase?.());
```

Swiddis:
Update: now depends on #7368, diff will look much larger until that one is merged

src/plugins/query_enhancements/public/search/sql_async_search_interceptor.ts
```diff
@@ -7,6 +7,8 @@ import { trimEnd } from 'lodash';
 import { BehaviorSubject, Observable, throwError } from 'rxjs';
 import { i18n } from '@osd/i18n';
 import { concatMap, map } from 'rxjs/operators';
+import { UiActionsStart } from 'src/plugins/ui_actions/public';
+import uuid from 'uuid';
```
AMoo-Miki: Let's use a specific version so if they change their default, we will still produce similar ones.
```suggestion
import { v4 as uuidv4 } from 'uuid';
```

Swiddis: Updated -- though I'm not sure if it matters since they're short-lived uuids that are generated on a per-query-run basis and don't need to be persisted

src/plugins/query_enhancements/public/search/sql_async_search_interceptor.ts
```diff
@@ -71,33 +78,43 @@ export class SQLAsyncSearchInterceptor extends SearchInterceptor {
       ...dataFrame.meta,
       queryConfig: {
         ...dataFrame.meta.queryConfig,
-        ...(this.connectionsService.getSelectedConnection() &&
-          this.connectionsService.getSelectedConnection()?.dataSource && {
-            dataSourceId: this.connectionsService.getSelectedConnection()?.dataSource.id,
-          }),
+        ...(this.connectionsService.getSelectedConnection() && {
+          dataSourceId: this.connectionsService.getSelectedConnection()?.id,
```
AMoo-Miki: I don't know how expensive `this.connectionsService.getSelectedConnection()` is but it would be a lot easier to understand if we just did:

```
const selectedConnection = this.connectionsService.getSelectedConnection();
if (selectedConnection) dataFrame.meta.queryConfig.dataSourceId = selectedConnection.id;

File: src/plugins/query_enhancements/public/search/sql_async_search_interceptor.ts [hunk 5]
```diff
@@ -71,33 +78,43 @@ export class SQLAsyncSearchInterceptor extends SearchInterceptor {
...
-      }
-      if (pollingResult && pollingResult.body.meta.status === 'FAILED') {
-        const jsError = new Error(pollingResult.data.error.response);
-        this.deps.toasts.addError(jsError, {
-          title: i18n.translate('queryEnhancements.sqlQueryError', {
-            defaultMessage: 'Could not complete the SQL async query',
-          }),
-          toastMessage: pollingResult.data.error.response,
+      if (pollingResult) {
+        const queryStatus = parseJobState(pollingResult.body.meta.status)!;
```
AMoo-Miki: I don't think we should trust that we will always get a valid `body.meta.status`.

AMoo-Miki (changes requested):
Changes not related to this PR were introduced.

src/plugins/query_enhancements/common/types.ts
Swiddis: Implemented, and grabbed some low-hanging-fruit test coverage

src/plugins/query_enhancements/public/search/sql_async_search_interceptor.ts
Swiddis: In the updates from #7368 we don't depend on `connectionsService` at all

src/plugins/query_enhancements/public/search/sql_async_search_interceptor.ts
Swiddis: Refactored to guard against null: if an invalid or missing status is returned we'll report an error and stop polling.

Swiddis:
Upstream PR is merged, ready for review again

src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -117,6 +123,21 @@ export class QueryEnhancementsPlugin
...
 
+    data.__enhance({
+      ui: {
+        queryEditorExtension: createDataSourceConnectionExtension(
+          this.connectionsService,
+          core.notifications.toasts,
+          this.config
+        ),
+      },
+    });
```
joshuali925, line 134: this shouldn't be added? it was removed in 7426

src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -16,6 +17,9 @@ import {
   QueryEnhancementsPluginStart,
   QueryEnhancementsPluginStartDependencies,
 } from './types';
+import { ASYNC_TRIGGER_ID } from '../common';
+
+export type PublicConfig = Pick<ConfigSchema, 'queryAssist'>;
```
joshuali925, line 22: this is also not used

src/plugins/query_enhancements/common/types.test.ts
```diff
@@ -0,0 +1,33 @@
...
+describe('parseJobState', () => {
+  it('should return the correct JobState for a valid string', () => {
+    expect(parseJobState('SUBMITTED')).toBe(JobState.SUBMITTED);
+    expect(parseJobState('waiting')).toBe(JobState.WAITING);
+    expect(parseJobState('SCHEDULED')).toBe(JobState.SCHEDULED);
+    expect(parseJobState('running')).toBe(JobState.RUNNING);
+    expect(parseJobState('FAILED')).toBe(JobState.FAILED);
+    expect(parseJobState('success')).toBe(JobState.SUCCESS);
+    expect(parseJobState('CANCELLING')).toBe(JobState.CANCELLING);
+    expect(parseJobState('cancelled')).toBe(JobState.CANCELLED);
```
ashwin-pc, line 17: nit: why do we have different cases for the states, some are capitalized while others arent

src/plugins/query_enhancements/server/search/ppl_search_strategy.ts
```diff
@@ -19,7 +18,7 @@ import { getFields } from '../../common/utils';
 import { Facet } from '../utils';
 
 export const pplSearchStrategyProvider = (
-  config$: Observable<SharedGlobalConfig>,
+  _config$: Observable<SharedGlobalConfig>,
```
ashwin-pc, line 21: Nice!

src/plugins/query_enhancements/public/search/sql_search_interceptor.ts
```diff
@@ -112,28 +119,42 @@ export class SQLSearchInterceptor extends SearchInterceptor {
         ...dataFrame.meta.queryConfig,
       },
     };
+    const queryId = uuidv4();
+    // Send an initial submit event to get faster feedback to clients waiting for info, since
+    // polling will wait for the first polling cycle to finish before sending anything
+    this.uiActions?.getTrigger(ASYNC_TRIGGER_ID).exec({
```
ashwin-pc, line 125: Nice!

ashwin-pc (commented):
This looks good, just curious on where are you displaying the loading state?

src/plugins/query_enhancements/common/types.test.ts
Swiddis: Testing case-insensitivity, though there's probably a way to make it more obvious

src/plugins/query_enhancements/common/types.ts
```diff
@@ -13,3 +14,48 @@ export interface FetchDataFrameContext {
...
+ * seems to use this state instead. See also: {@link parseJobState}.
+ */
+export enum JobState {
+  SUBMITTED = 'SUBMITTED',
+  WAITING = 'WAITING',
+  SCHEDULED = 'SCHEDULED',
+  RUNNING = 'RUNNING',
+  FAILED = 'FAILED',
+  SUCCESS = 'SUCCESS',
+  CANCELLING = 'CANCELLING',
```
sejli, line 30: Nit: We'll probably centralize all of these types post 2.16, think there's 3 locations now where this exists. Aside from that, is `CANCELLING` something new from the backend API? Don't remember this being in the old `DirectQueryLoadingStatus` type.

Swiddis:
> This looks good, just curious on where are you displaying the loading state?

This is the framework PR, rendering the progress on the UI is separate. There was a POC I used for testing but I haven't had time to implement the full UI.

src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -117,6 +121,11 @@ export class QueryEnhancementsPlugin
       },
     });
 
+    const ASYNC_TRIGGER: Trigger<typeof ASYNC_TRIGGER_ID> = {
+      id: ASYNC_TRIGGER_ID,
+    };
+    uiActions.registerTrigger(ASYNC_TRIGGER);
```
sejli, line 127: Piggy-backing off @ashwin-pc's comment, this PR sets up `queryEnhancements` so that after we have this trigger registered, we can use it to execute any progress updates, like when we add async stuff to PPL?

sejli (approved):
LGTM

src/plugins/query_enhancements/common/types.ts
Swiddis: I ported all the [EMR job run states](https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/job-states.html), with one of them changed since the flint API seems to use a different term. I'm not sure what all of the states we actually use are.

kavilla: definitely should be one. incase the job is just too long and we cant abort it from the front end

src/plugins/data/public/search/search_interceptor.ts
```diff
@@ -52,6 +53,7 @@ export interface SearchInterceptorDeps {
   startServices: Promise<[CoreStart, any, unknown]>;
   toasts: ToastsSetup;
   usageCollector?: SearchUsageCollector;
+  uiActions?: UiActionsStart;
```
kavilla, line 56: would we be able to add ui actions without adding to the search interceptor like the query enhancements takes a dependency on it

src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
```diff
@@ -9,7 +9,7 @@ import {
   IDataPluginServices,
   PersistedLog,
   QueryEditorExtensionDependencies,
-} from '../../../../data/public';
+} from 'src/plugins/data/public';
```
kavilla, line 12: this one might be fine now.  but sometimes safer based on the compiled down code and the tsconfig

kavilla (commented):
looks good just one comment if we could make this completely isolated to the query enhancements plugin

PR #7391, opened 2024-07-23
[Auto Suggest] DQL autosuggest with ANTLR

paulstn:
### Description

Implements ANTLR based autocomplete for DQL queries, currently within Discover.

Ignore files located in .generated and grammar/.antlr

Previously reviewed/approved in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7467 to be merged into the `feature/discover-2.0-1` branch. Was then reverted to isolate test failures, and then moved to this pr to get merged into main directly.

## Changelog

- feat: DQL Autocomplete

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/public/plugin.ts
```diff
@@ -168,6 +170,7 @@ export class DataPublicPlugin
 
     const ac = this.autocomplete.setup(core);
     ac.addQuerySuggestionProvider('SQL', getSQLSuggestions);
+    ac.addQuerySuggestionProvider('kuery', getDQLSuggestions);
```
sejli, line 173: Since the autocomplete service is initialized here, rather than create a getter/setter for `http`, we could just use it from the autocomplete service, since it gets set up with it.
```
const ac = this.autocomplete.setup(core);
```

sejli (approved):
Looks good, have some nits and one thing regarding `http`.

src/plugins/data/public/antlr/dql/code_completion.ts
```diff
@@ -0,0 +1,185 @@
...
+  indexPatterns,
+  position,
+}: QuerySuggestionGetFnArgs) => {
+  const currentIndexPattern = indexPatterns[0] as IndexPattern;
+
+  const inputStream = CharStream.fromString(query);
+  const lexer = new DQLLexer(inputStream);
+  const tokenStream = new CommonTokenStream(lexer);
+  const parser = new DQLParser(tokenStream);
+  parser.removeErrorListeners();
```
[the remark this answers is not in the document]
paulstn: This is only for the error listener on the parser, there's no behavior that needs the error listeners. Currently in the dql autocomplete, the parser is used to find the last field in the query when trying to find a value to suggest. Any errors found here would be about the current query not being grammatically correct, which doesn't happen to be useful.

src/plugins/data/public/antlr/dql/code_completion.ts
```diff
@@ -0,0 +1,185 @@
...
+import { DQLParserVisitor } from './.generated/DQLParserVisitor';
+
+const findCursorIndex = (
+  tokenStream: TokenStream,
+  cursor: monaco.Position,
+  whitespaceToken: number
+): number | undefined => {
+  const cursorCol = cursor.column - 1;
+
+  for (let i = 0; i < tokenStream.size; i++) {
```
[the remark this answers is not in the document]
paulstn: I think it might mess with the readability. This function essentially just walks through the tokens in tokenStream and checks the start and end column to see if its on the current token, then returns that iterator variable w/ modifications. Using `find` might be good for exactly how this function is now, but finding this cursor index could be complicated if something within the grammar changes, so personally, I think its better to have something that clearly walks through the tokens instead of a callback function.

src/plugins/data/public/antlr/dql/code_completion.test.ts
```diff
@@ -0,0 +1,281 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { monaco } from '@osd/monaco';
+import { getSuggestions } from './code_completion';
+import { IIndexPattern } from '../..';
+
+const testingIndex = ({
```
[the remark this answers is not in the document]
paulstn: thanks for the callout, done

src/plugins/data/public/plugin.ts
paulstn: Done. I pass the core setup along with the rest of the arguments whenever a query suggestion provider is called

src/plugins/data/public/antlr/dql/grammar/.antlr/DQLLexer.interp
```diff
@@ -0,0 +1,56 @@
+token literal names:
```
mengweieric, line 1: we should be able to remove these.

src/plugins/data/public/autocomplete/providers/query_suggestion_provider.ts
```diff
@@ -29,6 +29,7 @@
  */
 
 import { monaco } from 'packages/osd-monaco/target';
```
mengweieric, line 31: should use @osd/monaco

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -329,6 +336,8 @@ export default class QueryEditorUI extends Component<Props, State> {
         return monaco.languages.CompletionItemKind.Field;
       case 'value':
         return monaco.languages.CompletionItemKind.Value;
+      case 'keyword':
```
mengweieric, line 339: this is what i said, it is now returning monaco type directly with the new change and this is being removed. [PR](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7441)

src/plugins/data/public/autocomplete/providers/query_suggestion_provider.ts
paulstn: done

src/plugins/data/public/antlr/dql/grammar/.antlr/DQLLexer.interp
paulstn: will remove grammar/.antlr in fast follow

src/plugins/data/public/ui/query_editor/query_editor.tsx
paulstn: will updates completion item kind type in fast follow

sejli:
Could we close this one for #7467

sejli (approved):
Approving, more or less the same as a previously reviewed PR #7467.

sejli (dismissed):

src/plugins/data/public/antlr/dql/grammar/.antlr/DQLLexer.interp
mengweieric: Do we still want to remove these in this PR? should be a fast removal

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -264,47 +270,76 @@ export default class QueryEditorUI extends Component<Props, State> {
...
-  //             label: s.text,
-  //             kind: this.getCodeEditorSuggestionsType(s.type),
-  //             insertText: s.text,
-  //             range: wordRange,
-  //           }))
-  //         : [],
-  //     incomplete: false,
-  //   };
-  // };
+  private fetchIndexPatterns = async () => {
```
mengweieric, line 280: Synced with @paulstn offline on this, I believe we don't need this function and indexPatterns can refer to this.state.indexPatterns, i've done this in another PR and this should be a small changes

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -264,47 +270,76 @@ export default class QueryEditorUI extends Component<Props, State> {
...
+      fields: JSON.parse(savedObject.attributes?.fields),
+      type: SIMPLE_DATA_SET_TYPES.INDEX_PATTERN,
+    }));
+  };
+
+  provideCompletionItems = async (
+    model: monaco.editor.ITextModel,
+    position: monaco.Position
+  ): Promise<monaco.languages.CompletionList> => {
+    const enhancements = this.props.settings.getQueryEnhancements(this.props.query.language);
```
mengweieric, line 306: we don't need the enhancements do we?

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -264,47 +270,76 @@ export default class QueryEditorUI extends Component<Props, State> {
...
+    const connectionService = enhancements?.connectionService;
```
mengweieric, line 307: we don't need connectionService

src/plugins/data/public/ui/query_editor/query_editor.tsx
paulstn: Actually this fetchIndexPatterns happens to have the same name but its a new function, this one connects with the new dataset picker to get the most up to date index pattern

src/plugins/data/public/antlr/dql/grammar/.antlr/DQLLexer.interp
paulstn: since this change would be based on your pr, it would be best to handle it in a fast follow as to not remove actual functionality by accident. basically, i want keep changes separate so that there will be a working version before modifying again.

src/plugins/data/public/ui/query_editor/query_editor.tsx
paulstn: not sure, it was left over from early sql changes in `feature/discover-2.0-1`

src/plugins/data/public/ui/query_editor/query_editor.tsx
paulstn: this was also left over in early sql changes in `feature/discover-2.0-1`

src/plugins/data/public/ui/query_editor/query_editor.tsx
paulstn: made the change, removed it

src/plugins/data/public/ui/query_editor/query_editor.tsx
paulstn: removed

sejli (approved):

ashwin-pc:
ANTLR Grammer files are ignored in this PR: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7543/files

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -20,9 +20,14 @@ import { QueryEditorExtensions } from './query_editor_extensions';
...
+import { getQueryService } from '../../services';
+import { SIMPLE_DATA_SET_TYPES } from '../../../common';
 
-const LANGUAGE_ID = 'SQL';
-monaco.languages.register({ id: LANGUAGE_ID });
+const LANGUAGE_ID_SQL = 'SQL';
+monaco.languages.register({ id: LANGUAGE_ID_SQL });
+
+const LANGUAGE_ID_KUERY = 'kuery';
+monaco.languages.register({ id: LANGUAGE_ID_KUERY });
```
kavilla, line 30: nit: i dont think we should wait to register here. but nothing new for this PR.

fast follow or for 2.17 we refactor this all correctly and put it here:
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/5f19c37fadc98c05febc1010685280292fe58a0c/packages/osd-monaco/src/xjson/lexer_rules/index.ts#L38

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -264,47 +270,72 @@ export default class QueryEditorUI extends Component<Props, State> {
...
-  //         : [],
-  //     incomplete: false,
-  //   };
-  // };
+  private fetchIndexPatterns = async () => {
+    const client = this.services.savedObjects.client;
+    const dataSet = this.queryService.dataSetManager.getDataSet();
+    const title = dataSet?.title;
+
+    const resp = await client.find({
```
kavilla, line 285: i think we should use the index patterns service. because we are handling the caching there. we cache temporary index patterns (indices tables). this would hit savedobjects directly.

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
```diff
@@ -38,6 +38,14 @@ import { LIGHT_THEME, DARK_THEME } from './editor_theme';
 
 import './editor.scss';
 
+export interface LanguageSpecifiedConfiguration extends monaco.languages.LanguageConfiguration {
+  /**
+   * The language ID, meant to restrict the specified configuration for only this language. When
+   * not provided, will apply the language configuration for every language.
+   */
+  language?: string;
```
kavilla, line 46: what is this for? i think this should be a wrapper for the code editor but if there is logically differences then we add to osd-monaco

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
```diff
@@ -127,33 +135,6 @@ export class CodeEditor extends React.Component<Props, {}> {
       this.props.editorWillMount();
     }
 
-    monaco.languages.onLanguage(this.props.languageId, () => {
```
kavilla, line 130: this looks like existing code that technically anyone who is using the code_editor component will be getting rid of correct? without explicit opt in then these providers are gunna be set by default which could be a breaking change in some places.

kavilla:
do we know if the SQL implementation makes a request for fields per user input? what im seeing is this when i type a letter:

Screenshot 2024-07-29 at 10 22 21 PM

if so then we should put a debounce on getting suggestions.

kavilla:
do we have to wrap the values in quotes? for example:
Screenshot 2024-07-29 at 10 27 11 PM

kavilla:
have we verified with [field level security](https://opensearch.org/docs/latest/security/access-control/field-level-security/)? like values that can be sensitive in their nature. users who don't have permissions to see those fields won't see those fields in the autocomplete for values right?

kavilla (changes requested):
the feature is awesome and is great. great job.

just some concerns with security. could you provide some insight on those comments. if they have been researched and are not an issue then lemme know.

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
paulstn: its handled in the render() function. The reason I moved this there is because this method was within `_editorWillMount`, so all of this registration would only happen once. This resulted in the editor only registering the first language it sees, and if the language changes, none of these providers will be set for the new language. Functionality is exactly the same upon start.

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
paulstn: this was to allow for only specific languages to be specified for configurations. i can remove it

src/plugins/data/public/ui/query_editor/query_editor.tsx
paulstn: i'll try to refactor it for 2.17

paulstn:
> have we verified with [field level security](https://opensearch.org/docs/latest/security/access-control/field-level-security/)? like values that can be sensitive in their nature. users who don't have permissions to see those fields won't see those fields in the autocomplete for values right?
> 
> if i type values in the sensitive results I wont get back the hashed results right?
> 
> so i can't reverse engineer the actual value by guessing letters until i get a single result

The answer to this question is I'm not sure, but value suggestion directly leverages the same exactly functionality as filters. So, if this is insecure for autocomplete, it would be insecure for filters, which to my knowledge has been around for ~5 years now. It would be good to check, however, so I'll be following up on this for 2.17

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
paulstn: also another way of saying this is that this line:
```
monaco.languages.onLanguage(this.props.languageId, () => {
```
does nothing here

kavilla: have we verify no performance issues? that means anytime this is re-rendered its going to re register

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
kavilla: little bit confused, but you have more context than me. it seems like the only time this is set is in the loop where we check the language. and if we have to register the languages so we would still have to pass this prop. so as long as im registering the language then i can register the language configuration?

kavilla (commented):
qq in the PR

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -252,36 +258,64 @@ export default class QueryEditorUI extends Component<Props, State> {
     }
   };
 
+  private fetchIndexPatterns = debounce(
```
kavilla, line 261: yeah i believe index patterns service might be easier. it's ok can be fast follow

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
paulstn: i checked this when i made this change, it only triggered on every language switch

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
paulstn: so this interface is only used to specify which language the configuration should work for, its just an extra parameter that is checked within that loop to see if it should register the language configuration, or not

ashwin-pc (approved):

kavilla (dismissed):

ashwin-pc (approved):

ananzh (approved):

kavilla (approved):

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
SuZhou-Joe: We'd better move the code to like `componentDidUpdate` and detect if there are changes on languageId.

PR #7456, opened 2024-07-24
[2.16 MDS]Fix access control tab fetching for remote cluster

RyanL1997:
### Description
This PR will be forward port into both main and 2.x
Fix access control tab fetching for remote cluster

### Issues Resolved

- correct the behavior of fetching remote cluster roles instead of local for direct query datasource 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source_management/public/components/direct_query_data_sources_components/connection_detail/access_control_tab.tsx
+++ b/src/plugins/data_source_management/public/components/direct_query_data_sources_components/connection_detail/access_control_tab.tsx
@@ -25,6 +25,7 @@ interface AccessControlTabProps {
   connector: string;
   properties: unknown;
   allowedRoles: string[];
+  dataSourceMDSId: string;
 }
 
 export const AccessControlTab = (props: AccessControlTabProps) => {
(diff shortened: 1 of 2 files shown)


mengweieric (approved):

src/plugins/data_source_management/public/components/direct_query_data_sources_components/connection_detail/access_control_tab.tsx
```diff
@@ -34,17 +35,17 @@ export const AccessControlTab = (props: AccessControlTabProps) => {
   const { http } = useOpenSearchDashboards<DataSourceManagementContext>().services;
 
   useEffect(() => {
-    http!
-      .get(SECURITY_ROLES)
-      .then((data) =>
+    http
+      .get(SECURITY_ROLES, { query: { dataSourceId: props.dataSourceMDSId } })
```
derek-ho, line 39: There should ideally be some constant that core exposes that plugins can pull from, along same idea as https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6534. That way, this is not a hard coded value that can be changed/broken in the future

derek-ho (approved):

src/plugins/data_source_management/public/components/direct_query_data_sources_components/connection_detail/access_control_tab.tsx
RyanL1997: Yes, I was thinking the same and we should do this as a followup

BionIT (approved):

PR #7426, opened 2024-07-24
[Discover-next] data set picker

kavilla:
### Description

Thank you @ashwin-pc @sejli for the help on this one.

Stable dataset picker for indices and index patterns.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #7464, opened 2024-07-24
fix(query assist): update reading data source id from dataset manager

joshuali925:
### Description

- update the query assist logic for retrieving data source id through connection service to dataset manager, as connection service is removed in #7426 
- add dataset mock to query mock in data plugin

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/query_enhancements/public/query_assist/utils/create_extension.tsx
```diff
@@ -6,29 +6,30 @@
...
 const getAvailableLanguages$ = (
   availableLanguagesByDataSource: Map<string | undefined, string[]>,
-  connectionsService: ConnectionsService,
-  http: HttpSetup
+  http: HttpSetup,
+  data: DataPublicPluginSetup
 ) =>
-  connectionsService.getSelectedConnection$().pipe(
+  data.query.dataSet.getUpdates$().pipe(
+    startWith(data.query.dataSet.getDataSet()),
```
joshuali925, line 29: @kavilla why is there a skip 1 in dataset manager? i had to add a startWith to make it work
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/adbdd9501846cb4062697c4dabc79330eb1ddd9e/src/plugins/data/public/query/dataset_manager/dataset_manager.ts#L29

src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
```diff
@@ -37,19 +34,9 @@ export const QueryAssistBar: React.FC<QueryAssistInputProps> = (props) => {
   const { generateQuery, loading } = useGenerateQuery();
   const [callOutType, setCallOutType] = useState<QueryAssistCallOutType>();
   const dismissCallout = () => setCallOutType(undefined);
-  const [selectedIndex, setSelectedIndex] = useState<string>('');
-  const dataSourceIdRef = useRef<string>();
+  const selectedIndex = services.data.query.dataSet.getDataSet()?.title;
```
ashwin-pc, line 37: Will the component re-render if the dataset changes? If not you should subscribe to the dataset. use the useDataSetManager hook to get a stateful dataset object

src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
```diff
@@ -67,7 +54,7 @@ export const QueryAssistBar: React.FC<QueryAssistInputProps> = (props) => {
       question: inputRef.current.value,
       index: selectedIndex,
       language: props.dependencies.language,
-      dataSourceId: dataSourceIdRef.current,
+      dataSourceId: services.data.query.dataSet.getDataSet()?.dataSourceRef?.id,
```
ashwin-pc, line 57: same here. If you use the stateful dataset here, you can avoid fetchign the index again and causing a potential drift between the selected dataset and when this function is called

ashwin-pc (commented):
The rest looks good, ust one comment on how to use use the dataser

src/plugins/query_enhancements/public/query_assist/utils/create_extension.tsx
ashwin-pc: He copied it over from the query manager. Yeah i dont think it needs to be there either. Feel free to remove it. We need the value of the dataset immediately

src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
joshuali925: i agree. `dataSourceIdRef` previously was a ref and we never relied on re-rendering triggered by subscription, so current implementation shouldn't cause issues. But the re-rendering is probably a side effect and not good to rely on.

i wanted to subscribe but i didn't understand this part https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7464#discussion_r1690533991

it doesn't work unless i add a default initial value

src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
joshuali925: > avoid fetchign the index again and causing a potential drift between the selected dataset and when this function is called

didn't understand this part, when the function is called `services.data.query.dataSet.getDataSet()` returns the latest selected dataSet. but anyways i'll add a stateful dataset and this part can share the value

src/plugins/query_enhancements/public/query_assist/utils/create_extension.tsx
kavilla: +1

joshuali925: thanks for confirming, i removed it

kavilla (approved):

sejli (approved):

PR #7484, opened 2024-07-25
[Workspace] refactor saved object client addToWorkspaces and deleteFromWorkspaces function params structure

raintygao:
### Description

Refactor saved object client addToWorkspaces and deleteFromWorkspaces function params structure to make it extendable.

## Testing the changes

Go to workspace create/detail to assign and unassign data source, all would be as expected.

## Changelog

- refactor: Update saved object client addToWorkspaces and deleteFromWorkspaces function params structure

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/server/saved_objects/service/saved_objects_client.ts
+++ b/src/core/server/saved_objects/service/saved_objects_client.ts
@@ -473,12 +473,14 @@ export class SavedObjectsClient {
   deleteFromWorkspaces = async <T = unknown>(
     type: string,
     id: string,
-    options: { workspaces: string[] }
+    options: { workspaces?: string[] }
   ) => {
     const { workspaces } = options;
     if (!workspaces || workspaces.length === 0) {
       throw new TypeError(`Workspaces is required.`);
     }
+    // There may be some customized wrappers which will extend the options etc, so we will delete workspace field and pass options continuously.
+    delete options.workspaces;
     const object = await this.get<T>(type, id, options);
     const existingWorkspaces = object.workspaces ?? [];
     const newWorkspaces = existingWorkspaces.filter((item) => {
@@ -518,12 +520,14 @@ export class SavedObjectsClient {
   addToWorkspaces = async <T = unknown>(
     type: string,
     id: string,
-    options: { workspaces: string[] }
+    options: { workspaces?: string[] }
   ): Promise<any> => {
     const { workspaces } = options;
     if (!workspaces || workspaces.length === 0) {
       throw new TypeError(`Workspaces is required.`);
     }
+    // There may be some customized wrappers which will extend the options etc, so we will delete workspace field and pass options continuously.
+    delete options.workspaces;
     const object = await this.get<T>(type, id, options);
     const existingWorkspaces = object.workspaces ?? [];
     const mergedWorkspaces = existingWorkspaces.concat(workspaces);
--- a/src/core/server/saved_objects/service/saved_objects_client.ts
+++ b/src/core/server/saved_objects/service/saved_objects_client.ts
@@ -468,19 +468,25 @@ export class SavedObjectsClient {
    * Remove a saved object from workspaces
    * @param type
    * @param id
-   * @param workspaces
+   * @param options
    */
-  deleteFromWorkspaces = async <T = unknown>(type: string, id: string, workspaces: string[]) => {
+  deleteFromWorkspaces = async <T = unknown>(
+    type: string,
+    id: string,
+    options: { workspaces: string[] }
+  ) => {
+    const { workspaces } = options;
     if (!workspaces || workspaces.length === 0) {
       throw new TypeError(`Workspaces is required.`);
     }
-    const object = await this.get<T>(type, id);
+    const object = await this.get<T>(type, id, options);
     const existingWorkspaces = object.workspaces ?? [];
     const newWorkspaces = existingWorkspaces.filter((item) => {
       return workspaces.indexOf(item) === -1;
     });
     if (newWorkspaces.length > 0) {
       return await this.update<T>(type, id, object.attributes, {
+        ...options,
         workspaces: newWorkspaces,
         version: object.version,
       });
@@ -492,6 +498,8 @@ export class SavedObjectsClient {
           ...object.attributes,
         },
         {
+          ...options,
+          workspaces: [],
           id,
           permissions: object.permissions,
           overwrite: true,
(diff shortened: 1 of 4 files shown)


src/core/server/saved_objects/service/saved_objects_client.ts
```diff
@@ -492,6 +498,8 @@ export class SavedObjectsClient {
           ...object.attributes,
         },
         {
+          ...options,
+          workspaces: [],
```
SuZhou-Joe, line 502: Do we need this line?

raintygao: Yes, options contain a workspace field with some workspaces, otherwise the delete workspace field would fail.

src/core/server/saved_objects/service/saved_objects_client.ts
```diff
@@ -468,19 +468,27 @@ export class SavedObjectsClient {
...
+    type: string,
+    id: string,
+    options: { workspaces?: string[] }
+  ) => {
+    const { workspaces } = options;
     if (!workspaces || workspaces.length === 0) {
       throw new TypeError(`Workspaces is required.`);
     }
-    const object = await this.get<T>(type, id);
+    // There may be some customized wrappers which will extend the options etc, so we will delete workspace field and pass options continuously.
```
SuZhou-Joe, line 482:
```suggestion
    // There may be some customized wrappers which will extend the options, 
    // delete workspace field and pas options continuously for extensibility.
```

src/core/server/saved_objects/service/saved_objects_client.ts
SuZhou-Joe: This line is not needed as we've already delete the workspaces field?

src/core/server/saved_objects/service/saved_objects_client.ts
```diff
@@ -505,22 +515,26 @@ export class SavedObjectsClient {
...
     id: string,
-    workspaces: string[]
+    options: { workspaces?: string[] }
   ): Promise<any> => {
+    const { workspaces } = options;
     if (!workspaces || workspaces.length === 0) {
       throw new TypeError(`Workspaces is required.`);
     }
-    const object = await this.get<T>(type, id);
+    // There may be some customized wrappers which will extend the options etc, so we will delete workspace field and pass options continuously.
```
SuZhou-Joe, line 529:
```suggestion
    // There may be some customized wrappers which will extend the options,  so we will delete workspace field and pass options continuously for extensibility.
```

src/core/server/saved_objects/service/saved_objects_client.ts
raintygao: Yes, I just want to make it clear.

raintygao: I can also delete this line.

raintygao: Deleted.

SuZhou-Joe (dismissed):

src/core/server/saved_objects/service/saved_objects_client.ts
```diff
@@ -468,19 +468,27 @@ export class SavedObjectsClient {
...
    * @param type
    * @param id
-   * @param workspaces
+   * @param options
    */
-  deleteFromWorkspaces = async <T = unknown>(type: string, id: string, workspaces: string[]) => {
+  deleteFromWorkspaces = async <T = unknown>(
+    type: string,
+    id: string,
+    options: { workspaces?: string[] }
```
ruanyl, line 476: Then I think the type of `options` should be extending `SavedObjectsBaseOptions`?

raintygao: Updated.

src/core/server/saved_objects/service/saved_objects_client.ts
```diff
@@ -505,22 +514,26 @@ export class SavedObjectsClient {
...
    * @param type
    * @param id
-   * @param workspaces
+   * @param options
    */
   addToWorkspaces = async <T = unknown>(
     type: string,
     id: string,
-    workspaces: string[]
+    options: SavedObjectsBaseOptions
```
ruanyl, line 522: Discussed with @SuZhou-Joe, I may suggest to make the interface like this
```typescript
  deleteFromWorkspaces = async <T = unknown>(
    type: string,
    id: string,
    workspaces: string[],
    options: SavedObjectsBaseOptions = {}
  ) => {
```

SuZhou-Joe: Nit: rename the workspaces: string[] to targetWorkspaces: string[] to reduce the confusion of duplicate workspaces params.

raintygao: Update.

src/core/server/saved_objects/service/saved_objects_client.ts
```diff
@@ -492,6 +499,8 @@ export class SavedObjectsClient {
           ...object.attributes,
         },
         {
+          ...options,
+          workspaces: [],
```
ruanyl, line 503: `workspaces: [],` this is not needed because `options.workspaces` won't exist?

raintygao: Yes. From the type definition, options could also contain workspace field. I'm intended to add this because this will ensure the empty workspace to make the function completed and make it looking clear.

SuZhou-Joe: Can we try to get the rest of params within options without workspaces? There should not have any objects whose workspaces field is empty array.

raintygao: What unexpected errors will be led to? I suppose empty workspaces array just represent that this object doesn't belong to any workspaces and I think we should not parse the options any more if we add a new field.

raintygao: If empty workspace array would be not reliable, I think we can also pass undefined to redeclare. This could ensure rewrite raw option and would be set value by repository.

ruanyl: @raintygao You're right, user could set `options.workspaces`. It's safe to make sure we don't send `workspaces` to `this.create`
What about doing this:
```typescript
  deleteFromWorkspaces = async <T = unknown>(
    type: string,
    id: string,
    workspaces: string[],
    options: Omit<SavedObjectsBaseOptions, 'workspaces'> = {}
  ) => {
    if (!workspaces || workspaces.length === 0) {
      throw new TypeError(`Workspaces is required.`);
    }
    if ('workspaces' in options && options.workspaces) {
      throw new TypeError('Invalid options, options.workspaces should not exist.');
    }
```

raintygao: Sure, updated.

ruanyl (approved):

SuZhou-Joe (approved):

src/core/server/saved_objects/service/saved_objects_client.ts
SuZhou-Joe: I remember we use create with overwrite when updating objects with empty workspaces to prevent objects with `workspaces: []` from creating. I could not remember in which case the listing logic may find unexpected objects but we'd better keep the logic as it was before the change right? Or we do not need to use create with overwrite, just update with workspaces: [].

raintygao: Yes, I remember that update object with empty workspace would have some unexpected behaivor. Overwrite create with workspace: undefined would work well, but we can also throw error when options contain workspace field like what we have built for now.

SuZhou-Joe: Yes, what @ruanyl  proposed should work as well, thanks for making it consistent.

Issue #7280, opened 2024-07-17
[BUG] [Workspace] workspace name missing validation on blank input


Hailong-am:
**Describe the bug**

when we create a new workspace, the validation rule for workspace name is to check it don't have special character and only allow `a-z, A-Z, 0-9, (), [], _ (underscore), - (hyphen) and (space)`.
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/d2e7b9d41cf1725b49eb5826c0673d2bb1609d24/src/plugins/workspace/public/components/workspace_form/utils.ts#L39-L46 

But for blank input as workspace name, it can pass the validation and created successfully.

workspace with blank name will be not clickable on workspace list page and hard to tell what it is in workspace switcher.

![image](https://github.com/user-attachments/assets/f1d30b41-a3b7-4e4a-a76a-b62c3f016f5a)

![image](https://github.com/user-attachments/assets/d04328f3-8808-45f0-901a-e2e77adea5cc)

**To Reproduce**

**Expected behavior**

workspace name with blank should be blocked for creation and updation

**OpenSearch Version**
Please list the version of OpenSearch being used.

**Dashboards Version**
Please list the version of OpenSearch Dashboards being used.

**Plugins**

Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

ananzh:
@Hailong-am I will remove the untriaged label for now.

SuZhou-Joe:
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7512

PR #7512, opened 2024-07-26
[Worksapce]Add blank check for workspace name

Hailong-am:
### Description

Add blank or empty check for workspace name

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/7280

## Screenshot

when input a blank workspace name, validation will fail

image

## Changelog

- fix: [Workspace]add workspace name blank/empty check

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


wanglam (approved):
LGTM

src/plugins/workspace/server/routes/index.ts
```diff
@@ -46,13 +46,21 @@ const workspaceOptionalAttributesSchema = {
   reserved: schema.maybe(schema.boolean()),
 };
 
+const workspaceNameSchema = schema.string({
+  validate(value) {
+    if (!value || value.trim().length === 0) {
+      return "can't be empty or blank.";
```
SuZhou-Joe, line 52: Not sure if the description is clear enough for customers to understand what happens here. And should we wrap it with i18n?

Hailong-am: field name will prepend to actual message, the whole error message will be
```json
{
 "statusCode": 400,
 "error": "Bad Request",
 "message": "[request body.attributes.name]: can't be empty or blank."
}
```

>And should we wrap it with i18n?

as the default validation message of [schema.string()](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/packages/osd-config-schema/src/types/string_type.ts#L59)haven't been wrapped by i18n. i think we can keep it same

ruanyl (approved):

SuZhou-Joe (approved):

PR #7516, opened 2024-07-26
[Workspace]Optimize workspace permission validate for bulk

wanglam:
### Description

This PR is for optimizing saved objects's workspace permission validation logic for bulk operations. It mainly includes below three changes:
1. Add saved objects cache related methods to permission control client to enable saved objects cache.
2. Cache workspace saved objects when executing saved object client bulk create, bulk update and bulk get.
3. Clear saved objects cache when saved objects cache active before repsonse

## Changelog

- feat: [Workspace]Optimize workspace permission validation for bulk operations

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/workspace/server/permission_control/client.ts
```diff
@@ -175,4 +215,31 @@ export class SavedObjectsPermissionControl {
...
+    );
+  }
+
+  public clearSavedObjectsCache(request: OpenSearchDashboardsRequest) {
+    const requestKey = ensureRawRequest(request);
+    this._shouldCachedSavedObjects.delete(requestKey);
+    this._savedObjectCache.delete(requestKey);
+  }
+
+  public isSavedObjectsCacheActive(request: OpenSearchDashboardsRequest) {
```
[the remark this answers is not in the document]
wanglam: Sure, since this delete operation will be fired in every request response. Not sure if it too expensive to do the delete.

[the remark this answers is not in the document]
ruanyl: We can encapsulate the check in `clearSavedObjectsCache`

src/plugins/workspace/server/permission_control/client.ts
```diff
@@ -30,6 +33,8 @@ export class SavedObjectsPermissionControl {
   private readonly logger: Logger;
   private _getScopedClient?: SavedObjectsServiceStart['getScopedClient'];
   private auth?: HttpAuth;
+  private _savedObjectCache: WeakMap<LegacyRequest, { [key: string]: SavedObject }> = new WeakMap();
```
SuZhou-Joe, line 36: What about we making it `WeakMap<string(request.id), { [key: string]: SavedObject }>`? By doing so we do not need to import ensureRawRequest and LegacyRequest from core.

src/plugins/workspace/server/permission_control/client.ts
```diff
@@ -48,11 +53,46 @@ export class SavedObjectsPermissionControl {
...
+    );
+    const retrievedSavedObjects =
+      savedObjectsToGet.length > 0
+        ? (await this.getScopedClient?.(request)?.bulkGet(savedObjectsToGet))?.saved_objects || []
+        : [];
+
+    const retrievedSavedObjectsMap: { [key: string]: SavedObject } = {};
+    retrievedSavedObjects.forEach((savedObject) => {
+      const savedObjectKey = this.generateSavedObjectKey(savedObject);
+      if (this._shouldCachedSavedObjects.get(requestKey)?.includes(savedObjectKey)) {
```
SuZhou-Joe, line 77: Seems we can detect if there is cache already by looking up directly in the `_savedObjectCache` and making it the source of truth.

src/plugins/workspace/server/permission_control/client.ts
```diff
@@ -175,4 +215,30 @@ export class SavedObjectsPermissionControl {
       result: hasPermissionToAllObjects,
     };
   }
+
+  public cacheSavedObjects(
```
SuZhou-Joe, line 219: Actually it is not doing `cache`, it is adding a list of objects that are awaiting for cache. Maybe rename to like `addSavedObjectsToCachableList`?

src/plugins/workspace/server/permission_control/client.ts
```diff
@@ -48,11 +53,46 @@ export class SavedObjectsPermissionControl {
     this.logger = logger;
   }
 
+  private generateSavedObjectKey = ({ type, id }: { type: string; id: string }) => {
+    return `${type}${id}`;
```
SuZhou-Joe, line 57:
```suggestion
    return `${type}:${id}`;
```
Add a separator would be nice to avoid collision.

src/plugins/workspace/server/permission_control/client.ts
wanglam: We need to use Map instead of WeakMap if we want to use `request.id` as map key. Key of WeakMap should be an object.

SuZhou-Joe: Then what about using Map? I just do not think it is necessary to import the internal `ensureRawRequest`.

wanglam: In my point of view, `WeakMap` has better memory performance on large requests. There are many request level caches are using `WeakMap`. 
Here are the examples: 
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/61d5c4905bce1709d29629ade132723d87fb6897/src/core/server/http/base_path_service.ts#L41
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/61d5c4905bce1709d29629ade132723d87fb6897/src/core/server/http/auth_headers_storage.ts#L46
Some [external plugins](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/61d5c4905bce1709d29629ade132723d87fb6897/src/plugins/console/server/routes/api/console/proxy/create_handler.ts#L38) are using `ensureRawRequest` too. I'm open to change to `Map` and using `request.uuid` as the map key. What do you think about it?

wanglam: Update with `request.uuid` instead of `request.id`, the `request.id` is not guaranteed to be unique per request. See below comments:
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/http/router/request.ts#L165

SuZhou-Joe (approved):

ruanyl (approved):

PR #7543, opened 2024-07-28
[Auto Suggest] Add back reverted MDS support, tests and cleanup for sql auto suggest

mengweieric:
### Description

- Cherry-pick MDS support for sql autosuggest
- Tests update
- Minor cleanup

## Changelog
- feat: [Auto Suggest] Add MDS support, tests and cleanup for sql auto suggest

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


.gitignore
```diff
@@ -68,3 +68,6 @@ snapshots.js
 
 # Yarn local mirror content
 .yarn-local-mirror
+
+# Ignore the generated antlr files
+/src/plugins/data/public/antlr/opensearch_sql/grammar/.antlr
```
kavilla, line 73: i see we are git ignoring this.

do we have a script that ends up generating the rules? or how do we get these values now?

src/plugins/data/public/ui/query_editor/editors/default_editor/index.tsx
```diff
@@ -5,7 +5,7 @@
 
 import React from 'react';
 import { EuiText, EuiTextColorProps } from '@elastic/eui';
-// import { monaco } from '@osd/monaco';
+import { monaco } from '@osd/monaco';
```
kavilla, line 8: how come we did this?

@osd/monaco is the wrapper around the monaco editor. and the CodeEditor is the component that wires up the monaco editor. and then everyone should interact with the CodeEditor component. This way, if someone in another plugin wants to use the autocomplete and the code editor but don't want to use the data plugin UI then they can still have all that ready to go. but this would require developers to do the same code we have here.

we should refactor to ensure the CodeEditor is the source of truth.

kavilla:
Some questions that if you have some insight could unblock/give peace of mind.

* does autocomplete values display sensitive info and filter results if I guess the letters until I get a single value [[question in DQL](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7391#issuecomment-2257497558)]
* does SQL fire a request per key stroke? if so we should debounce [[question in DQL](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7391#issuecomment-2257490721)]

.gitignore
mengweieric: This is neither ignoring grammar files or generated lexer and parser files, this is for ignoring some intermediate local, auto generated files that are not needed for autocomplete logics or file reference, the entire .antlr folder is generated when you have changes to your grammar for development work. The grammar is located at `/src/plugins/data/public/antlr/opensearch_sql/grammar/` and generated lexer and parser are located at `/src/plugins/data/public/antlr/opensearch_sql/.generated`, the command to generate lexer and parser are located in package.json and they are already generated that code has reference to.

src/plugins/data/public/ui/query_editor/editors/default_editor/index.tsx
mengweieric: Not sure if I understand the suggestion. The monaco import here is only for specifying the correct type. If you take a look at few lines below
```
provideCompletionItems: monaco.languages.CompletionItemProvider['provideCompletionItems'];
``` 
that this is the only usage. The autocomplete itself is only hooked in on CodeEditor. What would be the suggestion if this is not what you are looking at?

mengweieric:
> Some questions that if you have some insight could unblock/give peace of mind.
> 
> * does autocomplete values display sensitive info and filter results if I guess the letters until I get a single value [[question in DQL](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7391#issuecomment-2257497558)]
> * does SQL fire a request per key stroke? if so we should debounce [[question in DQL](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7391#issuecomment-2257490721)]

Thanks for review.

This SQL implementation does not contain values suggestion as it is not a common practice to suggest values with a few concerns including the security issue you mentioned.

As respect to your second question if there's field match then no additional request is sent, but ideally the plan was to also leverage both index patterns and cache layer together maybe so that if there's no dataset changes the schema is only fetched once for all the following field suggestion. I believe as queryEnhancement and data plugin with MQL features are getting into a good shape, we should be able to leverage more core APIs from these plugins to fetch schema and manage request in a more optimal way.

ashwin-pc (approved):
Approving given the response.

changelogs/fragments/7463.yml
```diff
@@ -0,0 +1,2 @@
+feat:
+- Add MDS support along with a few cleanup and tests update ([#7463](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7463))
```
joshuali925, line 2: wrong PR number?

src/plugins/data/public/antlr/opensearch_sql/code_completion.ts
```diff
@@ -47,80 +48,59 @@ export const getSuggestions = async ({
...
-
-      schemas.forEach((schema) => {
-        if (schema.body?.fields?.length > 0) {
-          const columns = schema.body.fields.find((col: any) => col.name === 'COLUMN_NAME');
-          const fieldTypes = schema.body.fields.find((col: any) => col.name === 'DATA_TYPE');
+    if (suggestions.suggestColumns?.tables?.length) {
+      const tableNames = suggestions.suggestColumns.tables.map((table) => table.name);
+      const schemas = await fetchTableSchemas(tableNames, api, dataSetManager);
+
+      (schemas as IDataFrameResponse[]).forEach((schema: IDataFrameResponse) => {
```
joshuali925, line 65: nit. it would be better to define return types for `fetchTableSchemas` instead of assert

src/plugins/data/public/antlr/opensearch_sql/code_completion.ts
```diff
@@ -47,80 +48,59 @@ export const getSuggestions = async ({
...
           text: sk.value,
-          type: 'keyword',
+          type: monaco.languages.CompletionItemKind.Keyword,
         }))
       );
     }
   } catch (error) {
-    // TODO: pipe error to the UI
+    // TODO: Handle errors appropriately, possibly logging or displaying a message to the user
+    return [];
```
joshuali925, line 103: is this catching errors for `await fetchTableSchemas` or something else could also throw error? if the network call fails, do we want to show other suggestions like `suggestAggregateFunctions` if possible?

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -4,81 +4,85 @@
...
-        return from(defaultReuqstHandler());
+    switchMap((dataSet) => {
+      if (!dataSet) {
+        return from(defaultRequestHandler());
       }
-      const dataSourceId = connection?.dataSource?.id;
-      const title = connection?.attributes?.title;
-      return from(dataSourceReuqstHandler({ dataSourceId, title }));
+      const dataSourceId = dataSet?.dataSourceRef?.id;
+      const title = dataSet?.dataSourceRef?.name;
```
joshuali925, line 32:
```suggestion
      const dataSourceId = dataSet.dataSourceRef?.id;
      const title = dataSet.dataSourceRef?.name;
```

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -4,81 +4,85 @@
...
+      return from(dataSourceRequestHandler({ dataSourceId, title }));
```
joshuali925, line 33: if dataSourceId and title are undefined, would this fail? if not then make `IDataSourceRequestHandlerParams` to define them as optional string?

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -4,81 +4,85 @@
...
     })
   );
 
+const fetchFromAPI = async (api: any, body: string) => {
+  try {
+    return await api.http.fetch({
```
joshuali925, line 39: what is `api` and why not just return api.http.fetch without try catch?

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -4,81 +4,85 @@
...
+        const requests = tables.map(async (table) => {
+          const body = JSON.stringify(queryFormatter(table));
+          return fetchFromAPI(api, body);
+        });
+        return Promise.all(requests);
       }
     ).subscribe({
       next: (dataFrames: any) => resolve(dataFrames),
-      error: (err: any) => {
+      error: (err: Error) => {
```
joshuali925, line 76: seems like this resolves and rejects immediately and only once, do you need to make getRawSuggestionData an observable? and it would be better to avoid `any` for easier maintenance

joshuali925 (approved):

changelogs/fragments/7463.yml
mengweieric: this commit is cherry picking previous commit after revert, therefore it contains this as part of the pervious commit, the new changelog was added in the description and auto generated

src/plugins/data/public/antlr/opensearch_sql/code_completion.ts
mengweieric: Yea we could but the only concern here is if fetching the schema is having error, but user is still getting other suggestions, would this error being potentially hidden in this case that user still thinks everything is working accurately?

src/plugins/data/public/antlr/shared/utils.ts
mengweieric: yes, we should make `IDataSourceRequestHandlerParams` optional string

src/plugins/data/public/antlr/opensearch_sql/code_completion.ts
joshuali925: i would think that there are different completion sources, one source failing shouldn't affect others. the TODO `Handle errors appropriately, possibly logging or displaying a message to the user` should help to notify user about any failure

PR #7593, opened 2024-07-31
[Auto Suggest] DQL Updates

paulstn:
### Description

Update to DQL Autocomplete: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7391

Every file removal under `grammar/.antlr` can be ignored.

- [x] use official value suggestion methods instead of direct api call
- [x] removed language specified configuration
- [ ]  ~~added memoization for value suggestion to reduce number of calls made~~
- [x] removed core start from query suggestion function
- [x] added tests for value suggestion
- [x] added more test coverage for other general cases
- [ ] ~~[[RFC] Monaco Code Editor provider registration #7594](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/7594) made changes based on this RFC~~
- [x] remove grammar/.antlr auto generated files
- [x] updated types in code completion and related files
- [x] fixed many group value suggestion bugs and edge cases with more robust parser visitor
- [x] fix group value NOT suggestion bugs
- [ ] ~~added basic keyword syntax highlighting~~

## Changelog

- fix: Update DQL Autocomplete in code and functionality

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/public/antlr/dql/code_completion.ts
```diff
@@ -174,21 +142,123 @@ export const getSuggestions = async ({
...
     if (candidates.rules.has(DQLParser.RULE_field)) {
       completions.push(...findFieldSuggestions(indexPattern));
     }
 
+    interface FoundLastValue {
+      field: string | undefined;
+      value: string | undefined;
+    }
+
+    // visitor for parsing the current query
```
paulstn, line 162: The purpose of this 'visitor' here is to traverse the AbstractSyntaxTree generated by DQLParser and return what it thinks is the associated `field` and associated `value`, if there is one that is partially complete. 

`defaultResult` and `aggregateResult` define the default behavior at each node of what to do with the return values from visiting a particular node's children.

The other methods, in the format `visit[RULE_NAME]` override the default rule visitors in order to define custom behavior.

The starting point of the custom rule visitor functions is `KeyValueExpression`, which will match the parts of the dql query that have the format `field: value`, which is the only instance in which we want to provide value suggestions. The value here can be as simple as a 'tokenSearch' or a 'phrase', from which we can easily find the current value if there is one. However, it can also be a `groupExpression`, for which we have to define more custom behavior on those rule visitors. The general strategy is to enter and return nodes that contain the cursor position. 

Ultimately, the found field + value combinations are returned up the tree, giving deference to combinations found on the right side since more recent query input is on that side.

src/plugins/data/public/antlr/dql/code_completion.test.ts
```diff
@@ -28,82 +42,234 @@ const getSuggestionsAtPos = async (query: string, endPos: number) => {
...
+    } as IDataPluginServices,
   });
 };
 
-const getSuggestionAtEnd = async (query: string) => {
+const getSuggestionsAtEnd = async (query: string) => {
   return await getSuggestionsAtPos(query, query.length + 1);
 };
 
+const testAroundClosing = (
```
paulstn, line 58: This function contains 2-4 tests that can be triggered around a certain query. These were common patterns found in queries that needed to be tested, basically the behavior around closed/unclosed quotes and parenthesis

src/plugins/data/public/antlr/dql/code_completion.ts
```diff
@@ -45,38 +48,26 @@ const findFieldSuggestions = (indexPattern: IndexPattern) => {
...
-  const fieldSuggestions: Array<{
-    text: string;
-    type: monaco.languages.CompletionItemKind;
-  }> = fieldNames.map((field: string) => {
+  const fieldSuggestions: QuerySuggestion[] = fieldNames.map((field: string) => {
     return {
       text: field,
       type: monaco.languages.CompletionItemKind.Field,
       insertText: `${field}: `,
+      start: -1,
```
paulstn, line 56: These `start/end` parameters will be handled in an upcoming pr

src/plugins/data/public/antlr/shared/constants.ts
```diff
@@ -167,26 +167,67 @@ export const testingIndex = ({
...
+  { text: 'DestWeather', type: 3, insertText: 'DestWeather: ', end: -1, start: -1 },
+  { text: 'DistanceMiles', type: 3, insertText: 'DistanceMiles: ', end: -1, start: -1 },
+  { text: 'FlightDelay', type: 3, insertText: 'FlightDelay: ', end: -1, start: -1 },
+  { text: 'FlightNum', type: 3, insertText: 'FlightNum: ', end: -1, start: -1 },
+  { text: 'OriginWeather', type: 3, insertText: 'OriginWeather: ', end: -1, start: -1 },
+  { text: '_id', type: 3, insertText: '_id: ', end: -1, start: -1 },
+  { text: '_index', type: 3, insertText: '_index: ', end: -1, start: -1 },
+  { text: '_score', type: 3, insertText: '_score: ', end: -1, start: -1 },
+  { text: '_source', type: 3, insertText: '_source: ', end: -1, start: -1 },
+  { text: '_type', type: 3, insertText: '_type: ', end: -1, start: -1 },
```
joshuali925, line 195: should constants used for test be in a *.test.ts file?

joshuali925 (approved):

src/plugins/data/public/antlr/shared/constants.ts
paulstn: My assumption was that since the file didn't contain any tests, it shouldn't have that extension. I couldn't find any precedence for constants being used exclusively for tests, do you know if there is another way of handling these?

kavilla:
seeing some failures in the ci.

also qq, i know @mengweieric was working on a feature branch for his updates. does these updates have those and account for changes related to that work. since it has clean ups?

src/plugins/data/public/antlr/shared/constants.ts
```diff
@@ -167,26 +167,67 @@ export const testingIndex = ({
...
-  { text: 'DistanceMiles', type: 3, insertText: 'DistanceMiles: ' },
-  { text: 'FlightDelay', type: 3, insertText: 'FlightDelay: ' },
-  { text: 'FlightNum', type: 3, insertText: 'FlightNum: ' },
-  { text: 'OriginWeather', type: 3, insertText: 'OriginWeather: ' },
-  { text: '_id', type: 3, insertText: '_id: ' },
-  { text: '_index', type: 3, insertText: '_index: ' },
-  { text: '_score', type: 3, insertText: '_score: ' },
-  { text: '_source', type: 3, insertText: '_source: ' },
-  { text: '_type', type: 3, insertText: '_type: ' },
+export const fieldNameSuggestions: Array<{
```
kavilla, line 176: we shouldn't store test values in constants file. constants will get bundled up and will increase the release distribution size. test files do not get compiled with the distribution.

src/plugins/data/public/antlr/shared/constants.ts
kavilla: ah @joshuali925 similar comment https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7593/files#r1732297731

you can use `mock` or just include this in the test.

exporting public fieldNameSuggestions could get confusing for other plugins since now any plugin can access it and this will get compiled down with the release distribution.

src/plugins/data/public/autocomplete/autocomplete_service.ts
```diff
@@ -63,7 +61,7 @@ export class AutocompleteService {
     const provider = this.querySuggestionProviders.get(language);
 
     if (provider) {
-      return provider({ core: this.core, ...args });
+      return provider(args);
```
kavilla, line 64: is this a breaking change?

paulstn: no, i'm reverting a change that i made previously with dql autocomplete. this was before i knew about a clear way to get to the http core util. but now since we're using services, its redundant and so i put it back the way it was.

paulstn:
> also qq, i know @mengweieric was working on a feature branch for his updates. does these updates have those and account for changes related to that work. since it has clean ups?

The updates here and the interface changes Eric made in the feature branch don't conflict with each other, since I've included the necessary changes to DQL Autocomplete that handle those changes within that feature branch.

src/plugins/data/public/antlr/shared/constants.ts
paulstn: that makes sense, i'll move everything here into those tests

src/plugins/data/public/antlr/shared/constants.ts
paulstn: out of curiosity, could you point me to where its specified which files don't get compiled in the distribution? would it just be .test.ts files, or are there other files that wouldn't be included?

paulstn:
> seeing some failures in the ci.

The cypress failures seem to be coming and going for different tests with each trivial, unrelated commit. I'm not sure what I should be doing here other than retry until all the stars align.

src/plugins/data/public/antlr/shared/constants.ts
joshuali925: there are some hard coded ignores in the repo, like https://github.com/opensearch-project/OpenSearch-Dashboards/blob/25c23db905e1b0f8a83229def41151a6e275f097/packages/osd-pm/src/utils/targeted_build.ts#L58-L59

joshuali925 (approved):

src/plugins/data/public/antlr/dql/code_completion.ts
```diff
@@ -45,38 +48,26 @@ const findFieldSuggestions = (indexPattern: IndexPattern) => {
...
-) => {
-  if (!http) return [];
-  return await http.fetch(`/api/opensearch-dashboards/suggestions/values/${indexTitle}`, {
-    method: 'POST',
-    body: JSON.stringify({ query: currentValue, field: fieldName, boolFilter: [] }),
-  });
-};
-
 const findValueSuggestions = async (
   index: IndexPattern,
```
kavilla, line 65: is this the indexpattern service or just object. because actually can just take the query object which now has the dataset object which has all the information you will need.

kavilla (dismissed):

src/plugins/data/public/antlr/dql/code_completion.ts
paulstn: This is the actual object, I believe its set by the query_editor.tsx function which comes from the dataset object.

kavilla (approved):

joshuali925 (approved):

PR #7596, opened 2024-07-31
[Workspace] add validation for data source in get and bulk_get methods

raintygao:
### Description

Add validation for data source in get and bulk get

## Screenshot

No UX updates

## Testing the changes

In workspace, when getting a data source that doesn't belong to this workspace, this request will be forbidden. Before we didn't do this validation.

## Changelog

- fix: add validation for data source in get and bulk_get methods

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
```diff
@@ -179,6 +189,22 @@ export class WorkspaceSavedObjectsClientWrapper {
     return hasPermission;
   }
 
+  // Data source is a workspace level object, validate if data source has the permission to the given `workspaceIds`
+  private validateDataSourcePermissions = (
+    object: SavedObject,
+    request: OpenSearchDashboardsRequest
+  ) => {
+    const requestWorkspaceId = getWorkspaceState(request).requestWorkspaceId;
+    if (
```
wanglam, line 198: Nit: Shall we rewrite to below code?
```
return !requestWorkspaceId || !!object.workspaces?.includes(requestWorkspaceId)
```
This looks more clearable to me.

wanglam (approved):

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
raintygao: Do you mean we use this code to replace ` requestWorkspaceId &&
      !(object.workspaces && object.workspaces?.includes(requestWorkspaceId))`? I suppose their effect would not be the same.

wanglam: Not replace the if condition. We can replace the whole if statement.
```ts
if (
      requestWorkspaceId &&
      !(object.workspaces && object.workspaces?.includes(requestWorkspaceId))
    ) {
      return false;
    } else {
      return true;
    }
```
can be refactor to
```ts
return !requestWorkspaceId || !!object.workspaces?.includes(requestWorkspaceId)
```
It has better readability to me. The method will return true when request workspace id is empty or object's workspaces attributes includes request workspace id.

raintygao: `!!` is used for converting undefined to false, right?  This looks more clear, thanks.

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
```diff
@@ -179,6 +189,15 @@ export class WorkspaceSavedObjectsClientWrapper {
     return hasPermission;
   }
 
+  // Data source is a workspace level object, validate if data source has the permission to the given `workspaceIds`
```
ruanyl, line 192:
```
validate if data source has the permission to the given `workspaceIds`
```
what do you mean by this? should it be
```
validate if the request have access to the data source within the requested workspace?
```

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
```diff
@@ -179,6 +189,15 @@ export class WorkspaceSavedObjectsClientWrapper {
...
+  private validateDataSourcePermissions = (
+    object: SavedObject,
+    request: OpenSearchDashboardsRequest
+  ) => {
+    const requestWorkspaceId = getWorkspaceState(request).requestWorkspaceId;
+    return !requestWorkspaceId || !!object.workspaces?.includes(requestWorkspaceId);
```
ruanyl, line 198: `!requestWorkspaceId` this covers the case when outside of a workspace?

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
raintygao: Yes, updated.

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
raintygao: No, in `WorkspaceState` type, `requestWorkspaceId` could be undefined. Here just want to ensure that the workspaceId to be compared is existing.

wanglam: The `requestWorkspaceId` should be exists if user in a workspace. In another words,  user should be out of workspace if requestWorkspaceId is undefined right?

raintygao: We don't need to care whether is in or out of workspace especially here.

SuZhou-Joe: Imaging user is in global(requestWorkspaceId is undefined) and try to get a datasource(without workspaces) by `id`, we should prevent user from getting the datasource detail.

raintygao: I think this may be a common case in OSD, this doesn't belong to the concept of workspace/assign/unassign, we should not do extra limitation.

SuZhou-Joe: @wanglam @ruanyl what's your opinion?
IMO, regarding the restriction will only be turned on when permission control and workspace is enabled, `Reading` it from all kinds of unexpected approaches should be prevented.

SuZhou-Joe: Yeah, agreed. There may be some use cases that users may access data source detail in global.

SuZhou-Joe: But can we add a unit test to cover the case when requestWorkspaceId is empty?

src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts
ruanyl: Not a big concern, but I do like your original implementation, it's more readable to me, when I read the code, I can easily get that we only disallow the access to data source when request with workspace but the data source is not associated with the workspace. ;D

SuZhou-Joe (approved):

ruanyl (approved):

PR #7609, opened 2024-08-01
[dataset-nav][bug] get connections by cluster and update client namespace

kavilla:
### Description

There was a conflict in namespaces when registering a transport client with OpenSearch cluster. The schemas were also conflicting with observability and data source mgmt endpoints.

Modifying the namespace for query enhancements to avoid conflicts and then initialize the external datasource connections on load.

### Issues Resolved

n/a

## Changelog

- fix: dataset nav to load external connections and update namespace

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/query_enhancements/server/plugin.ts
```diff
@@ -18,17 +18,16 @@ import { ConfigSchema } from '../common/config';
 import { defineRoutes } from './routes';
 import {
   pplSearchStrategyProvider,
+  pplRawSearchStrategyProvider,
```
ashwin-pc, line 21: nit: are we using this?

src/plugins/query_enhancements/server/search/index.ts
```diff
@@ -4,5 +4,6 @@
  */
 
 export { pplSearchStrategyProvider } from './ppl_search_strategy';
+export { pplRawSearchStrategyProvider } from './ppl_raw_search_strategy';
```
ashwin-pc, line 7: nit: I dont see this being referenced anywhere

ashwin-pc (approved):
LGTM, small nits but none are blockers

abbyhu2000 (approved):

src/plugins/query_enhancements/server/search/index.ts
kavilla: it probably wasn't required but i just moved it to the right location it was previously being exported differently than the other search strategies.,

PR #7625, opened 2024-08-06
refactor: simplify theme configuration and defaulting

virajsanghvi:
### Description

Reduces the amount of logic needed to add a new theme, reduces the number of files to be touched, + documents the process. There should be no change in experience.

Additionally, it appears we were persisting the label value in some places, while using the version string in others. This worked because we only ever checked that themeTag versions matched 'v7', but v8/next was the else case so it just so happened to work correctly. This tries to use the version string consistently, while keeping the label used for persisting to advanced settings (because I couldn't find an easy way to translate persisted values for existing configurations).

Unfortunately, due to need to run in different contexts, I had to write this to the least common denominator, so changes across the board are still necessary.

Note: There is an existing bug with browser based settings that doesn't respect the user selected setting in the topnav theme selector (even though the selected theme is actually defined). This isn't a released feature and not making things worse, so didn't address. Bug opened: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/7689

### Issues Resolved

N/A

## Screenshot

| Context | Screenshot |
| ---- | ---- |
| v7 light | image|
| v7 dark | image|
| v8 light | image| 
| v8 dark | image| 

## Testing the changes

Most of the lines were already covered. The user header theme selection doesn't appear to be. Patch codecov was higher so I didn't add, but can if desired.

To test I:
- With user theme configuration turned off, made sure I could select the theme from advanced settings and see the correct result and css files were loaded as expected, and that returning to advanced settings the correct theme was selected
- With user theme configuration turned on, made sure sure I could select the theme from top nav and see correct result  and css files were loaded as expected, and that returning to the selector, the correct theme was selected
- Added a new theme (locally), and validated the above scenarios
- Bootstrapping/building locally still succeeds

## Changelog
- refactor: simplify theme configuration and defaulting
- deprecate: Deprecating `CssDistFilename` exports in favor of `themeCssDistFilenames` in `@osd/ui-shared-deps`

### Check List

- [X] All tests pass
  - [X] `yarn test:jest`
  - [X] `yarn test:jest_integration`
- [X] New functionality has been documented.


src/core/server/ui_settings/ui_settings_config.ts
```diff
@@ -56,7 +57,7 @@ const configSchema = schema.object({
   overrides: schema.object({}, { unknowns: 'allow' }),
   defaults: schema.object({
     'theme:darkMode': schema.maybe(schema.boolean({ defaultValue: false })),
-    'theme:version': schema.maybe(schema.string({ defaultValue: 'v8' })),
+    'theme:version': schema.maybe(schema.string({ defaultValue: defaultThemeVersion })),
```
virajsanghvi, line 60: Note: I'm not 100% how this is used, so continued to use the actual version strings instead of labels (like settings/theme.ts). I think this shouldn't make this worse, but if this is supposed to use the same values as the other file, lmk and will update.

packages/osd-ui-shared-deps/index.d.ts
```diff
@@ -44,29 +44,14 @@ export const jsFilename: string;
...
-
-/**
- * Filename of the dark-theme css file in the distributable directory
- */
-export const darkV8CssDistFilename: string;
-
-/**
- * Filename of the light-theme css file in the distributable directory
- */
-export const lightCssDistFilename: string;
```
AMoo-Miki, line 64: These are part of our public API. If we intend to keep these changes only in 3.0.0, removing these would be fine but if we intend to backport this PR, we will need to continue to provide these and mark them as deprecated.

packages/osd-ui-shared-deps/theme.ts
```diff
@@ -30,20 +30,21 @@
...
+import { resolveThemeTag, resolveThemeVersion } from './theme_config';
 
 const globals: any = typeof window === 'undefined' ? {} : window;
 
 export type Theme = typeof LightTheme;
 
 // in the OpenSearch Dashboards app we can rely on this global being defined, but in
 // some cases (like jest) the global is undefined
-export const tag: string = globals.__osdThemeTag__ || 'v8light';
-export const version = tag.startsWith('v7') ? 7 : 8;
```
AMoo-Miki, line 41: This would also be a breaking change.

src/core/server/ui_settings/ui_settings_config.ts
AMoo-Miki: I donno how I feel about a dependency defining the default value.

packages/osd-ui-shared-deps/index.d.ts
virajsanghvi: Is using jsdoc good enough for marking deprecated?

packages/osd-ui-shared-deps/theme.ts
virajsanghvi: How is this breaking?

src/core/server/ui_settings/ui_settings_config.ts
virajsanghvi: If the goal is the schema is the source of truth, it should be factored that way. Currently, the default is spread in multiple places and this really isn't factored to be the source of truth (a dependency needs this value). I can change this to be independent, but not sure if we're gaining more than we're losing. I'm not sure how this practically matters given how osd is built today, but maybe i'm missing something about how consumers use this setup.

AMoo-Miki: > If the goal is the schema is the source of truth, it should be factored that way

No piece of code should need to know what the default is; they should all rely on UISettings to get the theme that is active. If any other piece of code is making decisions about defaults, it is just bad logic and would need to change.

I will pull in @ashwin-pc who might know otherwise.

packages/osd-ui-shared-deps/index.d.ts
AMoo-Miki: What we do is add `@deprecated` here and call it out in the CHANGELOG.

packages/osd-ui-shared-deps/theme.ts
AMoo-Miki: Currect; only the version which was a number is now not a number. A simple way out would be to expose `version` as a number that we don't use but create a new `versionString` or anything you like for the future and then mark version as deprecated.

src/core/server/ui_settings/ui_settings_config.ts
ashwin-pc: @AMoo-Miki Correct me if im wrong but is you concern here that we are updating the config default from an import? or that we have multiple places where the default theme is referenced?

Setting the default value for the UI setting like this should be okay. It is after all referencing a static value, As for seting the default theme as a fallback in the other places seems unnecessary since the config should already have atleast the default value. 

Maybe that was out of scope for the PR, but why cant we keep just one source of truth for what the current theme is and let it also handle how to fallback to the default if a value is not set?

AMoo-Miki: > is you concern here that we are updating the config default from an import? or that we have multiple places where the default theme is referenced?

Both. I like that this PR tries to centralize it but we shouldn't even have to because everyone should fetch any such value from UISettings and since a default value is provided there, they should trust that it exists. Less importantly, it feel wrong to me that OSD itself doesn't own the default value and that it breaks the existing pattern.

virajsanghvi: > but we shouldn't even have to because everyone should fetch any such value from UISettings and since a default value is provided there

A problem with this statement is that @osd/ui-shared-deps, and I think optimizer, actually need this value. So a dependency of OSD would require OSD to define the default. If we want the schema to be the source of truth, that's fine, but IMO it likely needs to be factored into a package so it can used that way (at that point OSD would be depending on a package for its defaults - IMO not a problem if that's its responsibility - also would make it easier to swap out defaults).

I actually don't even understand how this value is used - uisettingsclient takes the default value from value of the uisettings definition, not the schema, e.g. https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a360b38506a2665e76059534328c8a86c2449cb8/src/core/public/ui_settings/ui_settings_client.ts#L100 (although a couple lines above it, a comment says to use the defaultValue property). 

Anyways, I'm fine reverting the change in this schema file as it makes not real difference today (I'm not changing the default), but I think

AMoo-Miki (approved):

zhongnansu (approved):

PR #7636, opened 2024-08-07
Move set default data source before workspace update

wanglam:
### Description

This PR is for fixing dev server crash after un-assign self as a workspace owner. In previous implementation, the set default data source will be called after permission updated. The user won't have `write` permission to the workspace. So the set default data source operation will be failed and throw exception. Then there will be a unhandled promise rejection. Move set default data source before workspace update to avoid this issue. 

## Screenshot

No UI Changes.

## Changelog

- fix: [Workspace] Move set default source order to avoid dev server crash

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/server/workspace_client.test.ts
+++ b/src/plugins/workspace/server/workspace_client.test.ts
@@ -181,6 +181,24 @@ describe('#WorkspaceClient', () => {
     expect(mockCheckAndSetDefaultDataSource).toHaveBeenCalledWith(uiSettingsClient, ['id1'], true);
   });
 
+  it('update# should log error when failed set default data source', async () => {
+    const client = new WorkspaceClient(coreSetup, logger);
+    await client.setup(coreSetup);
+    client?.setSavedObjects(savedObjects);
+    client?.setUiSettings(uiSettings);
+    mockCheckAndSetDefaultDataSource.mockRejectedValueOnce(new Error());
+
+    await client.update(mockRequestDetail, mockWorkspaceId, {
+      name: mockWorkspaceName,
+      permissions: {},
+      dataSources: ['id1'],
+    });
+
+    expect(logger.error).toHaveBeenLastCalledWith(
+      'Set default data source error during workspace updating'
+    );
+  });
+
   it('delete# should unassign data source before deleting related saved objects', async () => {
     const client = new WorkspaceClient(coreSetup, logger);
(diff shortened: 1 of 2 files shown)


SuZhou-Joe (dismissed):

Hailong-am:
> So the set default data source operation will be failed and throw exception. Then there will be a unhandled promise rejection

but we have try catch for whole update operation, not sure why it does not catch it and cause server crash

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/workspace/server/workspace_client.ts#L221-L293

wanglam:
> > So the set default data source operation will be failed and throw exception. Then there will be a unhandled promise rejection
> 
> but we have try catch for whole update operation, not sure why it does not catch it and cause server crash
> 
> https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/workspace/server/workspace_client.ts#L221-L293

The `checkAndSetDefaultDataSource‎ ` is an async function but we didn't wait it complete. The error occurs after `workspaceClient.update` method executed complete, so there will be an unhandled promise rejection.

Hailong-am:
> > > So the set default data source operation will be failed and throw exception. Then there will be a unhandled promise rejection
> > 
> > 
> > but we have try catch for whole update operation, not sure why it does not catch it and cause server crash
> > https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/workspace/server/workspace_client.ts#L221-L293
> 
> The `checkAndSetDefaultDataSource‎ ` is an async function but we didn't wait it complete in the `workspaceClient.update` method. The error occurs after `workspaceClient.update` method executed complete, so there will be an unhandled promise rejection.

I see, thanks!

Hailong-am (approved):

ruanyl:
@wanglam I guess the issue only happens when user self-removed from the workspace and assign/un-assign data source at the same time?

wanglam:
> @wanglam I guess the issue only happens when user self-removed from the workspace and assign/un-assign data source at the same time?

@ruanyl I think it's not related to the assign/un-assign data source, the default data source will always be set to undefined if passed data sources is empty. In this case, I didn't change the data sources of workspace.

ruanyl (approved):

SuZhou-Joe (approved):

PR #7640, opened 2024-08-07
[Workspace] Update workspace list page table

Qxisylolo:
### Description

In this pr, I refractor the workList table by adding 2 new fields, integrating 3 actions[delete, copyId, edit] , utilizing the Use Case selector, creating an empty state and enabling multiple deletion.

New update:
save a new field Use Case which is extracted from field features.
Add multiple deletion

## Screenshot
Screenshot 2024-08-16 at 09 33 19

When hovering over the description, you will see the tool tip.
Screenshot 2024-08-16 at 09 33 54

Perform filtering by Use case
Screenshot 2024-08-16 at 09 34 19
Screenshot 2024-08-16 at 09 34 40

Integrating 3 actions
Screenshot 2024-08-09 at 09 43 17

Performing search
Screenshot 2024-08-16 at 09 35 03

Empty state
Screenshot 2024-08-08 at 09 45 12

Enable multiple deletion
Screenshot 2024-08-08 at 21 39 35

Screenshot 2024-08-08 at 21 39 41

## Changelog
- feat: [Workspace] Update workspace list page table

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/workspace/public/components/workspace_list/index.tsx
```diff
@@ -38,23 +43,75 @@ export interface WorkspaceListProps {
...
+
+    return cleanWorkspaceId(appUrl);
+  }, [application, http]);
+
+  const [message, setMessage] = useState<React.ReactNode>(
+    <EuiEmptyPrompt
+      iconType="spacesApp"
+      title={<h3>No workspace available</h3>}
+      titleSize="s"
+      body="There are no workspace to display. Create workspace to get started."
```
raintygao, line 95: Better to wrap with i18n

src/plugins/workspace/public/components/workspace_list/index.tsx
```diff
@@ -65,129 +122,228 @@ export const WorkspaceList = ({ registeredUseCases$ }: WorkspaceListProps) => {
...
-        return (
-          item.id.toLowerCase().indexOf(normalizedQuery) > -1 ||
-          item.name.toLowerCase().indexOf(normalizedQuery) > -1
-        );
-      });
-      return result;
-    }
-    return workspaceList;
-  }, [workspaceList, queryInput]);
+  const formatDate = function (dateString: string) {
```
raintygao, line 125: Better to move `formatDate`  function from this component and tests file to a utils file to maintain. Besides, OSD uses `moment` as dependency, could we use some format functions in moment directly?

src/plugins/workspace/public/components/workspace_list/index.tsx
```diff
@@ -65,129 +122,228 @@ export const WorkspaceList = ({ registeredUseCases$ }: WorkspaceListProps) => {
...
+    const day = String(date.getUTCDate()).padStart(2, '0');
+    const year = date.getUTCFullYear();
+
+    const hours = String(date.getUTCHours()).padStart(2, '0');
+    const minutes = String(date.getUTCMinutes()).padStart(2, '0');
+    const seconds = String(date.getUTCSeconds()).padStart(2, '0');
+    const milliseconds = String(date.getUTCMilliseconds()).padStart(3, '0');
+
+    return `${month} ${day},${year}@${hours}:${minutes}:${seconds}.${milliseconds}`;
+  };
```
Hailong-am, line 153: can we use [moment.js](https://momentjs.com/) to achieve this?

src/plugins/workspace/public/components/workspace_list/index.tsx
Qxisylolo: thanks for your advice

src/plugins/workspace/public/components/workspace_list/index.tsx
Qxisylolo: Thanks for you review, [moment.js](https://momentjs.com/) is really useful, 
Screenshot 2024-08-07 at 18 20 01
but I'm afraid there is not the expected format shown in figma

src/plugins/workspace/public/components/workspace_list/index.tsx
Qxisylolo: thanks for your advice

src/plugins/workspace/public/components/workspace_list/index.tsx
Hailong-am: can you try with format `'MMM DD[,]YYYY [@] HH:mm:ss.SSS'` reference https://momentjs.com/docs/#/displaying/

Qxisylolo: Yeah, this works, thank you!

wanglam: image
Shall we use the date format setting in the advanced settings?

src/plugins/workspace/public/components/workspace_list/index.tsx
```diff
@@ -65,129 +131,198 @@ export const WorkspaceList = ({ registeredUseCases$ }: WorkspaceListProps) => {
...
+      render: (useCase: string[]) => {
+        if (!useCase || useCase.length === 0) {
           return '';
         }
-        const useCaseId = getFirstUseCaseOfFeatureConfigs(features);
-        const useCase =
+        const useCaseId = getFirstUseCaseOfFeatureConfigs(useCase);
+        if (useCaseId) {
+          addFeatureFilter(useCaseId);
+        }
```
ruanyl, line 212: Why need to call `addFeatureFilter` here?

src/plugins/workspace/public/components/workspace_list/index.tsx
```diff
@@ -38,23 +44,83 @@ export interface WorkspaceListProps {
...
   const initialSortField = 'name';
   const initialSortDirection = 'asc';
   const workspaceList = useObservable(workspaces?.workspaceList$ ?? of([]), []);
-  const [queryInput, setQueryInput] = useState<string>('');
+
+  const newWorkspaceList: WorkspaceAttributeWithUseCaseID[] = useMemo(() => {
+    return workspaceList.map(
+      (workspace): WorkspaceAttributeWithUseCaseID => ({
+        ...workspace,
+        useCase: [...(workspace.features || [])],
```
ruanyl, line 65: We can make it a string instead of array, then you don't need to call `getFirstUseCaseOfFeatureConfigs` many times
```suggestion
        useCase: getFirstUseCaseOfFeatureConfigs(workspace.features || []) ?? ''
```

src/plugins/workspace/public/components/workspace_list/index.tsx
```diff
@@ -65,129 +131,198 @@ export const WorkspaceList = ({ registeredUseCases$ }: WorkspaceListProps) => {
...
   ];
 
-  const workspaceCreateUrl = useMemo(() => {
-    if (!application || !http) {
-      return '';
-    }
-
-    const appUrl = application.getUrlForApp(WORKSPACE_CREATE_APP_ID, {
-      absolute: false,
+  useMemo(() => {
```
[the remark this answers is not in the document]
ruanyl: actually you don't need a state for `featureFilters`, it can be derived from the workspace list, assume `workspace.useCase` is a string as I suggested
```typescript
  const featureFilters = useMemo(() => {
    const allUseCases = new Set<string>();
    newWorkspaceList.forEach((workspace) => {
      if (workspace.useCase) {
        allUseCases.add(workspace.useCase);
      }
    });
    return [...allUseCases];
  }, [newWorkspaceList]);
```

src/plugins/workspace/public/components/workspace_list/index.tsx
ruanyl: Seems we can simply remove it

src/plugins/workspace/public/components/workspace_list/index.tsx
ruanyl: Seems there is bug when it's an array, see my screenshot
image

src/plugins/workspace/public/components/workspace_list/index.tsx
```diff
@@ -12,12 +13,17 @@ import {
...
+  copyToClipboard,
+  EuiButtonEmpty,
+  EuiEmptyPrompt,
 } from '@elastic/eui';
 import useObservable from 'react-use/lib/useObservable';
 import { BehaviorSubject, of } from 'rxjs';
 import { i18n } from '@osd/i18n';
-import { debounce, DEFAULT_NAV_GROUPS } from '../../../../../core/public';
+import { DEFAULT_NAV_GROUPS } from '../../../../../core/public';
 import { WorkspaceAttribute } from '../../../../../core/public';
```
ruanyl, line 27: We can combine them
```suggestion
import { DEFAULT_NAV_GROUPS, WorkspaceAttribute } from '../../../../../core/public';
```

src/plugins/workspace/public/components/workspace_list/index.tsx
```diff
@@ -38,23 +44,83 @@ export interface WorkspaceListProps {
...
+    );
+  }, [workspaceList]);
+
   const [pagination, setPagination] = useState({
     pageIndex: 0,
     pageSize: 5,
     pageSizeOptions: [5, 10, 20],
   });
   const [deletedWorkspace, setDeletedWorkspace] = useState<WorkspaceAttribute | null>(null);
+  const [featureFilters, setFeatureFilters] = useState<string[]>([]);
```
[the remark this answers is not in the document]
ruanyl: Perhaps we call it `useCaseFilters`?

src/plugins/workspace/public/components/workspace_list/index.tsx
```diff
@@ -65,129 +131,198 @@ export const WorkspaceList = ({ registeredUseCases$ }: WorkspaceListProps) => {
...
+          'data-test-subj': 'workspace-list-copy-id-icon',
+          render: ({ id }: WorkspaceAttribute) => {
+            return (
+              <EuiButtonEmpty
+                onClick={() => handleCopyId(id)}
+                size="xs"
+                iconType="copy"
+                color="text"
+              >
+                <EuiText size="m">Copy</EuiText>
```
ruanyl, line 267: Shall we say "Copy ID"?

src/plugins/workspace/public/components/workspace_list/index.tsx
Qxisylolo: Thank you for your comments. I have updated the array to a string and addressed the other issues mentioned. Could you please review the changes and let me know if the bug still persists? Thank you!

src/plugins/workspace/public/components/workspace_list/index.tsx
```diff
@@ -38,23 +45,100 @@ export interface WorkspaceListProps {
...
+  };
+
+  const newWorkspaceList: WorkspaceAttributeWithUseCaseID[] = useMemo(() => {
+    return workspaceList.map(
+      (workspace): WorkspaceAttributeWithUseCaseID => ({
+        ...workspace,
+        useCase: extractUseCaseFromFeatures(workspace.features ?? []),
+      })
+    );
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
raintygao, line 91: Why we use eslint ignore here?

Qxisylolo: Hi, I add it to prevent a warning about not including extractUseCaseFromFeatures in the dependency array, which is indirectly dependent on workspaceList.features, and it is already accounted for through workspaceList in the dependency array

raintygao: Using eslint ignore in component file would not look reasonable. What about adding extractUseCaseFromFeatures to dependency or wrapping extractUseCaseFromFeatures with `useCallback` and adding it to dependency.

Qxisylolo: Sure, I will update it, Thank you~

src/plugins/workspace/public/components/workspace_list/index.tsx
Qxisylolo: Thanks for your comment, already updated.

src/plugins/workspace/public/components/workspace_list/index.tsx
```diff
@@ -277,7 +277,7 @@ export const WorkspaceList = ({ registeredUseCases$ }: WorkspaceListProps) => {
       width: '25%',
       truncateText: false,
       render: (lastUpdatedTime: string) => {
-        return moment(lastUpdatedTime).format('MMM DD[,]YYYY [@] HH:mm:ss.SSS');
+        return moment(lastUpdatedTime).format('MMM D, YYYY @ HH:mm:ss.SSS');
```
wanglam, line 280: In fact, I'm not suggesting using the same date format string with advanced setting. I'm prefer read the ui settings directly. Here is sample code: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/dashboard/public/application/utils/get_table_columns.tsx#L19 Feel free to have a try.

Qxisylolo: okok, I see, thank you!

ruanyl (approved):

SuZhou-Joe (approved):

PR #7651, opened 2024-08-08
feat: [contentManagement] allow to update section input after page rendered

ruanyl:
This PR makes it possible to update the section input after the page is rendered, for example, when rendering a dashboard on the page, we need to apply the date range, filter and query on the fly.

## Changelog

- feat: [contentManagement] allow to update section input after page rendered

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/content_management/public/components/section_input.ts
+++ b/src/plugins/content_management/public/components/section_input.ts
@@ -32,6 +32,7 @@ export const createCardInput = (
     viewMode: ViewMode.VIEW,
     columns: section.columns,
     panels,
+    ...section.input,
   };
 
   contents.forEach((content) => {
@@ -74,7 +75,7 @@ export const createDashboardInput = async (
   const h = 15;
   const counter = new BehaviorSubject(0);
 
-  contents.forEach(async (content, i) => {
+  contents.forEach(async (content) => {
     counter.next(counter.value + 1);
     try {
       if (content.kind === 'dashboard') {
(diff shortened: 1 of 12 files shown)


TackAdam (approved):

SuZhou-Joe (approved):

src/plugins/content_management/public/components/section_input.ts
```diff
@@ -164,29 +165,27 @@ export const createDashboardInput = async (
...
-    id: section.id,
     timeRange: {
       to: 'now',
       from: 'now-7d',
     },
-    title: section.title ?? 'test',
     query: {
       query: '',
-      language: 'lucene',
+      language: 'kuery',
```
SuZhou-Joe, line 182: What's the difference between the language here? I guess it will be overwritten by `...section.input`

src/plugins/content_management/public/services/content_management/page.ts
```diff
@@ -31,6 +35,28 @@ export class Page {
...
+    callback: (section: Section | null, err?: Error) => Section | null
+  ) {
+    const section = this.sections.get(sectionId);
+    if (!section) {
+      callback(null, new Error(`Section id ${sectionId} not found`));
+      return;
+    }
+
+    const updated = callback(section);
+    if (updated) {
```
AMoo-Miki, line 49: How would you feel about doing
```
if (updated.kind !== section.kind) return;
```
... and then a switch-case for `dashboard`, `card`, and `custom`?

AMoo-Miki (approved):

src/plugins/content_management/public/services/content_management/page.ts
ruanyl: Good idea! I will adopt this change in the upcoming PR

src/plugins/content_management/public/components/section_input.ts
ruanyl: Yes, it can be override by `section.input`. Here it just changed the default language from `lucene` to `kuery(DQL)`.

src/plugins/content_management/public/services/content_management/page.ts
ruanyl: image
Seems TS compiler is not happy with that ;D

PR #7686, opened 2024-08-12
Change the locale dynamically by adding &i18n-locale to URL

ananzh:
### Description
The main issue was the inability to dynamically change the locale in OpenSearch Dashboards. Currently we need to update config file and i18nrc.json.

This PR allows  users to switch to a different locale (e.g., from English to Chinese) by appending or modifying the 'i18n-locale' parameter in the URL.

* getAndUpdateLocaleInUrl: If a non-default locale is found, this function reconstructs the URL with the locale parameter in the correct position.
* updated the ScopedHistory class, allowing it to detect locale changes and trigger reloads as necessary.
* modify the i18nMixin, which sets up the i18n system during server startup, to register all available translation files during server startup, not just the current locale.
* update the uiRenderMixin to accept requests for any registered locale and dynamically load and cache translations for requested locales.

### Issues Resolved
NA

## Screenshot

* For url without query params

https://github.com/user-attachments/assets/f18dde06-e8ef-4729-936e-b8db5e2ab096

* For url with query params

https://github.com/user-attachments/assets/21fc3c34-172f-4c6a-b43a-10ababacc717

* For unsupport locale, it will break with proper hint with a 404 not found error. This is different from current behavior which will use en without any hints.

Screenshot 2024-08-12 at 7 41 02 PM

* Share link 
Can be shared for both embed or link snapshot but not saved object. This is expected, bc this locale is attached to the url not saved in saved obj. 

https://github.com/user-attachments/assets/7e9c417c-2f63-48f6-a4fe-de9374330a7b

## Changelog
- feat: Change the locale dynamically by adding &i18n-locale to URL

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/packages/osd-i18n/src/core/i18n.ts
+++ b/packages/osd-i18n/src/core/i18n.ts
@@ -261,5 +261,12 @@ export async function load(translationsUrl: string) {
     throw new Error(`Translations request failed with status code: ${response.status}`);
   }
 
-  init(await response.json());
+  const data = await response.json();
+
+  if (data.warning) {
+    // Store the warning to be displayed after core system setup
+    (window as any).__i18nWarning = data.warning;
+  }
+
+  init(data.translations);
 }
--- a/src/core/public/locale_helper.test.ts
+++ b/src/core/public/locale_helper.test.ts
@@ -3,126 +3,48 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { extractLocaleInfo, getAndUpdateLocaleInUrl } from './locale_helper';
+import { getLocaleInUrl } from './locale_helper';
 
-describe('extractLocaleInfo', () => {
-  const testCases = [
-    {
-      description: 'After hash and slash',
-      input: 'http://localhost:5603/app/home#/&i18n-locale=fr-FR',
-      expected: {
-        localeValue: 'fr-FR',
-        localeParam: 'i18n-locale=fr-FR',
-        updatedUrl: 'http://localhost:5603/app/home#/',
-      },
-    },
-    {
-      description: 'After path and slash',
-      input: 'http://localhost:5603/app/home/&i18n-locale=de-DE',
-      expected: {
-        localeValue: 'de-DE',
-        localeParam: 'i18n-locale=de-DE',
-        updatedUrl: 'http://localhost:5603/app/home/',
-      },
-    },
-    {
-      description: 'No locale parameter',
-      input: 'http://localhost:5603/app/home',
-      expected: {
-        localeValue: 'en',
-        localeParam: null,
-        updatedUrl: 'http://localhost:5603/app/home',
-      },
-    },
-    {
-      description: 'Complex URL with locale',
-      input: 'http://localhost:5603/app/dashboards#/view/id?_g=(...)&_a=(...)&i18n-locale=es-ES',
-      expected: {
-        localeValue: 'es-ES',
-        localeParam: 'i18n-locale=es-ES',
-        updatedUrl: 'http://localhost:5603/app/dashboards#/view/id?_g=(...)&_a=(...)',
-      },
-    },
-  ];
-
-  testCases.forEach(({ description, input, expected }) => {
-    it(description, () => {
-      const result = extractLocaleInfo(input);
-      expect(result).toEqual(expected);
-    });
-  });
-});
-
-describe('getAndUpdateLocaleInUrl', () => {
-  let originalHistoryReplaceState: typeof window.history.replaceState;
-
+describe('getLocaleInUrl', () => {
   beforeEach(() => {
-    // Mock window.history.replaceState
-    originalHistoryReplaceState = window.history.replaceState;
-    window.history.replaceState = jest.fn();
+    // Clear any warnings before each test
+    delete (window as any).__localeWarning;
   });
 
-  afterEach(() => {
-    // Restore original window.history.replaceState
-    window.history.replaceState = originalHistoryReplaceState;
+  it('should return the locale from a valid query string', () => {
+    const url = 'http://localhost:5603/app/home?locale=en-US';
+    expect(getLocaleInUrl(url)).toBe('en-US');
   });
 
-  const testCases = [
-    {
-      description: 'Category 1: basePath + #/',
-      input: 'http://localhost:5603/app/home#/&i18n-locale=zh-CN',
-      expected: 'http://localhost:5603/app/home#/?i18n-locale=zh-CN',
-      locale: 'zh-CN',
-    },
-    {
-      description: 'Category 1: basePath + # (empty hashPath)',
-      input: 'http://localhost:5603/app/home#&i18n-locale=zh-CN',
-      expected: 'http://localhost:5603/app/home#?i18n-locale=zh-CN',
-      locale: 'zh-CN',
-    },
-    {
-      description: 'Category 2: basePath + # + hashPath + ? + hashQuery',
-      input: 'http://localhost:5603/app/dashboards#/view/id?_g=(...)&_a=(...)&i18n-locale=zh-CN',
-      expected: 'http://localhost:5603/app/dashboards#/view/id?_g=(...)&_a=(...)&i18n-locale=zh-CN',
-      locale: 'zh-CN',
-    },
-    {
-      description: 'Category 3: basePath only',
-      input: 'http://localhost:5603/app/management&i18n-locale=zh-CN',
-      expected: 'http://localhost:5603/app/management?i18n-locale=zh-CN',
-      locale: 'zh-CN',
-    },
-    {
-      description: 'Category 1: basePath + # + hashPath',
-      input: 'http://localhost:5603/app/dev_tools#/console&i18n-locale=zh-CN',
-      expected: 'http://localhost:5603/app/dev_tools#/console?i18n-locale=zh-CN',
-      locale: 'zh-CN',
-    },
-    {
-      description: 'URL without locale parameter',
-      input: 'http://localhost:5603/app/home#/',
-      expected: 'http://localhost:5603/app/home#/',
-      locale: 'en',
-    },
-    {
-      description: 'Complex URL with multiple parameters',
-      input:
-        "http://localhost:5603/app/dashboards#/view/7adfa750-4c81-11e8-b3d7-01146121b73d?_g=(filters:!(),refreshInterval:(pause:!f,value:900000),time:(from:now-24h,to:now))&_a=(description:'Analyze%20mock%20flight%20data',filters:!())&i18n-locale=zh-CN",
-      expected:
-        "http://localhost:5603/app/dashboards#/view/7adfa750-4c81-11e8-b3d7-01146121b73d?_g=(filters:!(),refreshInterval:(pause:!f,value:900000),time:(from:now-24h,to:now))&_a=(description:'Analyze%20mock%20flight%20data',filters:!())&i18n-locale=zh-CN",
-      locale: 'zh-CN',
-    },
-  ];
+  it('should return the locale from a valid hash query string', () => {
+    const url = 'http://localhost:5603/app/home#/?locale=fr-FR';
+    expect(getLocaleInUrl(url)).toBe('fr-FR');
+  });
 
-  testCases.forEach(({ description, input, expected, locale }) => {
-    it(description, () => {
-      const result = getAndUpdateLocaleInUrl(input);
-      expect(result).toBe(locale);
-      if (locale !== 'en') {
-        expect(window.history.replaceState).toHaveBeenCalledWith(null, '', expected);
-      } else {
-        expect(window.history.replaceState).not.toHaveBeenCalled();
-      }
-    });
+  it('should return null for a URL without locale', () => {
+    const url = 'http://localhost:5603/app/home';
+    expect(getLocaleInUrl(url)).toBeNull();
+  });
+
+  it('should return null and set a warning for an invalid locale format in hash', () => {
+    const url = 'http://localhost:5603/app/home#/&locale=de-DE';
+    expect(getLocaleInUrl(url)).toBeNull();
+    expect((window as any).__localeWarning).toBeDefined();
+    expect((window as any).__localeWarning.title).toBe('Invalid URL Format');
+  });
+
+  it('should return null for an empty locale value', () => {
+    const url = 'http://localhost:5603/app/home?locale=';
+    expect(getLocaleInUrl(url)).toBeNull();
+  });
+
+  it('should handle URLs with other query parameters', () => {
+    const url = 'http://localhost:5603/app/home?param1=value1&locale=ja-JP&param2=value2';
+    expect(getLocaleInUrl(url)).toBe('ja-JP');
+  });
+
+  it('should handle URLs with other hash parameters', () => {
+    const url = 'http://localhost:5603/app/home#/route?param1=value1&locale=zh-CN&param2=value2';
+    expect(getLocaleInUrl(url)).toBe('zh-CN');
   });
 });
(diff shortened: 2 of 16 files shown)


src/core/public/locale_helper.ts
```diff
@@ -0,0 +1,114 @@
...
+  url: string
+): { localeValue: string | null; localeParam: string | null; updatedUrl: string } {
+  const patterns = [
+    /[#&?](i18n-locale)=([^&/]+)/i, // Standard query parameter
+    /#\/&(i18n-locale)=([^&/]+)/i, // After hash and slash
+    /\/&(i18n-locale)=([^&/]+)/i, // After path and slash
+  ];
+
+  for (const pattern of patterns) {
+    const match = url.match(pattern);
```
joshuali925, line 22: can this do something like https://github.com/opensearch-project/OpenSearch-Dashboards/blob/75d67c06e72cf157368e5175cb44bc54c16607af/src/core/public/chrome/chrome_service.tsx#L141
instead of regex?

src/core/public/osd_bootstrap.ts
```diff
@@ -31,13 +31,41 @@
...
+    // This will affect all subsequent i18n.translate() calls
+    i18n.setLocale(urlLocale);
+
+    // Modify the translationsUrl to include the new locale
+    // This ensures that the correct translation file is requested from the server
+    // The replace function changes the locale in the URL, e.g.,
+    // from '/translations/en.json' to '/translations/zh-CN.json'
+    injectedMetadata.i18n.translationsUrl = injectedMetadata.i18n.translationsUrl.replace(
+      /\/([^/]+)\.json$/,
+      `/${urlLocale}.json`
```
joshuali925, line 63: if user sets an invalid locale through URL, is it the same behavior as setting it through config/opensearch_dashboards.yml and OSD UI won't break?

src/core/public/locale_helper.ts
```diff
@@ -0,0 +1,114 @@
...
+ *    Restructuring rationale: The '&' is changed to '?' because there were no existing
+ *    query parameters in the basePath. This ensures the locale is recognized as the
+ *    start of the query string rather than being misinterpreted as part of the path.
+ *
+ * The function performs the following steps:
+ * 1. Extracts the locale parameter from its current position in the URL.
+ * 2. Removes the locale parameter from its original position.
+ * 3. Reconstructs the URL, placing the locale parameter in the correct position
+ *    based on the URL structure to ensure proper parsing by OpenSearch Dashboards.
+ * 4. Updates the browser's URL without causing a page reload.
```
joshuali925, line 81: skimmed through seems the purpose is to handle invalid URL param syntax where `&` should be used instead of `?`. Do we need to handle that? i'm not sure if anything else in OSD does similar handling, since usually it's user's responsibility to correctly use URL parameters

src/core/public/osd_bootstrap.ts
```diff
@@ -31,13 +31,41 @@
...
+    );
+  } else {
+    i18n.setLocale('en');
```
joshuali925, line 66: why is this needed?

joshuali925 (commented):
thanks, just some questions

src/core/public/locale_helper.ts
ananzh: From my tests, I found 
1) http://localhost:5601/app/home#/&i18n-locale=zh-CN will not work
2) http://localhost:5601/app/home#/? query= (...) &i18n-locale=zh-CN can work.

The error will cause a blank page without proper prompt error and hard to trace. I think /&i18n-locale=zh-CN might be interpreted as a path rather than a query parameter, leading to routing errors or the parameter being ignored.

src/core/public/osd_bootstrap.ts
ananzh: I remember this is for user removing the `i18n-locale` part from url after first add it.

src/core/public/osd_bootstrap.ts
ananzh: It will break just as you set an invalid one in config/opensearch_dashboards.yml. Both will use load in i18n 
```
export async function load(translationsUrl: string) {
  // Once this package is integrated into core OpenSearch Dashboards we should switch to an abstraction
  // around `fetch` provided by the platform, e.g. `kfetch`.
  const response = await fetch(translationsUrl, {
    credentials: 'same-origin',
  });

  if (response.status >= 300) {
    throw new Error(`Translations request failed with status code: ${response.status}`);
  }

  init(await response.json());
}
```
this fetch will call locale route in src/legacy/ui/ui_render/ui_render_mixin.js
```
server.route({
    path: '/translations/{locale}.json',
    method: 'GET',
    config: { auth: false },
    handler: async (request, h) => {
      const { locale } = request.params;
      const normalizedLocale = locale.toLowerCase();
      const registeredLocales = i18nLoader.getRegisteredLocales().map((l) => l.toLowerCase());

      // Function to get or create cached translations
      const getCachedTranslations = async (localeKey, getTranslationsFn) => {
        if (!translationsCache[localeKey]) {
          const translations = await getTranslationsFn();
          translationsCache[localeKey] = {
            translations: JSON.stringify(translations),
            hash: createHash('sha1').update(JSON.stringify(translations)).digest('hex'),
          };
        }
        return translationsCache[localeKey];
      };

      let cachedTranslations;

      if (normalizedLocale === defaultLocale.toLowerCase()) {
        // Default locale
        cachedTranslations = await getCachedTranslations(defaultLocale, () =>
          i18n.getTranslation()
        );
      } else if (registeredLocales.includes(normalizedLocale)) {
        // Other registered locales
        cachedTranslations = await getCachedTranslations(normalizedLocale, () =>
          i18nLoader.getTranslationsByLocale(locale)
        );
      } else {
        // Locale not found
        throw Boom.notFound(`Unknown locale: ${locale}`);
      }

      return h
        .response(cachedTranslations.translations)
        .header('cache-control', 'must-revalidate')
        .header('content-type', 'application/json')
        .etag(cachedTranslations.hash);
    },
  });
```
for config change it will go to clause `normalizedLocale === defaultLocale.toLowerCase()` and for dynamic it will go to `else if (registeredLocales.includes(normalizedLocale))` because defaultLocale is the current i18n's locale. So if we have a dynamic locale, we will check if it has a translation file otherwise will throw `throw Boom.notFound(`Unknown locale: ${locale}`);`, which is same as if you pass an invalid locale to config.

src/core/public/locale_helper.ts
ananzh: will try this one.

src/core/public/locale_helper.ts
joshuali925: `http://localhost:5601/app/home#/&i18n-locale=zh-CN` this is not correct syntax, anything other parameters like `http://localhost:5601/app/home#/&xxx` also doesn't work. My point is there are many different support url params, do we need to do this special handling just for `i18n-locale`?

src/core/public/osd_bootstrap.ts
joshuali925: got it, thanks

src/core/public/locale_helper.ts
joshuali925: the reason i'm asking this is that if user puts something like `/app/home#/&i18n-locale=en&embed=true` or the correct way `/app/home#/?i18n-locale=en&embed=true` i don't think this logic handles it, the implementation might not be generic enough to handle URL params

ananzh: Thank you for this feedback. I'd like to clarify a few points:

1) I assumed that `&i18n-locale` would always be attached at the end of the URL. Is this assumption correct in all cases?
2) Regarding format and validation, what specific aspects should we be concerned about? Are there any edge cases we need to handle?
3) For cases like `http://localhost:5601/app/home#/&i18n-locale=zh-CN`, should our function handle this, or should the app be responsible for ensuring the correct URL structure (ex http://localhost:5601/app/home/&i18n-locale=zh-CN) before our function is called?
4. If the app is responsible for URL structure, how should we document this requirement for developers using this feature?

src/core/public/locale_helper.ts
ananzh: Oh I remember why I switch to regex. The searchParams property of the URL object is designed to work with standard query parameters, which are typically preceded by a question mark (?) in the URL.
For the URL like `http://localhost:5601/app/home/&i18n-locale-zh-CN`, it doesn't follow the standard format for query parameters.The & is being used as if it's starting a query string, but without a preceding ?.The searchParams approach would not work directly and searchParams.get('i18n-locale') will return null. I think this is the same issue as our later discussion. 

If we always have standard query parameters then we are good to switch to searchParams. But if in the future we want to translate urls like our current home page `http://localhost:5603/xqy/app/home#/` or management `http://localhost:5603/xqy/app/management` then we can't use this. I am trying to make this feature applicable for future extension.

src/core/public/osd_bootstrap.ts
ananzh: sorry. my comment is not correct. I think there is a bug in the previous behavior: once we defined in the config this `i18n.getLocale() !== locale.toLowerCase()` is always true so we never call ` throw Boom.notFound(`Unknown locale: ${locale}`);`
```
handler(request, h) {
      // OpenSearch Dashboards server loads translations only for a single locale
      // that is specified in `i18n.locale` config value.
      const { locale } = request.params;
      if (i18n.getLocale() !== locale.toLowerCase()) {
        throw Boom.notFound(`Unknown locale: ${locale}`);
      }

      // Stringifying thousands of labels and calculating hash on the resulting
      // string can be expensive so it makes sense to do it once and cache.
      if (translationsCache.translations == null) {
        translationsCache.translations = JSON.stringify(i18n.getTranslation());
        translationsCache.hash = createHash('sha1')
          .update(translationsCache.translations)
          .digest('hex');
      }

      return h
        .response(translationsCache.translations)
        .header('cache-control', 'must-revalidate')
        .header('content-type', 'application/json')
        .etag(translationsCache.hash);
    },
```

I am not sure if this a proper behavior because user has no idea that they pass the wrong locale. If we want to stick to this behavior we could change to 
```
else {
        // Locale not found
        cachedTranslations = await getCachedTranslations('en', () =>
          i18nLoader.getTranslationsByLocale('en')
        );
      }
```

ashwin-pc:
Nice! The feature looks good. A few questions and comments on the implementation though:
1. Will the locale always be there in the URL? if so can we make it optional?
2. Why did we change the behaviour when the locale is missing? Cant we load the default locale and show the error in the console instead? Why break the page for the user?
3. Why are you supporting invalid URL formats? Whats the need for this?

ananzh:
> Nice! The feature looks good. A few questions and comments on the implementation though:
> 
> 1. Will the locale always be there in the URL? if so can we make it optional?
> 2. Why did we change the behaviour when the locale is missing? Cant we load the default locale and show the error in the console instead? Why break the page for the user?
> 3. Why are you supporting invalid URL formats? Whats the need for this?

*For Q1:
locale can be removed. if no specified locale, we will use the default one specified in the config opensearch_dashboards.yml or `en` if not specified.

* For Q2:
I could change the current hehaviour to not break the page and add a prompt to hint user.

* For Q3:
We don't have to support invalid URL formats. Should we just let break the page if the url is not valid?

src/core/public/locale_helper.ts
joshuali925: > 1. I assumed that &i18n-locale would always be attached at the end of the URL. Is this assumption correct in all cases?

I think we just want to give a reasonable API for client to override OSD locale, which URL parameter should be enough in this case. When this feature exposes to users we should advertise it as "adding a URL parameter of `i18n-locale` to change locale", not "append `&i18n-locale` to change locale", so i don't think we need to handle invalid URL parameters

> 2. Regarding format and validation, what specific aspects should we be concerned about? Are there any edge cases we need to handle?

Since there are similar existing implementation in OSD (`embed` parameter), I think following the same implementation should be enough

for 3 and 4 i'm not sure what "app" refers to, could you clarify? But user/consumer of the feature should be responsible to provide correct URL parameter, the function should read it in a way respecting to searchParams standard

src/core/public/locale_helper.ts
joshuali925: I think it's answered in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7686#discussion_r1715682670
> When this feature exposes to users we should advertise it as "add a URL parameter of i18n-locale to change locale", not "append `&i18n-locale` to change locale", so i don't think we need to handle invalid URL parameters

kavilla:
sorry haven't dived into it deep and it probably was covered somewhere but how come we didn't want to base it on the locale from the user browser?

src/core/public/locale_helper.test.ts
```diff
@@ -0,0 +1,128 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { extractLocaleInfo, getAndUpdateLocaleInUrl } from './locale_helper';
+
+describe('extractLocaleInfo', () => {
+  const testCases = [
+    {
+      description: 'After hash and slash',
+      input: 'http://localhost:5603/app/home#/&i18n-locale=fr-FR',
```
kavilla, line 12: little bit worried about how long the url can get and as of late we have been able to hit the limit allowed. Is there another way we can indicate the locale without specifically calling it `i18n-locale`. I think I glanced it somewhere else but I think you called it out and makes sense why it is like this.

Because the i18n-locale is the template you provide with the value so it make sense. I've seen `g11n` be used for `globalization` by others to indicate it is the combination of both. is it worth it to do that?

joshuali925: i think we also want to support dark theme through URL params, not sure about other settings through URL. If we add support for other settings might be good to have some consistent convention to convert setting name to URL parameter name

ashwin-pc: HTTP headers is also another way to specify the locale. Can we support that?

ananzh:
Made some adjustments according to the comments

* no re-construction on invalid or non-standard url: only toast warning for invalid or non-standard url
* no break if locale is not supported (no translation file)


https://github.com/user-attachments/assets/8fc2093c-8cdf-4d38-a51a-d5460237abeb

ashwin-pc:
> Made some adjustments according to the comments
> 
> * no re-construction on invalid or non-standard url: only toast warning for invalid or non-standard url
> * no break if locale is not supported (no translation file)
> 
>  2024-08-14_09-05-25.mp4

@kgcreative any concerns with this toast pattern?

packages/osd-i18n/src/core/i18n.ts
```diff
@@ -261,5 +261,12 @@ export async function load(translationsUrl: string) {
     throw new Error(`Translations request failed with status code: ${response.status}`);
   }
 
-  init(await response.json());
+  const data = await response.json();
+
+  if (data.warning) {
+    // Store the warning to be displayed after core system setup
+    (window as any).__i18nWarning = data.warning;
```
joshuali925, line 268: is it common to store this in window?

ananzh: I think it is not uncommon, especially in situations where we need to pass information from sever side to browser that doesn't have a direct communication channel.

kgcreative:
Seems consistent with the way we handle toasts across the application today. I would prefer we use the `globe`, `questionInCircle`, or `alert` icons instead of the help icon.

ananzh:
> Seems consistent with the way we handle toasts across the application today. I would prefer we use the `globe`, `questionInCircle`, or `alert` icons instead of the help icon.

Hi @kgcreative since the feature is for developer only, toasts might not be necessary, cuz user will never managing the url. Therefore I just change to console warning.

joshuali925 (approved):
lgtm, thanks

LDrago27 (approved):

PR #7701, opened 2024-08-13
Add an unified storage class for data plugin

abbyhu2000:
### Description
* Created a more advanced storage class for data plugin 
* Refactor logics from previous storage usages
* Add Query History class for future recent query use

## Changelog

- feat: Add a unified storage class for data plugin

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/common/storage/index.ts
+++ b/src/plugins/data/common/storage/index.ts
@@ -1,0 +1,12 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+export { DataStorage, createStorage } from './storage';
--- a/src/plugins/data/common/storage/index.ts
+++ b/src/plugins/data/common/storage/index.ts
@@ -1,12 +1,6 @@
 /*
+ * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
- *
- * The OpenSearch Contributors require contributions made to
- * this file be licensed under the Apache-2.0 license or a
- * compatible open source license.
- *
- * Any modifications Copyright OpenSearch Contributors. See
- * GitHub history for details.
  */
 
 export { DataStorage, createStorage } from './storage';
--- a/src/plugins/data/common/storage/storage.ts
+++ b/src/plugins/data/common/storage/storage.ts
@@ -1,0 +1,74 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { transform, startsWith, keys } from 'lodash';
+import { parse, stringify } from '@osd/std';
+
+export enum StorageKeys {
+  WIDTH = 'widths',
+}
+
+type IStorageEngine = typeof window.localStorage;
+export class DataStorage {
+  constructor(private readonly engine: IStorageEngine, private readonly prefix: string) {}
+
+  encode(val: any) {
+    return stringify(val);
+  }
+
+  decode(val: any) {
+    if (typeof val === 'string') {
+      return parse(val);
+    }
+  }
+
+  encodeKey(key: string) {
+    return `${this.prefix}${key}`;
+  }
+
+  decodeKey(key: string) {
+    if (startsWith(key, this.prefix)) {
+      return `${key.slice(this.prefix.length)}`;
+    }
+  }
+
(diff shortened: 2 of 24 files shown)


src/plugins/data/common/storage/index.ts
```diff
@@ -0,0 +1,12 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
```
kavilla, line 4: i believe, the license should be:

```
/*
 */
```

src/plugins/data/common/storage/storage.ts
```diff
@@ -0,0 +1,74 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
```
kavilla, line 2: license header

src/plugins/data/common/storage/storage.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import { transform, startsWith, keys } from 'lodash';
+import { parse, stringify } from '@osd/std';
+
+export enum StorageKeys {
+  WIDTH = 'widths',
+}
+
+type IStorageEngine = typeof window.localStorage;
+export class DataStorage {
```
kavilla, line 20: i do like this structure. it makes it easier to not use `opensearchDashboards.` and just pass the key

src/plugins/data/public/ui/search_bar/search_bar.tsx
```diff
@@ -371,6 +373,14 @@ class SearchBarUI extends Component<SearchBarProps, State> {
         }
       }
     );
+    const dataSet = this.dataSetService.getDataSet();
```
kavilla, line 376: I think no other plugin utilizes datasets besides Discover. However, should we consider this while designing this recent query and history. Specifically the app where the query was sent. so that if the current app path is not the same app path then we do not display the query. in case one app supports different languages but another doesn't

kavilla (approved):
Awesome. Thank you and some nice clean ups. 

Just comments but nothing blocking. If potentially can add some tests where possible.

src/plugins/data/common/storage/storage.ts
abbyhu2000: fixed

src/plugins/data/public/ui/query_string_input/no_data_popover.tsx
```diff
@@ -31,8 +31,8 @@
 import { ReactElement, useEffect, useState } from 'react';
 import React from 'react';
 import { EuiButtonEmpty, EuiText, EuiTourStep } from '@elastic/eui';
-import { IStorageWrapper } from 'src/plugins/opensearch_dashboards_utils/public';
 import { i18n } from '@osd/i18n';
+import { DataStorage } from 'src/plugins/data/common';
```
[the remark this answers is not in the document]
abbyhu2000: changed to relative path

src/plugins/data/public/ui/settings/settings.ts
```diff
@@ -72,26 +71,26 @@ export class Settings {
...
   setUserQueryLanguageBlocklist(languages: string[]) {
     this.storage.set(
-      'opensearchDashboards.userQueryLanguageBlocklist',
+      'userQueryLanguageBlocklist',
       languages.map((language) => language.toLowerCase())
     );
     return true;
   }
 
   getUserQueryLanguage() {
```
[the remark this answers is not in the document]
abbyhu2000: Yea i plan to include these language related storage stuff in the language manager, and get rid of the settings class.

src/plugins/data/public/query/query_string/query_history.ts
```diff
@@ -0,0 +1,70 @@
...
+  // can listen for changes to history and update because changes to history can
+  // be triggered from different places in the app. The alternative would be to store
+  // this in state so that we hook into the React model, but it would require loading history
+  // every time the application starts even if a user is not going to view history.
+  change(listener: (reqs: any[]) => void) {
+    const subscription = this.changeEmitter.subscribe(listener);
+    return () => subscription.unsubscribe();
+  }
+
+  addQueryToHistory(dataSet: string, query: Query, dateRange?: TimeRange) {
```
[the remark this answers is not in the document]
abbyhu2000: Sure modified this prop to be simple data set.

src/plugins/data/public/query/query_string/query_history.ts
```diff
@@ -0,0 +1,70 @@
...
+      this.storage.remove(key);
+    });
+
+    const timestamp = new Date().getTime();
+    const k = 'query_' + timestamp;
+    this.storage.set(k, {
+      dataSet,
+      time: timestamp,
+      query,
+      dateRange,
```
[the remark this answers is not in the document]
abbyhu2000: so inside the `onQueryBarSubmit` function, we only get time range as props. But time filter includes both time range and refresh interval. So do you think we need to include refresh interval and the filters here?

```
public onQueryBarSubmit = (queryAndDateRange: { dateRange?: TimeRange; query?: Query }) => {
```

kavilla (approved):

sejli (approved):

PR #7745, opened 2024-08-18
[discover] update interfaces and move dataset manager

kavilla:
Updating the discover interfaces and move dataset manager into the proper location within the querystring manager.

datastructure

cached-datastructure

### DIFF
--- a/src/plugins/data/common/datasets/types.ts
+++ b/src/plugins/data/common/datasets/types.ts
@@ -138,7 +138,7 @@ export enum DATA_STRUCTURE_META_TYPES {
 /**
  * Metadata for a data structure, used for additional properties like icons or tooltips.
  */
-export interface DataStructureFeatureMeta extends DataSourceMeta {
+export interface DataStructureFeatureMeta extends DataStructureMeta {
   type: DATA_STRUCTURE_META_TYPES.FEATURE;
   icon: string;
   tooltip: string;
--- a/src/plugins/data/common/datasets/types.ts
+++ b/src/plugins/data/common/datasets/types.ts
@@ -36,7 +36,7 @@ export interface DataSourceMeta {
  * const openSearchCluster: DataStructure = {
  *   id: "b18e5f58-cf71-11ee-ad92-2468ce360004",
  *   title: "Production Cluster",
- *   type: "OPENSEARCH",
+ *   type: "DATA-SOURCE",
  *   children: [
  *     {
  *       id: "b18e5f58-cf71-11ee-ad92-2468ce360004::logs-2023.05",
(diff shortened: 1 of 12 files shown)


src/plugins/data/common/datasets/types.ts
```diff
@@ -88,12 +110,69 @@ export interface DataStructure {
...
   title: string;
-  /** The type of the data structure, registered by other classes */
+  /** The type of the data structure */
   type: string;
   /** Optional reference to the parent data structure */
   parent?: DataStructure;
   /** Optional array of child data structures */
   children?: DataStructure[];
+  /** Optional metadata for the data structure */
+  meta?: DataStructureMeta;
```
kavilla, line 120: do we think it makes sense to have the dataset just have a datastructure property instead of a datasource property? and we can construct the datasource from the datastructure?

ashwin-pc: The data source type has the connection details as well. Data structure requires 2 objects to show cluster and connection details and makes the reference bigger when you store it in the url

src/plugins/data/common/datasets/types.ts
```diff
@@ -88,12 +110,69 @@ export interface DataStructure {
...
+ */
+export enum DATA_STRUCTURE_META_TYPES {
+  FEATURE = 'FEATURE',
+  CUSTOM = 'CUSTOM',
+}
+
+/**
+ * Metadata for a data structure, used for additional properties like icons or tooltips.
+ */
+export interface DataStructureFeatureMeta extends DataSourceMeta {
```
ashwin-pc, line 141:
```suggestion
export interface DataStructureFeatureMeta extends DataStructureMeta {
```

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -100,6 +103,42 @@ export class QueryStringManager {
...
   };
+
+  /**
+   * TODO: verify if we want to just access the dataset manager directly or access each function
+   */
+  public getDatasetManager = () => {
+    return this.datasetManager;
+  };
+
+  // TODO: uncomment or use based on decision above
```
ashwin-pc, line 114: No need to re expose the same functions. It'll just add unnecessary boilerplate

ashwin-pc (approved):

src/plugins/data/common/datasets/types.ts
kavilla: Yeah that makes sense. Will keep as is.

I do like the data structure object. Thanks for the clever idea @ashwin-pc. Seems super obvious that data sources and connections are types of data structures and gives us a flexible framework.

src/plugins/data/common/datasets/types.ts
kavilla: Good catch.

I made the same mistake while making the diagrams and fixed it and just to do the same mistake in the interface

PR #7776, opened 2024-08-20
[Rebase #6730 and fix tests] Fix/UI settings overrides

ananzh:
### Description

This PR is originally from https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6730. We simply update the tests.

### From #6730 Description

There are a couple different fixes here, which can be viewed separately by looking at the separate commits. But the root cause of https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6704 was primarily the fact that the implementation of `getOverridesOrDefault` assumed the wrong object shape for the `overrides` property, when it tried to access a `value` property instead of just the value itself. Note that the existing unit test also had this same error.

Because the `ui_render_mixin` is an old legacy `.js` file, there's not much help from typescript, so I also added some enforcement to make sure we're always sending boolean values to the template.

Additionally, I realized that the validation for the `theme:version` setting was not correct, and needed updating to match our current themes.

Finally, as requested, I updated the new user theme controls to be deactivated by default.

No changelog necessary, as the [previous PR this fixes](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5652) has not yet been released.

### Future work

@AMoo-Miki @manasvinibs - The [past discussion](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/4475/files#r1251126091) and this [TODO comment](https://github.com/AMoo-Miki/OpenSearch-Dashboards/blob/31e8481fd2c0bf185d34c1140724e9320f6a676c/src/core/server/ui_settings/ui_settings_config.ts#L50-L53) seem to indicate that we don't actually want to maintain support for these overrides long term. But even if we do, I don't think the schema definition for overrides makes much sense:

```
```
We probably want users to be aware of if they specify invalid overrides and not silently ignore (which is what was happening with an incorrectly set value in @BionIT 's configuration.

### Issues Resolved
Fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6704

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/server/ui_settings/ui_settings_config.ts
+++ b/src/core/server/ui_settings/ui_settings_config.ts
@@ -58,7 +58,11 @@ const configSchema = schema.object({
   overrides: schema.object({}, { unknowns: 'allow' }),
   defaults: schema.object({
     'theme:darkMode': schema.maybe(schema.boolean({ defaultValue: false })),
-    'theme:version': schema.maybe(schema.string({ defaultValue: DEFAULT_THEME_VERSION })),
+    'theme:version': schema.maybe(
+      schema.oneOf([schema.literal('v7'), schema.literal('Next (preview)')], {
+        defaultValue: 'Next (preview)',
+      })
+    ),
   }),
 });
 
--- a/src/core/server/ui_settings/settings/theme.ts
+++ b/src/core/server/ui_settings/settings/theme.ts
@@ -47,7 +47,7 @@ export const getThemeSettings = (): Record<string, UiSettingsParams> => {
       name: i18n.translate('core.ui_settings.params.enableUserControlTitle', {
         defaultMessage: 'Enable user control',
       }),
-      value: true,
+      value: false,
       description: i18n.translate('core.ui_settings.params.enableUserControlText', {
         defaultMessage: `Enable users to control theming and dark or light mode via "Appearance" control in top navigation. When true, those settings can no longer be set globally by administrators.`,
       }),
--- a/src/core/server/ui_settings/types.ts
+++ b/src/core/server/ui_settings/types.ts
@@ -58,7 +58,7 @@ export interface IUiSettingsClient {
   /**
    * Returns the overridden uiSettings value if one exists, or the registered default if one exists {@link UiSettingsParams}
    */
-  getOverrideOrDefault: (key: string) => unknown;
+  getOverrideOrDefault: <T = unknown>(key: string) => T;
   /**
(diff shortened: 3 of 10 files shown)


ruanyl (dismissed):
Code changes look good to me, I'm just curious where to find `theme:enableUserControl` on the UI? Somehow I couldn't find it in advance settings page.

SuZhou-Joe (approved):

ananzh:
> Code changes look good to me, I'm just curious where to find `theme:enableUserControl` on the UI? Somehow I couldn't find it in advance settings page.

Not sure. If you search `theme`, will you see it?
Screenshot 2024-08-21 at 9 08 23 AM

ananzh:
@ruanyl @SuZhou-Joe I fixed the last test to make it more robust. Can you guys help me to re-approve it?

ruanyl (approved):

andy-townsend:
> Code changes look good to me, I'm just curious where to find `theme:enableUserControl` on the UI? Somehow I couldn't find it in advance settings page.

Did you manage to resolve this @ruanyl ? I'm running 2.16 but can't see this under advanced settings either.

joshuali925:
@andy-townsend the `enableUserControl` comes from a new theme feature introduced in #5652, it was reverted in 2.x by https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6978 due to some bugs and lack of test coverage, so it didn't get released in 2.16.

To release this feature, we should
1. add some integration tests
2. merge theme relevant PRs to main first: #7787, #7791
3. in 2.x undo #6978, then backport this PR and the two PRs mentioned above

AMoo-Miki:
Decided against backporting.

src/core/server/ui_settings/ui_settings_config.ts
```diff
@@ -58,7 +58,11 @@ const configSchema = schema.object({
   overrides: schema.object({}, { unknowns: 'allow' }),
   defaults: schema.object({
     'theme:darkMode': schema.maybe(schema.boolean({ defaultValue: false })),
-    'theme:version': schema.maybe(schema.string({ defaultValue: DEFAULT_THEME_VERSION })),
+    'theme:version': schema.maybe(
+      schema.oneOf([schema.literal('v7'), schema.literal('Next (preview)')], {
```
zhongnansu, line 62: what if user wants to use v8 and v9?

Issue #7689, opened 2024-08-12
[BUG] User Theme Selection doesn't retain theme selection


virajsanghvi:
**Describe the bug**

After selecting a theme from the user themes selector in the header, the theme gets applied, but the popover doesn't retain the setting. This is because browser based settings don't override persisted settings when the component gets the current value of the setting.

**To Reproduce**
1. Open theme selector in header
2. Select a theme that isn't next/v8.
3. theme will be applied, but if you reopen theme selector, next/v8 will be selected

**Expected behavior**
The theme selector shows the currently selected theme.

**OpenSearch Version**
N/A

**Dashboards Version**
OSD/main (unreleased feature)

**Plugins**

N/A

**Screenshots**

N/A

**Host/Environment (please complete the following information):**
all

**Additional context**

N/A

joshuali925:
i'm not able to reproduce this on current main
![Aug-19-2024 16-23-16](https://github.com/user-attachments/assets/9dc247ce-1803-4754-825b-5fc95e342b70)

virajsanghvi:
Turn off user based preferences, set another theme in advanced settings as an admin, and then turn on user based preferences - i think you should be able to repro.

virajsanghvi:
I think the issue is that it treats browser based settings as a default (https://github.com/opensearch-project/OpenSearch-Dashboards/blob/9a842024809218413c8000f380fd90032a90ea03/src/core/public/ui_settings/ui_settings_client.ts#L65), so if the admin has already selected a value, the admin configured value is returned when calling uiSettingsClient.get()

joshuali925:
thanks for the details, i'm able to reproduce now

PR #6978, opened 2024-06-07
Revert 815d2bd7c612450e2a15a6543c7e74558adf4a81 "[Manual Backport 2.x] Feat (core): Make theme settings user-specific and user-configurable (#5652) (#6681)"

seraphjiang:
This reverts commit https://github.com/opensearch-project/OpenSearch-Dashboards/commit/815d2bd7c612450e2a15a6543c7e74558adf4a81 on 2.x branch

We are lack of test coverage to release this feature. It should be ok to leave it in main. There is debt whether we should keep in 2.x. Since there is no clear plan for the test, the change is big. To avoid manual revert on every release (like what we did in 2.14), we'd like to revert on 2.x.

PR #7787, opened 2024-08-21
[BUG] Allow user Theme Selection retain theme selection

ananzh:
### Description

Add `browserStoredSettings` and `getWithBrowserSettings` method to allow the retrieval and resolution of values from the cache and browser settings separately. When user control is enabled, we update the browserStoredSettings without interrupting the cache (advance settings user values). 

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/7689

## Screenshot

https://github.com/user-attachments/assets/bcd0fc05-cc2a-4fb3-b11d-bd39b1db4495

## Changelog
- fix: [BUG] Allow user Theme Selection retain theme selection

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -62,7 +63,8 @@ export class UiSettingsClient implements IUiSettingsClient {
       this.cache['theme:enableUserControl']?.userValue ??
       this.cache['theme:enableUserControl']?.value
     ) {
-      this.cache = defaultsDeep(this.cache, this.getBrowserStoredSettings());
+      this.browserStoredSettings = this.getBrowserStoredSettings();
+      this.cache = defaultsDeep({}, this.defaults, this.cache, this.browserStoredSettings);
```
virajsanghvi, line 67: isn't the order of precedence left to right, so this doesn't really change any behavior, right?

it feels like we can just remove this if we're not utilizing it because there shouldn't be a case where there's a browser setting but not an advanced setting, right? Or if an admin doesn't explicitly save something, is there no userValue from advanced settings?

src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -114,6 +116,34 @@ You can use \`IUiSettingsClient.get("${key}", defaultValue)\`, which will just r
...
+    }
+
+    if (!declared) {
+      throw new Error(
+        `Unexpected \`IUiSettingsClient.get("${key}")\` call on unrecognized configuration setting "${key}".`
+      );
+    }
+
+    const type = this.cache[key].type;
+    const userValue = this.cache[key].userValue;
```
virajsanghvi, line 133: i wouldn't name variable userValue as its really the value from advanced settings, not related to the user

src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -114,6 +116,34 @@ You can use \`IUiSettingsClient.get("${key}", defaultValue)\`, which will just r
...
+    const browserValue = this.browserStoredSettings[key]?.userValue;
+    const defaultValue = defaultOverride !== undefined ? defaultOverride : this.cache[key].value;
+    
+    let value;
+    if (this.cache['theme:enableUserControl']?.userValue ?? this.cache['theme:enableUserControl']?.value) {
```
virajsanghvi, line 138: know this is a wip, but seems like this should be consumer logic

src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -114,6 +116,34 @@ You can use \`IUiSettingsClient.get("${key}", defaultValue)\`, which will just r
...
+      value = browserValue ?? userValue ?? defaultValue;
+    } else {
+      value = userValue ?? defaultValue;
```
virajsanghvi, line 141: can null be a valid value? if so, don't think we should use ??

ananzh:
This can only be backported to 2.x if https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5652, which is reverted in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6978, and https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7776 are backported.

src/core/public/ui_settings/types.ts
```diff
@@ -54,6 +54,24 @@ export interface IUiSettingsClient {
...
+   */
+  getWithBrowserSettings<T = any>(
+    key: string,
+    defaultOverride?: T
+  ): {
+    advancedSettingValue: T | undefined;
+    browserValue: T | undefined;
+    defaultValue: T;
+  };
+
```
virajsanghvi, line 74: callout: I think this is fine for now to avoid increasing scope - i think in the future, we should consider making get() aware of all contexts, and having experiences like advanced settings explicitly call another method or a variant to exclude browser settings in resolution

two suggestions: 
- should we call this getValues() or something more generic where this can provide all levels of values for this setting (not just browser as it also returns default value today)? 
- should this include a resolvedValue or a value that handles what the current active value is so consumers don't have to resolve this themselves?

src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -114,6 +115,43 @@ You can use \`IUiSettingsClient.get("${key}", defaultValue)\`, which will just r
...
+    const type = this.cache[key]?.type;
+    const advancedSettingValue = this.cache[key]?.userValue;
+    const browserValue = this.browserStoredSettings[key]?.userValue;
+    const defaultValue = defaultOverride !== undefined ? defaultOverride : this.cache[key]?.value;
+
+    // Resolve the value based on the type
+    const resolveValue = (val: any): T => this.resolveValue(val, type);
+
+    const resolvedAdvancedValue =
+      advancedSettingValue !== undefined ? resolveValue(advancedSettingValue) : undefined;
```
virajsanghvi, line 143: nit: could create a helper function to handle this undefined case and inline these with return

src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -230,8 +268,15 @@ You can use \`IUiSettingsClient.get("${key}", defaultValue)\`, which will just r
 
     const declared = this.isDeclared(key);
     const defaults = this.defaults;
-
-    const oldVal = declared ? this.cache[key].userValue : undefined;
+    const userControl =
+      this.cache['theme:enableUserControl']?.userValue ??
```
virajsanghvi, line 272: ui_settings_client shouldn't be aware of a theme specific setting right?

also, could other settings be null? if so, this probably should be checking that these aren't defined.

src/plugins/advanced_settings/public/header_user_theme_menu.tsx
```diff
@@ -82,6 +93,10 @@ export const HeaderUserThemeMenu = () => {
 
   const onAppearanceSubmit = async (e: SyntheticEvent) => {
     const actions = [setThemeVersion(themeOptions.find((t) => theme === t.value)?.value ?? '')];
+    const result = uiSettings.getWithBrowserSettings<string>('theme:darkMode');
+    const currentDarkMode = enableUserControl
+      ? result.browserValue ?? result.advancedSettingValue ?? result.defaultValue
+      : result.advancedSettingValue ?? result.defaultValue;
```
virajsanghvi, line 99: i don't think consumers should be resolving values - think this should come from uisettingsclient

angle943:
This is Kiro posting.

No activity in 16+ months and has merge conflicts. Recommend closing; if the bug still reproduces on main, a fresh PR will be easier to review.

PR #7791, opened 2024-08-21
Support overriding theme through URL search parameter

joshuali925:
### Description

This PR allows theme to be overridden by search parameter.

- If `theme` is `dark`, OSD sets theme to dark. 
- If `theme` is `browser`, OSD sets theme to match browser. 
- If `theme` is `light`, OSD sets theme to light. 
- If `theme` is absent or any other value, OSD behavior does not change.

Theme can be a search param in either `location.search` or `location.hash`. For example

- `http://localhost:5601/app/home#/?theme=dark`
- `http://localhost:5601/app/management?theme=light`
- `http://localhost:5601/app/home?theme=browser#/?theme=dark`
  - it will prioritize the value in search param over hash, so theme will be browser

This override will only be in-effect if the parameter is present in the URL. It only changes the actual theme and does not change theme selection on UI. This is more for the `embed`ed use case and not a user facing feature.

This PR depends on #5652. The handlebars files don't seem to have tests, will add tests for 5652 in general later.

## Changelog
- feat: support overriding theme through URL search parameter

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/legacy/ui/ui_render/bootstrap/startup.js.hbs
+++ b/src/legacy/ui/ui_render/bootstrap/startup.js.hbs
@@ -14,12 +14,7 @@ try {
   if (!urlThemeParam && window.location.hash.includes('?')) {
     urlThemeParam = new URLSearchParams(window.location.hash.split('?')[1]).get('theme');
   }
-  if (
-    urlThemeParam !== null &&
-    urlThemeParam !== 'dark' &&
-    urlThemeParam !== 'light' &&
-    urlThemeParam !== 'browser'
-  ) {
+  if (![null, 'dark', 'light', 'browser'].includes(urlThemeParam)) {
     throw new Error('theme must be "browser", "light", or "dark", received ' + urlThemeParam);
     urlThemeParam = null;
   }
--- a/src/legacy/ui/ui_render/bootstrap/startup.js.hbs
+++ b/src/legacy/ui/ui_render/bootstrap/startup.js.hbs
@@ -7,16 +7,41 @@ try {
   window.console.log(error);
 }
 
-var useBrowserColorScheme = window.localStorage.getItem('useBrowserColorScheme') === 'true' || false;
+/** @type {'browser' | 'light' | 'dark' | null} */
+var urlThemeParam = null;
+try {
+  urlThemeParam = new URLSearchParams(window.location.search).get('theme');
+  if (!urlThemeParam && window.location.hash.includes('?')) {
+    urlThemeParam = new URLSearchParams(window.location.hash.split('?')[1]).get('theme');
+  }
+  if (
+    urlThemeParam !== null &&
+    urlThemeParam !== 'dark' &&
+    urlThemeParam !== 'light' &&
+    urlThemeParam !== 'browser'
+  ) {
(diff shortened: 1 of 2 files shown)


AMoo-Miki:
Does it matter that #5652 is not in 2.x?

joshuali925:
> Does it matter that #5652 is not in 2.x?

we can still backport this PR to 2.x but it will be dead code due to https://github.com/opensearch-project/OpenSearch-Dashboards/pull/6978. we do want to revert #6978 for 2.17 so this gets out, but we might disable the UI part

src/legacy/ui/ui_render/bootstrap/startup.js.hbs
```diff
@@ -7,16 +7,41 @@ try {
   window.console.log(error);
 }
 
-var useBrowserColorScheme = window.localStorage.getItem('useBrowserColorScheme') === 'true' || false;
+/** @type {'browser' | 'light' | 'dark' | null} */
+var urlThemeParam = null;
+try {
+  urlThemeParam = new URLSearchParams(window.location.search).get('theme');
```
virajsanghvi, line 13: if this parameter only impacts dark/light mode, can we name it after that and not use `theme`?

src/legacy/ui/ui_render/bootstrap/startup.js.hbs
```diff
@@ -7,16 +7,41 @@ try {
...
+  if (!urlThemeParam && window.location.hash.includes('?')) {
+    urlThemeParam = new URLSearchParams(window.location.hash.split('?')[1]).get('theme');
```
virajsanghvi, line 15: we specify urls params in the url fragment?

src/legacy/ui/ui_render/bootstrap/startup.js.hbs
```diff
@@ -7,16 +7,41 @@ try {
...
+  }
+  if (
+    urlThemeParam !== null &&
+    urlThemeParam !== 'dark' &&
+    urlThemeParam !== 'light' &&
+    urlThemeParam !== 'browser'
```
virajsanghvi, line 21: nit: could use [...].includes

src/legacy/ui/ui_render/bootstrap/startup.js.hbs
joshuali925: yes it's like `'colorscheme'` vs `'background'`, and "theme" only impacts background. i thought about something like darkMode, but it also needs to support "browser". do you have any recommendation?

src/legacy/ui/ui_render/bootstrap/startup.js.hbs
AMoo-Miki: In a future enhancement, we MIGHT be having an interceptor on the server side to handle language and theme changes among others. After intercepting, it will clean the URL and redirect to the destination without the lang and theme. If we look into the hash here, we cannot unlook at it later and that would make the transition challenging if not impossble without a breaking change.

IMO, we should drop this piece and stick to the one above.

joshuali925:
this is no longer needed

Issue #7786, opened 2024-08-21
[Discover] Add registrable DiscoverAction to Discover page


gaobinlong:
**Is your feature request related to a problem? Please describe.**
As part of the olly project, the functionalities of `generate anomaly detector` and `text to visualization` use LLM to help users to create an anomaly detector and generate visualizations based on natural language, we want to add the entry points of these two functionality to the discover page as follows:
<img width="1310" alt="image" src="https://github.com/user-attachments/assets/20c868fa-d6bf-4666-a185-8ce9a6ad939e">
, so in order to implement this, we need to add a registrable DiscoverActions component to `DiscoverCanvas`, so that other plugins can register their own actions, and then display the actions in the Discover page.

**Describe the solution you'd like**

Add a registrable DiscoverActions component to `DiscoverCanvas`.

**Describe alternatives you've considered**

No other alternatives.

**Additional context**

The UI mockup:
<img width="831" alt="image" src="https://github.com/user-attachments/assets/cbb49a06-fd71-4a09-a24a-549e8ddc2b41">

A demo video:
https://github.com/user-attachments/assets/cd7b3f12-526f-480a-83b2-3623f92ca803

PR #7793, opened 2024-08-22
Add registrable DiscoverActions to discover page

gaobinlong:
### Description

This PR adds a registrable DiscoverActions component to DiscoverCanvas in the discover page, so that other plugins can register their own actions, and then display the actions in the Discover page.

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/7786

## Screenshot

image

## Testing the changes

If no actions are registered, then no impact on the discover page.
All of the registered actions will be associated with their own feature flags, if users don't enable the feature flags, then no actions will be shown in the discover page.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_explorer/public/plugin.ts
+++ b/src/plugins/data_explorer/public/plugin.ts
@@ -20,6 +20,7 @@ import {
   DataExplorerPluginStart,
   DataExplorerPluginStartDependencies,
   DataExplorerServices,
+  DiscoverAction,
 } from './types';
 import { PLUGIN_ID, PLUGIN_NAME } from '../common';
 import { ViewService } from './services/view_service';
@@ -43,11 +44,12 @@ export class DataExplorerPlugin
   private appStateUpdater = new BehaviorSubject<AppUpdater>(() => ({}));
   private stopUrlTracking?: () => void;
   private currentHistory?: ScopedHistory;
+  private discoverActions: DiscoverAction[] = [];
 
   public setup(
     core: CoreSetup<DataExplorerPluginStartDependencies, DataExplorerPluginStart>,
     { data }: DataExplorerPluginSetupDependencies
-  ): DataExplorerPluginSetup {
+  ) {
     const viewService = this.viewService;
 
     const { appMounted, appUnMounted, stop: stopUrlTracker } = createOsdUrlTracker({
@@ -105,6 +107,7 @@ export class DataExplorerPlugin
             ...withNotifyOnErrors(coreStart.notifications.toasts),
           }),
           viewRegistry: viewService.start(),
+          discoverActions: this.discoverActions,
         };
 
         // Get start services as specified in opensearch_dashboards.json
@@ -124,7 +127,10 @@ export class DataExplorerPlugin
     });
 
     return {
-      ...this.viewService.setup(),
+      registerView: this.viewService.setup().registerView,
+      registerDiscoverAction: (discoverAction: DiscoverAction) => {
+        this.discoverActions.push(discoverAction);
+      },
     };
   }
 
--- a/src/plugins/data_explorer/public/plugin.ts
+++ b/src/plugins/data_explorer/public/plugin.ts
@@ -49,7 +49,7 @@ export class DataExplorerPlugin
   public setup(
     core: CoreSetup<DataExplorerPluginStartDependencies, DataExplorerPluginStart>,
     { data }: DataExplorerPluginSetupDependencies
-  ) {
+  ): DataExplorerPluginSetup {
     const viewService = this.viewService;
 
     const { appMounted, appUnMounted, stop: stopUrlTracker } = createOsdUrlTracker({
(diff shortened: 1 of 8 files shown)


src/plugins/data_explorer/public/plugin.ts
```diff
@@ -43,11 +44,12 @@ export class DataExplorerPlugin
   private appStateUpdater = new BehaviorSubject<AppUpdater>(() => ({}));
   private stopUrlTracking?: () => void;
   private currentHistory?: ScopedHistory;
+  private discoverActions: DiscoverAction[] = [];
 
   public setup(
     core: CoreSetup<DataExplorerPluginStartDependencies, DataExplorerPluginStart>,
     { data }: DataExplorerPluginSetupDependencies
-  ): DataExplorerPluginSetup {
+  ) {
```
SuZhou-Joe, line 52: Seems unnecessary change.

src/plugins/data_explorer/public/plugin.ts
```diff
@@ -105,6 +107,7 @@ export class DataExplorerPlugin
             ...withNotifyOnErrors(coreStart.notifications.toasts),
           }),
           viewRegistry: viewService.start(),
+          discoverActions: this.discoverActions,
```
SuZhou-Joe, line 110: Nit: maybe `actions` is good enough as we are already under the context of discover.

SuZhou-Joe:
Do not know if we have taken a look on the existing features of ui_actions_service.

src/plugins/data_explorer/public/plugin.ts
gaobinlong: Changed that, thanks!

gaobinlong:
> Do not know if we have taken a look on the existing features of ui_actions_service.
> 
> https://github.com/opensearch-project/OpenSearch-> Dashboards/blob/cb273364086057b542a55d2bbe28631313a22df9/src/plugins/ui_actions/README.m

Yeah, we have considered that solution but it doesn't meet our requirement, ui actions only supports a menu button with multiple actions, the style is different from the UI mockup for `generate anomaly detector` and `text to visualization` which flatten all of the actions in the page, so we cannot use ui actions directly.
image

src/plugins/discover/public/application/view_components/canvas/discover_canvas.scss
```diff
@@ -12,6 +12,12 @@
     position: relative;
   }
 
+  &_actions {
+    position: absolute;
+    top: 0;
+    right: 30px;
```
AMoo-Miki, line 18: How did we land at 30px? We should use $euiSizeL or XL.

AMoo-Miki:
> Yeah, we have considered that solution but it doesn't meet our requirement, ui actions only supports a menu button with multiple actions, the style is different from the UI mockup for generate anomaly detector and text to visualization which flatten all of the actions in the page, so we cannot use ui actions directly.

If i understand correctly, all you need is a different appearance. What are your requirements specifically that are not supported by the existing system?

src/plugins/discover/public/application/view_components/canvas/discover_canvas.scss
gaobinlong: Changed it to $euiSizeL, thanks!

gaobinlong:
> > Yeah, we have considered that solution but it doesn't meet our requirement, ui actions only supports a menu button with multiple actions, the style is different from the UI mockup for generate anomaly detector and text to visualization which flatten all of the actions in the page, so we cannot use ui actions directly.
> 
> If i understand correctly, all you need is a different appearance. What are your requirements specifically that are not supported by the existing system?

I think that's because UI actions needs all registered actions behind a clickable element, when clicking that element, all actions are shown, something like a popover,  this is different from our requirement that no clickable entry point exists, but show all the registered actions in the page directly. Maybe we can change the existing UI actions to make it support our case, but the effort is much more than the implementation in this PR.

kavilla:
this seems a little bit of an anti pattern.

Discover has a set of registered actions that it can do into the top nav bar
Screenshot 2024-05-22 at 2 32 16 PM

the other thing is that the histogram is apart of the discover plugin if and only if there is date histogram aggregation which only occurs if timestamps exist. Logically it makes sense for AD that if no timestamp then the action doesn't really matter since I doubt it is possible to do AD without timestamps but for any other type of action this will will not be preset if no timestamp field present.

i think if we really wanted something we could thing bigger and apply it with the query editor component. if you run `yarn start:enhancements` you will get the future iteration of the component.

@joshuali925 has made it really easy to portal in a component from the AD dashboards plugin to the query editor header bar.

joshuali925:
yeah I'm not sure if anomaly detector button should go into the query editor extensions UI, but from user perspective i feel it's a bit weird to have AD next to histogram, since actions are generally on top. It might break the experience if some plugins put actions on top and some put them below

dagneyb:
My two cents: I dont think this action belongs with the top nav "high-level" actions. The call to actions we are putting in line are truly intended to be in in-line with the associated data. We shouldn't let our current architecture drive our design. 

I'm not opposed to putting these in the query editor component as @kavilla called out, but I think we may be up against the deadline to make that change for 2.17.

ruanyl:
Closing as this is not needed

PR #7805, opened 2024-08-22
[Workspace] Refactor the style of recent items card in the header

Qxisylolo:
### Description
In this pr, I refactor the style of recent items card in the header by
1. Enable maxHeight; if exceeded, the menu will become scrollable.
2. Display associated icons.
3. The workspace name will be attached to the asset.
4. Allow setting preferences for the number of recent items displayed.

## Screenshot
https://github.com/user-attachments/assets/dd8a364d-abd2-4274-a647-cd9bc6942e44

empty state
Screenshot 2024-08-22 at 17 40 35

## Changelog
- feat: refractor the style of recent items card

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/public/chrome/chrome_service.tsx
+++ b/src/core/public/chrome/chrome_service.tsx
@@ -43,7 +43,6 @@ import {
 } from 'rxjs';
 import { flatMap, map, takeUntil } from 'rxjs/operators';
 import { EuiLink } from '@elastic/eui';
-import { CoreStart } from 'opensearch-dashboards/public';
 import { mountReactNode } from '../utils/mount';
 import { InternalApplicationStart } from '../application';
 import { DocLinksStart } from '../doc_links';
@@ -116,7 +115,6 @@ export interface StartDeps {
   application: InternalApplicationStart;
   docLinks: DocLinksStart;
   http: HttpStart;
-  // core: CoreStart;
   injectedMetadata: InjectedMetadataStart;
   notifications: NotificationsStart;
   uiSettings: IUiSettingsClient;
--- a/src/core/public/chrome/chrome_service.tsx
+++ b/src/core/public/chrome/chrome_service.tsx
@@ -43,6 +43,7 @@ import {
 } from 'rxjs';
 import { flatMap, map, takeUntil } from 'rxjs/operators';
 import { EuiLink } from '@elastic/eui';
+import { CoreStart } from 'opensearch-dashboards/public';
 import { mountReactNode } from '../utils/mount';
 import { InternalApplicationStart } from '../application';
 import { DocLinksStart } from '../doc_links';
@@ -115,6 +116,7 @@ export interface StartDeps {
   application: InternalApplicationStart;
   docLinks: DocLinksStart;
   http: HttpStart;
+  // core: CoreStart;
   injectedMetadata: InjectedMetadataStart;
   notifications: NotificationsStart;
   uiSettings: IUiSettingsClient;
@@ -349,6 +351,7 @@ export class ChromeService {
 
       getHeaderComponent: () => (
         <Header
+          http={http}
           loadingCount$={http.getLoadingCount$()}
           application={application}
           appTitle$={appTitle$.pipe(takeUntil(this.stop$))}
--- a/src/core/public/chrome/ui/header/header.tsx
+++ b/src/core/public/chrome/ui/header/header.tsx
@@ -79,6 +79,7 @@ import { HomeLoader } from './home_loader';
 import { RecentItems } from './recent_items';
 
 export interface HeaderProps {
+  http: HttpStart;
   opensearchDashboardsVersion: string;
   application: InternalApplicationStart;
   appTitle$: Observable<string>;
@@ -120,6 +121,7 @@ export interface HeaderProps {
 }
 
 export function Header({
+  http,
   opensearchDashboardsVersion,
   opensearchDashboardsDocLink,
   application,
@@ -351,11 +353,12 @@ export function Header({
   const renderRecentItems = () => (
     <EuiHeaderSectionItem border={useUpdatedHeader ? 'none' : 'right'}>
       <RecentItems
+        http={http}
         recentlyAccessed$={observables.recentlyAccessed$}
         workspaceList$={observables.workspaceList$}
         navigateToUrl={application.navigateToUrl}
-        navLinks$={observables.navLinks$}
-        basePath={basePath}
+        // navLinks$={observables.navLinks$}
+        // basePath={basePath}
         renderBreadcrumbs={renderBreadcrumbs(true)}
         buttonSize={useApplicationHeader ? 's' : 'xs'}
       />
--- a/src/core/public/chrome/ui/header/header.tsx
+++ b/src/core/public/chrome/ui/header/header.tsx
@@ -357,8 +357,6 @@ export function Header({
         recentlyAccessed$={observables.recentlyAccessed$}
         workspaceList$={observables.workspaceList$}
         navigateToUrl={application.navigateToUrl}
-        // navLinks$={observables.navLinks$}
-        // basePath={basePath}
         renderBreadcrumbs={renderBreadcrumbs(true)}
         buttonSize={useApplicationHeader ? 's' : 'xs'}
       />
--- a/src/core/public/chrome/ui/header/header.tsx
+++ b/src/core/public/chrome/ui/header/header.tsx
@@ -354,6 +354,8 @@ export function Header({
     <EuiHeaderSectionItem border={useUpdatedHeader ? 'none' : 'right'}>
       <RecentItems
         http={http}
+        navLinks$={observables.navLinks$}
+        basePath={basePath}
         recentlyAccessed$={observables.recentlyAccessed$}
         workspaceList$={observables.workspaceList$}
         navigateToUrl={application.navigateToUrl}
--- a/src/core/public/chrome/ui/header/recent_items.tsx
+++ b/src/core/public/chrome/ui/header/recent_items.tsx
@@ -28,6 +28,8 @@ import { SavedObjectsNamespaceType } from 'src/core/public';
 import { ChromeRecentlyAccessedHistoryItem, SavedObject } from 'opensearch-dashboards/public';
 import { WorkspaceObject } from '../../../workspace';
 import { HttpStart } from '../../../http';
+import { createRecentNavLink } from './nav_link';
+import { ChromeNavLink } from '../../../';
 import './recent_items.scss';
 
 const widthForRightMargin = 8;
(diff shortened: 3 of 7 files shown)


src/core/public/chrome/chrome_service.tsx
```diff
@@ -43,6 +43,7 @@ import {
 } from 'rxjs';
 import { flatMap, map, takeUntil } from 'rxjs/operators';
 import { EuiLink } from '@elastic/eui';
+import { CoreStart } from 'opensearch-dashboards/public';
```
raintygao, line 46: We could remove if we don't need this.

src/core/public/chrome/chrome_service.tsx
```diff
@@ -115,6 +116,7 @@ export interface StartDeps {
   application: InternalApplicationStart;
   docLinks: DocLinksStart;
   http: HttpStart;
+  // core: CoreStart;
```
raintygao, line 119: Better not contain annotation in component file.

src/core/public/chrome/ui/header/recent_items.tsx
```diff
@@ -2,76 +2,129 @@
...
+};
+
+type DetailedRecentlyAccessedItem = ChromeRecentlyAccessedHistoryItem &
+  SavedObjectWithMetadata &
+  ChromeRecentlyAccessedHistoryItem['meta'] & {
+    updatedAt: number;
+    workspaceName?: string;
+  };
+
+export const bulkGetDetail = (
```
raintygao, line 62: What about move this into a utils file and rename to some others like bulkGetSavedObjectsDetail, which could better indicate the effect of this function. And we don't need to export if nowhere imports this.

src/core/public/chrome/ui/header/recent_items.tsx
```diff
@@ -2,76 +2,129 @@
...
-  basePath,
   renderBreadcrumbs,
   buttonSize = 's',
+  http,
 }: Props) => {
   const [isPopoverOpen, setIsPopoverOpen] = useState(false);
-
+  const [isPreferencesPopoverOpen, setIsPreferencesPopoverOpen] = useState(false);
   const recentlyAccessedItems = useObservable(recentlyAccessed$, []);
+  const [recentsRadioIdSelected, setRecentsRadioIdSelected] = useState('recent_5');
```
raintygao, line 113: What about `useState(recentsRadios[0].id) since recentsRadios is already a constant varible.

src/core/public/chrome/ui/header/recent_items.tsx
```diff
@@ -94,46 +147,132 @@ export const RecentItems = ({
...
-              }
-              color="text"
-              size="s"
-            />
-          ))}
-        </EuiListGroup>
-      ) : (
-        <EuiText color="subdued">No recently viewed items</EuiText>
-      )}
+        <EuiPopover
```
raintygao, line 239: Nit: we could extract preference popover into a new component, to make two different components isolate.

src/core/public/chrome/ui/header/recent_items.tsx
```diff
@@ -94,46 +147,132 @@ export const RecentItems = ({
...
+            workspaceName: findWorkspace?.name,
+          };
+        });
+        setDetailedSavedObjects(formatDetailedSavedObjects);
+      });
+    }
+  }, [recentlyAccessedItems, http, workspaceList]);
+
+  const selectedRecentsItems = useMemo(() => {
+    return detailedSavedObjects.slice(0, Number(recentsRadioIdSelected.split('recent_')[1]));
```
raintygao, line 183: Nit: We could add an attribute in recentsRadios to store actual value and find instead of doing string split next time.

src/core/public/chrome/ui/header/recent_items.tsx
```diff
@@ -94,46 +147,132 @@ export const RecentItems = ({
...
+            updatedAt: moment(obj?.updated_at).valueOf(),
+            workspaceName: findWorkspace?.name,
+          };
+        });
+        setDetailedSavedObjects(formatDetailedSavedObjects);
+      });
+    }
+  }, [recentlyAccessedItems, http, workspaceList]);
+
+  const selectedRecentsItems = useMemo(() => {
```
raintygao, line 182: Nit: What about adding a line to make it more clear.
`
const displayAmount = Number(recentsRadioIdSelected.split('recent_')[1]);
`

src/core/public/chrome/ui/header/recent_items.tsx
Qxisylolo: Thanks for the review, I delete export and I think it's better to keep it here as the required type SavedObjectWithMetadata has the import restriction. Here I just simply duplicate it in the file. [ Unexpected path "src/plugins/saved_objects_management/common/types" imported in restricted zone.]

src/core/public/chrome/chrome_service.tsx
Qxisylolo: Thank you, updated.

src/core/public/chrome/chrome_service.tsx
Qxisylolo: Thank you, updated.

src/core/public/chrome/ui/header/recent_items.tsx
Qxisylolo: Thank you, updated.

src/core/public/chrome/ui/header/recent_items.tsx
Qxisylolo: Thank you, updated.

src/core/public/chrome/ui/header/recent_items.tsx
Qxisylolo: Thank you, updated.

src/core/public/chrome/ui/header/recent_items.tsx
Qxisylolo: Thank you, updated.

src/core/public/chrome/ui/header/header.tsx
```diff
@@ -351,11 +353,12 @@ export function Header({
...
     <EuiHeaderSectionItem border={useUpdatedHeader ? 'none' : 'right'}>
       <RecentItems
+        http={http}
         recentlyAccessed$={observables.recentlyAccessed$}
         workspaceList$={observables.workspaceList$}
         navigateToUrl={application.navigateToUrl}
-        navLinks$={observables.navLinks$}
-        basePath={basePath}
+        // navLinks$={observables.navLinks$}
+        // basePath={basePath}
```
SuZhou-Joe, line 361: Please delete these 2 lines if they are useless.

src/core/public/chrome/ui/header/recent_items.tsx
```diff
@@ -2,76 +2,164 @@
...
+  },
+  {
+    id: '10',
+    label: '10 pages',
+  },
+  {
+    id: '15',
+    label: '15 pages',
+  },
+];
```
SuZhou-Joe, line 97:
```suggestion
const recentsRadios = [
  {
    id: 5,
    label: '5 pages',
  },
  {
    id: 10,
    label: '10 pages',
  },
  {
    id: 15,
    label: '15 pages',
  },
];
```

Nit: number seems to make more sense to me.

src/core/public/chrome/ui/header/recent_items.tsx
```diff
@@ -94,46 +182,98 @@ export const RecentItems = ({
...
+  useEffect(() => {
+    const savedObjects = recentlyAccessedItems
+      .filter((item) => item.meta?.type)
+      .map((item) => ({
+        type: item.meta?.type || '',
+        id: item.id,
+      }));
+
+    if (savedObjects.length) {
+      bulkGetDetail(savedObjects, http).then((res) => {
```
ruanyl, line 194: Why should it call this API to get the details? Isn't items from `recentlyAccessed$` already have saved object title, type and workspace?

src/core/public/chrome/ui/header/recent_items.tsx
```diff
@@ -2,76 +2,164 @@
...
   const recentlyAccessedItems = useObservable(recentlyAccessed$, []);
+  const [recentsRadioIdSelected, setRecentsRadioIdSelected] = useState(recentsRadios[0].id);
   const workspaceList = useObservable(workspaceList$, []);
-  const navLinks = useObservable(navLinks$, []).filter((link) => !link.hidden);
-
-  const items = useMemo(() => {
-    // Only display five most recent items
-    return recentlyAccessedItems.slice(0, 5).map((item) => {
-      return {
-        link: createRecentNavLink(item, navLinks, basePath, navigateToUrl).href,
```
SuZhou-Joe, line 57: Seems we removed the createRecentNavLink import, but the link should be generated from it.

src/core/public/chrome/ui/header/recent_items.tsx
Qxisylolo: Hi, Thanks for your review, we call the API to get the associated icon

src/core/public/chrome/ui/header/recent_items.tsx
Qxisylolo: Hi, I delete it since we could directly access link from recentlyAccessedItems

src/core/public/chrome/ui/header/recent_items.tsx
SuZhou-Joe: Thanks, get it.

src/core/public/chrome/ui/header/recent_items.tsx
SuZhou-Joe: But that link is not correct as it does not contain workspace info.

Qxisylolo: ok,I see, I will correct this. Thank you

src/core/public/chrome/ui/header/recent_items.tsx
```diff
@@ -33,7 +42,63 @@ export interface Props {
...
+    updatedAt: number;
+    workspaceName?: string;
+  };
+
+const bulkGetDetail = (savedObjects: Array<Pick<SavedObject, 'type' | 'id'>>, http: HttpStart) => {
+  return Promise.all(
+    savedObjects.map((obj) =>
+      http
+        .get<SavedObjectWithMetadata>(
+          `/api/opensearch-dashboards/management/saved_objects/${encodeURIComponent(
```
ruanyl, line 71: I'm not fully "comfortable" with calling a plugin API from a core module though the API is exposed by a core plugin. 

But I've seen the saved object management API path `/management/opensearch-dashboards` has already been used in multiple places, so I'm fine with it now.

In my opinion, perhaps this recent visited button should be owned by saved object management plugin because it's all about recently visited saved objects, I don't see a strong correlation between chrome module and the recently visited items, to my understanding, it's just another registration point that saved objects management plugin contributes to.

cc @SuZhou-Joe @AMoo-Miki

ruanyl (approved):

src/core/public/chrome/ui/header/recent_items.tsx
SuZhou-Joe: Agree, things will look more in place if we can `register` the recent items to the left of the breadcrumbs within saved objects management plugin.

PR #7808, opened 2024-08-22
[discover] query data manager

kavilla:
Open for review while currently working on:
- ~~Handle conflicts~~
- Cleaning up the dataset logic i added that was unneeded for this specific PR
- Indices are being treated as leaf data structures properly so they dont go to time stamp
- Index patterns let you select the timestamp field even though it should just close automatically
- Filter out categories that not in the supported flat list (we are focusing on ensuring on the flat list we see index patterns for this initial PR)
- Add S3
- Revisits some constants, I might have gone too far

### Description

Query service > Query string manager > data set manager for managing data structures and datasets.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/public/ui/dataset_selector/__mocks__/utils.ts
ashwin-pc, line 1: Looks like you didnt modify how this mock works. Use this as a reference on how to implement the actual dataset manager response. Specifically how the handlers work w.r.t returning the options. If you follow how the path and options work here, the dataset seletor should just wrk out of the box. If you think we need to do this differently, let me know.

src/plugins/data/public/ui/search_bar/lib/use_dataset_manager.ts
```diff
@@ -6,34 +6,34 @@
...
-  dataSetManager: DataSetContract;
+interface UseDatasetManagerProps {
+  dataset?: Dataset;
+  datasetManager: DatasetContract;
 }
 
-export const useDataSetManager = (props: UseDataSetManagerProps) => {
-  const [dataSet, setDataSet] = useState<Dataset | undefined>(
-    props.dataSet || props.dataSetManager.getDataSet()
+export const useDatasetManager = (props: UseDatasetManagerProps) => {
```
ashwin-pc, line 16:
```suggestion
export const useDataset = (props: UseDatasetManagerProps) => {
```

This is more accurate since thats how react hooks are usually named

src/plugins/discover/public/application/view_components/utils/use_dataset_manager.ts
ashwin-pc, line 1: Nit: Looks like we already have this in `src/plugins/data/public/ui/search_bar/lib/use_dataset_manager.ts`Cant we expose it from the data plugin so that any plugin that wants to use datasets can use it easily instead of having to define their own hook

src/plugins/discover/public/application/view_components/utils/use_dataset_manager.ts
```diff
@@ -5,35 +5,35 @@
...
-  dataSetManager: DataSetManager;
+interface UseDatasetManagerProps {
+  dataset?: Dataset;
+  datasetManager: DatasetContract;
 }
 
-export const useDataSetManager = (props: UseDataSetManagerProps) => {
-  const [dataSet, setDataSet] = useState<Dataset | undefined>(
-    props.dataSet || props.dataSetManager.getDataSet()
+export const useDatasetManager = (props: UseDatasetManagerProps) => {
```
ashwin-pc, line 16: If we want to keep this instead of using the one from the data plugin, lets rename this hook too
```suggestion
export const useDataset = (props: UseDatasetManagerProps) => {
```

src/plugins/query_enhancements/public/query_assist/utils/create_extension.tsx
```diff
@@ -25,27 +25,30 @@ const getAvailableLanguages$ = (
...
+  data.query.queryString
+    .getDatasetManager()
+    .getUpdates$()
+    .pipe(
+      startWith(data.query.queryString.getDatasetManager().getDataset()),
+      distinctUntilChanged(),
+      switchMap(async (dataset) => {
+        // currently query assist tool relies on opensearch API to get index
+        // mappings, external data source types (e.g. s3) are not supported
+        if (dataset?.dataSource?.type !== DEFAULT_QUERY.DATASET.DATASOURCE.TYPE) return [];
```
ashwin-pc, line 37: Not for this PR, but we need to update this to function to  simply use the query object. Query assist should also work with indices right? This change wont work for that

src/plugins/query_enhancements/public/query_assist/utils/create_extension.tsx
```diff
@@ -25,27 +25,30 @@ const getAvailableLanguages$ = (
...
-        .then((response) => response.configuredLanguages)
-        .catch(() => []);
-      availableLanguagesByDataSource.set(dataSourceId, languages);
-      return languages;
-    })
-  );
+        const dataSourceId = dataset?.dataSource?.id;
+        const cached = availableLanguagesByDataSource.get(dataSourceId);
+        if (cached !== undefined) return cached;
+        const languages = await http
```
ashwin-pc, line 42: Also not for this PR, but why are we making a HTTP call for this? Doesnt it simply need to check for `PPL` and whether its an `index` or `index-pattern`? All of which should be available in the the query object

ashwin-pc (commented):
Damm, I was expecting to have a lot more comments than this. This is great. I think besides how we handle datastructres in the dataset manager, there are just smaller nits and changes. Lets chat about the dataset manager.

PR #7817, opened 2024-08-23
[Workspace]Fix maximum call stack error in use case service

wanglam:
### Description

This PR is for fixing maximum call stack error in use case service, separate a function to compare features using `sort` instead of using `Array.property.some`. 
The root cause of this bug should be error features compare result see below example:
```
// use case 1
{
    features:[{id:"1"},{id:"2"}]
}
// use case 2
{
    features:[{id:"1"},{id:"2"}]
}
```
The result should be `true` which means two use case should be equal. But the old compare logic will return `false`. We should change `b.features.some` to `b.features.every`, then the result would be correct. Using `some` and `every` to implement the compare logic would be a little bit complicated, it can't handle duplicate features. So we change to use `sort` to compare the use case.

### Issues Resolved

#7820 

## Screenshot

No UI changes

## Testing the changes

- Clone branch code and run `yarn osd bootstrap --single-version ignore`
```
workspace.enabled: true
uiSettings:
  overrides:
    'home:useNewHomePage': true
```
- Login and visit workspace detail page
- Open devtools and should no `Maximum call stack error` in the console

## Changelog

- fix: [Workspace] maximum call stack error in use case service

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/public/utils.ts
+++ b/src/plugins/workspace/public/utils.ts
@@ -24,7 +24,7 @@ import {
   WorkspaceAvailability,
 } from '../../../core/public';
 import { DEFAULT_SELECTED_FEATURES_IDS, WORKSPACE_DETAIL_APP_ID } from '../common/constants';
-import { WorkspaceUseCase } from './types';
+import { WorkspaceUseCase, WorkspaceUseCaseFeature } from './types';
 import { formatUrlWithWorkspaceId } from '../../../core/public/utils';
 import { SigV4ServiceName } from '../../../plugins/data_source/common/data_sources';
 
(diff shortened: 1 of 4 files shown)


src/plugins/workspace/public/utils.ts
```diff
@@ -260,6 +260,24 @@ export const convertNavGroupToWorkspaceUseCase = ({
...
+    return false;
+  }
+  for (const feature of features1Set) {
+    if (!features2Set.has(feature)) {
+      return false;
+    }
+  }
+  return true;
+};
+
```
SuZhou-Joe, line 280: Perhaps we can simplify the compare by sorting the serialized array and see if JSON.stringify(sortArray) equals?

ruanyl (approved):

SuZhou-Joe (approved):

PR #7845, opened 2024-08-26
[WIP] Working on aggs for PPL

kavilla:
### Description

Need to update the dataframe logic to just take the type and parse it.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/dataset_service/lib/index_pattern_type.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/lib/index_pattern_type.ts
@@ -12,6 +12,7 @@ import {
   Dataset,
   IIndexPattern,
   DATA_STRUCTURE_META_TYPES,
+  DataStructureCustomMeta,
 } from '../../../../../common';
 import { DatasetTypeConfig } from '../types';
 import { getIndexPatterns } from '../../../../services';
@@ -26,10 +27,12 @@ export const indexPatternTypeConfig: DatasetTypeConfig = {
 
   toDataset: (path) => {
     const pattern = path[path.length - 1];
+    const patternMeta = pattern.meta as DataStructureCustomMeta;
     return {
       id: pattern.id,
       title: pattern.title,
       type: DEFAULT_DATA.SET_TYPES.INDEX_PATTERN,
+      timeFieldName: patternMeta?.timeFieldName,
       dataSource: pattern.parent
         ? {
             id: pattern.parent.id,
--- a/src/plugins/data/public/query/query_string/dataset_service/lib/index_pattern_type.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/lib/index_pattern_type.ts
@@ -17,15 +17,15 @@ import {
 import { DatasetTypeConfig } from '../types';
 import { getIndexPatterns } from '../../../../services';
 
-export const indexPatternTypeConfig: DatasetTypeConfig = {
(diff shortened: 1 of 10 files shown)


src/plugins/data/public/query/query_string/dataset_service/lib/index_pattern_type.ts
```diff
@@ -12,24 +12,27 @@ import {
...
+  DataStructureCustomMeta,
 } from '../../../../../common';
 import { DatasetTypeConfig } from '../types';
 import { getIndexPatterns } from '../../../../services';
 
-export const indexPatternTypeConfig: DatasetTypeConfig = {
-  id: DEFAULT_DATA.SET_TYPES.INDEX_PATTERN,
-  title: 'Index Patterns',
-  meta: {
+class IndexPatternTypeConfig extends DatasetTypeConfig {
```
ashwin-pc, line 20: why do we want to change it to a class based approach over the simpler config? I like the config because i dont have to reimplement a class per language. Classes are useful when you want to make multiple objects of the same class or if you want to reference a singleton. This also differs from the pattern we have in OSD to register stuff. Even registering a whole app does not require a class.

src/plugins/query_enhancements/common/types.ts
```diff
@@ -5,8 +5,16 @@
...
 
-export interface FetchDataFrameContext {
+export interface QueryAggConfig extends Partial<IAggConfig> {
+  [key: string]: any;
+  qs: {
+    [x: string]: string;
+  };
+}
+
+export interface EnhancedFetchContext {
```
ashwin-pc, line 17: Whats there in the regular fetch context?

src/plugins/query_enhancements/common/utils.ts
```diff
@@ -133,7 +133,20 @@ export const handleDataFrameError = (response: any) => {
   }
 };
 
-export const fetchDataFrame = (context: FetchDataFrameContext, query: Query, df: IDataFrame) => {
+export const fetch = (context: EnhancedFetchContext, query: Query, aggConfig?: QueryAggConfig) => {
```
ashwin-pc, line 136: Add some comments here to these functions. They seem to be specific to PPL and SQL. I.e SQL plugin related features and not the default DSL based fetch

src/plugins/query_enhancements/common/utils.ts
```diff
@@ -133,7 +133,20 @@ export const handleDataFrameError = (response: any) => {
...
+  const { http, path, signal } = context;
+  const body = JSON.stringify({ query: { ...query, format: 'jdbc' }, aggConfig });
+  return from(
+    http.fetch({
+      method: 'POST',
+      path,
+      body,
+      signal,
+    })
+  ).pipe(tap(handleDataFrameError));
```
ashwin-pc, line 146:
```suggestion
  ).pipe(tap(handleResponseError));
```
Nit: This is more readable if that its doing is just handling the http fetch error. If its doing anything else, ignore me

src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -95,12 +95,12 @@ export class QueryEnhancementsPlugin
     queryString.getLanguageService().registerLanguage(sqlLanguageConfig);
 
     data.__enhance({
```
ashwin-pc, line 97: nit: Can we change the `__enhance` name in the API and directly call the language service API's. We are exposing the language service right?

PR #7848, opened 2024-08-26
Use language service to render query editor

abbyhu2000:
### Description

* Get rid of settings class
* User language service class inside query string service to handle language related logics
* Use query enhancement to receive query editor props and render query editor

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_explorer/public/components/sidebar/index.tsx
```diff
@@ -24,23 +24,29 @@ export const Sidebar: FC = ({ children }) => {
...
+      },
       notifications: { toasts },
       application,
     },
   } = useOpenSearchDashboards<DataExplorerServices>();
 
   useEffect(() => {
-    const subscriptions = ui.Settings.getEnabledQueryEnhancementsUpdated$().subscribe(
-      (enabledQueryEnhancements) => {
+    const subscriptions = queryString
```
kavilla, line 39: i think we can clean more of this file and all we really need it to do is check the ui settings if enhancements enabled. if enhancements enabled don't show the other selector

kavilla (approved):

kavilla (approved):
follows up for more cleaning but this is nice already. merging . future prs can target main

src/plugins/data/public/query/query_string/language_service/language_service.ts
```diff
@@ -37,6 +58,87 @@ export class LanguageService {
   public getDefaultLanguage(): LanguageConfig {
     return this.languages.get('kuery') || this.languages.values().next().value;
   }
+
+  public getQueryEditorExtensionMap() {
+    return this.queryEditorExtensionMap;
+  }
+
+  resetUserQuery() {
```
kavilla, line 66: i dont think we call this anywhere right we should look to call this when it gets toggled off. in the query enhancements plugin we can listen to the ui settings and reset query if enhancements is false

src/plugins/data/public/query/query_string/language_service/types.ts
```diff
@@ -5,26 +5,22 @@
...
-    dateRange?: {
-      initialFrom?: string;
-      initialTo?: string;
-    };
-  };
+  editor: (
+    collapsedProps: any,
+    expandedProps: any,
+    bodyProps: any
+  ) => EditorInstance<any, any, any>;
```
kavilla, line 19: fast follow, export types and use them.

kavilla, line 19: nit: i think this could be optional. if they don't pass in an editor we use the default editor

kavilla: fast follow: export the default Editor, or required components

src/plugins/data/public/ui/index.ts
```diff
@@ -46,6 +39,10 @@ export {
   QueryEditorExtensions,
   QueryEditorExtensionDependencies,
   QueryEditorExtensionConfig,
+  createEditor,
+  DefaultInput,
```
kavilla, line 43: instead of exporting the type, export the createEditor from the ui service, and then you can also export the default input and single line input

src/plugins/data/public/ui/query_editor/editors/default_editor/index.tsx
```diff
@@ -27,7 +27,7 @@ interface DefaultInputProps extends React.JSX.IntrinsicAttributes {
   provideCompletionItems: monaco.languages.CompletionItemProvider['provideCompletionItems'];
 }
 
-const DefaultInput: React.FC<DefaultInputProps> = ({
+export const DefaultInput: React.FC<DefaultInputProps> = ({
```
kavilla, line 30: like above don't need to export completing just make it possible to access the instantied component

src/plugins/data/public/ui/query_editor/language_selector.tsx
```diff
@@ -57,8 +56,8 @@ export const QueryLanguageSelector = (props: QueryLanguageSelectorProps) => {
 
   languageService.getLanguages().forEach((language) => {
     if (
-      (language && props.appName && !language.supportedAppNames.includes(props.appName)) ||
-      uiService.Settings.getUserQueryLanguageBlocklist().includes(language?.id)
+      (language && props.appName && !language.editorSupportedAppNames?.includes(props.appName)) ||
```
kavilla, line 59: nit: i think we can clean this up. language select should already dispalyed when its in a supported app nap

src/plugins/data/public/ui/query_editor/language_selector.tsx
```diff
@@ -57,8 +56,8 @@ export const QueryLanguageSelector = (props: QueryLanguageSelectorProps) => {
...
+      languageService.getUserQueryLanguageBlocklist().includes(language?.id)
```
kavilla, line 60: here is where we can probably modify this to ensure that the language service.getlanguages for each language only returns backs languages that are available from the dataset.

tldr: we should just getLanguages return back already filtered instead of the lagnuage selector handling this

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -371,9 +352,11 @@ export default class QueryEditorUI extends Component<Props, State> {
       provideCompletionItems: this.provideCompletionItems,
     };
 
+    const languageEditorFunc = this.languageManager.getLanguage(this.props.query.language)!.editor;
+
     const languageEditor = useQueryEditor
-      ? createDefaultEditor(singleLineInputProps, {}, defaultInputProps)
-      : createDQLEditor(singleLineInputProps, singleLineInputProps, {
+      ? languageEditorFunc(singleLineInputProps, {}, defaultInputProps)
```
kavilla, line 358: like above i think this should already being the editor wired up with state but will need to export it properly from the UI service.

can be a clean up follow up non blocker

PR #7870, opened 2024-08-27
[discover] registered languages interceptor clean up and aggs

kavilla:
### Description

Clean up interceptor and strategies. Re-add aggregations working for PPL language

### Issues Resolved

n/a

## Screenshot

Screenshot 2024-08-27 at 4 14 51 PM

## Changelog

- fix: clean up language search interceptors and fix aggs for PPL

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/7870.yml
+++ b/changelogs/fragments/7870.yml
@@ -1,0 +1,2 @@
+fix:
+- Clean up language search interceptors and fix aggs for PPL ([#7870](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7870))
--- a/src/plugins/data/common/data_frames/utils.ts
+++ b/src/plugins/data/common/data_frames/utils.ts
@@ -199,45 +199,30 @@ export const convertResult = (response: IDataFrameResponse): SearchResponse<any>
   if (data.hasOwnProperty('aggs')) {
     const dataWithAggs = data as IDataFrameWithAggs;
     if (!dataWithAggs.aggs) {
-      // TODO: MQL best guess, get timestamp field and caculate it here
       return searchResponse;
     }
-    searchResponse.aggregations = Object.entries(dataWithAggs.aggs).reduce(
-      (acc: Record<string, unknown>, [id, value]) => {
-        const aggConfig = dataWithAggs.meta?.aggs;
-        if (id === 'other-filter') {
-          const buckets = value as DataFrameBucketAgg[];
-          buckets.forEach((bucket) => {
-            const bucketValue = bucket.value;
-            searchResponse.hits.total += bucketValue;
-          });
-          acc[id] = {
-            buckets: [{ '': { doc_count: 0 } }],
-          };
-          return acc;
-        }
-        if (aggConfig && aggConfig.type === 'buckets') {
-          const buckets = value as DataFrameBucketAgg[];
-          acc[id] = {
-            buckets: buckets.map((bucket) => {
-              const bucketValue = bucket.value;
-              searchResponse.hits.total += bucketValue;
-              return {
-                key_as_string: bucket.key,
-                key: (aggConfig as DataFrameAggConfig).date_histogram
-                  ? new Date(bucket.key).getTime()
-                  : bucket.key,
-                doc_count: bucketValue,
-              };
-            }),
-          };
-          return acc;
-        }
-        acc[id] = Array.isArray(value) ? value[0] : value;
-        return acc;
-      },
-      {}
-    );
+    searchResponse.aggregations = {};
+
+    const aggConfig = dataWithAggs.meta;
+    Object.entries(dataWithAggs.aggs).forEach(([id, value]) => {
+      if (aggConfig && aggConfig.date_histogram) {
+        const buckets = value as Array<{ key: string; value: number }>;
+        searchResponse.aggregations[id] = {
+          buckets: buckets.map((bucket) => {
+            const timestamp = new Date(bucket.key).getTime();
+            searchResponse.hits.total += bucket.value;
+            return {
+              key_as_string: bucket.key,
+              key: timestamp,
+              doc_count: bucket.value,
+            };
+          }),
+        };
+      } else {
+        // Handle other aggregation types here if needed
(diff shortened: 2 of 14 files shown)


ashwin-pc (approved):

ashwin-pc (changes requested):
Are you sure we want to move the PPL query logic to the search strategy? Here is how i envision the two being distinct:

Search Interceptor:
A search interceptor is a component that intercepts search queries or operations before they are executed. Its main purposes are:
- To modify or enhance the search query
- To perform pre-processing or validation on the search parameters
- To implement cross-cutting concerns like logging, security checks, or performance monitoring
- To potentially short-circuit the search process if certain conditions are met


Search Strategy:
A search strategy encapsulates the logic for querying the data and make them interchangeable. This should be done per dataset type

kavilla:
> Are you sure we want to move the PPL query logic to the search strategy?

I know we synced, but for the sake of the PR commenting on here. This doesn't move the query logic to to the search strategy it just cleans it up what is existing there. the interceptor.

Further clean up can be made to ensure that the interceptor calls twice to the search strategy one with the search and second with the aggs. it would make it easier to debug. But that can be a follow up as this just cleans up to not pass a dataframe in the request and updates the response to how we need it.

For PPL strategy it's not required to send agg. So looking at the SQL strategy is the basic strategy

ruanyl:
Maybe this is not related to this PR, but just want to share my thoughts:

I’ve been reading the codes of query enhancements these days, but I don’t quite capture the idea of “languages”, in my opinion, “search strategies” already did the work to add support of different languages, why introducing a new concept of “language”? To me, “language” looks simply a UI concept for end users.

And now, different language have different search interceptor, looking at the current ppl/sql search interceptor, it seems it does more than it should. For example, it even reads the UI state: time filter, agg config, etc. What if I don’t have a UI and just call search API directly? I think the search interceptor should simply take the search request and then return the search response of the corresponding search strategy. It shouldn’t read the UI state and construct the search query, the query(or search params) should be constructed before hands and then pass to `search()` as params. Because we cannot expect the caller to have the search bar UI, it may just be a arbitrary API call from anywhere.

Also, I have a bit concern on this piece of code:
```typescript
    const search = ((request, options) => {
      const isEnhancedEnabled = uiSettings.get(UI_SETTINGS.QUERY_ENHANCEMENTS_ENABLED);
      if (isEnhancedEnabled) {
        const language = queryStringManager.getQuery().language;
        const languageConfig = queryStringManager.getLanguageService().getLanguage(language);
        queryStringManager.getLanguageService().setUiOverridesByUserQueryLanguage(language);

        if (languageConfig) {
          return languageConfig.search.search(request, options);
        }
      }

      return this.defaultSearchInterceptor.search(request, options);
    }) as ISearchGeneric;
```

This is the `search()` api exposed by search_service, now we changed to use different search intercepter based on the “language” that it reads from language_service, BUT, the language it returned is deeply coupled with the language selector in the search bar on the UI. When I don’t have the selector on my page, it reads from local storage! It completely ignores the search strategy that I passed by `options.strategy`.

kavilla:
Just want to say I really appreciate the time to look into it. Last year was I was talking about this stuff and trying to research into it. It was so complicated and people didn't touch it. So happy people get it now and are able to form constructive feedback.
--
> I’ve been reading the codes of query enhancements these days, but I don’t quite capture the idea of “languages”

Apologies yeah we should have documented or RFC it a little bit better. Basically, we realized that the query object that goes through the search source contains:
* Query - the query string
* Language - string of the language (kuery, lucene, PPL, SQL)

With the ability to swap out the data source from MDS and the ability how to query the data depends on the query itself. We either had to access a bunch of services to find what we are doing which we found was causing out of sync'ness and errors. So we decided to modify the query string manager which is responsible for the current state of the query to have more information about how to query the data, and how the query the data is impacted by the components that are used to actually query the data. The monaco code editor passes in language so the UI and interceptor gets heavily impacted by the current language selected.

The default search interceptor ending up requiring building complicated ASTs to pass to an OpenSearch cluster. So there are client side changes that impacts. For example the polling. Unfortunately, the polling is short polling until the status is success, but will give 200s, and giving an indicator on the client side. So the UI has to have a sense of the language and dataset.

But you are right. We have too many interceptors. When initially build the MQL framework, my initial proposal was to have a single search interceptor for the non-native languages and then it reads the strategy off of the interceptor. But I was okay with SQL being built out as a different interceptor until we unified the final non-native interceptor. 

>  For example, it even reads the UI state: time filter, agg config, etc.

Those come from the data plugin and will be available although not required in this current approach so you would have to have a UI regardless of this with or without this work. However before you would either have to meet the complicated contract defined to add new languages globally. Updating the data plugin UI search source would require each language added to build and generate an AST, for example https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/common/opensearch_query/kuery/ast/_generated_/kuery.js. This current approach with the interceptor lets you add any language without building into something OpenSearch understands since the default route will still do too much to prep it for OpenSearch before it hits the strategy. Then each search strategy would have to parse the complicated AST.

> Because we cannot expect the caller to have the search bar UI.

For vega this is true but for the most part of the applications I would argue that developers should utilize the search bar. It handles state and integrates the entire application and keeps the same consistent behavior. The fact that there are plugins outside this repo is likely due to the result that they did not want to write the complicated ASTs to add a new language but then the experience is detached and lose a lot of the features that customers demands. Now those plugins can opt in anytime to have the search bar and  get all the features that are built out for them.

The search strategy is still available for developers to call from the server side so all they need to do is to expose an endpoint which calls directly to the search strategy they desired but overall.

> When I don’t have the selector on my page, it reads from local storage! It completely ignores the search strategy that I passed by options.strategy.

Yeah we should fix this. This likely was due to porting it over some of the query enhancements stuff.

ashwin-pc (approved):

abbyhu2000 (approved):

AMoo-Miki:
Manually backported to 2.x with https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7900

PR #7886, opened 2024-08-28
[Page Header]Render filter bar in full width with application header

zhongnansu:
### Description

- create a portal for filter bar component to render on its own row
- Add a  reference point that next to header
- Note that for discover, we still need filter bar to render as part of the SearchBar component

## Screenshot
- with filters

image
image

- without filters
image
image

## Changelog

- skip
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/public/chrome/ui/header/header.scss
+++ b/src/core/public/chrome/ui/header/header.scss
@@ -18,7 +18,6 @@
 
   &.primaryApplicationHeader {
     padding-top: $euiSizeM - $euiSizeXS;
-    border-bottom: 1px solid $euiColorLightShade;
 
     &:last-child {
       margin-bottom: 0;
@@ -104,4 +103,8 @@
 
 .stretchedActionMenu {
   width: 100%;
+}
+
+#globalHeaderBars {
+  border-bottom: 1px solid $euiColorLightShade;
 }
--- a/src/core/public/chrome/ui/header/header.tsx
+++ b/src/core/public/chrome/ui/header/header.tsx
@@ -447,6 +447,7 @@ export function Header({
         </EuiHeaderSection>
         <EuiHeaderSection side="right">{renderRightControls()}</EuiHeaderSection>
       </EuiHeader>
+      <div id="applicationHeaderFilterBar" />
     </div>
   );
 
--- a/src/plugins/data/public/ui/filter_bar/filter_bar.tsx
+++ b/src/plugins/data/public/ui/filter_bar/filter_bar.tsx
@@ -40,6 +40,7 @@ import {
 import { FormattedMessage, InjectedIntl, injectI18n } from '@osd/i18n/react';
 import classNames from 'classnames';
 import React, { useState } from 'react';
+import { createPortal } from 'react-dom';
 import { stringify } from '@osd/std';
 
 import { FilterEditor } from './filter_editor';
@@ -55,6 +56,7 @@ interface Props {
   className: string;
   indexPatterns: IIndexPattern[];
   intl: InjectedIntl;
+  isFilterBarPortable?: boolean;
 }
 
 const maxFilterWidth = 600;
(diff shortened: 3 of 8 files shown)


virajsanghvi:
Can you post a screenshot of what it looks like with no filters?

src/core/public/chrome/ui/header/header.scss
```diff
@@ -18,7 +18,6 @@
 
   &.primaryApplicationHeader {
     padding-top: $euiSizeM - $euiSizeXS;
-    border-bottom: 1px solid $euiColorLightShade;
```
virajsanghvi, line 21: was this just unnecessary or what's the before/after look like for this border? would this impact current headers (before page headers) in any way?

zhongnansu:
> Can you post a screenshot of what it looks like with no filters?

added to the PR description

src/core/public/chrome/ui/header/header.tsx
```diff
@@ -447,6 +448,8 @@ export function Header({
         </EuiHeaderSection>
         <EuiHeaderSection side="right">{renderRightControls()}</EuiHeaderSection>
       </EuiHeader>
+      <div id="applicationHeaderFilterBar" />
+      <EuiHorizontalRule margin="none" />
```
virajsanghvi, line 452: guessing this replaces the border? does this only take up 1px height?

src/plugins/data/public/ui/filter_bar/filter_bar.tsx
```diff
@@ -171,12 +173,14 @@ function FilterBarUI(props: Props) {
 
   const classes = classNames('globalFilterBar', props.className);
   const filterBarPrefixText = i18n.translate('data.search.filterBar.filterBarPrefixText', {
-    defaultMessage: 'Filters: ',
+    defaultMessage: 'Filters',
   });
+  const filterGroupClassName =
+    useNewHeader && !props.fixFilterBar ? 'globalFilterGroup--compressed' : 'globalFilterGroup';
```
virajsanghvi, line 179: nit: should we just add the compressed class in the compress case vs switching between the two?

src/plugins/data/public/ui/filter_bar/filter_bar.tsx
```diff
@@ -55,6 +56,7 @@ interface Props {
   className: string;
   indexPatterns: IIndexPattern[];
   intl: InjectedIntl;
+  fixFilterBar?: boolean;
```
virajsanghvi, line 59: nit: Can you just add a comment for what this "fixes"?

src/core/public/chrome/ui/header/header.tsx
zhongnansu: yes, otherwise it will be 2 rows and a line separator between the header and the filter bar, which looks really bad and not align with the mockups

zhongnansu: image

src/core/public/chrome/ui/header/header.scss
zhongnansu: I made this change along with adding the `EuiHorizontalRule`,  in order to let the line separator always stays at the bottom, to make it align with UX. This shouldn't affect current legacy headers, since this class is only added for new application header

src/plugins/data/public/ui/filter_bar/filter_bar.tsx
zhongnansu: will make the change

src/core/public/chrome/ui/header/header.tsx
virajsanghvi: Do we use horizontal rules for this elsewhere? If so, that's fine, otherwise, think this should probably be a css border based on the presence of siblings or last of operator.

src/plugins/data/public/ui/filter_bar/filter_bar.tsx
zhongnansu: updated to use a new naming `isFilterBarPortable` that self explains, and also updated the related logic

src/core/public/chrome/ui/header/header.tsx
zhongnansu: sure, switched back to use css bottom
image

virajsanghvi (approved):
Lgtm, although please validate failing tests are unrelated

virajsanghvi (approved):

PR #7926, opened 2024-08-29
fix(discover): disable cache only on `indexPatterns.get` to prevent crash

joshuali925:
### Description

Right now in new discover, if user adds/removes two fields on the left quickly, it will crash with
```
Something went wrong
Try refreshing the page. If that doesn't work, go back to the previous page or clear your session data.


Clear your session

Go back
Cannot read properties of undefined (reading 'type')
Version: 3.0.0
Build: 9007199254740991
TypeError: Cannot read properties of undefined (reading 'type')
```

console error
```
fatal_errors_service.tsx:89 TypeError: Cannot read properties of undefined (reading 'type')
    at isDefault (is_default.ts:33:1)
    at updateSearchSource (update_search_source.ts:32:1)
    at async use_search.ts:101:1
    at async use_search.ts:193:1
```

I believe this happens because cache is being updated when user adds a field, and the index pattern is not available in the cache during the update.
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a9f84be5665ff89bebe75bfef8053b42f76c610b/src/plugins/discover/public/application/helpers/popularize_field.ts#L45-L49

If user adds another field when cache is not available, the index pattern will be undefined, and calling `undefined.type` crashes the page
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a9f84be5665ff89bebe75bfef8053b42f76c610b/src/plugins/data/common/index_patterns/lib/is_default.ts#L33-L36

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/view_components/utils/update_search_source.ts
+++ b/src/plugins/discover/public/application/view_components/utils/update_search_source.ts
@@ -34,7 +34,7 @@ export const updateSearchSource = async ({
 
   const dataset =
     indexPattern.id === queryDataset?.id
-      ? await data.indexPatterns.get(queryDataset?.id!, true)
+      ? await data.indexPatterns.get(queryDataset?.id!)
       : indexPattern;
 
   const sortForSearchSource = getSortForSearchSource(


kavilla:
do we know if this works fine with indices on load?

joshuali925:
> do we know if this works fine with indices on load?

the fields from the last index-pattern search are showed on the left when searching indices, seems to be another bug. for testing i searched the index with the same fields, and adding/removing fields worked

ashwin-pc (approved):

AMoo-Miki:
Spoke to @joshuali925; the `true` removal makes it so the dataset doesn't rely on cache.

joshuali925:
adding some context, i sent this PR because page crash affects user experience and i want to fix it for 2.17. this seems to be the easiest fix, i'm not sure what is the best fix

AMoo-Miki:
Spoke to Rocky offline:
```
joshua's work will address this issue. it's a good observation regarding MQL. in the initial implementation, before we added the data explorer component in version 2.17, i created a temporary index pattern for the results. this was necessary because the search source required an explicit index pattern.
previously, the process worked like:
check if the index pattern exists in the cache
if not found in the cache, query the saved objects client
if still not found, attempt to query the cluster for the index
For PPL and SQL queries, I would:
create a dataframe
if the dataframe existed on the search source, create a temporary index pattern
insert this temporary pattern only in the cache (not creating a saved object)
bypass the normal logic and only check the index pattern cache if the dataframe existed in the search source
which prevented potential errors that could occur when querying an index or table. with the explorer component, we create the temporary index pattern much earlier in the process, before any search source interactions. which allows us to rely on the normal code path to work properly, eliminating the need for the previous workaround
```

AMoo-Miki (approved):

Issue #7632, opened 2024-08-06
[BUG] Alias fields not visible in Discover tab for non-date field types


ananzh:
**Describe the bug**
Alias fields in OpenSearch are not consistently visible in the Discover . The visibility depends on the data type of the field being aliased. Specifically, alias fields are visible when the aliased field is of type "date", but not visible when the aliased field is of type "keyword".

**To Reproduce**
1. Create two indices with different mappings:
```
   Index 1 (keyword type):
   PUT movies-1
   {
     "mappings": {
       "properties": {
         "year": {
           "type": "keyword"
         },
         "release_date_movies": {
           "type": "alias",
           "path": "year"
         }
       }
     }
   }

   Index 2 (date type):
   PUT movies-2
   {
     "mappings": {
       "properties": {
         "year": {
           "type": "date"
         },
         "release_date_movies": {
           "type": "alias",
           "path": "year"
         }
       }
     }
   }
```

2. Add sample data to each index:
```
   For movies-1:
   POST movies-keyword/_doc
   {
     "year": "2024"
   }

   For movies-2:
   POST movies-date/_doc
   {
     "year": "2024-01-01"
   }
```

3. Open Discover.
4. Select the index pattern for movies-1 and observe the available fields.
<img width="1498" alt="Screenshot 2024-08-06 at 4 36 40 PM" src="https://github.com/user-attachments/assets/f1fcaf98-b1b4-4705-a632-56efc841b71a">

5. Select the index pattern for movies-2 and observe the available fields.
<img width="1498" alt="Screenshot 2024-08-06 at 4 36 31 PM" src="https://github.com/user-attachments/assets/54dc83e5-dc49-436a-9712-19bd78897571">

**Expected behavior**
Both "year" and "release_date_movies" fields should be visible in the Discover for both indices.

Actual Behavior:
- For movies-1: Only the "year" field is visible in the Discover tab.
- For movies-2: Both "year" and "release_date_movies" fields are visible in the Discover tab.

**OpenSearch Version**
currently only try main

**Dashboards Version**
currently only try main

huyaboo:
Will triage and release feedback today

ashwin-pc:
We are currently working on supporting auto complete for Discover for 2.17. That should help mitigate some of the pain here until we dig deeper into this issue.

kavilla:
One thing that I noticed that the type of the first screenshot of the release date movies is of type text indicated by the 't' and doesn't display.

[...]

PR #7930, opened 2024-08-30
[Fix] Show Alias Fields in Discover Tab 

sejli:
### Description

Adds a new advanced setting that updates the Discover's search request to include `"fields": ["*"]`. Defaults to off, open to suggestions for the title and description of the new setting.

Thanks @LDrago27 for the [deep dive](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/7632#issuecomment-2307713526) and [fix](https://github.com/LDrago27/OpenSearch-Dashboards/commit/fddc1a933b6cbd5c2aa502a82d6de63603e3ef09)!

### Issues Resolved

#7632 

## Changelog

- fix: Show alias fields in Discover tab

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ashwin-pc (approved):

src/plugins/data/server/ui_settings.ts
```diff
@@ -762,5 +762,19 @@ export function getUiSettings(): Record<string, UiSettingsParams<unknown>> {
...
+      }),
+      value: false,
+      description: i18n.translate('data.advancedSettings.searchUseFieldsDescription', {
+        defaultMessage: `
+        <strong>Experimental</strong>:
+        Adds the <code>"fields": ["*"]</code> property to search request body`,
+      }),
+      schema: schema.boolean(),
+      category: ['search'],
+      requiresPageReload: true,
```
joshuali925, line 777: why does this require reload?

src/plugins/data/server/ui_settings.ts
```diff
@@ -762,5 +762,19 @@ export function getUiSettings(): Record<string, UiSettingsParams<unknown>> {
       }),
       schema: schema.arrayOf(schema.string()),
     },
+    [UI_SETTINGS.SEARCH_USE_FIELDS]: {
+      name: i18n.translate('data.advancedSettings.searchUseFieldsTitle', {
+        defaultMessage: 'Include all fields in search request',
```
joshuali925, line 767: this setting is to include all fields, would `search:includeAllFields` be better than `search:useFields`?

joshuali925 (dismissed):

src/plugins/data/server/ui_settings.ts
sejli: Just checked with @kavilla, `search_source` should access the dependencies every time it calls `flatten()`, so we don't need it. Nice catch.

src/plugins/data/server/ui_settings.ts
sejli: My thinking was that previously, we weren't using the `fields` field in the request body. Turning on this property "uses" `fields`. I think in this case, the title itself could be something different.

ashwin-pc: I like this suggestion

kavilla (approved):

ashwin-pc (approved):

kavilla (approved):

PR #7932, opened 2024-08-30
[Auto Suggest] PPL autocomplete and interface changes

paulstn:
### Description

Adds PPL Autocomplete using ANTLR and C3 Code Completion.
Also makes interface changes to include a 'details' aspect to the suggestions which will include a relevant label for each one. Another interface change allows a suggestion to specify the range of characters that it should replace. 

Note: Every file worth reviewing is under `src/plugins/data/public`

## Changelog

- feat: PPL Autocomplete
- fix: Auto Suggest interface updates
- fix: Single Line Editor Suggestion Window covered

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


docs/_sidebar.md
ashwin-pc, line 1: You may want to rebase. This should already be in main

package.json
```diff
@@ -85,7 +85,8 @@
...
     "osd:ciGroup10": "echo \"dashboard_sanity_test_spec.js\"",
     "osd:ciGroup11": "echo \"apps/vis_builder/*.js\"",
-    "generate:opensearchsqlantlr": "./node_modules/antlr4ng-cli/index.js -Dlanguage=TypeScript -o ./src/plugins/data/public/antlr/opensearch_sql/.generated -visitor -no-listener -Xexact-output-dir ./src/plugins/data/public/antlr/opensearch_sql/grammar/OpenSearchSQLLexer.g4 ./src/plugins/data/public/antlr/opensearch_sql/grammar/OpenSearchSQLParser.g4"
+    "generate:opensearchsqlantlr": "./node_modules/antlr4ng-cli/index.js -Dlanguage=TypeScript -o ./src/plugins/data/public/antlr/opensearch_sql/.generated -visitor -no-listener -Xexact-output-dir ./src/plugins/data/public/antlr/opensearch_sql/grammar/OpenSearchSQLLexer.g4 ./src/plugins/data/public/antlr/opensearch_sql/grammar/OpenSearchSQLParser.g4",
+    "generate:opensearchpplantlr": "./node_modules/antlr4ng-cli/index.js -Dlanguage=TypeScript -o ./src/plugins/data/public/antlr/opensearch_ppl/.generated -visitor -no-listener -Xexact-output-dir ./src/plugins/data/public/antlr/opensearch_ppl/grammar/OpenSearchPPLLexer.g4 ./src/plugins/data/public/antlr/opensearch_ppl/grammar/OpenSearchPPLParser.g4"
```
ashwin-pc, line 89: Cant we have a single command to generate all the antlr grammer? Alternatively, not for this PR by we can also use the `osd` script to add thsi as a command there to not bloat up the package.json commands. e.g. `yarn osd antlr-grammar`

src/plugins/data/public/ui/query_editor/query_editor.tsx
ashwin-pc, line 1: Nice!

src/plugins/data/public/plugin.ts
```diff
@@ -163,9 +164,10 @@ export class DataPublicPlugin
       }))
     );
 
-    const ac = this.autocomplete.setup(core);
-    ac.addQuerySuggestionProvider('SQL', getSQLSuggestions);
-    ac.addQuerySuggestionProvider('kuery', getDQLSuggestions);
+    const autoComplete = this.autocomplete.setup(core);
+    autoComplete.addQuerySuggestionProvider('SQL', getSQLSuggestions);
+    autoComplete.addQuerySuggestionProvider('kuery', getDQLSuggestions);
+    autoComplete.addQuerySuggestionProvider('PPL', getPPLSuggestions);
```
ashwin-pc, line 170: nit: we should move adding the suggestion provider for PPL and SQL to the query enhancements plugin since registering the language happens there as well. Making the suggestion provider also be a part of the language service's register function would bring all these language related features into one place too.

src/plugins/data/public/antlr/shared/utils.ts
ashwin-pc, line 1: Can we add some tests for this?

docs/_sidebar.md
paulstn: i believe its removed now, thanks

package.json
paulstn: We could, but I'd assume more granularity would be better when needing to regenerate the antlr files. Do you think we should make this change now or handle it when moving it to the osd script?

kavilla (dismissed):

package.json
kavilla: +1 i think in @mengweieric PR made a similiar comment. Like post 2.17, we should have something in the `scripts` folder that lets us run `yarn antlr:generateGrammar --sql` or something

changelogs/fragments/7810.yml
```diff
@@ -0,0 +1,2 @@
+feat:
+- Add OpenSearch PPL autocomplete to discover 2.0 with query enhancements ([#7810](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7810))
```
joshuali925, line 2: np. there are two changelogs but non of them is for #7932?

packages/osd-monaco/src/xjson/lexer_rules/opensearchppl.ts
```diff
@@ -0,0 +1,529 @@
...
+          },
+        },
+      ],
+      [/[()]/, '@brackets'],
+      [/--.*$/, 'comment'],
+      [/\/\*/, 'comment', '@comment'],
+      [/\/.*$/, 'comment'],
+
+      [/".*?"/, 'string'],
+      [/'.*?'/, 'constant'],
```
joshuali925, line 494: curious how are these defined? i don't think PPL supports comments, and the string definition is a bit different from sql

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -295,12 +296,17 @@ export default class QueryEditorUI extends Component<Props, State> {
...
       suggestions:
         suggestions && suggestions.length > 0
-          ? suggestions.map((s: QuerySuggestion) => ({
-              label: s.text,
-              kind: s.type as monaco.languages.CompletionItemKind,
-              insertText: s.insertText ?? s.text,
-              range,
-            }))
+          ? suggestions
+              .filter((s) => 'detail' in s) // remove suggestion not of type MonacoCompatible
```
joshuali925, line 300: what is detail?

joshuali925 (approved):
lgtm, just questions

changelogs/fragments/7810.yml
paulstn: Just updated it, added more details :)

src/plugins/data/public/ui/query_editor/query_editor.tsx
paulstn: detail is a field in a possible suggestion that is a label on what that suggestion is. for example, if we're suggesting a field of type geo_point, it will read `Field: geo_point` next to the suggestion. the reason it's being used here is as a type guard since detail is a monaco specific suggestion field

packages/osd-monaco/src/xjson/lexer_rules/opensearchppl.ts
paulstn: good catch, i'm not sure why that's in there. best guess is it was part of the boilerplate. i'll look it over and make the needed changes

joshuali925: is the boilerplate generated by something or copied from some pre-existing code? or is this written from scratch by this PR? asking because looks like the rules are from some standard but might not completely apply to PPL syntax. probably won't break but just curious

joshuali925 (approved):

ashwin-pc (approved):

packages/osd-monaco/src/xjson/lexer_rules/opensearchppl.ts
paulstn: i just mean copied from the other monarch lexer rule files, i agree on it not completely applying to ppl

PR #7936, opened 2024-08-30
feat: added a new API to allow adding additional editor controls

ruanyl:
### Description

Allow externals to add additional query editor controls

## Screenshot

image

## Changelog

- feat: added a new API to allow externals to add additional controls to query editor

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/query_editor/query_editor.tsx
+++ b/src/plugins/data/public/ui/query_editor/query_editor.tsx
@@ -16,6 +16,7 @@ import {
 } from '@elastic/eui';
 import classNames from 'classnames';
 import { isEqual } from 'lodash';
+import { BehaviorSubject, Subscription } from 'rxjs';
 import React, { Component, createRef, RefObject } from 'react';
 import { monaco } from '@osd/monaco';
 import {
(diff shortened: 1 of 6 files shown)


ruanyl:
@kavilla @ashwin-pc @abbyhu2000 Would love to see your comments :)

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -60,6 +62,7 @@ export interface QueryEditorProps {
   filterBar?: any;
   prepend?: React.ComponentProps<typeof EuiCompressedFieldText>['prepend'];
   savedQueryManagement?: any;
+  additionalControls$?: BehaviorSubject<SearchBarControl[]>;
```
ashwin-pc, line 65: Why do we need this to be an observable?

ruanyl: Might not be a hard requirement now, but I was thinking, it might make sense to make it an observable so that the controls can be added on the fly, similar to dynamic action triggers.

SuZhou-Joe (approved):

gaobinlong (approved):

src/plugins/data/public/ui/query_editor/query_editor.tsx
ashwin-pc: There are 2 problems I see with this approach.
1. Whats the usecase for such dynamic controls? Without a use I don't want to add more complexity to the query bar. It think it's more useful to pass the current app state to determine if the controls are needed rather than having an observable. 
2. Right now you can only add controls but not remove them. But this is an easy fix compared to the first one though

PR #7943, opened 2024-08-30
[discover] async query and caching

kavilla:
### Description

Async query feature for S3 type on Discover. Due to the time to click and verify I end up implementing the cache to avoid waiting too long per every re-build.

What this PR does:
- Poll for query
- Cache data structures and refetches them from session storage
- Poll based on the data type

What this PR does NOT do yet:
- SessionId for search strategy, working fine for selector
- isCacheable field property
- Abort signal on server

### Issues Resolved

n/a

## Changelog
- feat: async query search and caching, also adding tests to related components

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #7948, opened 2024-08-30
Discover Bug Fixes

LDrago27:
### Description

This changes fixes the following bugs
- SQL Search not working for IndexPatterns with Wildcards in the name
- Intial Page Load was not working for Non DQL default language

## Screenshot

https://github.com/user-attachments/assets/0d3f4929-8a05-45ab-bcb2-3d4fe3a4c576

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/server/utils/facet.ts
+++ b/src/plugins/query_enhancements/server/utils/facet.ts
@@ -40,7 +40,7 @@ export class Facet {
       const query: Query = request.body.query;
       const { format, df } = request.body;
       const params = {
-        body: { ...query },
+        body: { query: query.query },
         ...(format !== 'jdbc' && { format }),
       };
       const clientId = query.dataset?.dataSource?.id ?? df?.meta?.queryConfig?.dataSourceId;
--- a/src/plugins/data/public/query/query_string/query_string_manager.ts
+++ b/src/plugins/data/public/query/query_string/query_string_manager.ts
@@ -60,14 +60,32 @@ export class QueryStringManager {
   }
 
   public getDefaultQuery() {
-    return {
-      query: this.getDefaultQueryString(),
-      language: this.getDefaultLanguage(),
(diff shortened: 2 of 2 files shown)


src/plugins/query_enhancements/server/utils/facet.ts
```diff
@@ -40,7 +40,7 @@ export class Facet {
       const query: Query = request.body.query;
       const { format, df } = request.body;
       const params = {
-        body: { ...query },
+        body: { query: query.query },
```
sejli, line 43: Is there a reason why we removed the rest of the query object? I don't remember if the `fetch()` function for `PPL` and `SQL` use the rest of the object off the top of my head.

sejli (approved):
LGTM. Thanks for helping to fix these bugs!

src/plugins/query_enhancements/server/utils/facet.ts
LDrago27: Before this change the request body that we used to send was of the following format
```
{
    query: {
      query: 'SELECT * FROM aaa* LIMIT 10',
      language: 'SQL',
      dataset: [Object],
      format: 'jdbc'
    },
    aggConfig: null,
    df: null
}
```
However the SQL plugins request only needs the following parameters.
image

Also we observed sending in the extra unrecognized parameters was causing bugs where the SQL Search not working for IndexPatterns with Wildcards in the name, therefore this PR aims to cleanup the request body that we send.

abbyhu2000 (approved):

src/plugins/query_enhancements/server/utils/facet.ts
kavilla: yeah end up doing something similar here too https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7943/files

Issue #7821, opened 2024-08-23
[RFC] User level settings


ruanyl:
## Introduction
Currently, OpenSearch-Dashboards provide global settings that apply to all users. The purpose of this RFC is to introduce a new feature that allows to customize advance settings at user level. This feature ensures that when a user changes their settings, it will not affect other users. 

## Goals
- Implement user-level settings that allow individual customization without affecting other users.
- Ensure seamless integration with existing uiSettings client.
- Maintain backward compatibility with previous versions of OpenSearch-Dashboards.

## UX changes
1. There will be a new page "User settings" added to allow user to update their own settings from a dedicated place.
2. The current global advance settings page will remain, changes on this page will affect only the global settings.

## Design details
To support user-level settings we'll need to modify the current ui settings schema by introducing a new field `scopes` to indicate  whether a setting is `global` or `user` or both.

### Registering a setting with scope
The `scopes` is an array, it can contain `{type: 'global'}` or `{type: 'user'}` or both. If no scope provided, it defaults to `global`.

For example:
```diff
{
    dateFormat: {
      name: i18n.translate('core.ui_settings.params.dateFormatTitle', {
        defaultMessage: 'Date format',
      }),
      value: 'MMM D, YYYY @ HH:mm:ss.SSS',
      description: '...',
      schema: schema.string(),
+     scopes: [{type: 'global'}, {type: 'user'}]
    }
}
```

### Reading a setting with scope
When calling `uiSettings.get(key)`, it returns the value based on the following rules:
1. If the setting was registered with scope `global` or empty, the global setting will be returned.
2. If the setting was registered with scope `user`, the user setting will be returned
3. If the setting was registered with both `user` and `global`, the user setting will be returned if it's set, otherwise, return the global setting
4. If calling `uiSettings.get(key, defaultValue, 'user')` with scope specified, it will return the setting of the specified scope. If the value doesn't not exist, then return the `defaultValue`. **This requires to update the current uiSettings service's interface**

### Writing a setting with scope
When calling `uiSettings.set(key, value)`, it follows the following rules:
1. If the setting was registered with scope `global` or empty, it will write the value to global setting saved object.
2. If the setting was registered with scope `user`, it will write the value to user setting saved object.
3. If the setting was registered with both `user` and `global`, it will by default update the user setting saved object.
[...]

Issue #7909, opened 2024-08-29
[Workspace] User setting page


Hailong-am:
**Is your feature request related to a problem? Please describe.**

Add a new user settings page which allows user to set default workspace

## workspace enabled
![image](https://github.com/user-attachments/assets/ebaaea8e-9703-4613-9fa3-5e7f317b1e3d)

## workspace not enabled

![image](https://github.com/user-attachments/assets/b607668a-b888-4717-ac31-9d21bf8a5ef2)

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #7953, opened 2024-09-02
[Workspace] Add User level setting

Hailong-am:
### Description

This PR includes 
1. user settings page
2. ui_setting_client changes to support user personal settings
3. UserUISettingsClientWrapper to translate user id placeholder to real user id

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/7909

## Screenshot

https://github.com/user-attachments/assets/57e9962b-4869-4a41-911e-9a698480708e

## Testing the changes

1. Run OSD enabled `yarn start --no-base-path --workspace.enabled=true`
2. make sure security is enabled
3. open http://localhost:5601, and then go to dev tools
4. Run kibana search to make sure global setting been created
```
GET .kibana/_search
{
  "query": {
    "bool": {
      "must": [
        {
          "match": {
            "type": "config"
          }
        }
      ]
    }
  }
```
5. Run command to add user setting, after that run kibana query again to verify 
a config doc with id `config:admin` been indexed and it has defaultWorkspace as "ABC"
```
curl --request POST \
  --url http://localhost:5601/api/opensearch-dashboards/settings \
  --header 'Authorization: Basic YWRtaW46YWRtaW4=' \
  --header 'Content-Type: application/json' \
  --header 'osd-xsrf: osd-fetch' \
  --data '{
 "changes": {
  "defaultIndex": "logstash-*",
  "defaultWorkspace": "ABC"
 }
}'
```

## Changelog

- feat: [Workspace] Support user personal settings

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/core/server/ui_settings/ui_settings_client.ts
```diff
@@ -62,6 +64,9 @@ type UiSettingsRawValue = UiSettingsParams & UserProvidedValue;
 type UserProvided<T = unknown> = Record<string, UserProvidedValue<T>>;
 type UiSettingsRaw = Record<string, UiSettingsRawValue>;
 
+// identifier for current user
+export const CURRENT_USER = '<current_user>';
```
SuZhou-Joe, line 68:
```suggestion
export const CURRENT_USER_PLACEHOLDER = '<current_user>';
```
`CURRENT_USER` seems a little bit confusing to me, actually it is a placeholder if I understand correctly.

src/plugins/workspace/public/components/workspace_list/index.tsx
```diff
@@ -72,12 +107,14 @@ export const WorkspaceList = ({ registeredUseCases$ }: WorkspaceListProps) => {
...
     pageIndex: 0,
-    pageSize: 5,
+    pageSize: 10,
     pageSizeOptions: [5, 10, 20],
   });
   const [deletedWorkspaces, setDeletedWorkspaces] = useState<WorkspaceAttribute[]>([]);
   const [selection, setSelection] = useState<WorkspaceAttribute[]>([]);
   const [allDataSources, setAllDataSources] = useState<DataSourceAttributesWithWorkspaces[]>([]);
+  //  default workspace state
+  const [defaultWorkspace, setDefaultWorkspace] = useState<string | undefined>(undefined);
```
raintygao, line 117: Nit: Would it be more clear if it is named as `defaultWorkspaceId`?

src/plugins/workspace/public/components/workspace_list/index.tsx
```diff
@@ -361,124 +442,151 @@ export const WorkspaceList = ({ registeredUseCases$ }: WorkspaceListProps) => {
...
+      'data-test-subj': 'workspace-list-set-default-icon',
+      onClick: (item: WorkspaceAttribute) => handleSetDefaultWorkspace(item),
+    },
+    {
+      name: <EuiText key="leave">Leave</EuiText>,
+      icon: 'exit',
+      type: 'icon',
+      description: 'Leave workspace',
+      'data-test-subj': 'workspace-list-leave-icon',
+      available: () => false,
```
raintygao, line 477: Do we need this config if it would always be unavailable?

src/core/server/ui_settings/ui_settings_client.ts
```diff
@@ -201,16 +228,45 @@ export class UiSettingsClient implements IUiSettingsClient {
...
+      }
+    });
+
+    return [globalChanges, userChanges];
+  }
+
   private async write({
     changes,
     autoCreateOrUpgradeIfMissing = true,
+    userLevel = false,
```
ruanyl, line 260: I think a boolean flag `userLevel: boolean` may not flexible enough, imagine in the future, we may have uiSettings client to support workspace level settings.

src/plugins/workspace/public/components/workspace_list/index.tsx
Hailong-am: we plan to enable it in next PR to let user leave from selected workspace only if user permission granted by individual user id not group.

src/plugins/workspace/public/components/workspace_list/index.tsx
Hailong-am: thanks, will update to `defaultWorkspaceId`

src/core/server/ui_settings/ui_settings_client.ts
```diff
@@ -201,16 +228,45 @@ export class UiSettingsClient implements IUiSettingsClient {
...
+    Object.entries(this.defaults).forEach(([key, value]) => {
+      if (value.scope === 'user' || (Array.isArray(value.scope) && value.scope.includes('user'))) {
+        userLevelKeys.push(key);
+      }
+    });
+    const userChanges = {} as Record<string, any>;
+    const globalChanges = {} as Record<string, any>;
+
+    Object.entries(changes).forEach(([key, val]) => {
+      if (userLevelKeys.includes(key)) {
```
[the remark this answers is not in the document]
ruanyl: Here is just a few changes to demonstrate what I was referring:
```diff
diff --git a/src/core/server/ui_settings/ui_settings_client.ts b/src/core/server/ui_settings/ui_settings_client.ts
index e78f797465..b48a32e6d9 100644
--- a/src/core/server/ui_settings/ui_settings_client.ts
+++ b/src/core/server/ui_settings/ui_settings_client.ts
@@ -50,6 +50,7 @@ export interface UiSettingsServiceOptions {
 interface ReadOptions {
   ignore401Errors?: boolean;
   autoCreateOrUpgradeIfMissing?: boolean;
+  scope?: Scope;
 }
 
 interface UserProvidedValue<T = unknown> {
@@ -62,6 +63,8 @@ type UiSettingsRawValue = UiSettingsParams & UserProvidedValue;
 type UserProvided<T = unknown> = Record<string, UserProvidedValue<T>>;
 type UiSettingsRaw = Record<string, UiSettingsRawValue>;
 
+type Scope = 'user' | 'global';
+
 export class UiSettingsClient implements IUiSettingsClient {
   private readonly type: UiSettingsServiceOptions['type'];
   private readonly id: UiSettingsServiceOptions['id'];
@@ -100,13 +103,13 @@ export class UiSettingsClient implements IUiSettingsClient {
     return this.defaults[key]?.value;
   }
 
-  async get<T = any>(key: string): Promise<T> {
-    const all = await this.getAll();
+  async get<T = any>(key: string, scope?: Scope): Promise<T> {
+    const all = await this.getAll(scope);
     return all[key];
   }
 
-  async getAll<T = any>() {
-    const raw = await this.getRaw();
+  async getAll<T = any>(scope?: Scope) {
+    const raw = await this.getRaw(scope);
 
     return Object.keys(raw).reduce((all, key) => {
       const item = raw[key];
@@ -115,8 +118,10 @@ export class UiSettingsClient implements IUiSettingsClient {
     }, {} as Record<string, T>);
   }
 
-  async getUserProvided<T = unknown>(): Promise<UserProvided<T>> {
-    const userProvided: UserProvided<T> = this.onReadHook<T>(await this.read());
+  async getUserProvided<T = unknown>(scope?: Scope): Promise<UserProvided<T>> {
+    const userProvided: UserProvided<T> = this.onReadHook<T>(
+      await this.read({ scope, autoCreateOrUpgradeIfMissing: scope === 'global' })
+    );
 
     // write all overridden keys, dropping the userValue is override is null and
     // adding keys for overrides that are not in saved object
@@ -128,13 +133,13 @@ export class UiSettingsClient implements IUiSettingsClient {
     return userProvided;
   }
 
-  async setMany(changes: Record<string, any>) {
+  async setMany(changes: Record<string, any>, scope?: Scope) {
     this.onWriteHook(changes);
-    await this.write({ changes });
+    await this.write({ changes, scope });
   }
 
-  async set(key: string, value: any) {
-    await this.setMany({ [key]: value });
+  async set(key: string, value: any, scope?: Scope) {
+    await this.setMany({ [key]: value }, scope);
   }
 
   async remove(key: string) {
@@ -159,8 +164,8 @@ export class UiSettingsClient implements IUiSettingsClient {
     }
   }
 
-  private async getRaw(): Promise<UiSettingsRaw> {
-    const userProvided = await this.getUserProvided();
+  private async getRaw(scope?: Scope): Promise<UiSettingsRaw> {
+    const userProvided = await this.getUserProvided(scope);
     return defaultsDeep(userProvided, this.defaults);
   }
 
@@ -204,25 +209,32 @@ export class UiSettingsClient implements IUiSettingsClient {
   private async write({
     changes,
     autoCreateOrUpgradeIfMissing = true,
+    scope = 'global',
   }: {
     changes: Record<string, any>;
     autoCreateOrUpgradeIfMissing?: boolean;
+    scope?: Scope;
   }) {
     changes = this.translateChanges(changes, 'timeline', 'timelion');
+    const id = this.getId(scope);
     try {
-      await this.savedObjectsClient.update(this.type, this.id, changes);
+      await this.savedObjectsClient.update(this.type, id, changes);
     } catch (error) {
       if (!SavedObjectsErrorHelpers.isNotFoundError(error) || !autoCreateOrUpgradeIfMissing) {
         throw error;
       }
 
-      await createOrUpgradeSavedConfig({
-        savedObjectsClient: this.savedObjectsClient,
-        version: this.id,
-        buildNum: this.buildNum,
-        log: this.log,
-        handleWriteErrors: false,
-      });
+      if (scope === 'global') {
+        await createOrUpgradeSavedConfig({
+          savedObjectsClient: this.savedObjectsClient,
+          version: this.id,
+          buildNum: this.buildNum,
+          log: this.log,
+          handleWriteErrors: false,
+        });
+      } else {
+        await this.savedObjectsClient.create('config', {}, { id });
+      }
 
       await this.write({
         changes,
@@ -231,12 +243,21 @@ export class UiSettingsClient implements IUiSettingsClient {
     }
   }
 
+  getId(scope?: Scope) {
+    if (scope === 'user') {
+      return '__USER_SETTING_FLAG__';
+    }
+    return this.id;
+  }
+
   private async read({
     ignore401Errors = false,
     autoCreateOrUpgradeIfMissing = true,
+    scope = 'global',
   }: ReadOptions = {}): Promise<Record<string, any>> {
+    const id = this.getId(scope);
     try {
-      const resp = await this.savedObjectsClient.get<Record<string, any>>(this.type, this.id);
+      const resp = await this.savedObjectsClient.get<Record<string, any>>(this.type, id);
       return this.translateChanges(resp.attributes, 'timelion', 'timeline');
     } catch (error) {
       if (SavedObjectsErrorHelpers.isNotFoundError(error) && autoCreateOrUpgradeIfMissing) {
@@ -262,6 +283,10 @@ export class UiSettingsClient implements IUiSettingsClient {
         return {};
       }
 
+      if (scope === 'user') {
+        return {};
+      }
+
       throw error;
     }
   }
```

src/plugins/advanced_settings/public/management_app/advanced_settings.tsx
```diff
@@ -172,6 +172,7 @@ export class AdvancedSettingsComponent extends Component<
     const all = config.getAll();
     const userSettingsEnabled = config.get('theme:enableUserControl');
     return Object.entries(all)
+      .filter((setting) => setting[1].scope !== 'user')
```
SuZhou-Joe, line 175: Can we add an enum for scope?

src/core/server/ui_settings/ui_settings_client.ts
Hailong-am: make sense, will updated to suggested name.

src/plugins/advanced_settings/public/plugin.ts
```diff
@@ -142,6 +190,15 @@ export class AdvancedSettingsPlugin
       setupTopNavThemeButton(core, core.uiSettings.get('home:useNewHomePage'));
     }
 
+    this.appUpdater$.next((app) => {
+      const securityEnabled = core.application.capabilities.userSettings?.enabled;
```
SuZhou-Joe, line 194: userSettingsEnabled?

src/plugins/advanced_settings/public/plugin.ts
```diff
@@ -116,6 +126,44 @@ export class AdvancedSettingsPlugin
       },
     ]);
 
+    if (core.chrome.navGroup.getNavGroupEnabled()) {
+      setupUserSettingsPage(contentManagementSetup);
+
+      core.application.registerAppUpdater(this.appUpdater$);
```
SuZhou-Joe, line 132:
```suggestion
```
No need to registerAppUpdater for all of the apps because you have already add the updater$ when register the application.

src/plugins/advanced_settings/public/plugin.ts
```diff
@@ -116,6 +126,44 @@ export class AdvancedSettingsPlugin
...
+          const [coreStart, { contentManagement, navigation }] = await core.getStartServices();
+
+          return renderUserSettingsApp(params, { ...coreStart, contentManagement, navigation });
+        },
+      });
+
+      core.chrome.navGroup.addNavLinksToGroup(DEFAULT_NAV_GROUPS.settingsAndSetup, [
+        {
+          id: 'user_settings',
+          order: 101,
```
SuZhou-Joe, line 162:
```suggestion
          order: 101, // user_settings should sit between `application settings` and `Assets`.
```

src/plugins/advanced_settings/public/plugin.ts
```diff
@@ -116,6 +126,44 @@ export class AdvancedSettingsPlugin
...
+      setupUserSettingsPage(contentManagementSetup);
+
+      core.application.registerAppUpdater(this.appUpdater$);
+
+      const userSettingTitle = i18n.translate('advancedSettings.userSettingsLabel', {
+        defaultMessage: 'User settings',
+      });
+
+      core.application.register({
+        id: 'user_settings',
```
SuZhou-Joe, line 139: Nit: the id is used in 3 places, better to assign to a variable.

src/plugins/advanced_settings/server/plugin.ts
```diff
@@ -34,34 +34,73 @@ import {
...
+    const globalConfig = await this.globalConfig$.pipe(first()).toPromise();
+    const isPermissionControlEnabled = globalConfig.savedObjects.permission.enabled === true;
+
+    const userUiSettingsClientWrapper = new UserUISettingsClientWrapper(
+      this.logger,
+      isPermissionControlEnabled
+    );
+    this.userUiSettingsClientWrapper = userUiSettingsClientWrapper;
+    core.savedObjects.addClientWrapper(
+      3,
```
SuZhou-Joe, line 78:
```suggestion
      3, // The wrapper should be triggered after workspace_id_consumer wrapper to avoid creating user settings within any workspace.
```

src/plugins/advanced_settings/server/plugin.ts
```diff
@@ -34,34 +34,73 @@ import {
...
+      'user_ui_settings',
+      userUiSettingsClientWrapper.wrapperFactory
+    );
+
+    core.capabilities.registerSwitcher(async (request, capabilities) => {
+      const userName = extractUserName(request, this.coreStart);
```
SuZhou-Joe, line 84: There is a getPrincipalFromRequest util in #7959 and we can try using that util to get user_name.

src/core/server/ui_settings/ui_settings_client.ts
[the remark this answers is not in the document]
Hailong-am: > What if a setting is both `global` and `user`? This implementation implies a setting can only be `global` or `user`.
> 
> This is what I proposed in the RFC, any reason why this didn't work?
> 
> ## Reading a setting with scope
> When calling uiSettings.get(key), it returns the value based on the following rules:
> 
> 1. If the setting was registered with scope global or empty, the global setting will be returned.
> 2. If the setting was registered with scope user, the user setting will be returned
> 3. If the setting was registered with both user and global, the user setting will be returned if it's set, otherwise, return the global setting
> 4. If calling uiSettings.get(key, defaultValue, 'user') with scope specified, it will return the setting of the specified scope. If the value doesn't not exist, then return the defaultValue. This requires to update the current uiSettings service's interface
> 
> ## Writing a setting with scope
> When calling uiSettings.set(key, value), it follows the following rules:
> 
> 1. If the setting was registered with scope global or empty, it will write the value to global setting saved object.
> 2. If the setting was registered with scope user, it will write the value to user setting saved object.
> 3. If the setting was registered with both user and global, it will by default update the user setting saved object.
> 4. If calling uiSettings.set(key, value, 'global') with scope specified, it will write the value to the corresponding saved object. This requires to update the current uiSettings service's interface

If a setting is both `global` and `user`, it will set to user setting by default, when get it they will merge, and returns same value when set. 

This PR implementation didn't implement item 4 for both read and write with specified scope that requires client to know which setting is global and which is user. I agree that will give more flexibility to control how we are going to store the setting but that also need additional validation make sure changes are align with specified scope.


To support item 4, that will be similar change as you suggested plus router accept scope parameter and scope validation.

As the default behavior could handle most of case, that's the reason i didn't implement it this PR.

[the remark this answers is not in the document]
Hailong-am: scope added to both get and set method

src/core/server/ui_settings/ui_settings_client.ts
Hailong-am: update to scope as suggested, it's more flexible. thanks.

src/plugins/advanced_settings/public/plugin.ts
Hailong-am: variable name updated

src/core/server/ui_settings/create_or_upgrade_saved_config/create_or_upgrade_saved_config.ts
```diff
@@ -35,24 +35,28 @@ import { SavedObjectsErrorHelpers } from '../../saved_objects/';
 import { Logger } from '../../logging';
 
 import { getUpgradeableConfig } from './get_upgradeable_config';
+import { generateDocId } from '../ui_settings_client';
```
ruanyl, line 38: Seems `create_or_upgrade_saved_config` and `ui_settings_client` import from each other, maybe move `generateDocId` to somewhere else?

src/plugins/advanced_settings/server/plugin.ts
Hailong-am: sure, i will copy the file to my PR at this moment to unblock

src/core/server/ui_settings/create_or_upgrade_saved_config/create_or_upgrade_saved_config.ts
Hailong-am: good catch, move it to utils

src/plugins/advanced_settings/server/plugin.ts
Hailong-am: comments added

src/plugins/advanced_settings/public/management_app/advanced_settings.tsx
Hailong-am: sure, enum added

Hailong-am:
two cypress test failure are related to data-explorer, not with this change

SuZhou-Joe (dismissed):

src/plugins/advanced_settings/server/saved_objects/user_ui_settings_client_wrapper.ts
```diff
@@ -0,0 +1,150 @@
...
+          };
+
+          // remove buildNum from attributes
+          const { buildNum, ...others } = attributes as any;
+
+          // create with reference, the reference field will used for filter settings by user
+          return await wrapperOptions.client.create(type, others, {
+            ...options,
+            id: docId,
+            references: [
```
ruanyl, line 114: Do we still need to add this `references`

Hailong-am: It's used to differentiate it's user setting not application settings. It's not mandatory, remove it now.

src/core/server/ui_settings/utils.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { UiSettingScope } from './types';
+
+export const CURRENT_USER_PLACEHOLDER = '<current_user>';
```
ruanyl, line 8: I have an impression that the current `request` can injected into uiSettingClient when I was doing the investigation,  I will take another look, but I think that won't block this PR.

src/plugins/workspace/common/types.ts
```diff
@@ -27,3 +27,9 @@ export interface DataSourceConnection {
   description?: string;
   relatedConnections?: DataSourceConnection[];
 }
+
+export enum PermissionModeId {
```
ruanyl, line 31: Perhaps we can use the `PermissionModeId` from `src/core/types/workspace.ts`?

src/core/types/workspace.ts
```diff
@@ -19,4 +25,5 @@ export interface WorkspaceAttribute {
 
 export interface WorkspaceAttributeWithPermission extends WorkspaceAttribute {
   permissions?: Permissions;
+  permissionMode?: PermissionModeId;
```
ruanyl, line 28: Q: Is this the permission of current user to the workspace?

src/plugins/advanced_settings/public/management_app/advanced_settings.tsx
```diff
@@ -172,6 +173,7 @@ export class AdvancedSettingsComponent extends Component<
     const all = config.getAll();
     const userSettingsEnabled = config.get('theme:enableUserControl');
     return Object.entries(all)
+      .filter((setting) => setting[1].scope !== UiSettingScope.USER)
```
ruanyl, line 176:
```suggestion
      .filter(([, setting]) => setting.scope !== UiSettingScope.USER)
```

ruanyl (approved):

src/core/types/workspace.ts
Hailong-am: Yes, it is permission for current user which shows here 
image

src/plugins/workspace/common/types.ts
Hailong-am: I have trouble to combine them previously, don't remember the reason, but now it works. anyway, combine them into one.

SuZhou-Joe (approved):

ruanyl (approved):

Hailong-am:
i will do manual backport due to 

```ts
 const enableUserControl = core.uiSettings.get('theme:enableUserControl');
    if (enableUserControl) {
      setupTopNavThemeButton(core, core.uiSettings.get('home:useNewHomePage'));
    }
```
 not available in 2.x

AMoo-Miki:
Manually backported to 2.x with https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8057

PR #7961, opened 2024-09-02
[Workspace] Refactor 'Associate data sources' in create page to support direct query connections

Kapian1234:
### Description
Refactor 'Associate data sources' in create page to support direct query connections

## Screenshot
Screenshot 2024-08-13 at 17 18 24
Screenshot 2024-09-02 at 16 13 12

## Changelog

- feat: Support DQCs in create page

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/workspace/public/components/workspace_detail/data_source_connection_table.tsx
```diff
@@ -174,27 +182,26 @@ export const DataSourceConnectionTable = ({
       },
     },
     {
-      width: '10%',
+      width: '15%',
```
wanglam, line 185: @Kapian1234 Seems we are re-use the same component with workspace detail side. Could you help check with @yubonluo if this width updates can be applied in workspace detail? I'm prefer to add `flag` here to determine if we need to update the width.

src/plugins/workspace/public/components/workspace_form/select_data_source_panel.tsx
```diff
@@ -3,149 +3,171 @@
...
+        ({ connectionType }) => connectionType === DataSourceConnectionType.OpenSearchConnection
+      )
+      .map(({ id, type, name, description }) => ({
+        id,
+        title: name,
+        description,
+        dataSourceEngineType: type as DataSourceEngineType,
+      }));
+    const savedDataSources: DataSource[] = [...assignedDataSources, ...dataSources];
+    onChange(savedDataSources);
```
wanglam, line 66: Can we call `setAssignedDataSourceConnections` immediately in this function? The selected connections in table need to wait for fetching connections after data source change.

src/plugins/workspace/public/components/workspace_form/select_data_source_panel.tsx
```diff
@@ -3,149 +3,171 @@
...
-            title: '',
-            id: '',
-          };
-      const newSelectedOptions = [...selectedDataSources];
-      newSelectedOptions.splice(index, 1, newOption);
+  const handleUnassignDataSources = (dataSourceConnections: DataSourceConnection[]) => {
+    const savedDataSources = (assignedDataSources ?? [])?.filter(
+      ({ id }: DataSource) => !dataSourceConnections.some((item) => item.id === id)
+    );
+    onChange(savedDataSources);
```
wanglam, line 73: The same as above comments, prefer call `setAssignedDataSourceConnections` immediately in this function.

src/plugins/workspace/public/components/workspace_detail/data_source_connection_table.tsx
Kapian1234: The width updates will indeed be applied in workspace detail, but I've checked with @yubonluo and the new width still presents well there. Perhaps the flag isn't necessary?

src/plugins/workspace/public/components/workspace_form/select_data_source_panel.tsx
Kapian1234: Alright, maybe the `useEffect` here can be removed.
```
  useEffect(() => {
    fetchDataSourceConnections(assignedDataSources, http, notifications).then((connections) => {
      setAssignedDataSourceConnections(connections);
    });
  }, [assignedDataSources, http, notifications]);
```

src/plugins/workspace/public/components/workspace_creator/workspace_creator_form.tsx
```diff
@@ -176,8 +176,9 @@ export const WorkspaceCreatorForm = (props: WorkspaceCreatorFormProps) => {
                 errors={formErrors.selectedDataSources}
                 onChange={setSelectedDataSources}
                 savedObjects={savedObjects}
-                selectedDataSources={formData.selectedDataSources}
+                assignedDataSources={formData.selectedDataSources}
                 data-test-subj={`workspaceForm-dataSourcePanel`}
+                isDashboardAdmin={true}
```
SuZhou-Joe, line 181:
```suggestion
                isDashboardAdmin={true}
```
Is it by intension?

src/plugins/workspace/public/components/workspace_detail/data_source_connection_table.tsx
```diff
@@ -27,26 +26,40 @@ import { i18n } from '@osd/i18n';
...
+  onSelectedItems?: (dataSources: DataSourceConnection[]) => void;
+  inCreatePage?: boolean;
 }
 
 export const DataSourceConnectionTable = ({
   isDashboardAdmin,
   connectionType,
   dataSourceConnections,
   handleUnassignDataSources,
+  onSelectedItems,
```
SuZhou-Joe, line 46:
```suggestion
  onSelectItems,
```
Nit:

src/plugins/workspace/public/components/workspace_detail/data_source_connection_table.tsx
```diff
@@ -152,19 +165,20 @@ export const DataSourceConnectionTable = ({
...
         defaultMessage: 'Title',
       }),
       truncateText: true,
       render: (name: string, record) => {
         const origin = window.location.origin;
+        const basePath = http.basePath.serverBasePath;
         let url: string;
         if (record.connectionType === DataSourceConnectionType.OpenSearchConnection) {
-          url = `${origin}/app/dataSources/${record.id}`;
+          url = `${origin}${basePath}/app/dataSources/${record.id}`;
```
SuZhou-Joe, line 179: Can we use `basePath.prepend(`/app/dataSources/${record.id}`, { withoutClientBasePath: true })`?

src/plugins/workspace/public/components/workspace_creator/workspace_creator_form.tsx
Kapian1234: Yes, it's always true here, otherwise the component wouldn't be mounted. Above, there's `isDashboardAdmin && isDataSourceEnabled && (`

src/plugins/workspace/public/components/workspace_detail/data_source_connection_table.tsx
Kapian1234: `basePath` is a string and there's no `prepend` in `basePath`

src/plugins/workspace/public/components/workspace_detail/data_source_connection_table.tsx
Kapian1234: ok

src/plugins/workspace/public/components/workspace_creator/workspace_creator_form.tsx
ruanyl: I suggest we rename `isDashboardAdmin` to `showDataSourceManagement` or something like that, because it's a UI behavior, `isDashboardAdmin` has been widely used as a term that related to permission

Kapian1234: But in this SelectDataSourcePanel component, `isDashboardAdmin` will continue to be passed to components like DataSourceConnectionTable, where there are some behaviors like 'select' and 'delete' do indeed need permissions verification. Using `showDataSourceManagement` during the pass may cause confusion.
In SelectDataSourcePanel:
```
  const renderTableContent = () => {
    return (
      <EuiPanel paddingSize="none" hasBorder={false}>
        <DataSourceConnectionTable
          isDashboardAdmin={isDashboardAdmin}
          dataSourceConnections={assignedDataSourceConnections}
          handleUnassignDataSources={handleUnassignDataSources}
          onSelectItems={getSelectedItems}
          inCreatePage={true}
          connectionType={AssociationDataSourceModalMode.OpenSearchConnections}
        />
      </EuiPanel>
    );
  };
```

src/plugins/workspace/public/components/workspace_form/select_data_source_panel.tsx
```diff
@@ -3,149 +3,169 @@
...
   errors?: { [key: number]: WorkspaceFormError };
   savedObjects: SavedObjectsStart;
-  selectedDataSources: DataSource[];
+  assignedDataSources: DataSource[];
   onChange: (value: DataSource[]) => void;
+  isDashboardAdmin: boolean;
 }
 
 export const SelectDataSourcePanel = ({
   errors,
```
ruanyl, line 28: Seems `errors` is not used any more, is it by intention?

src/plugins/workspace/public/components/workspace_form/select_data_source_panel.tsx
```diff
@@ -3,149 +3,169 @@
...
+        {assignedDataSources.length > 0 && renderTableContent()}
+      </EuiFlexItem>
+      {modalVisible && chrome && (
+        <AssociationDataSourceModal
+          savedObjects={savedObjects}
+          assignedConnections={assignedDataSourceConnections}
+          closeModal={() => setModalVisible(false)}
+          handleAssignDataSourceConnections={handleAssignDataSources}
+          http={http}
+          mode={toggleIdSelected as AssociationDataSourceModalMode}
```
ruanyl, line 164: Seems you don't need `as AssociationDataSourceModalMode`

src/plugins/workspace/public/components/workspace_creator/workspace_creator_form.tsx
ruanyl: I see, so that's an existing props. I think you can change the `props.isDashboardAdmin` of `SelectDataSourcePanel` because this props is newly aded, let's fix this first if fixing `DataSourceConnectionTable` is out of the scope of this PR.

src/plugins/workspace/public/components/workspace_detail/data_source_connection_table.tsx
```diff
@@ -27,26 +26,40 @@ import { i18n } from '@osd/i18n';
...
     Record<string, React.ReactNode>
   >({});
+  const {
+    services: { http },
+  } = useOpenSearchDashboards<CoreStart>();
+  useEffect(() => {
+    if (onSelectItems) {
+      onSelectItems(selectedItems);
+    }
+  }, [selectedItems, onSelectItems]);
```
ruanyl, line 62: You can have this logic in `onSelectionChange` then you don't need this `useEffect`, then everywhere currently calls `setSelectedItems`, you can use `onSelectionChange`

src/plugins/workspace/public/components/workspace_form/select_data_source_panel.tsx
```diff
@@ -3,149 +3,169 @@
...
+          notifications={notifications}
+          logos={chrome?.logos}
```
ruanyl, line 166: Nit: you checked the existence of `chrome` already, optional chain is not required here
```suggestion
          logos={chrome.logos}
```

src/plugins/workspace/public/components/workspace_creator/workspace_creator_form.tsx
Kapian1234: ok, so in SelectDataSourcePanel it would become:
```
        <DataSourceConnectionTable
          isDashboardAdmin={showDataSourceManagement}
          ...
        />
```

src/plugins/workspace/public/components/workspace_detail/data_source_connection_table.tsx
```diff
@@ -267,12 +280,17 @@ export const DataSourceConnectionTable = ({
                 icon: 'unlink',
                 type: 'icon',
                 onClick: (item: DataSourceConnection) => {
-                  setSelectedItems([item]);
-                  setModalVisible(true);
+                  if (inCreatePage) {
```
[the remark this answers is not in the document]
Kapian1234: yes

src/plugins/workspace/public/components/workspace_form/select_data_source_panel.tsx
Kapian1234: ok

src/plugins/workspace/public/components/workspace_detail/data_source_connection_table.tsx
Kapian1234: ok, thx!

src/plugins/workspace/public/components/workspace_form/select_data_source_panel.tsx
Kapian1234: thx I forgot to delete it

src/plugins/workspace/public/components/workspace_detail/data_source_connection_table.tsx
```diff
@@ -27,26 +26,40 @@ import { i18n } from '@osd/i18n';
...
+  inCreatePage?: boolean;
 }
 
 export const DataSourceConnectionTable = ({
   isDashboardAdmin,
   connectionType,
   dataSourceConnections,
   handleUnassignDataSources,
+  onSelectItems,
+  inCreatePage = false,
```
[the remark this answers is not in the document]
Kapian1234: ok

src/plugins/workspace/public/components/workspace_form/select_data_source_panel.tsx
Kapian1234: no, I forgot to delete it

src/plugins/workspace/public/components/workspace_detail/data_source_connection_table.tsx
```diff
@@ -285,22 +303,34 @@ export const DataSourceConnectionTable = ({
...
-        isSelectable={true}
-        itemIdToExpandedRowMap={itemIdToExpandedRowMap}
-        isExpandable={true}
-        pagination={{
-          initialPageSize: 10,
-          pageSizeOptions: [10, 20, 30],
-        }}
-      />
-      <EuiSpacer />
+      {inCreatePage ? (
```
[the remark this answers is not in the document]
Kapian1234: Another difference is that we show search bar when not in create page, then we have to pass two props. The search props looks like this, we may not need to pass it.
```
   const renderToolsLeft = useCallback(() => {
    return selectedItems.length > 0 && !modalVisible
      ? [
          <EuiSmallButton
            color="danger"
            onClick={() => setModalVisible(true)}
            data-test-subj="workspace-detail-dataSources-table-bulkRemove"
          >
            {i18n.translate('workspace.detail.dataSources.table.remove.button', {
              defaultMessage: 'Remove {numberOfSelect} association(s)',
              values: { numberOfSelect: selectedItems.length },
            })}
          </EuiSmallButton>,
        ]
      : [];
  }, [selectedItems, modalVisible]);

  const search: EuiSearchBarProps = {
    toolsLeft: renderToolsLeft(),
    box: {
      incremental: true,
    },
    filters: [
      {
        type: 'field_value_selection',
        field: 'type',
        name: 'Type',
        multiSelect: 'or',
        options: Array.from(
          new Set(openSearchConnections.map(({ type }) => type).filter(Boolean))
        ).map((type) => ({
          value: type!,
          name: type!,
        })),
      },
    ],
  };
```

src/plugins/workspace/public/components/workspace_detail/data_source_connection_table.tsx
SuZhou-Joe: right, please use http.basePath to get the final path. We should avoid string concatenation when generating the path.

src/plugins/workspace/public/components/workspace_detail/data_source_connection_table.tsx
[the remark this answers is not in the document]
SuZhou-Joe: If we are trying to make this component a reusable component, we should let the parent of the component to pass in different props to control the behavior of the component. For this component, I think we should split this component into 2 components:
1. A table component that renders the built-in columns, and gives a `onUnlinkDatasource` callback.
2. The ConfirmModal and all the different props should be handled by workspace details page and workspace creation page.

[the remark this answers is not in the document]
wanglam: Agree with @SuZhou-Joe 's comments. Current implementation of `DataSourceConnectionTable` contains too much business logic should handled by workspace detail or create page self.  I've updated the code. Feel free to help me review it. Thank you.

src/plugins/workspace/public/components/workspace_form/data_source_connection_table.tsx
```diff
@@ -152,19 +107,25 @@ export const DataSourceConnectionTable = ({
...
-          url = `${origin}/app/dataSources/${record.id}`;
+          url = http.basePath.prepend(`/app/dataSources/${record.id}`, {
+            withoutClientBasePath: true,
+          });
         } else {
-          url = `${origin}/app/dataSources/manage/${name}?dataSourceMDSId=${record.parentId}`;
+          url = `${origin}${basePath}/app/dataSources/manage/${name}?dataSourceMDSId=${record.parentId}`;
+          url = http.basePath.prepend(`/app/dataSources/${record.id}`, {
+            withoutClientBasePath: true,
+          });
```
SuZhou-Joe, line 128:
```suggestion
          url = http.basePath.prepend(`/app/dataSources/manage/${name}?dataSourceMDSId=${record.parentId}`, {
            withoutClientBasePath: true,
          });
```

src/plugins/workspace/public/components/workspace_form/data_source_connection_table.tsx
```diff
@@ -152,19 +107,25 @@ export const DataSourceConnectionTable = ({
...
       truncateText: true,
       render: (name: string, record) => {
         const origin = window.location.origin;
+        const basePath = http.basePath.serverBasePath;
         let url: string;
         if (record.connectionType === DataSourceConnectionType.OpenSearchConnection) {
-          url = `${origin}/app/dataSources/${record.id}`;
+          url = http.basePath.prepend(`/app/dataSources/${record.id}`, {
+            withoutClientBasePath: true,
+          });
```
SuZhou-Joe, line 123:
```suggestion
          url = http.basePath.prepend(`/app/dataSources/${record.id}`);
```
If it's opensearch connection, shall we just go to the data source page within the workspace?

src/plugins/workspace/public/components/workspace_form/data_source_connection_table.tsx
```diff
@@ -3,106 +3,61 @@
...
+type DataSourceConnectionTableProps = Omit<
+  EuiInMemoryTableProps<DataSourceConnection>,
+  | 'columns'
+  | 'itemsId'
+  | 'isSelectable'
+  | 'itemIdToExpandedRowMap'
+  | 'isExpandable'
+  | 'selection'
+  | 'pagination'
+> & {
```
SuZhou-Joe, line 40:
```suggestion
type DataSourceConnectionTableProps = {
  tableProps: Partial<EuiInMemoryTableProps<DataSourceConnection>>
```
What about we moving all the table related props to a standalone entry so that it won't bother other props. And to make the props simple by using Partial.

src/plugins/workspace/public/components/workspace_form/data_source_connection_table.tsx
```diff
@@ -284,48 +244,14 @@ export const DataSourceConnectionTable = ({
...
-          confirmButtonText={i18n.translate('workspace.detail.dataSources.Modal.confirmButton', {
-            defaultMessage: 'Remove data source(s)',
-          })}
-          buttonColor="danger"
-          defaultFocusedButton="confirm"
-        />
-      )}
-    </>
+    <EuiInMemoryTable
+      {...(restProps as EuiInMemoryTableProps<DataSourceConnection>)}
```
SuZhou-Joe, line 248:
```suggestion
      {...restProps}
```
We'd better avoid type assertion.

src/plugins/workspace/public/components/workspace_form/data_source_connection_table.tsx
wanglam: I think we should go to the data source management page within the workspace. @yubonluo , could you help clarify this?

ruanyl (approved):

src/plugins/workspace/public/components/workspace_creator/workspace_creator_form.tsx
ruanyl: @wanglam @Kapian1234 maybe we can have a follow up task to align this prop name among these components.

wanglam: I agree that we should have a follow-up task update these components. The `isDashboardAdmin` in `DataSourceConnectionTable` controls two behaviors. The first one is to control whether table row can be selectable. The second one is to control whether data source connection can be unassociated. Shall we need break it into two properties?

src/plugins/workspace/public/components/workspace_form/select_data_source_panel.tsx
```diff
@@ -3,149 +3,166 @@
...
+
+  const handleUnassignDataSources = (dataSourceConnections: DataSourceConnection[]) => {
+    const savedDataSourcesConnections = (assignedDataSourceConnections ?? [])?.filter(
+      ({ id }: DataSourceConnection) => !dataSourceConnections.some((item) => item.id === id)
+    );
+    setAssignedDataSourceConnections(savedDataSourcesConnections);
+    const savedDataSources = (assignedDataSources ?? [])?.filter(
+      ({ id }: DataSource) => !dataSourceConnections.some((item) => item.id === id)
+    );
+    onChange(savedDataSources);
```
SuZhou-Joe, line 70: May I know in which case the `assignedDataSourceConnections` will not equals to `assignedDataSources`?

ruanyl (approved):

SuZhou-Joe (approved):

SuZhou-Joe:
`Run cypress tests (osd:ciGroup6)` should be irrelevant to the code changes.

PR #7991, opened 2024-09-04
[Auto Suggest] Look and Feel changes

paulstn:
## Videos

Light Mode:

https://github.com/user-attachments/assets/a4123882-8532-4a4d-90b7-aa118c93208a

Dark Mode:

https://github.com/user-attachments/assets/58fe0391-25d1-4321-a470-a80dd48751d0

# Changes

## Keeping the suggestion window open

This PR aims to keep the suggestion window open as long as possible by doing the following three things:
1. Triggers the suggestion window to open whenever the editor is 'focused' (clicked on)
2. Triggers the suggestion window after the user types a space
3. Triggers the suggestion window after a suggestion has been chosen, thereby showing the next suggested option immediately

Part of the behavior that already exists is that the suggestion window will close when the user presses ESC

## User hints below suggestions

This change appends user hints below the suggestion window: `Tab to insert, ESC to close window`. It does this through scss

## DQL Operators

`AND`, `OR`, and `NOT` have been distinguished as operators - the details and the icon have been changed

image

Also, DQL Autocomplete was modified slightly to work better with change number 3. PPL and SQL will also have to be modified, because currently the cursor is still sitting on the previous word so the next suggestions aren't available until a space is put.

## Changelog

- feat: Keep Autocomplete suggestion window open and put user hints below the suggestion window


ashwin-pc (approved):
Nice!

src/plugins/data/public/ui/query_editor/_query_editor.scss
```diff
@@ -199,3 +199,24 @@
     border: none;
   }
 }
+
+.suggest-widget {
+  position: relative;
+
+  &.visible::after {
```
paulstn, line 206: I'm appending the box onto the `suggest-widget` class, which is what's used by the monaco editor package. Also additionally using the `visible` class to only display it when the suggestion window is visible

kavilla: instead of this, could we create an element on the code_editor?

for example

```
_editorDidMount = (editor: monaco.editor.IStandaloneCodeEditor, __monaco: unknown) => {
 ...
  const suggestWidget = editor.getContribution('editor.contrib.suggestController')?.widget;
  if (suggestWidget) {
    const originalRender = suggestWidget.render;
    suggestWidget.render = (element: HTMLElement) => {
      originalRender.call(suggestWidget, element);
      
      let messageElement = element.querySelector('.suggest-widget-message');
      if (!messageElement) {
        messageElement = document.createElement('div');
        messageElement.className = 'suggest-widget-message';
        messageElement.textContent = 'Tab to insert, ESC to close window';
        element.appendChild(messageElement);
      }
    };
  }

...
};

**Review by ashwin-pc — DISMISSED** [review]

File: src/plugins/data/public/ui/query_editor/_query_editor.scss (hunk 1 above)
Comment (paulstn) [reply src/plugins/data/public/ui/query_editor/_query_editor.scss:206]: I tried this method out first but it didn't work for me. I could try to revisit it, but part of the issue could be that the version of monaco that we're using is old so it may not work the same compared to if it was up to date.

**Review by ashwin-pc — APPROVED** [review]

File: src/plugins/data/public/ui/query_editor/_query_editor.scss (hunk 1 above)
Comment (kavilla) [reply src/plugins/data/public/ui/query_editor/_query_editor.scss:206]: i see.

for  the styling though could we use the standard values?

File: src/plugins/data/public/ui/query_editor/_query_editor.scss [hunk 2]
```diff
@@ -199,3 +199,24 @@
...
+    position: absolute;
+    bottom: -30px;
+    left: 0;
+    width: 100%;
+    height: 30px;
+    background-color: $euiColorLightestShade;
+    border: 1px solid $euiColorLightShade;
+    padding: 5px;
```
kavilla, line 214: for example should this be
```
padding: $euiSizeS
```

ashwin-pc: Oh shoot, good catch. The border should also be an eui prop. `$euiBorderThin`, as for padding it should be `$euiSizeXS`. And why do we have padding 5px all around and 4px in the top?

And for font use `@include ouiFontSizeXS`

paulstn: Modified the padding, border, and font size to all use variables. As well as the height and the 'bottom' attribute.

changelogs/fragments/7991.yml
```diff
@@ -0,0 +1,6 @@
+feat:
+- Keep Autocomplete suggestion window open ([#7991](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7991))
```
kavilla, line 2: this fragment should just be one thing as it will be squashed

kavilla (approved):

sejli (approved):

PR #7992, opened 2024-09-04
[Workspace] Hide first home breadcrumbs within workspace

Hailong-am:
### Description

Hide the home breadcrumbs when we are in a workspace, but keep it in recent popover

image

## Changelog

- feat: [Workspace] Hide home breadcrumbs when in a workspace

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/public/chrome/ui/header/header_breadcrumbs.tsx
+++ b/src/core/public/chrome/ui/header/header_breadcrumbs.tsx
@@ -75,7 +75,7 @@ export function HeaderBreadcrumbs({
     crumbs = breadcrumbEnricher(crumbs);
   }
 
-  if (dropHomeFromBreadcrumb && Object.hasOwn(crumbs[0], 'home')) {
+  if (dropHomeFromBreadcrumb && crumbs.length && Object.hasOwn(crumbs[0], 'home')) {
     crumbs = crumbs.slice(1);
   }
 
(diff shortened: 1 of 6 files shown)


Hailong-am:
@AMoo-Miki @zhongnansu can you help to review this change?

src/core/public/chrome/ui/header/header_breadcrumbs.tsx
```diff
@@ -73,6 +75,10 @@ export function HeaderBreadcrumbs({
     crumbs = breadcrumbEnricher(crumbs);
   }
 
+  if (dropHomeFromBreadcrumb && crumbs.length && Object.hasOwn(crumbs[0], 'home')) {
```
ruanyl, line 78: Do we need to check if home is true?

Hailong-am: we could add this condition as well. There only have one place to set this additional property, i think it's ok only check existence.

zhongnansu (approved):
lgtm

SuZhou-Joe (approved):

src/core/public/chrome/ui/header/header_breadcrumbs.tsx
ruanyl: Can we have the assumption that when `home` is set, it must be true?

ruanyl (approved):

SuZhou-Joe:
`Run cypress tests (osd:ciGroup6)` should be irrelevant to the code changes.

PR #7998, opened 2024-09-04
[query assist] use badge to show agent errors

joshuali925:
### Description

this PR fixes some bugs caused by dataset and ml-commons api change, and adds error to badge on UI

- enable query assist for index patterns without data source 
- add agent error parsing utils 
- update ml-commons response schema processing 
   - previously ml-commons returns error.body as a string, now it is a JSON object. Ideally frontend should keep it as is to reduce serializing/deserializing, but since older version of ml-commons can be used through MDS, we'll keep it as a string for consistency
- use badge to show query assist error if possible 
- add unit tests

## Screenshot
image

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/query_enhancements/public/query_assist/utils/errors.ts
```diff
@@ -12,6 +13,33 @@ export class ProhibitedQueryError extends Error {
   }
 }
 
+interface IAgentError {
+  error: {
```
kavilla, line 17: this just got merged today but maybe we could use this: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/query/query_string/language_service/lib/query_result.tsx#L21

src/plugins/query_enhancements/public/query_assist/utils/errors.ts
```diff
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+// eslint-disable-next-line max-classes-per-file
 import { ResponseError } from '@opensearch-project/opensearch/lib/errors';
```
kavilla, line 7: nit: should we use the core type? https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/http/router/response.ts#L42

src/plugins/query_enhancements/public/query_assist/components/warning_badge.tsx
```diff
@@ -0,0 +1,72 @@
...
+interface WarningBadgeProps {
+  error: AgentError | undefined;
+}
+
+export const WarningBadge: React.FC<WarningBadgeProps> = (props) => {
+  const [isPopoverOpen, setIsPopoverOpen] = useState(false);
+  const [showMore, setShowMore] = useState(false);
+
+  if (!props.error) return null;
+  const error = props.error.error.error;
```
kavilla, line 19: nit: maybe

```typescript
const { error: { error } } = props.error
```

src/plugins/query_enhancements/public/query_assist/components/warning_badge.tsx
```diff
@@ -0,0 +1,72 @@
...
+        </EuiBadge>
+      }
+      isOpen={isPopoverOpen}
+      closePopover={() => setIsPopoverOpen(false)}
+      className="queryAssist__popover"
+    >
+      <EuiText size="s" className="queryAssist__popoverText">
+        <dl>
+          <dd id="queryAssistErrorTitle">
+            <b>Error</b>
```
kavilla, line 47: do we need i18n?

kavilla (dismissed):

kavilla:
cool stuff. just nitpicks but looks good. i think we should wrap around post 2.17 to make this component merge with the status component and make it generic so that it is reusable. since i can see this being very useful for other stuff

src/plugins/query_enhancements/public/query_assist/utils/errors.ts
joshuali925: i think that one is only used for async query. there are some differences since ml-commons error will have `type` but won't have `status`

src/plugins/query_enhancements/public/query_assist/utils/errors.ts
joshuali925: i think it should be the one from js library which has the `body` field. the core type gives a lot of lint errors since there's no `body` field https://github.com/opensearch-project/OpenSearch-Dashboards/blob/84d992ab0be48ef037b2c69baa5cfc9e77d64dd1/src/core/server/http/router/response.ts#L42-L48

and core type is in server, i don't think we can import it here

src/plugins/query_enhancements/public/query_assist/components/warning_badge.tsx
joshuali925: good point, forgot about i18n, updated

src/plugins/data/public/query/query_string/query_string_manager.mock.ts
```diff
@@ -44,10 +45,10 @@ const createSetupContractMock = (isEnhancementsEnabled: boolean = false) => {
   };
 
   const queryStringManagerMock: jest.Mocked<QueryStringContract> = {
-    getQuery: jest.fn(),
+    getQuery: jest.fn().mockReturnValue(defaultQuery),
     setQuery: jest.fn(),
-    getUpdates$: jest.fn(),
-    getDefaultQuery: jest.fn(),
+    getUpdates$: jest.fn().mockReturnValue(of(defaultQuery)),
```
kavilla, line 50: nice.

but i think one of the tests will fail now. so might have to just update the test to expect this now.

kavilla (commented):
dang nice! one comment

src/plugins/data/public/query/query_string/query_string_manager.mock.ts
joshuali925: weird i ran `yarn test:jest` before pushing, they passed locally

joshuali925:
@kavilla there is a cypress failure on cigroup6

```
  2 failing
  1) discover_advanced_setting
       doc_table:highlight advanced setting
         check new table respects doc_table:highlight setting:
     AssertionError: Timed out retrying after 60000ms: Expected to find element: `mark`, but never found it.
      at Context.eval (http://localhost:5601/__cypress/tests?p=cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover_advanced_settings.spec.js:443:22)
  2) discover_advanced_setting
       doc_table:highlight advanced setting
         check legacy table respects doc_table:highlight setting:
     AssertionError: Timed out retrying after 60000ms: Expected to find element: `mark`, but never found it.
      at Context.eval (http://localhost:5601/__cypress/tests?p=cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover_advanced_settings.spec.js:459:22)
  (Results)
  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ Tests:        18                                                                               │
  │ Passing:      15                                                                               │
  │ Failing:      2                                                                                │
  │ Pending:      1                                                                                │
  │ Skipped:      0                                                                                │
  │ Screenshots:  2                                                                                │
  │ Video:        true                                                                             │
  │ Duration:     6 minutes, 33 seconds                                                            │
  │ Spec Ran:     core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover_adv │
  │               anced_settings.spec.js                                                           │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
```

i don't think it's related to my PR though because it also [failed](https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/10725895055/job/29751027314) for #8025, which has no code change

the unit tests all passed

src/plugins/query_enhancements/public/query_assist/_index.scss
```diff
@@ -11,4 +11,19 @@
   &.queryAssist__callout {
     margin-top: $euiSizeXS;
   }
+
+  .queryAssist__badgeContainer {
+    position: fixed;
+    right: 48px;
```
AMoo-Miki, line 17: Where is this used? How did we land on 48? Why is it fixed?

src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
```diff
@@ -91,16 +96,20 @@ export const QueryAssistBar: React.FC<QueryAssistInputProps> = (props) => {
...
-        <EuiFlexGroup gutterSize="s" responsive={false} alignItems="center">
+        <EuiFlexGroup gutterSize="xs" responsive={false} alignItems="center">
           <EuiFlexItem>
             <QueryAssistInput
               inputRef={inputRef}
               persistedLog={persistedLog}
               isDisabled={loading}
               selectedIndex={selectedIndex}
               previousQuestion={previousQuestionRef.current}
+              style={agentError && { paddingRight: 96 }}
```
AMoo-Miki, line 107: Why are we adding inline styles?

src/plugins/query_enhancements/public/query_assist/components/warning_badge.tsx
```diff
@@ -0,0 +1,116 @@
...
+        <EuiBadge
+          color="warning"
+          // @ts-ignore this is needed to avoid enter key triggering warning badge instead of submit button
+          type="button"
+          iconType="alert"
+          onClick={(e: SyntheticEvent) => {
+            e.preventDefault();
+            setIsPopoverOpen(!isPopoverOpen);
+          }}
+          onClickAriaLabel="Click to show details"
```
AMoo-Miki, line 36: Needs to use i18n

src/plugins/query_enhancements/public/query_assist/utils/errors.test.ts
```diff
@@ -42,6 +42,24 @@ describe('formatError', () => {
     expect(formattedError.message).toEqual(error.body.message);
   });
 
+  it('should return an AgentError if possible', () => {
+    const error = new ResponseError({
+      statusCode: 400,
+      body: {
+        statusCode: 400,
+        message: `{"error":{"reason":"Invalid Request","details":"PPLTool doesn't support searching indices starting with '.' since it could be system index, current searching index name: .opensearch-sap-log-types-config","type":"IllegalArgumentException"},"status":400}`,
```
AMoo-Miki, line 50: The last time I spoke with the documentation team, we want to use `indexes` for anything new we are adding. 

@opensearch-project/docs can u plz provide guidance.

src/plugins/query_enhancements/server/routes/query_assist/routes.ts
```diff
@@ -78,13 +78,16 @@ export function registerQueryAssistRoutes(router: IRouter) {
         return response.ok({ body: responseBody });
       } catch (error) {
         if (isResponseError(error)) {
-          if (error.statusCode === 400 && error.body.includes(ERROR_DETAILS.GUARDRAILS_TRIGGERED))
+          if (
+            error.statusCode === 400 &&
+            // on opensearch >= 2.17, error.body is an object https://github.com/opensearch-project/ml-commons/pull/2858
+            JSON.stringify(error.body).includes(ERROR_DETAILS.GUARDRAILS_TRIGGERED)
+          )
```
AMoo-Miki, line 85: This will throw if OSD 2.17 is paired with older OS.

src/plugins/query_enhancements/public/query_assist/utils/errors.test.ts
kolchfa-aws: Yup, our style guide prescribes to use the word "indexes" (unless it's a variable; then "indices" is fine because we've inherited lots of settings/endpoints that use it in this form). So in error messages we should use "indexes".

src/plugins/query_enhancements/public/query_assist/_index.scss
joshuali925: the badge needs to overlay on top of the field text, 48 pixels to the right end. open to suggestions since field text in oui doesn't support this

src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
joshuali925: any issues or guidelines with inline styles? from the question i'm inferring we want to avoid inline styles?

src/plugins/query_enhancements/public/query_assist/utils/errors.test.ts
joshuali925: this is a unit test, the message came from the skills plugin. do we have any campaign to rename existing "indices" to "indexes" in all projects the org? interested to read more on the details

src/plugins/query_enhancements/public/query_assist/_index.scss
AMoo-Miki: We should use the `append` prop of the text field. https://oui.opensearch.org/1.12/#/forms/form-controls#prepend-and-append

src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
AMoo-Miki: Correct. We don't want to have inline styles and also would rather use OUI constructs as much as possible. 

Also, what is the purpose of this right padding?

src/plugins/query_enhancements/public/query_assist/_index.scss
joshuali925: append looks different from mock 
image

src/plugins/query_enhancements/public/query_assist/utils/errors.test.ts
AMoo-Miki: I don't know if there is or was a campaign to replace existing ones but if anything new is being added, we wanna make sure to use `indexes`. The skills plugin sounds new; is it? If yes, we should let them know.

src/plugins/query_enhancements/public/query_assist/_index.scss
AMoo-Miki: It does. @opensearch-project/opensearch-ux can you please provide guidance.

src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
joshuali925: > We don't want to have inline styles and also would rather use OUI constructs as much as possible

are there guidelines like documentation explaining this to developers?

> Also, what is the purpose of this right padding?

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/23ca7d4e59c86e28ac7d7279dda3a6c25c84c3c4/src/plugins/query_enhancements/public/query_assist/_index.scss#L17-L19

src/plugins/query_enhancements/public/query_assist/utils/errors.test.ts
joshuali925: based on [tags](https://github.com/opensearch-project/skills/tags) the oldest skills plugin is for OS_2.12

i think it would be good to have some documentation/guideline to align to

src/plugins/query_enhancements/public/query_assist/_index.scss
```diff
@@ -7,6 +7,29 @@
...
+    // the badge is 48px away from the right edge of the query assist form
+    position: fixed;
+    right: 48px;
+  }
+
+  // the query assist badge is overlayed on top of the query assist input, so
+  // we need to push for some blank space on the right to avoid text
+  // overlapping
+  .queryAssist__withBadge {
+    padding-right: 96px;
```
AMoo-Miki, line 21: Since the text length will be different in different language, this is not a suitable solution.

joshuali925: agree, @ashwin-pc suggested to add a padding to the right to avoid text overlapping with badge. i'm not sure if we want to dynamically calculate, do you know if there's a better solution?

joshuali925: i didn't find a better solution, changed this to dynamically calculate

src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
AMoo-Miki: I pinged @kgcreative offline but can you also plz reach out to them to see if they can revise the design. Of course, we can perhaps go ahead and unblock this PR and deal with it later.

src/plugins/query_enhancements/public/query_assist/_index.scss
AMoo-Miki: You know what...
1. Use `append` to add the warning.
2. Use CSS ti hide the right border of the text field.

With this, it will look identical to what UX wants and we won't need magic numbers or dynamically identified width.

joshuali925: that's a good idea. i tried it the problem is that the suggestion component's width doesn't expand to the `append` section

The other thing is do you know how to simulate `.euiTextField:focus` effect? if i add `euiTextField` class to the badge/popover, it will introduce extra euiTextField styles. if i directly copy the css (with variables), the style won't stay consistent if euiTextField style changes
```scss
.queryAssist__popover {
  &.focused {
    background-color: #fcfeff;
    background-image: linear-gradient(to top, #07827E, #07827E 2px, transparent 2px, transparent 100%);
    background-size: 100% 100%;
  }
}
```

I think even if we apply the same styling, but because the focus effect is an animation, it could be difficult to align the timing. Notice how the `append` section is highlighted first and the green line appears instantly

https://github.com/user-attachments/assets/dbe062d6-2882-4b2f-84d7-475a9636d9ce

src/plugins/query_enhancements/public/query_assist/_index.scss
ashwin-pc: Miki made a good suggestion about using the OUI components prepend option and using CSS to hide the parts that dont look as per mocks

joshuali925: @ashwin-pc see https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7998#discussion_r1752347506

joshuali925:
Thanks @AMoo-Miki for helping. It would be ideal to add a feature in OUI text field so that it can append element inside itself. For now we have 2 workarounds:

1. **the current implementation**: it is hacky because it involves dynamically retrieving badge width and doing math to compute the correct padding. The math won't make sense if for example the layout changes, or text field size changes.
2. **using css to change background on focus**: this is less hacky because it doesn't involve math calculations. The downside is that it doesn't look perfect as the mocks (suggestions not full width and no animation for the append section), but i think it's good enough as a compromise

Recording for option 2:
Sep-10-2024 13-51-20

@kgcreative @kavck do you think the above recording is acceptable behavior?

AMoo-Miki:
Another option is hide the `Error` on focus.

joshuali925:
> Another option is hide the `Error` on focus.

i actually like it better. showing "Error" when user writes the question doesn't really help, and hiding it gives more horizontal space
Sep-10-2024 14-30-32

kgcreative:
There is a very limited subset of `errors` that would show up in the input; i would imagine they are mostly around illegal characters that we might underline in red in some way, or submit or connection errors where there is a failure before a query or a summary is generated

We also need to think about where we might show query errors, or summary errors. Those should be in the query area and the summary area respectively.

src/plugins/query_enhancements/public/query_assist/components/warning_badge.tsx
```diff
@@ -0,0 +1,130 @@
...
+            />
+          </EuiBadge>
+        }
+        isOpen={isPopoverOpen}
+        closePopover={() => setIsPopoverOpen(false)}
+        className="queryAssist__popover"
+      >
+        <EuiText size="s" className="queryAssist__popoverText">
+          <dl>
+            <dd id="queryAssistErrorTitle">
```
AMoo-Miki, line 64: This should use `<EuiIcon type="alert" />` as opposed to CSS.

ashwin-pc (approved):

PR #7999, opened 2024-09-04
Filter Saved Search Based on AppName Supported by Language

LDrago27:
### Description

This PR aims to filter out unsupported Saved Searches (SQL and PPL based Saved Queries are not supported for Visualization) across the application namely in Dashboards and Visualize.

## Screenshot

Saved Search in Discover (contains all SQL, PPL, DQL, Lucene Searches)
image

Saved Search in Dashboards (contains DQL, Lucene based searches)
image

Saved Search in Visualize (contains DQL, Lucene based searches)
image

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/dashboard/public/application/utils/create_dashboard_container.tsx
+++ b/src/plugins/dashboard/public/application/utils/create_dashboard_container.tsx
@@ -282,7 +282,12 @@ const getEmptyScreenProps = (
             getFactory: embeddable.getEmbeddableFactory,
             notifications,
             overlays,
-            SavedObjectFinder: getSavedObjectFinder(savedObjects, uiSettings),
+            SavedObjectFinder: getSavedObjectFinder(
+              savedObjects,
+              uiSettings,
+              services.data,
+              services.application
+            ),
           });
         }
       } else {
--- a/src/plugins/saved_objects/public/finder/saved_object_finder.tsx
+++ b/src/plugins/saved_objects/public/finder/saved_object_finder.tsx
@@ -58,9 +58,12 @@ import {
   CoreStart,
   IUiSettingsClient,
   SavedObjectsStart,
+  ApplicationStart,
 } from 'src/core/public';
 
 import { DataSourceAttributes } from 'src/plugins/data_source/common/data_sources';
+import { DataPublicPluginStart } from 'src/plugins/data/public';
+import { first } from 'rxjs/operators';
 import { getIndexPatternTitle } from '../../../data/common/index_patterns/utils';
 import { LISTING_LIMIT_SETTING } from '../../common';
 
@@ -76,6 +79,7 @@ export interface SavedObjectMetaData<T = unknown> {
 interface FinderAttributes {
   title?: string;
   type: string;
+  kibanaSavedObjectMeta?: string;
 }
 
 interface SavedObjectFinderState {
@@ -122,6 +126,8 @@ export type SavedObjectFinderProps = SavedObjectFinderFixedPage | SavedObjectFin
 export type SavedObjectFinderUiProps = {
   savedObjects: CoreStart['savedObjects'];
   uiSettings: CoreStart['uiSettings'];
+  data: DataPublicPluginStart;
+  application: CoreStart['application'];
 } & SavedObjectFinderProps;
 
 class SavedObjectFinderUi extends React.Component<
@@ -137,10 +143,17 @@ class SavedObjectFinderUi extends React.Component<
     showFilter: PropTypes.bool,
   };
 
+  public async getCurrentAppId() {
+    const appId = await this.props.application.currentAppId$.pipe(first()).toPromise();
+    return appId;
+  }
+
+  readonly languageService = this.props.data.query.queryString.getLanguageService();
   private isComponentMounted: boolean = false;
 
   private debouncedFetch = _.debounce(async (query: string) => {
     const metaDataMap = this.getSavedObjectMetaDataMap();
+    const currentAppId = await this.getCurrentAppId();
 
     const fields = Object.values(metaDataMap)
       .map((metaData) => metaData.includeFields || [])
@@ -162,7 +175,7 @@ class SavedObjectFinderUi extends React.Component<
       return await client.get<DataSourceAttributes>('data-source', id);
     };
 
-    const savedObjects = await Promise.all(
+    let savedObjects = await Promise.all(
       resp.savedObjects.map(async (obj) => {
         if (obj.type === 'index-pattern') {
           const result = { ...obj };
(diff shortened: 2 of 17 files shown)


src/plugins/saved_objects/public/finder/saved_object_finder.tsx
```diff
@@ -178,6 +191,25 @@ class SavedObjectFinderUi extends React.Component<
       })
     );
 
+    // Filter search objects whose language are not supported by the current Application
+    savedObjects = savedObjects.filter((obj) => {
```
kavilla, line 195: i think we should filter the results similar to how it does `(obj.type === 'index-pattern')` on line 180. that way we keep const savedObjects

src/plugins/saved_objects/public/finder/saved_object_finder.tsx
```diff
@@ -178,6 +191,25 @@ class SavedObjectFinderUi extends React.Component<
...
+      if (obj.type === 'search') {
+        if (obj.attributes?.kibanaSavedObjectMeta) {
+          const sourceObject = JSON.parse(obj.attributes?.kibanaSavedObjectMeta?.searchSourceJSON);
+          const languageId = sourceObject.query.language;
+
+          const languageProperties = this.languageService.getLanguage(languageId);
+
+          if (languageProperties?.supportedAppNames.includes(currentAppId)) {
```
kavilla, line 203: can be similified to

```
return this.languageService.getLanguage(languageId)?.supportedAppNames.includes(currentAppId)) || true
```

LDrago27: That makes sense. I have updated the change to reflect this.

src/plugins/saved_objects/public/finder/saved_object_finder.tsx
LDrago27: Consolidated the logic in one location.

src/plugins/dashboard/public/application/utils/create_dashboard_container.tsx
```diff
@@ -282,7 +282,12 @@ const getEmptyScreenProps = (
             getFactory: embeddable.getEmbeddableFactory,
             notifications,
             overlays,
-            SavedObjectFinder: getSavedObjectFinder(savedObjects, uiSettings),
+            SavedObjectFinder: getSavedObjectFinder(
```
kavilla, line 285: oh sorry about. not catching this earlier. im not sure if we want to update the interface like this as it might be a breaking change for any plugin using this.

i see the saved objects plugin gets started up with the services and with it is the data plugin. potentially the same with the services.application? that way we don't need to update the interface. since i can see it already touches a bunch of plugins within OSD. But no positive any plugins outside this repo (non-OpenSearch project and OpenSearch project)

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/saved_objects/public/plugin.ts#L57

src/plugins/saved_objects/public/finder/saved_object_finder.tsx
```diff
@@ -571,9 +607,20 @@ class SavedObjectFinderUi extends React.Component<
...
+) => {
   return (props: SavedObjectFinderProps) => (
-    <SavedObjectFinderUi {...props} savedObjects={savedObject} uiSettings={uiSettings} />
+    <SavedObjectFinderUi
+      {...props}
+      savedObjects={savedObject}
+      uiSettings={uiSettings}
+      data={data}
+      application={application}
+    />
```
sejli, line 623: @kavilla If we want to do a quick fix for this breaking change, we could make the new parameters optional.
```
data?: DataPublicPluginStart,
application?: ApplicationStart
```

LDrago27: Thanks @sejli I have taken your idea and made the new parameters as optional. Therefore if you don't pass in these params they should work in a similar way as before. 

To highlight the non breaking nature of changes I have reverted back the changes made to the state where we were using it without the new params to ensure it does n't break the functionality elsewhere.

src/plugins/dashboard/public/application/utils/create_dashboard_container.tsx
LDrago27: @kavilla  I tried this approach, however I see the downstream applications are directly importing the component without using the pluginStart props. Hence I decided against it.

src/plugins/saved_objects/public/finder/saved_object_finder.tsx
```diff
@@ -172,13 +201,24 @@ class SavedObjectFinderUi extends React.Component<
             getDataSource
           );
           return result;
+        } else if (obj.type === 'search') {
+          const sourceObject = JSON.parse(
+            obj.attributes?.kibanaSavedObjectMeta?.searchSourceJSON ?? null
+          );
+          const languageId = sourceObject?.query?.language;
+          if (this.isSavedSearchLanguageSupported(languageId, currentAppId, this.languageService)) {
```
kavilla, line 209: i wonder if it makes it a little bit more scalable/safer to just have this as a pass a function

like from the dashboards it could be

```typescript
 SavedObjectFinder: getSavedObjectFinder(
          savedObjects,
          uiSettings,
          searchFilterFn: (obj) => {
              const sourceObject = JSON.parse(
              obj.attributes?.kibanaSavedObjectMeta?.searchSourceJSON ?? null);
              const languageId = sourceObject?.query?.language;
              services.data.languageService?.getLanguage(languageId)?.supportedAppNames?.includes(currentAppId) ?? true;
          }
      )
```

kavilla: i think i only worry ensuring we dont end up adding more required plugins since i know a good amount of community members hit the APIs directly and even though this is the public side the saved objects plugin has been historically didn't need to care about the application which made it reliably abstracted and enforced a pattern for developers.

do we forsee in issues with introducing these concepts of language service to the saved object finder?

kavilla (approved):

src/plugins/saved_objects/public/finder/saved_object_finder.tsx
LDrago27: The overall change seems safe and the default behavior without the additional parameters remains unchanged for SavedObjectFinder. The saved_object_finder.test.tsx tests all use the default version of SavedObjectFinder without the optional parameters and the behavior remains unchanged.
 
In the meanwhile I have created the issue https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8122 , which we can use to create a more scalable version of the above when required.

sejli (approved):

PR #8019, opened 2024-09-05
[WIP][discover] PPL and SQL use default internal search route

kavilla:
### Description

commit currently doesn't work.

likely because we need to parse the bare minimum required request for PPL endpoints as it seems to be sending too much an error.

After all the refactoring for 2.17. The default /internal/search with search strategy is now usable. Effectly ensuring we are using no new API routes.

also i think this should fix the issue where index patterns dont work with ppl and sql (could be a new issue)

### Issues Resolved
n/a

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/common/utils.ts
+++ b/src/plugins/query_enhancements/common/utils.ts
@@ -3,15 +3,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { Query } from 'src/plugins/data/common';
-import { from, throwError, timer } from 'rxjs';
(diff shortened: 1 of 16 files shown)


src/plugins/query_enhancements/common/utils.ts
```diff
@@ -37,6 +33,12 @@ export const getFields = (rawResponse: any) => {
   }));
 };
 
+export const getQuery = (searchRequest: IOpenSearchDashboardsSearchRequest): Query => {
```
kavilla, line 36: parses query

kavilla:
will close but will track this

PR #8053, opened 2024-09-06
[workspace]feat: bypass permission control when construct ui settings client

SuZhou-Joe:
### Description

In 2.16 we refactored the way how we construct the find query based on workspace and permission control, which makes normal user unable to `find` global objects. And this behavior breaks the upgrade flow of advanced settings as it is using find to list all of the global configs

This PR is to fix this issue by constructing a saved objects client that will bypass workspace permission control wrapper so that the upgrade flow can work as it was.

## Testing the changes

- Turn on workspace permission control by turning on `savedObjects.permission.enabled: true` and config a user as the dashboard admin.
- Login with a non-dashboard-admin user
- Create a config whose doc id is `config:2.9.0` with some customized attributes and delete the existing config whose doc id is `config:3.0.0`
- Refresh the page and you should see a new doc with id `config:3.0.0` should be created, with the customized attributes upgraded from `config:2.9.0`

## Changelog

- fix: bypass permission control when construct ui settings client

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/server/plugin.ts
+++ b/src/plugins/workspace/server/plugin.ts
@@ -12,9 +12,9 @@ import {
   Logger,
   CoreStart,
   SharedGlobalConfig,
+  WORKSPACE_SAVED_OBJECTS_CLIENT_WRAPPER_ID,
 } from '../../../core/server';
 import {
-  WORKSPACE_SAVED_OBJECTS_CLIENT_WRAPPER_ID,
   WORKSPACE_CONFLICT_CONTROL_SAVED_OBJECTS_CLIENT_WRAPPER_ID,
   WORKSPACE_ID_CONSUMER_WRAPPER_ID,
   PRIORITY_FOR_WORKSPACE_CONFLICT_CONTROL_WRAPPER,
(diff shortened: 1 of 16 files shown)


src/plugins/workspace/server/plugin.ts
```diff
@@ -134,7 +134,9 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
           }
           const [coreStart] = await core.getStartServices();
           const uiSettingsClient = coreStart.uiSettings.asScopedToClient(
-            coreStart.savedObjects.getScopedClient(request)
+            coreStart.savedObjects.getScopedClient(request, {
+              excludedWrappers: [WORKSPACE_SAVED_OBJECTS_CLIENT_WRAPPER_ID],
+            })
           );
           const defaultWorkspaceId = await uiSettingsClient.get(DEFAULT_WORKSPACE);
```
Hailong-am, line 141: how about call `await uiSettingsClient.get(DEFAULT_WORKSPACE);` no matter what's the path after the user authenticated, after this call setting object should been initialized which include upgrade logic. With this, we don't need to put exclude workspace wrapper logic in other place.

SuZhou-Joe: Get your point, yeah let me see if I can find a centralized place that will trigger the `get` method of ui settings client.

PR #8060, opened 2024-09-06
Fix row rendering in infinite scroll

Swiddis:
### Description

When rendering infinite scrolling PPL/SQL in the new Discover UI, the rendering was stalling and entering an infinite loading loop. The root cause turns out to be that there's a small amount of fade-out at the bottom of the OSD application:

image

This was causing an issue with the progress bar observer responsible for loading the next batch of rows: it had `threshold: 1.0` set which meant that it would only trigger if the progress bar is 100% visible. This isn't a problem for DQL where large amounts of data records have a limit disclaimer at the bottom:

image

But when that disclaimer isn't present (for PPL and SQL), this fade effect would cover the progress bar, causing it not to meet the `1.0` threshold:

image

In addition to fixing the bug, this PR adds a batch of unit tests for the discover table.

### Issues Resolved

N/A

## Screenshot

After reaching the end of the available records:

image

## Testing the changes

The PR features a new set of unit tests for the involved functionality, but given the nature of the bug it's hard to test the fault directly (we would need to mock the entire OSD application and assert that the intersection threshold is > 0.5 when we scroll to the end). Theoretically possible, but probably too much effort for a one-line fix.

## Changelog

- fix: Fix row rendering in Discover infinite scroll 

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/components/default_discover_table/default_discover_table.tsx
+++ b/src/plugins/discover/public/application/components/default_discover_table/default_discover_table.tsx
@@ -264,7 +264,16 @@ const DefaultDiscoverTableUI = ({
         </table>
         {!showPagination && renderedRowCount < rows.length && (
           <div ref={sentinelRef}>
-            <EuiProgress size="xs" color="accent" data-test-subj="discoverRenderedRowsProgress" />
+            <EuiProgress
+              size="xs"
+              color="accent"
+              data-test-subj="discoverRenderedRowsProgress"
+              style={{
+                // Add a little margin if we aren't rendering the truncation callout below, to make
+                // the progress bar render better when it's not present
+                marginBottom: rows.length !== sampleSize ? '5px' : '0',
+              }}
+            />
           </div>
         )}
         {!showPagination && rows.length === sampleSize && (
--- a/src/plugins/discover/public/application/components/default_discover_table/default_discover_table.tsx
+++ b/src/plugins/discover/public/application/components/default_discover_table/default_discover_table.tsx
@@ -5,7 +5,7 @@
 
 import './_doc_table.scss';
 
-import React, { useEffect, useMemo, useRef, useState, useCallback, SetStateAction } from 'react';
+import React, { useEffect, useMemo, useRef, useState, useCallback } from 'react';
 import { EuiSmallButtonEmpty, EuiCallOut, EuiProgress } from '@elastic/eui';
 import { FormattedMessage } from '@osd/i18n/react';
 import { TableHeader } from './table_header';
--- a/src/plugins/discover/public/application/components/default_discover_table/default_discover_table.tsx
+++ b/src/plugins/discover/public/application/components/default_discover_table/default_discover_table.tsx
@@ -123,7 +123,11 @@ const DefaultDiscoverTableUI = ({
             setDesiredRowCount((prevRowCount) => prevRowCount + DESIRED_ROWS_LOOKAHEAD);
           }
         },
-        { threshold: 1.0 }
+        {
+          // Important that 0 < threshold < 1, since there OSD application div has a transparent
+          // fade at the bottom which causes the sentinel element to sometimes not be 100% visible
+          threshold: 0.5,
+        }
       );
 
       observerRef.current.observe(sentinelElement);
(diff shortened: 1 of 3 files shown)


src/plugins/discover/public/application/components/default_discover_table/default_discover_table.tsx
```diff
@@ -166,12 +172,17 @@ const DefaultDiscoverTableUI = ({
...
+        if (renderedRowCount + DESIRED_ROWS_LOOKAHEAD > desiredRowCount) {
+          setDesiredRowCount(Math.min(renderedRowCount + DESIRED_ROWS_LOOKAHEAD, rows.length));
+        }
       };
       lazyLoadRequestFrameRef.current = requestAnimationFrame(loadMoreRows);
     }
 
     return () => cancelAnimationFrame(lazyLoadRequestFrameRef.current);
-  }, [showPagination, renderedRowCount, desiredRowCount]);
+  }, [showPagination, renderedRowCount, desiredRowCount, rows.length]);
```
kavilla, line 185: do you mind checking if `rows` is also recreated? if so then it might be worth

```typescript
const [rowCount, setRowCount] = useState(rows.length);
```

src/plugins/discover/public/application/components/default_discover_table/default_discover_table.tsx
```diff
@@ -254,7 +265,7 @@ const DefaultDiscoverTableUI = ({
         </table>
         {!showPagination && renderedRowCount < rows.length && (
           <div ref={sentinelRef}>
-            <EuiProgress size="xs" color="accent" />
+            <EuiProgress size="xs" color="accent" data-test-subj="discoverRenderedRowsProgress" />
```
kavilla, line 268: nice!

src/plugins/discover/public/application/components/default_discover_table/default_discover_table.tsx
Swiddis: Yes, it's safely set via `setRows` [in the canvas data subscription](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/4776fcd9e0a875ce4b5ac97a73b8952c846afdf2/src/plugins/discover/public/application/view_components/canvas/index.tsx#L89). The changes should propagate safely. I added a unit test for this to be 100% sure, though.

src/plugins/discover/public/application/components/default_discover_table/default_discover_table.tsx
```diff
@@ -118,10 +120,14 @@ const DefaultDiscoverTableUI = ({
...
             setRenderedRowCount((prevRowCount) => prevRowCount + INFINITE_SCROLLED_PAGE_SIZE);
-            setDesiredRowCount((prevRowCount) => prevRowCount + 5 * INFINITE_SCROLLED_PAGE_SIZE);
+            setDesiredRowCount((prevRowCount) => prevRowCount + DESIRED_ROWS_LOOKAHEAD);
           }
         },
-        { threshold: 1.0 }
+        {
+          // Important that 0 < threshold < 1, since there OSD application div has a transparent
+          // fade at the bottom which causes the sentinel element to sometimes not be 100% visible
+          threshold: 0.5,
```
AMoo-Miki, line 129: why not use .001 so the observer will trigger loading as soon as the user scrolls there?

Swiddis: The progress bar is 2 pixels wide, with 50% opacity that makes any threshold lower than 25% redundant... I'll set it to 0.1

src/plugins/discover/public/application/components/default_discover_table/default_discover_table.tsx
```diff
@@ -254,7 +264,16 @@ const DefaultDiscoverTableUI = ({
...
+            <EuiProgress
+              size="xs"
+              color="accent"
+              data-test-subj="discoverRenderedRowsProgress"
+              style={{
+                // Add a little margin if we aren't rendering the truncation callout below, to make
+                // the progress bar render better when it's not present
+                marginBottom: rows.length !== sampleSize ? '5px' : '0',
+              }}
+            />
```
ashwin-pc, line 276: Why arent we adding the truncation banner for PPL and SQL?

Swiddis: That would require being able to get the active selected language in-context since PPL and SQL have a different default number of records they return when they have no limit configured, I briefly tried to follow the stack trace to see if I could pass that context down but couldn't find where it exists. Figured fixing the core bug was more important than the bells & whistles for this PR, could look at it for a future PR.

ashwin-pc (approved):
Thanks Simeon!

sejli (approved):
LGTM, I reran the build and test workflow that failed. I don't think it's related to your changes, but will wait for that before merging first.

joshuali925 (approved):

PR #8133, opened 2024-09-11
[Discover] Adding Recent Datasets

sejli:
### Description

Adds recent datasets to dataset selector. Also fixes an issue where we were able to select an index without specifying a time field option.

image

image

## Changelog

- feat: Adds recently selected data to data selector

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
@@ -4,6 +4,7 @@
  */
 
 import { CoreStart } from 'opensearch-dashboards/public';
+import LRUCache from 'lru-cache';
 import {
   Dataset,
   DataStructure,
@@ -22,6 +23,7 @@ export class DatasetService {
   private indexPatterns?: IndexPatternsContract;
   private defaultDataset?: Dataset;
   private typesRegistry: Map<string, DatasetTypeConfig> = new Map();
+  private recentDatasets: LRUCache<string, Dataset> = new LRUCache({ max: 4 });
 
   constructor(
     private readonly uiSettings: CoreStart['uiSettings'],
@@ -59,6 +61,22 @@ export class DatasetService {
 
   public getDefault(): Dataset | undefined {
     return this.defaultDataset;
+  }
+
+  public getRecentDataset(dataset: string | undefined): Dataset | undefined {
+    if (dataset) {
+      return this.recentDatasets.get(dataset);
+    }
+  }
+
+  public getRecentDatasets(): Dataset[] {
+    return Array.from(this.recentDatasets.values());
+  }
+
+  public addRecentDataset(dataset: Dataset | undefined): void {
+    if (dataset) {
+      this.recentDatasets.set(dataset.id, dataset);
+    }
   }
 
   public async cacheDataset(dataset: Dataset): Promise<void> {
--- a/src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
@@ -78,12 +78,6 @@ export class DatasetService {
     }
   }
 
-  public getRecentDataset(datasetID: string | undefined): Dataset | undefined {
-    if (datasetID) {
-      return this.recentDatasets.get(datasetID);
-    }
-  }
-
   public getRecentDatasets(): Dataset[] {
     return this.recentDatasets.values();
   }
--- a/src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
@@ -23,7 +23,7 @@ export class DatasetService {
   private indexPatterns?: IndexPatternsContract;
   private defaultDataset?: Dataset;
   private typesRegistry: Map<string, DatasetTypeConfig> = new Map();
-  private recentDatasets: LRUCache<string, Dataset> = new LRUCache({ max: 4 });
+  private recentDatasets: LRUCache<string, Dataset>;
 
   constructor(
     private readonly uiSettings: CoreStart['uiSettings'],
@@ -32,6 +32,9 @@ export class DatasetService {
     if (this.uiSettings.get(UI_SETTINGS.QUERY_ENHANCEMENTS_ENABLED)) {
       this.registerDefaultTypes();
     }
+    this.recentDatasets = new LRUCache({
+      max: this.uiSettings.get(UI_SETTINGS.SEARCH_MAX_RECENT_DATASETS) ?? 4,
+    });
   }
 
   /**
@@ -63,9 +66,9 @@ export class DatasetService {
     return this.defaultDataset;
   }
 
-  public getRecentDataset(dataset: string | undefined): Dataset | undefined {
-    if (dataset) {
-      return this.recentDatasets.get(dataset);
+  public getRecentDataset(datasetID: string | undefined): Dataset | undefined {
+    if (datasetID) {
+      return this.recentDatasets.get(datasetID);
     }
   }
 
--- a/src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
@@ -35,6 +35,7 @@ export class DatasetService {
     this.recentDatasets = new LRUCache({
       max: this.uiSettings.get(UI_SETTINGS.SEARCH_MAX_RECENT_DATASETS),
     });
+    this.deserializeRecentDatasets();
   }
 
   /**
@@ -66,6 +67,17 @@ export class DatasetService {
     return this.defaultDataset;
   }
 
+  private serializeRecentDatasets(): void {
+    this.sessionStorage.set('recentDatasets', this.getRecentDatasets());
+  }
+
+  private deserializeRecentDatasets(): void {
+    const cacheData = this.sessionStorage.get('recentDatasets');
+    if (cacheData) {
+      cacheData.forEach((dataset: Dataset) => this.addRecentDataset(dataset, false));
+    }
+  }
+
   public getRecentDataset(datasetID: string | undefined): Dataset | undefined {
     if (datasetID) {
       return this.recentDatasets.get(datasetID);
@@ -73,12 +85,15 @@ export class DatasetService {
   }
 
   public getRecentDatasets(): Dataset[] {
-    return Array.from(this.recentDatasets.values());
+    return this.recentDatasets.values();
   }
 
-  public addRecentDataset(dataset: Dataset | undefined): void {
+  public addRecentDataset(dataset: Dataset | undefined, serialize: boolean = true): void {
     if (dataset) {
       this.recentDatasets.set(dataset.id, dataset);
+    }
+    if (serialize) {
+      this.serializeRecentDatasets();
     }
   }
 
--- a/src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
@@ -33,7 +33,7 @@ export class DatasetService {
       this.registerDefaultTypes();
     }
     this.recentDatasets = new LRUCache({
-      max: this.uiSettings.get(UI_SETTINGS.SEARCH_MAX_RECENT_DATASETS) ?? 4,
+      max: this.uiSettings.get(UI_SETTINGS.SEARCH_MAX_RECENT_DATASETS),
     });
   }
 
--- a/src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
+++ b/src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
@@ -17,9 +17,10 @@ import { FormattedMessage } from '@osd/i18n/react';
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { toMountPoint } from '../../../../opensearch_dashboards_react/public';
 import { Dataset, DEFAULT_DATA } from '../../../common';
-import { getQueryService } from '../../services';
 import { IDataPluginServices } from '../../types';
 import { AdvancedSelector } from './advanced_selector';
+import { getQueryService } from '../../services';
+import { DatasetServiceContract } from '../../query';
 
 interface DatasetSelectorProps {
   selectedDataset?: Dataset;
@@ -44,7 +45,7 @@ export const DatasetSelector = ({
 }: DatasetSelectorProps) => {
   const isMounted = useRef(false);
   const [isOpen, setIsOpen] = useState(false);
-  const [datasets, setDatasets] = useState<Dataset[]>([]);
+  const [datasets, setDatasets] = useState<Array<{ dataset: Dataset; recent: boolean }>>([]);
   const { overlays } = services;
   const datasetService = getQueryService().queryString.getDatasetService();
   const datasetIcon =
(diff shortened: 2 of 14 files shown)


src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -4,6 +4,7 @@
  */
 
 import { CoreStart } from 'opensearch-dashboards/public';
+import LRUCache from 'lru-cache';
```
ashwin-pc, line 7: Damm didnt know we had his package already available!

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -62,6 +64,22 @@ export class DatasetService {
     return this.defaultDataset;
   }
 
+  public getRecentDataset(dataset: string | undefined): Dataset | undefined {
```
ashwin-pc, line 67:
```suggestion
  public getRecentDataset(id: string | undefined): Dataset | undefined {
```
or `datasetId` passing in a  dataset and getting a dataset doesnt make sense right?

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -62,6 +64,22 @@ export class DatasetService {
...
+    if (dataset) {
+      return this.recentDatasets.get(dataset);
+    }
+  }
+
+  public getRecentDatasets(): Dataset[] {
```
ashwin-pc, line 73: it would be nice if we can specify a number a limit to that, default can be no limit. This will help us show this in other places such as the advanced selector without modifications

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
ashwin-pc, line 1: Can you add unit tests for these?

src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
```diff
@@ -99,22 +99,41 @@ export const DatasetSelector = ({
...
+      const label = dataSource ? `${dataSource.title}::${title}` : title;
+      recentlySelected.push({
+        label,
+        checked: id === selectedDataset?.id ? 'on' : undefined,
+        key: id,
+        prepend: <EuiIcon type={datasetService.getType(type)!.meta.icon.type} />,
+      });
+    });
+    if (recentlySelected.length > 1) {
+      newOptions.unshift(...recentlySelected);
```
ashwin-pc, line 118: Nit: Cant these simply be appended before index patterns?

src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
```diff
@@ -99,22 +99,41 @@ export const DatasetSelector = ({
...
   }, [datasets, selectedDataset?.id, datasetService]);
 
   const handleOptionChange = useCallback(
     (newOptions: EuiSelectableOption[]) => {
       const selectedOption = newOptions.find((option) => option.checked === 'on');
       if (selectedOption) {
-        const foundDataset = datasets.find((dataset) => dataset.id === selectedOption.key);
+        const foundDataset =
+          datasets.find((dataset) => dataset.id === selectedOption.key) ??
+          datasetService.getRecentDataset(selectedOption.key);
```
ashwin-pc, line 129: Cant the recents be added along with the fetched datasets so that we have a single source of truth and avoids adding another dependency to the useCallback hook?

ashwin-pc (commented):
Nice! Can we add some tests for this feature so that we can be sure that these dont break as we fix more things?

ashwin-pc:
How are you limiting the recents?

src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
```diff
@@ -99,22 +99,41 @@ export const DatasetSelector = ({
         prepend: <EuiIcon type={datasetService.getType(type)!.meta.icon.type} />,
       });
     });
-
+    const recentlySelected: EuiSelectableOption[] = [
+      {
+        label: 'Recently Selected Data',
```
kavilla, line 104: Shouldn't this be `Recently selected data`

Screenshot 2024-09-12 at 2 20 51 PM

src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
kavilla: +1. 

also do we know if we want to display this drop down with two checkmarks? or if the recents contain a dataset id then we won't display the dataset within the original type?

For example, if they have an index pattern like `sample_data_logs` and they select it. then `sample_data_logs` will display only in the recents section with a section and the index patterns section will not have `sample_data_logs`.

or did we want two checkmarks with `sample_data_logs` existing in both places?

src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
kavilla: +1

src/plugins/data/public/ui/dataset_selector/index.tsx
```diff
@@ -18,6 +18,7 @@ const ConnectedDatasetSelector = ({ onSubmit }: ConnectedDatasetSelectorProps) =
   const { services } = useOpenSearchDashboards<IDataPluginServices>();
   const queryString = services.data.query.queryString;
   const initialDataset = queryString.getQuery().dataset || queryString.getDefaultQuery().dataset;
+  services.data.query.queryString.getDatasetService().addRecentDataset(initialDataset);
```
kavilla, line 21:
```typescript
queryString.getDatasetService().addRecentDataset(initialDataset);
```

src/plugins/data/public/ui/dataset_selector/index.tsx
```diff
@@ -34,6 +35,7 @@ const ConnectedDatasetSelector = ({ onSubmit }: ConnectedDatasetSelectorProps) =
       const query = queryString.getInitialQueryByDataset(dataset);
       queryString.setQuery(query);
       onSubmit!(queryString.getQuery());
+      services.data.query.queryString.getDatasetService().addRecentDataset(dataset);
```
kavilla, line 38:
```typescript
queryString.getDatasetService().addRecentDataset(initialDataset);
```

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
kavilla: +1, UI settings is already passed in. we could just add a new advanced setting

src/plugins/data/public/ui/dataset_selector/index.tsx
kavilla: @ashwin-pc do you think we should be just handling this is the query string manager? like updating the recents based on the query?

src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
sejli: The `datasets` state in this file is fetched using `fetchDatasets()`, which just calls the `fetch()` function for index patterns. Since it is just an array of index patterns, we would need to identify which datasets in the array belong to the index pattern section and the recently selected datasets section. I separated them here because it's possible that the recent datasets are not index patterns. To use a single source of truth, I think the easiest way would be to build the options array in `fetchDatasets()` such that the `datasets` state already contains the labels differentiating the two sections in the array. Open to other suggestions though.

src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
ashwin-pc: Thats a good callout. For simplicity I think we should deduplicate it and show it only in the recents. That way showing the check mark will also be easier

src/plugins/data/public/ui/dataset_selector/index.tsx
ashwin-pc: Yes that would make it less prone to issues like when datasets are selected using ways that arent the dataset selector. Wouldnt block on that being a requirement, but that is a cleaner approach

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -31,6 +33,9 @@ export class DatasetService {
     if (this.uiSettings.get(UI_SETTINGS.QUERY_ENHANCEMENTS_ENABLED)) {
       this.registerDefaultTypes();
     }
+    this.recentDatasets = new LRUCache({
+      max: this.uiSettings.get(UI_SETTINGS.SEARCH_MAX_RECENT_DATASETS) ?? 4,
```
kavilla, line 37: nit: could this actually be undefined?

sejli: While running tests, `uiSettings` has everything set to be `jest.fn()`, which is the only place where it would be undefined. Could override the original `uiSettings` mock such that `uiSettings.get()` returns `4`, would that be better practice?

ashwin-pc: yes, dont change the UI implementation to meet a test infrastructure gap if possible.

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -62,6 +67,22 @@ export class DatasetService {
...
+    if (datasetID) {
+      return this.recentDatasets.get(datasetID);
+    }
+  }
+
+  public getRecentDatasets(): Dataset[] {
+    return Array.from(this.recentDatasets.values());
+  }
+
+  public addRecentDataset(dataset: Dataset | undefined): void {
```
kavilla, line 80: do you think instead of relying on the UI component for setting the dataset the recent dataset is the one that is used? so like nobody could add recent dataset they can only access the recent data sets?

kavilla (approved):

src/plugins/data/public/ui/dataset_selector/index.tsx
kavilla: @sejli maybe could be an open issue

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -165,6 +165,7 @@ export const Configurator = ({
             onConfirm(dataset);
           }}
           fill
+          disabled={timeFields && timeFields.length > 0 && timeFieldName === undefined}
```
paulstn, line 168: nit: but couldn't you combine the first two expressions down to `timeFields?.length > 0`. 
do you think that the way you did it would be better style?

src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
```diff
@@ -57,15 +58,26 @@ export const DatasetSelector = ({
...
+    ];
+    setDatasets(allDatasets);
 
     // If no dataset is selected, select the first one
     if (!selectedDataset && fetchedDatasets.length > 0) {
-      setSelectedDataset(fetchedDatasets[0]);
+      setSelectedDataset(
+        allDatasets.filter((ds) => ds.recent)[0].dataset ??
+          allDatasets.filter((ds) => !ds.recent)[0].dataset
+      );
```
joshuali925, line 80: is the second filter needed? since if there's no recent datasets, then every dataset should satisfy `!ds.recent`

also recent datasets are added at first anyways, this is just `allDatasets[0].dataset`?

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -23,6 +24,7 @@ export class DatasetService {
   private indexPatterns?: IndexPatternsContract;
   private defaultDataset?: Dataset;
   private typesRegistry: Map<string, DatasetTypeConfig> = new Map();
+  private recentDatasets: LRUCache<string, Dataset>;
```
joshuali925, line 27: is this just in memory or stored in some storage?

src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
```diff
@@ -57,15 +58,26 @@ export const DatasetSelector = ({
...
     const fetchedDatasets =
       fetchedIndexPatternDataStructures.children?.map((pattern) =>
         typeConfig.toDataset([pattern])
       ) ?? [];
-    setDatasets(fetchedDatasets);
+    const filteredFetchedDatasets = fetchedDatasets.filter(
+      (dataset) => !recentDatasets.some((ds) => ds.id === dataset.id)
+    );
+    const allDatasets = [
+      ...recentDatasets.map((ds) => ({ dataset: ds, recent: true })),
```
joshuali925, line 70: would there be invalid/outdated datasets? should we check and make sure every recent dataset appears in the fetchedDatasets?

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
sejli: Currently, we'll just store it in memory.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
ashwin-pc: This will conflict with your time field fixes PR. use the logic from there.

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -31,6 +33,9 @@ export class DatasetService {
     if (this.uiSettings.get(UI_SETTINGS.QUERY_ENHANCEMENTS_ENABLED)) {
       this.registerDefaultTypes();
     }
+    this.recentDatasets = new LRUCache({
```
ashwin-pc, line 36: We should store this in session storage so that when the user returns we can show them this. Not a blocker for this PR, but if we are skipping this, lets add a fast follow. Recent datasets are not very useful if they dont persist across my session

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
ashwin-pc: +1 Again not a blocker for this PR, but we should add a fast follow to do this automatically when a dataset is set in the query manager. additionally if we dont want certain setQuery.dataset calls to update the LRU cache, we can always pass a option to not cache it. This can be used for on page load updates that load the default dataset on page load if no dataset is present in the URL. 

This will be a source of bugs in the future when someone in another app or usecase forgets to call this.

src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
```diff
@@ -57,15 +58,26 @@ export const DatasetSelector = ({
...
 
+    const recentDatasets = datasetService.getRecentDatasets();
     const fetchedDatasets =
       fetchedIndexPatternDataStructures.children?.map((pattern) =>
         typeConfig.toDataset([pattern])
       ) ?? [];
-    setDatasets(fetchedDatasets);
+    const filteredFetchedDatasets = fetchedDatasets.filter(
+      (dataset) => !recentDatasets.some((ds) => ds.id === dataset.id)
+    );
```
ashwin-pc, line 68:
```suggestion
    const recentDatasetIds = new Set(recentDatasets.map(ds => ds.id));

    const filteredFetchedDatasets = fetchedDatasets.filter(
      dataset => !recentDatasetIds.has(dataset.id)
   );
```
nit: slightly easier to read

src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
ashwin-pc: No, because fetched datasets only contain index patterns

ashwin-pc (commented):
Lemme know what you think about my comments on:
1. Caching the recent datasets in the session storage
2. Adding to recent datasets when the query in the query string manager is updated

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
kavilla: +1

yeah I agree i think sean and i had similar idea. we should def backlog an item posting.

copy pasta for transparency:

-------
in the interest of timelines, potential approach use hook for setRecentDatasets and then a new const recentOptions to useMemo and then join the the options in the html component

* **create backlog issue**: storing the datasets created in the dataset service
  * new function in dataset service like `getDatasets()`
  * populate the selector for all the datasets. when users create a dataset it stores the dataset service and updates the LRU recent datasets. when user select a dataset it will update the LRU recent datasets. 
  * can use the session storage to store datasets since it data structure could be of type `DATASET`. 
  * when the user opens the dataset selector for the first time it calls getRecentDatasets and and getDatasets from the data service. then subsequent calls will just be the get recent datasets.
* **benefits**: dataset service is a single source of truth of recent datasets and datasets. another plugin can access the datasets without using our component. we can easily update the session id in the dataset service without the query state being updated.

joshuali925 (dismissed):

joshuali925 (approved):

ananzh (approved):

PR #8136, opened 2024-09-11
Fix the display and jump logic for recent assets

Kapian1234:
### Description
1. Filter out global objects in recently viewed assets when workspace is enabled.
2. In recently accessed, when workspace disabled, it should jump to the global visualization/dashboard page (ignore the workspace of the current saved object).

## Changelog

- fix: Fix the display and jump logic for recent assets

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/core/public/chrome/recently_accessed/recently_accessed_service.ts
```diff
@@ -81,10 +84,16 @@ export class RecentlyAccessedService {
...
       /** Gets the current array of history items. */
-      get: () => history.get(),
+      get: () => history.get().filter((item) => !!item.workspaceId),
 
       /** Gets an observable of the current array of history items. */
-      get$: () => history.get$(),
+      get$: () => {
+        return history.get$().pipe(
+          map((items) => {
+            return items.filter((item) => (workspaceEnabled ? !!item.workspaceId : true));
```
wanglam, line 93: Can we add some UT to test this filtering logic?

src/core/public/chrome/ui/header/nav_link.tsx
```diff
@@ -125,13 +125,17 @@ export function createRecentNavLink(
...
+  navigateToUrl: InternalApplicationStart['navigateToUrl'],
+  workspaceEnabled: boolean = true
 ): RecentNavLink {
   const { link, label, workspaceId } = recentLink;
   const href = relativeToAbsolute(
-    basePath.prepend(formatUrlWithWorkspaceId(link, workspaceId || '', basePath), {
-      withoutClientBasePath: true,
-    })
+    basePath.prepend(
+      formatUrlWithWorkspaceId(link, workspaceEnabled ? workspaceId || '' : '', basePath),
```
wanglam, line 134: This logic looks a little bit strange to me, I think we don't need to call `formatUrlWithWorkspaceId` if workspaceEnabled equal false. How about refactor with below code?
```suggestion
      workspaceEnabled ? formatUrlWithWorkspaceId(link, workspaceId || '', basePath) : link,
```
By the way, can we add some unit tests about it?

src/core/public/chrome/recently_accessed/recently_accessed_service.ts
```diff
@@ -81,10 +84,16 @@ export class RecentlyAccessedService {
       },
 
       /** Gets the current array of history items. */
-      get: () => history.get(),
+      get: () => history.get().filter((item) => !!item.workspaceId),
```
Hailong-am, line 87: will this impact workspace disabled case? I think we should check workspaceEnabled flag

Kapian1234: yes I forgot it here, thx for reminding!

src/core/public/chrome/ui/header/nav_link.tsx
Kapian1234: sure

ruanyl (approved):

src/core/public/chrome/ui/header/nav_link.tsx
```diff
@@ -125,13 +125,17 @@ export function createRecentNavLink(
   recentLink: ChromeRecentlyAccessedHistoryItem,
   navLinks: ChromeNavLink[],
   basePath: HttpStart['basePath'],
-  navigateToUrl: InternalApplicationStart['navigateToUrl']
+  navigateToUrl: InternalApplicationStart['navigateToUrl'],
+  workspaceEnabled: boolean = true
```
SuZhou-Joe, line 129: The default value should be false I think?

SuZhou-Joe: And I do not think the flag is needed here, if the recent item has workspaceId but the feature flag workspace is turned off, it should be be displayed to users.

src/plugins/saved_objects_management/public/management_section/recent_work.tsx
```diff
@@ -232,9 +232,9 @@ export const RecentWork = (props: { core: CoreStart; workspaceEnabled?: boolean
                 recentAccessItem,
                 navLinks,
                 core.http.basePath,
-                core.application.navigateToUrl
+                core.application.navigateToUrl,
+                !!workspaceEnabled
```
SuZhou-Joe, line 236: Shall we do a search for the places that are using `createRecentNavLink`? IIRC there are at least 2 places:
- recent menu in breadcrumb.
- recent works in home page.

src/core/public/chrome/ui/header/nav_link.tsx
Kapian1234: Indeed, it should be false.

src/core/public/chrome/recently_accessed/recently_accessed_service.ts
```diff
@@ -81,10 +84,16 @@ export class RecentlyAccessedService {
       },
 
       /** Gets the current array of history items. */
-      get: () => history.get(),
+      get: () => history.get().filter((item) => (workspaceEnabled ? !!item.workspaceId : true)),
```
ruanyl, line 87: I'm thinking, should the filtering logic be done in the component where the data is used?

wanglam: Correct me if I missing understanding. I remember we discuss this offline. All recent accessed items should applied the same filtering logic.

SuZhou-Joe: Right, if we do the filtering logic in component, then we need to applied the same logic in multiple places.

But it occurs to me that we can use different storage key for workspace enabled / workspace disabled, then we do not need to use filter.

ruanyl: > Correct me if I missing understanding. I remember we discuss this offline. All recent accessed items should applied the same filtering logic.

I see, I'm fine with it now. 

> But it occurs to me that we can use different storage key for workspace enabled / workspace disabled, then we do not need to use filter.

Sounds good to me, could a future improvement/refactor task

ananzh:
@Kapian1234 thanks for the contribution. I update it to 2.19. could you check and fix the comments? If no response, we will need to convert it to draft.

Kapian1234:
> @Kapian1234 thanks for the contribution. Could you check and fix the comment? I think we could release this for 2.19. Will leave @SuZhou-Joe and @ruanyl to make the call.

ok

SuZhou-Joe (dismissed):

Kapian1234:
> @Kapian1234 , bootstrap failed, could you please take a look?

ok

SuZhou-Joe (approved):

src/core/public/chrome/ui/header/__snapshots__/header.test.tsx.snap
```diff
@@ -8005,7 +8005,7 @@ exports[`Header handles visibility and lock changes 1`] = `
                                   Object {
                                     "aria-label": "dashboard, type: opensearchDashboards",
                                     "data-test-subj": "collapsibleNavAppLink--recent",
-                                    "href": "http://localhost/test/",
+                                    "href": "http://localhost/",
```
wanglam, line 8008: Seems this snapshot missing base path here. I think this should not be updated.

src/core/public/chrome/ui/header/nav_link.test.tsx
```diff
@@ -103,6 +103,6 @@ describe('createRecentNavLink', () => {
       mockedNavigateToUrl
     );
 
-    expect(recentLink.href).toEqual('http://localhost/test/w/foo/app/foo');
+    expect(recentLink.href).toEqual('http://localhost/test/app/foo');
```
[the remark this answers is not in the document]
Kapian1234: Added

SuZhou-Joe (approved):

Hailong-am (approved):

PR #8166, opened 2024-09-12
[Discover] UI Updates to Dataset Configurator

sejli:
### Description

* Fixes blank row at bottom of Discover, thanks @ashwin-pc @AMoo-Miki for the change
* Fixes an issue where indexes were able to be selected with a placeholder time field, which would cause issues
* Disables time field selector in index pattern configurator if the index pattern already has a time field.

## Screenshot

image

image

image

## Changelog

- fix: Fixes UI issues in Discover and data configurator

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/mocks.ts
+++ b/src/plugins/data/public/mocks.ts
@@ -91,6 +91,12 @@ const createStartContract = (isEnhancementsEnabled: boolean = false): Start => {
         })
       ),
       clearCache: jest.fn(),
+      create: jest.fn().mockResolvedValue({
+        id: 'test-index-pattern',
+        title: 'Test Index Pattern',
+        type: 'INDEX_PATTERN',
+      }),
+      saveToCache: jest.fn(),
     } as unknown) as IndexPatternsContract,
     dataSources: dataSourceServiceMock.createStartContract(),
   };
--- a/src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
@@ -9,7 +9,6 @@ import {
   DataStructure,
   IndexPatternSpec,
   DEFAULT_DATA,
-  IFieldType,
   UI_SETTINGS,
   DataStorage,
   CachedDataStructure,
@@ -64,14 +63,11 @@ export class DatasetService {
 
   public async cacheDataset(dataset: Dataset): Promise<void> {
     const type = this.getType(dataset.type);
-    if (dataset) {
+    if (dataset && dataset.type !== DEFAULT_DATA.SET_TYPES.INDEX_PATTERN) {
       const spec = {
         id: dataset.id,
         title: dataset.title,
-        timeFieldName: {
-          name: dataset.timeFieldName,
-          type: 'date',
-        } as Partial<IFieldType>,
+        timeFieldName: dataset.timeFieldName,
         fields: await type?.fetchFields(dataset),
         dataSourceRef: dataset.dataSource
           ? {
--- a/src/plugins/data/public/query/query_string/language_service/types.ts
+++ b/src/plugins/data/public/query/query_string/language_service/types.ts
@@ -52,5 +52,5 @@ export interface LanguageConfig {
   showDocLinks?: boolean;
   editorSupportedAppNames?: string[];
   supportedAppNames?: string[];
-  noTimeField?: boolean;
+  disableDatePicker?: boolean;
 }
--- a/src/plugins/data/public/query/query_string/language_service/types.ts
+++ b/src/plugins/data/public/query/query_string/language_service/types.ts
@@ -52,4 +52,5 @@ export interface LanguageConfig {
   showDocLinks?: boolean;
   editorSupportedAppNames?: string[];
   supportedAppNames?: string[];
+  noTimeField?: boolean;
 }
--- a/src/plugins/data/public/query/query_string/language_service/types.ts
+++ b/src/plugins/data/public/query/query_string/language_service/types.ts
@@ -52,5 +52,5 @@ export interface LanguageConfig {
   showDocLinks?: boolean;
   editorSupportedAppNames?: string[];
   supportedAppNames?: string[];
-  disableDatePicker?: boolean;
+  hideDatePicker?: boolean;
 }
--- a/src/plugins/data/public/ui/dataset_selector/configurator.tsx
+++ b/src/plugins/data/public/ui/dataset_selector/configurator.tsx
@@ -19,7 +19,7 @@ import {
 import { i18n } from '@osd/i18n';
 import { FormattedMessage } from '@osd/i18n/react';
 import React, { useEffect, useState } from 'react';
-import { BaseDataset, Dataset, DatasetField } from '../../../common';
+import { BaseDataset, DEFAULT_DATA, Dataset, DatasetField } from '../../../common';
 import { getIndexPatterns, getQueryService } from '../../services';
 
 export const Configurator = ({
@@ -112,16 +112,17 @@ export const Configurator = ({
                     text: field.displayName || field.name,
                     value: field.name,
                   })),
-                  { text: '-----', value: '', disabled: true },
-                  { text: 'No time field', value: undefined },
+                  { text: '-----', value: '-----', disabled: true },
+                  { text: 'I dont want to use time filter', value: '' },
                 ]}
                 value={timeFieldName}
                 onChange={(e) => {
-                  const value = e.target.value === 'undefined' ? undefined : e.target.value;
-                  setTimeFieldName(value);
+                  setTimeFieldName(e.target.value);
+                  const value = e.target.value === '' ? undefined : e.target.value;
                   setDataset({ ...dataset, timeFieldName: value });
                 }}
-                disabled={dataset?.timeFieldName !== undefined}
+                hasNoInitialSelection={dataset.type === DEFAULT_DATA.SET_TYPES.INDEX}
+                disabled={dataset.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN}
               />
             </EuiFormRow>
           )}
@@ -166,7 +167,7 @@ export const Configurator = ({
             onConfirm(dataset);
           }}
           fill
-          disabled={timeFields && timeFields.length > 0 && timeFieldName === undefined}
+          disabled={timeFieldName === undefined && dataset.type === DEFAULT_DATA.SET_TYPES.INDEX}
         >
           <FormattedMessage
             id="data.explorer.datasetSelector.advancedSelector.confirm"
--- a/src/plugins/data/public/ui/dataset_selector/configurator.tsx
+++ b/src/plugins/data/public/ui/dataset_selector/configurator.tsx
@@ -52,7 +52,7 @@ export const Configurator = ({
     return languages[0];
   });
 
-  const displayTimeFieldSelector = (ds: Dataset, lang: string, fields: DatasetField[]) => {
+  const shouldDisplayTimeFieldSelector = (ds: Dataset, lang: string, fields: DatasetField[]) => {
     return (
       !(
         languageService.getLanguage(lang)?.disableDatePicker ||
@@ -113,7 +113,7 @@ export const Configurator = ({
             !languageService.getLanguage(language)?.disableDatePicker && (
               <EuiFormRow
                 label={i18n.translate(
-                  'data.explorer.datasetSelector.advancedSelector.configurator.datasetLabel',
+                  'data.explorer.datasetSelector.advancedSelector.configurator.indexPatternTimeFieldLabel',
                   {
                     defaultMessage: 'Time field',
                   }
@@ -122,7 +122,7 @@ export const Configurator = ({
                 <EuiFieldText disabled value={dataset.timeFieldName ?? 'No time field'} />
               </EuiFormRow>
             )}
-          {displayTimeFieldSelector(dataset, language, timeFields) && (
+          {shouldDisplayTimeFieldSelector(dataset, language, timeFields) && (
             <EuiFormRow
               label={i18n.translate(
                 'data.explorer.datasetSelector.advancedSelector.configurator.timeFieldLabel',
@@ -192,7 +192,8 @@ export const Configurator = ({
           }}
           fill
           disabled={
-            timeFieldName === undefined && displayTimeFieldSelector(dataset, language, timeFields)
+            timeFieldName === undefined &&
+            shouldDisplayTimeFieldSelector(dataset, language, timeFields)
           }
         >
           <FormattedMessage
--- a/src/plugins/data/public/ui/dataset_selector/configurator.tsx
+++ b/src/plugins/data/public/ui/dataset_selector/configurator.tsx
@@ -52,9 +52,9 @@ export const Configurator = ({
     return languages[0];
   });
 
-  const requiresTimeField = (ds: Dataset, lang: string, fields: DatasetField[]) => {
+  const displayTimeFieldSelector = (ds: Dataset, lang: string, fields: DatasetField[]) => {
     if (
-      languageService.getLanguage(lang)?.noTimeField ||
+      languageService.getLanguage(lang)?.disableDatePicker ||
       ds.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN
     ) {
       return false;
@@ -62,6 +62,7 @@ export const Configurator = ({
     if (fields && fields.length > 0) {
       return true;
     }
+    return false;
   };
 
   useEffect(() => {
@@ -111,7 +112,7 @@ export const Configurator = ({
             <EuiFieldText disabled value={dataset.title} />
           </EuiFormRow>
           {dataset.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN &&
-            !languageService.getLanguage(language)?.noTimeField && (
+            !languageService.getLanguage(language)?.disableDatePicker && (
               <EuiFormRow
                 label={i18n.translate(
                   'data.explorer.datasetSelector.advancedSelector.configurator.datasetLabel',
@@ -123,7 +124,7 @@ export const Configurator = ({
                 <EuiFieldText disabled value={dataset.timeFieldName ?? 'No time field'} />
               </EuiFormRow>
             )}
-          {requiresTimeField(dataset, language, timeFields) && (
+          {displayTimeFieldSelector(dataset, language, timeFields) && (
             <EuiFormRow
               label={i18n.translate(
                 'data.explorer.datasetSelector.advancedSelector.configurator.timeFieldLabel',
@@ -147,7 +148,7 @@ export const Configurator = ({
                   setTimeFieldName(e.target.value);
                   setDataset({ ...dataset, timeFieldName: value });
                 }}
-                hasNoInitialSelection={dataset.type === DEFAULT_DATA.SET_TYPES.INDEX}
+                hasNoInitialSelection={dataset.type !== DEFAULT_DATA.SET_TYPES.INDEX_PATTERN}
               />
             </EuiFormRow>
           )}
@@ -192,7 +193,9 @@ export const Configurator = ({
             onConfirm(dataset);
           }}
           fill
-          disabled={timeFieldName === undefined && requiresTimeField(dataset, language, timeFields)}
+          disabled={
+            timeFieldName === undefined && displayTimeFieldSelector(dataset, language, timeFields)
+          }
         >
           <FormattedMessage
             id="data.explorer.datasetSelector.advancedSelector.confirm"
--- a/src/plugins/data/public/ui/dataset_selector/configurator.tsx
+++ b/src/plugins/data/public/ui/dataset_selector/configurator.tsx
@@ -41,8 +41,9 @@ export const Configurator = ({
   const languages = type?.supportedLanguages(baseDataset) || [];
 
   const [dataset, setDataset] = useState<Dataset>(baseDataset);
-  const [timeFields, setTimeFields] = useState<DatasetField[]>();
+  const [timeFields, setTimeFields] = useState<DatasetField[]>([]);
   const [timeFieldName, setTimeFieldName] = useState<string | undefined>(dataset.timeFieldName);
+  const noTimeFilter = 'I dont want to use the time filter';
   const [language, setLanguage] = useState<string>(() => {
     const currentLanguage = queryString.getQuery().language;
     if (languages.includes(currentLanguage)) {
@@ -50,6 +51,18 @@ export const Configurator = ({
     }
     return languages[0];
   });
+
+  const requiresTimeField = (ds: Dataset, lang: string, fields: DatasetField[]) => {
+    if (
+      languageService.getLanguage(lang)?.noTimeField ||
+      ds.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN
+    ) {
+      return false;
+    }
+    if (fields && fields.length > 0) {
+      return true;
+    }
+  };
 
   useEffect(() => {
     const fetchFields = async () => {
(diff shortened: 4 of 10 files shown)


ashwin-pc:
> Disables time field selector in index pattern configurator if the index pattern already has a time field.

The time field selector should be disabled no matter what for index patterns. So even if they have an index pattern where they have selected no time field, this should not let the user edit it.

ashwin-pc:
These changes also need changelogs since these are fixes for whats in 2.17

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -165,6 +166,7 @@ export const Configurator = ({
             onConfirm(dataset);
           }}
           fill
+          disabled={timeFields && timeFields.length > 0 && timeFieldName === undefined}
```
ashwin-pc, line 169: What will happen if my index does not have any time fields? or if the user has selected the option to not have a time field?

ashwin-pc: Also lets add a unit test for this logic so that we show the fields when we expect the disabled behaviour to show up

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -121,6 +121,7 @@ export const Configurator = ({
                   setTimeFieldName(value);
                   setDataset({ ...dataset, timeFieldName: value });
                 }}
+                disabled={dataset?.timeFieldName !== undefined}
```
kavilla, line 124: does this disable the moment the user switches the timeFieldName?

for dataset.type === `INDEX_PATTERN` did we want to consider just hiding this select? but i haven't verified that we are updating the cache correctly if the user wants to update the timefield for non index patterns. so perhaps we should consider letting them set this even if the data set is created and hiding it in the case of index_pattern.

kavilla: for some reason cant seem to set `No time field`  to be undefined for latest main. It is setting it to `No time field` can we verify this?

No time field should map to undefined and time field should should then be undefined.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
kavilla: +1,

the correct flow should be the timeFieldName can be undefined and it should not set it in the dataset.

Honestly I think we should just be consistent with the index pattern creation.

Insert an empty value as the first value and that is the default value and disables the submit.

The `-----` does not set the value of timeFieldName

The `I dont want to use time filter` then enables the submit but ensures the time field name is not set.

Screenshot 2024-09-12 at 1 57 13 PM

src/plugins/data/public/ui/dataset_selector/configurator.tsx
sejli: > Or disable only type index pattern
@ashwin-pc also suggested this, updated in 3b77d5b. I think it'll be helpful to users to show whether or not there is already a time field assigned to the index pattern. Also, if there are no date fields, then the selector won't show at all.

> for some reason cant seem to set No time field to be undefined for latest main. It is setting it to No time field can we verify this?
No time field should map to undefined and time field should should then be undefined.

It works for me on `main`; I'm able to set `No time field` to undefined and it'll load correctly. One thing to note is that for PPL, we may need to update the search interceptor/strategy since it'll automatically append a `timefield` clause, which will break if we select no time field.

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -64,14 +63,11 @@ export class DatasetService {
...
   public async cacheDataset(dataset: Dataset): Promise<void> {
     const type = this.getType(dataset.type);
-    if (dataset) {
+    if (dataset && dataset.type !== DEFAULT_DATA.SET_TYPES.INDEX_PATTERN) {
       const spec = {
         id: dataset.id,
         title: dataset.title,
-        timeFieldName: {
-          name: dataset.timeFieldName,
-          type: 'date',
```
kavilla, line 73: i think i added type here for querying purposes.

src/plugins/data/public/query/query_string/language_service/types.ts
```diff
@@ -51,4 +51,5 @@ export interface LanguageConfig {
   showDocLinks?: boolean;
   editorSupportedAppNames?: string[];
   supportedAppNames?: string[];
+  noTimeField?: boolean;
```
kavilla, line 54: nit: perhaps it should be apart of the fields? as `aggregatable` as there is no field that can get any aggregations for

kavilla: if not that then perhaps something like disableDatePicker

since noTimeField might be a little confusion for it's functionality

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -51,6 +52,18 @@ export const Configurator = ({
...
   });
 
+  const requiresTimeField = (ds: Dataset, lang: string, fields: DatasetField[]) => {
+    if (
+      languageService.getLanguage(lang)?.noTimeField ||
+      ds.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN
+    ) {
+      return false;
+    }
+    if (fields && fields.length > 0) {
```
kavilla, line 62: what if they have no fields of type `date`? i think it's right though and matches the behavior othe index pattern management page that forces them still to click `i dont want to use a time filter`

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -97,7 +110,20 @@ export const Configurator = ({
           >
             <EuiFieldText disabled value={dataset.title} />
           </EuiFormRow>
-          {timeFields && timeFields.length > 0 && (
+          {dataset.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN &&
```
kavilla, line 113: what if type index? or s3? this will never display correct?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -112,15 +138,16 @@ export const Configurator = ({
                     text: field.displayName || field.name,
                     value: field.name,
                   })),
-                  { text: '-----', value: '', disabled: true },
-                  { text: 'No time field', value: undefined },
+                  { text: '-----', value: '-----', disabled: true },
+                  { text: noTimeFilter, value: noTimeFilter },
```
kavilla, line 142: oh this is weird. so undefined never worked? or was it because we were trying to compare to a string `undefined` ?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -112,15 +138,16 @@ export const Configurator = ({
...
                 ]}
                 value={timeFieldName}
                 onChange={(e) => {
-                  const value = e.target.value === 'undefined' ? undefined : e.target.value;
-                  setTimeFieldName(value);
+                  const value = e.target.value === noTimeFilter ? undefined : e.target.value;
+                  setTimeFieldName(e.target.value);
```
kavilla, line 147: how come not `value`?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -112,15 +138,16 @@ export const Configurator = ({
...
                   setDataset({ ...dataset, timeFieldName: value });
                 }}
+                hasNoInitialSelection={dataset.type === DEFAULT_DATA.SET_TYPES.INDEX}
```
kavilla, line 150: how about !== INDEX_PATTERN?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -165,6 +192,7 @@ export const Configurator = ({
             onConfirm(dataset);
           }}
           fill
+          disabled={timeFieldName === undefined && requiresTimeField(dataset, language, timeFields)}
```
kavilla, line 195: this might be an issue because we don't want to query the dataset source with a string it should be undefined

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
sejli: Is there any functionality I can test to see if changing this breaks something? The `timeFieldName` not being set properly is why time fields never worked when going through the configurator.

src/plugins/data/public/query/query_string/language_service/types.ts
sejli: Just to clarify, is `aggregatable` only referring to `date` fields? Is there anything else that could mean aggregatable for SQL? Putting as `disableDatePicker` until I have a better understanding of that.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
sejli: Yeah, this should only be displayed for index patterns. Spoke with @ashwin-pc on this one, index patterns are the only type with a time field set, so we should display what it is and not allow any edits.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
sejli: `OuiSelectable` has some weird properties when attempting to match an `undefined` value prop with what's in the array. Made it easier to just set the values as what the text is and change it in the `onChange`.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
sejli: `e.target.value` refers to the `text` displayed, so it won't actually be set as the `value` prop for the option.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
sejli: Yes

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -51,6 +52,18 @@ export const Configurator = ({
     return languages[0];
   });
 
+  const requiresTimeField = (ds: Dataset, lang: string, fields: DatasetField[]) => {
```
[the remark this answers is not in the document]
sejli: Yeah, I agree. Was just a quick name I made on the fly, I'll change it to `displayTimeFieldSelector`.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
sejli: Should return false if no fields of type `date`, means there isn't any fields for the date picker to use.

src/plugins/data/public/query/query_string/language_service/types.ts
kavilla: disabledatepicker is cool for me. or even if we want to keep in line with `showDocLinks` it could be `showDatePicker`

src/plugins/data/public/ui/dataset_selector/configurator.tsx
kavilla: true but i think we might want to still get the user to select that they will have no fields to use.

i've seen OpenSearch automatically mapped a timestamp value to a string, so if we hide the timefield selector if we don't find any date fields I can see that being confusing if the user expects that they should have a timestamp field. but if they have a dropdown that provides no option besides that they dont want a time filter then it will provide a little bit more insight to the user that they are choosing to not aggregate their data on a timestamp

src/plugins/data/public/ui/dataset_selector/configurator.tsx
kavilla: this seem like a bug

src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -100,6 +100,7 @@ export class QueryEnhancementsPlugin
       editor: enhancedSQLQueryEditor,
       editorSupportedAppNames: ['discover'],
       supportedAppNames: ['discover', 'data-explorer'],
+      disableDatePicker: true,
```
kavilla, line 103: nit: this my fault for suggesting something inaccurate and feel free to ignore this comment. but after thinking about it `disable` for me would make me believe that this will still render it but show it disabled. If it does that then this is perfect name but if it's hiding then perhaps we should go with the `showDatePicker`

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -41,8 +41,9 @@ export const Configurator = ({
   const languages = type?.supportedLanguages(baseDataset) || [];
 
   const [dataset, setDataset] = useState<Dataset>(baseDataset);
-  const [timeFields, setTimeFields] = useState<DatasetField[]>();
+  const [timeFields, setTimeFields] = useState<DatasetField[]>([]);
   const [timeFieldName, setTimeFieldName] = useState<string | undefined>(dataset.timeFieldName);
```
kavilla, line 45: oh i think i see why the issue above you mentioned. it is a string here but it's not a string it's a field object

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -51,6 +52,19 @@ export const Configurator = ({
...
+    if (
+      languageService.getLanguage(lang)?.disableDatePicker ||
+      ds.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN
+    ) {
+      return false;
+    }
+    if (fields && fields.length > 0) {
+      return true;
+    }
+    return false;
```
kavilla, line 65: nit: regardless i think we can simplify this logic out more like 

```typescript
return !(languageService.getLanguage(lang)?.disableDatePicker || ds.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN) || !fields && fields.length > 0;
```

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -97,7 +111,20 @@ export const Configurator = ({
           >
             <EuiFieldText disabled value={dataset.title} />
           </EuiFormRow>
-          {timeFields && timeFields.length > 0 && (
+          {dataset.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN &&
```
kavilla, line 114: could this call the same function displaytimefieldselector? and negate it?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -41,8 +41,9 @@ export const Configurator = ({
   const languages = type?.supportedLanguages(baseDataset) || [];
 
   const [dataset, setDataset] = useState<Dataset>(baseDataset);
-  const [timeFields, setTimeFields] = useState<DatasetField[]>();
+  const [timeFields, setTimeFields] = useState<DatasetField[]>([]);
   const [timeFieldName, setTimeFieldName] = useState<string | undefined>(dataset.timeFieldName);
+  const noTimeFilter = 'I dont want to use the time filter';
```
kavilla, line 46: should be able to pass in double quotes for `I don't`

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/index_pattern_management/public/components/create_index_pattern_wizard/lib/extract_time_fields.ts#L59

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -51,6 +52,17 @@ export const Configurator = ({
...
+        ds.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN
+      ) &&
+      fields &&
+      fields.length > 0
+    );
+  };
+
   useEffect(() => {
     const fetchFields = async () => {
       const datasetFields = await queryString
```
kavilla, line 68: we should add a follow up or a spinner to prevent the date field from loading in afterwards

kavilla (approved):
lgtm! just a comment about the name of the config to ensure consistency.

also should we just with @opensearch-project/opensearch-ux . the language selector is below the timefield selection but the language will display or hide the the timefield selection so the language is shifting. should we consider a different order?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
joshuali925: what about cloudwatch and other custom types? i thought we want to filter cloudwatch logs by timestamp

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -51,6 +52,17 @@ export const Configurator = ({
     return languages[0];
   });
 
+  const displayTimeFieldSelector = (ds: Dataset, lang: string, fields: DatasetField[]) => {
```
joshuali925, line 55: `shouldDisplayTimeFieldSelector`? since calling this doesn't seem to display the selector

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -112,15 +137,16 @@ export const Configurator = ({
                     text: field.displayName || field.name,
                     value: field.name,
                   })),
-                  { text: '-----', value: '', disabled: true },
-                  { text: 'No time field', value: undefined },
+                  { text: '-----', value: '-----', disabled: true },
+                  { text: noTimeFilter, value: noTimeFilter },
```
joshuali925, line 141: `noTimeFilter` is storing user facing text, could `text` and `value` be separated? text should have i18n and value should be a fixed key

src/plugins/data/public/ui/dataset_selector/configurator.tsx
sejli: This specific label is for index patterns only and shows a disabled `OuiFieldText` that displays the time field associated with the index pattern if there is one. It's only disabled for index patterns because they already have a time field attached upon creation, if specified. If not, there will be an option to select a time field with [this selector](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/5ac93609c7947de5f772b6443e285c9f84de55f8/src/plugins/data/public/ui/dataset_selector/configurator.tsx#L125-L152).

src/plugins/data/public/ui/dataset_selector/configurator.tsx
sejli: Keeping the value as an empty string or undefined would result in some weird behaviors when the `value` prop for `OuiSelectable` is undefined, which is why I made them the same.

kavilla (approved):

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -51,6 +52,17 @@ export const Configurator = ({
     return languages[0];
   });
 
+  const shouldDisplayTimeFieldSelector = (ds: Dataset, lang: string, fields: DatasetField[]) => {
```
ashwin-pc, line 55:
```suggestion
  const shouldDisplayTimeFieldSelector = (ds: Dataset, lang: string, timeFields: DatasetField[]) => {
```
nit: reads easier. Had to go lower in the code to see if you were actually only passing time fields to make sure that you were passing the correct list

src/plugins/data/public/ui/dataset_selector/configurator.tsx
ashwin-pc: The logic here reads wrong. Correct me if my understanding is wrong but shouldnt the logic be like this:

```js
if(languageService.getLanguage(language)?.disableDatePicker) {
  return null
}

if(dataset.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN) {
  return DisabledInput
} else {
  return Select
}
```

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -112,15 +137,16 @@ export const Configurator = ({
...
                 ]}
                 value={timeFieldName}
                 onChange={(e) => {
-                  const value = e.target.value === 'undefined' ? undefined : e.target.value;
-                  setTimeFieldName(value);
+                  const value = e.target.value === noTimeFilter ? undefined : e.target.value;
+                  setTimeFieldName(e.target.value);
                   setDataset({ ...dataset, timeFieldName: value });
                 }}
+                hasNoInitialSelection={dataset.type !== DEFAULT_DATA.SET_TYPES.INDEX_PATTERN}
```
ashwin-pc, line 149: Do we still need this with the logic i proposed?

src/plugins/query_enhancements/public/plugin.tsx
ashwin-pc: +1

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -51,6 +52,17 @@ export const Configurator = ({
...
+  const shouldDisplayTimeFieldSelector = (ds: Dataset, lang: string, fields: DatasetField[]) => {
+    return (
+      !(
+        languageService.getLanguage(lang)?.disableDatePicker ||
+        ds.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN
+      ) &&
+      fields &&
+      fields.length > 0
+    );
+  };
```
ashwin-pc, line 64:
```suggestion
  const submitDisabled = useMemo (() => {
    return !(
      timeFieldName === undefined &&
      !(
        languageService.getLanguage(lang)?.disableDatePicker ||
        dataset.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN
      ) &&
      timeFields &&
      timeFields.length > 0
    );
  }, [dataset, language, timeFieldName, timeFields, languageService]);
```
nit: Also if you use the logic i have suggested before, this name probably wont be appropriate anymore. Then its only usecase is to disable the select button on the bottom. And for that, id just keep the logic colocated here.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -165,6 +191,10 @@ export const Configurator = ({
             onConfirm(dataset);
           }}
           fill
+          disabled={
+            timeFieldName === undefined &&
+            shouldDisplayTimeFieldSelector(dataset, language, timeFields)
+          }
```
ashwin-pc, line 197:
```suggestion
          disabled={submitDisabled}
```

ashwin-pc (commented):
Just lemme know if my functional logic makes sense.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
ashwin-pc: Irrelevant if you accept my suggested logic

ashwin-pc:
> lgtm! just a comment about the name of the config to ensure consistency. but if we wanted to do that we should do that as a fast follow and not block this pr.
> 
> also should we just with @opensearch-project/opensearch-ux . the language selector is below the timefield selection but the language will display or hide the the timefield selection so the language is shifting. should we consider a different order?

Yeah lets flip the order!

src/plugins/data/public/ui/dataset_selector/configurator.tsx
sejli: Adding this logic in 51e4490

src/plugins/data/public/ui/dataset_selector/configurator.tsx
sejli: Nope, should default to true then. Updated in 51e4490

src/plugins/data/public/mocks.ts
```diff
@@ -91,6 +91,12 @@ const createStartContract = (isEnhancementsEnabled: boolean = false): Start => {
         })
       ),
       clearCache: jest.fn(),
+      create: jest.fn().mockResolvedValue({
```
kavilla, line 94: might need to update the sanpshot for the related specific failed test

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -144,6 +130,46 @@ export const Configurator = ({
...
+                )}
+              >
+                <EuiSelect
+                  options={[
+                    ...timeFields.map((field) => ({
+                      text: field.displayName || field.name,
+                      value: field.name,
+                    })),
+                    { text: '-----', value: '-----', disabled: true },
+                    { text: noTimeFilter, value: noTimeFilter },
```
joshuali925, line 161: i think this is fine but it's better to separate them, since `text` can change (e.g. different translation, future UX wording update), and `value` should not change

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -41,8 +41,9 @@ export const Configurator = ({
   const languages = type?.supportedLanguages(baseDataset) || [];
 
   const [dataset, setDataset] = useState<Dataset>(baseDataset);
-  const [timeFields, setTimeFields] = useState<DatasetField[]>();
+  const [timeFields, setTimeFields] = useState<DatasetField[]>([]);
   const [timeFieldName, setTimeFieldName] = useState<string | undefined>(dataset.timeFieldName);
+  const noTimeFilter = "I don't want to use the time filter";
```
joshuali925, line 46: this should have the translation id in the create index pattern page 
image

joshuali925 (dismissed):

src/plugins/data/public/ui/dataset_selector/configurator.tsx
ashwin-pc: +1

ashwin-pc (approved):

joshuali925 (approved):

PR #8168, opened 2024-09-12
Sync query and filter when refreshing discover page

abbyhu2000:
### Description

This PR resolves previously identified three similar bugs when using legacy discover:

Query and filter failed to persist when: 
* refresh the page
* switching between default discover table and datagrid
* using the share link

## Screenshot

https://github.com/user-attachments/assets/2b8e5b54-756a-4435-ba52-284e7c875616

## Changelog

- fix: Sync query and filter when refreshing discover page

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/view_components/utils/use_search.ts
+++ b/src/plugins/discover/public/application/view_components/utils/use_search.ts
@@ -324,6 +324,11 @@ export const useSearch = (services: DiscoverViewServices) => {
       const savedSearchInstance = await getSavedSearchById(savedSearchId);
       setSavedSearch(savedSearchInstance);
 
+      // if saved search does not exist, do not atempt to sync filters and query from savedObject
+      if (!savedSearch) {
+        return;
+      }
+
       // sync initial app filters from savedObject to filterManager
       const filters = cloneDeep(savedSearchInstance.searchSource.getOwnField('filter'));
       const query =
--- a/src/plugins/discover/public/application/view_components/utils/use_search.ts
+++ b/src/plugins/discover/public/application/view_components/utils/use_search.ts
@@ -321,13 +321,13 @@ export const useSearch = (services: DiscoverViewServices) => {
   // Get savedSearch if it exists
   useEffect(() => {
     (async () => {
-      // if saved search does not exist, do not atempt to sync filters and query from savedObject
-      if (!savedSearchId) {
-        return;
-      }
-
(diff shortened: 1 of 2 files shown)


ashwin-pc (approved):
Can we add a test for this as a fast follow?

kavilla (approved):

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -324,6 +324,11 @@ export const useSearch = (services: DiscoverViewServices) => {
       const savedSearchInstance = await getSavedSearchById(savedSearchId);
       setSavedSearch(savedSearchInstance);
 
+      // if saved search does not exist, do not atempt to sync filters and query from savedObject
+      if (!savedSearch) {
```
kavilla, line 328: i think the linter will complain that this isn't included in the dependency right? but then 325 sets this value.

i can see in the future someone listening to the linter and including the savedsearch in the dependencies.

shouldn't we just do 
```typescript
if (!savedSearchInstance) {
```

abbyhu2000: changed to use `savedSearchId` instead since it is already a dependency

abbyhu2000: savedSearchInstance does not work here because it is always defined

kavilla (approved):

sejli (dismissed):

src/plugins/discover/public/application/view_components/utils/use_search.ts
abbyhu2000: `savedSearchId` here wont work as well, because when we create a new saved search, the id is null, but we do not want to return here.

kavilla (approved):

sejli (approved):

abbyhu2000:
Manual 2.17 backport : https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8179

abbyhu2000:
link check failure is not related: `* [404] [http://sites.google.com/site/murmurhash/](http://sites.google.com/site/murmurhash/) | Failed: Network error: Not Found`

neodescis:
This fix appears to have caused a very ugly bug where filters added on a dashboard or in the visualization editor then appear on the Discover page when you navigate there.

PR #8171, opened 2024-09-12
[Page Header] Add responsiveness for page header

zhongnansu:
### Description

- Add responsiveness for page header
- Stop converting data source picker actions to popover

## Screenshot

https://github.com/user-attachments/assets/fb34e737-dfd9-42d5-b948-596edd6d815d

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


zhongnansu:
> Just to confirm, the additional whitespace after removing the popover behavior will be addressed separately?
@virajsanghvi What's the whitespace you are referring to? Is it the one from this screenshot?
image

src/core/public/chrome/ui/header/header.tsx
```diff
@@ -443,22 +445,39 @@ export function Header({
...
+          </EuiHeaderSection>
+
+          {/* Right Section */}
+          <EuiHeaderSection side="right">
+            <EuiFlexGroup gutterSize="s">
+              <EuiFlexItem grow={false}>{renderCenterControls()}</EuiFlexItem>
+
+              <EuiFlexItem grow={false}>{renderActionMenu()}</EuiFlexItem>
+
+              <EuiFlexItem>{renderRightControls()}</EuiFlexItem>
```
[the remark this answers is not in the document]
zhongnansu: Yes,  side="right" prop ensures that the section is aligned to the right, and its child elements will not grow. Removing all grow=false

virajsanghvi:
> > Just to confirm, the additional whitespace after removing the popover behavior will be addressed separately?
> > @virajsanghvi What's the whitespace you are referring to? Is it the one from this screenshot?
> > image

Sorry, the space I'm referring to is the whitespace above the area selected in your screenshot.

zhongnansu:
@virajsanghvi Fixed the spacing issue, and updated the attached video, could you re-review?

src/core/public/chrome/ui/header/header.tsx
```diff
@@ -281,93 +284,186 @@ export function Header({
...
+    observables.navControlsExpandedCenter$.subscribe((value) => {
+      hasNavControlsExpandedCenter = isTruthy(value);
+    });
+    observables.navControlsRight$.subscribe((value) => {
+      hasNavControlsRight = isTruthy(value);
+    });
+
+    if (
+      !hasRightControls &&
+      !hasNavControlsExpandedCenter &&
```
virajsanghvi, line 387: curious why does this matter?

src/core/public/chrome/ui/header/header.tsx
```diff
@@ -443,22 +539,45 @@ export function Header({
...
+            <EuiFlexGroup gutterSize="s">
+              <EuiFlexItem grow={false}>
+                <EuiHeaderSectionItem border="none" data-test-subj="headerApplicationTitle">
+                  <EuiTitle size="l" className="newTopNavHeaderTitle">
+                    {breadcrumbs && <h1>{breadcrumbs[breadcrumbs.length - 1]?.text}</h1>}
+                  </EuiTitle>
+                </EuiHeaderSectionItem>
+              </EuiFlexItem>
+
+              {renderBadge() && <EuiFlexItem grow={false}>{renderBadge()}</EuiFlexItem>}
```
virajsanghvi, line 558: instead of evaluating these twice, can we assign to a variable and reuse?

src/core/public/chrome/utils.ts
```diff
@@ -221,3 +221,11 @@ export const getVisibleUseCases = (navGroupMap: Record<string, NavGroupItemInMap
...
   );
 };
+
+export const isTruthy = (value: any) => {
+  if (Array.isArray(value) && value.length === 0) {
+    return false;
+  }
+
+  return Boolean(value);
+};
```
virajsanghvi, line 231: The word "truthy" has an expectation in js, but the empty array check here breaks that. Can we rename this or if its only used in one file (which it looks like), can just put it there?

src/core/public/chrome/ui/header/header.tsx
```diff
@@ -281,93 +284,186 @@ export function Header({
...
 
-  const renderCenterControls = () => (
-    <>
-      {useUpdatedHeader && (
-        <EuiHeaderSectionItem border="none">
-          <HeaderNavControls navControls$={observables.navControlsExpandedCenter$} />
-        </EuiHeaderSectionItem>
-      )}
+    if (!hasLeftControls && !hasNavControlsLeft && useUpdatedHeader) {
+      return null;
```
virajsanghvi, line 299: Are there any consumers that rely on the presence of the dom element to insert content? Mostly just asking if this sparks any known use cases we could potentially break here, but I'm hoping no one is doing manual dom lookups to insert content.

virajsanghvi (commented):
I think we will want to address vertical padding between items but can be done after this cr: 
image

zhongnansu:
> I think we will want to address vertical padding between items but can be done after this cr: image

@virajsanghvi This comes with the default EuiFlexItem, to fix it, we need to update in OUI

virajsanghvi:
> > I think we will want to address vertical padding between items but can be done after this cr: image
> 
> @virajsanghvi This comes with the default EuiFlexItem, to fix it, we need to update in OUI

Can we not address with using smaller padding sizes on the elements or a css override? Anyways, may be easier to identify options once its pushed.

src/core/public/chrome/ui/header/header.tsx
zhongnansu: So based on my observation, even if there's nothing mounted to those header controls mountpoint . It still renders a empty <EuiHeaderSectionItem>, and taking up spaces with no content. That's where the empty spaces are coming from. So adding this check to remove the container when no content

src/core/public/chrome/ui/header/header.tsx
zhongnansu: No as far as I know, finger crossed

src/core/public/chrome/ui/header/header.tsx
virajsanghvi: To be clear, my question is caring about expandedCenter when rendering right items... Just curious why this matters here whereas it doesn't anywhere else

src/core/public/chrome/ui/header/header.tsx
zhongnansu: yeah will do,

src/core/public/chrome/utils.ts
zhongnansu: will rename to hasValue, and move to the file that uses it

src/core/public/chrome/ui/header/header.tsx
zhongnansu: I added here as the condition, since for old UI, we still have this mountpoint and in case anyone is using it(not as far as I know, also Miki documented this is unused), it can still be rendered as expected, instead of returning null. Just aligning with existing code, which made this "expandedCenter" par of overall right control
```

    return (
      <>
        {useUpdatedHeader && (
          <EuiHeaderSectionItem border="none">
            <HeaderNavControls navControls$={observables.navControlsExpandedRight$} />
          </EuiHeaderSectionItem>
        )}

        {useUpdatedHeader && (
          <EuiHeaderSectionItem border="none">
            <HeaderControlsContainer
              data-test-subj="headerRightControl"
              controls$={application.currentRightControls$}
            />
          </EuiHeaderSectionItem>
        )}

        <EuiHeaderSectionItem border={useUpdatedHeader ? 'none' : 'left'}>
          <HeaderNavControls side="right" navControls$={observables.navControlsRight$} />
        </EuiHeaderSectionItem>
      </>
    );
```

virajsanghvi: Sorry, where is expanded center here? I only see expandedRight

zhongnansu: it's a typo, it should be expandedRight

src/core/public/chrome/ui/header/header.scss
```diff
@@ -110,3 +110,11 @@
 #globalHeaderBars:has(.primaryApplicationHeader) {
   border-bottom: 1px solid $euiColorLightShade;
 }
+
+/* Control flex items to wrap when the viewport is less than the medium size */
+@media (max-width: 768px) {
```
virajsanghvi, line 115: nit: are $euiBreakpoints available here? If so, could use `$euiBreakpoints['m']` instead of the hard coded value

virajsanghvi (commented):
Change looks good assuming checks pass

virajsanghvi (approved):

ZilongX (approved):

Issue #330, opened 2021-05-05
[Bug] Ensure OpenSearch Dashboards stays available in large clusters


boktorbb-amzn:
### Problem Statement

For sufficiently large OpenSearch clusters, when Dashboards sends a health check out there can be a failure and Dashboards becomes unavailable.

### Root Cause of the issue

The default Dashboards behavior is to fan out healthcheck requests across the entire cluster. For large clusters, if any nodes are processing or ingestion heavy and time out then it fails the healthcheck and Dashboards becomes unavailable.

### Proposed Dashboards solution

The proposal is to create effectively a similar node attribute and use that for an optimized healthcheck

Dashboards Configuration: Create a setting that’s called `optimized_healthcheck` in `opensearch_dashboards.yml` that looks for the OpenSearch Node attribute `cluster_id`  By default, `optimized_healthcheck` will default to `null` which lets Dashboards continue fanning out healthcheck requests across all nodes. If the value is `cluster_id` , then it will switch to the logic outlined in the below algorithm section.

OpenSearch Configuration: `cluster_id` → a new node attribute to be added in by customers that would differentiate cluster instances 

cluster_id can take the form of an integer that is assigned during cluster creation to all OpenSearch nodes. It will increment when new instances of the cluster are spun up

Using this cluster_id we can follow a general algorithm: 

Step 1: Aggregate all cluster_id for OpenSearch nodes
Step 2: if all the nodes have same `cluster_id`, retrieve `nodes.info` from `_local` node only.
    - Using `_cluster/state/nodes` to retrieve from each master node 
Step 3: if the nodes have different `cluster_id`, fan out the request to all the nodes.

![Screen Shot 2021-05-14 at 11 13 32 AM](https://user-images.githubusercontent.com/65934617/118311919-6ca09980-b4a5-11eb-9446-fd65872a83c5.png)

- [x] Design Doc
- [ ] Implementation

PR #8187, opened 2024-09-13
Use local clusterState call during healthchecks

gargharsh3134:
### Description

During health checks, while trying to get the optimised health checks property for all the nodes, a cluster state call is made to Opensearch which eventually is served by the ClusterManager node and hence stresses it out. For large clusters having high number of indices and shards, the cluster state size grows and serving cluster state requests from each node starts to degrade the cluster manager's performance. The attribute for each node can very well be served from local cluster state of the node. As per cluster manager's functioning, a node whose cluster state is lagging will be kicked out of the cluster after 90 seconds of waiting. So, serving clusterState from local for OSD process should not lead to any functionality change, except a little time lag in case the cluster state is lagging on the local node. 

This change just updates the clusterState call to use the ```local=true``` query param, so that the request doesn't impact ClusterManager node.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


ruanyl (dismissed):

src/core/server/opensearch/version_check/ensure_opensearch_version.ts
```diff
@@ -76,6 +76,7 @@ export const getNodeId = async (
      */
     const state = (await internalClient.cluster.state({
       metric: 'nodes',
+      local: true,
```
ruanyl, line 79: Looking at the comment on this function call, it seems calling cluster manager node is by intention. Shall we adjust the comment if using `local: true` is what we want?

ruanyl:
Could you please mark this PR to be ready to review? 
The code change looks fine to me, but I left a question about this old comment:
```
    /*
     * Using _cluster/state/nodes to retrieve the cluster_id of each node from cluster manager node which
     * is considered to be a lightweight operation to aggegrate different cluster_ids from the OpenSearch nodes.
     */
```

src/core/server/opensearch/version_check/ensure_opensearch_version.ts
gargharsh3134: Ack, will update the code comment. Thanks for pointing it out.

ashwin-pc (commented):
Nice! Can you add a changeset to the PR. I think this change is worth calling out in the changelog. Will approve once thats ready.

gargharsh3134:
> Nice! Can you add a changeset to the PR. I think this change is worth calling out in the changelog. Will approve once thats ready.

@ashwin-pc Thanks for taking a look. I have added the changelog fragment, please take a look again.

gargharsh3134:
> Could you please mark this PR to be ready to review? The code change looks fine to me, but I left a question about this old comment:
> 
> ```
>     /*
>      * Using _cluster/state/nodes to retrieve the cluster_id of each node from cluster manager node which
>      * is considered to be a lightweight operation to aggegrate different cluster_ids from the OpenSearch nodes.
>      */
> ```

@ruanyl Thanks for looking into it. I have updated the code comment, please check again.

ruanyl:
@gargharsh3134 While I was diving into the original implementation, I found the related issue of it: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/330

I'm not familiar with the context of it, would you mind to take a look and see if your changes are compatible with it?

gargharsh3134:
> @gargharsh3134 While I was diving into the original implementation, I found the related issue of it: #330
> 
> I'm not familiar with the context of it, would you mind to take a look and see if your changes are compatible with it?

@ruanyl I went through the linked issue, and looks like the new changes should maintain compatibility. The issue calls out the problem of fanning out the health check calls directly to all the nodes, and incase some of the nodes are slow and under resource constraint, they can bring down the OSD process. The proposed solution was to just make a single _cluster/state call to figure out when to really fan out these health checks to all the nodes, instead of doing it each time.

In the existing implementation, the request path was OSD making a call to local OS process which then is calling clusterManager node, once clusterManager responds back to local OS, the same response is served to OSD. With the new changes, the only difference is, instead of local node calling cluster manager, it would serve the request from it's own local cluster state (which might be lagging if the node is slow but would eventually catch up).

ashwin-pc (approved):

ruanyl (approved):

PR #8013, opened 2024-09-05
[Workspace] Integrate workspace front end with data connection type saved object

raintygao:
### Description

Integrate workspace with data connection type saved object in frontend

## Screenshot
image
image

## Testing the changes

Use APIs in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/7925 to create data connection object, then goes to workspace create page to assign/unassign data connection objects.

## Changelog

- feat: Integrate workspace with data connections in front end

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #8200, opened 2024-09-14
[Workspace] feat: update workspace server to support data connection type

raintygao:
### Description

Update workspace server to support data connection type

This is a PR picked from https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8013, as we want to separate frontend and server changes and merge server changes first for testing, so we extract server changes from https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8013 and create this PR.

### Create workspace
post http://localhost:6601/api/workspaces
```
{
    "attributes": {
        "name": "workspace-name1",
        "color": "#54B399",
        "features": [
            "use-case-observability"
        ]
    },
    "settings": {
        "dataSources": [],
        "dataConnections": [
            "12345" // data connection object id
        ],
        "permissions": {
            "library_write": {
                "users": [
                    "%me%"
                ]
            },
            "write": {
                "users": [
                    "%me%"
                ]
            }
        }
    }
}
```
Result
```
{
    "success": true,
    "result": {
        "id": "_IlxNR"
    }
}
```

### Update workspace
put http://localhost:6601/api/_IlxNR
```
{
    "attributes": {
        "name": "workspace-name1",
        "color": "#54B399",
        "features": [
            "use-case-observability"
        ]
    },
    "settings": {
        "dataSources": [],
        "dataConnections": [
            "12346" // new data connection object ids that you want to associate
        ],
        "permissions": {
            "library_write": {
                "users": [
                    "%me%"
                ]
            },
            "write": {
                "users": [
                    "%me%"
                ]
            }
        }
    }
}
```
Result
```
{
    "success": true,
    "result": true
}
```

### Verify
get data-connection objects with workspace field 
http://localhost:6601/api/saved_objects/_find?fields=id&fields=title&fields=workspaces&per_page=10000&type=data-connection&workspaces=*

Result
image

## Changelog

- feat: update workspace server to support data connection type

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source/opensearch_dashboards.json
+++ b/src/plugins/data_source/opensearch_dashboards.json
@@ -6,5 +6,5 @@
   "ui": true,
   "requiredPlugins": ["opensearchDashboardsUtils"],
   "optionalPlugins": [],
-  "extraPublicDirs": ["common/data_sources", "common/util"]
+  "extraPublicDirs": ["common","common/data_sources", "common/util"]
 }
--- a/src/plugins/workspace/public/workspace_client.ts
+++ b/src/plugins/workspace/public/workspace_client.ts
@@ -204,6 +204,7 @@ export class WorkspaceClient implements IWorkspaceClient {
     settings: {
       dataSources?: string[];
       permissions?: SavedObjectPermissions;
+      dataConnections?: string[];
     }
   ): Promise<IResponse<Pick<WorkspaceAttributeWithPermission, 'id'>>> {
     const path = this.getPath();
@@ -298,6 +299,7 @@ export class WorkspaceClient implements IWorkspaceClient {
     settings: {
       dataSources?: string[];
       permissions?: SavedObjectPermissions;
+      dataConnections?: string[];
     }
   ): Promise<IResponse<boolean>> {
     const path = this.getPath(id);
--- a/src/plugins/workspace/server/workspace_client.ts
+++ b/src/plugins/workspace/server/workspace_client.ts
@@ -28,7 +28,10 @@ import {
   WORKSPACE_ID_CONSUMER_WRAPPER_ID,
   WORKSPACE_SAVED_OBJECTS_CLIENT_WRAPPER_ID,
 } from '../common/constants';
-import { DATA_SOURCE_SAVED_OBJECT_TYPE } from '../../data_source/common';
+import {
+  DATA_SOURCE_SAVED_OBJECT_TYPE,
+  DATA_CONNECTION_SAVED_OBJECT_TYPE,
+} from '../../data_source/common';
 
 const WORKSPACE_ID_SIZE = 6;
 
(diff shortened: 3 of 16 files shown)


src/plugins/data_source/opensearch_dashboards.json
```diff
@@ -6,5 +6,5 @@
   "ui": true,
   "requiredPlugins": ["opensearchDashboardsUtils"],
   "optionalPlugins": [],
-  "extraPublicDirs": ["common/data_sources", "common/util"]
+  "extraPublicDirs": ["common","common/data_sources", "common/util"]
```
SuZhou-Joe, line 9: Seems like a new pattern to me, so it will bypass the lint check?

raintygao: I suppose would, I checked some other usage before modified.
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/opensearch_dashboards.json
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/expressions/opensearch_dashboards.json

SuZhou-Joe (approved):
LGTM, so the permission on the data connections will be raised in another PR?

raintygao:
> LGTM, so the permission on the data connections will be raised in another PR?

Yes, this will be addressed and included in later PRs.

src/plugins/workspace/server/workspace_client.ts
```diff
@@ -260,11 +280,37 @@ export class WorkspaceClient implements IWorkspaceClientImpl {
...
+          }
+        }
+        if (dataConnectionsToBeAdded.length > 0) {
+          for (const dataConnectionId of dataConnectionsToBeAdded) {
+            promises.push(
+              client.addToWorkspaces(DATA_CONNECTION_SAVED_OBJECT_TYPE, dataConnectionId, [id])
+            );
+          }
         }
       }
```
SuZhou-Joe, line 309: Nit: seems like the logic to handle the data connection are the same as data source, would it be possible to make it into a function like `addOperationsToPromise(type: string)`?

So the code here will be like
```
const promises = [];
const addOperationsToPromise = (type: string) => { xxx }

// Add data sources operations into promise
addOperationsToPromise(DATA_SOURCE_SAVED_OBJECT_TYPE);

// Add data connections operations into promise
addOperationsToPromise(DATA_CONNECTION_SAVED_OBJECT_TYPE);
```

raintygao: Sounds good, will optimize this part in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8013.

src/plugins/workspace/public/workspace_client.ts
```diff
@@ -204,6 +204,7 @@ export class WorkspaceClient implements IWorkspaceClient {
     settings: {
       dataSources?: string[];
       permissions?: SavedObjectPermissions;
+      dataConnections?: string[];
```
ruanyl, line 207: Why not reuse `dataSources`? It doesn't seem necessary to add an extra parameter `dataConnections`. Because for workspace, they are all data sources, I personally don't want it to be implemented in a way that every time adding new type of data source, it requires to change the API interface.

ruanyl: what's passed in here is just saved object ids, we could handle different type of data source internally.

raintygao: Yes, if we reuse dataSources field, we need to update this field from [‘id1’,‘id2’] to [{type:‘data-source’,id:‘id1’},{type:‘data-connection’,id:‘id2’], which would make a breaking change to existing API.

ruanyl (approved):

PR #8205, opened 2024-09-16
fix for link not opening in new tab

### LINKED ISSUE #3500: [BUG] open in a new tab broken in 2.5.0
**Describe the bug**

Opensearch version 2.5.0 (was working in previous versions)

Go to “Stack Management”, then “index patterns”, select an index pattern and edit a field to format it as a URL. The option “Open in a new tab” doesn’t make the links open in a new tab:
Screenshot 2023-02-24 at 7 24 05 PM

The link opens correctly but in the same window and tab whether this option is turned on or off.

This option worked in a previous version (not sure which version that was)

**To Reproduce**
Go to “Stack Management”, then “index patterns”, select an index pattern and edit a field to format it as a URL. The option “Open in a new tab” doesn’t make the links open in a new tab.

**Expected behavior**
The link should open in a new window.

**OpenSearch Version**
2.5.0

**Dashboards Version**
2.5.0

 - OS: docker version on ubuntu host
 - Browser: chrome and safari

niteshthakur:
### Description

This change fixes the issue listed in [Bug 3500](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/3500)
### Issues Resolved

closes #3500 

## Testing the changes

A datatable visualization needs to be created on an index pattern which has a field formatted as URL and set to open in a new tab. On clicking, the url confiured should open in a new tab,

## Changelog
- skip

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

## Release Notes

* **Bug Fixes**
  * Improved security handling for links displayed in table visualizations by ensuring proper attribute preservation and adding security-related protections.

* **Tests**
  * Enhanced test coverage for table cell rendering, including HTML validation, boundary conditions, and link security attribute verification.


src/plugins/vis_type_table/public/components/table_vis_cell.tsx
```diff
@@ -23,6 +23,11 @@ export const getTableVisCellValue = (
   const rawContent = row[columnId];
   const colIndex = columns.findIndex((col) => col.id === columnId);
   const htmlContent = columns[colIndex].formatter.convert(rawContent, 'html');
+  dompurify.addHook('uponSanitizeAttribute', function(currentNode, hookEvent) {
+	if (hookEvent.attrName == 'target'){
+		hookEvent.forceKeepAttr = true;
+	}
+  });
```
AMoo-Miki, line 30: I don't think we should blindly allow any `target` and we shouldn't allow `target="_blank" without the `rel`.
```suggestion
  dompurify.addHook('uponSanitizeAttribute', (node) => {
    if ('target' in node) {
      node.setAttribute('target', '_blank');
      node.setAttribute('rel', 'noopener noreferrer');
    }
  });
```

AMoo-Miki (changes requested):
Also, please let the pre-commit hooks run so you catch the formatting errors.

rayliu2007:
It will be great if the fix could be merged into the next release.

src/plugins/vis_type_table/public/components/table_vis_cell.tsx
niteshthakur: How can you assume that target should be blank every time. It can be self also. Isn't it ? I can do a check to see the value of the attr and set the rel only if the value is _blank

niteshthakur: Also even in the code suggested by you, we are looking at all all the nodes with target attr.

niteshthakur:
> Also, please let the pre-commit hooks run so you catch the formatting errors.

Done.

src/plugins/vis_type_table/public/components/table_vis_cell.tsx
niteshthakur: I have the made the changes related to what I mentioned in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8205#discussion_r1790754687

niteshthakur: added multiple test cases as well.

AMoo-Miki: After further digging, I found that the `target` attribute is managed by the specific formatters. While OSD internally only has the URL formatter that generates anchor tags, other plugins or third-party forks could potentially introduce additional formatters. Therefore, it is the responsibility of each formatter to ensure that a valid `target` value is assigned. However, it is also OSD's responsibility to exercise due diligence and add `rel="noopener noreferrer"` to all links, regardless of whether the `target` attribute is set or its assigned value. This approach will help mitigate potential security risks associated with the improper handling of links.

```
  dompurify.addHook('afterSanitizeElements', (node) => {
    if (node.tagName?.toUpperCase?.() === 'A') {
      node.setAttribute('rel', 'noopener noreferrer');
    }
  });
```

I haven't tested this but lemme know what you think.

AMoo-Miki:
I also noticed that DCO failed. Please see this: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/CONTRIBUTING.md#developer-certificate-of-origin

src/plugins/vis_type_table/public/components/table_vis_cell.tsx
niteshthakur: Why do you want to do this in a new event _afterSanitizeElements_. I would do this as part of _uponSanitizeAttribute_ event. We can add these two values for all the anchor tags when processing the latter event. I agree that adding them to all the anchor tags does help improve the security.

niteshthakur: @AMoo-Miki  I have made the changes. Can it be merged now ?

niteshthakur: @AMoo-Miki gentle reminder to get this reviewed.

m-steen:
@AMoo-Miki @niteshthakur Why has this PR not been merged yet? This bug has been open since v2.5. As far as I can see, all requested changes were done, although your branch probably needs to be rebased with the current version of main.

abbyhu2000:
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11456 has fixed the open link issue. will close duplicate.

PR #8214, opened 2024-09-18
[Discover] Display Cache Time and Clear Cache Button

sejli:
### Description

* Stores last cache updated time in session storage. Last updated time refers to the last time that the cache was updated across all datasets (overall cache rather than specific types).
* Adds display for last cache update time in Advanced Dataset Selector
* Adds refresh cache button which clears session storage

Still early implementation, open to suggestions

### Screenshot

Empty Cache State:

image

Cache has been populated:

image

## Changelog

- feat: Add last updated time and cache refresh button to Discover Advanced Dataset Selector

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -164,6 +165,18 @@ export class DatasetService {
     });
   }
 
+  public refreshCache(): void {
+    this.sessionStorage.clear();
+  }
+
+  public getLastCacheTime(): string | undefined {
```
virajsanghvi, line 172: want to convert to a Date given its passed a date? otherwise consumers just need to handle this

src/plugins/data/public/ui/dataset_selector/dataset_explorer.tsx
```diff
@@ -92,6 +92,29 @@ export const DatasetExplorer = ({
               </EuiLink>
             </p>
           </EuiText>
+          {queryString.getDatasetService().getLastCacheTime() && (
+            <EuiText>
```
virajsanghvi, line 96: size='s'

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -164,6 +165,18 @@ export class DatasetService {
     });
   }
 
+  public refreshCache(): void {
```
virajsanghvi, line 168: nit: is this the right name? not sure what refreshing means...

virajsanghvi (commented):
should we add tests for new methods + snapshot test to bump up patch cov?

kavilla:
a good catch all but did we want to consider making it more granular?

for example:
Screenshot 2024-09-18 at 3 26 56 PM

if i press the refresh on `Data` then it clears out the full cache. If i press refresh on `Database` then it will refresh all the Databases and Tables and not the Connections and Data Sources?

This way to avoid destroying the cache if only one source is out of date?

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
kavilla: yeah we should make sure to do this.

@sejli 
we should probably include the day and should also pulling from UI Settings to get the format of the date
Screenshot 2024-09-18 at 3 37 48 PM

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
kavilla: +1 ideally this should be renamed or this function should refetching the children for the current data structure.

kavilla (commented):
+1 to @virajsanghvi comments. nothing blocking

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -138,6 +138,7 @@ export class DatasetService {
   }
 
   private cacheDataStructure(dataType: string, dataStructure: DataStructure) {
+    this.setLastCacheTime(new Date());
```
AMoo-Miki, line 141: Since this is getting set in the sessionStorage, the Date object gets converted to a string. To consume this, we will need to parse it and what not. To programmatically use date, it is best to stick with the unix timestamp of the date which is number and can be easily parsed and formatted.

```suggestion
    this.setLastCacheTime(Date.now());
```

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -164,6 +165,18 @@ export class DatasetService {
     });
   }
 
+  public clearCache(): void {
+    this.sessionStorage.clear();
+  }
+
+  public getLastCacheTime(): string | undefined {
+    return this.sessionStorage.get('lastCacheTime');
```
AMoo-Miki, line 173: Assuming you do use `Date.now()` this line changes to
```suggestion
    return Number(this.sessionStorage.get('lastCacheTime')) || undefined;
```

The reason I propose `Number` and not `parseInt` is that `parseInt` will try to parse the numeric part of a mixed alphanumeric string which we don't want because the value should be fully numeric, and if it is not, we need to fail to parse.

The reason for `|| undefined` is that we should return `undefined` for all invalid values: `NaN` and `0`.

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -164,6 +165,18 @@ export class DatasetService {
     });
   }
 
+  public clearCache(): void {
+    this.sessionStorage.clear();
+  }
+
+  public getLastCacheTime(): number | undefined {
```
kavilla, line 172: nit i think there is precedent with `updatedAt`

kavilla (approved):
agreeing with @virajsanghvi  with some snapshots will be awesome.

one nit with the naming of the field

one fast follow we should do is pull from ui settings for the date format

src/plugins/data/public/ui/dataset_selector/dataset_explorer.tsx
```diff
@@ -68,31 +72,71 @@ export const DatasetExplorer = ({
...
+              <EuiFlexGroup gutterSize="xs">
+                <EuiFlexItem grow={false}>
+                  <EuiText size="s">
+                    <FormattedMessage
+                      id="data.explorer.datasetSelector.advancedSelector.lastUpdatedTime"
+                      defaultMessage={`Last updated at: ${moment(
+                        queryString.getDatasetService().getLastCacheTime()
+                      )
+                        .format(uiSettings.get('dateFormat'))
+                        .toString()}. `}
```
joshuali925, line 114: this should use variables instead of having a dynamic default message, like
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/b83fdecd9370e5c73d2f7408084f02b6fc5cae56/packages/osd-i18n/README.md?plain=1#L174-L180

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -138,6 +138,7 @@ export class DatasetService {
   }
 
   private cacheDataStructure(dataType: string, dataStructure: DataStructure) {
+    this.setLastCacheTime(Date.now());
```
joshuali925, line 141: is it just a single timestamp for all caches or can different data structures have different cache time?

sejli: It's just a single timestamp for all caches for now. Currently, we haven't had any requirements to make it per data structure/data type.

kavilla (approved):

ashwin-pc (approved):

PR #8247, opened 2024-09-19
[BUG] fixed wrapping of date

sumukhswamy:
### Description
image

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/discover/public/application/components/default_discover_table/table_row.tsx
```diff
@@ -107,18 +107,33 @@ const TableRowUI = ({
...
-              data-test-subj="docTableField"
-              className="osdDocTableCell eui-textBreakAll eui-textBreakWord"
-            >
-              <div className="truncate-by-height">
-                {/* eslint-disable-next-line react/no-danger */}
-                <span dangerouslySetInnerHTML={{ __html: sanitizedCellValue }} />
-              </div>
-            </td>
-          );
+          if (colName === '@timestamp') {
```
AMoo-Miki, line 110: Duplicating this big block just to add a class is detrimental to the maintainability of this piece of code. Please use `classNames` to conditionally add `eui-textBreakAll`.

AMoo-Miki (approved):

src/plugins/discover/public/application/components/default_discover_table/table_row.tsx
```diff
@@ -111,7 +111,9 @@ const TableRowUI = ({
             <td
               key={colName}
               data-test-subj="docTableField"
-              className="osdDocTableCell eui-textBreakAll eui-textBreakWord"
+              className={`osdDocTableCell ${
+                colName === '@timestamp' ? 'eui-textNoWrap' : 'eui-textBreakAll eui-textBreakWord'
```
Hailong-am, line 115: does the column name is fixed as `@timestamp`, i think we can select different filed as time field for index pattern
image

sumukhswamy: I dont quite understand this

Hailong-am: > I dont quite understand this
> 
As you hardcode the column name as `@timestamp`, do we have other column name?

sumukhswamy: have updated the condition for it add this for only timefield columns

src/plugins/discover/public/application/components/default_discover_table/table_row.tsx
```diff
@@ -111,7 +111,11 @@ const TableRowUI = ({
             <td
               key={colName}
               data-test-subj="docTableField"
-              className="osdDocTableCell eui-textBreakAll eui-textBreakWord"
+              className={`osdDocTableCell ${
+                indexPattern.timeFieldName === colName
```
sejli, line 115: Out of curiosity, will there ever be a case where `indexPattern.timeFieldName !== colName`?

sumukhswamy: don't think so, there is already another condition which used the same to check if timefield using the same condition
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/discover/public/application/components/default_discover_table/table_row.tsx#L129

ashwin-pc (approved):

src/plugins/discover/public/application/components/default_discover_table/table_row.tsx
kavilla: > Out of curiosity, will there ever be a case where indexPattern.timeFieldName !== colName?

without diving too deep based on the snippet I'm guessing that this is for all the fields and it was applying the text break all to each colName. For example like if they select timestamp, useriD, statusCode, etc. are all col names. It just timefieldName sticks out because it's the first one we see. Other columns might have a child element that don't let it get wrapped as easily.

To be honest I think this rule could actually be applied to all fields of type `date`. But that shouldn't block this PR.

kavilla (approved):

src/plugins/discover/public/application/components/default_discover_table/table_row.tsx
kavilla: created an issue: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8441

PR #8252, opened 2024-09-19
Fix running recent query button

abbyhu2000:
### Description

Fix the bug of running recent query jump to another language.

## Changelog

- fix: Run recent query button work properly

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/language_service/lib/recent_query.tsx
+++ b/src/plugins/data/public/query/query_string/language_service/lib/recent_query.tsx
@@ -51,7 +51,10 @@ export function RecentQueriesTable({
           icon: 'play',
           type: 'icon',
           onClick: (item: RecentQueryTableItem) => {
-            onClickRecentQuery(recentQueries[item.id].query, recentQueries[item.id].timeRange);
+            onClickRecentQuery(
+              recentQueries.find((recentQuery) => recentQuery.query.query === item.query)?.query!,
+              recentQueries[item.id].timeRange
+            );
           },
           'data-test-subj': 'action-run',
         },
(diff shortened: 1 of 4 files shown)


ashwin-pc (approved):

src/plugins/data/public/query/query_string/language_service/lib/recent_query.tsx
```diff
@@ -51,7 +51,10 @@ export function RecentQueriesTable({
           icon: 'play',
           type: 'icon',
           onClick: (item: RecentQueryTableItem) => {
-            onClickRecentQuery(recentQueries[item.id].query, recentQueries[item.id].timeRange);
+            onClickRecentQuery(
```
kavilla, line 54: to mind explaining this part a little bit? like how this fixes the problem?

without knowing too much about the implementation, for me this basically looks for the query and assumes its the same query which might not be accurate still i think.

i dont think we show this feature for DQL but in the case of DQL the query could actually just be 
`""` for multiple queries for different index patterns. And this logic to me would find the first query that is `""` and assume that is the recent query.

in SQL and PPL we dont limit what they type so technically they could select `data_logs`

and in SQL it would be `SELECT * FROM data_logs`
and then they type it `SELECT * FROM eccomerce` and then select the eccommerce sample data. so then the query editor says `SELECT * FROM eccomerce` which query is the right query here.

i think the solution here should be assigning a GUID or using the timestamp of execution as the identifier

abbyhu2000: make sense. i revised it to use uuid

src/plugins/data/public/query/query_string/language_service/types.ts
```diff
@@ -13,6 +13,7 @@ import {
 import { EditorInstance } from '../../../ui/query_editor/editors';
 
 export interface RecentQueryItem {
+  id: number;
```
kavilla, line 16: Nit: I think it's a string now right ?

kavilla (approved):

LDrago27 (approved):

src/plugins/data/public/query/query_string/language_service/types.ts
abbyhu2000: Will fix in later PR

PR #7925, opened 2024-08-29
Add Data connection saved-object type for external connections

goyamegh:
### Description

Today MDS plugin only supports OpenSearch datasources. With the new saved-object type data-connection, the intention is to allow the external data connections like CloudWatch and Security Lake.

### Issues Resolved
NA

## Screenshot
NA

## Testing the changes
Saved objects API testing sample inputs (unit tests added as well):

```
1. Create:
Input: 
POST http://localhost:5601/api/saved_objects/data-connection/12345
{
  "attributes": {
    "connectionId": "security-lake",
    "type": "AWS CloudWatch"
  }
}
Output: 
{
"type":  "data-connection",
"id":  "1234567",
"attributes":  {
"connectionId":  "security-lake",
"type":  "AWS CloudWatch"
},
"references":  [],
"updated_at":  "2024-08-29T21:38:20.344Z"
}

2. Update
Input:
PUT http://localhost:5601/api/saved_objects/data-connection/12345
{
  "attributes": {
    "connectionId": "security-lake",
    "type": "AWS CloudWatch"
  }
}

3. Bulk create - failure
Input:
POST http://localhost:5601/api/saved_objects/_bulk_create
[
    {
      "type": "data-connection",
      "id": "12345",
      "attributes": {
        "connectionId": "security-lake",
        "type": "AWS Security "
      }
    },
    {
      "type": "data-connection",
      "id": "67890",
      "attributes": {
        "connectionId": "cloudwatch",
        "type": "AWS CloudWatch1"
      }
    }
  ]
Output:
{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "Invalid attribute type: 'AWS Security ' for connection id: 'security-lake': Bad Request"
}

4. Bulk update
PUT http://localhost:5601/api/saved_objects/_bulk_update
[
    {
      "type": "data-connection",
      "id": "12345",
      "attributes": {
        "connectionId": "security-lake",
        "type": "AWS Security "
      }
    },
    {
      "type": "data-connection",
      "id": "67890",
      "attributes": {
        "connectionId": "cloudwatch",
        "type": "AWS CloudWatch1"
      }
    }
  ]
Output: 
{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "Updating a data-connection saved-object type is not supported: Bad Request"
}

3. Bulk create - success
Input:
POST http://localhost:5601/api/saved_objects/_bulk_create
[
    {
      "type": "data-connection",
      "id": "12345",
      "attributes": {
[...]

PR #8255, opened 2024-09-19
[Discover] Support data connections and multi-select table in dataset picker

joshuali925:
### Description

#7925 added data-connection saved object type, this PR sets the framework for using possible external connections in discover dataset selector. This PR does not add any specific connection types, things like CloudWatch logs or security lake can be added later.

Additionally this PR introduces multi-select datasets. If multiSelect is enabled, selecting a data structure will concatenate it's id with `,` with the existing selected dataset.

## Screenshot

cloudwatch implementation is not included in this PR, data is mocked to show the component added. UX is not final
Sep-19-2024 17-01-55

## Changelog
- feat: Support data connections and multi-select table in dataset picker

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/common/datasets/types.ts
```diff
@@ -202,6 +206,8 @@ export interface BaseDataset {
   type: string;
   /** Optional reference to the data source */
   dataSource?: DataSource;
+  /** Optional reference to the data connection */
+  dataConnection?: DataConnection;
```
ashwin-pc, line 210: Why do we need this? cant we use the existing `dataSource` property? Event they type above indicates that they are the same. having 2 properties where only one should be populated at any time is not ideal. Lets use the `dataSource` property until we need a distinction. And if we do, we should still only have one property that can be either type so that we cant accidentally set both somehow or check for only one of the two properties. e.g. while crating the title. right now all of Discover only checks for the dataSource property and it will be really easy to miss the data connections property

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -145,6 +151,8 @@ export class DatasetService {
       parent: dataStructure.parent?.id || '',
       children: dataStructure.children?.map((child) => child.id) || [],
       hasNext: dataStructure.hasNext,
+      nextToken: dataStructure.nextToken,
```
ashwin-pc, line 154: `hasNext` and `nextToken` are a little confusing. can we callit something that more generic that can be used for other pagination api's too like `paginationToken` or `cursor` like graphql?

src/plugins/data/common/datasets/types.ts
kavilla: +1 we shouldn't do this.

i would check out https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/query_enhancements/public/datasets/s3_type.ts#L110 to see how we abstracted it

src/plugins/data_source/opensearch_dashboards.json
```diff
@@ -6,5 +6,5 @@
   "ui": true,
   "requiredPlugins": ["opensearchDashboardsUtils"],
   "optionalPlugins": [],
-  "extraPublicDirs": ["common","common/data_sources", "common/util"]
+  "extraPublicDirs": ["common","common/data_sources", "common/data_connections", "common/util"]
```
kavilla, line 9: is data_connections always available?

then i think this will cause a dependency tree that data plugin is dependent on the connections work

src/plugins/data/public/ui/dataset_selector/_dataset_table.scss
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.datasetTable {
```
kavilla, line 6: is this technically just apart of the explorer?

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
kavilla: +1

src/plugins/data/public/ui/dataset_selector/dataset_table.tsx
```diff
@@ -0,0 +1,99 @@
...
+  explorerDataset: DataStructure | undefined;
+  selectDataStructure: (item: DataStructure | undefined, newPath: DataStructure[]) => Promise<void>;
+  fetchNextDataStructure: (
+    nextPath: DataStructure[],
+    dataType: string,
+    options?: DataStructureFetchOptions
+  ) => Promise<DataStructure>;
+}
+
+export const DataSetTable: React.FC<DatasetTableProps> = (props) => {
```
kavilla, line 26: `Dataset`

src/plugins/data/public/ui/dataset_selector/dataset_table.tsx
```diff
@@ -0,0 +1,99 @@
...
+        selection={{
+          onSelectionChange: (items) => {
+            if (items.length === 0) {
+              props.selectDataStructure(undefined, props.path.slice(0, props.index + 1));
+              return;
+            }
+            if (!items.every((item) => item.type === items[0].type)) {
+              throw new Error('All items must be of the same type');
+            }
+            const newItem: DataStructure = {
```
kavilla, line 71: what if we made multidatasets apart of the dataset service so if another plugin would want to implement a solution instead of using our selector they could.

so the [cacheDataset](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/8d52134734857561ad580d2567c68c01fe3e9fa8/src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts#L65) has the ability to take in an array of data structures. 

for example, the dataset type config for example S3 tables will let the dataset service so we dont need to worry about creating some data structure on the fly but just keeping an array of data structures in the ui portion. then when the user selected the data structures and hit next that is when we construct the dataset properly joining the ids.

this way then we dont have to worry about slicing and joining ID before they are ready. also lets plugin developers use the dataset service without using the data plugin ui

src/plugins/data/public/ui/dataset_selector/dataset_explorer.tsx
```diff
@@ -22,6 +22,8 @@ import { FormattedMessage } from '@osd/i18n/react';
 import { BaseDataset, DATA_STRUCTURE_META_TYPES, DataStructure } from '../../../common';
 import { QueryStringContract } from '../../query';
 import { IDataPluginServices } from '../../types';
+import { DataSetTable } from './dataset_table';
+import { DataStructureFetchOptions } from '../../query/query_string/dataset_service';
```
kavilla, line 26: nit: i would export this so from query_string and then query so that line 23 could be `QueryStringContract, DataStructureFetchOptions` incase other plugins would like to use this feature without the linter complaining about this not being exported from the top level folder of the plugin

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -91,22 +91,28 @@ export class DatasetService {
   public async fetchOptions(
     services: IDataPluginServices,
     path: DataStructure[],
-    dataType: string
+    dataType: string,
+    options?: DataStructureFetchOptions
```
kavilla, line 95: 1 more thing. i think @amsiglan  implemented a `getSearchOptions` perhaps we should introduce a `getFetchOptions`

src/plugins/data/common/datasets/types.ts
joshuali925: good point. i kept them separated because `data-connection` is a different saved object type, there might be distinctions in the future to make them not interchangeable, and calling both types `dataSource` might be confusing. But I agree since they are mutually exclusive and are very similar now, it's better keep both under `dataSource`

src/plugins/data_source/opensearch_dashboards.json
joshuali925: > is data_connections always available

this depends on whether the plugin that depends on data_source plugin calls it required or optional

> this will cause a dependency tree that data plugin is dependent on the connections work

this is not used by the data plugin, i exported it for cloudwatch implementation. it should be exported because `common/data_connections` should behave similar like `common/data_sources`, they are same level concepts

src/plugins/data/public/ui/dataset_selector/_dataset_table.scss
joshuali925: yes. i can combine them

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
joshuali925: i think `fetchOptions` should be defined by the caller of `fetch` to specify whether it wants to fetch next page or with some search keyword, not by the `DatasetTypeConfig`? whereas [`getSearchOptions`](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/1750cadee4d94c62532ddff377c74180c2b19ceb/src/plugins/data/public/query/query_string/dataset_service/types.ts#L57-L61) should be determined by the type config

i feel the `fetch` vs `search` is getting confusing, `fetch` is actually populating the properties like `children` of the given datastructure. i'll add/update jsdocs

src/plugins/data/public/ui/dataset_selector/dataset_table.tsx
joshuali925: do you mean changing `selectedDatasets` to an array in advanced selector https://github.com/opensearch-project/OpenSearch-Dashboards/blob/1750cadee4d94c62532ddff377c74180c2b19ceb/src/plugins/data/public/ui/dataset_selector/advanced_selector.tsx#L51

when user clicks next, configurator receives the list of datasets and uses dataset service to combine them into a single dataset?

joshuali925: actually it would change the `path` definition, might be too much change https://github.com/opensearch-project/OpenSearch-Dashboards/blob/f027944f67cf2c6a6439fb3b6c271920f4d60efc/src/plugins/data/public/ui/dataset_selector/dataset_explorer.tsx#L52-L73
if we defer combining ids until `next`, we need to properly represent `path` to include the multiple selected datasets in `dataset_explorer`. but `path` is `DataStructure[]`, we can't put multiple datasets on the same level. or did i understand your suggestion correctly?

kavilla: i see what you mean. this MVP to get multi select working without being too intrusive of a change. i can stand by that. should we create a backlog item to revisit this to make the dataset service to the heavy lifting of combining data structures

src/plugins/data/public/ui/dataset_selector/dataset_table.tsx
```diff
@@ -0,0 +1,101 @@
...
+            [],
+        }}
+      />
+
+      {paginationToken && (
+        <div className="datasetTable__loadMore">
+          <EuiLink
+            onClick={() => onTableChange({ paginationToken, search: searchRef.current?.value })}
+          >
+            <EuiText size="s">
```
kavilla, line 90: @virajsanghvi @zhongnansu @AMoo-Miki  this is the right size right?

src/plugins/data/public/ui/dataset_selector/dataset_table.tsx
```diff
@@ -0,0 +1,101 @@
...
+        loading={loading}
+        isSelectable
+        selection={{
+          onSelectionChange: (items) => {
+            if (items.length === 0) {
+              props.selectDataStructure(undefined, props.path.slice(0, props.index + 1));
+              return;
+            }
+            if (!items.every((item) => item.type === items[0].type)) {
+              throw new Error('All items must be of the same type');
```
kavilla, line 69: yes!

should we confirm with ux the proper wording? since this is an important detail i feel like people will ask.

could we add a test for this one. i can see one day someone might try to add this feature and not ask why we didn't want this. the test can be the safeguard

src/plugins/data/public/ui/dataset_selector/dataset_table.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const dataStructures = props.path[props.index].children || [];
+  const paginationToken = props.path[props.index].paginationToken;
+
+  const onTableChange = async (options: DataStructureFetchOptions) => {
+    const typeConfig = datasetService.getType(props.path[1].id);
+    if (!typeConfig) return;
+
+    setLoading(true);
+    await props
+      .fetchNextDataStructure(props.path, typeConfig.id, options)
```
kavilla, line 41: not a review but a question. taking in consideration potential amount of data structures that could return and pages. do you think we need to have some limit on this?

like an advanced setting that lets users set max page size default to a value like 10 and then max pages? that way we prevent the ability to just manually scan through 100s of pages for a tables and force the user to use the search bar.

not positive if the data source that is being queried could incur a lot of costs.

joshuali925: the OUI table component is a bit weird that the selections get cleared if items gets unloaded (i.e. when user go to the next page, search by something else, etc). 

So the issue with search alone is that if user wants to select `foo`, `bar`, after searching and selecting `foo`, searching `bar` would clear the selection. i haven't figured out a good UX for this

src/plugins/data/public/ui/dataset_selector/dataset_table.tsx
joshuali925: seems the EuiTable eats the error, it will probably need to be set some state and display error on UI. When we have some proper wording i can add a follow up PR. will create an issue for this

joshuali925: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8362

src/plugins/data/public/ui/dataset_selector/dataset_table.tsx
joshuali925: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8362

kavilla (approved):
should we add s3 as multi select? lgtm.

ashwin-pc (approved):

PR #8268, opened 2024-09-20
[workspace]feat: validate workspace when find objects

SuZhou-Joe:
### Description

We need to validate the workspace ids users pass, if it is not valid, we should just throw a 400 error and save the resource of our search engine.

## Changelog

- feat: validate if the workspaces param is valid or not when consume it in the wrapper.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
+++ b/src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
@@ -85,22 +85,33 @@ export class WorkspaceIdConsumerWrapper {
             ? options
             : this.formatWorkspaceIdParams(wrapperOptions.request, options);
         if (finalOptions.workspaces?.length) {
-          const workspaceList = await this.workspaceClient.list(
-            {
-              request: wrapperOptions.request,
-            },
-            {
-              perPage: 9999,
-            }
-          );
           let isAllTargetWorkspaceExisting = false;
-          if (workspaceList.success) {
-            const workspaceIdsSet = new Set(
-              workspaceList.result.workspaces.map((workspace) => workspace.id)
+          // If only has one workspace, we should use get to optimize performance
+          if (finalOptions.workspaces.length === 1) {
+            const workspaceGet = await this.workspaceClient.get(
+              { request: wrapperOptions.request },
+              finalOptions.workspaces[0]
             );
-            isAllTargetWorkspaceExisting = finalOptions.workspaces.every((targetWorkspace) =>
-              workspaceIdsSet.has(targetWorkspace)
+            if (workspaceGet.success) {
+              isAllTargetWorkspaceExisting = true;
+            }
+          } else {
+            const workspaceList = await this.workspaceClient.list(
+              {
+                request: wrapperOptions.request,
+              },
+              {
+                perPage: 9999,
+              }
             );
+            if (workspaceList.success) {
+              const workspaceIdsSet = new Set(
+                workspaceList.result.workspaces.map((workspace) => workspace.id)
+              );
+              isAllTargetWorkspaceExisting = finalOptions.workspaces.every((targetWorkspace) =>
+                workspaceIdsSet.has(targetWorkspace)
+              );
+            }
           }
 
           if (!isAllTargetWorkspaceExisting) {
--- a/src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
+++ b/src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
@@ -95,12 +95,11 @@ export class WorkspaceIdConsumerWrapper {
           );
           let isAllTargetWorkspaceExisting = false;
           if (workspaceList.success) {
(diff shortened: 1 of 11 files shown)


src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
```diff
@@ -74,15 +76,42 @@ export class WorkspaceIdConsumerWrapper {
...
+              request: wrapperOptions.request,
+            },
+            {
+              perPage: 9999,
+            }
+          );
+          let isAllTargetWorkspaceExisting = false;
+          if (workspaceList.success) {
+            const workspaceIds = workspaceList.result.workspaces.map((workspace) => workspace.id);
+            isAllTargetWorkspaceExisting = finalOptions.workspaces.every((targetWorkspace) =>
```
wanglam, line 99: Do we still support passing `*`? It seems the  `finalOptions.workspaces` could be `*` in the data source assignment case.

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
```diff
@@ -74,15 +76,42 @@ export class WorkspaceIdConsumerWrapper {
...
+            : this.formatWorkspaceIdParams(wrapperOptions.request, options);
+        if (finalOptions.workspaces?.length) {
+          const workspaceList = await this.workspaceClient.list(
+            {
+              request: wrapperOptions.request,
+            },
+            {
+              perPage: 9999,
+            }
+          );
```
Hailong-am, line 95: nit: if there only have one workspace in `finalOptions.workspaces` we can use `this.workspaceClient.get`, may or may not have performance improvement.

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
SuZhou-Joe: Oh, nice catch. Update that and add a unit test accordingly.

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
SuZhou-Joe: Yeah, we should do that.

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
Hailong-am: But `formatWorkspaceIdParams` will remove `*`, that's before the check. do we need to add * here?

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
```diff
@@ -74,15 +76,45 @@ export class WorkspaceIdConsumerWrapper {
...
+          let isAllTargetWorkspaceExisting = false;
+          if (workspaceList.success) {
+            const workspaceIds = [
```
ruanyl, line 98: Nit: this could a `Set`

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
```diff
@@ -74,15 +76,45 @@ export class WorkspaceIdConsumerWrapper {
...
+        // we need to make sure the find call for upgrade config should be able to find all the global configs as it was before.
+        // It is a workaround for 2.17, should be optimized in the upcoming 2.18 release.
+        const finalOptions =
           this.isConfigType(options.type as string) && options.sortField === 'buildNum'
             ? options
-            : this.formatWorkspaceIdParams(wrapperOptions.request, options)
-        );
+            : this.formatWorkspaceIdParams(wrapperOptions.request, options);
+        if (finalOptions.workspaces?.length) {
+          const workspaceList = await this.workspaceClient.list(
```
ruanyl, line 88: Shall we simply call `wrapperOptions.client.find` with `WORKSPACE_TYPE` to get the the workspace list? Just thinking might be better to not introduce another dependency to saved object client wrapper.

SuZhou-Joe: We can not because workspace is a hidden type.

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
SuZhou-Joe: Yeah, `*` does not need to be add here, removed that.

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
SuZhou-Joe: Sure, updated.

ruanyl (approved):

wanglam (approved):

Hailong-am (approved):

PR #8087, opened 2024-09-09
[Auto Suggest] PPL autocomplete fixes

paulstn:
### Description

Fixes various issues for ppl autocomplete.

Note: Can ignore changes in the `.generated` folder

## Changelog

- fix: PPL Autocomplete functions, fields, & table suggestion

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #8275, opened 2024-09-20
[Auto Suggest] PPL & SQL Value Suggestion

paulstn:
### Description

Only done for opensearch SQL + PPL. For both languages, value completion only occurs inside of a binary comparison predicate (for example `column = value`, and for an in-predicate with the keyword `IN` (e.g. `column IN ("value1", "value2", "value3")`). These are the only cases in which it makes sense to suggest values for a particular column.

SQL:

https://github.com/user-attachments/assets/54518f20-f349-4bc0-840e-9a38767b1aba

PPL:

https://github.com/user-attachments/assets/71c41b93-e199-48f8-acf2-7ba5aed16f73

## `src/plugins/data/public/antlr/*`

For PPL & SQL, uses the SQL query 
```
SELECT <column> FROM <table> GROUP BY <column> ORDER BY COUNT(<column>) DESC LIMIT <limit>
```
to find the most popular column values to display within autocomplete whenever a value should be suggested.

## `src/plugins/data/server/ui_settings.ts`

There were two UI settings added.
`query:enhancements:suggestValues` is a boolean that determines if value suggestion will be used
`query:enhancements:suggestValuesLimit` is a number that specifies how many values should be queried, defaulting to 200

## `src/plugins/data/public/autocomplete/autocomplete_service.ts`

The ability to add a value suggestion provided was included in this pr, but isn't in use so far. Creating a value suggestion provider that would make the api call (with all of the same parameters that exist right now) could be done but would require a lot of complexity for what is essentially done in 10 lines today. 

This PR also contains various fixes and quality of life updates within SQL & PPL suggestions.

## Testing the changes

Field level security testing:

https://github.com/user-attachments/assets/77bae369-8a30-42a3-acbc-bbab07184ed3

Field level masking testing:

https://github.com/user-attachments/assets/a7bc7f93-be32-4c72-9ce9-8dd9a15179ff

## Changelog

- feat: Autocomplete Value Suggestion

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -285,6 +285,7 @@ export default class QueryEditorUI extends Component<Props, State> {
       indexPattern,
       position,
       services: this.services,
+      dataset: this.queryString.getQuery().dataset,
```
sejli, line 288: Do we need to pass `dataset` in here if `services` is already being passed in? It's already included in `services` if you use `services.data.query.queryString.getQuery().dataset`.

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -63,7 +65,7 @@ const fetchFromAPI = async (api: any, body: string) => {
 export const fetchData = (
   tables: string[],
   queryFormatter: (table: string, dataSourceId?: string, title?: string) => any,
-  api: any,
+  http: any,
```
sejli, line 68: Define `http: HttpSetup`

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -93,24 +95,27 @@ export const fetchData = (
...
-          queryConfig: {
-            dataSourceId: dataSourceId || undefined,
-            title: title || undefined,
-          },
-        },
+export const fetchColumnValues = (
+  tables: string[],
+  column: string,
+  services: IDataPluginServices,
+  dataset?: Dataset
```
sejli, line 102: Why is `dataset` optional here? Would it still work without it?

src/plugins/data/server/ui_settings.ts
sejli, line 1: Do these settings need to have the page reloaded? @kavilla

src/plugins/data/public/ui/query_editor/query_editor.tsx
paulstn: i guess a lot of the parameters can be removed if we can get querystring from services, let me see about that

src/plugins/data/public/antlr/shared/utils.ts
paulstn: The original reason for this is the dataset would be used solely for mds, if there is another cluster then it will pass in the dataset to use the same data source that is already engaged. Also, the dataset being passed into the getSuggestion functions don't require dataset, so its type is `Dataset | undefined` before calling this fetchColumnValues helper function. 
Should I change this to have dataset be required?

src/plugins/data/server/ui_settings.ts
paulstn: btw I believe it does, when I tried changing the settings without reloading it didn't have an effect.

src/plugins/data/public/antlr/opensearch_ppl/grammar/OpenSearchPPLParser.g4
```diff
@@ -177,7 +177,7 @@ tableSourceClause
    ;
 
 renameClasue
-   : orignalField = wcFieldExpression AS renamedField = wcFieldExpression
+   : orignalField = fieldExpression AS renamedField = ID
```
mengweieric, line 180: Seems like we've changed a lot in grammar files which will make these differ a lot from original source from sql plugins. This is fine but did we do thorough tests to make sure it parses the query correctly? Do we also have plan to maintain overall grammar functionally to make it always align with grammar in sql plugin?

src/plugins/data/public/antlr/opensearch_ppl/grammar/OpenSearchPPLParser.g4
```diff
@@ -696,11 +696,11 @@ stringLiteral
    ;
 
 integerLiteral
-   : (PLUS | MINUS)? INTEGER_LITERAL
+   : INTEGER_LITERAL
```
mengweieric, line 699: if we don't suggest plus or minus, what would be the issue keeping them?

src/plugins/data/public/antlr/opensearch_sql/code_completion.ts
```diff
@@ -44,7 +44,33 @@ export const getSuggestions = async ({
...
+        dataset
+      );
+
+      let i = 0;
+      finalSuggestions.push(
+        ...res.body.fields[0].values.map((val: any) => {
+          i++;
+          return {
+            text: val.toString(),
+            insertText: typeof val === 'string' ? `"${val}" ` : `${val} `,
```
mengweieric, line 67: have we tested number value?

src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
```diff
@@ -40,16 +40,50 @@ export const getSuggestions = async ({
...
+        dataset
+      );
+
+      let i = 0;
+      finalSuggestions.push(
+        ...res.body.fields[0].values.map((val: any) => {
+          i++;
+          return {
+            text: val.toString(),
+            insertText: typeof val === 'string' ? `"${val}" ` : `${val} `,
```
mengweieric, line 61: do we need to consider use case like when user doesn't have permission to see values for certain field? like if the values for certain sensitive fields are masked.

src/plugins/data/public/antlr/opensearch_ppl/grammar/OpenSearchPPLParser.g4
paulstn: I think some of the changes to the grammar need to happen in order for it to fit for autocompletion. Some of the issues that would appear for example would be that tables would be candidates twice since the lexer rule for the table name allows for a space, and there is no good way of getting around that other than restricting the parser grammar to only consider a single table as being qualified. As for having it always align with the grammar in the sql plugin, I'm not sure how it would be done without having some way of having some way of keeping tabs on changes to the sql plugin's parsing grammar and making those changes to the autocomplete in dql. Do we expect there to be many changes to the sql grammar? Do you think we should keep these grammars exactly the same to make sure that changes to the sql grammar can be suggested?

src/plugins/data/public/antlr/opensearch_ppl/grammar/OpenSearchPPLParser.g4
paulstn: I've removed plus and minus from the ignore token list, since I found that commands such as `fields` need them.

src/plugins/data/public/antlr/opensearch_sql/code_completion.ts
paulstn: yeah so i have the logic like this since if the type isn't of type string, such as number, it should be unwrapped of quotes. i haven't seen any other data types that can be aggregated other than string or number, do you know if there are any others?

src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
```diff
@@ -40,16 +41,47 @@ export const getSuggestions = async ({
...
+      const res = await fetchColumnValues(
+        [indexPattern.title],
+        suggestions.suggestValuesForColumn,
+        services,
+        dataset
+      );
+
+      let i = 0;
+      finalSuggestions.push(
+        ...res.body.fields[0].values.map((val: any) => {
```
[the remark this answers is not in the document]
paulstn: changing this to be either of type number or string

src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
paulstn: testing field masking

src/plugins/data/public/antlr/opensearch_sql/grammar/OpenSearchSQLParser.g4
```diff
@@ -291,7 +291,7 @@ expression
 
 predicate
    : expressionAtom                                         # expressionAtomPredicate
-   | left = predicate comparisonOperator right = predicate  # binaryComparisonPredicate
+   | left = predicate comparisonOperator right = constant   # binaryComparisonPredicate
```
joshuali925, line 294: why do we need this change?

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -93,24 +96,27 @@ export const fetchData = (
...
+) => {
+  if (!services.uiSettings.get(UI_SETTINGS.QUERY_ENHANCEMENTS_SUGGEST_VALUES)) {
+    return new Promise(() => []);
```
joshuali925, line 106: should this guard happen before calling `fetchColumnValues`? so this function is not coupled with the setting

src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
```diff
@@ -40,16 +40,50 @@ export const getSuggestions = async ({
...
+      // get dataset for connecting to the cluster currently engaged
+      const dataset = services.data.query.queryString.getQuery().dataset;
+
+      // take the column and push in values for that column
+      const res = await fetchColumnValues(
+        [indexPattern.title],
+        suggestions.suggestValuesForColumn,
+        services,
+        dataset
+      );
```
joshuali925, line 53: how does it handle errors?

src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
```diff
@@ -40,16 +40,50 @@ export const getSuggestions = async ({
...
+      let i = 0;
+      finalSuggestions.push(
+        ...res.body.fields[0].values.map((val: any) => {
+          i++;
+          return {
+            text: val.toString(),
+            insertText: typeof val === 'string' ? `"${val}" ` : `${val} `,
+            type: monaco.languages.CompletionItemKind.Value,
+            detail: SuggestionItemDetailsTags.Value,
+            sortText: i.toString().padStart(3, '0'),
```
joshuali925, line 64: seems it's sorting by insertion order, is sortText with the 0 padding required? why not sort alphabetically?

src/plugins/data/public/autocomplete/providers/query_suggestion_provider.ts
```diff
@@ -31,6 +31,7 @@
 import { monaco } from '@osd/monaco';
 import { IFieldType, IndexPattern } from '../../../common/index_patterns';
 import { IDataPluginServices } from '../../types';
+import { Dataset } from '../../../common';
```
joshuali925, line 34: unused?

src/plugins/data/public/autocomplete/providers/value_suggestion_provider.ts
```diff
@@ -39,7 +39,16 @@ function resolver(title: string, field: IFieldType, query: string, boolFilter: a
   return [ttl, query, title, field.name, JSON.stringify(boolFilter)].join('|');
 }
 
-export type ValueSuggestionsGetFn = (args: ValueSuggestionsGetFnArgs) => Promise<any[]>;
+export type ValueSuggestionsGetFn = (
+  args: ValueSuggestionsGetFnArgs | ValueSuggestionsSQLGetFnArgs
+) => Promise<any[]>;
```
joshuali925, line 44: np. might be better to put args type and return type as generic, to allow extending and keeps autocomplete generic to what languages are used

joshuali925 (commented):
seems it's simplifying some of the comparison grammars, any context on why they are needed?

src/plugins/data/public/antlr/opensearch_ppl/grammar/OpenSearchPPLParser.g4
mengweieric: We should detect context as much from code as possible like modify visitors, changing the grammar directly would risk changing the original grammar sometimes, and complexities to sync new grammar

src/plugins/data/public/antlr/shared/utils.ts
paulstn: I think it might be better kept this way since fetchColumnValues is fully expected by its callers to return nothing if there's any reason why columns should not be returned, for example if the field type isn't supported.

src/plugins/data/public/antlr/opensearch_sql/grammar/OpenSearchSQLParser.g4
paulstn: This line, along with line 307/308, was changed so that values don't get suggested for the left side of a predicate, while the right side doesn't have field names suggested and only values suggested instead. Although having a constant on the left and a field on the right does get parsed, suggesting it is practically useless/ information overload. There would be too many suggestions and it doesn't make sense to show every single possibility.

src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
paulstn: what it would do here is just default to no suggestions, since at this point if res is a rejection from the api, it will be caught by lines 120-122. since the code completion method is called for any partially formed query, it should be fine to not show any suggestions if fetchColumnValues doesn't return anything.

src/plugins/data/public/autocomplete/providers/value_suggestion_provider.ts
paulstn: I'm not sure I fully understand what making the args or the return type generic would accomplish. Since this file is meant to include the implementation for functions that would do the value completion, they require specific inputs that may not be related to the output. 
To keep autocomplete generic to the languages used, should I just change the name of the function to move away from "SQL"?

mengweieric:
Just summarize some of the concerns for this PR that I feel we need to address:

**Grammar Changes**:
Modifying the grammar directly is not considered an optimal approach, as it introduces risks of mismatches with removed or modified grammar rules. We should prioritize leveraging visitors, token streams, and logical code adjustments to achieve the intended behavior.

**Test Coverage**:
There is insufficient test coverage to ensure the modified grammar behaves consistently with previous expectations. Additional tests should be added to confirm the correctness of the changes if we want to go ahead with grammar changes.

**Value Suggestions**:
While value suggestions are not a typical feature for SQL/PPL languages, it is acceptable in this context. However, the current implementation still has gaps in providing a meaningful suggestion experience, such as suggesting masked values.

**Unaddressed Comments**:
Some comments have not been addressed yet. Please ensure that all feedback is reviewed and handled accordingly.

src/plugins/data/public/antlr/opensearch_ppl/grammar/OpenSearchPPLParser.g4
```diff
@@ -263,9 +263,9 @@ comparisonExpression
    ;
 
 valueExpression
-   // : left = valueExpression binaryOperator = (STAR | DIVIDE | MODULE) right = valueExpression   # binaryArithmetic
-   // | left = valueExpression binaryOperator = (PLUS | MINUS) right = valueExpression             # binaryArithmetic
-   : primaryExpression                                                                          # valueExpressionDefault
+   : left = valueExpression binaryOperator = (STAR | DIVIDE | MODULE) right = valueExpression   # binaryArithmetic
```
paulstn, line 266: NOTE: this is reverting back a change made in a previous pr, to go back to the original grammar.

src/plugins/data/public/antlr/opensearch_sql/code_completion.test.ts
```diff
@@ -0,0 +1,149 @@
...
+
+      [',', ')'].forEach((term) => {
+        checkSuggestionsContain(result, {
+          text: term,
+          type: monaco.languages.CompletionItemKind.Keyword,
+        });
+      });
+    });
+
+    it.skip('should suggest aggregate functions when appropriate', async () => {
```
paulstn, line 123: NOTE: This test is skipped for now due to the working functionality requiring a new 'rerun and combine' functionality, where preferred rules are removed one at a time and then the completion candidate process is run without that rule. 
This new system is already written in this PR: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9120 
Decoupling it from the rest of that PR would be too much unnecessary effort.

src/plugins/data/public/antlr/opensearch_sql/code_completion.ts
```diff
@@ -59,17 +133,32 @@ export const getSuggestions = async ({
...
+        text: indexPattern.title,
+        type: monaco.languages.CompletionItemKind.Struct,
+        insertText: `${indexPattern.title} `,
+        detail: SuggestionItemDetailsTags.Table,
+      });
+    }
+
+    const suggestionImportance = new Map<number, string>();
+    suggestionImportance.set(OpenSearchSQLParser.STAR, '1');
+    suggestionImportance.set(OpenSearchSQLParser.IN, '09');
```
mengweieric, line 147: Do we want to use constants for all these text sorting values?

mengweieric (commented):
Still need more time to test it in app.

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -93,41 +96,82 @@ export const fetchData = (
...
+          dataset,
         },
-      },
-    }),
-    api,
-    queryString
-  );
+      })
+    )
+  ).body.fields[0].values;
```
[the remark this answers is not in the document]
paulstn: In the case when `fetchFromAPI` rejects the promise with the error, the rejected promise will travel up to the caller of `fetchColumnValues`. I wrote a test for this mocking an error from the http fetch function that `fetchFromAPI` calls:
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8275/files/2a167e91cc27b7be0091cd010495ed87ef9cd0e7#diff-b547688af7b19722a725a97405371301f010a16e1d6dd8dfbfdae74b0e3c784eR374-R382 

In the case where the fetched object doesn't match this caller chain, it will throw a rejected promise about the format. I wrote it this way because currently the returned object has this exact format, and there is no other standard to flatted this object to that I'm aware of.

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -93,41 +96,82 @@ export const fetchData = (
...
-          queryConfig: {
-            dataSourceId: dataSourceId || undefined,
-            title: title || undefined,
-          },
+export const fetchColumnValues = async (
+  table: string,
+  column: string,
+  services: IDataPluginServices,
+  fieldInOsd: IndexPatternField | undefined
+): Promise<any[]> => {
```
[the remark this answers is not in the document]
paulstn: I originally wrote this to return a string array, but I changed it to `any` because if the value suggestions change in the future and return more types than just string, this method would be able to handle those cases too. 
Should I change it back to `string[]`?

src/plugins/data/public/antlr/opensearch_ppl/opensearch_ppl_autocomplete.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  it('should correctly generate ignored tokens', () => {
+    const ignoredTokens = new Set(openSearchPplAutocompleteData.ignoredTokens);
+    expect(ignoredTokens.has(OpenSearchPPLParser.CASE)).toBe(true);
+    expect(ignoredTokens.has(OpenSearchPPLParser.CAST)).toBe(true);
+  });
+});
+
+describe('processVisitedRules', () => {
+  //   // createTokenStream takes a list of 'tokens' defined only by its type or an actual object to be returned as the Token
```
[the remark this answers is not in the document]
paulstn: I left this in as an accident, I was going to write these tests but I was able to cover these tests with end to end tests in `code_completion.test.ts`. 
I'll put some basic processVisitedRules tests in just as a placeholder if more testing is needed for PPL's visitedRules

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -263,7 +263,7 @@ export const QueryEditorUI: React.FC<Props> = (props) => {
                   range: s.replacePosition ?? defaultRange,
                   detail: s.detail,
                   command: { id: 'editor.action.triggerSuggest', title: 'Trigger Next Suggestion' },
-                  sortText: s.sortText, // when undefined, the falsy value will default to the label
+                  sortText: s.sortText ?? s.text, // when undefined, the falsy value will default to the label
```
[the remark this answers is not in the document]
paulstn: s.text can't be undefined or null because its type (`MonacoCompatibleQuerySuggestion`) has `text` as a required field

src/plugins/data/public/autocomplete/autocomplete_service.ts
```diff
@@ -65,16 +66,37 @@ export class AutocompleteService {
...
+  private getValueSuggestions: ValueSuggestionsGetFn = (args) => {
+    if ('language' in args) {
+      const { language } = args;
+      const provider = this.valueSuggestionProviders.get(language);
+
+      if (provider) {
+        return provider(args);
+      }
+    }
+    return this.defaultGetValueSuggestions
```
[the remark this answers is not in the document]
paulstn: the default value provider exists for the discover filter bar, i've set the default like that because i'm aiming to make it so that it will still exist for every case other than if the language is sql/ppl

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -93,41 +96,82 @@ export const fetchData = (
...
-    }),
-    api,
-    queryString
-  );
+      })
+    )
+  ).body.fields[0].values;
+};
+
+export const formatValuesToSuggestions = (
```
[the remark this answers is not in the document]
paulstn: changed

src/plugins/data/public/antlr/opensearch_sql/code_completion.ts
paulstn: moved all of these to constants folder

src/plugins/data/public/autocomplete/providers/value_suggestion_provider.ts
paulstn: Thanks for the clarification, I misunderstood Josh the first time. Looking at this section I put it in a while ago but now I realize the value suggestion provider is not really being used, so I'm going to remove all of these changes and just leave the default suggestion provider as is.

src/plugins/data/public/autocomplete/providers/value_suggestion_provider.ts
```diff
@@ -51,7 +60,9 @@ interface ValueSuggestionsGetFnArgs {
 
 export const getEmptyValueSuggestions = (() => Promise.resolve([])) as ValueSuggestionsGetFn;
 
-export const setupValueSuggestionProvider = (core: CoreSetup): ValueSuggestionsGetFn => {
+export const setupValueSuggestionProvider = (
```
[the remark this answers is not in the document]
paulstn: I've removed these changes entirely, see other comment

src/plugins/data/public/antlr/opensearch_ppl/grammar/OpenSearchPPLParser.g4
paulstn: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8087/files#diff-3dc92487ddb63ebd741b8639f39b9440158a1f31317978cbd84af144daf3adc0L266-R268

src/plugins/data/public/antlr/shared/utils.ts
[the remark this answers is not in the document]
mengweieric: I think we should be able to follow the similar above: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8275#discussion_r1896967901, or if string is the only type supported as of right now we can stick with string and extend it later when we want dynamic typing.

mengweieric (approved):

joshuali925 (approved):

src/plugins/data/public/antlr/shared/utils.ts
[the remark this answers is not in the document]
paulstn: Changed it to this in an earlier commit, agree with extending it later since currently only string is supported

PR #8271, opened 2024-09-20
[discover] cached data structures TTL

kavilla:
### Description

PR builds off of this PR: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8226/files

If the dataset type config passes a meta object with `updateAt`
within the data structure we know that the config opts into the TTL.

If TTL over configuring UI setting then refetch in the case of S3
to get a new session ID in the cache.

Signed-off-by: Kawika Avilla <kavilla414@gmail.com>

### Issues Partially Resolved

n/a

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #8304, opened 2024-09-23
  [WIP][discover][s3] fetch options with fields and caches it

kavilla:
### Description

NOTE: Contains commit from https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8271. For this PR just look at: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8304/commits/a9fa1dfad809cd7522ebd150137dda983bd1d1f6

NOTE: Currently does not work. Could it be the wrong query being sent?

Follow-up item should be what happens in the case where fetch fields fails? Or takes too long? Should we put a spinner on the configurator to not display until fetch fields has return a result (regardless of success or fail). Should we disable the `Select data` if fields have not returned a result?

With sample it did send
    `DESCRIBE default.http_logs_partitioned`
    
But failed with: Total size of serialized results of 6 task is bigger than maxResult size.

Otherwise it should be calling the same helper functions that cache the data structure.

### Issues Resolved

n/a

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/public/datasets/s3_type.ts
+++ b/src/plugins/query_enhancements/public/datasets/s3_type.ts
@@ -146,6 +146,7 @@ const setMeta = (dataStructure: DataStructure, response: any) => {
     ...dataStructure.meta,
     queryId: response.queryId,
     sessionId: response.sessionId,
+    updatedAt: Date.now(),
   } as DataStructureCustomMeta;
 };
 
--- a/src/plugins/query_enhancements/public/datasets/s3_type.ts
+++ b/src/plugins/query_enhancements/public/datasets/s3_type.ts
@@ -82,6 +82,14 @@ export const s3TypeConfig: DatasetTypeConfig = {
           children: tables,
         };
       }
+      case 'TABLE': {
+        const fields = await fetchFields(http, path);
+        return {
+          ...dataStructure,
+          hasNext: false,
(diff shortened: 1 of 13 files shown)


src/plugins/query_enhancements/public/datasets/s3_type.ts
```diff
@@ -246,3 +294,30 @@ const fetchTables = async (http: HttpSetup, path: DataStructure[]): Promise<Data
...
+  const connection = path.find((ds) => ds.type === 'CONNECTION');
+  const database = path.find((ds) => ds.type === 'DATABASE');
+  const sessionId = (connection?.meta as DataStructureCustomMeta).sessionId;
+  const table = path[path.length - 1];
+  const response = await http.fetch({
+    method: 'POST',
+    path: trimEnd(`${API.DATA_SOURCE.ASYNC_JOBS}`),
+    body: JSON.stringify({
+      lang: 'sql',
+      query: `DESCRIBE ${database?.title}.${table.title}`,
```
ps48, line 310: Should be `DESCRIBE ${datasource?.name}.${database?.title}.${table.title}`

kavilla:
closing in favor of: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8337/files

PR #8325, opened 2024-09-24
Filter Saved Query by supported language

LDrago27:
### Description

This PR aims at filtering Saved Queries on the basis of language. 

## Screenshot

In Discover that supports additional SQL and PPL languages
image

In Visualize page where we only support DQL, other language Saved Queries are filtered
 
image

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/public/query/saved_query/saved_query_service.ts
```diff
@@ -117,12 +121,29 @@ export const createSavedQueryService = (
...
+    const currentAppId =
+      (await application?.currentAppId$?.pipe(first()).toPromise()) ?? Promise.resolve(undefined);
+    const languageService = queryStringManager?.getLanguageService();
+
+    // Filtering saved queries based on language supported by cirrent application
+    if (currentAppId && languageService) {
+      queries = queries.filter((query) => {
+        const languageId = query.attributes.query.language;
+        return (
+          languageService?.getLanguage(languageId)?.supportedAppNames?.includes(currentAppId) ??
```
sejli, line 138: nit: Will `languageService` ever not have `getLanguage()`?

sejli (approved):
Nice changes!  Just some non-blocking comments, otherwise LGTM

src/plugins/data/public/query/saved_query/saved_query_service.ts
```diff
@@ -41,7 +43,9 @@ type SerializedSavedQueryAttributes = SavedObjectAttributes &
   };
 
 export const createSavedQueryService = (
```
kavilla, line 45: we should add a back log item to extend this service. i think at one point we were going to have a toggle to save the query with or without the dataset. which id imagine that is something we would want since in the past the saved query wouldn't replace the selected dataset (index pattern) it would just replace the current query

kavilla (approved):

kavilla:
lgtm its for 2.18 correct?

PR #8337, opened 2024-09-25
Discover support for S3 fields

ps48:
### Description

Discover 2.0 support for S3 fields
This PR doesn't add fields to cache. 

### Issues Resolved

Screenshot:
Screenshot 2024-09-26 at 8 59 26 PM

Demo Video:

https://github.com/user-attachments/assets/04af2e2b-b531-4617-8c5d-06f761ec0964

## Changelog

- fix: Discover 2.0 support for S3 fields

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/common/osd_field_types/types.ts
```diff
@@ -93,3 +93,27 @@ export enum OSD_FIELD_TYPES {
   NESTED = 'nested',
   HISTOGRAM = 'histogram',
 }
+
+/** @public **/
+export enum OPENSEARCH_SQL_TYPES {
```
kavilla, line 98: nit: for consistency `OPENSEARCH_SQL_FIELD_TYPES`

src/plugins/data/common/osd_field_types/osd_field_types.ts
```diff
@@ -84,3 +84,43 @@ export const setOsdFieldOverrides = (newOverrides: { [key: string]: any } | unde
...
   return osdFieldOverrides;
 };
+
+/**
+ *  Mapping function from SQL_TYPES to OSD_FIELD_TYPES
+ *
+ *  @param {OPENSEARCH_SQL_TYPES} sqlType
+ *  @return {OSD_FIELD_TYPES}
+ */
+export function castSQLTypeToOSDFieldType(sqlType: OPENSEARCH_SQL_TYPES): OSD_FIELD_TYPES {
```
kavilla, line 94: this is one way did we want to consider making [OsdFieldType](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/common/osd_field_types/osd_field_type.ts#L37) have an optional property like
```typescript
public readonly sqlTypes: readonly OPENSEARCH_SQL_FIELD_TYPES[];
```

src/plugins/data/common/osd_field_types/osd_field_types.ts
```diff
@@ -84,3 +84,43 @@ export const setOsdFieldOverrides = (newOverrides: { [key: string]: any } | unde
...
+  switch (sqlType) {
+    case OPENSEARCH_SQL_TYPES.BOOLEAN:
```
kavilla, line 96: Nice! i think we should add a section about this information in the reference area: https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/src/plugins/query_enhancements/public/query_editor/ppl_docs for sql or update these docs here: https://opensearch.org/docs/latest/search-plugins/sql/datatypes/

since there are functionality implications. thank you so much for handling this!

src/plugins/query_enhancements/public/datasets/s3_type.ts
```diff
@@ -246,3 +278,68 @@ const fetchTables = async (http: HttpSetup, path: DataStructure[]): Promise<Data
...
+    const database = path.find((ds) => ds.type === 'DATABASE');
+    const sessionId = (connection?.meta as DataStructureCustomMeta).sessionId;
+    const table = path[path.length - 1];
+
+    const response = await http.fetch({
+      method: 'POST',
+      path: trimEnd(`${API.DATA_SOURCE.ASYNC_JOBS}`),
+      body: JSON.stringify({
+        lang: 'sql',
+        query: `DESCRIBE TABLE ${connection?.title}.${database?.title}.${table.title}`,
```
kavilla, line 316: nice!

kavilla (commented):
like it so far! some comments

src/plugins/data/public/query/query_string/dataset_service/types.ts
```diff
@@ -38,7 +38,7 @@ export interface DatasetTypeConfig {
    * Fetches fields for the dataset.
    * @returns {Promise<DatasetField[]>} A promise that resolves to an array of DatasetFields.
    */
-  fetchFields: (dataset: Dataset) => Promise<DatasetField[]>;
+  fetchFields: (services: IDataPluginServices, dataset: Dataset) => Promise<DatasetField[]>;
```
amsiglan, line 41: This is a breaking change and plugins like NeoDashboards with fail at runtime if this gets deployed without the required changes in that plugin.

Can we introduce a different method and deprecate this? Though I know this is behind a flag...

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
```diff
@@ -72,7 +73,7 @@ export const indexTypeConfig: DatasetTypeConfig = {
     }
   },
 
-  fetchFields: async (dataset) => {
+  fetchFields: async (_, dataset: Dataset): Promise<DatasetField[]> => {
```
amsiglan, line 76: NIt: Let's add the full name along with the _ for code readability

src/plugins/data/common/osd_field_types/types.ts
ps48: updated

src/plugins/data/public/query/query_string/dataset_service/types.ts
ps48: > Can we introduce a different method and deprecate this? Though I know this is behind a flag...

IMO Adding a new function which does the same would lead to more confusion here. Things are changing everyday on this code piece so not sure how can we keep track of how downstream is using this feature.

src/plugins/data/common/osd_field_types/osd_field_types.ts
ps48: I'm talking to SQL maintainers on this. The documentation from SQL is not updated when it comes to return types, there are some discrepancies:  https://github.com/opensearch-project/sql/issues/1296

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
ps48: Sure, added this in update commit [ee287e8](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8337/commits/ee287e8f87a614516ee49c2f7946145f3ef91092)

src/plugins/data/common/osd_field_types/osd_field_types.ts
ps48: Updated the implementation for consistency. Now it is similar to the esTypes  [ee287e8](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8337/commits/ee287e8f87a614516ee49c2f7946145f3ef91092)

src/plugins/query_enhancements/common/constants.ts
```diff
@@ -54,3 +54,6 @@ export const UI_SETTINGS = {
 };
 
 export const ERROR_DETAILS = { GUARDRAILS_TRIGGERED: 'guardrails triggered' };
+
+export const S3_PARTITION_FIELD = '# Partition Information';
```
kavilla, line 58: nit:

will this be unique specifically for S3 data types. if not should we just have a constant for fields some thing like

```typescript
FIELD_TYPES = {
   PARTITION: '# Partition Information',
   COMMENT: 'comment',
```

if it is specific to s3 should we modify the constant on line 11 to be something like

```typescript
export const DATASET = {
  S3: {
     id: 'S3',
     FIELD_TYPES: {
        PARTITION: '# Partition Information',
        COMMENT: 'comment',
     }
  }
};
```

would make the delta smaller for importing these constants

src/plugins/query_enhancements/public/datasets/s3_type.ts
```diff
@@ -94,8 +102,9 @@ export const s3TypeConfig: DatasetTypeConfig = {
     }
   },
 
-  fetchFields: async (dataset: Dataset): Promise<DatasetField[]> => {
-    return [];
+  fetchFields: async (services: IDataPluginServices, dataset: Dataset): Promise<DatasetField[]> => {
+    const { http } = services;
+    return fetchFields(http, dataset);
```
kavilla, line 107: do we need an await here?

kavilla (commented):
couple comments mostly nits, one about the type check in the data plugin

kavilla:
do you mind updating the s3 tests for the expected response?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -53,17 +57,20 @@ export const Configurator = ({
 
   useEffect(() => {
     const fetchFields = async () => {
-      const datasetFields = await queryString
-        .getDatasetService()
-        .getType(baseDataset.type)
-        ?.fetchFields(baseDataset);
+      const datasetFields =
+        baseDataset.type !== 'S3'
```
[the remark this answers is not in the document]
ps48: I see, both PPL and SQL support aggregations. Not sure what's the product requirement here do we need time filter for S3 datasource? If not I remember @sejli  mentioned that we'll remove this all together for S3 datasources. If required, we can remove this line when we add in cache support as the subsequent call would just take it from cache.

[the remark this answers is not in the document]
sejli: I do have a PR that should remove the time field selector if the language is SQL If this is to fetch fields on creation of the dataset, I don't think my change is related. 

Currently, we don't need time filter for S3 datasource because SQL doesn't support time filters.

src/plugins/data/public/query/query_string/dataset_service/types.ts
amsiglan: Agreed we need to be flexible in expecting changes to the signature of experimental methods, so let's make services the second parameter to avoid runtime issues

ps48: updated - made services optional and second param.

src/plugins/query_enhancements/common/constants.ts
ps48: Updated [63b7d40](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8337/commits/63b7d40f8c89461c834da09a71c53bf848ab8d61) to mention them as excluded fields as they might be used by other flint based datatypes like security lake.

src/plugins/query_enhancements/public/datasets/s3_type.ts
```diff
@@ -246,3 +255,62 @@ const fetchTables = async (http: HttpSetup, path: DataStructure[]): Promise<Data
 
   return fetch(http, path, 'TABLE');
 };
+
+// Function to process the input and map types using the new SQL to OSD mapping
+export function mapResponseToFields(sqlOutput: {
+  status: string;
```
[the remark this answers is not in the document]
ps48: Yes this would be common for PPL/SQL and async queries. Converted to a type [here](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8337/files#diff-e3aee9041606e7aa03e0c86eb518a3a15d1ce3a5d2af35f1064be36b417270ffR38)

ps48:
> do you mind updating the s3 tests for the expected response?

Updated tests for see actual field response

src/plugins/data/public/ui/dataset_selector/configurator.tsx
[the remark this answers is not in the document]
ps48: Updated the Datasets and DataTypeConfig to use a meta tag `requireTimefilter` This is now used to see, if we need to fetch fields for time filter or not.

src/plugins/data/common/datasets/types.ts
```diff
@@ -28,6 +28,8 @@ export interface DataSourceMeta {
   name?: string;
   /** Optional session ID for faster responses when utilizing async query sources */
   sessionId?: string;
+  /** Optional requiresTimeFilter determines if a time filter is needed */
+  requiresTimeFilter?: boolean;
```
sejli, line 32: `supportsTimeFilter` might be a better name here, since it's possible that we choose no time field even if there are time fields available. One of the options in the time field drop down (if there are time fields) is `I don't want to use the time filter`.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -161,14 +169,18 @@ export const Configurator = ({
...
             onConfirm(dataset);
           }}
           fill
+          isLoading={isLoading}
+          disabled={isLoading}
         >
           <FormattedMessage
             id="data.explorer.datasetSelector.advancedSelector.confirm"
-            defaultMessage="Select Data"
+            defaultMessage={isLoading ? 'Loading' : 'Select Data'}
```
sejli, line 183: Might need to check with UX on the wording for this, because there would be a loading for the submit button to be ready and then another loading state when we actually run the query.

src/plugins/data/public/query/query_string/dataset_service/types.ts
```diff
@@ -53,7 +55,7 @@ export interface DatasetTypeConfig {
    * Fetches fields for the dataset.
    * @returns {Promise<DatasetField[]>} A promise that resolves to an array of DatasetFields.
    */
-  fetchFields: (dataset: Dataset) => Promise<DatasetField[]>;
+  fetchFields: (dataset: Dataset, services?: IDataPluginServices) => Promise<DatasetField[]>;
```
sejli, line 58: Do we need the entire services object? We're only using `http` for fetching S3 fields, and the other types don't require `services`. Unless, the future data types require using `services` instead of just `http`.

src/plugins/query_enhancements/public/datasets/s3_type.ts
```diff
@@ -246,3 +263,63 @@ const fetchTables = async (http: HttpSetup, path: DataStructure[]): Promise<Data
...
+  for (const row of sqlOutput.datarows) {
+    const [colName, dataType] = row;
+
+    // Stop processing once we hit the partition info row
+    if (colName === S3_PARTITION_INFO_COLUMN) {
+      break;
+    }
+
+    // Only include rows with valid types
+    if (dataType && dataType !== '') {
```
sejli, line 280:
```suggestion
    if (dataType) {
```
Is `dataType` always a string?

sejli (approved):
Some general non-blocking questions, but looks good so far

ashwin-pc:
The behaviour im seeing differs from the PR video a lil. 

1. The spinner after i select the data says "Loading" instead of "Fetching fields". I feel like loading is less useful than "Fetch fields" or "Loading Schema" or something more descriptive than "Loading"
2. While i dont like that the loading is happening before the modal closes. It not a bad solution. Did you look to see if you could close the modal after the click and show the loading state elsewhewhere? Like the . That being said, if i close the menu, its still firing the polling requests in the background
3. After selecting the flint dataset and running a query successfully, when i refresh the page, the available fields are empty
Screenshot 2024-10-04 at 5 36 47 AM. This is likely because you only fetch for fields in the data selector. I think fetch fields should happen twice. Once in the dataset selector (if the dataset type supports time filter) and once after the dataset is selected but the index pattern is not cached yet. During this time the data seletor modal can be closed and the available fields can be replaced with a spinner like in the mocks. Not 
4. Can you just pull in Rocky's cache changes into your PR? He's going to be out until the end of next week so it will be quicker to just incorporate his changes in your PR.

ps48:
Hi @ashwin-pc  & @sejli, thanks for the reviews. Replying here cause looks like each maintainer has their own perspective of the PR, which I like. 
> The spinner after i select the data says "Loading" instead of "Fetching fields". I feel like loading is less useful than "Fetch fields" or "Loading Schema" or something more descriptive than "Loading"

This was initially Loading fields (which is what is used in datasource management plugin). But was called out by Rocky to just keep as https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8337#discussion_r1779023524. I like Sean's comment to get this from UX rather we designing or doing copy changes.

> While i dont like that the loading is happening before the modal closes. It not a bad solution. Did you look to see if you could close the modal after the click and show the loading state elsewhewhere? Like the . That being said, if i close the menu, its still firing the polling requests in the background

While I agree to this, the current design for index patterns and indices is to load fields within the data selector.  Do we want to change this for all types of datasets?

>  After selecting the flint dataset and running a query successfully, when i refresh the page, the available fields are empty
This is likely because you only fetch for fields in the data selector. I think fetch fields should happen twice. Once in the dataset selector (if the dataset type supports time filter) and once after the dataset is selected but the index pattern is not cached yet. During this time the data seletor modal can be closed and the available fields can be replaced with a spinner like in the mocks

How this works today is in the data-selector the fields are called twice. 1) for the time filter 2) to load the available fields. I removed the fields call for the time filter with a flag in dataTypeConfig. Not sure about the scope of this change to move this out of data selector. My thinking was this would be solved with cached fields. 

>  Can you just pull in Rocky's cache changes into your PR? He's going to be out until the end of next week so it will be quicker to just incorporate his changes in your PR.

Sure, I can but not sure how much of it is working: https://github.com/opensearch-project/OpenSearch-Dashboards/commit/a9fa1dfad809cd7522ebd150137dda983bd1d1f6#diff-06a5092e8cbf6b8504113441021731cf1fa1bcc6f5a5d355a26dfc490b7ce46bR105. Can someone from the team validate if this is all what is needed? Might not have the bandwidth to take the cache part right now.

src/plugins/query_enhancements/public/datasets/s3_type.ts
ps48: if `(dataType && dataType !== '')` specifically excludes empty strings.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
ps48: Totally agree, have already changed it once. Would do as a fast follow https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8337#discussion_r1779023524

src/plugins/data/public/query/query_string/dataset_service/types.ts
ps48: Keeping it future proof. Any change in the second parameter needs changes in 9 places 🗡️  as of now. As Amar has already pointed out the downstream consumers of this API are not liking these core changes to be made anymore https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8337#discussion_r1777939295.

src/plugins/data/common/datasets/types.ts
ps48: `supportsTimeFilter` can change the name, if you have a hard opinion. But I disagree with your point that SQL and PPL async queries don't support time filter. We might have opted out of time filters support for these due to time constraints or product requirements. 

Example of time filters in Async SQL/PPL:
```
SELECT * FROM sample_table  WHERE  startTime < CAST('01-sep-2022' as date) AND (startTime >= CAST('02-sep-2022' as date) 

source = sample_table | where startTime < '09-01-2022:00.00.00.000' and startTime >= '09-02-2022:00.00.00.000' 
```

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -62,29 +62,34 @@ export class DatasetService {
...
-              name: dataset.dataSource.title,
-              type: dataset.dataSource.type,
-            }
-          : undefined,
-      } as IndexPatternSpec;
-      const temporaryIndexPattern = await this.indexPatterns?.create(spec, true);
-      if (temporaryIndexPattern) {
-        this.indexPatterns?.saveToCache(dataset.id, temporaryIndexPattern);
+    try {
+      if (dataset) {
```
amsiglan, line 68: Given we check for `dataset` should we use `dataset?.type` in line 66?

src/plugins/data/public/query/query_string/dataset_service/dataset_service.ts
```diff
@@ -62,29 +62,34 @@ export class DatasetService {
...
+              }
+            : undefined,
+        } as IndexPatternSpec;
+        const temporaryIndexPattern = await this.indexPatterns?.create(spec, true);
+        if (temporaryIndexPattern) {
+          this.indexPatterns?.saveToCache(dataset.id, temporaryIndexPattern);
+        }
       }
+    } catch (error) {
+      throw new Error(`Failed to load dataset: ${dataset.id}`);
```
amsiglan, line 92: same here use `dataset?.id`?

ps48: updated here: [af973c2](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8337/commits/af973c2457eac1754fc48b07c6f240b326fe55c4)

amsiglan (approved):

ashwin-pc:
@ps48 thanks
1.  For the copy changes, yes UX should be our PoC, can you check with them?
2. As for the rest of the comments, i agree, and they can be fast follows. Can you open an issue tagged with discover to track it? 
3. Same goes for caching. @kavilla should be back monday to pick it back up

ps48:
Incorporated all the requested changes in this PR: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8609 Closing this one in favor of the new PR.

PR #8345, opened 2024-09-25
Add UI Metric for Discover 2.0

LDrago27:
### Description

The PR aims at capturing following UI Metrics. 

1.  No of times discover has been loaded
2. No of times new discover has been loaded
3. No of times each datasource type queries have been executed
4. No of times each language queries have been executed

## Screenshot

image

## Testing the changes

1. Set usageCollection.uiMetric.enabled to true
5. Call the API /api/stats?extended=true&legacy=true&exclude_usage=false
6. Verify the ui_metric doesn't contain an entry corresponding to Discover
7. Navigate to Discover
8. Interact with Discover change the datasources, language
9. Wait for 90s
10. Call the API /api/stats?extended=true&legacy=true&exclude_usage=false and verify ui_metric contains an entry corresponding to Discover with the metrics corresponding to above interaction

## Changelog

- skip
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/index_pattern_select/index_pattern_select.test.tsx
+++ b/src/plugins/data/public/ui/index_pattern_select/index_pattern_select.test.tsx
@@ -49,7 +49,7 @@ describe('IndexPatternSelect', () => {
 
     bulkGetMock.mockResolvedValue({ savedObjects: [{ attributes: { title: 'test1' } }] });
     compInstance.debouncedFetch('');
-    await new Promise((resolve) => setTimeout(resolve, 300));
+    await new Promise((resolve) => setTimeout(resolve, 600));
     await nextTick();
     expect(bulkGetMock).toBeCalledWith([{ id: 'testDataSourceId', type: 'data-source' }]);
   });
--- a/src/plugins/data_explorer/public/components/app_container.tsx
+++ b/src/plugins/data_explorer/public/components/app_container.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import React, { memo, useRef } from 'react';
+import React, { memo, useEffect, useRef } from 'react';
 import {
   EuiFlexGroup,
   EuiFlexItem,
@@ -23,6 +23,13 @@ import './app_container.scss';
 import { useOpenSearchDashboards } from '../../../opensearch_dashboards_react/public';
 import { IDataPluginServices } from '../../../data/public';
 import { QUERY_ENHANCEMENT_ENABLED_SETTING } from './constants';
+import {
+  DISCOVER_LOAD_EVENT,
+  NEW_DISCOVER_LOAD_EVENT,
+  NEW_DISCOVER_OPT_IN,
+  NEW_DISCOVER_OPT_OUT,
+  trackUiMetric,
+} from '../ui_metric';
 
 export const AppContainer = React.memo(
   ({ view, params }: { view?: View; params: AppMountParameters }) => {
@@ -35,8 +42,14 @@ export const AppContainer = React.memo(
 
     const topLinkRef = useRef<HTMLDivElement>(null);
     const datePickerRef = useRef<HTMLDivElement>(null);
+
     if (!view) {
       return <NoView />;
+    }
+
+    trackUiMetric(DISCOVER_LOAD_EVENT);
+    if (isEnhancementsEnabled) {
+      trackUiMetric(NEW_DISCOVER_LOAD_EVENT);
     }
 
     const { Canvas, Panel, Context } = view;
--- a/src/plugins/data_explorer/public/services.ts
+++ b/src/plugins/data_explorer/public/services.ts
@@ -1,0 +1,10 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { createGetterSetter } from '../../opensearch_dashboards_utils/public';
+import { UsageCollectionSetup } from '../../usage_collection/public';
+
+export const [getUsageCollector, setUsageCollector] = createGetterSetter<UsageCollectionSetup>(
+  'UsageCollector'
+);
--- a/src/plugins/data_explorer/public/ui_metric/constants.ts
+++ b/src/plugins/data_explorer/public/ui_metric/constants.ts
@@ -1,0 +1,8 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const NEW_DISCOVER_LOAD_EVENT = 'new_discover_load_count';
+export const DISCOVER_LOAD_EVENT = 'discover_load_count';
+export const NEW_DISCOVER_APP_NAME = 'New_Discover';
--- a/src/plugins/data_explorer/public/ui_metric/index.ts
+++ b/src/plugins/data_explorer/public/ui_metric/index.ts
@@ -1,0 +1,7 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './constants';
+export * from './report_ui_metric';
--- a/src/plugins/discover/public/application/view_components/utils/use_search.ts
+++ b/src/plugins/discover/public/application/view_components/utils/use_search.ts
@@ -8,7 +8,7 @@ import { BehaviorSubject, Subject, merge } from 'rxjs';
 import { debounceTime } from 'rxjs/operators';
 import { i18n } from '@osd/i18n';
 import { useEffect } from 'react';
-import { cloneDeep } from 'lodash';
+import { cloneDeep, isEqual } from 'lodash';
 import { useLocation } from 'react-router-dom';
 import { RequestAdapter } from '../../../../../inspector/public';
 import { DiscoverViewServices } from '../../../build_services';
@@ -33,6 +33,7 @@ import {
 } from '../../../opensearch_dashboards_services';
 import { SEARCH_ON_PAGE_LOAD_SETTING } from '../../../../common';
 import { syncQueryStateWithUrl } from '../../../../../data/public';
+import { trackQueryMetric } from '../../../ui_metric';
 
 export enum ResultStatus {
   UNINITIALIZED = 'uninitialized',
@@ -183,6 +184,12 @@ export const useSearch = (services: DiscoverViewServices) => {
       searchSource.getSearchRequestBody().then((body: object) => {
         inspectorRequest.json(body);
       });
+
+      // Track the dataset type and language used
+      const query = searchSource.getField('query');
+      if (query && query.dataset?.type && query.language) {
+        trackQueryMetric(query);
+      }
 
       // Execute the search
       const fetchResp = await searchSource.fetch({
--- a/src/plugins/discover/public/ui_metric/constants.ts
+++ b/src/plugins/discover/public/ui_metric/constants.ts
@@ -1,0 +1,8 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const DATASET_METRIC_SUFFIX = 'dataset_queries_count';
+export const LANGUAGE_METRIC_SUFFIX = 'language_queries_count';
+export const NEW_DISCOVER_APP_NAME = 'New_Discover';
--- a/src/plugins/discover/public/ui_metric/report_ui_metric.ts
+++ b/src/plugins/discover/public/ui_metric/report_ui_metric.ts
@@ -1,0 +1,37 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { UiStatsMetricType } from 'packages/osd-analytics/target/types';
+import { METRIC_TYPE } from '../../../usage_collection/public';
+import { DATASET_METRIC_SUFFIX, LANGUAGE_METRIC_SUFFIX, NEW_DISCOVER_APP_NAME } from './constants';
+import { getUsageCollector } from '../opensearch_dashboards_services';
+import { Query } from '../../../data/public';
+
+export const getDatasetTypeMetricEventName = (datasource: string) => {
+  return `${datasource}_${DATASET_METRIC_SUFFIX}`;
+};
+
+export const getLanguageMetricEventName = (language: string) => {
+  return `${language}_${LANGUAGE_METRIC_SUFFIX}`;
+};
+
+export const trackUiMetric = (
+  eventName: string,
+  appName: string = NEW_DISCOVER_APP_NAME,
+  metricType: UiStatsMetricType = METRIC_TYPE.COUNT
+) => {
+  try {
+    const usageCollector = getUsageCollector();
+    usageCollector.reportUiStats(appName, metricType, eventName);
+  } catch (error) {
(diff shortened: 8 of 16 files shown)


src/plugins/data_explorer/public/services.ts
```diff
@@ -0,0 +1,16 @@
+/*
```
kavilla, line 1: i think this is the license header

```
/*
 */
```

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -236,6 +241,19 @@ export const useSearch = (services: DiscoverViewServices) => {
...
+      const query = searchSource.getField('query');
+      const queryStringHistory = data.query.queryString.getQueryHistory();
+      if (
+        query &&
+        !isEqual(query, queryStringHistory.slice(-1)[0]) &&
+        query.dataset?.type &&
+        query?.language
+      ) {
+        trackUiMetric(`${query?.dataset?.type}_${DATASOURCE_METRIC_SUFFIX}`);
+        trackUiMetric(`${query?.language}_${LANGUAGE_METRIC_SUFFIX}`);
```
kavilla, line 255: i never got the chance to add it for sql and ppl but here is how we were doing it for DQL

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/ui/query_editor/query_editor.tsx#L202

src/plugins/data_explorer/public/ui_metric/index.ts
kavilla, line 1: appreciation comment for adding an index file

kavilla (commented):
couple nits and potential suggestion

src/plugins/discover/public/ui_metric/constants.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const DATASOURCE_METRIC_SUFFIX = 'datasource_queries_count';
```
kavilla, line 6: we could probably put the non discover specific constants in the data plugin common so it can be reusable

src/plugins/discover/public/application/view_components/utils/use_search.ts
LDrago27: For DQL they were using a custom version of usageCollector and this Api route basically carters to it. For our purposes the default Ui Metric usage collector is good enough. Therefore we are using the `reportUiStats` methods that does server side batching and writes the report onto the savedObject every 60s.

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -236,6 +241,19 @@ export const useSearch = (services: DiscoverViewServices) => {
           elapsedMs,
         },
       });
+
+      // Add tracking for datasource type and query if the query is not same as last query
+      const query = searchSource.getField('query');
+      const queryStringHistory = data.query.queryString.getQueryHistory();
+      if (
```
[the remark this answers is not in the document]
LDrago27: @kavilla . That actually seems more neat. Will move the tracking to over there.

src/plugins/data_explorer/public/services.ts
LDrago27: Thanks. Fixed it.

src/plugins/data_explorer/public/ui_metric/constants.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const DATASOURCE_METRIC_SUFFIX = 'datasource_queries_count';
```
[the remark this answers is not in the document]
LDrago27: Yeah that is good idea I will have other helper methods which will convert the data point into a qualified event name.

src/plugins/discover/public/ui_metric/constants.ts
LDrago27: @kavilla I was thinking of slightly different approach where each plugin can have a folder called ui_metric which will contain all the information related to ui_metric specific to that plugin over there. So each plugin can independently maintain all the constants and events that they want to be tracked in one place.

src/plugins/discover/public/application/view_components/utils/use_search.ts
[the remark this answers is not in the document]
kavilla: apologies i wasn't too clear with my suggestion. 🤕 

what i meant that here you are access the data plugin to get the information to track it and which point you are interacting with the value after you get the response.

but instead of accessing the data plugin after you get a response, you move this logic within the data plugin where it has context of the previous query and the current query and the link above was more so to indicate how it has access to the previous query. so we could add a function in the data plugin that reports a ui metric if there was a new query. the ui service could even just subscribe to the query service or maybe the query service itself reports the metric since it already has access to the querystring history

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -236,6 +241,19 @@ export const useSearch = (services: DiscoverViewServices) => {
...
+        query &&
+        !isEqual(query, queryStringHistory.slice(-1)[0]) &&
+        query.dataset?.type &&
+        query?.language
+      ) {
+        trackUiMetric(getDatasourceMetricEventName(query.dataset.type));
```
kavilla, line 254: nit: this only tracks on success. did we want to track the ui metric regardless of response?

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -236,6 +241,19 @@ export const useSearch = (services: DiscoverViewServices) => {
           elapsedMs,
         },
       });
+
+      // Add tracking for datasource type and query if the query is not same as last query
+      const query = searchSource.getField('query');
```
kavilla, line 246: can't remember if discover creates a new search source and it doesn't have a parent but we could also check if searchSource.parent.getField('query') . but then again i haven't checked so it parent might not ever be set for discover

src/plugins/data_explorer/public/components/app_container.tsx
```diff
@@ -39,6 +40,11 @@ export const AppContainer = React.memo(
       return <NoView />;
     }
 
+    trackUiMetric(DISCOVER_LOAD_EVENT);
+    if (isEnhancementsEnabled) {
+      trackUiMetric(NEW_DISCOVER_LOAD_EVENT);
```
kavilla, line 45: nit: did we want to mimic the `opt_in` `opt_out`  metrics. 

correct me if im wrong though, but i think this will just increment a counter any time it loads and passes this condition but it might hide when a user decides to initially enables it and decides they don't like the new discover so then they disable it. one user might not realize that there is a new Discover experience until a few weeks later so then the Discover load count would be way higher than the new Discover load count but they never opted out since they prefer the new Discover experience. But to us we might deduce that people do not like the new experience.

So the opt in and opt out count actually helps give us a better story that people opted in and decided to opt out

src/plugins/discover/public/application/view_components/utils/use_search.ts
LDrago27: Updated it to track the event before the search actually happens.

src/plugins/data_explorer/public/components/app_container.tsx
LDrago27: Thanks @kavilla added a new metrics to track opt_in and opt_out

src/plugins/data_explorer/public/ui_metric/constants.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const NEW_DISCOVER_LOAD_EVENT = 'new_discover_load_count';
```
kavilla, line 6: super duper nit: don't change this PR specifically for this but the import line for src/plugins/data_explorer/public/components/app_container.tsx could be a smaller and not require future updates if we utilize an object. and easier to access for developers. and probably should keep the casing and the same.

suggestion like:

```typescript
const APP_NAME = `discover`;
const LOAD_EVENT = `load_count`;
const OPT_IN_EVENT = `opt_in_count`;
const OPT_OUT_EVENT = `opt_out_count`;
export const METRIC = {
  DEFAULT: {
     LOAD_EVENT: `${APP_NAME}_${LOAD_EVENT}`,
     OPT_IN_EVENT: `${APP_NAME}_${OPT_IN_EVENT}`,
     OPT_OUT_EVENT: `${APP_NAME}_${OPT_OUT_EVENT}`
  },
  LEGACY: {
     LOAD_EVENT: `${APP_NAME}_legacy_${LOAD_EVENT}`,
     OPT_IN_EVENT: `${APP_NAME}_legacy_${OPT_IN_EVENT}`,
     OPT_OUT_EVENT: `${APP_NAME}_legacy_${OPT_OUT_EVENT}`
  }
}
```

then we can just import `METRIC` and access the constant with the dot notation.

some things to mention is that proposing using `legacy` instead of classifying vs using the `new` because with new or utilizing the `2.0` is that we version this ui metric to be 2.0 and then if we make the new discover the default then it becomes confusing for those who don't have context that the the default is the `new` ui metric and the previous version was the one without any classifying word on the ui metric. however, im not super tied to calling legacy. but i think we should go with legacy or new and avoid the `2.0` or `2`.

also opted to ensure all the metrics are the same structure like

```typescript
appname_classifier_event_metrictype
```

classifier being like `new` or `legacy` or nothing. this way if other plugins utilize the same standard then we have a nice way to aggregate on the results in a similar structure for visualizations and decision making. vs writing a query that filters out only the new and making the regex be something so it filters for `new_discover` or `discover_new`.

the final is using of interpolation this way if another developer wants to track another metric they can add by seeing the standard of using interpolation and do it as well. the reason why because i've seen people hardcode type constants over and over again and then there was a typo at which point we used that in key for local storage. but since the users used it already we are kind of stuck. so if we reduce the amount of opportunities for someone to make a typo then we could avoid an issue like queries failing to show up if the query was `appName: discover` but the developer added a new metric `discoever_refresh_count` where they mixed the `e` and the `v` in discover.

src/plugins/data/public/ui/index_pattern_select/index_pattern_select.test.tsx
```diff
@@ -49,7 +49,7 @@ describe('IndexPatternSelect', () => {
 
     bulkGetMock.mockResolvedValue({ savedObjects: [{ attributes: { title: 'test1' } }] });
     compInstance.debouncedFetch('');
-    await new Promise((resolve) => setTimeout(resolve, 300));
+    await new Promise((resolve) => setTimeout(resolve, 600));
```
ashwin-pc, line 52: Why this change?

src/plugins/data_explorer/public/components/app_container.tsx
```diff
@@ -23,6 +23,13 @@ import './app_container.scss';
 import { useOpenSearchDashboards } from '../../../opensearch_dashboards_react/public';
 import { IDataPluginServices } from '../../../data/public';
 import { QUERY_ENHANCEMENT_ENABLED_SETTING } from './constants';
+import {
+  DISCOVER_LOAD_EVENT,
+  NEW_DISCOVER_LOAD_EVENT,
+  NEW_DISCOVER_OPT_IN,
+  NEW_DISCOVER_OPT_OUT,
```
ashwin-pc, line 30: Why do we need all 4? New discover cant be triggered without a page refresh. So the useEffect based tracking should be unnecessary.

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -184,6 +189,14 @@ export const useSearch = (services: DiscoverViewServices) => {
         inspectorRequest.json(body);
       });
 
+      // Track the dataset type and language used
+
+      const query = searchSource.getField('query');
+      if (query && query.dataset?.type && query.language) {
```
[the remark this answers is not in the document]
ashwin-pc: Its the former. Tracking dataset type and language are important only in the new discover

ashwin-pc (commented):
Lemme know your thoughts on the useEffect based tracking. The rest look good!

src/plugins/discover/public/application/view_components/utils/use_search.ts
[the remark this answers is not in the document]
LDrago27: The requirements are specific to the new Discover experience , therefore in those cases where we don't have the datasets and language with enhancements off are not something we need to be worried about.

src/plugins/discover/public/ui_metric/report_ui_metric.ts
```diff
@@ -0,0 +1,35 @@
...
+import { UiStatsMetricType } from 'packages/osd-analytics/target/types';
+import { METRIC_TYPE } from '../../../usage_collection/public';
+import {
+  DATASOURCE_METRIC_SUFFIX,
+  LANGUAGE_METRIC_SUFFIX,
+  NEW_DISCOVER_APP_NAME,
+} from './constants';
+import { getUsageCollector } from '../opensearch_dashboards_services';
+
+export const getDatasourceMetricEventName = (datasource: string) => {
```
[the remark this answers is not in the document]
LDrago27: This is neat. I have updated the code to reflect it.

src/plugins/data/public/ui/index_pattern_select/index_pattern_select.test.tsx
LDrago27: This test case was a bit flakey in nature. hence to resolve it I had increased the timeout.

src/plugins/data_explorer/public/components/app_container.tsx
LDrago27: Oh. Removing the OPT_IN and OPT_OUT tracking from here.

src/plugins/discover/public/ui_metric/constants.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const DATASOURCE_METRIC_SUFFIX = 'data_source_queries_count';
```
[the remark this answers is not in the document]
LDrago27: Updated the Metric to dataset_queries_count

src/plugins/data_explorer/public/services.ts
LDrago27: It is updated in the latest PR

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -184,6 +189,14 @@ export const useSearch = (services: DiscoverViewServices) => {
         inspectorRequest.json(body);
       });
 
+      // Track the dataset type and language used
+
```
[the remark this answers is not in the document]
LDrago27: Fixed it.

sejli (approved):
+1 to @kavilla's nonblocking comment on the string constants, might need that documented somewhere as part of OSD best practices. Otherwise, LGTM.

src/plugins/data/public/ui/index_pattern_select/index_pattern_select.test.tsx
AMoo-Miki: This is not the right way to deal with timers.

PR #8347, opened 2024-09-25
[discover] fix too many queries sent

kavilla:
### Description

Fixes the issue where too many queries were being fired. A number of things were being called because the dependency array was triggering updates and re-calls. Forcing a re-render of elements that had the possibility of updating the query manager.

Includes some tests.

### Issues Resolved

n/a

## Testing the changes

locally and tests

## Changelog

- fix: prevent too many queries sent from dataset selector

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/query_string_manager.ts
+++ b/src/plugins/data/public/query/query_string/query_string_manager.ts
@@ -31,7 +31,7 @@
 import { BehaviorSubject } from 'rxjs';
 import { skip } from 'rxjs/operators';
 import { CoreStart } from 'opensearch-dashboards/public';
-import { isEqual, debounce } from 'lodash';
+import { isEqual } from 'lodash';
 import { Dataset, DataStorage, Query, TimeRange, UI_SETTINGS } from '../../../common';
 import { createHistory, QueryHistory } from './query_history';
 import { DatasetService, DatasetServiceContract } from './dataset_service';
@@ -123,14 +123,6 @@ export class QueryStringManager {
       this.query$.next(newQuery);
     }
   };
-
-  /**
-   * Updates the query with debounce
-   * @param {Query} query
-   */
-  public debouncedSetQuery = debounce((query: Partial<Query>) => {
-    this.setQuery(query);
-  }, 100);
 
   /**
    * Resets the query to the default one.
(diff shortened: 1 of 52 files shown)


abbyhu2000 (approved):

src/plugins/data/public/ui/dataset_selector/index.tsx
```diff
@@ -17,25 +17,21 @@ interface ConnectedDatasetSelectorProps {
...
-  const initialDataset = queryString.getQuery().dataset || queryString.getDefaultQuery().dataset;
-  const [selectedDataset, setSelectedDataset] = useState<Dataset | undefined>(initialDataset);
-
-  useEffect(() => {
-    const subscription = queryString.getUpdates$().subscribe((query) => {
-      setSelectedDataset(query.dataset);
-    });
-
-    return () => subscription.unsubscribe();
-  }, [queryString]);
```
sejli, line 29: I remember being confused why this is here. Was there a reason why we needed two places to set the selected dataset?

sejli (approved):
Amazing

src/plugins/data/public/ui/dataset_selector/index.tsx
kavilla: not positive. probably was just code that was intended to be cleaned up i think i at one point we would set the query at the configurator level instead of using the function that we passed in to set it and then when we got time to wire up and use the passed in function it caused some issue with re-rendering.

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -68,6 +68,7 @@ export class QueryStringManager {
     const query = {
       query: defaultQuery,
       language: defaultLanguageId,
+      dataset: undefined,
```
abbyhu2000, line 71: why is this necessary? i see query is used to construct `newQuery` and dataset is set to be defaultDataset in line 81

kavilla: my linter was complaining because i thought the return type was just query language

joshuali925 (approved):
nice, lgtm

sejli (approved):

abbyhu2000 (approved):

kavilla:
failures unrelated and i think another PR is fixing it.

PR #8368, opened 2024-09-27
fix(queryEnhancements): data.search() should not ignore the `strategy` passed as parameter

ruanyl:
This is quick fix for the issue that the search strategy passed to data.search() API been completely ignored when query enhancement is enabled and the localStorage has a language preserved.

For example, selecting a language from discover language selector will make data.search() called with a specific strategy to use the language config stored in localStorage which is unexpected.

With this fix, language config is only used when search strategy is not passed explicitly.

Sharing my two cents:
1. I think it's better the `language` is passed as a parameter to `data.search()` instead of reading directly from query string manager(which has default value from local storage that different feature could share the same value). I think `data.search()` needs to be UI/feature-agnostic
2. I'm fine with having different search intercepters for different languages, but it seems the current intercepters(ppl & sql) does way too much than than an "intercepter" should. IMHO, a search intercepter will just intercepts the search params(input) and search response(output), and does things like error handling, validation, etc. But now it seems we're building the search query in the intercepter by reading from some global state: time fitter, aggregation, etc. Shouldn't this be done upfront and passed to `data.search()` as search params?

I may not have a full picture of the entire idea of query enhancement, and I'm open to any other ideas :)

## Changelog

- fix:  data.search() should not ignore the strategy passed as parameter

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/8368.yml
+++ b/changelogs/fragments/8368.yml
@@ -1,0 +1,2 @@
+fix:
+- Data.search() should not ignore the strategy passed as parameter ([#8368](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8368))
--- a/src/plugins/data/public/search/search_service.ts
+++ b/src/plugins/data/public/search/search_service.ts
@@ -155,7 +155,7 @@ export class SearchService implements Plugin<ISearchSetup, ISearchStart> {
   ): ISearchStart {
     const search = ((request, options) => {
       const isEnhancedEnabled = uiSettings.get(UI_SETTINGS.QUERY_ENHANCEMENTS_ENABLED);
-      if (isEnhancedEnabled) {
+      if (isEnhancedEnabled && !options?.strategy) {
         const queryStringManager = getQueryService().queryString;
         const language = queryStringManager.getQuery().language;
         const languageConfig = queryStringManager.getLanguageService().getLanguage(language);


SuZhou-Joe (approved):

ashwin-pc:
@AMoo-Miki @kavilla Can you review this?

@ruanyl Can you add details about why this is needed now?

ashwin-pc:
@ruanyl i dont this the interceptors do that much anymore. afaik the ppl search interceptor should only add the time field to the ppl query from the query state and nothign more. SQL should make no modifications to the query. Is this not whats happening? I agree with you that the inerceptor should not modify anything in the query. I think the time selector is the only exception here but i could be swayed there as well. 

If this is true, do we still need your change?

ruanyl:
@ashwin-pc You're right, that's true for sql interceptor. For ppl interceptor it adds time filter even when there is a time filter in the ppl query, for example, I have ppl like `source = opensearch_dashboards_sample_data_ecommerce | where order_date > DATE_SUB(NOW(), INTERVAL 7 DAY) | stats count()`

The actually query send becomes this which adds the time range from the ui
```
query: "source = opensearch_dashboards_sample_data_ecommerce  | where `order_date` >= '2024-09-30 23:57:59' and `order_date` <= '2024-10-01 00:12:59' | where order_date > DATE_SUB(NOW(), INTERVAL 7 DAY) | stats count()"
```

And also it seems it modifies the aggregation which I don't fully understand the reason, maybe it's the default aggregation based on the time range?
```
        aggsConfig.qs = {
          [key]: `${query.query} | stats count() by span(${query.dataset!.timeFieldName}, ${
            value[aggTypeKey].fixed_interval ??
            value[aggTypeKey].calendar_interval ??
            this.aggsService.calculateAutoTimeExpression({
              from: fromDate,
              to: toDate,
              mode: 'absolute',
            })
          })`,
        };
```

> If this is true, do we still need your change?

We still need this change because this change is about the 1st point for the language selector. The issue that I'm having is I'm using another search strategy, but after I selected PPL or SQL on the discover page, because of the following code, it will always use what's selected(ppl or sql) on discover app even I'm on another app. And actually, it calls `data.search(request, {strategy: "whatever_strategy"})` directly and has nothing to do with the language selected on discover app.
```
      if (isEnhancedEnabled) {
        const language = queryStringManager.getQuery().language;
        const languageConfig = queryStringManager.getLanguageService().getLanguage(language);
        queryStringManager.getLanguageService().setUiOverridesByUserQueryLanguage(language);

        if (languageConfig) {
          return languageConfig.search.search(request, options);
        }
      }
```

ashwin-pc:
I'll let @kavilla answer why we modify aggs. Based on my understanding, we shouldn't. Also adding the time filter even if we already have on in the query is intended behavior to be consistent. If the UI has a time filter, it makes more sense to apply it to the query than not because it simpler to reason that way. Optionally adding it to the query makes it harder to reason when the UI time filter is applied and when the query time filter is used. 

As for the change you are making, it seems fine temporarily to unblock your usecase but long term the goal is to have a strategy per dataset type and a interceptor per language and this solution breaks that. @kavilla will do a post 2.18 review of the design since we'll have to consolidate all the recent changes we've been making to the search pipeline.

joshuali925 (commented):
i briefly looked into this, the fix looks good, but how do we know the `options.strategy` is supposed to be used with the `defaultSearchInteceptor` but not the `languageConfig.search` interceptor?

ashwin-pc:
That was my concern as well.

ruanyl:
> i briefly looked into this, the fix looks good, but how do we know the `options.strategy` is supposed to be used with the `defaultSearchInteceptor` but not the `languageConfig.search` interceptor?

@ashwin-pc @joshuali925 
Thanks for following up :)

Both the current `sql_search_interceptor` and `ppl_search_interceptor` don't "care" the `options.strategy`, they have it's own logic to "compute" the strategy internally:
For sql_search_interceptor:
```
    let strategy = datasetType === DATASET.S3 ? SEARCH_STRATEGY.SQL_ASYNC : SEARCH_STRATEGY.SQL;

    if (datasetType) {
      const datasetTypeConfig = this.queryService.queryString
        .getDatasetService()
        .getType(datasetType);
      strategy = datasetTypeConfig?.getSearchOptions?.().strategy ?? strategy;
    }
```

For ppl_search_interceptor:
```
    let strategy = SEARCH_STRATEGY.PPL;

    if (datasetType) {
      const datasetTypeConfig = this.queryService.queryString
        .getDatasetService()
        .getType(datasetType);
      strategy = datasetTypeConfig?.getSearchOptions?.().strategy ?? strategy;
    }
```

So this PR won't affect the existing two languageConfig interceptors. However, I'd say this PR is just a temporary workaround to fix the issue that the passed strategy been ignore. I believe we would need a throughly refactor on this.

joshuali925 (approved):
i'm ok with this temporary workaround. i think it would be good to have some comments to call out that the framework now assumes all search interceptors from language configs do not use `options.strategy` and do expect `options.strategy` to be absent

@kavilla could you also take a look since you are more familiar with this part

ashwin-pc (approved):

PR #8422, opened 2024-10-01
[Workspace] feat: support use case populate for workspace create and list page

raintygao:
### Description

1. support use case as router search populate for workspace create and list page
2. remove owners column in workspace list page

## Screenshot

image

image

## Testing the changes

Go to workspace create/list page and use useCase search, for example
 http://localhost:6601/app/workspace_list#/?useCase=Security%20Analytics , then you could see security use case to be populated.

## Changelog

- feat: Support use case populate for workspace create and list page

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/public/application.tsx
+++ b/src/plugins/workspace/public/application.tsx
@@ -25,7 +25,13 @@ export const renderCreatorApp = (
 ) => {
   ReactDOM.render(
     <OpenSearchDashboardsContextProvider services={services}>
-      <WorkspaceCreatorApp {...props} />
+      <Router>
+        <Switch>
+          <Route>
+            <WorkspaceCreatorApp {...props} />
+          </Route>
+        </Switch>
+      </Router>
     </OpenSearchDashboardsContextProvider>,
     element
   );
@@ -56,7 +62,13 @@ export const renderListApp = (
 ) => {
   ReactDOM.render(
     <OpenSearchDashboardsContextProvider services={services}>
-      <WorkspaceListApp {...props} />
+      <Router>
+        <Switch>
+          <Route>
+            <WorkspaceListApp {...props} />
+          </Route>
+        </Switch>
+      </Router>
     </OpenSearchDashboardsContextProvider>,
     element
   );
--- a/src/plugins/workspace/public/components/utils/workspace.ts
+++ b/src/plugins/workspace/public/components/utils/workspace.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { WORKSPACE_DETAIL_APP_ID } from '../../../common/constants';
+import { WORKSPACE_DETAIL_APP_ID, WORKSPACE_LIST_APP_ID } from '../../../common/constants';
 import { CoreStart } from '../../../../../core/public';
 import { formatUrlWithWorkspaceId } from '../../../../../core/public/utils';
 import { DetailTab } from '../workspace_form/constants';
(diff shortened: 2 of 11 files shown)


src/plugins/workspace/public/application.tsx
```diff
@@ -25,7 +25,13 @@ export const renderCreatorApp = (
...
   ReactDOM.render(
     <OpenSearchDashboardsContextProvider services={services}>
-      <WorkspaceCreatorApp {...props} />
+      <Router>
+        <Switch>
+          <Route>
+            <WorkspaceCreatorApp {...props} />
+          </Route>
+        </Switch>
+      </Router>
```
Hailong-am, line 34: why do we need add router?

src/plugins/workspace/public/components/utils/workspace.ts
```diff
@@ -23,6 +23,12 @@ export const navigateToWorkspaceDetail = (
   );
 };
 
+export const navigateToWorkspaceListWithUseCase = ({ application }: Core, useCaseTitle: string) => {
+  const url = application.getUrlForApp(WORKSPACE_LIST_APP_ID, { absolute: true });
+  const search = `/?useCase=${useCaseTitle}`;
```
Hailong-am, line 28: can we use hash, so that router change is not needed? similar with workspace detail page

Hailong-am (approved):

src/plugins/workspace/public/application.tsx
SuZhou-Joe: I have the same concern that is the HashRouter here necessary?

src/plugins/workspace/public/components/workspace_creator/workspace_creator.tsx
```diff
@@ -53,16 +54,28 @@ export const WorkspaceCreator = (props: WorkspaceCreatorProps) => {
...
-          features: [getUseCaseFeatureConfig(defaultSelectedUseCase.id)],
-        }
-      : {}),
-  };
+  const location = useLocation();
+
+  const defaultWorkspaceFormValues = useMemo(() => {
+    let defaultSelectedUseCase;
+    const params = new URLSearchParams(location.search);
+    const useCaseTitle = params.get('useCase');
```
SuZhou-Joe, line 62: Shall we use id instead of title in workspace creator page?

src/plugins/workspace/public/components/utils/workspace.ts
```diff
@@ -23,6 +23,12 @@ export const navigateToWorkspaceDetail = (
   );
 };
 
+export const navigateToWorkspaceListWithUseCase = ({ application }: Core, useCaseTitle: string) => {
```
SuZhou-Joe, line 26: I found this function hard to use in my local because it requires caller to pass in `http`, while the http is not used in the function actually.

src/plugins/workspace/public/application.tsx
raintygao: Yes, we need hashRouter to wrap switch and route.

src/plugins/workspace/public/components/workspace_creator/workspace_creator.tsx
raintygao: No, we should use title to display and search, because query would populate the search key in the input, if we use id, the search query would be `all` which is inconsistent with the `Analytics (all)` in the column.

src/plugins/workspace/public/components/utils/workspace.ts
raintygao: Oh I missed it, updated.

src/plugins/workspace/public/components/utils/workspace.ts
raintygao: Updated. This may just be a boilerplate for now, it actually depends on how @yubonluo uses later.

yubonluo: If necessary in the future, I will make some adjustments when integrating it to the homepage

src/plugins/workspace/public/components/workspace_creator/workspace_creator.tsx
SuZhou-Joe: This is workspace creator page right? Not workspace list page. We are `select` an option instead of `search` by title.

SuZhou-Joe (approved):

SuZhou-Joe (approved):

Hailong-am (approved):

PR #8481, opened 2024-10-04
Poll for results from frontend for async search

amsiglan:
### Description
Currently for async search we depend on the search strategy to start the query and poll for results and resolve once the query completes/fails but since can to the order of 1 to 2 min, the frontend will timeout. This PR adds logic in frontend to poll for the results once the query is started.

The search strategy's search method will be invoked with the query id when polling for results, else it should start search for the given query.

## Testing the changes
Tested using local setup and neo app 

## Changelog
- feat: Add logic to poll for async query result

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/common/utils.ts
+++ b/src/plugins/query_enhancements/common/utils.ts
@@ -4,7 +4,7 @@
  */
 
 import { Query } from 'src/plugins/data/common';
-import { from, throwError, timer } from 'rxjs';
+import { from, timer } from 'rxjs';
 import { filter, mergeMap, take, takeWhile } from 'rxjs/operators';
 import {
   EnhancedFetchContext,
@@ -49,7 +49,11 @@ export const handleFacetError = (response: any) => {
 
 export const fetch = (context: EnhancedFetchContext, query: Query, aggConfig?: QueryAggConfig) => {
   const { http, path, signal } = context;
-  const body = JSON.stringify({ query: { ...query, format: 'jdbc' }, aggConfig });
+  const body = JSON.stringify({
+    query: { ...query, format: 'jdbc' },
+    aggConfig,
+    queryId: context.body?.queryId,
+  });
   return from(
     http.fetch({
       method: 'POST',
--- a/src/plugins/query_enhancements/common/utils.ts
+++ b/src/plugins/query_enhancements/common/utils.ts
@@ -6,6 +6,7 @@
 import { Query } from 'src/plugins/data/common';
 import { from, timer } from 'rxjs';
 import { filter, mergeMap, take, takeWhile } from 'rxjs/operators';
+import { stringify } from '@osd/std';
 import {
   EnhancedFetchContext,
   QueryAggConfig,
@@ -49,10 +50,10 @@ export const handleFacetError = (response: any) => {
 
 export const fetch = (context: EnhancedFetchContext, query: Query, aggConfig?: QueryAggConfig) => {
   const { http, path, signal } = context;
-  const body = JSON.stringify({
+  const body = stringify({
     query: { ...query, format: 'jdbc' },
     aggConfig,
-    queryId: context.body?.queryId,
+    pollQueryResultsParams: context.body?.pollQueryResultsParams,
   });
   return from(
     http.fetch({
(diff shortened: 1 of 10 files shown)


src/plugins/query_enhancements/common/utils.ts
```diff
@@ -49,7 +49,11 @@ export const handleFacetError = (response: any) => {
 
 export const fetch = (context: EnhancedFetchContext, query: Query, aggConfig?: QueryAggConfig) => {
   const { http, path, signal } = context;
-  const body = JSON.stringify({ query: { ...query, format: 'jdbc' }, aggConfig });
+  const body = JSON.stringify({
```
AMoo-Miki, line 52: Should this use `stringify` from `@osd/std`?

AMoo-Miki (dismissed):

ashwin-pc (changes requested):
Thanks, love the changes. Can you also update the SQL search strategy to use this. Right now this change breaks the SQL strategy when querying remote sources like this.

ashwin-pc:
Screenshot 2024-10-04 at 4 53 02 AM

For reference

src/plugins/query_enhancements/common/utils.ts
amsiglan: I am going to go with your suggestion here since I am new to this part of the code. @ashwin-pc let me know if think this is wrong

amsiglan:
Fixed the async sql strategy and also the issue that we see `Cannot read message of undefined`

src/plugins/query_enhancements/server/search/sql_async_search_strategy.ts
```diff
@@ -44,45 +38,61 @@ export const sqlAsyncSearchStrategyProvider = (
...
+        const inProgressQueryId = pollQueryResultsParams?.queryId;
 
-        if (!rawResponse.success) handleFacetError(rawResponse);
+        if (!inProgressQueryId) {
+          request.body = { ...request.body, lang: SEARCH_STRATEGY.SQL };
+          const rawResponse: any = await sqlAsyncFacet.describeQuery(context, request);
 
-        const statusConfig = buildQueryStatusConfig(rawResponse);
-        request.params = { queryId: statusConfig.queryId };
+          if (!rawResponse.success) handleFacetError(rawResponse);
```
ashwin-pc, line 48: Shouldnt this return 

```js
{
  type: DATA_FRAME_TYPES.POLLING,
  status: 'failed',
  body: {
    error: new Error(`JOB: ${pollQueryResultsParams.queryId} failed`),
  }
}
```

ashwin-pc (approved):
This looks good. Verified that its working. Had one comment that can be a fast follow, but if the query fails while starting the request, it does not follow the same poling response structure.

src/plugins/query_enhancements/server/search/sql_async_search_strategy.ts
amsiglan: Let me update this in follow up PR

src/plugins/query_enhancements/server/search/sql_async_search_strategy.ts
```diff
@@ -44,45 +38,61 @@ export const sqlAsyncSearchStrategyProvider = (
...
+          if (queryStatus?.toUpperCase() === 'SUCCESS') {
+            const dataFrame = createDataFrame({
+              name: query.dataset?.id,
+              schema: queryStatusResponse.data.schema,
+              meta: { ...pollQueryResultsParams },
+              fields: getFields(queryStatusResponse),
+            });
 
-        dataFrame.size = response.data.datarows.length;
+            dataFrame.size = queryStatusResponse.data.datarows.length;
```
AMoo-Miki, line 73: Shouldn't we check that `datarows` is available and an array?

AMoo-Miki (approved):

src/plugins/query_enhancements/server/search/sql_async_search_strategy.ts
amsiglan: I will leave that decision to the owners of this strategy, this PR is only updating the strategy to use the polling mechanism

amsiglan: Will follow up on this with the right owners and make change if needed in follow up PR

Issue #8256, opened 2024-09-20
[BUG] Redirection issue for direct query datasource in dashboards management


RyanL1997:
**Describe the bug**

When MDS is enabled, we are still having the card view in datasource detail page in here shows "query data in observability logs" and forwards user to log exploer page which should be disabled when MDS is enabled
Same redirecting action is also being applied for specific accelerations

**To Reproduce**
- enable MDS
- go to a direct query datasource's detail page
-  video

https://github.com/user-attachments/assets/f5f78804-d679-4a5c-8fbe-43d6262a5b72

**Expected behavior**
when MDS is enabled, it should be directed into Discover

**OpenSearch Version**
main

**Dashboards Version**
Please list the version of OpenSearch Dashboards being used.

**Plugins**

need dashboards observability plugin
**Screenshots**

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

dblock:
[Catch All [Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/dblock), [2](https://github.com/reta),  [3](https://github.com/krisfreedain),  [4](https://github.com/lukas-vlcek/)]

RyanL1997:
Hi team, according to the recent gameday sanity testing, me and @ps48 observed that the current behavior of this re-direction is only handling the case for redirection to discover. The enhanced behavior should be:
- mds enabled: redirect to discover (https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8492 has handled that)
- mds disabled: keep the original redirection for log explorer 

According ot the above, I'm re-opening this issue, and I will update the description

PR #8492, opened 2024-10-04
Update Observability Data Sources with Type and Redirection

paulstn:
## Added data source type to the data sources page
For S3:
image
And Prometheus:
image

## Updating redirection to Discover now that Explorer cannot be redirected to
For the flows in associated objects and accelerations that require redirection, all instances can be separated out into two kinds:
- for tables and skipping indices: redirection to a 'flint' datasource with the datasource, database, and table name.
- for covering indices and materialized views: redirection to the associated index name. 
For both kinds of redirection, now the MDS id has to be used to query correctly. All of these flows were updated accordingly.

There is one last kind of redirection with this card:
image
That previously would redirect to Log Explorer with only the datasource and nothing else. In the current version of Discover, doing this would be impossible, and even alternatives such as opening the DataSet picker modal in discover with the datasource selected isn't feasible. Currently it is set to redirect to Discover without the datasource. 

### Issues Resolved

Resolves: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8256

## Testing the changes

Testing with MDS off:

https://github.com/user-attachments/assets/5c94910f-71d1-4a8d-b2bf-68aa2eeb7bca

(video cut for length but query eventually completes)

When query:enhancements:enabled is off, every redirection point is disabled as such:
image
This is done for 
- assc. obj. table
- assc. obj. flyout
- acc. table
- acc. flyout

## Changelog

- fix: Updated DataSource Management to redirect to Discover as well as display the type of the DataSource

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_source_management/public/components/direct_query_data_sources_components/associated_object_management/utils/associated_objects_tab_utils.tsx
```diff
@@ -43,49 +43,38 @@ export const isCatalogCacheFetching = (...statuses: DirectQueryLoadingStatus[])
...
-    state: {
-      datasourceName,
-      datasourceType: DATA_SOURCE_TYPES.S3Glue,
-    },
+  application.navigateToApp('data-explorer', {
+    path: `discover#?_a=(discover:(columns:!(_source),isDirty:!f,sort:!()),metadata:(view:discover))&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))&_q=(filters:!(),query:(dataset:(dataSource:(id:'${
+      datasourceMDSId ?? ''
+    }',meta:(name:${datasourceName},type:CUSTOM),title:'',type:DATA_SOURCE),id:'${
+      datasourceMDSId ?? ''
+    }::${datasourceName}.${databaseName}.${tableName}',title:${datasourceName}.${databaseName}.${tableName},type:S3),language:SQL,query:'SELECT%20*%20FROM%20${datasourceName}.${databaseName}.${tableName}%20LIMIT%2010'))`,
```
RyanL1997, line 58: Does this re-direction also needs the `query:enhancements:enabled` to be `true` to work?

src/plugins/data_source_management/public/components/direct_query_data_sources_components/connection_detail/direct_query_connection_detail.tsx
```diff
@@ -339,6 +342,25 @@ export const DirectQueryDataConnectionDetail: React.FC<DirectQueryDataConnection
   const S3DatasourceOverview = () => (
     <EuiPanel>
       <EuiFlexGroup>
+        <EuiFlexItem grow={false}>
```
RyanL1997, line 345: Nit: what is this change for?

RyanL1997 (commented):
Hi @paulstn , thanks for taking this on. I just left some comments. In addition, could you also add
> - Resolve https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8256

In the PR description, so that the above issue can be auto resolved by merging this PR.

RyanL1997:
Also, can we record some tests, for both MDS enabled and disabled?

src/plugins/data_source_management/public/components/direct_query_data_sources_components/acceleration_management/acceleration_table.tsx
```diff
@@ -206,11 +206,11 @@ export const AccelerationTable = ({
   const tableActions = [
     {
       name: 'Query Data',
-      description: 'Query in Observability Logs',
+      description: 'Query in Discover',
       icon: 'discoverApp',
       type: 'icon',
       onClick: (acc: CachedAcceleration) => {
-        onDiscoverIconClick(acc, dataSourceName, application);
+        onDiscoverIconClick(acc, dataSourceName, dataSourceMDSId, application);
```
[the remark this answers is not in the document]
paulstn: With a local cluster dataSourceMDSId would be undefined here and later in the actual function that does the redirection it will pass it in as '', i.e. within rison encoding format

src/plugins/data_source_management/public/components/direct_query_data_sources_components/acceleration_management/acceleration_table.tsx
```diff
@@ -206,11 +206,11 @@ export const AccelerationTable = ({
   const tableActions = [
     {
       name: 'Query Data',
-      description: 'Query in Observability Logs',
+      description: 'Query in Discover',
```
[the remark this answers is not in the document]
paulstn: Yes this PR would be deprecating that redirection as well

src/plugins/data_source_management/public/components/direct_query_data_sources_components/connection_detail/direct_query_connection_detail.tsx
paulstn: This makes the change that the first two screenshots are about, they have the type of the datasource displayed.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/associated_object_management/utils/associated_objects_tab_utils.tsx
paulstn: It would need it to be true since the SQL language would be hidden under that setting

RyanL1997: Got it. So what happened if we disabled the query enhancement and perform a redirection in data source management?

paulstn: I've thought about it and I think the best solution here is to stop redirection with maybe a toast that says to enable query enhancements. With query enhancements off, there's no guarantee if we're in the new home to use explorer as a backup so we can only prevent it since there's no way to query flint tables.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/acceleration_management/acceleration_details_flyout.tsx
```diff
@@ -231,8 +232,9 @@ export const AccelerationDetailsFlyout = (props: AccelerationDetailsFlyoutProps)
   const DiscoverIcon = () => {
     return (
       <EuiButtonEmpty
+        isDisabled={!getUiSettings().get('query:enhancements:enabled')}
```
sejli, line 235: nit: `data_source_management` doesn't take a dependency on the `data` plugin, but could maybe add it [here](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source_management/common/index.ts) since we also have `DEFAULT_DATA_SOURCE_UI_SETTINGS_ID`.

changelogs/fragments/8492.yml
```diff
@@ -0,0 +1,2 @@
+fix:
+- Updated DataSource Management to redirect to Discover as well as display the type of the DataSource ([#8492](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8492))
```
sejli, line 2: Just FYI, the release notes generator requires the `CHANGELOG` description to be under 100 characters. Not sure if it counts the PR number, but the description itself is at 99 characters.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/associated_object_management/associated_objects_details_flyout.test.tsx
```diff
@@ -106,13 +106,9 @@ describe('AssociatedObjectsDetailsFlyout', () => {
...
     fireEvent.click(screen.getAllByRole('button')[0]);
     await waitFor(() => {
-      expect(mockApplication.navigateToApp).toHaveBeenCalledWith('observability-logs', {
-        path: '#/explorer',
-        state: {
-          datasourceName: 'testDatasource',
-          datasourceType: 's3glue',
-          queryToRun: 'source = testDatasource.testDatabase.testTable | head 10',
-        },
+      expect(mockApplication.navigateToApp).toHaveBeenCalledWith('data-explorer', {
```
sejli, line 109: nit: could use the util functions you made here in the tests too right?

src/plugins/data_source_management/public/components/direct_query_data_sources_components/associated_object_management/utils/associated_objects_tab_utils.tsx
```diff
@@ -43,49 +43,38 @@ export const isCatalogCacheFetching = (...statuses: DirectQueryLoadingStatus[])
...
-
-export const redirectToExplorerS3 = (datasourceName: string, application: ApplicationStart) => {
-  application.navigateToApp(observabilityLogsID, {
-    path: `#/explorer`,
-    state: {
-      datasourceName,
-      datasourceType: DATA_SOURCE_TYPES.S3Glue,
-    },
+  application.navigateToApp('data-explorer', {
+    path: `discover#?_a=(discover:(columns:!(_source),isDirty:!f,sort:!()),metadata:(view:discover))&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))&_q=(filters:!(),query:(dataset:(dataSource:(id:'${
```
sejli, line 54: nit: Feel like these two functions have a lot of duplicate code, could be worth a try to either combine them or wrap the actual path in another function.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/associated_object_management/utils/associated_objects_tab_utils.tsx
```diff
@@ -43,49 +43,38 @@ export const isCatalogCacheFetching = (...statuses: DirectQueryLoadingStatus[])
...
+      datasourceMDSId ?? ''
+    }::${datasourceName}.${databaseName}.${tableName}',title:${datasourceName}.${databaseName}.${tableName},type:S3),language:SQL,query:'SELECT%20*%20FROM%20${datasourceName}.${databaseName}.${tableName}%20LIMIT%2010'))`,
   });
 };
 
-const redirectToExplorerWithQuery = (
-  datasourceName: string,
-  datasourceType: string,
-  queriedIndex: string,
+export const redirectToDiscoverOSIdx = (
```
sejli, line 62: What does `redirectToDiscoverOSIdx` mean?

sejli (commented):
Looks good so far, just some general questions and nitpicks.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/associated_object_management/utils/associated_objects_tab_utils.tsx
paulstn: redirect to discover with opensearch index. the naming scheme was exactly the same except i just replaced 'explorer' with 'discover'

src/plugins/data_source_management/public/components/direct_query_data_sources_components/associated_object_management/utils/associated_objects_tab_utils.tsx
paulstn: Originally when it was just the 'state' being redirected to explorer through navigateToApp, that made a lot of sense. However, since Discover's url redirection has different forms between redirecting to a flint index vs an index pattern, the differences i would need to put in between both strings would be too much for it to be worth it, as I think it would greatly reduce readability.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/associated_object_management/associated_objects_details_flyout.test.tsx
paulstn: Yeah I think the previous tests that existed were just there for coverage purposes, I don't think I should call on the util functions since I believe that's what the tests are testing here.

changelogs/fragments/8492.yml
paulstn: Oh yeah my original description was over 100 characters and when I shortened it the changelog was accepted so I'd assume that the changebot checking the character count would be consistent with the release notes generator. That's just an assumption though.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/acceleration_management/acceleration_details_flyout.tsx
paulstn: Would there be any benefit to a dependency on the data plugin if we can just check the ui settings? Or were you pointing me to adding this UI setting as a constant to that folder?

RyanL1997 (approved):

sejli (approved):

ashwin-pc (approved):
Proxy approval for Ryan since he isnt a maintainer yet but is more familiar with this section of the codebase than me.

PR #8513, opened 2024-10-07
Add SearchOnLoad Preference for dataset

LDrago27:
### Description

This introduces a searchOnPageLoad preference for each dataset type. 
- When Advanced setting of searchOnPageLoad is OFF we respect the Advanced setting 
- When Advanced setting of searchOnPageLoad is ON we respect the selected dataset's searchOnLoad preference

## Testing the changes

For this demonstration below I have explicitly updated the `searchOnLoad` preference for indexes to be false.

https://github.com/user-attachments/assets/78cb0c86-959d-4b2a-b6d1-88874ac483eb

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/dataset_service/types.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/types.ts
@@ -30,6 +30,8 @@ export interface DatasetTypeConfig {
     icon: EuiIconProps;
     /** Optional tooltip text */
     tooltip?: string;
+    /** Optional preference for search on page load else defaulted to true */
+    searchOnLoad?: boolean;
   };
   /**
    * Converts a DataStructure to a Dataset.
--- a/src/plugins/discover/public/application/view_components/utils/use_search.ts
+++ b/src/plugins/discover/public/application/view_components/utils/use_search.ts
@@ -88,6 +88,7 @@ export const useSearch = (services: DiscoverViewServices) => {
   const [savedSearch, setSavedSearch] = useState<SavedSearch | undefined>(undefined);
   const { savedSearch: savedSearchId, sort, interval } = useSelector((state) => state.discover);
   const indexPattern = useIndexPattern(services);
+  const skipInitialFetch = useRef(false);
   const {
     data,
     filterManager,
@@ -111,6 +112,15 @@ export const useSearch = (services: DiscoverViewServices) => {
     requests: new RequestAdapter(),
   };
 
+  const getDatasetAutoSearchOnPageLoadPreference = () => {
+    // Checks the searchOnpageLoadPreference for the current dataset if not specifed defaults to true
+    const datasetType = data.query.queryString.getQuery().dataset?.type;
+
+    const datasetService = data.query.queryString.getDatasetService();
+
+    return !datasetType || (datasetService?.getType(datasetType)?.meta?.searchOnLoad ?? true);
+  };
+
   const shouldSearchOnPageLoad = useCallback(() => {
     // A saved search is created on every page load, so we check the ID to see if we're loading a
     // previously saved search or if it is just transient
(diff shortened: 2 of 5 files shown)


src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -111,6 +112,15 @@ export const useSearch = (services: DiscoverViewServices) => {
     requests: new RequestAdapter(),
   };
 
+  const getDatasetAutoSearchOnPageLoadPreference = () => {
+    // Checks the searchOnpageLoadPreference for the current dataset if not specifed defaults to true
+    const datasetType = data.query.queryString.getQuery().dataset?.type;
+
+    const datasetService = data.query.queryString.getDatasetService();
+
+    return datasetType ? datasetService?.getType(datasetType)?.meta?.searchOnLoad ?? true : true;
```
amsiglan, line 121: Nit: `!datasetType || datasetService?.getType(datasetType)?.meta?.searchOnLoad ?? true`

LDrago27: Thanks. Fixed it.

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -125,10 +135,13 @@ export const useSearch = (services: DiscoverViewServices) => {
...
       new BehaviorSubject<SearchData>({
-        status: shouldSearchOnPageLoad() ? ResultStatus.LOADING : ResultStatus.UNINITIALIZED,
+        status:
+          shouldSearchOnPageLoad() && !skipInitialFetch.current
+            ? ResultStatus.LOADING
+            : ResultStatus.UNINITIALIZED,
         queryStatus: { startTime },
       }),
-    [shouldSearchOnPageLoad, startTime]
+    [shouldSearchOnPageLoad, startTime, skipInitialFetch]
```
amsiglan, line 144: should we check for `skipInitialFetch.current` instead?

src/plugins/data/public/query/query_string/dataset_service/types.ts
```diff
@@ -30,6 +30,8 @@ export interface DatasetTypeConfig {
     icon: EuiIconProps;
     /** Optional tooltip text */
     tooltip?: string;
+    /** Optional preference for search on page load else defaulted to true */
+    searchOnLoad?: boolean;
```
ananzh, line 34: nit: So you set this one as an optional property. Seems searchOnLoad property is added to some dataset types but not all? Seems we have searchOnLoad on index pattern type, index type and s3 type. what else types need this searchOnLoad that allows this property to be missing? To make it consistent, maybe remove ? because this `return !datasetType || (datasetService?.getType(datasetType)?.meta?.searchOnLoad ?? true` already set the default value to true right? so `?` is not necessary.

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -111,6 +112,15 @@ export const useSearch = (services: DiscoverViewServices) => {
     requests: new RequestAdapter(),
   };
 
+  const getDatasetAutoSearchOnPageLoadPreference = () => {
```
ananzh, line 115: I know it is a bit too much to ask. maybe we should add a test here given we have more and more logic in useSearch.

ananzh (approved):
I am okay with merging this fix if it is urgent. A few nit points and one question: When a search is not performed automatically due to this setting, consider adding a visual indicator to inform the user that they need to manually trigger the search given that most users are familiar with old auto search and might not know they need to click.

src/plugins/data/public/query/query_string/dataset_service/types.ts
ashwin-pc: nice callout. I think we should have different types for the config we provide and the config we receive from the service. Not a blocker for the PR though

src/plugins/discover/public/application/view_components/utils/use_search.ts
ashwin-pc: +1 we definitely need to add a test for this

src/plugins/discover/public/application/view_components/utils/use_search.ts
ashwin-pc: +1.

ashwin-pc (approved):
Can be a fast follow but lets add tests for the fix.

PR #8535, opened 2024-10-09
Implement correct normalization of locales in i18n

AMoo-Miki:
### Description

Implement correct normalization of locales in i18n

Also:
* Stop examining the fragment identifier for possible locale overrides
* Update relative formats and plural rules for certain locales

## Changelog
- feat: Implement correct normalization of locales in i18n
- feat: Update relative formats and plural rules for certain locales

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


packages/osd-i18n/src/core/i18n.ts
```diff
@@ -64,8 +74,16 @@ function getMessageById(id: string): string | undefined {
  * Normalizes locale to make it consistent with IntlMessageFormat locales
  * @param locale
  */
-function normalizeLocale(locale: string) {
-  return locale.toLowerCase();
+export function normalizeLocale(locale: string) {
+  const { lang, script, region } = localeParser.exec(locale)?.groups || {};
+  // If parsing failed or the language code was not extracted, return the locale
+  if (!lang) return locale;
```
virajsanghvi, line 80: Is the right thing to accept the bad input or default in this case?

src/core/server/rendering/rendering_service.tsx
```diff
@@ -100,6 +100,15 @@ export class RenderingService {
           opensearchDashboardsConfig as OpenSearchDashboardsConfigType
         );
 
+        let locale = i18n.getLocale();
```
virajsanghvi, line 103: Is this always a defaulted/configured valid locale?

src/legacy/ui/ui_render/ui_render_mixin.js
```diff
@@ -55,7 +55,8 @@ import { getApmConfig } from '../apm';
  */
 export function uiRenderMixin(osdServer, server, config) {
   const translationsCache = { translations: null, hash: null };
-  const defaultLocale = i18n.getLocale() || 'en'; // Fallback to 'en' if no default locale is set
+  // `i18n.getLocale` will always return a value; the 'en' is just to accommodate tests
+  const defaultLocale = i18n.getLocale() || 'en';
```
virajsanghvi, line 59: will this always be comparable to a normalized locale? (given toLowerCase was dropped)

virajsanghvi:
> Stop examining the fragment identifier for possible locale overrides

Was this used by anything

packages/osd-i18n/src/core/i18n.ts
AMoo-Miki: This guy only does the normalization; (in)validation is handled by comparing it against those that have been registered.

src/core/server/rendering/rendering_service.tsx
AMoo-Miki: defaulted: yes
valid: no

i18n starts with `en` as the default locale and if configured in the YAML config, it uses that without validating that it has been registered or not; I think this is because while reading the config, the filesystem might not have been read to prime the translation registration cache. At runtime, the locale is validation and if found to not have been registered, `en` is used.

src/legacy/ui/ui_render/ui_render_mixin.js
AMoo-Miki: Everyone now goes through `normalizeLocale`, as opposed to `.toLowerCase()` which is what the older normalize was doing.

src/core/server/rendering/rendering_service.tsx
virajsanghvi: From the code, if invalid, then does it potentially lead to a translation file not being loaded? Is this okay because english translations are loaded worst case?

ananzh (commented):
I saw this PR removes `getLocaleInUrl` then introduces a new way to handle locale specification in URLs by checking query parameter `const localeOverride = (request.query as any)?.locale?.trim?.();`

I think the previous way is to handle both standard query or non-standard one, given the two formats of current osd urls. For the URL format like (https://localhost:9200/app/home#/?locale=cn), where the locale is specified in the hash fragment rather than as a regular query parameter, it seems we can't handle. 

Does it mean we will not work on such url and request the standard query format if user need this feature? Should we add some documents somewhere for user?

AMoo-Miki:
> > Stop examining the fragment identifier for possible locale overrides
> 
> Was this used by anything

No. We haven’t started using any of the theme or locale overrides yet.

src/core/server/rendering/rendering_service.tsx
AMoo-Miki: If the locale is not registered, it would just use English, yes.

virajsanghvi (approved):

AMoo-Miki:
> I think the previous way is to handle both standard query or non-standard one, given the two formats of current osd urls. 
> 
> Does it mean we will not work on such url and request the standard query format if user need this feature? 

That is correct. We cannot parse and handle hash on the server-side and as such would rather not entertain it. A similar pattern is used for the theme. These two parameters are exclusively being introduced to temporarily override a setting and are not meant to be suffixed to every URL fragment. 

> Should we add some documents somewhere for user?

I don't believe these are to be part of our public API. If we decide to make them, we should.

ananzh (approved):

Issue #4055, opened 2023-05-17
[Enhancement] Global search on dashboard menus/objects


arunx2:
**Is your feature request related to a problem? Please describe.**

It is challenging to navigate through menus to find a page to perform a certain action. For ex. If I want to change roles, I need to navigate through Hamburger--> Security --> Roles. For accessing index patterns, Hamburger--> Stack management --> Index patterns.

**Describe the solution you'd like**

A global search bar on the top banner with a type ahead functionality. I should be able to type "roles" and select the link, that takes to "Roles".

**Additional context**

joshuarrrr:
Hi @arunx2 - do you think what you're asking for would be satisfied by https://github.com/opensearch-project/OpenSearch-Dashboards/issues/1854? If so, I think we can close this as a duplicate, and you can add any additional ideas or requests over there.

Issue #1854, opened 2022-07-06
Quick access menu


ashwin-pc:
**Is your feature request related to a problem? Please describe.**

Accessing the different sub-apps and pages within OSD easier e.g. Dasboards, Visualize, Discover and Dev Tools using keyboard shortcuts.

**Describe the solution you'd like**

Pressing a key command like `ctrl/cmd + k` opens up a floating search bard with suggestions on some of the frequently visited routes or views in Dashboards that a user can easily navigate to without having to click through menus or links. The initial goal can be just accessing the specific pages, but in future can extend to reaching specific visualizations or dashboards directly.

e.g. Githubs quick access tool

![image](https://user-images.githubusercontent.com/20453492/177638702-59e61f78-5d89-4aa7-85a6-ae47d6dafc7a.png)

**Describe alternatives you've considered**

Adding a search bar to the header that can also serve the same purpose. This is trickier since at space is in demand and already quite overused.

**Additional context**

PR #8538, opened 2024-10-10
[Workspace] Add global search bar into left nav

Hailong-am:
### Description

This PR is add search bar into left nav, the search bar supports search pages in and out of workspace.

saved objects is not supported yet.

https://github.com/user-attachments/assets/84fd06a5-a701-4ddb-baa3-1273c87061ac

### Issues Resolved

part of https://github.com/opensearch-project/OpenSearch-Dashboards/issues/1854
and https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4055

## Changelog

- feat: [Workspace] Add global search bar into left nav

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/core/public/chrome/global_search/global_search_service.test.ts
```diff
@@ -0,0 +1,47 @@
...
+        return [];
+      },
+    });
+
+    expect(start.getAllSearchStrategies()).toHaveLength(1);
+    expect(start.getAllSearchStrategies()[0].id).toEqual('test');
+    expect(start.getAllSearchStrategies()[0].type).toEqual(SearchObjectTypes.PAGES);
+  });
+
+  it('registerSearchStrategy with duplicate id', async () => {
```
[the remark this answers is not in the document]
Hailong-am: Yes, the expected behavior is not override existing one. Test case updated to check the content of search strategy not been override.

src/core/public/chrome/ui/header/collapsible_nav_group_enabled.tsx
```diff
@@ -27,6 +27,8 @@ import { ALL_USE_CASE_ID, DEFAULT_APP_CATEGORIES } from '../../../../../core/uti
 import { CollapsibleNavTop } from './collapsible_nav_group_enabled_top';
 import { HeaderNavControls } from './header_nav_controls';
 import { NavGroups } from './collapsible_nav_groups';
+import { HeaderSearchBar, HeaderSearchBarIcon } from './header_search_bar';
+import { GlobalSearchStrategy } from '../../global_search';
```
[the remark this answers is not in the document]
Hailong-am: update as suggested

src/core/public/chrome/ui/header/header_search_bar.tsx
```diff
@@ -0,0 +1,240 @@
...
+  );
+};
+
+export const HeaderSearchBar = ({ globalSearchStrategies, panel, onSearchResultClick }: Props) => {
+  const [results, setResults] = useState([] as React.JSX.Element[]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isPopoverOpen, setIsPopoverOpen] = useState(false);
+  const [panelWidth, setPanelWidth] = useState(0);
+  const inputRef = (node: HTMLElement | null) => setInputEl(node);
+  const [inputEl, setInputEl] = useState<HTMLElement | null>(null);
```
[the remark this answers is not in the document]
Hailong-am: This is by intention, inputEl is used as dependency for resize.

```ts
const onResize = useCallback(() => {
    if (inputEl) {
      const width = inputEl.getBoundingClientRect().width;
      setPanelWidth(width);
    }
  }, [inputEl, setPanelWidth]);
```

src/plugins/dev_tools/public/dev_tools_icon.tsx
```diff
@@ -35,6 +36,20 @@ export function DevToolsIcon({
   title: string;
 }) {
   const [modalVisible, setModalVisible] = useState(false);
+
+  const [devToolTab, setDevToolTab] = useState('');
+
+  const createOpenDevToolAction = createAction<typeof DEVTOOL_OPEN_ACTION>({
+    type: DEVTOOL_OPEN_ACTION,
+    getDisplayName: () => 'Open DevTools',
```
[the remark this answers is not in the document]
Hailong-am: this display name will not show on in UI as this is the only action registered.

src/core/public/chrome/global_search/global_search_service.ts
```diff
@@ -0,0 +1,84 @@
...
+import { ReactNode } from 'react';
+
+export enum SearchObjectTypes {
+  PAGES = 'pages',
+  SAVED_OBJECTS = 'saved_objects',
+}
+/**
+ * @experimental
+ */
+export interface GlobalSearchStrategy {
```
ashwin-pc, line 15: Can we not use search strategy as the name? We already have a concept of search strategy in the application. This is used for the actual search queries. Lets call these commands. Here is a really good extensible model that VS code uses for their global search. https://code.visualstudio.com/api/references/contribution-points#contributes.commands

src/core/public/chrome/global_search/global_search_service.ts
```diff
@@ -0,0 +1,84 @@
...
+   * search object type
+   * @type {SearchObjectTypes}
+   */
+  type: SearchObjectTypes;
+  /**
+   * do the search and return search result with a React element
+   * @param value search query
+   * @param callback callback function when search is done
+   */
+  doSearch(value: string, callback?: () => void): Promise<ReactNode[]>;
```
ashwin-pc, line 30: How about `run`. This works well with the term command too. So it will be `command.run(props)`

src/core/public/chrome/ui/header/header_search_bar.tsx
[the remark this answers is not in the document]
Hailong-am: Found that we can remove `onResize` as it will adjust the panel width when focus.

and changed to `const inputElRef = useRef<HTMLElement | null>(null);` as well.

src/core/public/chrome/global_search/global_search_service.ts
Hailong-am: rename it to searchCommand

src/core/public/chrome/global_search/global_search_service.ts
Hailong-am: rename this method name to `run` as well

Hailong-am:
> So here's what i had in mind for that project (And we can impliment a subset of that today for the sake of time)
> 
> 1. We have a global command service
> 2. We also have a global search bar
> 3. The default behaviour of the search bar is to search for applications/pages
> 4. if the @ symbol is specified we search for saved objects
> 5. If the > is specified, we show the available commands using the command service
> 
> Plugins can only register commands. Pages and saved objects are used by default but we dont expose this
> 
> This is very similar to how VS code does it and its really elagant.
> 
> cc: @kgcreative
> 
> If we dont want to go that far right now, lets keep the search bar simple and not expose any API's. This is not something we need to rush and will not be easy to walk back on in the future :)

Thanks for sharing your thought on this, 
1. `saved objects` search will come with next PR
2. we can have some examples/use cases of commands after that we can have it implemented

ruanyl:
@Hailong-am From the screen record you share, if seem the highlighted text color is black and the rest is blue, while the blue text looks more like highlighted text to me, is this by design?

Hailong-am:
> @Hailong-am From the screen record you share, it seems the highlighted text color is black and the rest are blue, while the blue text looks more like highlighted text to me, is this by design?

That's OUI default highlight behavior and don't provide a way to customized it. Confirmed with @kgcreative we don't do customization to the default OUI behavior.

Reference: https://oui.opensearch.org/1.16/#/utilities/highlight-and-mark

src/core/public/chrome/global_search/global_search_service.ts
```diff
@@ -0,0 +1,85 @@
...
+export interface GlobalSearchCommand {
+  /**
+   * unique id of this strategy
+   */
+  id: string;
+  /**
+   * search object type
+   * @type {SearchObjectTypes}
+   */
+  type: SearchObjectTypes;
```
SuZhou-Joe, line 25: Seems the type here is useless, all the search / navigation logic is within `run` callback actually.

src/core/public/chrome/ui/header/collapsible_nav_group_enabled.scss
```diff
@@ -114,4 +114,25 @@
...
   }
+
+  .searchBar-wrapper {
+    padding: $euiSize;
+    padding-top: $euiSizeS;
+    background-color: transparent;
+    flex-grow: 0;
+
+    .searchInput {
+      background-color: transparentize(white, 0.5);
```
SuZhou-Joe, line 125: Should we use some variable from theme? Will `white` work in dark mode?

src/core/public/chrome/ui/header/header_search_bar.tsx
```diff
@@ -0,0 +1,259 @@
...
+  panel?: boolean;
+  onSearchResultClick?: () => void;
+}
+
+/**
+ * search input match with `@` will handled by saved objects search command
+ * search input match with `>` will handled by plugin customized commands
+ */
+export const SAVED_OBJECTS_SYMBOL = '@';
+export const COMMANDS_SYMBOL = '>';
```
SuZhou-Joe, line 34: It looks a little bit weird to me. Why don't we just aggregate all the results from `globalSearchCommands`?

src/core/public/chrome/ui/header/collapsible_nav_group_enabled.scss
Hailong-am: good catch, dark mode looks not correct with white, updated to variable `$ouiFormBackgroundColor`

src/core/public/chrome/ui/header/header_search_bar.tsx
Hailong-am: That's to filter what kind of searchCommand we want to run. If user input is start with `@` like  `@flights` it will only search for saved objects title has flights.

SuZhou-Joe: Oh, that's a big vision in the future but seems weird that we implement in this PR. I think these should be done in an enhancement PR if we really have the use case.

SuZhou-Joe (approved):
Looks amazing to me and finally we get a global search function in left nav.

src/plugins/workspace/public/components/global_search/page_item.tsx
```diff
@@ -0,0 +1,96 @@
...
+          workspace={currentWorkspace}
+          availableUseCases={availableUseCases || []}
+        />
+      ),
+    });
+  } else {
+    breadcrumbs.push({ text: link.navGroup.title });
+  }
+  const appId = link.id.toLowerCase();
+  const isLanding = appId.endsWith('landing');
```
ruanyl, line 53: Q: Which page ids are ended with `landing`?

src/plugins/workspace/public/components/global_search/search_pages_command.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { ChromeNavLink, ChromeRegistrationNavLink, CoreStart } from 'opensearch-dashboards/public';
+import { first } from 'rxjs/operators';
+import React, { ReactNode } from 'react';
+import { BehaviorSubject } from 'rxjs';
+import { getFirstUseCaseOfFeatureConfigs } from '../../utils';
+import { GlobalSearchPageItem } from './page_item';
+import { DEFAULT_NAV_GROUPS, NavGroupType } from '../../../../../core/public';
+import { WorkspaceUseCase } from '../../types';
+
+export const searchPageWithInWorkspace = async (
```
ruanyl, line 15:
```suggestion
export const searchPageWithinWorkspace = async (
```

src/plugins/workspace/public/components/global_search/search_pages_command.tsx
```diff
@@ -0,0 +1,78 @@
...
+): Promise<ReactNode[]> => {
+  if (coreStart) {
+    const currentWorkspace = await coreStart.workspaces.currentWorkspace$.pipe(first()).toPromise();
+
+    const navGroupMap = await coreStart.chrome.navGroup
+      .getNavGroupsMap$()
+      .pipe(first())
+      .toPromise();
+
+    const allAvailableCaseId = Object.entries(DEFAULT_NAV_GROUPS)
```
ruanyl, line 29: So these are global nav item, it will be searchable even when outside of a workspace, if that's the case, is the function name `searchPageWithInWorkspace` still appropriate?

src/plugins/dev_tools/public/dev_tools_icon.tsx
```diff
@@ -35,6 +36,20 @@ export function DevToolsIcon({
   title: string;
 }) {
   const [modalVisible, setModalVisible] = useState(false);
+
+  const [devToolTab, setDevToolTab] = useState('');
+
+  const createOpenDevToolAction = createAction<typeof DEVTOOL_OPEN_ACTION>({
```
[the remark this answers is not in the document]
Hailong-am: > Just sharing my thought, instead of implementing it with Action/Trigger, I'd simply use a global state for the `modalVisible` and `devToolTab`, perhaps just a `BahaviorSubject`.

sounds like Observable is more lightweight, haven't got a chance to try it.

src/plugins/workspace/public/components/global_search/page_item.tsx
Hailong-am: settings and data administration landing page, those two.

src/plugins/workspace/public/components/global_search/search_pages_command.tsx
Hailong-am: make sense, update the method name to `searchPages` so that it will not limit to pages in workspace.

src/core/public/chrome/ui/header/header_search_bar.tsx
```diff
@@ -0,0 +1,259 @@
...
+    setIsPopoverOpen(false);
+  };
+
+  const pagesSection = (items: ReactNode[]) => {
+    return (
+      <EuiFlexGroup direction="column" gutterSize="xs">
+        <EuiFlexItem>
+          <EuiTitle size="s">
+            <EuiText size="xs" color="subdued">
+              {i18n.translate('core.globalSearch.pageSection.title', { defaultMessage: 'Pages' })}
```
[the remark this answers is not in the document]
Hailong-am: > This is again just for discussion.
> 
> It seems we need a specific `pagesSection` based on the `type` just to render the title, can we make it so that it's part of the command interface? Like:
> 
> ```
> {
>   id: 'workspacePagesSearchCommand',
>   category: 'Pages',
> }
> ```
> 
> Then for all the items returns by this command, we display it under the `Pages` title. Seems then we don't need `type: SearchObjectTypes;`

Yeah, the section render could be a general component/method, the only difference is the section header. Refactor it to accept a section header props to make it reusable.

src/core/public/chrome/ui/header/header_search_bar.tsx
```diff
@@ -0,0 +1,259 @@
...
+      searchValue: value,
+    };
+  },
+  [SearchObjectTypes.SAVED_OBJECTS]: (value: string) => {
+    return {
+      match: value.startsWith(SAVED_OBJECTS_SYMBOL),
+      searchValue: value.replace(SAVED_OBJECTS_SYMBOL, '').trim(),
+    };
+  },
+};
```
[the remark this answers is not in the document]
Hailong-am: That's make sense, to let each command decide what's the behavior. But for builtin pages and saved objects search command, we don't want to provide the flexibility to customized the alias, name and behavior of whether to run the command or not.

src/core/public/chrome/global_search/global_search_service.ts
```diff
@@ -0,0 +1,85 @@
...
+  PAGES = 'pages',
+  SAVED_OBJECTS = 'saved_objects',
+}
+
+/**
+ * @experimental
+ */
+export interface GlobalSearchCommand {
+  /**
+   * unique id of this strategy
```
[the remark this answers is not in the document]
Hailong-am: sure, updated

src/core/public/chrome/ui/header/header_search_bar.tsx
Hailong-am: Agree these should come with future enhancement PR, these are built in symbols for saved objects and commands, so i put it here for in advance.

regarding to prefix registration, that's might be long term plan if we plan to support more command types.

src/core/public/chrome/ui/header/header_search_bar.tsx
```diff
@@ -0,0 +1,259 @@
...
+  const searchResultSections = results && (
+    <EuiFlexGroup direction="column" gutterSize="none">
+      {results.map((result) => (
+        <EuiFlexItem key={result.key}>{result}</EuiFlexItem>
+      ))}
+    </EuiFlexGroup>
+  );
+
+  const onSearch = async (value: string) => {
+    // do page search
```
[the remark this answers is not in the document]
Hailong-am: It's not needed anymore, remove it.

src/core/public/chrome/ui/header/header_search_bar.tsx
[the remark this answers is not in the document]
Hailong-am: >If the input not start with any command name/alias, the it should call run for all commands with the current input. 

I think we should search pages only, to search other types it will need explicit type the alias

src/core/public/chrome/ui/header/header_search_bar.tsx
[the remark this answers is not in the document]
Hailong-am: I combine category into type and change type to a object like this 

```ts
export const SearchCommandTypes = {
  PAGES: {
    description: i18n.translate('core.globalSearch.pages.description', { defaultMessage: 'Pages' }),
    alias: null,
  },
  SAVED_OBJECTS: {
    description: i18n.translate('core.globalSearch.assets.description', {
      defaultMessage: 'Assets',
    }),
    alias: SAVED_OBJECTS_SYMBOL,
  },
} as const;
```

src/plugins/dev_tools/public/dev_tools_icon.tsx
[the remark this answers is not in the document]
Hailong-am: By comparing uiActions and Observable, they are both able to handle events. the uiActions is a central place to manage events and could have context object associate with. It's more powerful and extensible. I could prefer keep the current implementation.

SuZhou-Joe (approved):

src/core/public/chrome/ui/header/header_search_bar.tsx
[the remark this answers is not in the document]
ruanyl: I think header_search_bar should better not maintain `page` or `save_objects`, because in the future, there might be `alerts` and `anomaly_detectors`, these are registered by plugins, which core should NOT be aware of, so the `description` and `alias` should be better registered by the plugin. What do you think?

ruanyl (approved):

src/core/public/chrome/ui/header/header_search_bar.tsx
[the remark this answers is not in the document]
Hailong-am: Let plugin to register it makes more sense. we could add `description` and `alias` to `GlobalSearchCommand` interface that will take precedence. And `SearchCommandTypes` serve as the builtin for pages and assets.

PR #8541, opened 2024-10-10
[workspace]fix the workspace selector jump that occurs when hovering and clicking.

Qxisylolo:
### Description

This pr fixes the workspace selector jump that occurs when hovering and clicking.

## Screenshot
Before:

https://github.com/user-attachments/assets/ac72b982-87f1-4fd6-885b-d8c242bb07bd

After:

https://github.com/user-attachments/assets/e3aecc24-9d0b-4934-a8d4-9846f0c83952

## Changelog
- fix: the workspace selector jump that occurs when hovering and clicking.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/public/components/workspace_selector/workspace_selector.tsx
+++ b/src/plugins/workspace/public/components/workspace_selector/workspace_selector.tsx
@@ -66,6 +66,7 @@ export const WorkspaceSelector = ({ coreStart, registeredUseCases$ }: Props) => 
   };
   const button = currentWorkspace ? (
     <EuiPanel
+      className="popoverButton"
       paddingSize="none"
       color="transparent"
       hasBorder={false}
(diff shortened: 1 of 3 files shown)


src/plugins/workspace/public/components/workspace_selector/workspace_selector.tsx
```diff
@@ -66,6 +66,7 @@ export const WorkspaceSelector = ({ coreStart, registeredUseCases$ }: Props) =>
   };
   const button = currentWorkspace ? (
     <EuiPanel
+      className="popoverButton"
```
SuZhou-Joe, line 69:
```suggestion
      className="workspaceSelectorPopoverButton"
```
Adding some prefix makes me feel safer, can we?

Qxisylolo: Thanks for the advice, always updated.

SuZhou-Joe (dismissed):

SuZhou-Joe (approved):

Hailong-am (approved):
need confirm whether we need the shadow when hover

ruanyl:
The animation is the default behavior of a clickable OuiPanel, instead of override the styles from component library, maybe we shouldn't use `OuiPanel` for this component

ruanyl (approved):

PR #8565, opened 2024-10-11
[Discover] Add Footer Bar for Single Line Editor

sejli:
### Description

* Adds footer bar in multiline editor to single line editor when editor is focused
* Fixes query result component not displaying errors
* Correctly surfaces errors from async search strategy

## Screenshot

image

image

image

image

image

## Changelog

- feat: Adds editor footer to single line editor on focus

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/common/utils/helpers.ts
+++ b/src/plugins/data/common/utils/helpers.ts
@@ -29,7 +29,12 @@
  */
 
 import { i18n } from '@osd/i18n';
-import { PollQueryResultsHandler, FetchStatusResponse } from '../data_frames';
+import {
+  PollQueryResultsHandler,
+  FetchStatusResponse,
+  IDataFrameResponse,
+  DATA_FRAME_TYPES,
+} from '../data_frames';
 
 export interface QueryStatusOptions {
   pollQueryResults: PollQueryResultsHandler;
@@ -53,11 +58,15 @@ export const handleQueryResults = async <T>(
   } while (queryStatus !== 'SUCCESS' && queryStatus !== 'FAILED');
 
   if (queryStatus === 'FAILED') {
-    throw new Error(
+    const error = Error(
       i18n.translate('data.search.request.failed', {
         defaultMessage: 'An error occurred while executing the search query',
       })
     );
+    return {
+      status: 'failed',
+      body: { error },
+    } as FetchStatusResponse;
   }
 
   return queryResultsRes;
--- a/src/plugins/data/common/utils/helpers.ts
+++ b/src/plugins/data/common/utils/helpers.ts
@@ -28,13 +28,7 @@
  * under the License.
  */
 
-import { i18n } from '@osd/i18n';
-import {
-  PollQueryResultsHandler,
-  FetchStatusResponse,
-  IDataFrameResponse,
-  DATA_FRAME_TYPES,
-} from '../data_frames';
+import { PollQueryResultsHandler, FetchStatusResponse } from '../data_frames';
 
 export interface QueryStatusOptions {
   pollQueryResults: PollQueryResultsHandler;
@@ -58,15 +52,7 @@ export const handleQueryResults = async <T>(
   } while (queryStatus !== 'SUCCESS' && queryStatus !== 'FAILED');
 
   if (queryStatus === 'FAILED') {
-    const error = Error(
-      i18n.translate('data.search.request.failed', {
-        defaultMessage: 'An error occurred while executing the search query',
-      })
-    );
-    return {
-      status: 'failed',
-      body: { error },
-    } as FetchStatusResponse;
+    throw new Error(queryResultsRes?.body.error);
   }
 
   return queryResultsRes;
--- a/src/plugins/data/public/query/query_string/language_service/lib/query_result.tsx
+++ b/src/plugins/data/public/query/query_string/language_service/lib/query_result.tsx
@@ -13,7 +13,6 @@ import React, { useEffect, useState } from 'react';
 export enum ResultStatus {
   UNINITIALIZED = 'uninitialized',
   LOADING = 'loading', // initial data load
-  LOADING_MORE = 'loading_more', // loading additional data while existing results are presents
   READY = 'ready', // results came back
   NO_RESULTS = 'none', // no results came back
   ERROR = 'error', // error occurred
@@ -59,12 +58,23 @@ export function QueryResult(props: { queryStatus: QueryStatus }) {
     };
   }, [props.queryStatus.startTime]);
 
-  if (
-    props.queryStatus.status === ResultStatus.LOADING ||
-    props.queryStatus.status === ResultStatus.LOADING_MORE
-  ) {
-    if (elapsedTime < BUFFER_TIME) {
-      return null;
+  if (elapsedTime > BUFFER_TIME) {
+    if (props.queryStatus.status === ResultStatus.LOADING) {
+      const time = Math.floor(elapsedTime / 1000);
+      return (
+        <EuiButtonEmpty
+          color="text"
+          size="xs"
+          onClick={() => {}}
+          isLoading
+          data-test-subj="queryResultLoading"
+        >
+          {i18n.translate('data.query.languageService.queryResults.loadTime', {
+            defaultMessage: 'Loading {time} s',
+            values: { time },
+          })}
+        </EuiButtonEmpty>
+      );
     }
     const time = Math.floor(elapsedTime / 1000);
     return (
--- a/src/plugins/data/public/query/query_string/language_service/lib/query_result.tsx
+++ b/src/plugins/data/public/query/query_string/language_service/lib/query_result.tsx
@@ -160,15 +160,7 @@ export function QueryResult(props: { queryStatus: QueryStatus }) {
       data-test-subj="queryResultError"
     >
       <EuiPopoverTitle>ERRORS</EuiPopoverTitle>
-      <div style={{ width: '250px' }}>
-        <EuiText size="s">
-          <strong>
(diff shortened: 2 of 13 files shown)


src/plugins/data/public/ui/query_editor/_query_editor.scss
```diff
@@ -215,3 +215,40 @@
     display: block;
   }
 }
+
+.queryEditor__footer {
```
LDrago27, line 219: For the values below can we use the constants from oui  like `$ouiSizeM`.

ashwin-pc (dismissed):

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -259,9 +259,9 @@ export const useSearch = (services: DiscoverViewServices) => {
       }
       let errorBody;
       try {
-        errorBody = JSON.parse(error.message);
+        errorBody = JSON.parse(error.body.message);
       } catch (e) {
-        errorBody = error.message;
+        errorBody = error.body.message;
```
ashwin-pc, line 264: this broke the app when i tried it with an S3 datasource

ashwin-pc (commented):
Forgot one comment

src/plugins/data/public/query/query_string/language_service/lib/query_result.tsx
```diff
@@ -99,7 +104,13 @@ export function QueryResult(props: { queryStatus: QueryStatus }) {
     }
 
     return (
-      <EuiButtonEmpty iconSide="left" iconType={'checkInCircleEmpty'} size="xs" onClick={() => {}}>
+      <EuiButtonEmpty
+        iconSide="left"
+        iconType={'checkInCircleEmpty'}
+        iconGap="s"
+        size="xs"
+        onClick={() => {}}
```
virajsanghvi, line 112: why do we have a button that does nothing? is this the right ui element?

src/plugins/data/public/ui/query_editor/_query_editor.scss
virajsanghvi: +1

src/plugins/data/public/ui/query_editor/_query_editor.scss
```diff
@@ -215,3 +215,40 @@
...
+
+// TODO: Temporary workaround to disable padding for single line editor footer
+.euiFormControlLayout--group.euiFormControlLayout--compressed .osdQuerEditor__singleLine .euiText {
+  padding-top: 0 !important;
+  padding-bottom: 0 !important;
+}
+
+.euiFormControlLayout--group .osdQuerEditor__singleLine .euiText {
+  background-color: unset !important;
+  line-height: 21px !important;
```
virajsanghvi, line 249: if we do something custom to override default behavior, can we include a comment with the rationale?

src/plugins/data/public/ui/query_editor/editors/shared.tsx
```diff
@@ -61,56 +62,106 @@ export const SingleLineInput: React.FC<SingleLineInputProps> = ({
...
+        if (blurTimeoutRef.current) {
+          clearTimeout(blurTimeoutRef.current);
+        }
+      };
+    },
+    [editorDidMount]
+  );
+
+  return (
+    <div className="euiFormControlLayout euiFormControlLayout--compressed euiFormControlLayout--group osdQueryBar__wrap">
```
virajsanghvi, line 99: qq: is there a reason to use the classes instead of components?

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -430,6 +431,34 @@ export default class QueryEditorUI extends Component<Props, State> {
...
+          <EuiButtonEmpty
+            iconSide="left"
+            iconType="clock"
+            iconGap="s"
+            size="xs"
+            onClick={this.toggleRecentQueries}
+            className="queryEditor__footerItem"
+            flush="both"
+          >
+            <EuiText size="xs" color="subdued">
```
virajsanghvi, line 456: is there a reason we're styling the text over the button itself?

virajsanghvi:
just fyi, there's a test that's failing on both linux and windows

src/plugins/data/public/ui/query_editor/query_editor.tsx
AMoo-Miki: Button's sizes are not in sync with text size; an `xs` button uses `s` text and there isn't a size below it. Additionally, button doesn't have a `subdued` color.

src/plugins/data/public/ui/query_editor/_query_editor.scss
AMoo-Miki: OUI's flexboxes are messed up; they simulate `gap` by adding margins on children and negative margins on parent and this occasionally brings in unpleasant side effects.

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -272,15 +276,15 @@ export const useSearch = (services: DiscoverViewServices) => {
       }
       let errorBody;
       try {
-        errorBody = JSON.parse(error.message);
+        errorBody = JSON.parse(error.body);
       } catch (e) {
-        errorBody = error.message;
+        errorBody = error.body;
```
joshuali925, line 281: how do you know that there is `error.body`? and does it follow type correctly? https://github.com/opensearch-project/OpenSearch-Dashboards/blob/7da544372cd24c80eaf286a143d4c844c6087c60/src/plugins/discover/public/application/view_components/utils/use_search.ts#L55-L59

src/plugins/data/common/utils/helpers.ts
```diff
@@ -53,11 +58,15 @@ export const handleQueryResults = async <T>(
...
   if (queryStatus === 'FAILED') {
-    throw new Error(
+    const error = Error(
       i18n.translate('data.search.request.failed', {
         defaultMessage: 'An error occurred while executing the search query',
       })
     );
+    return {
+      status: 'failed',
+      body: { error },
```
joshuali925, line 68: the returned error only contains the hard coded message "An error occurred while executing the search query", should we return the actual backend error to tell users what the error was?

ashwin-pc:
@sejli can you call out all that your PR is fixing here.

src/plugins/data/public/ui/query_editor/query_editor.tsx
virajsanghvi: I guess my question is why we are creating designs that oui doesn't support, but that is not blocking. If this is to try to match mocks, that's fine.

src/plugins/data/public/query/query_string/language_service/lib/query_result.tsx
```diff
@@ -22,10 +22,9 @@ export interface QueryStatus {
   status: ResultStatus;
   body?: {
     error?: {
-      reason?: string;
-      details: string;
+      statusCode?: number;
```
ashwin-pc, line 25: why do we have this here? We dont want to show the user the status code, but rather the error and the reason

src/plugins/query_enhancements/server/routes/index.ts
```diff
@@ -85,8 +85,14 @@ export function defineSearchStrategyRouteProvider(logger: Logger, router: IRoute
...
         } catch (err) {
+          let error;
+          try {
+            error = JSON.parse(err.message);
+          } catch (e) {
+            error = err;
+          }
           return res.custom({
-            statusCode: err.name,
+            statusCode: error.status,
```
ashwin-pc, line 95: doesnt this mean we will return a 400 error instead of 200 when the job fails? Shouldnt we be returning a 200 here?

sejli: Since the search strategy doesn't throw an error, the request won't reach this catch. Instead, it'll just be returned by the `res.ok({ body: { ...queryRes } });`.

src/plugins/data/public/ui/query_editor/query_editor.tsx
ashwin-pc: It is to match mocks for now

ashwin-pc (approved):
The PR is good, has some follow up work though. Lets follow up on how we display the error. I think for now this is better than what we have, but we need to fix:
1. Error when the editor is not focussed
2. How the error is displayed
3. Loading state when the query is run after the initial load

src/plugins/query_enhancements/server/routes/index.ts
virajsanghvi, line 95: all error types above won't necessarily have status, right? Is that okay?

virajsanghvi (approved):
Approving. There are some open questions, but none are blocking

src/plugins/query_enhancements/server/routes/index.ts
```diff
@@ -85,8 +85,14 @@ export function defineSearchStrategyRouteProvider(logger: Logger, router: IRoute
           const queryRes: IDataFrameResponse = await searchStrategy.search(context, req as any, {});
           return res.ok({ body: { ...queryRes } });
         } catch (err) {
+          let error;
+          try {
+            error = JSON.parse(err.message);
```
joshuali925, line 90: the status code might be in err object, not in err.message. If err.message is a valid string, then status code is lost

src/plugins/query_enhancements/server/routes/index.ts
joshuali925, line 95: where is `error.status` coming from?

PR #8570, opened 2024-10-12
[Workspace] Fix workspace update issue

yubonluo:
### Description

1. After updating workspace detail, should not show the navigate away modal.
2. After updating Collaborators, switch to other tab should not show the navigate away modal

https://github.com/user-attachments/assets/9dd0f262-07ca-43ff-bb8b-59a7b26a58c8

## Screenshot

https://github.com/user-attachments/assets/e0fc5895-e8ec-4ad4-a068-0a3f36aeab9a

## Changelog

- fix: Fix workspace update issue
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


SuZhou-Joe:
It looks a little bit weird to me that we are refreshing the page after update the workspace detail, I think we already have the refresh logic in workspace client when calling `update` interface and the only thing we need to do is to update the state of the detail panel component.

wanglam:
> It looks a little bit weird to me that we are refreshing the page after update the workspace detail, I remember we already have the refresh logic for `currentWorkspace$` and `workspaces$` in workspace client when calling `update` interface and the only thing we need to do is to update the state of the detail panel component. By only updating the state, we won't have the issue of the `navigate away modal` for workspace detail tab at least.

Agree with you. I think the navigate away modal issue should come from default values not updated. I'm prefer to add some method to update the defaultValuesRef in useWorkspaceForm hook.

yubonluo:
> > It looks a little bit weird to me that we are refreshing the page after update the workspace detail, I remember we already have the refresh logic for `currentWorkspace$` and `workspaces$` in workspace client when calling `update` interface and the only thing we need to do is to update the state of the detail panel component. By only updating the state, we won't have the issue of the `navigate away modal` for workspace detail tab at least.
> 
> Agree with you. I think the navigate away modal issue should come from default values not updated. I'm prefer to add some method to update the defaultValuesRef in useWorkspaceForm hook.

Sure, the code has been updated. When user updates the workspace, the detail page will not be refreshed and `defaultValuesRef.current` will be updated as `currentFormData`

src/plugins/workspace/public/components/workspace_form/use_workspace_form.ts
```diff
@@ -107,34 +107,34 @@ export const useWorkspaceForm = ({
...
-        true
-      );
+      const submitFormData = {
+        name: currentFormData.name!,
+        description: currentFormData.description,
+        color: currentFormData.color || '#FFFFFF',
+        features: currentFormData.features,
+        permissionSettings: currentFormData.permissionSettings as WorkspacePermissionSetting[],
+        selectedDataSourceConnections: currentFormData.selectedDataSourceConnections,
+      };
```
SuZhou-Joe, line 117: Nit: we can have a unified function like `getSubmitFormData` to generate the data payload for submit, it can be used here and line 130.

yubonluo: Thanks, the code has been updated

src/plugins/workspace/public/components/workspace_form/use_workspace_form.ts
```diff
@@ -107,34 +123,19 @@ export const useWorkspaceForm = ({
...
-        name: currentFormData.name!,
-        description: currentFormData.description,
-        color: currentFormData.color || '#FFFFFF',
-        features: currentFormData.features,
-        permissionSettings: settings,
-        selectedDataSourceConnections: currentFormData.selectedDataSourceConnections,
-      },
-      false
-    );
+    const result = await onSubmit?.(getSubmitFormData(currentFormData, settings));
```
SuZhou-Joe, line 138:
```suggestion
    const result = await onSubmit?.({ ...getSubmitFormData(currentFormData), permissionSettings: settings });
```
Should be good enough to use spread and override.

SuZhou-Joe (dismissed):

src/plugins/workspace/public/components/workspace_form/use_workspace_form.ts
yubonluo: Thanks for your suggestion. The code has been updated

src/plugins/workspace/public/components/workspace_form/use_workspace_form.ts
```diff
@@ -89,8 +89,19 @@ export const useWorkspaceForm = ({
...
+  const getSubmitFormData = (submitFormData: WorkspaceFormDataState) => {
+    return {
+      name: submitFormData.name!,
+      description: submitFormData.description,
+      color: submitFormData.color || '#FFFFFF',
+      features: submitFormData.features,
+      permissionSettings: submitFormData.permissionSettings as WorkspacePermissionSetting[],
+      selectedDataSourceConnections: submitFormData.selectedDataSourceConnections,
+    };
+  };
```
Hailong-am, line 101: looks like the return value is same as WorkspaceFormDataState? can we use `WorkspaceFormDataState` diectly? or `{...WorkspaceFormDataState}`

Hailong-am (approved):

src/plugins/workspace/public/components/workspace_form/use_workspace_form.ts
yubonluo: No, there are some differences between the type of return values and WorkspaceFormDataState, such as the useCase attribute.

SuZhou-Joe (approved):

PR #8575, opened 2024-10-12
fix: link check failure

SuZhou-Joe:
### Description

Fix link check for `helpmenow.com` is down.

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ruanyl (approved):
I think we've encountered similar issue multiple times when a domain become inaccessible due to various reasons.

Shall we have a convention to use only, for example, `example.com` in all unit tests which need a dummy url. And add `example.com` to `.lycheeignore`, then no needs to bother link checker.

@ashwin-pc @AMoo-Miki @virajsanghvi what do you think?

PR #8581, opened 2024-10-14
[Workspace] Disable copy all button when there is no saved objects in assets page 

yubonluo:
### Description

Page crashed after click "Copy assets to ..." button in workspace assets page when assets table is empty
image

image (5)

## Screenshot

image

## Changelog

- fix: Disable cope all button when there is no saved objects in assets page 
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/saved_objects_management/public/management_section/objects_table/components/duplicate_result_flyout.tsx
+++ b/src/plugins/saved_objects_management/public/management_section/objects_table/components/duplicate_result_flyout.tsx
@@ -91,7 +91,7 @@ export class DuplicateResultFlyout extends React.Component<DuplicateResultFlyout
         {copiedCount && (
           <EuiFlexItem grow={false}>
             <EuiTitle size="xs">
-              <h4 className="savedObjectsManagementImportSummary__copiedCountCount">
+              <h4 className="savedObjectsManagementImportSummary__createdCount">
                 <FormattedMessage
                   id="savedObjectsManagement.copyResult.copiedCountHeader"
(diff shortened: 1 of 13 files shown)


SuZhou-Joe (approved):

SuZhou-Joe:
It seems we only disable the button instead of fixing the root cause. Can we fix the place giving the error?

src/plugins/saved_objects_management/public/management_section/objects_table/components/duplicate_result_flyout.tsx
```diff
@@ -91,7 +91,7 @@ export class DuplicateResultFlyout extends React.Component<DuplicateResultFlyout
         {copiedCount && (
           <EuiFlexItem grow={false}>
             <EuiTitle size="xs">
-              <h4 className="savedObjectsManagementImportSummary__copiedCountCount">
+              <h4 className="savedObjectsManagementImportSummary__createdCount">
```
wanglam, line 94: Nit: This change does not seem related to the copy issue, was it made intentionally?

wanglam (approved):
Great fix. After testing, it looks good on my local environment.

yubonluo:
> It seems we only disable the button instead of fixing the root cause. Can we fix the place giving the error?

Sure, the root cause is when selectedObjectCount is 0, we can not get the `objectName: allSelectedObjects[0].meta.title,`. After updating the code, it should be `objectName: allSelectedObjects[0]?.meta.title,`

Hailong-am (approved):

PR #8632, opened 2024-10-17
[discover-new] fix spelling for localstorage key name

hanzhinstas:
### Description
Key name was introduced with typo

## Testing the changes
https://dictionary.cambridge.org/dictionary/english/experience

## Changelog
- skip

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/components/utils/local_storage.ts
+++ b/src/plugins/discover/public/application/components/utils/local_storage.ts
@@ -5,7 +5,7 @@
 
 import { Storage } from '../../../../../opensearch_dashboards_utils/public';
 
-export const NEW_DISCOVER_KEY = 'discover:newExpereince';
+export const NEW_DISCOVER_KEY = 'discover:newExperience';
 
 export const getNewDiscoverSetting = (storage: Storage): boolean => {
   const storedValue = storage.get(NEW_DISCOVER_KEY);


LDrago27 (approved):

src/plugins/discover/public/application/components/utils/local_storage.ts
```diff
@@ -5,7 +5,7 @@
 
 import { Storage } from '../../../../../opensearch_dashboards_utils/public';
 
-export const NEW_DISCOVER_KEY = 'discover:newExpereince';
+export const NEW_DISCOVER_KEY = 'discover:newExperience';
```
kavilla, line 8: Yeah this always stuck out like a red thumb for me. One risk to call out here is that previous users who set this will need to reset this. Should we call this out in a changelog?

hanzhinstas: Something like that?

> fixed typo in option name that disabled legacy Discover UI. Those using new Discover have to re-enable it once again

ashwin-pc: yep that works

angle943:
This is Kiro posting.

Small spelling fix with merge conflicts. @hanzhinstas — please rebase if still interested, otherwise this can be closed.

PR #8641, opened 2024-10-18
[Workspace]Response forbidden error for not permitted workspace

wanglam:
### Description

This PR is for fixing 500 error when import sample data to a not permitted workspace. It should return a forbidden error.

## Screenshot

image
image

## Testing the changes

- Clone branch code and run `yarn osd bootstrap --single-version loose`
```
workspace.enabled: true
uiSettings:
  overrides:
    'home:useNewHomePage': true
opensearchDashboards.dashboardAdmin.users: ['admin']
```
- Login with admin user
- Visit internal users page and create a test-user
- Visit a existing workspace settings page and add test-user as a read-only collaborator
- Open an incognito window and login with test-user
- Goto the sample data page of the same workspace
- Click "Add data" button, it should show error like screenshot below
- Should see "403" response in devtools
 
## Changelog

- feat: [Workspace] Response forbidden error for not permitted workspace

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/home/server/services/sample_data/routes/install.ts
+++ b/src/plugins/home/server/services/sample_data/routes/install.ts
@@ -31,6 +31,7 @@
 import { schema } from '@osd/config-schema';
 import { IRouter, LegacyCallAPIOptions, Logger } from 'src/core/server';
 import { getWorkspaceState } from '../../../../../../core/server/utils';
+import { SavedObjectsErrorHelpers } from '../../../../../../core/server';
 import { SampleDatasetSchema } from '../lib/sample_dataset_registry_types';
 import { createIndexName } from '../lib/create_index_name';
 import {
@@ -217,6 +218,9 @@ export function createInstallRoute(
       } catch (err) {
         const errMsg = `bulkCreate failed, error: ${err.message}`;
         logger.warn(errMsg);
+        if (workspaceId && SavedObjectsErrorHelpers.isForbiddenError(err)) {
+          return res.forbidden({ body: errMsg });
+        }
         return res.internalError({ body: errMsg });
       }
       const errors = createResults.saved_objects.filter((savedObjectCreateResult) => {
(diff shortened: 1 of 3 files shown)


Hailong-am (approved):

src/plugins/home/server/services/sample_data/routes/install.ts
```diff
@@ -217,6 +218,9 @@ export function createInstallRoute(
       } catch (err) {
         const errMsg = `bulkCreate failed, error: ${err.message}`;
         logger.warn(errMsg);
+        if (workspaceId && SavedObjectsErrorHelpers.isForbiddenError(err)) {
```
Hailong-am, line 221: do we have any other error except forbidden error?

wanglam: I've searched the whole org, there are [three places](https://github.com/search?q=repo%3Aopensearch-project%2FOpenSearch-Dashboards%20SavedObjectsErrorHelpers.decorateForbiddenError&type=code) will generate ForbiddenError:
1. OpenSearch response forbidden error: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/32b11de6e20f3d233a0c337a7ab0a36588d54c72/src/core/server/saved_objects/service/lib/decorate_opensearch_error.ts#L80
2. Workspace saved object permission wrapper forbidden error: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/32b11de6e20f3d233a0c337a7ab0a36588d54c72/src/plugins/workspace/server/saved_objects/workspace_saved_objects_client_wrapper.ts#L47
3. Data source permission wrapper forbidden error: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/32b11de6e20f3d233a0c337a7ab0a36588d54c72/src/plugins/data_source_management/server/saved_objects/data_source_premission_client_wrapper.ts#L163
I think it's OK to response 403 for case 2 and 3, seems they are user permission related issues. 
For case 1, the server response 500 in the old implementation. The current implementation will return 403. 
We may need to check the error message if we want to distinguish these cases. Do you have other suggestions? I think response 403 would be OK for case 1.

SuZhou-Joe: Nit: Why do we check if there is workspaceId present? I think returning 403 error if the error is a forbidden error seems straightforward to me.

SuZhou-Joe (approved):

Issue #8612, opened 2024-10-16
[BUG][Discover-next] Discover UI stuck on 'Searching' after deleting index pattern from Index Management


ananzh:
**Describe the bug**

When using Discover with query enhancement enabled, deleting an index pattern from Index Management does not properly update the "Recently selected data" list in Discover. This causes the UI to become stuck in a "Searching" state when attempting to use Discover after deleting an index pattern.

**To Reproduce**
1. Enable Discover query enhancement
2. Add some index patterns and indices (e.g. "aaa*" (default) and "test-flat*")
3. Go to Index Management page
4. Delete the "aaa*" index pattern
5. Return to the Discover page
6. Observe that the "Recently selected data" list still shows the deleted "aaa*" pattern
7. Attempt to use Discover

**Expected behavior**
After deleting an index pattern: The "Recently selected data" list in Discover should update to remove the deleted index pattern. Discover should function normally without getting stuck in a "Searching" state. The UI should reflect the current state of available index patterns accurately.

**OpenSearch Version**
main

**Dashboards Version**
main

**Plugins**
security-dashboards

**Screenshots**
<img width="1677" alt="Screenshot 2024-10-16 at 1 52 35 PM" src="https://github.com/user-attachments/assets/f2413c68-0bfb-4b89-a384-cfa1e17a794c">

PR #8659, opened 2024-10-18
[bug] Discover UI stuck on searching after deleting index pattern

kavilla:
### Description

When using Discover with query enhancement enabled, deleting an index pattern from Index Management does not properly update the "Recently selected data" list in Discover. This causes the UI to become stuck in a "Searching" state when attempting to use Discover after deleting an index pattern.

Handle the error case where the use index patterns hook caught error when enhancements was enabled.

Now it goes to the next available index pattern

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8612

## Screenshot

Screenshot 2024-10-18 at 2 13 56 PM

## Testing the changes

re-creation steps in issue

## Changelog
- fix: Discover UI stuck on searching after failing to find deleted index pattern

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/view_components/utils/use_index_pattern.ts
+++ b/src/plugins/discover/public/application/view_components/utils/use_index_pattern.ts
@@ -74,25 +74,25 @@ export const useIndexPattern = (services: DiscoverViewServices) => {
             handleIndexPattern();
           }
         } else {
-          try {
-            const ip = await fetchIndexPatternDetails(indexPatternIdFromState);
-            if (isMounted) {
-              setIndexPattern(ip);
-            }
-          } catch (error) {
-            if (isMounted) {
-              const warningMessage = i18n.translate('discover.indexPatternFetchErrorWarning', {
-                defaultMessage: 'Error fetching index pattern: {error}',
-                values: { error: (error as Error).message },
-              });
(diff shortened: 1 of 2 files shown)


src/plugins/discover/public/application/view_components/utils/use_index_pattern.ts
```diff
@@ -74,25 +74,25 @@ export const useIndexPattern = (services: DiscoverViewServices) => {
...
+          if (isMounted) {
+            setIndexPattern(ip);
           }
         }
       }
     };
 
-    handleIndexPattern();
+    try {
+      handleIndexPattern();
```
kavilla, line 86: so instead of just handling the error case when enhancements is in not enabled. i wrapped the function in the same try catch. so in reality the delta change here was just moving where we put the try catch on

virajsanghvi (approved):
code looks fine, should there be a test case to prevent a regression here in the future?

approving to unblock + assuming it passes ci

ananzh (approved):

AMoo-Miki:
Manually backported to 2.x with https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8670

PR #8671, opened 2024-10-19
Select the correct language when a different one is chosen by the dataset selector

AMoo-Miki:
### Description

Select the correct language when a different one is chosen by the dataset selector

## Changelog
- fix: Select the correct language when a different one is chosen by the dataset selector

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/query_editor/language_selector.tsx
+++ b/src/plugins/data/public/ui/query_editor/language_selector.tsx
@@ -44,7 +44,7 @@ export const QueryLanguageSelector = (props: QueryLanguageSelectorProps) => {
     : undefined;
 
   useEffect(() => {
-    const subscription = queryString.getUpdates$().subscribe((query: Query) => {
+    const subscription = queryString.getUpdates$(0).subscribe((query: Query) => {
       if (query.language !== currentLanguage) {
         setCurrentLanguage(query.language);
       }
--- a/src/plugins/data/public/ui/query_editor/query_editor.tsx
+++ b/src/plugins/data/public/ui/query_editor/query_editor.tsx
@@ -240,11 +240,10 @@ export default class QueryEditorUI extends Component<Props, State> {
 
     if (!isEqual(this.props.query.dataset, prevQuery.dataset)) {
       if (this.inputRef) {
-        const newQuery = this.queryString.getInitialQuery();
-        const newQueryString = newQuery.query;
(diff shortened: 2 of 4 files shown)


ashwin-pc (approved):

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -240,11 +240,10 @@ export default class QueryEditorUI extends Component<Props, State> {
 
     if (!isEqual(this.props.query.dataset, prevQuery.dataset)) {
       if (this.inputRef) {
-        const newQuery = this.queryString.getInitialQuery();
-        const newQueryString = newQuery.query;
-        if (this.inputRef.getValue() !== newQueryString) {
+        const newQueryString = this.props.query.query;
```
kavilla, line 243: i think now when i select the dataset for the new language the query string is not updated with the updated dataset.

kavilla: recreation steps:
* have an index pattern and navigate to discover. load results 
* select an index from a cluster.
* notice how the query (if for sql and ppl) still references the old dataset

src/plugins/data/public/ui/query_editor/language_selector.tsx
```diff
@@ -44,7 +44,7 @@ export const QueryLanguageSelector = (props: QueryLanguageSelectorProps) => {
     : undefined;
 
   useEffect(() => {
-    const subscription = queryString.getUpdates$().subscribe((query: Query) => {
+    const subscription = queryString.getUpdates$(0).subscribe((query: Query) => {
```
kavilla, line 47: is it potentially just a race condition or something?

could we verify if the props.query.language is the current query always. and this component gets re-rendered. i need to verify since last checking these things out. or if you have the insight already.

because what im thinking is that potentially the dataset is checking which languages are supported and adding to a list while this is listening to changes and before the language is added this setter is called but set to the first language it found.

there's a couple of places i think were we need loading spinners. shooting in the dark perhaps we just need to check if everything is done loading before setting the current language?

kavilla: or opt for just disabling unsupported languages and setting the current value vs removing them from view

kavilla:
closing as i think this might. be addressed by another PR. @AMoo-Miki lemme know if that was incorrect

PR #8701, opened 2024-10-24
Add framework to show related indexed views inside dataset selector and show banner in the Discover canvas for the selected dataset

amsiglan:
### Description
In this PR we add support for
1. Selecting indexed view inside dataset selector. The indexed view is an Opensearch index which can be created for an external source like S3. When user selects an indexed view, we use this index for querying the data
2. Showing a banner at the top of Discover results canvas. This banner can be used to show actions related to the selected dataset. The first use we will add for this banner would be to show a callout to create an indexed view while searching an external source, or when an indexed view is selected, we can show an entry point to select other indexed view

### Issues Resolved
New feature

#### Note: Below screenshots were produced by making changes to the S3 type config for testing and are not part of this PR since this PR is only adding the framework.

- Callout for creating indexed view
image

- Action from callout can open the create flyout (this is all handled by the type config provider and not in this PR's code)
image

- Create indexed view banner:
image

## Testing the changes
Set up workspace with query enhancements enabled
- For the banner in Discover canvas
1. [For testing only]: Update the S3 config type to return the banner props based on the fetch status
2. Discover will render the banner based on provided props

- For showing indexed views in dataset selector
1. Update the S3 config type to return implemented instance of IndexedViewsService
2. Dataset selector will show the indexed views in the second step

## Changelog
- feat: Add framework to show banner at the top in discover results canvas
- feat: Show indexed views in dataset selector

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/8701.yml
+++ b/changelogs/fragments/8701.yml
@@ -1,0 +1,2 @@
+feat:
+- Show banner at the top in discover canvas; show indexed views in dataset selector ([#8701](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8701))
--- a/changelogs/fragments/8701.yml
+++ b/changelogs/fragments/8701.yml
@@ -1,2 +1,3 @@
 feat:
-- Show banner at the top in discover canvas; show indexed views in dataset selector ([#8701](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8701))
+- Add framework to show banner at the top in discover results canvas ([#8701](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8701))
+- Show indexed views in dataset selector ([#8701](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8701))
--- a/src/plugins/data/common/datasets/types.ts
+++ b/src/plugins/data/common/datasets/types.ts
@@ -247,6 +247,8 @@ export interface Dataset extends BaseDataset {
   timeFieldName?: string;
   /** Optional language to default to from the language selector */
   language?: string;
+  /** Optional name of the indexed view to search data from */
+  indexedView?: string;
 }
 
 export interface DatasetField {
@@ -256,6 +258,30 @@ export interface DatasetField {
   // TODO:  osdFieldType?
 }
 
+export type CanvasBannerProps =
+  | {
+      componentType: 'callout';
+      title: string;
+      iconType?: string;
+      content?: React.ReactNode;
+    }
+  | {
+      componentType: 'callout';
+      title?: string;
+      iconType?: string;
+      content: React.ReactNode;
+    }
+  | {
+      componentType: 'custom';
+      content: React.ReactNode;
+    };
+
 export interface DatasetSearchOptions {
   strategy?: string;
+  /**
+   * Returns props used to render a banner on the Discover canvas/results section
+   */
+  getBannerProps?: (
+    searchStatus: string
+  ) => CanvasBannerProps | Promise<CanvasBannerProps> | undefined;
 }
--- a/src/plugins/data/common/datasets/types.ts
+++ b/src/plugins/data/common/datasets/types.ts
@@ -258,21 +258,26 @@ export interface DatasetField {
   // TODO:  osdFieldType?
 }
 
-export type CanvasBannerProps =
+export enum DatasetCanvasBannerComponentType {
+  Callout = 'callout',
+  Custom = 'custom',
+}
+
+export type DatasetCanvasBannerProps =
   | {
-      componentType: 'callout';
+      componentType: DatasetCanvasBannerComponentType.Callout;
       title: string;
       iconType?: string;
       content?: React.ReactNode;
     }
   | {
-      componentType: 'callout';
+      componentType: DatasetCanvasBannerComponentType.Callout;
       title?: string;
       iconType?: string;
       content: React.ReactNode;
     }
   | {
-      componentType: 'custom';
+      componentType: DatasetCanvasBannerComponentType.Custom;
       content: React.ReactNode;
     };
 
--- a/src/plugins/data/common/datasets/types.ts
+++ b/src/plugins/data/common/datasets/types.ts
@@ -259,25 +259,25 @@ export interface DatasetField {
 }
 
 export enum DatasetCanvasBannerComponentType {
-  Callout = 'callout',
-  Custom = 'custom',
+  CALLOUT = 'callout',
+  CUSTOM = 'custom',
 }
 
 export type DatasetCanvasBannerProps =
   | {
-      componentType: DatasetCanvasBannerComponentType.Callout;
+      componentType: DatasetCanvasBannerComponentType.CALLOUT;
       title: string;
       iconType?: string;
       content?: React.ReactNode;
     }
   | {
-      componentType: DatasetCanvasBannerComponentType.Callout;
+      componentType: DatasetCanvasBannerComponentType.CALLOUT;
       title?: string;
       iconType?: string;
       content: React.ReactNode;
     }
   | {
-      componentType: DatasetCanvasBannerComponentType.Custom;
+      componentType: DatasetCanvasBannerComponentType.CUSTOM;
       content: React.ReactNode;
     };
 
@@ -288,5 +288,5 @@ export interface DatasetSearchOptions {
    */
   getBannerProps?: (
     searchStatus: string
-  ) => CanvasBannerProps | Promise<CanvasBannerProps> | undefined;
+  ) => DatasetCanvasBannerProps | Promise<DatasetCanvasBannerProps> | undefined;
 }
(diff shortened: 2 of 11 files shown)


src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -52,25 +53,49 @@ export const Configurator = ({
...
+  useEffect(() => {
+    setLanguages(type?.supportedLanguages(dataset) || []);
+  }, [dataset, type]);
+
+  useEffect(() => {
     const currentLanguage = queryString.getQuery().language;
     if (languages.includes(currentLanguage)) {
-      return currentLanguage;
+      setSelectedLanguage(currentLanguage);
     }
```
virajsanghvi, line 70: don't you need to return, otherwise, don't you always set languages[0] below?

virajsanghvi (commented):
Is there a plan for unit tests?

Also, can you double check ci failures? One I think is a known issue but the other i'm not sure about

src/plugins/data/public/ui/dataset_selector/configurator.tsx
amsiglan: Fixed, thanks for catching

amsiglan:
> Is there a plan for unit tests?
> 
> Also, can you double check ci failures? One I think is a known issue but the other i'm not sure about

For now all changes are behind a flag, unit tests will be added as fast follow next week

changelogs/fragments/8701.yml
```diff
@@ -0,0 +1,2 @@
+feat:
+- Show banner at the top in discover canvas; show indexed views in dataset selector ([#8701](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8701))
```
AMoo-Miki, line 2: Please separate these and rephrase them.

src/plugins/data/common/datasets/types.ts
```diff
@@ -256,6 +258,30 @@ export interface DatasetField {
   // TODO:  osdFieldType?
 }
 
+export type CanvasBannerProps =
```
AMoo-Miki, line 261: Please prefix this new type with `Data` to avoid collisions.

src/plugins/data/common/datasets/types.ts
```diff
@@ -256,6 +258,30 @@ export interface DatasetField {
...
+  | {
+      componentType: 'callout';
```
AMoo-Miki, line 263: Please use an exported `enum` for `componentType`.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -52,25 +53,50 @@ export const Configurator = ({
       defaultMessage: "I don't want to use the time filter",
     }
   );
-  const [language, setLanguage] = useState<string>(() => {
+  const [selectedLanguage, setSelectedLanguage] = useState<string>('');
+  const indexedViewsService = type?.indexedViewsService;
+  const [selectedIndexedView, setSelectedIndexedView] = useState<string | undefined>();
+  const [indexedViews, setIndexedViews] = useState<IndexedView[]>([]);
+  const [loadingIndexedViews, setLoadingIndexedViews] = useState(false);
```
AMoo-Miki, line 60: Current OSD naming convention is to prefix these with `is`.
```suggestion
  const [isLoadingIndexedViews, setIsLoadingIndexedViews] = useState(false);
```

src/plugins/data/common/datasets/types.ts
```diff
@@ -256,6 +258,35 @@ export interface DatasetField {
   // TODO:  osdFieldType?
 }
 
+export enum DatasetCanvasBannerComponentType {
+  Callout = 'callout',
+  Custom = 'custom',
```
AMoo-Miki, line 263: While I know OSD has made a mess of using enums but the correct convention is for enum names to be all uppercase to indicate their constant-ness.
```suggestion
  CALLOUT = 'callout',
  CUSTOM = 'custom',
```

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -52,25 +53,57 @@ export const Configurator = ({
...
+  useEffect(() => {
     const currentLanguage = queryString.getQuery().language;
     if (languages.includes(currentLanguage)) {
-      return currentLanguage;
+      setSelectedLanguage(currentLanguage);
+      return;
     }
-    return languages[0];
-  });
+    setSelectedLanguage(languages[0]);
```
AMoo-Miki, line 72: I noticed that the default value for `selectedLanguage` was `''`. This line could potentially set it to `undefined`. Is the logic capable of handling that?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -52,25 +53,57 @@ export const Configurator = ({
...
+  const [indexedViews, setIndexedViews] = useState<IndexedView[]>([]);
+  const [isLoadingIndexedViews, setIsLoadingIndexedViews] = useState(false);
+
+  useEffect(() => {
+    setLanguages(type?.supportedLanguages(dataset) || []);
+  }, [dataset, type]);
+
+  useEffect(() => {
     const currentLanguage = queryString.getQuery().language;
     if (languages.includes(currentLanguage)) {
```
AMoo-Miki, line 68: I would sleep a lot better at night if we had a check here for `Array.isArray(languages)` considering others can register dataset types and not honor our types and that `registerType` doesn't perform any validation.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
AMoo-Miki: Do we even do any validation on these languages? What would happen if `selectedLanguage = 'invalid'`? It would be great if we could add some comments (if none exist already) about the treatment of unset, empty, and other invalid values.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
AMoo-Miki, line 1: Line 119:
```
 if (baseDataset?.dataSource?.meta?.supportsTimeFilter === false && timeFields.length > 0) {
      setTimeFields([]);
      return;
}
```
I am not sure this logic is adequate enough. It doesn't handle `baseDataset?.dataSource?.meta?.supportsTimeFilter === undefined`. Wouldn't `baseDataset?.dataSource?.meta?.supportsTimeFilter !== true` be a better condition?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
amsiglan: This PR is not changing the existing behavior for languages and only making it reactive so that we can decide whether to show the indexed views. I would suggest we add validations with proper plan for handling the invalid values

src/plugins/data/public/ui/dataset_selector/configurator.tsx
amsiglan: I tried this change but it breaks the configurator and ends up calling fetch fields in a loop, keeping it as is

src/plugins/data/common/datasets/types.ts
```diff
@@ -256,6 +258,35 @@ export interface DatasetField {
...
+    };
+
 export interface DatasetSearchOptions {
   strategy?: string;
+  /**
+   * Returns props used to render a banner on the Discover canvas/results section
+   */
+  getBannerProps?: (
+    searchStatus: string
+  ) => DatasetCanvasBannerProps | Promise<DatasetCanvasBannerProps> | undefined;
```
ashwin-pc, line 291: i dont think this is the right place to put this code. Daatset search options is meant for options to augment the search request itself. we can add this directly to `DatasetTypeConfig` as `getSearchBanner` in `src/plugins/data/public/query/query_string/dataset_service/types.ts`.

ashwin-pc (commented):
Besides the one change i called out the rest of the code flow seems fine to me. I have not validated the change for any regressions or coverage. Can we make the change I mentioned?

src/plugins/data/public/query/query_string/dataset_service/types.ts
```diff
@@ -16,6 +16,14 @@ export interface DataStructureFetchOptions {
   paginationToken?: string;
 }
 
+export interface IndexedView {
```
AMoo-Miki, line 19: Please prefix this and everything else that exported from here with `Dataset`

src/plugins/data/common/datasets/types.ts
riysaxen-amzn: addressed in -. https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8818

src/plugins/data/public/query/query_string/dataset_service/types.ts
riysaxen-amzn: addressed in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8818

PR #8702, opened 2024-10-24
[Discover] notify user about field_type_tolerance for SQL/PPL in query footer

joshuali925:
### Description

Add icon in query editor footer to notify user about limitation in array type in SQL/PPL responses. The icon is present if `plugins.query.field_type_tolerance` is unset or set to `false`, clicking on it will open a popover with details. The popover is open by default if user hasn't seen it before (state persisted in localStorage).

This only applies to OpenSearch data sources. If request to cluster settings API failed, it will not display the icon.

This PR also renames `SQL` to `OpenSearch SQL`

### Issues Resolved

related: https://github.com/opensearch-project/sql/pull/3118

## Screenshot

image

## Changelog
- feat: notify user about field_type_tolerance for SQL/PPL in discover query footer

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/public/plugin.tsx
+++ b/src/plugins/query_enhancements/public/plugin.tsx
@@ -101,7 +101,7 @@ export class QueryEnhancementsPlugin
     // Register SQL language
     const sqlLanguageConfig: LanguageConfig = {
       id: 'SQL',
-      title: 'SQL',
+      title: 'OpenSearch SQL',
       search: sqlSearchInterceptor,
       getQueryString: (query: Query) => {
         return `SELECT * FROM ${query.dataset?.title} LIMIT 10`;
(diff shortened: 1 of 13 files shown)


src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -87,13 +92,16 @@ export class QueryEnhancementsPlugin
       editor: enhancedPPLQueryEditor,
       editorSupportedAppNames: ['discover'],
       supportedAppNames: ['discover', 'data-explorer'],
+      inputFooterItems: {
```
ashwin-pc, line 95: Why cant we reuse the editor prop or pass it directly in the editor?

joshuali925:
this is no longer needed

PR #8703, opened 2024-10-25
Add query enhancement cypress tests to Github CI

abbyhu2000:
### Description

Add query enhancement cypress tests workflow to github CI to test discover 2.0 functionalities: 
- sql, ppl queries
- dataset selector
- ...

The tests can be added under folder `cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancement` in the functional repo, and will be run in ciGroup10 in the CI.

## Changelog

- feat: Add query enhancement cypress tests to Github CI
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/.github/workflows/cypress_workflow.yml
+++ b/.github/workflows/cypress_workflow.yml
@@ -93,8 +93,7 @@ jobs:
         matrix.config == 'dashboard' &&
         'node scripts/opensearch_dashboards --dev --no-base-path --no-watch --savedObjects.maxImportPayloadBytes=10485760 --server.maxPayloadBytes=1759977 --logging.json=false --data.search.aggs.shardDelay.enabled=true' ||
         'node ../scripts/opensearch_dashboards --dev --no-base-path --no-watch --savedObjects.maxImportPayloadBytes=10485760 --server.maxPayloadBytes=1759977 --logging.json=false --data.search.aggs.shardDelay.enabled=true --csp.warnLegacyBrowsers=false --uiSettings.overrides["query:enhancements:enabled"]=false' }}
-
-      SNAPSHOT_CMD: ${{ matrix.config == 'query_enhanced' &&
+      OPENSEARCH_SNAPSHOT_CMD: ${{ matrix.config == 'query_enhanced' &&
         'node ../scripts/opensearch snapshot -E --sql cluster.routing.allocation.disk.threshold_enabled=false' ||
         matrix.config == 'dashboard' &&
         'node scripts/opensearch snapshot -E cluster.routing.allocation.disk.threshold_enabled=false' ||
@@ -203,13 +202,18 @@ jobs:
       - run: npx cypress cache list
       - run: npx cypress cache path
 
+      - name: Verify start commands
+        run: |
+          echo "OPENSEARCH_SNAPSHOT_CMD=${{ env.OPENSEARCH_SNAPSHOT_CMD }}"
+          echo "START_CMD=${{ env.START_CMD }}"
+
       # Run tests based on configuration
       - name: Run FT repo tests
         if: matrix.test_location == 'ftr'
         uses: cypress-io/github-action@v2
         with:
           working-directory: ${{ env.FTR_PATH }}
-          start: ${{ env.SNAPSHOT_CMD }}, ${{ env.START_CMD }}
+          start: ${{ env.OPENSEARCH_SNAPSHOT_CMD }}, ${{ env.START_CMD }}
           wait-on: 'http://localhost:9200, http://localhost:5601'
           command: yarn cypress:run-without-security --browser ${{ env.CYPRESS_BROWSER }} --spec ${{ env.SPEC }}
 
@@ -224,7 +228,7 @@ jobs:
         uses: cypress-io/github-action@v6
         with:
           install-command: npx cypress install --force
-          start: ${{ env.SNAPSHOT_CMD }}, ${{ env.START_CMD }}
+          start: ${{ env.OPENSEARCH_SNAPSHOT_CMD }}, ${{ env.START_CMD }}
           wait-on: 'http://localhost:9200, http://localhost:5601'
           command: yarn cypress:run-without-security --browser ${{ env.CYPRESS_BROWSER }} --spec ${{ env.DASHBOARDS_SPEC }}
 
--- a/.github/workflows/cypress_workflow.yml
+++ b/.github/workflows/cypress_workflow.yml
@@ -205,11 +205,6 @@ jobs:
       - run: npx cypress cache list
       - run: npx cypress cache path
 
-      - name: Verify start commands
-        run: |
-          echo "OPENSEARCH_SNAPSHOT_CMD=${{ env.OPENSEARCH_SNAPSHOT_CMD }}"
-          echo "START_CMD=${{ env.START_CMD }}"
-
       # Run tests based on configuration
       - name: Run FT repo tests with no query enhancements
         if: matrix.test_location == 'ftr' && matrix.config == 'standard'
--- a/.github/workflows/cypress_workflow.yml
+++ b/.github/workflows/cypress_workflow.yml
@@ -95,7 +95,7 @@ jobs:
         'node ../scripts/opensearch_dashboards --dev --no-base-path --no-watch --savedObjects.maxImportPayloadBytes=10485760 --server.maxPayloadBytes=1759977 --logging.json=false --data.search.aggs.shardDelay.enabled=true --csp.warnLegacyBrowsers=false --uiSettings.overrides["query:enhancements:enabled"]=false' }}
 
       SNAPSHOT_CMD: ${{ matrix.config == 'query_enhanced' &&
-        'node ../scripts/opensearch snapshot -E cluster.routing.allocation.disk.threshold_enabled=false --sql' ||
+        'node ../scripts/opensearch snapshot -E --sql cluster.routing.allocation.disk.threshold_enabled=false' ||
         matrix.config == 'dashboard' &&
         'node scripts/opensearch snapshot -E cluster.routing.allocation.disk.threshold_enabled=false' ||
         'node ../scripts/opensearch snapshot -E cluster.routing.allocation.disk.threshold_enabled=false' }}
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -85,8 +85,7 @@
     "release_note:generate": "scripts/use_node scripts/generate_release_note",
     "cypress:run-without-security": "env TZ=America/Los_Angeles NO_COLOR=1 cypress run --headless --env SECURITY_ENABLED=false",
     "cypress:run-with-security": "env TZ=America/Los_Angeles NO_COLOR=1 cypress run --headless --env SECURITY_ENABLED=true,openSearchUrl=https://localhost:9200,WAIT_FOR_LOADER_BUFFER_MS=500",
-    "osd:ciGroup10": "echo \"dashboard_sanity_test_spec.js\"",
-    "osd:ciGroup11": "echo \"apps/vis_builder/*.js\"",
```
abbyhu2000, line 89: removed `"osd:ciGroup11": "echo \"apps/vis_builder/*.js\"",` because vis_builder tests are already running in ciGroup1 as shown here https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/package.json#L28

abbyhu2000:
Tests under ciGroup 11: https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/1598/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

abbyhu2000:
Cypress runner: 
https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/11515231239/job/32057142616 

CiGroup10 is testing query enhancement tests that are added in this functional repo PR: https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/1598

All passing except for CiGroup 6, i think currently ciGroup6 is failing on main, should not be relate to this PR.

.github/workflows/cypress_workflow.yml
```diff
@@ -42,18 +40,65 @@ env:
...
+  LATEST_VERSION: '2.17.0'
 
 jobs:
   cypress-tests:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-          group: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+        include:
```
ananzh, line 51: Will this work and help organize?
```
matrix:
  config: [standard, query_enhanced, dashboard]
  group: [1, 2, 3, 4, 5, 6, 7, 8, 9]
  include:
    - config: query_enhanced
      group: 10
      test_location: ftr
    - config: dashboard
      group: 11
      test_location: source
  exclude:
    - config: query_enhanced
      group: [1,2,3,4,5,6,7,8,9]
    - config: dashboard
      group: [1,2,3,4,5,6,7,8,9]
```

ananzh:
QQ: so for writing tests in ciGroup10, do we need add some code to create a workspace? and test in the workspace?

abbyhu2000:
> QQ: so for writing tests in ciGroup10, do we need add some code to create a workspace? and test in the workspace?

i 

> QQ: so for writing tests in ciGroup10, do we need add some code to create a workspace? and test in the workspace?

I think i asked @ashwin-pc during standup and he responded that right now we only covers the flows on discover pages. 

Do we want to add test to include the workspace flows? or we only test what discover page covers.

.github/workflows/cypress_workflow.yml
kavilla: +1,  using matrix to create the right config could make it shorter. but i do see a benefit of making it this granular as it we have more control. as it leave spaces in the future where we migrate the tests from ftr to this repo

.github/workflows/cypress_workflow.yml
```diff
@@ -42,18 +40,65 @@ env:
...
+          # Standard test configs (groups 1-9)
+          - group: 1
+            config: standard
+            test_location: ftr
+          - group: 2
+            config: standard
+            test_location: ftr
+          - group: 3
+            config: standard
+            test_location: ftr
```
kavilla, line 61: we have a variable FTR_PATH, should we use this/change FTR_PATH to something that makes more sense to usage and add the other locations? or should we get rid of the FTR_PATH variable

.github/workflows/cypress_workflow.yml
```diff
@@ -42,18 +40,65 @@ env:
   COMMENT_TAG: '[MANUAL CYPRESS TEST RUN RESULTS]'
   COMMENT_SUCCESS_MSG: ':white_check_mark: Cypress test run succeeded!'
   COMMENT_FAILURE_MSG: ':x: Cypress test run failed!'
+  LATEST_VERSION: '2.17.0'
```
kavilla, line 43: maybe in an inline comment on why we are using this. i believe it was because of issues trying to use main and install plugins for github virtual env

kavilla: one thing we could also do is actually use to make this dynamic is rely on git tags. every release infra will cut a tag for the release in all the repos post release and then we use that to make a github release. we could potentially just get the highest version tag from our repo and assume that is the highest release version of OpenSearch that we can download

.github/workflows/cypress_workflow.yml
```diff
@@ -146,70 +203,93 @@ jobs:
       - run: npx cypress cache list
       - run: npx cypress cache path
 
-      # Setup spec files for Dashboards in-house cypress tests
-      - name: Setup spec files for Dashboards tests
-        if: ${{ inputs.specs == '' && matrix.group > 9 }}
+      # Run tests based on configuration
+      - name: Run FT repo tests with no query enhancements
```
kavilla, line 207: nit: historically we called this FTRepo.

and I think the specification `with no query enhancements ` might be unnecessary. we should go with the actual one run the query enhancements to say `with query enhancements` incase we want to follow this format in the future and add different type of situations.

.github/workflows/cypress_workflow.yml
```diff
@@ -146,70 +203,93 @@ jobs:
...
+        if: matrix.test_location == 'ftr' && matrix.config == 'standard'
+        uses: cypress-io/github-action@v2
```
kavilla, line 209: did we want to bump the cypress github action if available and works? might have improvements https://github.com/cypress-io/github-action/blob/master/CHANGELOG.md. but i see somethings in our file will need to change so maybe not worht it at this time

.github/workflows/cypress_workflow.yml
```diff
@@ -146,70 +203,93 @@ jobs:
...
+      - name: Download OpenSearch
+        if: matrix.test_location == 'ftr' && matrix.config == 'query_enhanced'
+        uses: suisei-cn/actions-download-file@v1.4.0
+        with:
+          url: https://artifacts.opensearch.org/releases/bundle/opensearch/${{ env.LATEST_VERSION }}/opensearch-${{ env.LATEST_VERSION }}-linux-x64.tar.gz
+
+      - name: Extract OpenSearch
+        if: matrix.test_location == 'ftr' && matrix.config == 'query_enhanced'
+        run: |
+          tar -xzf opensearch-*.tar.gz
```
kavilla, line 225: we should be able to untar and call it a specific name so line 238 doesn't have to be a specific version

.github/workflows/cypress_workflow.yml
```diff
@@ -146,70 +203,93 @@ jobs:
...
+        shell: bash
 
-      # Run FT repo tests
-      - name: Run FT repo tests
-        if: ${{ matrix.group < 10 }}
+      - name: Run OpenSearch
+        if: matrix.test_location == 'ftr' && matrix.config == 'query_enhanced'
+        run: |
+          /bin/bash -c "./opensearch-2.17.0/opensearch-tar-install.sh &"
+          sleep 30
```
kavilla, line 239: confused about this one.

wait-on should be actually a better resolution to this: https://github.com/cypress-io/github-action?tab=readme-ov-file#wait-on

are you seeing issues with that?

.github/workflows/cypress_workflow.yml
```diff
@@ -42,18 +40,65 @@ env:
...
+            test_location: ftr
+          # Dashboard tests
+          - group: 11
+            config: dashboard
+            test_location: source
     container:
       image: docker://opensearchstaging/ci-runner:ci-runner-rockylinux8-opensearch-dashboards-integtest-v2
       options: --user 1001
     env:
+      START_CMD: ${{ matrix.config == 'query_enhanced' &&
```
kavilla, line 92: i really like this concept of `config` because it actually follows the format well with how OSD functional tests are structured!

only nit I have here is that this actually is a little bit hard to read. and any fixes or additional updates to this will require us to know to update both places.

so i think minimum we should consider the logic being where we append extra configurations if the matrix config is query enhanced instead of having two.

if that is not good enough then it might be time to start sourcing a .env config file that stores this information so that we can modify this stuff and make this file more approachable.

.github/workflows/cypress_workflow.yml
```diff
@@ -42,18 +40,65 @@ env:
...
+            config: standard
+            test_location: ftr
+          - group: 2
+            config: standard
+            test_location: ftr
+          - group: 3
+            config: standard
+            test_location: ftr
+          - group: 4
+            config: standard
```
kavilla, line 63: nit: `default` has been used for this type of situation

.github/workflows/cypress_workflow.yml
```diff
@@ -128,14 +173,26 @@ jobs:
...
 
+      # Setup spec files for Dashboards in-house cypress tests
+      - name: Setup spec files for Dashboards tests
+        if: ${{ inputs.specs == '' && matrix.test_location == 'source' }}
+        shell: bash
+        run: |
+          IFS="," read -a SPEC_ARRAY <<< $(yarn --silent osd:ciGroup${{ matrix.group }})
+          DASHBOARDS_SPEC=''
+          for i in "${SPEC_ARRAY[@]}"; do
+          DASHBOARDS_SPEC+="cypress/integration/core_opensearch_dashboards/${i},"
```
kavilla, line 184: we desperately need to be able to tag cypress test so that when we write them inhouse the test can decide which workflow it will run in

kavilla (commented):
i really like the proposal. i mentioned this in the comments moving towards the config route is actually more in line with how OSD functional tests are so the system is better built for that. So nice call!

Some comments though related to future proving the solution

kavilla (approved):
everything i think can just be a fast follow

ananzh (approved):

AMoo-Miki:
manual backport was merged into 2.x

PR #8707, opened 2024-10-25
[BUG][Discover] Ensure save query loaded properly from asset

ananzh:
### Description

Saved query workflow 
```
useSavedQuery → populateStateFromSavedQuery with new saved query → queryString.setQuery(new saved query)
useSavedQuery → populateStateFromSavedQuery with new saved query → filterManager.setFilters(new saved query)
```
Then they are clean out in use_search
```
useEffect(() => {
    (async () => {

      const query =
        savedSearchInstance.searchSource.getField('query') ||
        data.query.queryString.getDefaultQuery();

      const isEnhancementsEnabled = await uiSettings.get('query:enhancements:enabled');
      if (isEnhancementsEnabled && query.dataset) {
        let pattern = await data.indexPatterns.get(
          query.dataset.id,
          query.dataset.type !== 'INDEX_PATTERN'
        );
        if (!pattern) {
          await data.query.queryString.getDatasetService().cacheDataset(query.dataset, {
            uiSettings: services.uiSettings,
            savedObjects: services.savedObjects,
            notifications: services.notifications,
            http: services.http,
            data: services.data,
          });
          pattern = await data.indexPatterns.get(
            query.dataset.id,
            query.dataset.type !== 'INDEX_PATTERN'
          );
          savedSearchInstance.searchSource.setField('index', pattern);
        }
      }

      // sync initial app filters from savedObject to filterManager
      const filters = cloneDeep(savedSearchInstance.searchSource.getOwnField('filter'));
      const actualFilters = [];

      if (filters !== undefined) {
        const result = typeof filters === 'function' ? filters() : filters;
        if (result !== undefined) {
          actualFilters.push(...(Array.isArray(result) ? result : [result]));
        }
      }

      data.query.queryString.setQuery(query);
      setSavedSearch(savedSearchInstance);

      if (savedSearchInstance?.id) {
        chrome.recentlyAccessed.add(
          savedSearchInstance.getFullPath(),
          savedSearchInstance.title,
          savedSearchInstance.id,
          {
            type: savedSearchInstance.getOpenSearchType(),
          }
        );
      }
    })();
    // This effect will only run when getSavedSearchById is called, which is
    // only called when the component is first mounted.
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [getSavedSearchById, savedSearchId]);
```

After change:

* legacy 

https://github.com/user-attachments/assets/fe203bb8-e752-4995-ba94-4c365c93f5ee

* with enhancement

https://github.com/user-attachments/assets/29228da0-b2e0-408d-ac1d-8b390069abef

### Issues Resolved

NA

## Screenshot

NA

## Changelog

- fix: Ensure save query loaded properly from asset

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/view_components/utils/use_search.ts
+++ b/src/plugins/discover/public/application/view_components/utils/use_search.ts
@@ -86,7 +86,9 @@ export const useSearch = (services: DiscoverViewServices) => {
   const { pathname } = useLocation();
   const initalSearchComplete = useRef(false);
   const [savedSearch, setSavedSearch] = useState<SavedSearch | undefined>(undefined);
-  const { savedSearch: savedSearchId, sort, interval } = useSelector((state) => state.discover);
+  const { savedSearch: savedSearchId, sort, interval, savedQuery } = useSelector(
+    (state) => state.discover
+  );
   const indexPattern = useIndexPattern(services);
   const skipInitialFetch = useRef(false);
   const {
@@ -387,9 +389,12 @@ export const useSearch = (services: DiscoverViewServices) => {
 
       // sync initial app filters from savedObject to filterManager
       const filters = cloneDeep(savedSearchInstance.searchSource.getOwnField('filter'));
-      const actualFilters = [];
 
-      if (filters !== undefined) {
+      let actualFilters: any[] = [];
+
+      if (savedQuery) {
+        actualFilters = data.query.filterManager.getFilters();
+      } else if (filters !== undefined) {
         const result = typeof filters === 'function' ? filters() : filters;
         if (result !== undefined) {
           actualFilters.push(...(Array.isArray(result) ? result : [result]));
@@ -397,7 +402,7 @@ export const useSearch = (services: DiscoverViewServices) => {
       }
 
       filterManager.setAppFilters(actualFilters);
-      data.query.queryString.setQuery(query);
+      data.query.queryString.setQuery(savedQuery ? data.query.queryString.getQuery() : query);
       setSavedSearch(savedSearchInstance);
 
       if (savedSearchInstance?.id) {
--- a/changelogs/fragments/8707.yml
+++ b/changelogs/fragments/8707.yml
@@ -1,0 +1,2 @@
+fix:
+- Ensure save query loaded properly from asset ([#8707](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8707))


src/plugins/discover/public/application/view_components/utils/use_search.ts
ashwin-pc, line 1: @ananzh i think its high time we wrote tests for this hook. Its critical to how the app works and reviewing a PR that modifies it without tests makes me nervous. You dont need to test every flow right now, but lets start to add tests so that the main expected behaviour of the hook is tested and the changes you are making are also tested

yarn.lock
ashwin-pc, line 1: Can we skip this from the PR?

ashwin-pc (commented):
While the change looks fine to me, the fact that its changing how useSearch works makes me nervous. Lets add tests here, both unit tests and functional

src/plugins/discover/public/application/view_components/utils/use_search.ts
ananzh: It is a relative safe change bc it only modify the behavior when there is a saved query id. I will update a test for the changes.

ashwin-pc (approved):

virajsanghvi (approved):

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -387,17 +389,20 @@ export const useSearch = (services: DiscoverViewServices) => {
...
+      } else if (filters !== undefined) {
         const result = typeof filters === 'function' ? filters() : filters;
         if (result !== undefined) {
           actualFilters.push(...(Array.isArray(result) ? result : [result]));
         }
       }
 
       filterManager.setAppFilters(actualFilters);
-      data.query.queryString.setQuery(query);
+      data.query.queryString.setQuery(savedQuery ? data.query.queryString.getQuery() : query);
```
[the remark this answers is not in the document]
ananzh: Yes, the QueryStringManager is also updated when a saved query is applied. In the same populateStateFromSavedQuery function, the query string is updated:

```
queryString.setQuery({
  ...savedQuery.attributes.query,
  // We should keep the currently selected dataset intact
  dataset: queryString.getQuery().dataset,
});
```

This means that when the code in use_search.ts runs:
```
data.query.queryString.setQuery(query);
```
The query variable is being set with a value that might override what was previously set by the saved query.
My previous settings
```
data.query.queryString.setQuery(savedQuery ? data.query.queryString.getQuery() : query);
```
I think my goal was to ensure that when a saved query exists, its query string settings take precedence over any other query settings. But I saw it was removed. I think there might be some other concerns or other updates. I am not sure why it was removed. Did we see regression bugs by changing back to `data.query.queryString.setQuery(query);`.

PR #8708, opened 2024-10-26
[Discover] Modify the search bar info box content for sql/ppl

RyanL1997:
### Description
- Modify the search bar info box content for sql/ppl
- The info box will only show up when it is user's first time to select SQL/PPL
- Change the language picker's title from "SQL" into "OpenSearch SQL" 

### Issues Resolved

* Relate https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8702
* we can close the above since this PR reflects the new design from UIUX

## Screenshot

Version Final:

https://github.com/user-attachments/assets/5a34c5a6-0504-4771-a949-f1bb00bf0d3a

## Changelog

- fix: [Discover] Modify the search bar info box content for sql/ppl #8708

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/language_service/lib/language_reference.test.tsx
+++ b/src/plugins/data/public/query/query_string/language_service/lib/language_reference.test.tsx
@@ -1,0 +1,130 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from 'react';
+import { render, screen, fireEvent, cleanup, act, waitFor } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { LanguageReference } from './language_reference';
+import { IntlProvider } from 'react-intl';
+
+// Helper to wrap component with IntlProvider
+const renderWithIntl = (ui) => {
+  return render(<IntlProvider locale="en">{ui}</IntlProvider>);
+};
+
+describe('LanguageReference Component', () => {
+  beforeEach(() => {
+    // Clear localStorage and DOM before each test
+    localStorage.clear();
+    document.body.innerHTML = '';
+  });
+
+  afterEach(() => {
+    cleanup();
+  });
+
+  test('auto-opens the info box on first load if autoShow is true and sets localStorage key', async () => {
+    await act(async () => {
+      renderWithIntl(
+        <LanguageReference
+          body={<div data-test-subj="test-body">Test Body</div>}
+          autoShow={true}
+          selectedLanguage="SQL"
+        />
+      );
+    });
+
+    // Check localStorage was set
+    const storageKey = 'hasSeenInfoBox_SQL';
+    expect(localStorage.getItem(storageKey)).toBe('true');
+
+    // Wait for and check if popover content is visible
+    await waitFor(() => {
+      const popoverContent = screen.getByTestId('test-body');
+      expect(popoverContent).toBeInTheDocument();
+    });
+  });
+
+  test('toggles the info box open and close on button click', async () => {
+    await act(async () => {
+      renderWithIntl(
+        <LanguageReference
+          body={<div data-test-subj="test-body">Test Body</div>}
+          autoShow={false}
+          selectedLanguage="PPL"
+        />
+      );
+    });
+
+    // Wait for button to be available
+    await waitFor(() => {
+      const button = screen.getByTestId('languageReferenceButton');
+      expect(button).toBeInTheDocument();
+    });
+
+    // Initially closed
+    expect(screen.queryByTestId('test-body')).not.toBeInTheDocument();
+
+    // Click to open
+    const button = screen.getByTestId('languageReferenceButton');
+    await act(async () => {
+      fireEvent.click(button);
+    });
+
+    // Wait for content to be visible
+    await waitFor(() => {
+      const content = screen.getByTestId('test-body');
+      expect(content).toBeInTheDocument();
+    });
+
+    // Click to close
+    await act(async () => {
+      fireEvent.click(button);
+    });
+
+    // Wait for content to be removed
+    await waitFor(() => {
+      expect(screen.queryByTestId('test-body')).not.toBeInTheDocument();
+    });
+  });
+
+  test('shows correct title in popover', async () => {
+    await act(async () => {
+      renderWithIntl(
+        <LanguageReference
+          body={<div data-test-subj="test-body">Test Body</div>}
+          autoShow={true}
+          selectedLanguage="SQL"
+        />
+      );
+    });
+
+    // Wait for and check if the title is rendered
+    await waitFor(() => {
+      expect(screen.getByText('Syntax options')).toBeInTheDocument();
+    });
+  });
+
+  test('respects autoShow prop when false', async () => {
+    await act(async () => {
+      renderWithIntl(
+        <LanguageReference
+          body={<div data-test-subj="test-body">Test Body</div>}
+          autoShow={false}
+          selectedLanguage="SQL"
+        />
+      );
+    });
+
+    // Wait for component to be fully rendered
+    await waitFor(() => {
+      const button = screen.getByTestId('languageReferenceButton');
+      expect(button).toBeInTheDocument();
+    });
+
+    // Content should not be visible
+    expect(screen.queryByTestId('test-body')).not.toBeInTheDocument();
+  });
+});
--- a/src/plugins/data/public/query/query_string/language_service/lib/language_reference.tsx
+++ b/src/plugins/data/public/query/query_string/language_service/lib/language_reference.tsx
@@ -21,13 +21,13 @@ export const LanguageReference = (props: {
   useEffect(() => {
     if (
       props.selectedLanguage === 'SQL' &&
-      window.localStorage.getItem('hasSeenSQLInfoBox') === 'false'
+      window.localStorage.getItem('hasSeenSQLInfoBox') === null
     ) {
       setIsLanguageReferenceOpen(true);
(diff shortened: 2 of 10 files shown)


RyanL1997:
Unit testing is WIP. I'd gather some feedback first

src/plugins/data/public/query/query_string/language_service/lib/language_reference.tsx
```diff
@@ -7,11 +7,32 @@ import { i18n } from '@osd/i18n';
...
+      setIsLanguageReferenceOpen(true);
+      window.localStorage.setItem('hasSeenSQLInfoBox', 'true');
+    } else if (
+      props.selectedLanguage === 'PPL' &&
+      window.localStorage.getItem('hasSeenPPLInfoBox') === 'false'
+    ) {
+      setIsLanguageReferenceOpen(true);
+      window.localStorage.setItem('hasSeenPPLInfoBox', 'true');
+    }
+  }, [props.selectedLanguage]);
```
ashwin-pc, line 35: Why do we need to specify this here? can we make it generic. Something like this?

```suggestion
  useEffect(() => {
    if (autoShow) {
      const storageKey = `hasSeenInfoBox_${selectedLanguage}`;
      const hasSeenInfoBox = window.localStorage.getItem(storageKey);

      if (hasSeenInfoBox === 'false') {
        setIsLanguageReferenceOpen(true);
        window.localStorage.setItem(storageKey, 'true');
      }
    }
  }, [props.selectedLanguage]);
```

src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -42,133 +42,137 @@ export class QueryEnhancementsPlugin
...
-      startServices: core.getStartServices(),
-      usageCollector: data.search.usageCollector,
-    });
+    // Initialize local storage keys when the plugin loads, if they don't exist
+    if (!window.localStorage.getItem('hasSeenSQLInfoBox')) {
+      window.localStorage.setItem('hasSeenSQLInfoBox', 'false');
+    }
+    if (!window.localStorage.getItem('hasSeenPPLInfoBox')) {
+      window.localStorage.setItem('hasSeenPPLInfoBox', 'false');
+    }
```
ashwin-pc, line 52: Why do we ned this? why not just use undefined as the intiial state. If the key does not exist, its false by default

RyanL1997: Fixed.

src/plugins/data/public/query/query_string/language_service/lib/language_reference.tsx
RyanL1997: Good catch. I just applied the change

src/plugins/data/public/query/query_string/language_service/lib/language_reference.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+    // Clear localStorage and DOM before each test
+    localStorage.clear();
+    document.body.innerHTML = '';
+  });
+
+  afterEach(() => {
+    cleanup();
+  });
+
+  test('auto-opens the info box on first load if autoShow is true and sets localStorage key', async () => {
```
ashwin-pc, line 28: Not a blocker, but can you add a fast follow test to ensure that this only opens once. Thats the key behavior here. Lets not forget to test that.

src/plugins/data/public/query/query_string/language_service/lib/language_reference.tsx
ashwin-pc, line 1: Hmm, weird why its still complaining that these lines arent tested. Your tests have covered these lines

src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -42,133 +42,126 @@ export class QueryEnhancementsPlugin
...
-      http: core.http,
-      uiSettings: core.uiSettings,
-      startServices: core.getStartServices(),
-      usageCollector: data.search.usageCollector,
-    });
 
-    const pplControls = [pplLanguageReference()];
-    const sqlControls = [sqlLanguageReference()];
+    // Subscribe to updates to track language changes
+    queryString.getUpdates$().subscribe((query) => {
```
ashwin-pc, line 47: Oh no, i missed this part. This is risky. We should not be reregistering the language every time the query changes! Did you sanity test discover with this change?

ashwin-pc: You can avoid this logic entirely since you only need query for the selected language, but looking at the code, you dont really need it right? on line 51&52, you can remove `selectedLanguage` since each instance of `LanguageReference` is created separately for that language.

ashwin-pc (changes requested):
I'm sorry i missed this in the first review, but we should not be reregistering the language every time the query updates.

src/plugins/data/public/query/query_string/language_service/lib/language_reference.test.tsx
RyanL1997: Sure lets add that case as a followup

src/plugins/query_enhancements/public/plugin.tsx
RyanL1997: fixed. I removed the subscriber

src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -42,74 +41,59 @@ export class QueryEnhancementsPlugin
...
-      usageCollector: data.search.usageCollector,
-    });
-
-    const pplControls = [pplLanguageReference()];
-    const sqlControls = [sqlLanguageReference()];
 
-    const enhancedPPLQueryEditor = createEditor(SingleLineInput, null, pplControls, DefaultInput);
+    // Define controls once for each language and register language configurations outside of `getUpdates$`
+    const pplControls = [pplLanguageReference('PPL')];
+    const sqlControls = [sqlLanguageReference('SQL')];
```
ashwin-pc, line 47: Nit: you dont even need this since `pplLanguageReference` will always only be used for PPL right? the additional prop you are sending is redundant

src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -42,74 +41,59 @@ export class QueryEnhancementsPlugin
...
       title: 'PPL',
-      search: pplSearchInterceptor,
-      getQueryString: (query: Query) => {
-        return `source = ${query.dataset?.title}`;
-      },
-      fields: {
-        filterable: false,
-        visualizable: false,
-      },
+      search: new PPLSearchInterceptor({
```
ashwin-pc, line 53: Nit: This is a nice change, but when making a change, keep the scope limited to just whats changing since its harder to review the diff quickly. Especially when there is no functinal benifit from the change. Thsi is just my personal opinion, so take it with a grain of salt.

ashwin-pc (approved):
Thanks! This looks much better 😊

src/plugins/data/public/query/query_string/language_service/lib/language_reference.test.tsx
```diff
@@ -0,0 +1,130 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from 'react';
+import { render, screen, fireEvent, cleanup, act, waitFor } from '@testing-library/react';
+import '@testing-library/jest-dom';
```
ruanyl, line 8: Guess you don't need to import this? Because `@testing-library/jest-dom` is imported in jest setup files.

src/plugins/data/public/query/query_string/language_service/lib/language_reference.tsx
```diff
@@ -7,11 +7,27 @@ import { i18n } from '@osd/i18n';
...
+  selectedLanguage?: string;
+}) => {
+  const [isLanguageReferenceOpen, setIsLanguageReferenceOpen] = useState(props.autoShow || false);
+
+  useEffect(() => {
+    if (props.autoShow) {
+      const storageKey = `hasSeenInfoBox_${props.selectedLanguage}`;
+      const hasSeenInfoBox = window.localStorage.getItem(storageKey);
+
+      if (hasSeenInfoBox === null && props.autoShow) {
```
ruanyl, line 25: It doesn't seem to need check `props.autoShow` here again because it's checked on line 21
```suggestion
      if (hasSeenInfoBox === null) {
```

ruanyl (approved):

Gokul-Radhakrishnan:
[opensearch-release-notes-2.18.0.md](https://github.com/opensearch-project/opensearch-build/blob/main/release-notes/opensearch-release-notes-2.18.0.md#experimental-features)
image

How do we [enable this experimental feature](https://opensearch.org/docs/latest/install-and-configure/configuring-opensearch/experimental/)?
image

What is the name of this feature?

RyanL1997:
Hi @Gokul-Radhakrishnan , thanks for dropping the above comment. May I ask if the above question is related to the specific implementation of this PR? If it is only a general question that related to Discover feature for version `2.18.0`, I suggest that to file an separate issue under OpenSearch Dashboards repo, so that more maintainers can be looped into the conversation to answer your questions. 

And here are some useful links that related to the issue filing:

- Where to open an issue: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/new/choose
- Documentation for how to contribute: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/CONTRIBUTING.md#contributing-code


cc @ashwin-pc

Gokul-Radhakrishnan:
@RyanL1997 I want to try out the PPL in Discover. How can I do that?

PR #8716, opened 2024-10-28
Add data2summary check in data summary panel in discover

Qxisylolo:
### Description
This pr adds data2summary agent check in data summary panel in discover. 
if there is a query assistant agent but not to summary agent then the summary also won’t be displayed

## Screenshot
no agent:
截屏2024-10-28 11 39 23

local：
截屏2024-10-28 11 41 45

data2summary agent
截屏2024-10-28 11 39 36

https://github.com/user-attachments/assets/52f8c455-59fd-49d8-91ec-76cfdeab0688

## Changelog
- feat: adds data2summary agent check in data summary panel in discover.

### Check List

- [ ] All tests pass
 - [ ] `yarn test:jest`


src/plugins/query_enhancements/public/query_assist/components/query_assist_summary.tsx
```diff
@@ -206,6 +207,19 @@ export const QueryAssistSummary: React.FC<QueryAssistSummaryProps> = (props) =>
     });
   }, [props.core]);
 
+  useEffect(() => {
+    const checkSummaryAgent = async () => {
+      setIsSummaryAgent(false);
+      if (selectedDataset.current?.dataSource?.id) {
+        const res = await getIsSummaryAgent(selectedDataset.current.dataSource.id);
```
wanglam, line 214: Should we check if `assistantClient` was assigned before calling `getIsSummaryAgent`?

src/plugins/query_enhancements/opensearch_dashboards.json
```diff
@@ -4,7 +4,7 @@
   "server": true,
   "ui": true,
   "requiredPlugins": ["data", "opensearchDashboardsReact", "opensearchDashboardsUtils", "savedObjects", "uiActions"],
-  "optionalPlugins": ["dataSource", "usageCollection"],
+  "optionalPlugins": ["dataSource", "usageCollection","assistantDashboards"],
```
wanglam, line 7: That means an OSD plugin will depend on a third-party plugin. Should we need to provide some way to register the assistant client instead of adding an optional dependency?

chishui: I don't think we need any dependency on "assistantDashboards", this is just a API call to check if agent exists, right?

wanglam: Yeah, it's an optional plugin. The OSD will still work fine if this plugin is missing. I'm just wondering if it's suitable to add a third-party plugin dependency in the OSD. I can't find any existing examples that do the same thing.

src/plugins/query_enhancements/public/types.ts
```diff
@@ -7,7 +7,7 @@ import { CoreSetup, CoreStart } from 'opensearch-dashboards/public';
 import { DataSourcePluginStart } from 'src/plugins/data_source/public';
 import { DataPublicPluginSetup, DataPublicPluginStart } from '../../data/public';
 import { UsageCollectionSetup } from '../../usage_collection/public';
-
+import { AssistantPublicPluginStart } from '../../../../plugins/dashboards-assistant/public';
```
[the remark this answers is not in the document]
kavilla: actually think make break it since this is an optional plugin it but is trying to access it the file directly. in the case where there is no dashboards assist plugin installed what happens @Qxisylolo ? does the application crash?

src/plugins/query_enhancements/public/types.ts
```diff
@@ -22,6 +22,7 @@ export interface QueryEnhancementsPluginSetupDependencies {
 export interface QueryEnhancementsPluginStartDependencies {
   data: DataPublicPluginStart;
   dataSource?: DataSourcePluginStart;
+  assistantDashboards: AssistantPublicPluginStart;
 }
```
kavilla, line 26: this should be optional if optional plugin

src/plugins/query_enhancements/public/query_assist/utils/get_is_summary_agent.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { getAssistantClient } from '../../services/index';
+import { DATA2SUMMARY_AGENT_CONFIG_ID } from './constant';
+
+export async function getIsSummaryAgent(dataSourceID: string) {
+  const assistantClient = getAssistantClient();
```
kavilla, line 10: imo if it's easier to add this code directly into the query enhancements and do the inverse of this where you access the query enhancements client. that would be safer option as the bundle will always be distributed for query enhancements

kavilla: but there might be an even better solution that is scalable like updating a service that lets you register a summary provider and then this plugin just needs to access it from the data plugin

kavilla: but the desired solution is that the reference to this banner is free real estate for any plugin. you should be able to do this all from within the query assist plugin directly

src/plugins/query_enhancements/public/types.ts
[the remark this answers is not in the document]
ruanyl: I think this will break the build if the plugin was not pulled in to the `plugins` folder, OSD should not depends on 3rd party plugins.

src/plugins/query_enhancements/opensearch_dashboards.json
ruanyl: Agree with that, OSD core should not rely on 3rd party plugins.

ruanyl: I think we can call the API endpoint directly for now, we will move the entire assistant and summary feature to dashboards-assistant plugin in the future.

src/plugins/query_enhancements/public/types.ts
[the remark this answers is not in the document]
Qxisylolo: ok, I see, I will call the API endpoint , Thanks for all the comments

src/plugins/query_enhancements/public/query_assist/components/query_assist_summary.tsx
```diff
@@ -70,6 +72,7 @@ export const QueryAssistSummary: React.FC<QueryAssistSummaryProps> = (props) =>
   const [loading, setLoading] = useState(false); // track loading state
   const [feedback, setFeedback] = useState(FeedbackStatus.NONE);
   const [isEnabledByCapability, setIsEnabledByCapability] = useState(false);
+  const [isSummaryAgent, setIsSummaryAgent] = useState(false);
```
chishui, line 75: What about `isSummaryAgentAvailable`?

src/plugins/query_enhancements/common/constants.ts
```diff
@@ -32,6 +33,10 @@ export const API = {
     ASYNC_JOBS: `${BASE_API}/jobs`,
     CONNECTIONS: `${BASE_API}/connections`,
   },
+  AGENT_API: {
+    EXECUTE: `${BASE_API_AGENT}/agent/_execute`,
```
chishui, line 37: We don't need this `EXECUTE`, right?

src/plugins/query_enhancements/public/query_assist/components/query_assist_summary.tsx
```diff
@@ -206,6 +209,21 @@ export const QueryAssistSummary: React.FC<QueryAssistSummaryProps> = (props) =>
     });
   }, [props.core]);
 
+  useEffect(() => {
+    setIsSummaryAgent(false);
+    const fetchSummaryAgent = async () => {
+      if (selectedDataset.current?.dataSource?.id) {
+        const summaryAgentStatus = await checkAgentsExist(
```
chishui, line 216: What will happen when `checkAgentsExist` throw exception? Should we try/catch the exception?

chishui:
LGTM. @ruanyl @kavilla @wanglam  do you know why there are no auto build/test in Checks? Is this a workflow bug or was the workflow changed?

src/plugins/query_enhancements/common/constants.ts
```diff
@@ -7,6 +7,7 @@ export const PLUGIN_ID = 'queryEnhancements';
 export const PLUGIN_NAME = 'queryEnhancements';
 
 export const BASE_API = '/api/enhancements';
+export const BASE_API_AGENT = '/api/assistant';
```
ruanyl, line 10: Nit
```suggestion
export const BASE_API_ASSISTANT = '/api/assistant';
```

ruanyl (approved):
Approved with a minor comment

src/plugins/query_enhancements/public/query_assist/utils/get_is_summary_agent.ts
```diff
@@ -0,0 +1,30 @@
...
+      queryParams.append('agentConfigName', agentConfigName);
+    }
+    queryParams.append('dataSourceId', dataSourceId);
+
+    const response = await http.get(API.AGENT_API.CONFIG_EXISTS, {
+      query: Object.fromEntries(queryParams),
+    });
+    return response;
+  } catch (error) {
+    throw error;
```
ruanyl, line 28: btw, I don't see the purpose of this `catch` because it throws it immediately which is the same as not having the try-catch block.

Qxisylolo: ok, I see. thank you

ruanyl (approved):

src/plugins/query_enhancements/public/query_assist/utils/get_is_summary_agent.ts
```diff
@@ -0,0 +1,26 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { HttpSetup } from 'opensearch-dashboards/public';
+import { API } from '../../../common/constants';
+
+export async function checkAgentsExist(
+  http: HttpSetup,
+  agentConfigName: string | string[],
+  dataSourceId: string
```
Hailong-am, line 12: nit: dataSourceId should be optional

src/plugins/query_enhancements/public/query_assist/components/query_assist_summary.tsx
```diff
@@ -206,6 +209,26 @@ export const QueryAssistSummary: React.FC<QueryAssistSummaryProps> = (props) =>
     });
   }, [props.core]);
 
+  useEffect(() => {
+    setIsSummaryAgentAvailable(false);
+    const fetchSummaryAgent = async () => {
+      if (selectedDataset.current?.dataSource?.id) {
+        try {
```
Hailong-am, line 216: we should support datasource not enabled case

Qxisylolo: Thanks for the comments, updated

ruanyl: We may not need to have `if (selectedDataset.current?.dataSource?.id)`, call `await checkAgentsExist()` with `selectedDataset.current?.dataSource?.id` directly should be good. because:
1. when data source is not enabled, the selectedDataset.dataSource is undefined, in such case, we passed `undefined` to `checkAgentsExist`, this will make it to check local cluster
2. when data source is enabled, but the selected dataset is a local cluster, the selectedDataset.dataSource is undefined as well, in this case, we passed `undefined` to `checkAgentsExist` which makes it to call local cluster.

src/plugins/query_enhancements/public/query_assist/utils/get_is_summary_agent.ts
```diff
@@ -0,0 +1,26 @@
...
+  const queryParams = new URLSearchParams();
+  if (Array.isArray(agentConfigName)) {
+    agentConfigName.forEach((name) => queryParams.append('agentConfigName', name));
+  } else {
+    queryParams.append('agentConfigName', agentConfigName);
+  }
+  queryParams.append('dataSourceId', dataSourceId);
+
+  const response = await http.get(API.AGENT_API.CONFIG_EXISTS, {
+    query: Object.fromEntries(queryParams),
```
ruanyl, line 23: There might be problem to call `Object.fromEntries(queryParams)`, for example:
```typescript
// if calling append two times
queryParams.append('agentConfigName', 'name1')
queryParams.append('agentConfigName', 'name2')
// this returns {agentConfigName: 'name2'}, which name1 is missing
Object.fromEntries(queryParams),
```
I think you can simply call in this way
```
  const response = await http.get(API.AGENT_API.CONFIG_EXISTS, {
    query: { agentConfigName, dataSourceId },
  });
```

ruanyl (approved):

src/plugins/query_enhancements/public/query_assist/utils/get_is_summary_agent.ts
Qxisylolo: thanks for the advice，already updated~

src/plugins/query_enhancements/public/query_assist/components/query_assist_summary.tsx
```diff
@@ -206,6 +209,24 @@ export const QueryAssistSummary: React.FC<QueryAssistSummaryProps> = (props) =>
...
+          DATA2SUMMARY_AGENT_CONFIG_ID,
+          selectedDataset.current?.dataSource?.id
+        );
+        setIsSummaryAgentAvailable(summaryAgentStatus.exists);
+      } catch (error) {
+        // eslint-disable-next-line no-console
+        console.error(error);
+      }
+    };
+    fetchSummaryAgent();
```
SuZhou-Joe, line 227: Should we call this method only when `isEnabledByCapability` is true?

Qxisylolo: Thanks for the reminder, added it

SuZhou-Joe (approved):

ruanyl (approved):

PR #8723, opened 2024-10-28
Update QueryHistory to use sessionStorage

ananzh:
### Description
Update QueryHistory to use sessionStorage

### Issues Resolved
NA

## Screenshot

Screenshot 2024-10-28 at 1 43 00 PM

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/query_history.ts
+++ b/src/plugins/data/public/query/query_string/query_history.ts
@@ -14,12 +14,12 @@ export const HISTORY_KEY_PREFIX = 'query_';
 export class QueryHistory {
   private changeEmitter: BehaviorSubject<any[]>;
 
-  constructor(private readonly storage: DataStorage) {
+  constructor(private readonly sessionStorage: DataStorage) {
     this.changeEmitter = new BehaviorSubject<any[]>(this.getHistory());
   }
 
   public getHistoryKeys(): string[] {
-    return this.storage
+    return this.sessionStorage
       .keys()
       .filter((key: string) => key.startsWith(HISTORY_KEY_PREFIX))
       .sort((a, b) => {
@@ -31,7 +31,7 @@ export class QueryHistory {
 
   public getHistory(): any[] {
     return this.getHistoryKeys()
-      .map((key) => this.storage.get(key))
+      .map((key) => this.sessionStorage.get(key))
       .sort((a, b) => b.time - a.time);
   }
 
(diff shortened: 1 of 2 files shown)


virajsanghvi (commented):
To confirm, the fix here is using a local storage mechanism over storing the query history?

yarn.lock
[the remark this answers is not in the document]
ananzh: sure.

ananzh:
> To confirm, the fix here is using a local storage mechanism over storing the query history?

this fix is to use sessionStorage for query history.

ashwin-pc (approved):

src/plugins/data/public/query/query_string/query_history.ts
```diff
@@ -14,12 +14,12 @@ export const HISTORY_KEY_PREFIX = 'query_';
 export class QueryHistory {
   private changeEmitter: BehaviorSubject<any[]>;
 
-  constructor(private readonly storage: DataStorage) {
+  constructor(private readonly sessionStorage: DataStorage) {
```
kavilla, line 17: nit: do you mind restoring this? i think it might be a little bit confusing.

i might have been unclear. `DataStorage` is the the storage class that has an storage engine of either `sessionStorage` or `localStorage`.

I think calling this `sessionStorage` might be confusing and potentially incorrect because I can swap the engine under dataStorage without it being transparent. `storage` is more generic.

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -54,7 +54,7 @@ export class QueryStringManager {
     private readonly notifications: NotificationsSetup
   ) {
     this.query$ = new BehaviorSubject<Query>(this.getDefaultQuery());
-    this.queryHistory = createHistory({ storage });
+    this.queryHistory = new QueryHistory(this.sessionStorage);
```
kavilla, line 57: how come? should we remove the `createHistory` function now?

cc: @abbyhu2000

src/plugins/data/public/query/query_string/query_history.ts
kavilla: as you can see https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/plugin.ts#L132 i could modify this to be window.localStorage which could leave us open for regressions later.

abbyhu2000: Yes `DataStorage` is a general storage class, one of its function `createStorage` can be used to create either a localStorage or sessionStorage. https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/plugin.ts#L130

We should change this to storage or dataStorage perhaps?

kavilla: yeah i can see the case where a regression can happen that is obfuscated with the naming since someone can change the the engine used to be `window.localStorage` but then it gets passed down into query_history as renamed to `sessionStorage` when it is actually using localStorage

ananzh: @abbyhu2000 @kavilla thanks for reviewing. I added a follow-up PR https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8733

PR #8725, opened 2024-10-28
Update connections API error handling to use 503 for service unavailability

ananzh:
### Description
Update connections API error handling to use 503 for service unavailability. Previously, the API would return 500 Internal Server Error for cases like "No Living connections", which could be misleading to clients. Now it returns 503 Service Unavailable, which more accurately represents the nature of the failure.

### Issues Resolved
NA

## Screenshot
NA

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/server/routes/data_source_connection/routes.ts
+++ b/src/plugins/query_enhancements/server/routes/data_source_connection/routes.ts
@@ -31,7 +31,12 @@ export function registerDataSourceConnectionsRoutes(
         if (error.statusCode === 404 || error.statusCode === 400) {
           return response.ok({ body: [] });
         }
-        return response.custom({ statusCode: error.statusCode || 500, body: error.message });
+        // Transform 500 errors to 503 to indicate service availability issues
+        const statusCode = error.statusCode === 500 ? 503 : error.statusCode || 503;
+        return response.custom({
+          statusCode,
+          body: error.message,
+        });
       }
     }
   );


ashwin-pc (approved):

src/plugins/query_enhancements/server/routes/data_source_connection/routes.ts
```diff
@@ -31,7 +31,12 @@ export function registerDataSourceConnectionsRoutes(
         if (error.statusCode === 404 || error.statusCode === 400) {
           return response.ok({ body: [] });
         }
-        return response.custom({ statusCode: error.statusCode || 500, body: error.message });
+        // Transform 500 errors to 503 to indicate service availability issues
+        const statusCode = error.statusCode === 500 ? 503 : error.statusCode || 503;
```
virajsanghvi, line 35: Are all 500s 503s?

virajsanghvi (approved):
Approving to unblock, but left a comment

src/plugins/query_enhancements/server/routes/data_source_connection/routes.ts
ananzh: Currently we only see 500 "No Living connections". Might have other 5** errors.

joshuali925: OSD should only return 500 if itself runs into internal server error. Here if an `error` has a `.statusCode` that is 500, the error must be coming from a different service that had internal server error. They are actual 500s in other services, not Neo

PR #8749, opened 2024-10-29
[data] updates query and lang if language is not supported by query data

kavilla:
### Description

If the query currently has a language set that is not valid for that dataset type then update the language and get the initial query for that language and dataset and then send the query.

The scenario we can see this is when we are on an index pattern, and DQL selected. Then from recent datasets select an S3 connection. The dataset updates but the language does not update so therefore the query is invalid.

### Issues Resolved

n/a

## Changelog

- fix: updates query and language if language is not supported by query data

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/query_string_manager.ts
+++ b/src/plugins/data/public/query/query_string/query_string_manager.ts
@@ -31,7 +31,7 @@
 import { BehaviorSubject } from 'rxjs';
 import { skip } from 'rxjs/operators';
 import { CoreStart, NotificationsSetup } from 'opensearch-dashboards/public';
-import { debounce, isEqual } from 'lodash';
+import { isEqual } from 'lodash';
 import { i18n } from '@osd/i18n';
 import { Dataset, DataStorage, Query, TimeRange, UI_SETTINGS } from '../../../common';
 import { createHistory, QueryHistory } from './query_history';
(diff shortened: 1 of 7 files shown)


ruanyl (dismissed):

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -154,10 +154,41 @@ export class QueryStringManager {
...
+          });
+
+          // Show warning about language change
+          showWarning(this.notifications, {
+            title: i18n.translate('data.datasetLanguageChangeTitle', {
+              defaultMessage: 'Query Language Changed',
+            }),
+            text: i18n.translate('data.datasetLanguageChangeBody', {
+              defaultMessage:
+                'Selected dataset only supports {supportedLanguage} queries. Language has been changed.',
```
virajsanghvi, line 181: We don't expose the term dataset to customers, right?

There could also be more than one dataset supported?

Why not just "Query language changed to {supportedLanguage}."

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -154,10 +154,41 @@ export class QueryStringManager {
...
+        // If we have supported languages and current language isn't supported
+        if (supportedLanguages && !supportedLanguages.includes(newQuery.language)) {
+          // Get initial query with first supported language and new dataset
+          newQuery = this.getInitialQuery({
+            language: supportedLanguages[0],
+            dataset: newQuery.dataset,
+          });
+
+          // Show warning about language change
+          showWarning(this.notifications, {
```
virajsanghvi, line 175: Shouldn't this just be an info?

virajsanghvi:
Btw, looks like there's a cypress test failure related to discover

src/plugins/data/public/query/query_string/query_string_manager.ts
joshuali925: +1

src/plugins/data/public/query/query_string/query_string_manager.ts
kavilla: addressed

src/plugins/data/public/query/query_string/query_string_manager.ts
kavilla: it could be show warning is the helper function for this manager that passes an id that prevents duplicate notifications from popping up. so if the user constantly switches between recent data and somehow the current language is always invalid it won't keep adding a notification on the right side and keep pushing the duplicate ones upwards. it will just have a single notification available.

kavilla: Screenshot 2024-10-30 at 12 59 26 PM

sejli (approved):

sejli:
Does this need a `backport 2.18` label?

joshuali925 (approved):

virajsanghvi (approved):

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -154,10 +154,40 @@ export class QueryStringManager {
...
+            title: i18n.translate('data.languageChangeTitle', {
+              defaultMessage: 'Language Changed',
```
AMoo-Miki, line 177:
```suggestion
              defaultMessage: 'Language changed',
```

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -154,10 +154,40 @@ export class QueryStringManager {
...
+        // Get supported languages for the dataset
+        const supportedLanguages = this.datasetService
+          .getType(newQuery.dataset.type)
+          ?.supportedLanguages(newQuery.dataset);
+
+        // If we have supported languages and current language isn't supported
+        if (supportedLanguages && !supportedLanguages.includes(newQuery.language)) {
+          // Get initial query with first supported language and new dataset
+          newQuery = this.getInitialQuery({
+            language: supportedLanguages[0],
```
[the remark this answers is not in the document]
kavilla: it shouldn't but with new features i'd imagine there prolly is some way a plugin developer can attempt. will add an issue

src/plugins/data/public/query/query_string/query_string_manager.ts
kavilla: i'm not sure if this will retrigger requiring approvals so will do this in fast follow! thanks

PR #8758, opened 2024-10-30
[Discover] Filter out OpenSearch Serverless Collections

sejli:
### Description

Filters out OpenSearch Serverless Collections after fetching data sources

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
@@ -123,6 +123,7 @@ const fetchDataSources = async (client: SavedObjectsClientContract) => {
     response.savedObjects
       .filter(
         (savedObject) =>
+          typeof savedObject.attributes?.dataSourceEngineType === 'string' &&
           !savedObject.attributes?.dataSourceEngineType?.includes('OpenSearch Serverless')
       )
       .map((savedObject) => ({
--- a/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
@@ -37,10 +37,10 @@ export const indexTypeConfig: DatasetTypeConfig = {
       timeFieldName: indexMeta?.timeFieldName,
       dataSource: dataSource
         ? {
-            id: dataSource.id,
-            title: dataSource.title,
-            type: dataSource.type,
-          }
+          id: dataSource.id,
+          title: dataSource.title,
+          type: dataSource.type,
+        }
         : DEFAULT_DATA.STRUCTURES.LOCAL_DATASOURCE,
     };
   },
@@ -120,7 +120,7 @@ const fetchDataSources = async (client: SavedObjectsClientContract) => {
     perPage: 10000,
   });
   const dataSources: DataStructure[] = [DEFAULT_DATA.STRUCTURES.LOCAL_DATASOURCE].concat(
-    response.savedObjects.map((savedObject) => ({
+    response.savedObjects.filter((savedObject) => savedObject.attributes.dataSourceEngineType !== 'OpenSearch Serverless').map((savedObject) => ({
       id: savedObject.id,
       title: savedObject.attributes.title,
       type: 'DATA_SOURCE',
--- a/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
@@ -37,10 +37,10 @@ export const indexTypeConfig: DatasetTypeConfig = {
       timeFieldName: indexMeta?.timeFieldName,
       dataSource: dataSource
         ? {
-          id: dataSource.id,
-          title: dataSource.title,
-          type: dataSource.type,
-        }
+            id: dataSource.id,
+            title: dataSource.title,
+            type: dataSource.type,
+          }
         : DEFAULT_DATA.STRUCTURES.LOCAL_DATASOURCE,
     };
   },
@@ -120,11 +120,15 @@ const fetchDataSources = async (client: SavedObjectsClientContract) => {
     perPage: 10000,
   });
   const dataSources: DataStructure[] = [DEFAULT_DATA.STRUCTURES.LOCAL_DATASOURCE].concat(
-    response.savedObjects.filter((savedObject) => savedObject.attributes.dataSourceEngineType !== 'OpenSearch Serverless').map((savedObject) => ({
-      id: savedObject.id,
-      title: savedObject.attributes.title,
-      type: 'DATA_SOURCE',
-    }))
+    response.savedObjects
+      .filter(
+        (savedObject) => savedObject.attributes.dataSourceEngineType !== 'OpenSearch Serverless'
+      )
+      .map((savedObject) => ({
+        id: savedObject.id,
+        title: savedObject.attributes.title,
+        type: 'DATA_SOURCE',
+      }))
   );
 
   return injectMetaToDataStructures(dataSources);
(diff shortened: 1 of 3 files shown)


ashwin-pc (approved):
Lgtm

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
```diff
@@ -120,11 +120,15 @@ const fetchDataSources = async (client: SavedObjectsClientContract) => {
...
   });
   const dataSources: DataStructure[] = [DEFAULT_DATA.STRUCTURES.LOCAL_DATASOURCE].concat(
-    response.savedObjects.map((savedObject) => ({
-      id: savedObject.id,
-      title: savedObject.attributes.title,
-      type: 'DATA_SOURCE',
-    }))
+    response.savedObjects
+      .filter(
+        (savedObject) => savedObject.attributes.dataSourceEngineType !== 'OpenSearch Serverless'
```
kavilla, line 125: if we got the okay to filter out serverless then this looks good to me!

definitely not a blocker for this PR or more of a question: do we think it would be worth it for the dataset type config to have a `supportedEngineType` (following the naming pattern set by @manasvinibs) a while back?

a value we could respect is `ALL` and all the data source engine types work. that way if the engine type and i think it actually should follow the complete paradigm set by mana. where it actually just be `supportedEngines` which is a record of semver engines. 

this will help future proof us like

```
supportedEngines: {
  'OpenSearch': "^1.0.0",
}
```

for scenarios like maybe a specific feature on a cluster.

cc: @virajsanghvi  @AMoo-Miki any opinions on this one

src/plugins/query_enhancements/public/datasets/s3_type.ts
```diff
@@ -198,17 +198,21 @@ const fetchDataSources = async (client: SavedObjectsClientContract): Promise<Dat
...
-      meta: {
-        query: {
-          id: savedObject.id,
-        },
-        type: DATA_STRUCTURE_META_TYPES.CUSTOM,
-      } as DataStructureCustomMeta,
-    }))
+    resp.savedObjects
+      .filter(
+        (savedObject) => savedObject.attributes.dataSourceEngineType !== 'OpenSearch Serverless'
```
kavilla, line 203: prolly could add a test for this

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
AMoo-Miki: A. If AWS wants to filter something out, they are more than welcome to patch it in their fork. OpenSearch logic should not contain no opinions for AWS.
B. @kavilla, while i like the flexibility your proposal provides, without many users asking for this flexibility, I cannot convince myself that it is worth complicating the code.

ananzh (approved):
We could a follow up on @kavilla's question

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
kavilla: yeah that's true. something that we sometimes run into is like the query assist be available on only certain clusters and version.

but yeah that would be too complicated and i think we would should solve that more universally with capability APIs

src/plugins/query_enhancements/public/datasets/s3_type.ts
virajsanghvi, line 203: Is attributes guaranteed to be defined?

virajsanghvi (commented):
How did we test these changes?

kavilla (approved):

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
```diff
@@ -120,11 +120,15 @@ const fetchDataSources = async (client: SavedObjectsClientContract) => {
     perPage: 10000,
   });
   const dataSources: DataStructure[] = [DEFAULT_DATA.STRUCTURES.LOCAL_DATASOURCE].concat(
-    response.savedObjects.map((savedObject) => ({
-      id: savedObject.id,
-      title: savedObject.attributes.title,
-      type: 'DATA_SOURCE',
-    }))
+    response.savedObjects
```
kavilla, line 123: i think this file has tests as well we can add which should be copy pasta from the other one

kavilla:
i see test failures on the ci but i feel like github actions have been having some issue yesterday and today

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
joshuali925: can this use the enum instead of string https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a70a7b83b0165c66391433ba98d4c09ad5127902/src/plugins/data_source/common/data_sources/types.ts#L60

> OpenSearch logic should not contain opinions for AWS

If OSD allows users to add OpenSearch serverless as datasources to it, I think this is not an opinion for AWS, it's a limitation that this combination does not work in OSD?

> OSD is not the one who defined dataSourceEngineType: 'OpenSearch Serverless'

q: who is defining the `dataSourceEngineType: 'OpenSearch Serverless'`

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
kavilla: but it can be another PR

kavilla (approved):

ashwin-pc (approved):

src/plugins/query_enhancements/public/datasets/s3_type.ts
sejli: Attributes is guaranteed to be defined, but not sure about `dataSourceEngineType`. This should be some of the `data_source` plugin work using the saved objects client. I'm not super familiar, but it looks like it should also be defined for data sources [here](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/workspace/public/utils.ts#L249).
```
.then((response) => {
      const objects = response?.savedObjects;
      if (objects) {
        return objects.map((source) => {
          const id = source.id;
          // Data connection doesn't have title for now, would use connectionId instead to display.
          const title = source.get('title') ?? source.get('connectionId') ?? '';
          const workspaces = source.workspaces ?? [];
          const auth = source.get('auth');
          const description = source.get('description');
          const dataSourceEngineType = source.get('dataSourceEngineType');
          const type = source.type;
```

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
sejli: Should be the `data_source` plugin when a [data source is created](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source_management/public/components/create_data_source_wizard/create_data_source_wizard.tsx#L85).
```
const handleSubmit = async (attributes: DataSourceAttributes) => {
    setIsLoading(true);
    try {
      // Fetch data source metadata from added OS/ES domain/cluster
      const metadata = await fetchDataSourceMetaData(http, attributes);
      attributes.dataSourceVersion = metadata.dataSourceVersion;
      attributes.dataSourceEngineType = metadata.dataSourceEngineType;
      attributes.installedPlugins = metadata.installedPlugins;
```

AMoo-Miki: > If OSD allows users to add OpenSearch serverless as datasources to it, I think this is not an opinion for AWS, it's a limitation that this combination does not work in OSD?

Thanks for finding that. It should have never happened.

> who is defining the dataSourceEngineType: 'OpenSearch Serverless'

Found it in the data_source:
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/190dab0570929085dd20f1c009440e97001009d7/src/plugins/data_source/server/routes/data_source_connection_validator.ts#L41-L71

Apparently all of these was added to OSD :/

PR #8772, opened 2024-10-31
[data] reset query when disabling query enhancements if type is not supported

kavilla:
### Description

Reset the query to DQL and an empty string and update to use a default index pattern when query enhancements went from enabled to disabled.

For example if the user selects SQL, then disables query enhancements we will default it back to DQL. This is to avoid bad state or failures.

Another example would be having an S3 data type which is not supported by default.

### Issues Resolved

n/a

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/query_string_manager.ts
+++ b/src/plugins/data/public/query/query_string/query_string_manager.ts
@@ -33,7 +33,7 @@ import { skip } from 'rxjs/operators';
 import { CoreStart, NotificationsSetup } from 'opensearch-dashboards/public';
 import { isEqual } from 'lodash';
 import { i18n } from '@osd/i18n';
-import { Dataset, DataStorage, Query, TimeRange, UI_SETTINGS } from '../../../common';
+import { Dataset, DataStorage, DEFAULT_DATA, Query, TimeRange, UI_SETTINGS } from '../../../common';
 import { createHistory, QueryHistory } from './query_history';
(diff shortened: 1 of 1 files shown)


src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -57,6 +57,28 @@ export class QueryStringManager {
     this.queryHistory = createHistory({ storage: this.sessionStorage });
     this.datasetService = new DatasetService(uiSettings, this.sessionStorage);
     this.languageService = new LanguageService(this.defaultSearchInterceptor, this.storage);
+
+    this.uiSettings.get$(UI_SETTINGS.QUERY_ENHANCEMENTS_ENABLED).subscribe((isEnabled) => {
```
ashwin-pc, line 61: This isn't safe enough. It should be something on page load ideally that checks for the flag and does any cleanup actions. Since we require a page reload for the flag, that should take care of the change listener missing the change and letting the bad state persist

ananzh: Is it possible that the setting changes before subscription set up or already in a bad state (ex. SQL) when the subscription is established?

ananzh:
@kavilla is this for 2.18 or 2.19?

angle943:
This is Kiro posting.

Merge conflicts and no recent progress. @kavilla — is this still active? If yes, please rebase. If not, please close.

PR #8777, opened 2024-10-31
[2.x-manual-bp][Discover] Fix error handling for ppl jobs API (#8771)

ananzh:
### Description
Backport PR:
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8771

From 8771:
* fix error handling for ppl jobs API

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/server/search/ppl_async_search_strategy.ts
+++ b/src/plugins/query_enhancements/server/search/ppl_async_search_strategy.ts
@@ -59,6 +59,9 @@ export const pplAsyncSearchStrategyProvider = (
         } else {
           request.params = { queryId: inProgressQueryId };
           const queryStatusResponse: any = await pplAsyncJobsFacet.describeQuery(context, request);
+
+          if (!queryStatusResponse.success) handleFacetError(queryStatusResponse);
+
           const queryStatus = queryStatusResponse?.data?.status;
           logger.info(`pplAsyncSearchStrategy: JOB: ${inProgressQueryId} - STATUS: ${queryStatus}`);
 
--- a/changelogs/fragments/8771.yml
+++ b/changelogs/fragments/8771.yml
@@ -1,0 +1,2 @@
+fix:
+- Fix error handling for ppl jobs API ([#8771](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8771))


sejli (approved):

src/plugins/query_enhancements/server/search/ppl_async_search_strategy.ts
```diff
@@ -59,6 +59,9 @@ export const pplAsyncSearchStrategyProvider = (
         } else {
           request.params = { queryId: inProgressQueryId };
           const queryStatusResponse: any = await pplAsyncJobsFacet.describeQuery(context, request);
+
+          if (!queryStatusResponse.success) handleFacetError(queryStatusResponse);
```
AMoo-Miki, line 63: `queryStatusResponse` could be undefined.

AMoo-Miki (changes requested):

src/plugins/query_enhancements/server/search/ppl_async_search_strategy.ts
ananzh: So this describeQuery calls fetchJobs or fetch. Both wraps the API call in a try-catch block: If the API call succeeds, it returns { success: true, data: queryRes }. If the API call fails, it returns { success: false, data: err }. So the result is either return { ...response, data: shimFunctions[dataType](response.data) } or response, with success either true or false right?

So the result from describeQuery could be
```
 if (response.success === false || !this.shimResponse) return response;
```
where it has success either true or false.

or 
```
shimFunctions[dataType]
      ? { ...response, data: shimFunctions[dataType](response.data) }
      : response;
```
where it still has response in both ways. So seems the exception for describeQuery can be captured in facet error right? What do you think?

joshuali925: it won't be undefined:
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/8d847a5dcaa010db5dcc154333c2e653cc893cac/src/plugins/query_enhancements/server/utils/facet.ts#L99
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/8d847a5dcaa010db5dcc154333c2e653cc893cac/src/plugins/query_enhancements/server/types.ts#L48-L51

joshuali925 (approved):

src/plugins/query_enhancements/server/search/ppl_async_search_strategy.ts
AMoo-Miki: The issue i am having is `queryStatusResponse: any`. Will need to understand why the code doesn't use `queryStatusResponse: FacetResponse` and why the nest line does `queryStatusResponse?.data`.

joshuali925: yes i don't know why it uses `any`, type can be omitted since for `describeQuery`, the return type is correct and there are no type errors in it

`data` shouldn't be `any` as well, it should be either `Error` or the response type. 

And i think the real problem is that i'm not sure why it goes from
```
error response -> catch -> return with `success: false` -> check if `success === false` -> throw new error wrapping response error
```
instead of just
```
error response -> throw response error
```

in the current logic it's likely miss checking `success === false`. This PR and my PR https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8743 were fixing the same problem that this check is missing

PR #8778, opened 2024-10-31
[Discover] Remove Default Cluster 

sejli:
### Description

Removes default cluster from index and S3 dataset types cluster lists

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
@@ -119,19 +119,17 @@ const fetchDataSources = async (client: SavedObjectsClientContract) => {
     type: 'data-source',
     perPage: 10000,
   });
-  const dataSources: DataStructure[] = [DEFAULT_DATA.STRUCTURES.LOCAL_DATASOURCE].concat(
-    response.savedObjects
-      .filter(
-        (savedObject) =>
-          typeof savedObject.attributes?.dataSourceEngineType === 'string' &&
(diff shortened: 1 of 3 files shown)


ashwin-pc (approved):

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
```diff
@@ -119,19 +119,17 @@ const fetchDataSources = async (client: SavedObjectsClientContract) => {
...
-          typeof savedObject.attributes?.dataSourceEngineType === 'string' &&
-          !savedObject.attributes?.dataSourceEngineType?.includes('OpenSearch Serverless')
-      )
-      .map((savedObject) => ({
-        id: savedObject.id,
-        title: savedObject.attributes.title,
-        type: 'DATA_SOURCE',
-      }))
-  );
+  const dataSources: DataStructure[] = response.savedObjects
```
kavilla, line 122: will need to fast follow on checking if local cluster is available

ashwin-pc:
Lets fast follow to allow local cluster to show up as well

joshuali925 (approved):
approving but we should add some logic to check and add if local datasource exists, and it should say local cluster, not default cluster when we do

kavilla (approved):
LGTM. Could we verify the source of the error. Is from the API and we should have fixed the response, or is it because we always need to have a data source if now. I would be surprised we are the only ones that ran into this issue. I thought it was save to assume if no data source Id then the default data source Is used.

PR #8785, opened 2024-11-01
[Discover] Update Version Filter for Data Sources

sejli:
### Description

* Uses `semver` to filter out data sources with invalid versions or versions that are lower than `1.0.0`.
* Adding tests for `index_type.ts`

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.test.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.test.ts
@@ -9,11 +9,21 @@ import { indexTypeConfig } from './index_type';
 import { SavedObjectsClientContract } from 'opensearch-dashboards/public';
 import { DATA_STRUCTURE_META_TYPES, DataStructure, Dataset } from '../../../../../common';
 import * as services from '../../../../services';
+import { IDataPluginServices } from 'src/plugins/data/public';
 
-jest.mock('../../../../services', () => ({
-  getSearchService: jest.fn(),
-  getIndexPatterns: jest.fn(),
-}));
+jest.mock('../../../../services', () => {
+  return {
+    getSearchService: jest.fn(),
+    getIndexPatterns: jest.fn(),
+    getQueryService: () => ({
+      queryString: {
+        getLanguageService: () => ({
+          getQueryEditorExtensionMap: jest.fn().mockReturnValue({}),
+        }),
+      },
+    }),
+  };
+});
 
 describe('indexTypeConfig', () => {
   const mockSavedObjectsClient = {} as SavedObjectsClientContract;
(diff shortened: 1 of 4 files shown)


src/plugins/query_enhancements/public/datasets/s3_type.ts
```diff
@@ -197,11 +198,10 @@ const fetchDataSources = async (client: SavedObjectsClientContract): Promise<Dat
...
   const dataSources: DataStructure[] = resp.savedObjects
-    .filter(
-      (savedObject) =>
-        typeof savedObject.attributes?.dataSourceEngineType === 'string' &&
-        !savedObject.attributes?.dataSourceEngineType?.includes('OpenSearch Serverless')
-    )
+    .filter((savedObject) => {
+      const coercedVersion = semver.coerce(savedObject.attributes.dataSourceVersion);
+      return coercedVersion ? semver.satisfies(coercedVersion, '>=1.0.0') : false;
+    })
```
Hailong-am, line 204: should we use `semver.vaild(savedObject.attributes.dataSourceVersion)` to check datsource version is valid first and then do satisfies check?

src/plugins/query_enhancements/public/datasets/s3_type.test.ts
```diff
@@ -142,36 +142,40 @@ describe('s3TypeConfig', () => {
...
           {
             id: 'ds2',
-            attributes: { title: 'DataSource 2', dataSourceEngineType: 'OpenSearch Serverless' },
+            attributes: { title: 'DataSource 2', dataSourceVersion: '' },
+          },
+          { id: 'ds3', attributes: { title: 'DataSource 3', dataSourceVersion: '2.17.0' } },
+          {
+            id: 'ds4',
+            attributes: { title: 'DataSource 4', dataSourceVersion: '.0' },
           },
```
Hailong-am, line 170: can we add some test to check invalid semver version and semver version with snapshot/alpha etc?

src/plugins/query_enhancements/public/datasets/s3_type.ts
Flyingliuhub: +1, If semver.coerce returns null, this will lead to semver.satisfies(null, '>=x.x.x'), could cause an issue? You may want to add a null check before using semver.satisfies, as semver.satisfies doesn't handle the null

src/plugins/query_enhancements/public/datasets/s3_type.test.ts
virajsanghvi: What happens on an invalid version? I assume coerce is just not touching it and it fails satisfies? Data source 2 is an example of an invalid version, but the truthy check on coercedVersion handles that case (I assume).

Docs looks like coerce should handle second case so don't think that's a big issue.

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
```diff
@@ -120,11 +121,10 @@ const fetchDataSources = async (client: SavedObjectsClientContract) => {
     perPage: 10000,
   });
   const dataSources: DataStructure[] = response.savedObjects
-    .filter(
-      (savedObject) =>
-        typeof savedObject.attributes?.dataSourceEngineType === 'string' &&
-        !savedObject.attributes?.dataSourceEngineType?.includes('OpenSearch Serverless')
-    )
+    .filter((savedObject) => {
+      const coercedVersion = semver.coerce(savedObject.attributes.dataSourceVersion);
```
virajsanghvi, line 125: This attribute always exists?

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.test.ts
```diff
@@ -77,4 +87,47 @@ describe('indexTypeConfig', () => {
...
+    });
+
+    const result = await indexTypeConfig.fetch(mockServices as IDataPluginServices, [
+      { id: 'unknown', title: 'Unknown', type: 'UNKNOWN' },
+    ]);
+
+    expect(result.children).toHaveLength(2);
+    expect(result.children?.[0].title).toBe('DataSource 1');
+    expect(result.children?.[1].title).toBe('DataSource 3');
+    expect(result.children?.some((child) => child.title === 'DataSource 2')).toBe(false);
```
virajsanghvi, line 130: nit: not sure how valuable this is

virajsanghvi (commented):
just blocking approval on the question on invalid version

src/plugins/query_enhancements/public/datasets/s3_type.ts
virajsanghvi: Doesn't the truthy check before satisfies handle your concern?

src/plugins/query_enhancements/public/datasets/s3_type.test.ts
virajsanghvi: Confirmed an invalid version returns null locally, so lgtm

virajsanghvi (approved):

src/plugins/query_enhancements/public/datasets/s3_type.ts
Flyingliuhub: Yes. truthy check should handle the concerns about `null` and `undefined`, should we validate the datasource version is valid? Thanks

Flyingliuhub (approved):

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
```diff
@@ -120,11 +121,10 @@ const fetchDataSources = async (client: SavedObjectsClientContract) => {
...
+      return coercedVersion ? semver.satisfies(coercedVersion, '>=1.0.0') : false;
```
kavilla, line 126: ah i see this what we meant with putting version. if this is the case would we be able to go with the [original suggestion](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8758#discussion_r1823351611) of adding an optional property of the dataset type ?

src/plugins/query_enhancements/public/datasets/s3_type.ts
sejli: It's my understanding that if the `dataSourceVersion` is not valid in our situation, then [`semver.coerce()`](https://docs.npmjs.com/cli/v6/using-npm/semver#coercion) should return null.
> Only text which lacks digits will fail coercion

 `dataSourceVersion` is "invalid" when it is an empty string and gets caught by the truthy check.

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
kavilla: yeah @Flyingliuhub @ZilongX confirming this is the correct attribute?

We should get cypress tests once i merge the workflow in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8703

kavilla (approved):

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
ZilongX: yes `dataSourceVersion` would always exist, it may carry empty value for invalid datasource but the filed itself would be there

ZilongX (approved):

kavilla:
Merging as validated it is working but we should revisit this

PR #8799, opened 2024-11-03
[Discover] Makes caching dataset options optional

ashwin-pc:
### Description

Makes caching dataset options optional. Enables it only for s3 since its metadtata calls are slower.

Additinal fix: When caching was no longer optional, qury assist api calls held the Clusters menu in a perpetual loading state causing the menu to not load. This fixes that by timing out only the is enabled queries to not prevent that menu from loading.

## Testing the changes

See unit test.

Or for Manual testing:
- Open Discover
- Open the advanced data selector
- Select index patterns, the menu shoud always fire a network request
- Select Indices, the menu should always fire a network request
- Select S3, the menu should only fire the query oince

## Changelog

- fix: [Discover] Makes cachign dataset options optional and configurable by the dataset type

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/dataset_service/types.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/types.ts
@@ -36,6 +36,8 @@ export interface DatasetTypeConfig {
     supportsTimeFilter?: boolean;
     /** Optional isFieldLoadAsync determines if field loads are async */
     isFieldLoadAsync?: boolean;
+    /** Optional cacheOptions determines if the data structure is cacheable. Defaults to false */
+    cacheOptions?: boolean;
   };
   /**
    * Converts a DataStructure to a Dataset.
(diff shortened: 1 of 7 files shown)


kavilla (approved):

kavilla (approved):

src/plugins/data/public/query/query_string/dataset_service/types.ts
```diff
@@ -36,6 +36,8 @@ export interface DatasetTypeConfig {
     supportsTimeFilter?: boolean;
     /** Optional isFieldLoadAsync determines if field loads are async */
     isFieldLoadAsync?: boolean;
+    /** Optional cacheOptions determines if the data structure is cacheable. Defaults to false */
+    cacheOptions?: boolean;
```
SuZhou-Joe, line 40: Why don't we just name it cacheable since it is a boolean.

SuZhou-Joe (approved):

PR #8795, opened 2024-11-02
[Discover] Get value of query from query editor instead of props

kavilla:
### Description

Not able to recreate the issue where if in DQL and press `Enter` then the text field is automatically cleared out. Not positive if this will resolve this issue but it feels inaccurate to submit the props.query instead of relying on the value of the query

### Issues Resolved

n/a

## Changelog

- fix: [Discover] get query value from editor on submit instead of query state 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [x] Update [CHANGELOG.md](./../CHANGELOG.md)
- [x] Commits are signed per the DCO using --signoff

PR #8800, opened 2024-11-03
[BUG] Fix DQL & PPL Autocomplete Regression Issues

mengweieric:
### Description

Regressions issues fixed
- query.query for onSubmit of query editor is empty on suggestion select therefore causes the entire query to be cleared out as selecting a suggestion triggers query submit
- query retrieved from getQueryString() is never being updated with query changes in query editor

## Changelog
- fix: [Discover] Fix DQL & PPL Autocomplete Regression Issues

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/query_editor/query_editor.tsx
+++ b/src/plugins/data/public/ui/query_editor/query_editor.tsx
@@ -127,7 +127,7 @@ export const QueryEditorUI: React.FC<Props> = (props) => {
   };
 
   const onSubmit = (query: Query, dateRange?: TimeRange) => {
-    // TODO: this is a hack to get the current query string as getQueryString() is not fetching the updated query string
+    // TODO: this is a hack to get the current query string as query.query in some use cases are empty
     const currentQuery = query.query || inputRef.current?.getModel()?.getValue();
     if (!currentQuery || currentQuery.trim() === '') return;
     if (props.onSubmit) {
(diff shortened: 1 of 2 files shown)


ashwin-pc:
@mengweieric we don't need to hack it. The value on submit should always come from the editor. I prefer https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8795/files as it does just that.

mengweieric:
> @mengweieric we don't need to hack it. The value on submit should always come from the editor. I prefer https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8795/files as it does just that.

This didn't work on my end.

ashwin-pc:
Why are we falling back to the incorrect query here?

ashwin-pc:
Also are you sure you applied the change to the single line editor? It works for me when we add it there

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -127,14 +127,17 @@ export const QueryEditorUI: React.FC<Props> = (props) => {
   };
 
   const onSubmit = (query: Query, dateRange?: TimeRange) => {
+    // TODO: this is a hack to get the current query string as query.query in some use cases are empty
+    const currentQuery = inputRef.current?.getModel()?.getValue() ?? query.query;
+    if (!currentQuery || currentQuery.trim() === '') return;
```
ashwin-pc, line 132: We should not do this. IF i have a vlaid query and then i clear it. This will prevent me from seeing my results

PR #8803, opened 2024-11-04
[Discover] Query editor state sync fixes

ashwin-pc:
### Description

The Query editor and similarly auto complete use the query state to corretcly query the Datasource. However the query state value used by different parts of the editor UI are out of sync. This PR clearly demarcates which instance of the query you are using and calls out why that is the rught query object to reference.

It also addresses a quirk of monaco which after initialization does not use the latest value of the query object resulting in incorrect  values being used for language and dataset. This is fixed by using useRef to store a referenc of the object that the monaco callback can use when the shortcut commands are fired.

### Issues Resolved

1. When the user is using the single line editor, they type a word and hit enter, the query clears the current value. Ths is because the monaco callback on`Enter` did not use the vlue from the editor. Fixed by using the editor value directly
2. When the page is loaded and the query states are changed e.g. langauge is changed. When `Cmd+Enter` is hit, the language and datasource revert back to the default values when the app was loaded. Thsi is fixed using the ref hack

## Changelog

- fix: Query Editor state sync issues in Discover

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/query_editor/query_editor.tsx
+++ b/src/plugins/data/public/ui/query_editor/query_editor.tsx
@@ -25,6 +25,7 @@ import {
   RecentQueriesTable,
   QueryResult,
   QueryStatus,
+  useQueryStringManager,
 } from '../..';
 import { OpenSearchDashboardsReactContextValue } from '../../../../opensearch_dashboards_react/public';
 import { fromUser, getQueryLog, PersistedLog, toUser } from '../../query';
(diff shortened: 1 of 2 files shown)


src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -74,15 +75,26 @@ export const QueryEditorUI: React.FC<Props> = (props) => {
   const headerRef = useRef<HTMLDivElement>(null);
   const bannerRef = useRef<HTMLDivElement>(null);
   const queryControlsContainer = useRef<HTMLDivElement>(null);
+  // TODO: There should only be one source of truth for query state and it should match the app query state. Firing the query should be restricted to when the user is ready, but should not prevent the app query from changing.
```
kavilla, line 78:
```suggestion
  // TODO: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8801
```

kavilla: added an issue for this

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -74,15 +75,26 @@ export const QueryEditorUI: React.FC<Props> = (props) => {
...
   const queryString = getQueryService().queryString;
   const languageManager = queryString.getLanguageService();
   const extensionMap = languageManager.getQueryEditorExtensionMap();
   const services = props.opensearchDashboards.services;
+  const { query } = useQueryStringManager({
+    queryString,
+  });
+  const queryRef = useRef(query);
+
+  // Monaco editor commands dont reference the query state directly since the commands are registered at startup, so we need to keep a ref to the query state that it can use when needed
```
kavilla, line 90:
```suggestion
  // Monaco commands are registered once at startup, we need a ref to access the latest query state inside command callbacks
```

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -155,13 +165,14 @@ export const QueryEditorUI: React.FC<Props> = (props) => {
   const onQueryStringChange = (value: string) => {
     onChange({
       query: value,
-      language: props.query.language,
-      dataset: props.query.dataset,
+      // Its fiune to reference the query state her since it is not updated yet
```
kavilla, line 168:
```suggestion
```

kavilla (approved):

kavilla (approved):

SuZhou-Joe (approved):

PR #8806, opened 2024-11-04
[Discover] Decouple `data$` updates to prevent rows clearing on hook re-render

### LINKED ISSUE #8552: [BUG] Discover runs query again when adding or removing selected fields
**Describe the bug**

In Discover, adding or removing a field from "Selected fields" on the left side bar will trigger the query to run again. It's an acceptable behavior but doesn't work well for async queries which are slow to run.

**To Reproduce**
1. Go to https://playground.opensearch.org/app/data-explorer/discover#
2. query anything
3. open browser network tab, clear
4. on the left, drag a field to "Selected fields"
5. See two APIs triggered, the first one calls `PUT /api/saved_objects/index-pattern/<id>` and received 403. There is also a `POST /internal/search/opensearch-with-long-numerals` which reruns the query 
image

**Expected behavior**
Do not rerun the query unless necessary

**OpenSearch Version**
playground

**Dashboards Version**
playground

**Plugins**

**Screenshots**

**Additional context**

joshuali925:
### Description

While https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8552 is fixed by not letting query automatically run again, updating selected field still clears all rows in discover. This happens because any re-render to hook `useSearch` will clear `data$` by creating a new observable.

This PR fixes it by decoupling update logic based on dependencies.

### Issues Resolved

#8552 

## Screenshot

No UI change

## Testing the changes

i don't see UT for `useSearch`, @LDrago27 could you take a look if they will break anything?

## Changelog

- fix: decouple data$ updates to prevent rows clearing on hook re-render

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ashwin-pc:
@joshuali925 can you add a UT atleast for your part of the change?

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -143,8 +144,45 @@ export const useSearch = (services: DiscoverViewServices) => {
...
+
+  useEffect(() => {
+    data$.next({
+      ...data$.value,
+      status:
+        shouldSearchOnPageLoad() && !skipInitialFetch.current
+          ? ResultStatus.LOADING
+          : ResultStatus.UNINITIALIZED,
+    });
+  }, [data$, shouldSearchOnPageLoad, skipInitialFetch]);
```
sejli, line 184: Is there a reason why this is separated into two `useEffect()`'s? If mimicking the original `useMemo()`, could be done in a single update and reduce the number of `useEffects()`.

joshuali925: startTime updates should only update `queryStatus.startTime` and not reset `status`. it's the bug this PR is trying to fix, we should decouple queryStatus and status updates

virajsanghvi (approved):

sejli (approved):

PR #8818, opened 2024-11-06
Add framework to show related indexed views inside dataset selector and show banner in the Discover canvas for the selected dataset 

riysaxen-amzn:
### Description

This has all the changes from Base PR: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8701 

```

In this PR we add support for

Selecting indexed view inside dataset selector. The indexed view is an Opensearch index which can be created for an external source like S3. When user selects an indexed view, we use this index for querying the data
Showing a banner at the top of Discover results canvas. This banner can be used to show actions related to the selected dataset. The first use we will add for this banner would be to show a callout to create an indexed view while searching an external source, or when an indexed view is selected, we can show an entry point to select other indexed view
```

and UTs for Configurator 

```

 PASS  src/plugins/data/public/ui/dataset_selector/configurator.test.tsx (7.025 s)
  Configurator Component
    ✓ should render the component with the correct title and description (153 ms)
    ✓ should call onCancel when cancel button is clicked (83 ms)
    ✓ should call onPrevious when previous button is clicked (72 ms)
    ✓ should disable the submit button when conditions are not met (70 ms)
    ✓ should update state correctly when language is selected (68 ms)
    ✓ should fetch indexed views on mount and display them (53 ms)
    ✓ should initialize selectedLanguage with the current language from queryString (46 ms)
    ✓ should default selectedLanguage to the first language if currentLanguage is not supported (39 ms)
    ✓ should display the supported language dropdown correctly (41 ms)
    ✓ should disable the confirm button when submit is disabled (151 ms)

Tests:       10 passed, 10 total
Snapshots:   0 total
Time:        7.964 s, estimated 8 s
Ran all test suites matching /src\/plugins\/data\/public\/ui\/dataset_selector\/configurator.test.tsx/i.

```

## Screenshot

iv1
iv2

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/public/ui/dataset_selector/configurator.test.tsx
```diff
@@ -0,0 +1,182 @@
+/*
```
virajsanghvi, line 1: Are the lines called out as missed by codecov in configurator.tsx not considered critical?

virajsanghvi (commented):
Is there a plan to add additional tests (primarily concerned about for canvas/index.ts)?

Is the idea to rebase this after https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8701 is merged?

src/plugins/data/public/query/query_string/dataset_service/types.ts
```diff
@@ -16,6 +16,14 @@ export interface DataStructureFetchOptions {
   paginationToken?: string;
 }
 
+export interface IndexedView {
```
AMoo-Miki, line 19: Please prefix with `Dataset`

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -39,13 +39,13 @@ export const Configurator = ({
 }) => {
   const queryService = getQueryService();
   const queryString = queryService.queryString;
-  const languageService = queryService.queryString.getLanguageService();
+  const languageService = queryService.queryString?.getLanguageService();
```
virajsanghvi, line 42: Can we not mock languageService over introducing optional chaining throughout this file?

AMoo-Miki:
I would rather have these combined with https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8701 to avoid conflicts and the trouble of keeping in sync.

src/plugins/data/public/ui/dataset_selector/configurator.test.tsx
```diff
@@ -0,0 +1,317 @@
...
+const mockGetQueryService = jest.fn();
+const mockGetIndexPatterns = jest.fn();
+const mockQueryService = {
+  queryString: {
+    getLanguageService: jest.fn().mockReturnValue({
+      getLanguage: jest.fn().mockImplementation((langId) => ({
+        title:
+          {
+            en: 'English',
+            fr: 'French',
```
AMoo-Miki, line 23: I don't think this is the "language" `getLanguageService` provides. @kavilla am i wrong?

ashwin-pc: Lets combine the 2 Pr's so that they are contextually together. Want to ensure that the tests are covering the right functionality

src/plugins/data/public/query/query_string/dataset_service/types.ts
```diff
@@ -82,4 +96,14 @@ export interface DatasetTypeConfig {
    * Returns a list of sample queries for this dataset type
    */
   getSampleQueries?: (dataset: Dataset, language: string) => any;
+  /**
+   * Service used for indexedViews related operations
+   */
+  indexedViewsService?: DatasetIndexedViewsService;
```
kavilla, line 102: is this the right name? i don tthink the config should define the service. should it not be a service defined?

src/plugins/data/public/query/query_string/dataset_service/types.ts
riysaxen-amzn: addressed

src/plugins/data/public/ui/dataset_selector/configurator.tsx
riysaxen-amzn: addressed

src/plugins/data/public/ui/dataset_selector/configurator.test.tsx
riysaxen-amzn: addressed

changelogs/fragments/8818.yml
```diff
@@ -0,0 +1,3 @@
+feat:
+- Add framework to show banner at the top in discover results canvas ([#8701](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8818))
```
virajsanghvi, line 2: think you also want to update 8701

virajsanghvi (dismissed):
Re-running ci to see if some checks will pass (different set than last time) but approving to unblock.

Will want to follow up and correct the pr number in the changelog fragment

changelogs/fragments/8818.yml
riysaxen-amzn: updated

virajsanghvi (dismissed):
Checked screenshot for cigroup6, refresh button was present so likely a delay in showing up

src/plugins/data/common/datasets/types.ts
```diff
@@ -256,6 +258,29 @@ export interface DatasetField {
...
+export enum DatasetCanvasBannerComponentType {
+  CALLOUT = 'callout',
+  CUSTOM = 'custom',
+}
+
+export type DatasetCanvasBannerProps =
+  | {
+      componentType: DatasetCanvasBannerComponentType.CALLOUT;
+      title: string;
+      iconType?: string;
```
kavilla, line 270: how come iconType vs the actual icon following the current pattern: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/common/datasets/types.ts#L155

src/plugins/data/common/datasets/types.ts
```diff
@@ -256,6 +258,29 @@ export interface DatasetField {
...
 }
 
+export enum DatasetCanvasBannerComponentType {
+  CALLOUT = 'callout',
+  CUSTOM = 'custom',
+}
+
+export type DatasetCanvasBannerProps =
+  | {
+      componentType: DatasetCanvasBannerComponentType.CALLOUT;
```
kavilla, line 268: if my understanding of this is correct. This should just be `type` as it is redundant if this component is already a component and developers will understand that the `type` is associated to the type of component.

src/plugins/data/common/datasets/types.ts
```diff
@@ -256,6 +258,29 @@ export interface DatasetField {
   // TODO:  osdFieldType?
 }
 
+export enum DatasetCanvasBannerComponentType {
+  CALLOUT = 'callout',
```
kavilla, line 262: thanks for making this a string. but could we follow current pattern could we make this all caps: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/common/datasets/types.ts#L136

might be a nitpick but it's easier to write code expecting strings to be in a certain casing

kavilla, line 262:
```suggestion
  CALLOUT = 'CALLOUT',
```

src/plugins/data/common/datasets/types.ts
```diff
@@ -256,6 +258,29 @@ export interface DatasetField {
...
+  CUSTOM = 'custom',
```
kavilla, line 263:
```suggestion
  CUSTOM = 'CUSTOM',
```

src/plugins/data/common/datasets/types.ts
kavilla, line 268:
```suggestion
      type: DatasetCanvasBannerComponentType.CALLOUT;
```

src/plugins/data/public/index.ts
```diff
@@ -493,6 +493,7 @@ export {
   QueryStart,
   PersistedLog,
   LanguageReference,
+  IndexedViewsService,
```
AMoo-Miki, line 496: Didn't we rename this to `DatasetIndexedViewsService`?

src/plugins/data/public/query/query_string/index.ts
```diff
@@ -34,6 +34,7 @@ export {
   DatasetService,
   DatasetServiceContract,
   DatasetTypeConfig,
+  DatasetIndexedViewsService as IndexedViewsService,
```
AMoo-Miki, line 37: The whole point of prefixing is to avoid name collision. Exporting it without a prefix out of here doesn't mitigate that.

src/plugins/data/common/datasets/types.ts
```diff
@@ -256,6 +258,29 @@ export interface DatasetField {
...
+}
+
+export type DatasetCanvasBannerProps =
```
kavilla, line 266: i think we should have a `BaseDatasetCanvasBannerProps`  or `DatasetCanvasBannerProps` that has required props to add a new banner. Then create a type called `DatasetCanvasCalloutBannerProps` and `DatasetCanvasCustomBannerProps` or something.

if developing these union types become quiet difficult to understand sometimes and caused the linter to complain because the type isn't strict enough. easier to work with explicit types.

as far as the base props i think we should have at least 
```typescript
export type DatasetCanvasBannerProps = {
 // required. unique id, this will be used to associate a dataset to a canvas props. this is different than title because title is something that could be displayed but if i have two devs trying to add a new banner called `Index Callout Banner` (haven't looked at the code yet) but if the component doesn't change in an ID then maybe we won't render an update by accident.
  id: string;
  title: string; // required. going from optional to required would be a breaking change. and haven't looked at how this is used but i can imagine we don't really want to have a case where title is undefined.

File: src/plugins/data/common/datasets/types.ts [hunk 14]
```diff
@@ -256,6 +258,29 @@ export interface DatasetField {
   // TODO:  osdFieldType?
 }
 
+export enum DatasetCanvasBannerComponentType {
```
kavilla, line 261:
```suggestion
export enum DATASET_CANVAS_BANNER_COMPONENT_TYPES {
```

src/plugins/data/common/datasets/types.ts
kavilla: that or `title` to be `name` similar to the flyout stuff. 

we should also consider maybe `key` but we should def include
```
className?: string;
'data-test-subj'?: string;
```

writing tests on this abstract types would make this so much easier and if we have to style them the className will help

src/plugins/data/common/datasets/types.ts
```diff
@@ -247,6 +247,8 @@ export interface Dataset extends BaseDataset {
   timeFieldName?: string;
   /** Optional language to default to from the language selector */
   language?: string;
+  /** Optional name of the indexed view to search data from */
+  indexedView?: string;
```
kavilla, line 251: could you add more insight on what this is?
is this to be the id of the indexedView component. is this a unique id generated from something?

cc: @virajsanghvi curious your opinion here

without much context this for me this seems something like a data view is that correct @riysaxen-amzn ?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -40,10 +41,21 @@ export const Configurator = ({
   const queryString = queryService.queryString;
   const languageService = queryService.queryString.getLanguageService();
   const indexPatternsService = getIndexPatterns();
-  const type = queryString.getDatasetService().getType(baseDataset.type);
+  const type = queryString?.getDatasetService().getType(baseDataset.type);
```
kavilla, line 44: queryString manager shouldn't be optional how come? the type might be undefined but the app won't work if queryString is undefined

ruanyl:
@riysaxen-amzn FYI, noticed there is a conflict :)

PR #8842, opened 2024-11-11
[Workspace]Fix error toasts in sample data page

wanglam:
### Description

This PR addresses the issue of error toasts appearing on the sample data page when installing or uninstalling sample data. The root cause was that the user did not have permission to update UI settings at the workspace level. The changes in this PR include:

1. Remove saved objects first when uninstalling sample data.
2. Disable the behavior of updating the default index pattern UI setting for non workspace admin when installing or uninstalling sample data inside a workspace.

## Screenshot

No UI Changes

## Testing the changes

1. Clone the branch and run `yarn osd bootstrap --single-version loose`.
2. Add the following configurations to `config/opensearch_dashboards.yml`:
   ```
   workspace.enabled: true
   uiSettings:
     overrides:
       'home:useNewHomePage': true
   ```
3. Run `yarn start --no-base-path`.
4. Log in with the admin user and create a new workspace named "test1".
5. Visit the sample data page of the "test1" workspace.
6. Click the "Add data" and "Remove" buttons on the sample data cards to ensure you can install and uninstall sample data without any issues.
7. Create a new user with the "admin" backend_role in the "Internal users" page.
8. Assign the "Read and write" access level to the new user inside the "test1" workspace.
9. Open an incognito window and log in with the new user.
10. Visit the sample data page of the "test1" workspace and click the "Add data" button. Verify that no error toasts appear.
11. Switch back to the admin user tab, visit the same workspace, and set the "flights" sample data as the default index pattern in the index patterns page.
12. Switch to the new user tab and click the "Remove" button on the "flights" sample data card. Verify that an error toast appears, similar to the provided screenshot.
image
13. Click the "Remove" buttons on other sample data cards and ensure no error toasts appear.
14. Switch to the admin user tab and set the new user as an "Admin" of the "test1" workspace.
15. Switch to the new user tab and click the "Remove" button on the "flights" sample data card. Verify that no error toasts appear.
16. Test with the workspace disabled (for regression testing):
    - Remove the newly added configurations.
    - Visit the home page and click the "Add sample data" button.
    - Click the "Add data" buttons. Sample data should be installed without error toasts.
    - Click the "Remove" buttons. Sample data should be uninstalled without error toasts.

## Changelog

- fix: [Workspace]Fix error toasts in sample data page

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/home/public/application/sample_data_client.js
+++ b/src/plugins/home/public/application/sample_data_client.js
@@ -41,16 +41,26 @@ export async function listSampleDataSets(dataSourceId) {
   return await getServices().http.get(sampleDataUrl, { query });
 }
 
-const isWorkspaceEnabled = () => {
-  const workspaces = getServices().application.capabilities.workspaces;
-  return !!(workspaces && workspaces.enabled);
+const canUpdateUISetting = () => {
+  const {
+    application: { capabilities },
+    workspaces,
+  } = getServices();
+  if (
+    capabilities.workspaces &&
+    capabilities.workspaces.enabled &&
+    capabilities.workspaces.permissionEnabled
+  ) {
+    return !!workspaces?.currentWorkspace$.getValue()?.owner;
+  }
+  return true;
 };
 
 export async function installSampleDataSet(id, sampleDataDefaultIndex, dataSourceId) {
   const query = buildQuery(dataSourceId);
   await getServices().http.post(`${sampleDataUrl}/${id}`, { query });
 
-  if (!isWorkspaceEnabled() && getServices().uiSettings.isDefault('defaultIndex')) {
+  if (canUpdateUISetting() && getServices().uiSettings.isDefault('defaultIndex')) {
     getServices().uiSettings.set('defaultIndex', sampleDataDefaultIndex);
   }
 
@@ -64,7 +74,7 @@ export async function uninstallSampleDataSet(id, sampleDataDefaultIndex, dataSou
   const uiSettings = getServices().uiSettings;
 
   if (
-    !isWorkspaceEnabled() &&
+    canUpdateUISetting() &&
     !uiSettings.isDefault('defaultIndex') &&
     uiSettings.get('defaultIndex') === sampleDataDefaultIndex
   ) {
--- a/src/plugins/home/public/application/sample_data_client.js
+++ b/src/plugins/home/public/application/sample_data_client.js
@@ -41,11 +41,16 @@ export async function listSampleDataSets(dataSourceId) {
   return await getServices().http.get(sampleDataUrl, { query });
 }
 
+const isWorkspaceEnabled = () => {
+  const workspaces = getServices().application.capabilities.workspaces;
+  return !!(workspaces && workspaces.enabled);
+};
+
 export async function installSampleDataSet(id, sampleDataDefaultIndex, dataSourceId) {
   const query = buildQuery(dataSourceId);
   await getServices().http.post(`${sampleDataUrl}/${id}`, { query });
 
-  if (getServices().uiSettings.isDefault('defaultIndex')) {
+  if (!isWorkspaceEnabled() && getServices().uiSettings.isDefault('defaultIndex')) {
     getServices().uiSettings.set('defaultIndex', sampleDataDefaultIndex);
   }
 
@@ -59,6 +64,7 @@ export async function uninstallSampleDataSet(id, sampleDataDefaultIndex, dataSou
   const uiSettings = getServices().uiSettings;
 
   if (
+    !isWorkspaceEnabled() &&
     !uiSettings.isDefault('defaultIndex') &&
     uiSettings.get('defaultIndex') === sampleDataDefaultIndex
   ) {
--- a/src/plugins/home/server/services/sample_data/routes/uninstall.ts
+++ b/src/plugins/home/server/services/sample_data/routes/uninstall.ts
@@ -62,27 +62,10 @@ export function createUninstallRoute(
         return response.notFound();
       }
 
-      const caller = dataSourceId
-        ? context.dataSource.opensearch.legacy.getClient(dataSourceId).callAPI
-        : context.core.opensearch.legacy.client.callAsCurrentUser;
-
-      for (let i = 0; i < sampleDataset.dataIndices.length; i++) {
-        const dataIndexConfig = sampleDataset.dataIndices[i];
-        const index =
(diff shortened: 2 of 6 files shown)


SuZhou-Joe (dismissed):

src/plugins/home/public/application/sample_data_client.js
```diff
@@ -41,11 +41,16 @@ export async function listSampleDataSets(dataSourceId) {
   return await getServices().http.get(sampleDataUrl, { query });
 }
 
+const isWorkspaceEnabled = () => {
```
SuZhou-Joe, line 44: I am thinking if it is workspaceEnabled or permissionControlEnabled.

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -144,6 +148,59 @@ export class WorkspaceUiSettingsClientWrapper {
       return wrapperOptions.client.update(type, id, attributes, options);
     };
 
+    const deleteUiSettingsWithWorkspace = async (
```
SuZhou-Joe, line 151:
```suggestion
    const deleteWithDefaultIndexPatternCheck = async (
```

src/plugins/home/public/application/sample_data_client.js
```diff
@@ -41,11 +41,16 @@ export async function listSampleDataSets(dataSourceId) {
...
+  const workspaces = getServices().application.capabilities.workspaces;
+  return !!(workspaces && workspaces.enabled);
+};
+
 export async function installSampleDataSet(id, sampleDataDefaultIndex, dataSourceId) {
   const query = buildQuery(dataSourceId);
   await getServices().http.post(`${sampleDataUrl}/${id}`, { query });
 
-  if (getServices().uiSettings.isDefault('defaultIndex')) {
+  if (!isWorkspaceEnabled() && getServices().uiSettings.isDefault('defaultIndex')) {
```
SuZhou-Joe, line 53: Do we still need to check the feature flag? I am thinking line 51 will throw permission error if user happens to delete the default index pattern.

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -144,6 +148,59 @@ export class WorkspaceUiSettingsClientWrapper {
...
+          id: workspaceId,
+        }))
+      );
+
+      const defaultIndexPatternWorkspaces = workspaceObjects.saved_objects.filter(
+        ({ attributes }) => attributes?.uiSettings?.[DEFAULT_INDEX_PATTERN_UI_SETTINGS_ID] === id
+      );
+      try {
+        await this.getWorkspaceTypeEnabledClient(wrapperOptions.request).bulkUpdate(
+          defaultIndexPatternWorkspaces.map((savedObject) => ({
```
Hailong-am, line 180: can we check defaultIndexPatternWorkspaces is not empty first to reduce unnecessary workspace update? 

also will bulkUpdate an empty list throw errors?

wanglam: The `bulkUpdate` won't throw errors for empty list. But you're right, we can add a empty check logic for it.

SuZhou-Joe (dismissed):

src/plugins/home/server/services/sample_data/routes/uninstall.ts
```diff
@@ -66,6 +66,40 @@ export function createUninstallRoute(
...
+      let savedObjectsList = getFinalSavedObjects({
+        dataset: sampleDataset,
+        workspaceId,
+        dataSourceId,
+      });
+
+      /**
+       * Try to delete index patterns first to avoid partial delete issues.
+       * This is necessary because default index patterns cannot be deleted
+       * when a workspace is enabled and the user lacks the required permissions.
```
ruanyl, line 78: > This is necessary because default index patterns cannot be deleted
> when a workspace is enabled and the user lacks the required permissions.

I might miss something here, do we have special check on index pattern when deleting?

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -144,6 +148,61 @@ export class WorkspaceUiSettingsClientWrapper {
       return wrapperOptions.client.update(type, id, attributes, options);
     };
 
+    const deleteWithDefaultIndexPatternCheck = async (
```
ruanyl, line 151: I'm a bit concern that we're over complicating this saved object wrapper, it's intended to be a wrapper for `config` type of object, but now we're doing an implicit cascading updating when a `index-pattern` is deleted.
Could you please elaborate why this change is necessary?

src/plugins/home/server/services/sample_data/routes/uninstall.ts
wanglam: In this PR, we will delete index patterns of sample data first when user want to uninstall sample data. So there are different cases when user try to uninstall sample data.
- Case 1: Index pattern is default index pattern and user is `dashboards admin` or `workspace admin`, Index pattern can be removed
- Case 2: Index pattern is default index pattern and user is `workspace read and write` and `workspace read only`, index pattern can't be removed
- Case 3: Index pattern is not default index pattern and user is `dashboards admin`, `workspace admin` or `workspace read & write`, index pattern can be removed
- Case 4: Index pattern is not default index pattern and user is `workspace read only`, index pattern can't be removed
This PR will enhance case 1 and case 2, will try to remove ui settings first when deleted index patterns are default index pattern.

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
wanglam: You are right, we can keep the default index pattern validation logic in the ui setting PR when trying to delete index patterns. Then it will throw an error if a user tries to remove a default index pattern. The benefits of leaving the cascading updating logic were that the user doesn't need to call the UI setting separately. I'm open to this. After this change, the user will get an error when trying to uninstall sample data and the index pattern is a default index pattern. Do you think if we should remove the default index patterns ui setting before sample data uninstall?

ruanyl (approved):

Hailong-am (approved):

PR #8843, opened 2024-11-11
[Autocomplete] PPL table name parser fix

paulstn:
### Description

Every file under `.generated` can be ignored

There were two changes made. 
- The first one was to the PPL grammar file, reverting back a fix to prevent the table name being suggested twice and ultimately preventing any further suggestions from appearing.
- The second change was in the [opensearch_ppl_autocomplete.ts](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8843/files#diff-af0a2c7eded6acde6a93d6221337c7e8038a5fff28e08f917bdd80fd045d390d) file, doing the fix the first change handled by instead intercepting a deeper rule in the grammar, preventing the table from incorrectly being suggested. The code completion engine was hitting the first table rule twice, possibly due to an option where the lexer rule allows for whitespace.

Ultimately, this helps to solve an issue where correctly formatted table names (with backticks, etc.) prevented other suggestions from appearing. The original grammar would've parsed it correctly.

## Changelog

- fix: PPL table name parser

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/public/antlr/opensearch_ppl/grammar/OpenSearchPPLParser.g4
```diff
@@ -775,11 +775,11 @@ valueList
    ;
 
 qualifiedName
-   : ID // (DOT ident)* # identsAsQualifiedName
+   : ID // ident (DOT ident)* # identsAsQualifiedName
    ;
 
 tableQualifiedName
-   : ID // tableIdent (DOT ident)* # identsAsTableQualifiedName
+   : tableIdent (DOT ident)* # identsAsTableQualifiedName
```
joshuali925, line 782: would issues from https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8087#discussion_r1765952671 come back?

src/plugins/data/public/antlr/opensearch_ppl/opensearch_ppl_autocomplete.ts
```diff
@@ -99,7 +99,7 @@ export function processVisitedRules(
         shouldSuggestColumns = true;
         break;
       }
-      case OpenSearchPPLParser.RULE_tableQualifiedName: {
+      case OpenSearchPPLParser.RULE_tableIdent: {
```
joshuali925, line 102: what does this do?

joshuali925 (approved):

src/plugins/data/public/antlr/opensearch_ppl/grammar/OpenSearchPPLParser.g4
paulstn: The issue I mentioned in that comment is what the change I've done in `opensearch_ppl_autocomplete.ts` should handle, since we can prevent the grammar from causing the code completion engine to stay on that rule, thereby suggesting a table name twice.

joshuali925:
there are some issues with CI right now

PR #8798, opened 2024-11-03
Added UTs for Saved queries new UI

riysaxen-amzn:
### Description

Created UTs for Saved queries new UI(https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8469)

## Testing the changes

```
yarn test:jest src/plugins/data/public/query/saved_query/saved_query_service.test.ts
yarn run v1.22.22
$ scripts/use_node scripts/jest src/plugins/data/public/query/saved_query/saved_query_service.test.ts
jest-haste-map: duplicate manual mock found: index
  The following files share their name; please delete one of them:
    * <rootDir>/src/plugins/vis_type_vega/public/expressions/__mocks__/index.ts
    * <rootDir>/src/core/server/saved_objects/import/__mocks__/index.ts

 PASS  src/plugins/data/public/query/saved_query/saved_query_service.test.ts (5.786 s)
  saved query service
    saveQuery
      ✓ should create a saved object for the given attributes (3 ms)
      ✓ should allow overwriting an existing saved query (1 ms)
      ✓ should optionally accept filters and timefilters in object format (1 ms)
      ✓ should throw an error when saved objects client returns error
      ✓ should throw an error if the saved query does not have a title (1 ms)
      ✓ should include dataset in the query when query enhancement is enabled and dataset exists (1 ms)
      ✓ should include isTemplate in the query object when new saved queries UI is enabled and isTemplate is true
    findSavedQueries
      ✓ should find and return saved queries without search text or pagination parameters (1 ms)
      ✓ should return the total count along with the requested queries
      ✓ should find and return saved queries with search text matching the title field (1 ms)
      ✓ should find and return parsed filters and timefilters items
      ✓ should return an array of saved queries
      ✓ should accept perPage and page properties (1 ms)
      ✓ should correctly parse a json query string
      ✓ should correctly parse a json object
      ✓ should handle null string with single quote
      ✓ should handle null string with double quote
      ✓ should handle null quoted string
      ✓ should not lose double quotes
    getSavedQuery
      ✓ should retrieve a saved query by id (1 ms)
      ✓ should only return saved queries
    deleteSavedQuery
      ✓ should delete the saved query for the given ID
    getAllSavedQueries
      ✓ should return all the saved queries (1 ms)
    getSavedQueryCount
      ✓ should return the total number of saved queries

```

```
yarn test:jest src/plugins/data/public/plugin.*                                     
yarn run v1.22.22
[...]

PR #8848, opened 2024-11-12
Fix template queries loading and update getSampleQuery interface

jowg-amazon:
### Description
- Instead of loading template queries through saved objects api, this pr changes it so that we can load template queries in through the getSampleQuery function directly
- Updates URL state when opening a saved query so datapicker state is correct
- Updates getSampleQuery interface so it can also return a promise

```
 PASS  src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.test.tsx
  OpenSavedQueryFlyout
    ✓ should render the flyout with correct tabs and content (58 ms)
    ✓ should filter saved queries based on search input (29 ms)
    ✓ should select a query when clicking on it and enable the "Open query" button (14 ms)
    ✓ should call handleQueryDelete when deleting a query (21 ms)
    ✓ should handle pagination controls correctly (43 ms)

Tests:       5 passed, 5 total
Snapshots:   0 total
Time:        3.351 s
Ran all test suites matching /src\/plugins\/data\/public\/ui\/saved_query_flyouts\/open_saved_query_flyout.test.tsx/i.
```

Screenshot 2024-11-11 at 10 52 01 PM
Screenshot 2024-11-11 at 10 51 54 PM
Screenshot 2024-11-12 at 9 23 49 AM
Screenshot 2024-11-11 at 10 51 48 PM
Screenshot 2024-11-11 at 10 51 42 PM

https://github.com/user-attachments/assets/c43d7a51-4740-4c85-8e64-bfc2281db62f

## Changelog

- fix: Fix template queries loading and update getSampleQuery interface

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/dataset_service/types.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/types.ts
@@ -83,5 +83,5 @@ export interface DatasetTypeConfig {
   /**
    * Returns a list of sample queries for this dataset type
    */
-  getSampleQueries?: (dataset: Dataset, language: string) => any;
+  getSampleQueries?: (dataset?: Dataset, language?: string) => Promise<any> | any;
 }
--- a/src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
+++ b/src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
@@ -13,6 +13,7 @@ import {
   EuiFlyoutBody,
   EuiFlyoutFooter,
   EuiFlyoutHeader,
+  EuiLoadingSpinner,
   EuiSearchBar,
   EuiSearchBarProps,
   EuiSpacer,
@@ -62,10 +63,11 @@ export function OpenSavedQueryFlyout({
   const [languageFilterOptions, setLanguageFilterOptions] = useState<string[]>([]);
   const [selectedQuery, setSelectedQuery] = useState<SavedQuery | undefined>(undefined);
   const [searchQuery, setSearchQuery] = useState(EuiSearchBar.Query.MATCH_ALL);
-
+  const [isLoading, setIsLoading] = useState(false);
   const currentTabIdRef = useRef(selectedTabId);
 
   const fetchAllSavedQueriesForSelectedTab = useCallback(async () => {
+    setIsLoading(true);
     if (queryStringManager.getQuery()?.dataset?.type === 'SECURITY_LAKE') {
       setHasTemplateQueries(true);
     }
@@ -85,6 +87,7 @@ export function OpenSavedQueryFlyout({
         if (Array.isArray(templateQueries)) setSavedQueries(templateQueries);
       }
     }
+    setIsLoading(false);
   }, [savedQueryService, currentTabIdRef, setSavedQueries, queryStringManager]);
 
   const updatePageIndex = useCallback((index: number) => {
@@ -194,7 +197,13 @@ export function OpenSavedQueryFlyout({
         onChange={onChange}
       />
       <EuiSpacer />
-      {queriesOnCurrentPage.length > 0 ? (
+      {isLoading ? (
+        <EuiFlexGroup justifyContent="center" alignItems="center" style={{ height: '200px' }}>
+          <EuiFlexItem grow={false}>
+            <EuiLoadingSpinner size="xl" />
+          </EuiFlexItem>
+        </EuiFlexGroup>
+      ) : queriesOnCurrentPage.length > 0 ? (
         queriesOnCurrentPage.map((query) => (
           <SavedQueryCard
             key={query.id}
@@ -220,7 +229,7 @@ export function OpenSavedQueryFlyout({
         />
       )}
       <EuiSpacer />
-      {queriesOnCurrentPage.length > 0 && (
+      {!isLoading && queriesOnCurrentPage.length > 0 && (
         <EuiTablePagination
           itemsPerPageOptions={[5, 10, 20]}
           itemsPerPage={itemsPerPage}
--- a/src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
+++ b/src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
@@ -23,14 +23,16 @@ import {
 } from '@elastic/eui';
 import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { i18n } from '@osd/i18n';
-import { SavedQuery, SavedQueryService } from '../../query';
+import { QueryStringManager, SavedQuery, SavedQueryService } from '../../query';
 import { SavedQueryCard } from './saved_query_card';
+import { Query } from '../../../common';
 
 export interface OpenSavedQueryFlyoutProps {
   savedQueryService: SavedQueryService;
   onClose: () => void;
   onQueryOpen: (query: SavedQuery) => void;
   handleQueryDelete: (query: SavedQuery) => Promise<void>;
+  queryStringManager: QueryStringManager;
 }
 
 interface SavedQuerySearchableItem {
@@ -47,6 +49,7 @@ export function OpenSavedQueryFlyout({
   onClose,
   onQueryOpen,
   handleQueryDelete,
+  queryStringManager,
 }: OpenSavedQueryFlyoutProps) {
   const [selectedTabId, setSelectedTabId] = useState<string>('mutable-saved-queries');
   const [savedQueries, setSavedQueries] = useState<SavedQuery[]>([]);
@@ -60,17 +63,29 @@ export function OpenSavedQueryFlyout({
   const [selectedQuery, setSelectedQuery] = useState<SavedQuery | undefined>(undefined);
   const [searchQuery, setSearchQuery] = useState(EuiSearchBar.Query.MATCH_ALL);
 
+  const currentTabIdRef = useRef(selectedTabId);
+
   const fetchAllSavedQueriesForSelectedTab = useCallback(async () => {
-    const allQueries = await savedQueryService.getAllSavedQueries();
-    const templateQueriesPresent = allQueries.some((q) => q.attributes.isTemplate);
-    const queriesForSelectedTab = allQueries.filter(
-      (q) =>
-        (selectedTabId === 'mutable-saved-queries' && !q.attributes.isTemplate) ||
-        (selectedTabId === 'template-saved-queries' && q.attributes.isTemplate)
-    );
-    setSavedQueries(queriesForSelectedTab);
-    setHasTemplateQueries(templateQueriesPresent);
-  }, [savedQueryService, selectedTabId, setSavedQueries]);
+    if (queryStringManager.getQuery()?.dataset?.type === 'SECURITY_LAKE') {
+      setHasTemplateQueries(true);
+    }
+    if (currentTabIdRef.current === 'mutable-saved-queries') {
+      const allQueries = await savedQueryService.getAllSavedQueries();
+      const mutableSavedQueries = allQueries.filter((q) => !q.attributes.isTemplate);
+      if (currentTabIdRef.current === 'mutable-saved-queries') {
+        setSavedQueries(mutableSavedQueries);
+      }
+    } else if (currentTabIdRef.current === 'template-saved-queries') {
+      const query = queryStringManager.getQuery();
+      if (query?.dataset?.type) {
+        const templateQueries = await queryStringManager
+          .getDatasetService()
+          ?.getType(query.dataset.type)
+          ?.getSampleQueries?.();
+        if (Array.isArray(templateQueries)) setSavedQueries(templateQueries);
+      }
+    }
+  }, [savedQueryService, currentTabIdRef, setSavedQueries, queryStringManager]);
 
   const updatePageIndex = useCallback((index: number) => {
     pager.current.goToPageIndex(index);
@@ -252,6 +267,7 @@ export function OpenSavedQueryFlyout({
           initialSelectedTab={tabs[0]}
           onTabClick={(tab) => {
             setSelectedTabId(tab.id);
+            currentTabIdRef.current = tab.id;
           }}
         />
       </EuiFlyoutBody>
@@ -268,7 +284,19 @@ export function OpenSavedQueryFlyout({
               fill
               onClick={() => {
                 if (selectedQuery) {
-                  onQueryOpen(selectedQuery);
+                  if (
+                    // Template queries are not associated with data sources. Apply data source from current query
+                    selectedQuery.attributes.isTemplate
+                  ) {
+                    const updatedQuery: Query = {
+                      ...queryStringManager?.getQuery(),
+                      query: selectedQuery.attributes.query.query,
+                      language: selectedQuery.attributes.query.language,
+                    };
+                    queryStringManager.setQuery(updatedQuery);
+                  } else {
+                    onQueryOpen(selectedQuery);
+                  }
                   onClose();
                 }
               }}
--- a/src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
+++ b/src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
@@ -68,9 +68,23 @@ export function OpenSavedQueryFlyout({
 
   const fetchAllSavedQueriesForSelectedTab = useCallback(async () => {
     setIsLoading(true);
-    if (queryStringManager.getQuery()?.dataset?.type === 'SECURITY_LAKE') {
-      setHasTemplateQueries(true);
+    const query = queryStringManager.getQuery();
+    let templateQueries: any[] = [];
+
+    // fetch sample query based on dataset type
+    if (query?.dataset?.type) {
+      templateQueries =
+        (await queryStringManager
+          .getDatasetService()
+          ?.getType(query.dataset.type)
+          ?.getSampleQueries?.()) || [];
+
+      // Check if any sample query has isTemplate set to true
+      const hasTemplates = templateQueries.some((q) => q?.attributes?.isTemplate);
+      setHasTemplateQueries(hasTemplates);
     }
+
+    // Set queries based on the current tab
     if (currentTabIdRef.current === 'mutable-saved-queries') {
       const allQueries = await savedQueryService.getAllSavedQueries();
       const mutableSavedQueries = allQueries.filter((q) => !q.attributes.isTemplate);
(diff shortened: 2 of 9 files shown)


src/plugins/data/public/query/query_string/dataset_service/types.ts
```diff
@@ -83,5 +83,5 @@ export interface DatasetTypeConfig {
   /**
    * Returns a list of sample queries for this dataset type
    */
-  getSampleQueries?: (dataset: Dataset, language: string) => any;
+  getSampleQueries?: (dataset?: Dataset, language?: string) => Promise<any> | any;
```
virajsanghvi, line 86: Have we updated all callers to handler promises with this interface change?

Pre-existing, but can we be more specific than any?

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
```diff
@@ -60,17 +63,29 @@ export function OpenSavedQueryFlyout({
   const [selectedQuery, setSelectedQuery] = useState<SavedQuery | undefined>(undefined);
   const [searchQuery, setSearchQuery] = useState(EuiSearchBar.Query.MATCH_ALL);
 
+  const currentTabIdRef = useRef(selectedTabId);
```
virajsanghvi, line 67: Can we add unit tests for this file?

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
```diff
@@ -60,17 +63,29 @@ export function OpenSavedQueryFlyout({
...
-    const templateQueriesPresent = allQueries.some((q) => q.attributes.isTemplate);
-    const queriesForSelectedTab = allQueries.filter(
-      (q) =>
-        (selectedTabId === 'mutable-saved-queries' && !q.attributes.isTemplate) ||
-        (selectedTabId === 'template-saved-queries' && q.attributes.isTemplate)
-    );
-    setSavedQueries(queriesForSelectedTab);
-    setHasTemplateQueries(templateQueriesPresent);
-  }, [savedQueryService, selectedTabId, setSavedQueries]);
+    if (queryStringManager.getQuery()?.dataset?.type === 'SECURITY_LAKE') {
```
virajsanghvi, line 71: Is Security Lake a specific aws type? Is there something more generic we could use or could we abstract this for any dataset type?

src/plugins/discover/public/application/components/no_results/no_results.tsx
```diff
@@ -174,6 +174,7 @@ export const DiscoverNoResults = ({ queryString, query, savedQuery, timeFieldNam
   // }
 
   const [savedQueries, setSavedQueries] = useState<SavedQuery[]>([]);
+  const [sampleQueries, setSampleQueries] = useState<any>([]);
```
virajsanghvi, line 177: can we add unit tests for this file?

virajsanghvi (commented):
Only skimmed so far, but posting initial comments

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
jowg-amazon: We only want this flyout to be shown when the data source picker is a security lake connection

src/plugins/discover/public/application/components/no_results/no_results.tsx
jowg-amazon: UT's for saved queries are all in this PR: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8798

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
jowg-amazon: UT's for saved queries are all in this PR: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8798

src/plugins/data/public/query/query_string/dataset_service/types.ts
jowg-amazon: Yes it's only being called in `no_results.tsx` and we modified it so that it handles the promise

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
virajsanghvi: @ashwin-pc @kavilla @AMoo-Miki Is this okay? Is there a better way of going about this?

Or can we track following up on implementing this in a generic way?

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
```diff
@@ -179,7 +197,13 @@ export function OpenSavedQueryFlyout({
         onChange={onChange}
       />
       <EuiSpacer />
-      {queriesOnCurrentPage.length > 0 ? (
+      {isLoading ? (
+        <EuiFlexGroup justifyContent="center" alignItems="center" style={{ height: '200px' }}>
```
virajsanghvi, line 201: nit: why are we hardcoding height here?

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
```diff
@@ -268,7 +293,19 @@ export function OpenSavedQueryFlyout({
...
+                  if (
+                    // Template queries are not associated with data sources. Apply data source from current query
+                    selectedQuery.attributes.isTemplate
+                  ) {
+                    const updatedQuery: Query = {
+                      ...queryStringManager?.getQuery(),
+                      query: selectedQuery.attributes.query.query,
+                      language: selectedQuery.attributes.query.language,
+                    };
+                    queryStringManager.setQuery(updatedQuery);
```
virajsanghvi, line 305: curious, why does this use setQuery over onQueryOpen?

src/plugins/discover/public/application/components/no_results/no_results.tsx
virajsanghvi: Same questions as above wrt to UTs. There looks like there's new functionality, so I'm curious how we're unit testing/validating those changes or why it doesn't matter right now.

src/plugins/discover/public/application/components/no_results/no_results.tsx
```diff
@@ -186,6 +187,39 @@ export const DiscoverNoResults = ({ queryString, query, savedQuery, timeFieldNam
...
+      }
+    }
+
+    // Samples for the dataset type
+    if (query?.dataset?.type) {
+      const datasetType = queryString.getDatasetService()?.getType(query.dataset.type);
+      if (datasetType?.getSampleQueries) {
+        const sampleQueriesResponse = datasetType.getSampleQueries(query.dataset, query.language);
+        if (Array.isArray(sampleQueriesResponse)) {
+          setSampleQueries([...sampleQueriesResponse, ...newSampleQueries]);
```
virajsanghvi, line 207: nit, could create a helper function for setting sample queries that gets used in both places to prevent divergence

src/plugins/discover/public/application/components/no_results/no_results.tsx
```diff
@@ -229,7 +244,7 @@ export const DiscoverNoResults = ({ queryString, query, savedQuery, timeFieldNam
                   <EuiSpacer size="s" />
                   {sampleQueries
                     .slice(0, 5)
-                    .map((sampleQuery) =>
+                    .map((sampleQuery: any) =>
```
virajsanghvi, line 247: why any? can we be more specific?

jowg-amazon: `any` was part of the preexisting logic. It wasn't adhering to any object and was just returning the title and the query. 
Ex: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/76cf823e0a52e101e3e9b5fbe03836ec529229da/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts#L94

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
ashwin-pc: Why are we hardcoding this? Isn't it just as easy to fetch it from the dataset config for the selected dataset?

jowg-amazon: Updated in this commit: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8848/commits/e14f0be53fb68cb09743a01f2acd7ef730712929. So that we check `isTemplate` to determine if the tab should be shown.

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
jowg-amazon: Added unit tests for this file in this commit: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8848/commits/c73d3579b3c97c6cef0cc77a670865a1fabe06eb

src/plugins/discover/public/application/components/no_results/no_results.tsx
jowg-amazon: Added a github issue to track this UT: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8853

virajsanghvi: Follow up: In addition to tracking unit tests, saw demonstration of code paths offline

virajsanghvi (approved):
For some reason codecov is no longer loading, but think the two main things are covered at this point. 

I restarted some failed CI that I think is unrelated, but please validate.

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
```diff
@@ -59,18 +63,39 @@ export function OpenSavedQueryFlyout({
...
+      const allQueries = await savedQueryService.getAllSavedQueries();
+      const mutableSavedQueries = allQueries.filter((q) => !q.attributes.isTemplate);
+      if (currentTabIdRef.current === 'mutable-saved-queries') {
+        setSavedQueries(mutableSavedQueries);
+      }
+    } else if (currentTabIdRef.current === 'template-saved-queries') {
+      setSavedQueries(templateQueries);
+    }
+    setIsLoading(false);
+  }, [savedQueryService, currentTabIdRef, setSavedQueries, queryStringManager]);
```
ashwin-pc, line 98: Does this mean that now while my sample queries load my saved query flyout is unusable? Also if thats true, whats the user experience when these queries take time to load or fail?

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
```diff
@@ -59,18 +63,39 @@ export function OpenSavedQueryFlyout({
...
+    }
+
+    // Set queries based on the current tab
+    if (currentTabIdRef.current === 'mutable-saved-queries') {
+      const allQueries = await savedQueryService.getAllSavedQueries();
+      const mutableSavedQueries = allQueries.filter((q) => !q.attributes.isTemplate);
+      if (currentTabIdRef.current === 'mutable-saved-queries') {
+        setSavedQueries(mutableSavedQueries);
+      }
+    } else if (currentTabIdRef.current === 'template-saved-queries') {
```
ashwin-pc, line 94: nit: Can we use enums or constants for the tabs and not raw strings?

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
ashwin-pc: This is the line that concerns me a little and want to know more about why we need this. What are we trying to do here? 

There are 2 states at play here. Query manager state and the local searchbar state. `onQueryOpen` updates the search bar state and not the query state because the act of opening a saved query today does not update the query state until the user has decided to fire the query (similar to how you dont want typing the query to update query state until you are ready to fire the query)

Updating the query manager state here for templates changes this behaviour for users when it comes to templates, because now it means that template queries DO change the query state even if the user has not been given the opportunity to confirm that thats the query that they want to run. 

This change will not affect datasources that dont run the query by default, but for indices and index patterns, this means that opening a saved query template runs the query automatically while non templates do not.

src/plugins/data/public/ui/saved_query_management/saved_query_management_component.tsx
```diff
@@ -256,6 +258,7 @@ export function SavedQueryManagementComponent({
                   onClose={() => openSavedQueryFlyout?.close().then()}
                   onQueryOpen={onLoad}
                   handleQueryDelete={handleDelete}
+                  queryStringManager={queryStringManager}
```
ashwin-pc, line 261: Nit: Instead of prop drilling `queryStringManager` throughout, `savedQueryService` should ideally have been responsible for templates too.

src/plugins/discover/public/application/components/no_results/no_results.tsx
ashwin-pc: +1 having the savedQueryService take care of this would be ideal. Sample queries and templates are an extension of that

src/plugins/discover/public/application/components/no_results/no_results.tsx
```diff
@@ -186,6 +187,39 @@ export const DiscoverNoResults = ({ queryString, query, savedQuery, timeFieldNam
...
+          setSampleQueries([...sampleQueriesResponse, ...newSampleQueries]);
+        } else if (sampleQueriesResponse instanceof Promise) {
+          sampleQueriesResponse
+            .then((datasetSampleQueries: any) => {
+              if (Array.isArray(datasetSampleQueries)) {
+                setSampleQueries([...datasetSampleQueries, ...newSampleQueries]);
+              }
+            })
+            .catch((error: any) => {
+              // noop
```
ashwin-pc, line 216: Lets atleast log this?

ashwin-pc (commented):
Did we sanity test this with the new UI with different datasources and languages?

kavilla (changes requested):
i would avoid injecting a new prop to global component. will add more comments to the pr

src/plugins/data/public/ui/search_bar/search_bar.tsx
```diff
@@ -467,6 +474,7 @@ class SearchBarUI extends Component<SearchBarProps, State> {
             useSaveQueryMenu={useSaveQueryMenu}
             isQueryEditorControl={isQueryEditorControl}
             saveQuery={this.onSave}
+            queryStringManager={this.props.queryStringManager}
```
kavilla, line 477: the query string manager being pass in which will impact the state of the components in search bar which is also utilized by other plugins besides the Discover plugin (like dashboards and visualizations) so i would watch out and generally avoid touching this file.

this already was able to access the query string manager like this:

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/ui/search_bar/search_bar.tsx#L123

this should be available in similar locations within this ui

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
```diff
@@ -59,18 +63,39 @@ export function OpenSavedQueryFlyout({
...
-      (q) =>
-        (selectedTabId === 'mutable-saved-queries' && !q.attributes.isTemplate) ||
-        (selectedTabId === 'template-saved-queries' && q.attributes.isTemplate)
-    );
-    setSavedQueries(queriesForSelectedTab);
-    setHasTemplateQueries(templateQueriesPresent);
-  }, [savedQueryService, selectedTabId, setSavedQueries]);
+    setIsLoading(true);
+    const query = queryStringManager.getQuery();
+    let templateQueries: any[] = [];
```
kavilla, line 72: we should just be passing around a `Query` object.

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
```diff
@@ -59,18 +63,39 @@ export function OpenSavedQueryFlyout({
...
+          ?.getType(query.dataset.type)
+          ?.getSampleQueries?.()) || [];
+
+      // Check if any sample query has isTemplate set to true
+      const hasTemplates = templateQueries.some((q) => q?.attributes?.isTemplate);
+      setHasTemplateQueries(hasTemplates);
+    }
+
+    // Set queries based on the current tab
+    if (currentTabIdRef.current === 'mutable-saved-queries') {
```
kavilla, line 88: apologies i believe i missed the original design PR for template queries. 

Key Issues with Current Approach:

- Saved Query Purpose Change
  - Historically, saved queries stored exactly what users typed
  - Used for cross-application consistency
  - one major use case: Create query in Discover, apply same query in Dashboards against many saved searches and visualizations
- Cross-Plugin Impact
  - Potential unexpected behavior in Dashboards when using templated queries
  - Need to verify behavior after dataset type selection
  - Other plugins may not be aware of template functionality
Suggested Alternative Structure
- Make saved queries and saved query templates separate entities
- Create new saved object type: "saved query template"
- Keep existing saved query behavior intact
- Share UI components between both types

This maintains backwards compatibility while cleanly separating template functionality from core saved query behavior.

that said have we verified existing saved query impacts? do we need to add a saved object migration for saved queries? could we verify the impact to other plugins for example the dashboards plugin?

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
AWSHurneyt: @ashwin-pc That's good to know. Given that context, it sounds like the way forward would involve retrieving the dataset from the query state when the saved query doesn't have a dataset (i.e., when the saved query is a template, or when the saved query is not associated with a dataset), and applying it to the saved query before calling `onQueryOpen`. Does that sound appropriate?
E.g.,
```
onClick={() => {
                if (selectedQuery) {
                  if (
                    // Template queries are not associated with data sources. Apply data source from current query
                    selectedQuery.attributes.isTemplate ||
                    // Associating a saved query with a data source is optional. If no data source is present, apply the current source.
                    !selectedQuery.attributes.query.dataset?.dataSource
                  ) {
                    selectedQuery.attributes.query.dataset = queryStringManager?.getQuery().dataset;
                  }
                  onQueryOpen(selectedQuery);
                  onClose();
                }
              }}

src/plugins/data/public/ui/search_bar/search_bar.tsx
AWSHurneyt: Thanks for calling that out. I'll revert line 205 in [create_search_bar.tsx](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8848/files#diff-c9b47a56764f7a94d072874feb9b1f3473dedb7d32480ace6405475c3e953ce6).

ashwin-pc (approved):
Conditionally approving it to unblock testing the saved query template changes. Saved query as a feature is still broken. That needs to be fixed in a separate PR.

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
AWSHurneyt: Since this PR has been merged, I'll adjust this in a follow-up PR.

src/plugins/data/public/ui/search_bar/search_bar.tsx
AWSHurneyt: Addressed in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8864

PR #8851, opened 2024-11-12
Indexed views framework

amsiglan:
### Description
In this PR we add support for
1. Selecting indexed view inside dataset selector. The indexed view is an Opensearch index which can be created for an external source like S3. When user selects an indexed view, we use this index for querying the data
2. Showing a banner at the top of Discover results canvas. This banner can be used to show actions related to the selected dataset. The first use we will add for this banner would be to show a callout to create an indexed view while searching an external source, or when an indexed view is selected, we can show an entry point to select other indexed view

### Issues Resolved
New feature

#### Note: Below screenshots were produced by making changes to the S3 type config for testing and are not part of this PR since this PR is only adding the framework.

- Callout for creating indexed view
image

- Action from callout can open the create flyout (this is all handled by the type config provider and not in this PR's code)
image

- Create indexed view banner:
image

## Testing the changes
Set up workspace with query enhancements enabled
- For the banner in Discover canvas
1. [For testing only]: Update the S3 config type to return the banner props based on the fetch status
2. Discover will render the banner based on provided props

- For showing indexed views in dataset selector
1. Update the S3 config type to return implemented instance of IndexedViewsService
2. Dataset selector will show the indexed views in the second step

## Changelog
- feat: Add framework to show banner at the top in discover results canvas
- feat: Show indexed views in dataset selector

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/common/datasets/types.ts
+++ b/src/plugins/data/common/datasets/types.ts
@@ -247,8 +247,13 @@ export interface Dataset extends BaseDataset {
   timeFieldName?: string;
   /** Optional language to default to from the language selector */
   language?: string;
-  /** Optional name of the indexed view to search data from */
-  indexedView?: string;
+  /** Optional reference to the source dataset. Example usage is for indexed views to store the
+   * reference to the table dataset 
+   */
+  ref?: {
+    id: string;
+    type: string;
+  };
 }
 
 export interface DatasetField {
@@ -257,29 +262,6 @@ export interface DatasetField {
   displayName?: string;
   // TODO:  osdFieldType?
 }
-
-export enum DatasetCanvasBannerComponentType {
-  CALLOUT = 'CALLOUT',
-  CUSTOM = 'custom',
-}
-
-export type DatasetCanvasBannerProps =
-  | {
-      componentType: DatasetCanvasBannerComponentType.CALLOUT;
-      title: string;
-      iconType?: string;
-      content?: React.ReactNode;
-    }
-  | {
-      componentType: DatasetCanvasBannerComponentType.CALLOUT;
-      title?: string;
-      iconType?: string;
-      content: React.ReactNode;
-    }
-  | {
-      componentType: DatasetCanvasBannerComponentType.CUSTOM;
-      content: React.ReactNode;
-    };
 
 export interface DatasetSearchOptions {
   strategy?: string;
--- a/src/plugins/data/common/datasets/types.ts
+++ b/src/plugins/data/common/datasets/types.ts
@@ -247,6 +247,8 @@ export interface Dataset extends BaseDataset {
   timeFieldName?: string;
   /** Optional language to default to from the language selector */
   language?: string;
+  /** Optional name of the indexed view to search data from */
+  indexedView?: string;
 }
 
 export interface DatasetField {
@@ -255,6 +257,29 @@ export interface DatasetField {
   displayName?: string;
   // TODO:  osdFieldType?
 }
+
+export enum DatasetCanvasBannerComponentType {
+  CALLOUT = 'CALLOUT',
+  CUSTOM = 'custom',
+}
+
+export type DatasetCanvasBannerProps =
+  | {
+      componentType: DatasetCanvasBannerComponentType.CALLOUT;
+      title: string;
+      iconType?: string;
+      content?: React.ReactNode;
+    }
+  | {
+      componentType: DatasetCanvasBannerComponentType.CALLOUT;
+      title?: string;
+      iconType?: string;
+      content: React.ReactNode;
+    }
+  | {
+      componentType: DatasetCanvasBannerComponentType.CUSTOM;
+      content: React.ReactNode;
+    };
 
 export interface DatasetSearchOptions {
   strategy?: string;
--- a/src/plugins/data/public/query/query_string/dataset_service/types.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/types.ts
@@ -3,7 +3,13 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import { EuiIconProps } from '@elastic/eui';
-import { Dataset, DatasetField, DatasetSearchOptions, DataStructure } from '../../../../common';
+import {
+  Dataset,
+  DatasetCanvasBannerProps,
+  DatasetField,
+  DatasetSearchOptions,
+  DataStructure,
+} from '../../../../common';
 import { IDataPluginServices } from '../../../types';
 
 /**
@@ -14,6 +20,14 @@ export interface DataStructureFetchOptions {
   search?: string;
   /** Token for paginated results */
   paginationToken?: string;
+}
+
+export interface DatasetIndexedView {
+  name: string;
+}
+
+export interface DatasetIndexedViewsService {
+  getIndexedViews: (dataset: Dataset) => Promise<DatasetIndexedView[]>;
 }
 
 /**
@@ -74,7 +88,7 @@ export interface DatasetTypeConfig {
    * Retrieves the search options to be used for running the query on the data connection associated
    * with this Dataset
    */
-  getSearchOptions?: () => DatasetSearchOptions;
+  getSearchOptions?: (dataset: Dataset) => DatasetSearchOptions;
   /**
    * Combines a list of user selected data structures into a single one to use in discover.
    * @see https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8362.
@@ -84,4 +98,15 @@ export interface DatasetTypeConfig {
    * Returns a list of sample queries for this dataset type
    */
   getSampleQueries?: (dataset: Dataset, language: string) => any;
+  /**
+   * Service used for indexedViews related operations
+   */
+  indexedViewsService?: DatasetIndexedViewsService;
+  /**
+   * Returns props used to render a banner on the Discover canvas/results section
+   */
+  getSearchBannerProps?: (
+    searchStatus: string,
+    dataset: Dataset
+  ) => DatasetCanvasBannerProps | Promise<DatasetCanvasBannerProps> | undefined;
 }
--- a/src/plugins/data/public/query/query_string/dataset_service/types.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/types.ts
@@ -5,10 +5,10 @@
 import { EuiIconProps } from '@elastic/eui';
 import {
   Dataset,
-  DatasetCanvasBannerProps,
   DatasetField,
   DatasetSearchOptions,
   DataStructure,
+  SavedObject,
 } from '../../../../common';
 import { IDataPluginServices } from '../../../types';
 
@@ -28,6 +28,12 @@ export interface DatasetIndexedView {
 
 export interface DatasetIndexedViewsService {
   getIndexedViews: (dataset: Dataset) => Promise<DatasetIndexedView[]>;
+  /**
+   * Returns the data source saved object connected with the data connection object
+   * @param dataConnectionId Id of the data-connection saved object for which we want the data-source
+   * @returns Data source saved object for the Collection/Cluster attached with the data-connection
+   */
+  getConnectedDataSource: (dataConnectionId: string) => Promise<SavedObject>;
 }
 
 /**
@@ -102,11 +108,4 @@ export interface DatasetTypeConfig {
    * Service used for indexedViews related operations
    */
   indexedViewsService?: DatasetIndexedViewsService;
-  /**
-   * Returns props used to render a banner on the Discover canvas/results section
-   */
-  getSearchBannerProps?: (
-    searchStatus: string,
-    dataset: Dataset
-  ) => DatasetCanvasBannerProps | Promise<DatasetCanvasBannerProps> | undefined;
 }
--- a/src/plugins/data/public/query/query_string/query_results_service/index.ts
+++ b/src/plugins/data/public/query/query_string/query_results_service/index.ts
@@ -1,0 +1,7 @@
+/*
+* Copyright OpenSearch Contributors
+* SPDX-License-Identifier: Apache-2.0
+*/
+
+export * from './query_result_service';
+export * from './types';
--- a/src/plugins/data/public/query/query_string/query_results_service/query_result_service.ts
+++ b/src/plugins/data/public/query/query_string/query_results_service/query_result_service.ts
@@ -1,0 +1,26 @@
+/*
+* Copyright OpenSearch Contributors
+* SPDX-License-Identifier: Apache-2.0
+*/
+
+import { QueryResultEnhancements, QueryResultExtensionConfig } from "./types";
+
+
+export class QueryResultService {
+  private queryResultsExtensionMap: Record<string, QueryResultExtensionConfig>;
+
+  constructor() {
+    this.queryResultsExtensionMap = {};
+  }
+
+  public __enhance = (enhancements: QueryResultEnhancements) => {
+    if (enhancements.queryResultExtension) {
+      this.queryResultsExtensionMap[enhancements.queryResultExtension.id] =
+        enhancements.queryResultExtension;
+    }
+  };
+
+  public getQueryResultExtensionMap = () => {
+    return this.queryResultsExtensionMap;
+  }
+}
--- a/src/plugins/data/public/query/query_string/query_results_service/types.ts
+++ b/src/plugins/data/public/query/query_string/query_results_service/types.ts
@@ -1,0 +1,37 @@
+/*
+* Copyright OpenSearch Contributors
+* SPDX-License-Identifier: Apache-2.0
+*/
+
+import { PublicMethodsOf } from "packages/osd-utility-types/target";
+import { QueryResultService } from "./query_result_service";
+import { Query } from "../../../../common";
+
+export interface QueryResultExtensionConfig {
+  /**
+   * The id for the discover results extension.
+   */
+  id: string;
+  /**
+   * Lower order indicates higher position on UI.
+   */
+  order: number;
+  /**
+   * A function that returns the discover results extension banner. The banner is a
+   * component that will be displayed above the results.
+   * @param dependencies - The dependencies required for the extension.
+   * @returns The component to display the banner.
+   */
+  getBanner?: (dependencies: QueryResultExtensionDependencies) => React.ReactElement | null;
+}
+
+export interface QueryResultExtensionDependencies {
+  query: Query;
+  queryStatus: string;
+}
+
+export interface QueryResultEnhancements {
+  queryResultExtension?: QueryResultExtensionConfig;
+}
+
+export type QueryResultServiceContract = PublicMethodsOf<QueryResultService>;
--- a/src/plugins/data/public/ui/dataset_selector/configurator.tsx
+++ b/src/plugins/data/public/ui/dataset_selector/configurator.tsx
@@ -22,6 +22,7 @@ import React, { useEffect, useMemo, useState } from 'react';
 import { BaseDataset, DEFAULT_DATA, Dataset, DatasetField, Query } from '../../../common';
 import { getIndexPatterns, getQueryService } from '../../services';
 import { IDataPluginServices } from '../../types';
+import { DatasetIndexedView } from '../../query/query_string/dataset_service';
 
 export const Configurator = ({
   services,
@@ -40,10 +41,21 @@ export const Configurator = ({
   const queryString = queryService.queryString;
   const languageService = queryService.queryString.getLanguageService();
   const indexPatternsService = getIndexPatterns();
-  const type = queryString.getDatasetService().getType(baseDataset.type);
+  const type = queryString?.getDatasetService().getType(baseDataset.type);
   const languages = type?.supportedLanguages(baseDataset) || [];
 
-  const [dataset, setDataset] = useState<Dataset>(baseDataset);
+  const [selectedLanguage, setSelectedLanguage] = useState<string>(() => {
+    const currentLanguage = queryString?.getQuery().language;
+    if (languages.includes(currentLanguage)) {
+      return currentLanguage;
+    }
+    return languages[0];
+  });
+
+  const [dataset, setDataset] = useState<Dataset>({
+    ...baseDataset,
+    language: selectedLanguage,
+  });
   const [timeFields, setTimeFields] = useState<DatasetField[]>([]);
   const [timeFieldName, setTimeFieldName] = useState<string | undefined>(dataset.timeFieldName);
   const noTimeFilter = i18n.translate(
@@ -52,30 +64,48 @@ export const Configurator = ({
       defaultMessage: "I don't want to use the time filter",
     }
   );
-  const [language, setLanguage] = useState<string>(() => {
-    const currentLanguage = queryString.getQuery().language;
-    if (languages.includes(currentLanguage)) {
-      return currentLanguage;
-    }
-    return languages[0];
-  });
+  const indexedViewsService = type?.indexedViewsService;
+  const [selectedIndexedView, setSelectedIndexedView] = useState<string>();
+  const [indexedViews, setIndexedViews] = useState<DatasetIndexedView[]>([]);
+  const [isLoadingIndexedViews, setIsLoadingIndexedViews] = useState(false);
+
+  useEffect(() => {
+    const getIndexedViews = async () => {
+      if (indexedViewsService) {
+        setIsLoadingIndexedViews(true);
+        const fetchedIndexedViews = await indexedViewsService.getIndexedViews(baseDataset);
+        setIsLoadingIndexedViews(false);
+        setIndexedViews(fetchedIndexedViews || []);
+      }
+    };
+
+    getIndexedViews();
+  }, [indexedViewsService, baseDataset]);
 
   const submitDisabled = useMemo(() => {
     return (
-      timeFieldName === undefined &&
-      !(
-        languageService.getLanguage(language)?.hideDatePicker ||
-        dataset.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN
-      ) &&
-      timeFields &&
-      timeFields.length > 0
+      isLoadingIndexedViews ||
+      (timeFieldName === undefined &&
+        !(
+          languageService?.getLanguage(selectedLanguage)?.hideDatePicker ||
+          dataset.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN
+        ) &&
+        timeFields &&
+        timeFields.length > 0)
     );
-  }, [dataset, language, timeFieldName, timeFields, languageService]);
+  }, [
+    dataset,
+    selectedLanguage,
+    timeFieldName,
+    timeFields,
+    languageService,
+    isLoadingIndexedViews,
+  ]);
 
   useEffect(() => {
     const fetchFields = async () => {
       const datasetFields = await queryString
-        .getDatasetService()
+        ?.getDatasetService()
         .getType(baseDataset.type)
         ?.fetchFields(baseDataset);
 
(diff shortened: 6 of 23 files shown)


virajsanghvi (commented):
Callout that there are multiple ci items failing - please take a look (looks like lint and snapshot tests, and patch coverage). Also, given patch coverage % and the fact that this is a framework for adding additional functionality, would expect unit test coverage to be higher.

src/plugins/data/common/datasets/types.ts
```diff
@@ -247,6 +247,13 @@ export interface Dataset extends BaseDataset {
   timeFieldName?: string;
   /** Optional language to default to from the language selector */
   language?: string;
+  /** Optional reference to the source dataset. Example usage is for indexed views to store the
+   * reference to the table dataset 
+   */
+  ref?: {
```
virajsanghvi, line 253: Shouldn't we name this something less generic?

@kavilla Is this the right place for this? Isn't the source an attribute of the index that was created? Would expect that to be part of DataSource definition I guess, although if this is a two-way door seems fine for now (I think it would be as not having source in the right place wouldn't impact functionality for saved query/recent dataset?).

src/plugins/data/public/query/query_string/dataset_service/types.ts
```diff
@@ -74,7 +94,7 @@ export interface DatasetTypeConfig {
    * Retrieves the search options to be used for running the query on the data connection associated
    * with this Dataset
    */
-  getSearchOptions?: () => DatasetSearchOptions;
+  getSearchOptions?: (dataset: Dataset) => DatasetSearchOptions;
```
ashwin-pc, line 97: Do we need this change? I dont see it being used anywhere

src/plugins/data/public/query/query_string/query_results_service/index.ts
ashwin-pc, line 1: Can you fix the lint errors

src/plugins/vis_builder/public/application/components/data_tab/field_selector.tsx
```diff
@@ -33,7 +33,7 @@ export const FieldSelector = () => {
   const [filteredFields, setFilteredFields] = useState<IndexPatternField[]>([]);
 
   useEffect(() => {
-    const indexFields = indexPattern?.fields.getAll() ?? [];
+    const indexFields = indexPattern?.fields?.getAll() ?? [];
```
ashwin-pc, line 36: Why is this being included here?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -123,6 +158,79 @@ export const Configurator = ({
...
+                label={
+                  <EuiFormLabel>
+                    {i18n.translate(
+                      'data.explorer.datasetSelector.advancedSelector.configurator.showAvailableIndexedViewsLabel',
+                      {
+                        defaultMessage: 'Query indexed view'
+                      }
+                    )}
+                  </EuiFormLabel>  }
+                onChange={(e) => setSelectIndexedView(e.target.checked)}
```
virajsanghvi, line 176: Didn't @riysaxen-amzn add tests for a bunch of these? Looks like they're all here, but not sure why coverage report looks different

src/plugins/data/public/query/query_string/query_results_service/query_result_service.ts
ashwin-pc, line 1: @joshuali925 is this the correct way to expose an extention?

src/plugins/discover/public/application/view_components/canvas/index.tsx
```diff
@@ -41,6 +42,8 @@ export default function DiscoverCanvas({ setHeaderActionMenu, history, optionalR
       data,
     },
   } = useOpenSearchDashboards<DiscoverViewServices>();
+  const queryResultsBannerRef = useRef<HTMLDivElement>(null);
+  const extensionMap = data.query.queryString.getQueryResultService().getQueryResultExtensionMap();
```
virajsanghvi, line 46: Can we add unit tests for this file? Not sure if @riysaxen-amzn was able to

src/plugins/discover/public/application/view_components/canvas/query_result_extensions/query_result_extension.tsx
```diff
@@ -0,0 +1,43 @@
...
+import ReactDOM from 'react-dom';
+import { QueryResultExtensionConfig, QueryResultExtensionDependencies } from '../../../../../../data/public/query/query_string/query_results_service';
+
+export interface QueryResultExtensionProps {
+  config: QueryResultExtensionConfig;
+  dependencies: QueryResultExtensionDependencies;
+  bannerContainer: Element;
+}
+
+const QueryResultExtensionPortal: React.FC<{ container: Element }> = (props) => {
```
virajsanghvi, line 17: Can we add tests for this?

virajsanghvi (commented):
Does this incorporate all feedback addressed in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8818?

Can you point out the major new changes from previous PRs?

ashwin-pc (commented):
Didnt get a chance to review the configurator changes but had a look at the rest. Added my comments below. I agree with Viraj, can we add unit tests for the new code introduced?

src/plugins/data/public/query/query_string/dataset_service/types.ts
amsiglan: Will remove

src/plugins/vis_builder/public/application/components/data_tab/field_selector.tsx
amsiglan: Not sure how it got picked, removing

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -42,8 +47,20 @@ export const Configurator = ({
...
+    if (languages.includes(currentLanguage)) {
+      return currentLanguage;
+    }
+    return languages[0];
+  });
+
+  const [dataset, setDataset] = useState<Dataset>({
+    ...cloneDeep(baseDataset),
+    language,
+  });
```
joshuali925, line 63: avoid clone on re-render?
```suggestion
  const [dataset, setDataset] = useState<Dataset>(() => ({
    ...cloneDeep(baseDataset),
    language,
  }));
```

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -52,25 +69,36 @@ export const Configurator = ({
...
+  const [indexedViews, setIndexedViews] = useState<DatasetIndexedView[]>([]);
+  const [isLoadingIndexedViews, setIsLoadingIndexedViews] = useState(false);
+
+  useEffect(() => {
+    const getIndexedViews = async () => {
+      if (indexedViewsService) {
+        setIsLoadingIndexedViews(true);
+        const fetchedIndexedViews = await indexedViewsService.getIndexedViews(baseDataset);
+        setIsLoadingIndexedViews(false);
+        setIndexedViews(fetchedIndexedViews || []);
```
joshuali925, line 83: nit. avoid setting state if component is unmonted?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -52,25 +69,36 @@ export const Configurator = ({
...
-      !(
-        languageService.getLanguage(language)?.hideDatePicker ||
-        dataset.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN
-      ) &&
-      timeFields &&
-      timeFields.length > 0
+      isLoadingIndexedViews ||
+      (timeFieldName === undefined &&
+        !(
+          languageService?.getLanguage(language)?.hideDatePicker ||
```
joshuali925, line 95: languageService can be undefined?

src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
```diff
@@ -82,7 +82,8 @@ export const DatasetSelector = ({
   const { overlays } = services;
   const datasetService = getQueryService().queryString.getDatasetService();
   const datasetIcon =
-    datasetService.getType(selectedDataset?.type || '')?.meta.icon.type || 'database';
+    datasetService.getType(selectedDataset?.ref?.type || selectedDataset?.type || '')?.meta.icon
```
joshuali925, line 85: q. is indexView the only use case for `ref` right now? do we have plan to add other use cases?

src/plugins/data/common/datasets/types.ts
kavilla: that is a good point. for example we do store the session id wtih the meta info about the data source since it's the data source that decides that where as the data source would also include the index that is pointing to. originally i think we though it should go in the data source meta but didn't have too much of hard buy in on the data source. but reflecting it makes more sense there.

agreeed as well this is a two way door. as i believe we can still be able to query the info i think the ref now is used for ui purposes (for now) but def will need to update it.

src/plugins/data/public/query/query_string/query_results_service/query_result_service.ts
joshuali925: i think it's fine, the query editor extension is exposed through language service because the extensions are language specific. whereas query result extension is not specific to language so it makes sense to not reuse language service

src/plugins/data/public/query/query_string/dataset_service/types.ts
```diff
@@ -16,6 +22,20 @@ export interface DataStructureFetchOptions {
...
+}
+
+export interface DatasetIndexedViewsService {
+  getIndexedViews: (dataset: Dataset) => Promise<DatasetIndexedView[]>;
+  /**
+   * Returns the data source saved object connected with the data connection object
+   * @param dataConnectionId Id of the data-connection saved object for which we want the data-source
+   * @returns Data source saved object for the Collection/Cluster attached with the data-connection
+   */
+  getConnectedDataSource: (dataConnectionId: string) => Promise<SavedObject>;
```
kavilla, line 36: isn't this already present in the dataset.dataSource? This might be putting too much on the dataset service or another thing to maintain. if the devs need the access the data source then can just use the saved objects client directly.

kavilla, line 36: nit:

```suggestion
  getDataSource: (dataset: Dataset) => Promise<SavedObject>;
```

if we were to go with this it might just be easier to just pass in the dataset as you have above instead of having it pass by id

src/plugins/data/public/query/query_string/query_results_service/types.ts
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { PublicMethodsOf } from 'packages/osd-utility-types/target';
+import { QueryResultService } from './query_result_service';
+import { Query } from '../../../../common';
+
+export interface QueryResultExtensionConfig {
+  /**
+   * The id for the discover results extension.
```
kavilla, line 12:
```suggestion
   * The id for the query results extension.
```

since the data plugin can be used by any plugin and is used in dashboards and visualizations

src/plugins/discover/public/application/view_components/canvas/query_result_extensions/query_result_extension.tsx
amsiglan: done

src/plugins/data/public/ui/dataset_selector/configurator.tsx
amsiglan: The state won't get initialized again on re-render even if it is an object

src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
amsiglan: For now it is only for index views

src/plugins/data/public/query/query_string/dataset_service/types.ts
amsiglan: The data source contains info about the data connection itself, this is need to be able to use the data connection's attributes and fetch the matching data source (collection/cluster). It is added to the indexed view service and only the type lives here in the dataset_service

amsiglan: Don't want to name it `getDataSource` because it doesn't get the data source object for the given id but instead it looks for the data source saved object connected with the given data connection object

src/plugins/data/public/ui/dataset_selector/configurator.tsx
amsiglan: Not sure why it got changed

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -52,25 +69,36 @@ export const Configurator = ({
...
+          dataset.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN
+        ) &&
+        timeFields &&
+        timeFields.length > 0)
     );
-  }, [dataset, language, timeFieldName, timeFields, languageService]);
+  }, [dataset, language, timeFieldName, timeFields, languageService, isLoadingIndexedViews]);
```
kavilla, line 101: nit: do we need to prevent the user from clicking submit if the indexed view is still being loaded? they should be able to submit and would be able to query for results with or without indexed views so it seems like preventing the user from submitting would make it seem like an indexed view is required to query.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -123,6 +151,81 @@ export const Configurator = ({
...
+                    options={indexedViews.map(({ name }) => ({
+                      text: name,
+                      value: name,
+                    }))}
+                    value={selectedIndexedView}
+                    onChange={async (e) => {
+                      const value = e.target.value;
+                      setSelectedIndexedView(value);
+                      let connectedDataSource;
+                      if (dataset.dataSource?.id) {
```
kavilla, line 199: isn't dataSource have everything you need? 

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/common/datasets/types.ts#L224

logically if the type is not an INDEX_PATTERN then it seems like we just want to replace the dataset.dataSource.type = DATA_SOURCE

vs making this unique to the saved object if im understanding this correctly

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -202,7 +305,7 @@ export const Configurator = ({
         </EuiButton>
         <EuiButton
           onClick={async () => {
-            await queryString.getDatasetService().cacheDataset(dataset, services);
+            await queryString?.getDatasetService().cacheDataset(dataset, services);
```
kavilla, line 308: queryString manager isn't optional. if this related to a test though we should fix the test to instantiate these properly vs updating the code.

src/plugins/data/common/datasets/types.ts
virajsanghvi: Can we create an issue to track the rename and move?

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -186,6 +186,12 @@ export const QueryEditorUI: React.FC<Props> = (props) => {
 
   const onSelectLanguage = (languageId: string) => {
     const newQuery = queryString.getInitialQueryByLanguage(languageId);
+    if (newQuery.dataset) {
```
kavilla, line 189: i might need a little bit more context here. within the query editor this newQuery will always have a dataset defined. 

the getInitialQueryByDataset seems to be doing the same thing as well.

src/plugins/discover/public/application/view_components/canvas/query_result_extensions/query_resultextension.test.tsx
```diff
@@ -0,0 +1,64 @@
+/*
```
virajsanghvi, line 1: nit: file is not named correctly (missing underscore)

src/plugins/discover/public/application/view_components/canvas/index.tsx
```diff
@@ -50,10 +53,11 @@ export default function DiscoverCanvas({ setHeaderActionMenu, history, optionalR
     };
   });
   const isEnhancementsEnabled = uiSettings.get(QUERY_ENHANCEMENT_ENABLED_SETTING);
+  const defaultColumns = uiSettings.get(DEFAULT_COLUMNS_SETTING) || [];
   const filteredColumns = filterColumns(
     columns,
     indexPattern,
-    uiSettings.get(DEFAULT_COLUMNS_SETTING),
+    defaultColumns,
```
kavilla, line 60: how come?

src/plugins/discover/public/application/view_components/canvas/query_result_extensions/query_resultextension.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+  });
+
+  it('does not render when banner is not present', async () => {
+    getBannerMock.mockReturnValue(null);
+
+    const { queryByText } = render(<QueryResultExtension {...defaultProps} />);
+
+    await waitFor(() => {
+      expect(queryByText('Test Banner')).toBeNull();
+    });
```
virajsanghvi, line 62: nit: are snapshot tests set up? would probably be easier to write, but not a blocker

src/plugins/discover/public/application/view_components/canvas/query_result_extensions/query_resultextension.test.tsx
kavilla, line 1: file name doesn't match others

virajsanghvi (commented):
Do we know why codecov report isn't correct? Hard to validate what we've addressed

src/plugins/discover/public/application/view_components/utils/filter_columns.ts
```diff
@@ -27,7 +27,7 @@ export function filterColumns(
     return columns.length > 0 ? columns : ['_source'];
   }
   // if true, we keep columns that exist in the new index pattern
-  const fieldsName = indexPattern?.fields.getAll().map((fld) => fld.name) || [];
```
kavilla, line 30: i think we should revert if this was just needed for the tests

src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -62,7 +62,7 @@ export class QueryEnhancementsPlugin
         startServices: core.getStartServices(),
         usageCollector: data.search.usageCollector,
       }),
-      getQueryString: (currentQuery: Query) => `source = ${currentQuery.dataset?.title}`,
+      getQueryString: (currentQuery: Query) => `source = ${currentQuery.dataset?.title} | head 10`,
```
kavilla, line 65: we had to remove this because PPL head 10 will cause some null pointer exception based on the time filter

src/plugins/data/public/query/query_string/dataset_service/types.ts
kavilla: could you check out this PR here? 

https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8845

this might reduce the complexity of adding another service

src/plugins/discover/public/application/view_components/canvas/index.tsx
kavilla: +1 it's kind of difficult to get the insight on how this one is being used. 

some unit tests and example implementation of this being used would help a lot.


apologies for the confusion but the original suggestion was to re-utilize the existing framework to achieve the same results. the existing framework for the query editor extensions could have been used to register a banner with functionality to accomplished what is required. the query editor is built with dataset selector. this is how we are able to avoid passing around stateful props.

but current implementation exposes a new service that now has it's own state and has to keep track of the state and communicate the state it's interacting with. which might cause issues anytime the UI state is out of sync or if a new feature is added.

the data plugin is more of a global plugin. Discover isn't the only plugin that uses it. Dashboards and visualizations in OSD core uses it as well. So plugins shouldn't register to a service that only ends up being renderable from the Discover page. So to futureproof it, for example if we wanted to toggle on visualizations or the workbench plugin to utilize your feature we wouldn't have to do new development again. It would just be just a opt in the top nav bar the data plugin search bar and we have access to all your features.

we should be able to take what you have and just register it towards the footer of the query editor. so to the user and to the mock up it looks the same but we are not creating a new service or new component with it's own state to manage.

please see this PR as well which is one way we can simplify same of the indexed view stuff https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8845

src/plugins/data/public/ui/query_editor/query_editor.tsx
ashwin-pc: Can be a fast follow. We should remove langugae fromt he dataset and pass the query object to the dataset config instead

src/plugins/query_enhancements/public/plugin.tsx
amsiglan: okay, will remove it for as discussed with the team but this causes queries to run for a long time and fail as they try to fetch lot of content without the limit

src/plugins/data/public/ui/dataset_selector/configurator.tsx
amsiglan: Removed this change

src/plugins/data/public/ui/dataset_selector/configurator.tsx
amsiglan: Removing the check

src/plugins/data/public/ui/query_editor/_query_editor.scss
```diff
@@ -140,6 +140,10 @@
   padding: 0 $euiSizeXS $euiSizeS;
 }
 
+.osdQueryEditor__bottomPanel {
+  margin: 5px 0;
```
kavilla, line 144: @AMoo-Miki i think this should be an OUI variable correct?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -91,6 +124,38 @@ export const Configurator = ({
...
+    return {
+      ...dataset,
+      id: `${dataset.id}.${selectedIndexedView}`,
+      title: selectedIndexedView,
+      type: DEFAULT_DATA.SET_TYPES.INDEX,
+      sourceDatasetRef: {
+        id: dataset.id,
+        type: dataset.type,
+      },
+      dataSource: connectedDataSource ?? dataset.dataSource,
```
kavilla, line 155: apologies. still not able to understand the difference here. dataset.dataSource should always be the same as connectedDataSource. If this is required because title is not present that I think we should include the title within the `sourceDatasetRef` at the time of creation and all times rather than creating a new function

src/plugins/data/common/datasets/types.ts
```diff
@@ -247,6 +247,13 @@ export interface Dataset extends BaseDataset {
   timeFieldName?: string;
   /** Optional language to default to from the language selector */
   language?: string;
+  /** Optional reference to the source dataset. Example usage is for indexed views to store the
+   * reference to the table dataset
+   */
+  sourceDatasetRef?: {
```
kavilla, line 253: nit:
where you able to comment on @virajsanghvi previous feedback?
def could be follow up but just some insight url length is limited:
- Application and global state are stored in URL parameters
- Although we're far from URL length limits, field name length matters
- Should follow existing naming conventions (e.g. 'timeFieldName' from index patterns)
- Consider shorter property names when possible

src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extensions.tsx
```diff
@@ -62,6 +64,7 @@ const QueryEditorExtensions: React.FC<QueryEditorExtensionsProps> = React.memo((
             dependencies={dependencies}
             componentContainer={extensionComponentContainer}
             bannerContainer={bannerContainer}
+            bottomPanelContainer={bottomPanelContainer}
```
kavilla, line 67: yes! exactly! thank you

virajsanghvi (commented):
Can we resolve @kavilla's concern around the connected data source?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -123,6 +188,57 @@ export const Configurator = ({
...
+                    'data.explorer.datasetSelector.advancedSelector.configurator.indexedViewHelpText',
+                    {
+                      defaultMessage: 'Select an indexed view to speed up your query.',
+                    }
+                  )}
+                >
+                  <EuiSelect
+                    isLoading={isLoadingIndexedViews}
+                    options={indexedViews.map(({ name }) => ({
+                      text: name,
```
kavilla, line 228: is there no option to select an option and then unselect it if the user did not want to have an indexed view? similar to how the timefieldname provides an option. we should follow the same UX for consistency

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -52,13 +69,29 @@ export const Configurator = ({
...
   );
-  const [language, setLanguage] = useState<string>(() => {
-    const currentLanguage = queryString.getQuery().language;
-    if (languages.includes(currentLanguage)) {
-      return currentLanguage;
-    }
-    return languages[0];
-  });
+  const indexedViewsService = type?.indexedViewsService;
+  const [selectedIndexedView, setSelectedIndexedView] = useState<string>();
```
kavilla, line 73: technically the state could be string or undefined right?

src/plugins/data/public/ui/dataset_selector/configurator.tsx
amsiglan: The connected data source is the collection attached to the data-connection ds

src/plugins/data/public/ui/dataset_selector/configurator.tsx
amsiglan: Adding in a followup

src/plugins/data/public/ui/query_editor/_query_editor.scss
```diff
@@ -141,7 +141,7 @@
 }
 
 .osdQueryEditor__bottomPanel {
-  margin: 5px 0;
+  margin: $euiSizeXS 0;
```
virajsanghvi, line 144: Just a call out - this isn't equivalent to 5px

src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extension.tsx
```diff
@@ -103,6 +119,11 @@ export const QueryEditorExtension: React.FC<QueryEditorExtensionProps> = (props)
     props.dependencies,
   ]);
 
+  const bottomPanel = useMemo(() => props.config.getBottomPanel?.(props.dependencies), [
```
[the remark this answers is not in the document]
amsiglan: We have UTs to validate that the given component gets rendered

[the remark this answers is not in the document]
amsiglan: Updated them for the bottom panel

src/plugins/data/public/ui/query_editor/_query_editor.scss
amsiglan: Removed it from here and yes 4px would also be fine in this case

src/plugins/data/public/ui/dataset_selector/configurator.tsx
amsiglan: Fixed

src/plugins/data/public/ui/query_editor/_query_editor.scss
```diff
@@ -140,6 +140,10 @@
   padding: 0 $euiSizeXS $euiSizeS;
 }
 
+.osdQueryEditor__bottomPanel {
+  margin: 5px 0;
+}
+
```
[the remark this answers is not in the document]
amsiglan: Removed it from here, will add it to the callout directly

src/plugins/data/public/ui/query_editor/query_editor.tsx
amsiglan: Fixed

src/plugins/data/common/datasets/types.ts
kavilla: > Do you have a suggested name that accurately reflects what this is?

nope :). i didn't have a good suggestion and was being pedantic and something to keep track of.

src/plugins/data/public/ui/dataset_selector/configurator.tsx
```diff
@@ -91,6 +120,38 @@ export const Configurator = ({
     fetchFields();
   }, [baseDataset, indexPatternsService, queryString, timeFields.length]);
 
+  const updateDatasetForIndexedView = useCallback(async () => {
+    if (!indexedViewsService || !selectedIndexedView) {
+      return dataset;
+    }
+
+    let connectedDataSource;
```
kavilla, line 128: can be a fast follow but i think the indexed view service could handle fetching the datasource and creating and returning an indexed view vs the configurator doing this work

kavilla (dismissed):

src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extension.tsx
```diff
@@ -34,6 +36,14 @@ export interface QueryEditorExtensionDependencies {
    * Set whether the query editor is collapsed.
    */
   setIsCollapsed: (isCollapsed: boolean) => void;
+  /**
+   * Currently set Query
+   */
+  query: Query;
```
kavilla, line 42: nit: what are we using this for? if it's for accessing the current query since it's with the same component it should already have access to the same query. therefore we could just access this query directly

amsiglan: This is used to access the dataset on the query, passed in query to be future safe

virajsanghvi (approved):
Approving assuming ci passes

virajsanghvi (approved):
Restarted failing ci, did pass on linux so assuming it'll pass on windows

ashwin-pc (approved):

ashwin-pc:
Skipping CI group 10 since the failed test case seems to be flakey. The rest of the tests in the suite pass

https://github.com/user-attachments/assets/5f7fcb9d-d841-4109-9b39-47c6e231ffd5

PR #8864, opened 2024-11-14
Adjusted source of QueryStringManager functions for flyout.

AWSHurneyt:
### Description
1. Addressed comment https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8848#discussion_r1840286560
2. Added try/catch block.

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
+++ b/src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
@@ -27,7 +27,7 @@ import { i18n } from '@osd/i18n';
 import { SavedQuery, SavedQueryService } from '../../query';
 import { SavedQueryCard } from './saved_query_card';
 import { Query } from '../../../common';
-import { getQueryService } from 'src/plugins/visualize/public/services';
+import { getQueryService } from '../../services';
 
 export interface OpenSavedQueryFlyoutProps {
   savedQueryService: SavedQueryService;
--- a/src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
+++ b/src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
@@ -68,33 +68,38 @@ export function OpenSavedQueryFlyout({
 
   const fetchAllSavedQueriesForSelectedTab = useCallback(async () => {
     setIsLoading(true);
-    const query = queryStringManager.getQuery();
-    let templateQueries: any[] = [];
+    try {
+      const query = queryStringManager.getQuery();
+      let templateQueries: any[] = [];
 
-    // fetch sample query based on dataset type
-    if (query?.dataset?.type) {
-      templateQueries =
-        (await queryStringManager
-          .getDatasetService()
-          ?.getType(query.dataset.type)
-          ?.getSampleQueries?.()) || [];
+      // fetch sample query based on dataset type
+      if (query?.dataset?.type) {
+        templateQueries =
+          (await queryStringManager
+            .getDatasetService()
+            ?.getType(query.dataset.type)
+            ?.getSampleQueries?.()) || [];
 
-      // Check if any sample query has isTemplate set to true
-      const hasTemplates = templateQueries.some((q) => q?.attributes?.isTemplate);
-      setHasTemplateQueries(hasTemplates);
-    }
-
-    // Set queries based on the current tab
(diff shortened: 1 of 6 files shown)


src/plugins/data/public/ui/search_bar/search_bar.tsx
```diff
@@ -43,7 +43,6 @@ import {
   SavedQueryAttributes,
   TimeHistoryContract,
   QueryStatus,
```
kavilla, line 45: mind fixing the linter issue

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
```diff
@@ -33,7 +33,7 @@ export interface OpenSavedQueryFlyoutProps {
   onClose: () => void;
   onQueryOpen: (query: SavedQuery) => void;
   handleQueryDelete: (query: SavedQuery) => Promise<void>;
-  queryStringManager: QueryStringManager;
+  queryStringManager: QueryStringContract;
```
kavilla, line 36: we actually shouldn't need this at all. since it's apart of the the data public folder the query string manager can be accessed from the fetchAllSavedQueriesForSelectedTab callback like

```typescript
import { getQueryService } from '../../services';

const fetchAllSavedQueriesForSelectedTab = useCallback(async () => {
    setIsLoading(true);
    try {
      const query = queryStringManager.getQuery();
```

AWSHurneyt: Oh thanks for that context! Revised.

src/plugins/data/public/ui/search_bar/search_bar.tsx
AWSHurneyt: Fixed.

kavilla:
thank you so much! one more thing, we use a changelog bot to generate changelog fragment and then release notes. do you mind updating the PR description. in this case i think it can be `skip` as this is fixing something that hasn't been released yet. but i'll leave that up to you. you might need to install the bot as well

https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8864#issuecomment-2475306629

i think regardless it might fail the ci check because it's detecting commits t hat had a changelog fragment

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
```diff
@@ -65,36 +64,43 @@ export function OpenSavedQueryFlyout({
...
-    // Set queries based on the current tab
-    if (currentTabIdRef.current === 'mutable-saved-queries') {
-      const allQueries = await savedQueryService.getAllSavedQueries();
-      const mutableSavedQueries = allQueries.filter((q) => !q.attributes.isTemplate);
+      // Set queries based on the current tab
       if (currentTabIdRef.current === 'mutable-saved-queries') {
-        setSavedQueries(mutableSavedQueries);
+        const allQueries = await savedQueryService.getAllSavedQueries();
+        const mutableSavedQueries = allQueries.filter((q) => !q.attributes.isTemplate);
+        if (currentTabIdRef.current === 'mutable-saved-queries') {
```
kavilla, line 92: is this required? line 89 should be covering it right?

AWSHurneyt: @jowg-amazon could clarify?

jowg-amazon: We were noticing that the getAllSavedQueries call can take a little while to return. If a user switches tabs before this call is finished, we were seeing that `setSavedQueries(mutableSavedQueries)` was overwriting the saved queries incorrectly. This check makes sure that we're still on the mutable saved queries tab.

kavilla: instead of just have a state variable that sets a boolean if it `hasTemplateQueries`, why don't we just have a state variable for templateQueries and a state variable mutableSavedQueries.

Or even more so just set saved queries to all queries and then in the component if the selected tab is template-saved-queries you filter out all the saved queries not matching template-saved-queries etc.

ashwin-pc (approved):

AWSHurneyt:
@kavilla @ashwin-pc could one of you merge the PR? I don't seem to have the permissions to do so.

PR #8866, opened 2024-11-14
[Discover] Hide Date Picker For Unsupported Types

sejli:
### Description

Hides date picker in situations where it is unsupported (when using languages/dataset types that explicitly say that it's not supported)

## Screenshot

https://github.com/user-attachments/assets/b33de83b-cfda-4016-9616-766ffc285df1

## Changelog

- fix: Hide Date Picker for Unsupported Types

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/language_service/language_service.mock.ts
+++ b/src/plugins/data/public/query/query_string/language_service/language_service.mock.ts
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { createEditor, DQLBody, SingleLineInput } from '../../../ui';
 import { LanguageServiceContract } from './language_service';
 import { LanguageConfig } from './types';
 
@@ -14,7 +15,7 @@ const createSetupLanguageServiceMock = (): jest.Mocked<LanguageServiceContract> 
     title: 'DQL',
     search: {} as any,
     getQueryString: jest.fn(),
-    editor: {} as any,
+    editor: createEditor(SingleLineInput, SingleLineInput, [], DQLBody),
     fields: {
       filterable: true,
       visualizable: true,
@@ -28,7 +29,7 @@ const createSetupLanguageServiceMock = (): jest.Mocked<LanguageServiceContract> 
     title: 'Lucene',
     search: {} as any,
     getQueryString: jest.fn(),
-    editor: {} as any,
+    editor: createEditor(SingleLineInput, SingleLineInput, [], DQLBody),
     fields: {
       filterable: true,
       visualizable: true,
@@ -42,7 +43,9 @@ const createSetupLanguageServiceMock = (): jest.Mocked<LanguageServiceContract> 
 
   return {
     __enhance: jest.fn(),
-    registerLanguage: jest.fn(),
+    registerLanguage: jest.fn((language: LanguageConfig) => {
+      languages.set(language.id, language);
+    }),
     getLanguage: jest.fn((id: string) => languages.get(id)),
     getLanguages: jest.fn(() => Array.from(languages.values())),
     getDefaultLanguage: jest.fn(() => languages.get('kuery') || languages.values().next().value),
--- a/src/plugins/data/public/ui/query_editor/query_editor_top_row.tsx
+++ b/src/plugins/data/public/ui/query_editor/query_editor_top_row.tsx
@@ -72,7 +72,7 @@ export default function QueryEditorTopRow(props: QueryEditorTopRowProps) {
   const [isDateRangeInvalid, setIsDateRangeInvalid] = useState(false);
   const [isQueryEditorFocused, setIsQueryEditorFocused] = useState(false);
   const opensearchDashboards = useOpenSearchDashboards<IDataPluginServices>();
-  const { uiSettings, storage, appName } = opensearchDashboards.services;
+  const { uiSettings, storage, appName, data } = opensearchDashboards.services;
 
   const queryLanguage = props.query && props.query.language;
   const persistedLog: PersistedLog | undefined = React.useMemo(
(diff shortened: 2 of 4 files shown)


src/plugins/data/public/ui/query_editor/query_editor_top_row.tsx
```diff
@@ -225,7 +225,17 @@ export default function QueryEditorTopRow(props: QueryEditorTopRowProps) {
   }
 
   function shouldRenderDatePicker(): boolean {
-    return Boolean(props.showDatePicker ?? true) ?? (props.showAutoRefreshOnly && true);
+    return (
+      Boolean(props.showDatePicker ?? true) &&
+      !(
+        queryLanguage &&
+        data.query.queryString.getLanguageService().getLanguage(queryLanguage)?.hideDatePicker
```
kavilla, line 232: should we have updating cypress tests? or write some here

ashwin-pc: +1. A unit test too atleast

src/plugins/data/public/ui/query_editor/query_editor_top_row.tsx
```diff
@@ -225,7 +225,17 @@ export default function QueryEditorTopRow(props: QueryEditorTopRowProps) {
   }
 
   function shouldRenderDatePicker(): boolean {
-    return Boolean(props.showDatePicker ?? true) ?? (props.showAutoRefreshOnly && true);
+    return (
+      Boolean(props.showDatePicker ?? true) &&
```
kavilla, line 229: nit: ouch to the  3 things that can control the date picker from rendering but we call it three different things. 

we should consider field consistency and if the language and dataset type has the ability to configure the date picker is there. we should be asking the query string manager if we should show date picker. since dataset service and  language service are children of the query string manager . instead accessing the querystring than the services.

the querystring manager could have a helper function like

`supportsTimeFilter(query: Query)` then it accessing the services and return back if that query supports time filters. would make it a little bit easier to handle the logic and a lower lvl

kavilla: not a blocker for this pr

src/plugins/data/public/ui/query_editor/query_editor_top_row.tsx
```diff
@@ -225,7 +225,17 @@ export default function QueryEditorTopRow(props: QueryEditorTopRowProps) {
   }
 
   function shouldRenderDatePicker(): boolean {
-    return Boolean(props.showDatePicker ?? true) ?? (props.showAutoRefreshOnly && true);
```
kavilla, line 228: whoops we lost this `showAutoRefreshOnly`

src/plugins/data/public/ui/query_editor/query_editor_top_row.tsx
```diff
@@ -225,7 +225,17 @@ export default function QueryEditorTopRow(props: QueryEditorTopRowProps) {
...
+    return (
+      Boolean(props.showDatePicker ?? true) &&
+      !(
+        queryLanguage &&
+        data.query.queryString.getLanguageService().getLanguage(queryLanguage)?.hideDatePicker
+      ) &&
+      (props.query?.dataset
+        ? data.query.queryString.getDatasetService().getType(props.query.dataset.type)?.meta
+            ?.supportsTimeFilter !== false
```
kavilla, line 236:
```suggestion
            ?.supportsTimeFilter
```

i think?

src/plugins/data/public/query/query_string/language_service/language_service.mock.ts
```diff
@@ -14,7 +15,7 @@ const createSetupLanguageServiceMock = (): jest.Mocked<LanguageServiceContract>
     title: 'DQL',
     search: {} as any,
     getQueryString: jest.fn(),
-    editor: {} as any,
+    editor: createEditor(SingleLineInput, SingleLineInput, [], DQLBody),
```
kavilla, line 18: ty!!

kavilla (commented):
lgtm. just nit pick backlog items.

the only thing is re-adding the showautofrefreshonly conditional

src/plugins/data/public/ui/query_editor/query_editor_top_row.tsx
sejli: We want to let `undefined` values pass. Since, if `supportsTimeFilter` is `undefined`, then `.supportsTimeFilter` would be false. However, since `supportsTimeFilter` is an optional parameter, we want the date picker to still show when it is `undefined`  (it's currently only defined in the `s3_type`). This checks if the `supportsTimeFilter` flag is explicitly set to false, which is the only case where the date picker should be hidden.

src/plugins/data/public/ui/query_editor/query_editor_top_row.tsx
sejli: Discussed offline, was originally fixing a regression since this conditional does not reach `(props.showAutoRefreshOnly && true)` as `Boolean(props.showDatePicker ?? true)` would never be undefined. Fixed the regression by adding it back in as discussed with @kavilla in [1640375](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8866/commits/1640375843f5ff28537f9461969b7cb11d14d3d6).

ananzh (approved):

src/plugins/data/public/ui/query_editor/query_editor_top_row.tsx
```diff
@@ -225,7 +225,17 @@ export default function QueryEditorTopRow(props: QueryEditorTopRowProps) {
   }
 
   function shouldRenderDatePicker(): boolean {
-    return Boolean(props.showDatePicker ?? true) ?? (props.showAutoRefreshOnly && true);
+    return (
+      Boolean((props.showDatePicker || props.showAutoRefreshOnly) ?? true) &&
```
kavilla, line 229: thanks! good catch. i think this wouldn't have been caught until a user complained about it

kavilla (approved):

PR #8871, opened 2024-11-15
[Discover] Dataset search on page load issues

kavilla:
### Description

Required a double click on search and then also potentially loading issue.

### Issues Resolved

n/a

## Testing the changes

Unit tests. Also was able to recreate the bug. Switch the index pattern type config to not search on page load then switch off settings. I was able to run into the bug that required me to click twice to actually get results

## Changelog

- fix: Search on page load out of sync state when clicking submit.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/view_components/utils/use_search.ts
+++ b/src/plugins/discover/public/application/view_components/utils/use_search.ts
@@ -127,11 +127,11 @@ export const useSearch = (services: DiscoverViewServices) => {
     // previously saved search or if it is just transient
     return (
       datasetPreference ||
-      services.uiSettings.get(SEARCH_ON_PAGE_LOAD_SETTING) ||
+      uiSettings.get(SEARCH_ON_PAGE_LOAD_SETTING) ||
       savedSearch?.id !== undefined ||
       timefilter.getRefreshInterval().pause === false
     );
-  }, [data.query, savedSearch?.id, services.uiSettings, timefilter, uiSettings]);
+  }, [data.query, savedSearch, uiSettings, timefilter]);
 
   const startTime = Date.now();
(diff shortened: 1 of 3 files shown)


src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -115,24 +115,23 @@ export const useSearch = (services: DiscoverViewServices) => {
...
+    const typeConfig = dataset ? queryString.getDatasetService().getType(dataset.type) : undefined;
+    const datasetPreference =
+      typeConfig?.meta?.searchOnLoad ?? uiSettings.get(SEARCH_ON_PAGE_LOAD_SETTING);
+
     // A saved search is created on every page load, so we check the ID to see if we're loading a
     // previously saved search or if it is just transient
     return (
-      services.uiSettings.get(SEARCH_ON_PAGE_LOAD_SETTING) ||
+      datasetPreference ||
+      uiSettings.get(SEARCH_ON_PAGE_LOAD_SETTING) ||
```
virajsanghvi, line 130: why repeat this when part of datasetPreference assignment above?

ananzh: +1

kavilla: i believe i was just over thinking. in terms of code it is redundant but i was little bit worried that return true if undefined isn't the most accurate depiction of what should happen.

but with fresh eyes that doesn't really make sense. in terms of the ci taking an hour, could this be a fast follow?

abbyhu2000 (approved):

PR #8872, opened 2024-11-15
Remove filter out serverless cluster and add support to extract index name

ananzh:
### Description
Allow extract index name for both serverless and non-serverless clusters Allow different key formats:
- datasource-id::TIMESERIES::<index-name>:0
- datasource-id::<index-name>:0
- <index-name> (non-serverless case)

### Issues Resolved
NA

## Screenshot
https://github.com/user-attachments/assets/127b851c-23db-4d65-9751-63b307332b3c

## Testing the changes
NA 

## Changelog
- fix: Remove filter out serverless cluster and add support to extract index name

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.test.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.test.ts
@@ -10,11 +10,18 @@ import { SavedObjectsClientContract } from 'opensearch-dashboards/public';
 import { DATA_STRUCTURE_META_TYPES, DataStructure, Dataset } from '../../../../../common';
 import * as services from '../../../../services';
 import { IDataPluginServices } from 'src/plugins/data/public';
+import { of } from 'rxjs';
 
 jest.mock('../../../../services', () => {
+  const mockSearchFunction = jest.fn();
+
   return {
-    getSearchService: jest.fn(),
     getIndexPatterns: jest.fn(),
+    getSearchService: jest.fn(() => ({
+      getDefaultSearchInterceptor: () => ({
+        search: mockSearchFunction,
+      }),
+    })),
     getQueryService: () => ({
       queryString: {
         getLanguageService: () => ({
@@ -90,9 +97,7 @@ describe('indexTypeConfig', () => {
 
   test('should fetch data sources for unknown type', async () => {
     mockSavedObjectsClient.find = jest.fn().mockResolvedValue({
-      savedObjects: [
-        { id: 'ds1', attributes: { title: 'DataSource 1', dataSourceVersion: '3.0' } },
-      ],
+      savedObjects: [{ id: 'ds1', attributes: { title: 'DataSource 1' } }],
     });
 
     const result = await indexTypeConfig.fetch(mockServices as IDataPluginServices, [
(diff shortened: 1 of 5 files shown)


src/plugins/data/public/query/query_string/dataset_service/lib/index_type.test.ts
```diff
@@ -104,30 +109,54 @@ describe('indexTypeConfig', () => {
...
-          id: 'ds2',
-          attributes: { title: 'DataSource 2', dataSourceVersion: '' },
+  describe('fetchIndices', () => {
+    test('should extract index names correctly from different formats', async () => {
+      const mockResponse = {
+        rawResponse: {
+          aggregations: {
+            indices: {
+              buckets: [
+                // Serverless format with TIMESERIES
```
kavilla, line 119: nit: i think serverless is a product. probably shouldn't mention here

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
```diff
@@ -158,9 +152,22 @@ const fetchIndices = async (dataStructure: DataStructure): Promise<string[]> =>
...
+    if (!rawResponse.aggregations) {
+      return [];
+    }
+
+    return rawResponse.aggregations.indices.buckets.map((bucket: { key: string }) => {
+      const key = bucket.key;
+      // Handle the case of serverless cluster where key format is either:
+      // - datasource-id::TIMESERIES::<index-name>:0
+      // - datasource-id::<index-name>:0
+      // Note: Index names cannot contain ':' or '::' in OpenSearch, so these delimiters
```
kavilla, line 164: we have verified this correct?

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
```diff
@@ -158,9 +152,22 @@ const fetchIndices = async (dataStructure: DataStructure): Promise<string[]> =>
...
+      // are guaranteed to be part of the serverless format, not the index name
+      const parts = key.split('::');
+      const lastPart = parts[parts.length - 1] || '';
```
kavilla, line 167: just verifying there's no way we would run into an exception here right?

src/plugins/query_enhancements/public/datasets/s3_type.test.ts
```diff
@@ -154,33 +152,6 @@ describe('s3TypeConfig', () => {
       expect(result.children?.[0].title).toBe('DataSource 1');
       expect(result.hasNext).toBe(true);
     });
-
-    it('should filter out data sources with versions lower than 1.0.0', async () => {
```
kavilla, line 158: nit: perhaps its worth to keep this test to but flip it so it shows that we do NOT filter out data sources

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
```diff
@@ -158,9 +152,22 @@ const fetchIndices = async (dataStructure: DataStructure): Promise<string[]> =>
...
-    return rawResponse.aggregations
-      ? rawResponse.aggregations.indices.buckets.map((bucket: { key: any }) => bucket.key)
-      : [];
+    if (!rawResponse.aggregations) {
+      return [];
+    }
+
+    return rawResponse.aggregations.indices.buckets.map((bucket: { key: string }) => {
+      const key = bucket.key;
+      // Handle the case of serverless cluster where key format is either:
```
kavilla, line 161: nit: same as above i think serverless is a product. so prolly should be taken out

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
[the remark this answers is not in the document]
kavilla:
```suggestion
      const lastPart = parts[parts.length - 1];
```

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
kavilla: can we get sign off or buy in from the team formatting indices this way? it might even be worth export a const like DELIMINATER and using it. i think we have one in some plugins cant remember if it is global.

 ideally in the future collections functions like workspaces accurate results since the datasource id i believe is getting long. and then our dataset takes the id over the parent and appends itself to it. our url is getting quiet long. Discover doesn't really need to know this information if the data source already knows what information. 
 
 also might need some insight on the collection concept.

kavilla (approved):
awesome! i think we should get buy in and committment on the format not changing if this is how it will be otherwise we should consider adding a new data type `collections`

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
ananzh: Yes. I verified both : and ::. Error is 400 invalid index name. Was trying to add screenshot but failed.

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.test.ts
ananzh: will remove this comment

kavilla (dismissed):

abbyhu2000 (approved):

ananzh:
loll new commit will wipe out changelog then update changelog will wipe out approval 
@kavilla could you re-approve? 🫨

kavilla (approved):

PR #8888, opened 2024-11-19
[Workspace] Isolate objects based on workspace when calling get/bulkGet

yubonluo:
### Description

Isolate objects based on workspace when calling get/bulkGet

1. Global operation (option/request workspace id is null): Return object;
2. `option.worksapces.length` > 1: Throw error; 
3. Workspace operation: 
	1. Object is data source:
		1. User is data source admin: Return object;
		2. Global data source: Do not return the object;
		3. Validate data source whether belongs to the workspace
	2. Other object:
		1. Global object: Return object;
		2. Validate object whether belongs to the workspace

## Screenshot
No UI change

## Changelog

- refactor: [Workspace] Isolate objects based on workspace when calling get/bulkGet

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/workspace/server/saved_objects/integration_tests/workspace_id_consumer_wrapper.test.ts
```diff
@@ -312,5 +318,128 @@ describe('workspace_id_consumer integration test', () => {
...
+        .expect(200);
+      expect(getResultWithRequestWorkspace.body.id).toEqual('foo');
+      expect(getResultWithRequestWorkspace.body.workspaces).toEqual([createdFooWorkspace.id]);
+
+      const getResultWithoutRequestWorkspace = await osdTestServer.request
+        .get(root, `/api/saved_objects/${dashboard.type}/bar`)
+        .expect(200);
+      expect(getResultWithoutRequestWorkspace.body.id).toEqual('bar');
+
+      await osdTestServer.request
```
SuZhou-Joe, line 355: Can we add one more case that `get global objects within a workspace` won't throw any error? It is to ensure we won't introduce regression for data sources and config.

yubonluo: Sure, I will add more cases

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
```diff
@@ -126,8 +131,54 @@ export class WorkspaceIdConsumerWrapper {
...
+      ): Promise<SavedObject<T>> => {
+        const objectToGet = await wrapperOptions.client.get<T>(type, id, options);
+
+        if (!excludeTypes.includes(type)) {
+          const { workspaces } = this.formatWorkspaceIdParams(wrapperOptions.request, options);
+
+          // If there is a workspace in options or request, object should belong to the workspace.
+          if (workspaces?.length === 1 && !objectToGet.workspaces?.includes(workspaces[0])) {
+            throw SavedObjectsErrorHelpers.createGenericNotFoundError(type, id);
+          }
```
SuZhou-Joe, line 177: I am thinking we can extract the code to a method like `preflightGet(type: string, id: string)`, it can throw error if the workspace check fails and it can be used both in `get` and `bulkGet`.

yubonluo: Sure, I have extract the common validate code.

kavilla (dismissed):

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
```diff
@@ -48,6 +51,26 @@ export class WorkspaceIdConsumerWrapper {
     return type === UI_SETTINGS_SAVED_OBJECTS_TYPE;
   }
 
+  // If the object.workspaces is null/[] (as global object), return it directly.
+  // If the workspace is specified, validate whether the object exists in the workspace.
+  private validateObjectExistInWorkspaces<T>(
```
kavilla, line 56: nit: the name of this function is a little bit confusing as it will not really just validate which implies returning true or false but it would modify the object if it exists in the workspace. perhaps this could be a return of true return if the object workspaces includes the workspace.

if this is the current standard of saved objects then please ignore.

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
```diff
@@ -48,6 +51,26 @@ export class WorkspaceIdConsumerWrapper {
...
+    object: SavedObject<T>,
+    workspace: string
+  ): SavedObject<T> {
+    if (object.workspaces && object.workspaces.length > 0) {
+      if (!object.workspaces.includes(workspace)) {
```
kavilla, line 61: general question, and im not really sure if this could be possible because I don't think workspaces references any associated saved objects

if the `object.workspaces` does not include the current workspace is there ever a `workspace.objects` or something like that? or only the object keeps reference to the workspace?

yubonluo: Hi @kavilla , only the object keeps reference to the workspace like `object.workspaces` and there is no `workspace.objects`

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
yubonluo: Thanks for your suggestion, I have updated the function name to `formatObjectForWorkspaceValidation`

SuZhou-Joe (approved):

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
```diff
@@ -126,8 +156,60 @@ export class WorkspaceIdConsumerWrapper {
...
         return wrapperOptions.client.find(finalOptions);
       },
-      bulkGet: wrapperOptions.client.bulkGet,
-      get: wrapperOptions.client.get,
+      bulkGet: async <T = unknown>(
+        objects: SavedObjectsBulkGetObject[] = [],
+        options: SavedObjectsBaseOptions = {}
+      ): Promise<SavedObjectsBulkResponse<T>> => {
+        const { workspaces } = this.formatWorkspaceIdParams(wrapperOptions.request, options);
+        if (!!workspaces && workspaces.length > 1) {
```
wanglam, line 164: Not a blocking issue, seems the options.workspaces is useless in `bulkGet` and `get`, can we remove this parameter in the options?

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
```diff
@@ -48,6 +52,32 @@ export class WorkspaceIdConsumerWrapper {
...
+    object: SavedObject<T>,
+    workspace: string,
+    request: OpenSearchDashboardsRequest
+  ) {
+    // Data source is a workspace level object, validate if the request has access to the data source within the requested workspace.
+    if (validateIsWorkspaceDataSourceAndConnectionObjectType(object.type)) {
+      if (!!getWorkspaceState(request).isDataSourceAdmin) {
+        return true;
+      }
+      // Deny access if the object is a global data source (no workspaces assigned)
```
wanglam, line 65: Seems the dashboards admin will be denied too, is it expected?

yubonluo: Yes, the dashboard admin can not access the the global data source in the workspace

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
```diff
@@ -48,6 +52,32 @@ export class WorkspaceIdConsumerWrapper {
...
+      if (!object.workspaces || object.workspaces.length === 0) {
```
SuZhou-Joe, line 66: I am thinking there will be an issue that dashboard admin can list all the data sources but if they switch to a global data source, they will be able to denied. Can we allow dashboard admin to `get` data sources when outside workspace?

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
```diff
@@ -48,6 +52,32 @@ export class WorkspaceIdConsumerWrapper {
...
+      }
+    }
+    /*
+     * Allow access if the requested workspace matches one of the object's assigned workspaces
+     * This ensures that the user can only access data sources within their current workspace
+     */
+    if (object.workspaces && object.workspaces.length > 0) {
+      return object.workspaces.includes(workspace);
+    }
+    // Allow access if the object is a global object (object.workspaces is null/[])
```
SuZhou-Joe, line 77: The comment looks weird to me, if the object is global, they will be denied in line 66.

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
yubonluo: If user is outside the workspace, this method will not be called and bulkGet/Get will return object directly.

SuZhou-Joe: Oh, yes. Thanks for that.

SuZhou-Joe (approved):

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
```diff
@@ -126,8 +156,60 @@ export class WorkspaceIdConsumerWrapper {
...
+          throw SavedObjectsErrorHelpers.createBadRequestError('Multiple workspace parameters');
+        }
+
+        const objectToGet = await wrapperOptions.client.get<T>(type, id, options);
+
+        if (
+          workspaces?.length === 1 &&
+          !this.validateObjectInAWorkspace(objectToGet, workspaces[0], wrapperOptions.request)
+        ) {
+          throw SavedObjectsErrorHelpers.createGenericNotFoundError(type, id);
```
Hailong-am, line 207: As discussed we could use 403 forbidden here to differentiate object not fould case.

Hailong-am (approved):

src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts
yubonluo: Sure I will update the code.

SuZhou-Joe (approved):

Hailong-am (approved):

PR #8896, opened 2024-11-19
Added framework to get default query string using dataset and language combination

amsiglan:
### Description
This PR adds framework to get the default query for a dataset and language combination from the dataset. If not provided we fallback to the defaults from the language service for each language.

## Testing the changes
Tested using local setup

## Changelog
- feat: Added framework to get default query string using dataset and language combination

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/dataset_service/types.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/types.ts
@@ -106,4 +106,8 @@ export interface DatasetTypeConfig {
    * Service used for indexedViews related operations
    */
   indexedViewsService?: DatasetIndexedViewsService;
+  /**
+   * Returns the default query that is added to the query editor when a dataset is selected.
+   */
+  getDefaultQueryString?: (dataset: Dataset, language: string) => string | void;
 }
--- a/src/plugins/data/public/query/query_string/dataset_service/types.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/types.ts
@@ -8,6 +8,7 @@ import {
   DatasetField,
   DatasetSearchOptions,
   DataStructure,
+  Query,
   SavedObject,
 } from '../../../../common';
 import { IDataPluginServices } from '../../../types';
@@ -107,7 +108,7 @@ export interface DatasetTypeConfig {
    */
   indexedViewsService?: DatasetIndexedViewsService;
   /**
-   * Returns the default query that is added to the query editor when a dataset is selected.
+   * Returns the initial query that is added to the query editor when a dataset is selected.
    */
-  getDefaultQueryString?: (dataset: Dataset, language: string) => string | void;
+  getInitialQueryString?: (query: Query) => string | void;
 }
--- a/src/plugins/data/public/query/query_string/query_string_manager.ts
+++ b/src/plugins/data/public/query/query_string/query_string_manager.ts
@@ -63,11 +63,16 @@ export class QueryStringManager {
     return this.storage.get('userQueryString') || '';
   }
 
-  private getInitialDatasetQueryString(query: Query, dataset: Dataset) {
-    const { language } = query;
+  private getInitialDatasetQueryString(query: Query) {
+    const { language, dataset } = query;
 
     const languageConfig = this.languageService.getLanguage(language);
-    const typeConfig = this.datasetService.getType(dataset.type);
+    let typeConfig;
+
+    if (dataset) {
+      typeConfig = this.datasetService.getType(dataset.type);
+    }
+
     return (
       typeConfig?.getInitialQueryString?.(query) ?? (languageConfig?.getQueryString(query) || '')
     );
@@ -93,7 +98,7 @@ export class QueryStringManager {
 
       return {
         ...newQuery,
-        query: this.getInitialDatasetQueryString(newQuery, defaultDataset),
+        query: this.getInitialDatasetQueryString(newQuery),
       };
     }
 
@@ -257,7 +262,7 @@ export class QueryStringManager {
         dataset,
         query: '',
       };
-      newQuery.query = this.getInitialDatasetQueryString(newQuery, dataset);
+      newQuery.query = this.getInitialDatasetQueryString(newQuery);
       return newQuery;
     }
 
@@ -281,16 +286,12 @@ export class QueryStringManager {
    */
   public getInitialQueryByLanguage = (languageId: string) => {
     const curQuery = this.query$.getValue();
-    const language = this.languageService.getLanguage(languageId);
     const newQuery = {
       ...curQuery,
       language: languageId,
     };
-    let queryString = language?.getQueryString(newQuery) || '';
 
-    if (curQuery.dataset) {
-      queryString = this.getInitialDatasetQueryString(newQuery, curQuery.dataset) ?? queryString;
-    }
+    const queryString = this.getInitialDatasetQueryString(newQuery);
     this.languageService.setUserQueryString(queryString);
 
     return {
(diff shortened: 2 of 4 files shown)


kavilla (changes requested):
needs tests

src/plugins/data/public/query/query_string/dataset_service/types.ts
```diff
@@ -106,4 +106,8 @@ export interface DatasetTypeConfig {
    * Service used for indexedViews related operations
    */
   indexedViewsService?: DatasetIndexedViewsService;
+  /**
+   * Returns the default query that is added to the query editor when a dataset is selected.
+   */
+  getDefaultQueryString?: (dataset: Dataset, language: string) => string | void;
```
kavilla, line 112: the language service passes in a `Query` object since it already has dataset and language. 

Could we provide this as well? If functionality it is similar to the language config `getQueryString` wouldn't it be better to call it `getQueryString`. And I don't think we want to return type of `string | void` if people decide to implement and return `void` then this function name is confusing.

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -63,6 +63,11 @@ export class QueryStringManager {
     return this.storage.get('userQueryString') || '';
   }
 
+  private getDefaultDatasetQueryString(dataset: Dataset, language: string) {
```
kavilla, line 66: this appears to get actaully get the initial dataset query string.

`default` within this context is called sparingly compared the get intial query. we should update the name because it will become confusing

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -81,7 +86,13 @@ export class QueryStringManager {
...
       const language = this.languageService.getLanguage(defaultLanguageId);
       const newQuery = { ...query, dataset: defaultDataset };
-      const newQueryString = language?.getQueryString(newQuery) || '';
+
+      const defaultDatasetQueryString = this.getDefaultDatasetQueryString(
+        defaultDataset,
+        newQuery.language
+      );
+      const newQueryString =
+        defaultDatasetQueryString ?? (language?.getQueryString(newQuery) || '');
```
kavilla, line 95:
```suggestion
        this.getDefaultDatasetQueryString(
        defaultDataset,
        newQuery.language
      ) || language?.getQueryString(newQuery) || '';
```

for consistency

kavilla (commented):
this still doesn't make a lot of sense to me with the current **existing** framework. why would we stop here at the query string level? if the dataset type config being able to override the query syntax provided by the language service then why shouldn't the dataset type config being able to override the editor, fields? it seems like we are overriding the sample queries already so pretty much the language service serves no real purpose if we just keep overriding by the dataset service

virajsanghvi (commented):
are there no unit tests for query_string_manager.ts?

src/plugins/data/public/query/query_string/query_string_manager.ts
kavilla: perhaps we need a default query string at the workspace level but yeah for DQL and Lucene it's a valid query and the original behavior.

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -244,13 +255,15 @@ export class QueryStringManager {
...
+      const defaultDatasetQueryString = this.getDefaultDatasetQueryString(dataset, language);
       const languageService = this.languageService.getLanguage(language);
       const newQuery = {
         language,
         dataset,
         query: '',
       };
-      newQuery.query = languageService?.getQueryString(newQuery) || '';
+      newQuery.query =
+        defaultDatasetQueryString ?? (languageService?.getQueryString(newQuery) || '');
```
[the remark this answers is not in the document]
amsiglan: Let me put it all together

src/plugins/data/public/query/query_string/dataset_service/types.ts
amsiglan: Let's go with `getInitialQueryString`

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -63,6 +63,16 @@ export class QueryStringManager {
     return this.storage.get('userQueryString') || '';
   }
 
+  private getInitialDatasetQueryString(query: Query, dataset: Dataset) {
```
amsiglan, line 66: The reason to have the additional `dataset` parameter is to avoid unnecessary checks for `query.dataset` since all the caller methods call this only with valid dataset

src/plugins/data/public/query/query_string/query_string_manager.ts
amsiglan: Updated the name to use `getInitialQueryString`

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -302,7 +319,8 @@ export class QueryStringManager {
       language: languageId,
       dataset: newDataset,
     };
-    const queryString = language?.getQueryString(newQuery) || '';
+    const defaultDatasetQueryString = this.getDefaultDatasetQueryString(newDataset, languageId);
+    const queryString = defaultDatasetQueryString ?? (language?.getQueryString(newQuery) || '');
```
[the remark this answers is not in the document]
amsiglan: Refactored code to dedupe code pieces for language config as well

src/plugins/data/public/query/query_string/query_string_manager.ts
amsiglan: No op

amsiglan:
> needs tests

Working on tests

src/plugins/data/public/query/query_string/query_string_manager.ts
virajsanghvi: It it possible for dataset to not match query.dataset? If this is not checked for by calling code, can we throw when these do not match as an invariant?

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -279,7 +286,11 @@ export class QueryStringManager {
       ...curQuery,
       language: languageId,
     };
-    const queryString = language?.getQueryString(newQuery) || '';
+    let queryString = language?.getQueryString(newQuery) || '';
```
virajsanghvi, line 289: isn't this duplicating what getInitialDatasetQueryString is doing?

amsiglan: I can put this behind else

virajsanghvi: Yeah, `queryString = curQuery.dataset ? this.getInitialDatasetQueryString(newQuery, curQuery.dataset) : (language?.getQueryString(newQuery) || '')` might make this a little less confusing. Or just handling the undefined dataset case in the common method (preferred imo, but not sure I'm missing something)

src/plugins/data/public/query/query_string/query_string_manager.ts
virajsanghvi: Just in case I'm not around, this is the only real blocking comment to approval from me - just worried about a potential mismatch between dataset and query's dataset - if there's something that prevents that situation, or we enforce we don't get into that situation, I'm good.

amsiglan: updated to just use the dataset from the query

kavilla (approved):

virajsanghvi (approved):
approving assuming it passes ci

PR #8899, opened 2024-11-20
Only support copy for query templates

amsiglan:
### Description
For query templates we don't want to run them when they load because they need user input to be valid. Currently Discover does not support a good way to avoid running the query on load.
To avoid this, this PR only lets the user copy the query, which can then be pastes into the Discover editor for modification before running them.

## Screenshot

https://github.com/user-attachments/assets/52508d7a-a9fd-4c8e-b4d0-1c3bc0f159c4

## Testing the changes
Tested changes locally

## Changelog
- fix: Only support copy action for query templates

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
+++ b/src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
@@ -21,15 +21,18 @@ import {
   EuiTablePagination,
   EuiTitle,
   Pager,
+  copyToClipboard,
 } from '@elastic/eui';
 import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { i18n } from '@osd/i18n';
+import { NotificationsStart } from 'opensearch-dashboards/public';
 import { SavedQuery, SavedQueryService } from '../../query';
 import { SavedQueryCard } from './saved_query_card';
 import { getQueryService } from '../../services';
 
 export interface OpenSavedQueryFlyoutProps {
   savedQueryService: SavedQueryService;
+  notifications?: NotificationsStart;
   onClose: () => void;
   onQueryOpen: (query: SavedQuery) => void;
   handleQueryDelete: (query: SavedQuery) => Promise<void>;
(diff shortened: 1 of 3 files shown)


src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
```diff
@@ -301,25 +343,12 @@ export function OpenSavedQueryFlyout({
           </EuiFlexItem>
           <EuiFlexItem grow={false}>
             <EuiButton
-              disabled={!selectedQuery}
+              // disabled={!selectedQuery}
```
virajsanghvi, line 346: not blocking, but can we clean up dead code?

virajsanghvi (approved):
code looks safe to me although I think it'll end up failing tests for the flyout (you may want to run locally to get ahead of it) - approving to unblock and assuming it passes ci 

let's make sure we follow-up with testing if it does pass

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
amsiglan: Fixed

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
```diff
@@ -111,6 +122,7 @@ export function OpenSavedQueryFlyout({
     fetchAllSavedQueriesForSelectedTab();
     setSearchQuery(EuiSearchBar.Query.MATCH_ALL);
     updatePageIndex(0);
+    setSelectedQuery(undefined);
```
virajsanghvi, line 125: curious, why do you have to do this?

virajsanghvi: nevermind, assuming you want to unset when reloading...

virajsanghvi (approved):

src/plugins/data/public/ui/saved_query_flyouts/open_saved_query_flyout.tsx
amsiglan: Yes when the queries are reloaded or tab is changed

ashwin-pc (approved):

huyaboo (approved):

PR #8930, opened 2024-11-26
[Discover] Fix Initialization if No Saved Query

sejli:
### Description

Upon Discover mount (when `getSavedSearchById` is called), we were either setting the query to be the saved search, if it exists, or the default query. In this PR, I updated the logic such that we use the current query instead, which should be the single entry point that also handles the default query state.

## Changelog

- fix: Update saved search initialization logic to use current query instead of default query

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/view_components/utils/use_search.ts
+++ b/src/plugins/discover/public/application/view_components/utils/use_search.ts
@@ -392,8 +392,7 @@ export const useSearch = (services: DiscoverViewServices) => {
       const savedSearchInstance = await getSavedSearchById(savedSearchId);
 
       const query =
-        savedSearchInstance.searchSource.getField('query') ||
-        data.query.queryString.getDefaultQuery();
+        savedSearchInstance.searchSource.getField('query') || data.query.queryString.getQuery();
 
       const isEnhancementsEnabled = await uiSettings.get('query:enhancements:enabled');
       if (isEnhancementsEnabled && query.dataset) {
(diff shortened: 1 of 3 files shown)


src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -392,8 +392,7 @@ export const useSearch = (services: DiscoverViewServices) => {
       const savedSearchInstance = await getSavedSearchById(savedSearchId);
 
       const query =
-        savedSearchInstance.searchSource.getField('query') ||
-        data.query.queryString.getDefaultQuery();
```
paulstn, line 396: Since `.getDefaultQuery()` isn't being used here, is it still used in other locations? I would've assumed I would see the removal of this function as well after reading the description

kavilla: i dont know if we want removal of the function in general. but i understand what you mean it should probably be a private function within the query string manager.

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -432,7 +431,7 @@ export const useSearch = (services: DiscoverViewServices) => {
       }
 
       filterManager.setAppFilters(actualFilters);
-      data.query.queryString.setQuery(savedQuery ? data.query.queryString.getQuery() : query);
+      data.query.queryString.setQuery(query);
```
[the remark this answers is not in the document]
sejli: I removed this here because originally, `query` was either the query from the `savedSearch` or the default query. Since I changed `query` to `data.query.queryString.getQuery()`, it shouldn't have any impact since we will instead only be using `data.query.queryString.getQuery()` if there is no saved search.

LDrago27 (approved):

ananzh (approved):

PR #8931, opened 2024-11-26
[Auto Suggest] DQL Parsing updates and Fixes

paulstn:
### Description

~NOTE: PR is based on https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8275, can only be merged when the aforementioned is~
They are relatively separate PRs, can decouple

Fixes DQL issues such as:
- DQL 'ID' definition, so that ANTLR can parse field names that contain an 'at' + slight adjustments
- Putting double quotes on values to let ANTLR parse the query even if the values contain offending symbols

https://github.com/user-attachments/assets/9b86b5db-2076-4af4-9b49-9a164b09abd0

- Changing field to only add an EQ (`:`) if the field is of type boolean or keyword, and suggest every other operator type
image

Every file under `.generated` can be ignored

## Changelog

- fix: DQL autocomplete better parsing and fixes

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


joshuali925 (approved):

src/plugins/data/public/antlr/dql/code_completion.ts
```diff
@@ -247,14 +247,21 @@ export const getSuggestions = async ({
...
+              insertText: `"${val}" `,
             };
           })
         );
       }
     }
 
-    const dqlOperators = new Set([DQLParser.AND, DQLParser.OR, DQLParser.NOT]);
+    const booleanOperators = new Set([DQLParser.AND, DQLParser.OR, DQLParser.NOT]);
+    const operators = new Set([
```
mengweieric, line 258: nit. may use more descriptive naming here, something like `singleQuoteWrappedOperators` to align with the above `booleanOperators` naming

src/plugins/data/public/antlr/dql/code_completion.ts
```diff
@@ -135,7 +127,15 @@ export const getSuggestions = async ({
 
     // check to see if field rule is a candidate. if so, suggest field names
     if (candidates.rules.has(DQLParser.RULE_field)) {
-      completions.push(...fetchFieldSuggestions(indexPattern, (f: any) => `${f}: `));
+      completions.push(
```
mengweieric, line 130: tests?

mengweieric (commented):
do we know what are the list of offending symbols that this PR addresses?

src/plugins/data/public/antlr/dql/code_completion.ts
paulstn: they're in `code_completion.test.ts`

paulstn:
> do we know what are the list of offending symbols that this PR addresses?

the symbol that resulted in this pr being made was `@`, but I expanded the definition to align with PPL/SQL fields, since they would have the right pattern to match with OSD field names

src/plugins/data/public/antlr/dql/code_completion.ts
paulstn: changed it to relationalOperators

src/plugins/data/public/antlr/dql/code_completion.ts
paulstn: To clarify, these lines are definitely covered in the tests. There's one test that explicitly covers it, under "Test field suggestions" -> "basic field suggestion". I just put console logs on all of these lines and verified that they're each covered through this test and many others in the file.

mengweieric (approved):

PR #8917, opened 2024-11-23
datepicker conditional render

kavilla:
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #8901, opened 2024-11-20
Backport 8896 and 8851 to 2.x

amsiglan:
### Description
Backport #8896 and #8851 to 2.x

## Screenshot
N/A

## Testing the changes
N/A

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #8932, opened 2024-11-26
[Discover] Support custom logic to insert time filter based on dataset type

joshuali925:
### Description

Continue on https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8917

This PR
- allows dataset to override `hideDatePicker` per language for custom logic to insert time filter into query
- disables submit button if time field is not selected in dataset configurator

### Issues Resolved

closes https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8917

## Changelog
- feat: Support custom logic to insert time filter based on dataset type

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/dataset_service/types.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/types.ts
@@ -45,8 +45,8 @@ export interface DatasetTypeConfig {
   title: string;
   languageOverrides?: {
     [language: string]: {
-      /** The overrides transfer the responsibility of handling the input from
-       * the language interceptor to the dataset types strategy. */
+      /** The override transfers the responsibility of handling the input from
+       * the language interceptor to the dataset type search strategy. */
       hideDatePicker?: boolean;
     };
   };
--- a/src/plugins/data/public/query/query_string/dataset_service/types.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/types.ts
@@ -43,6 +43,13 @@ export interface DatasetTypeConfig {
   id: string;
   /** Human-readable title for the dataset type */
   title: string;
+  languageOverrides?: {
+    [language: string]: {
+      /** The overrides transfer the responsibility of handling the input from
+       * the language interceptor to the dataset types strategy. */
+      hideDatePicker?: boolean;
+    };
+  };
   /** Metadata for UI representation */
   meta: {
     /** Icon to represent the dataset type */
@@ -51,7 +58,7 @@ export interface DatasetTypeConfig {
     tooltip?: string;
     /** Optional preference for search on page load else defaulted to true */
     searchOnLoad?: boolean;
-    /** Optional supportsTimeFilter determines if a time filter is needed */
+    /** Optional supportsTimeFilter determines if a time field is supported */
     supportsTimeFilter?: boolean;
     /** Optional isFieldLoadAsync determines if field loads are async */
     isFieldLoadAsync?: boolean;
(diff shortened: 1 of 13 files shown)


src/plugins/data/public/query/query_string/dataset_service/types.ts
```diff
@@ -43,6 +43,13 @@ export interface DatasetTypeConfig {
   id: string;
   /** Human-readable title for the dataset type */
   title: string;
+  languageOverrides?: {
```
ruanyl, line 46: Is there an example of how or when `languageOverrides` will be used?

src/plugins/query_enhancements/public/search/ppl_search_interceptor.ts
```diff
@@ -68,15 +69,33 @@ export class PPLSearchInterceptor extends SearchInterceptor {
...
   }
 
   private buildQuery() {
-    const query: Query = this.queryService.queryString.getQuery();
+    const { queryString } = this.queryService;
+    const query: Query = queryString.getQuery();
     const dataset = query.dataset;
     if (!dataset || !dataset.timeFieldName) return query;
+    const datasetService = queryString.getDatasetService();
+    if (datasetService.getType(dataset.type)?.languageOverrides?.PPL?.hideDatePicker === false)
```
ruanyl, line 96: It seems when `hideDatePicker === false`, `timeRange` is added to `params.body`, otherwise, the time filter is encapsulated in the PPL query on line 102. The time filter will anyway be added, is this by intention?

joshuali925: yes it's by intention, if `hideDatePicker === false`, then it's not automatically inserted. This allows custom logic to insert the time filter

ruanyl: I see, so when `timeRange` is added to `params.body`, then it's up to server side search strategy to decide how to use it, did I get it right?

src/plugins/query_enhancements/public/search/sql_search_interceptor.ts
```diff
@@ -62,6 +63,16 @@ export class SQLSearchInterceptor extends SearchInterceptor {
         .getDatasetService()
         .getType(datasetType);
       strategy = datasetTypeConfig?.getSearchOptions?.().strategy ?? strategy;
+
+      if (datasetTypeConfig?.languageOverrides?.SQL?.hideDatePicker === false) {
```
ruanyl, line 67: Is this just for consistency? Because I don't see it add time range in the query automatically in sql search interceptor.

ruanyl (approved):
Approved with some non-blocking questions :)

src/plugins/query_enhancements/public/search/ppl_search_interceptor.ts
joshuali925: right

src/plugins/query_enhancements/public/search/sql_search_interceptor.ts
joshuali925: the idea is that if there is a language override that sets `hideDatePicker: false`, then the search interceptor will not automatically insert the timeRange into query, but will pass `timeRange` to search strategy. Passing `timeRange` happens regardless of whether the search interceptor has logic to insert `timeRange`

Currently we don't have SQL query parser + builder so it's hard to insert time range to SQL, but if the search API for a specific dataset takes a SQL query and a separate time range parameter, then this will be useful

ananzh (approved):

joshuali925:
backport 2.x depends on https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8901

PR #8981, opened 2024-11-27
add a new flag in timefilter service to disable time filter

ruanyl:
### Description

Update query assistant UX:

1. when query assistant is active(assistant input expanded), disable the time filter and do NOT add the time filter to the PPL query
2. when query assistant is collapse, enable the time filter and add the time filter to PPL

## Screenshot

https://github.com/user-attachments/assets/cbca812b-2f64-4d74-9c79-5e132a1ede2e

## Changelog

- feat: hide time filter when query assistant input is expanded

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


yubonluo (approved):

ruanyl:
> how come? apologies little bit defensive about the super date picker render because it's a source of a number of bugs due to the different variations of state it will be in. does this also mean if i expand query assist in discover and navigate to dashboards is the timefilter disabled still?

Hey @kavilla, no worry, I totally understand your concern.
> does this also mean if i expand query assist in discover and navigate to dashboards is the timefilter disabled still?

In this particular case(assistant toggle), the state is reset whenever the query assistant bar unmount. So the state won't persist when navigated to dashboards.

And back to the question why introducing a new API in timefilter service?
I've been thinking of keeping this state as a local state of query assistant, but the value needs to be read in several places:
1. The time filter UI component in `QueryEditorTopRow`
2. In the `PPLSearchInterceptor` to decide whether to add time filter to query or not

It ends up with spreading a local state to multiple component which is even a worse design in my opinion. So I decided to have it in a central place where all the components have access to. I may not be the discover profession, really appreciate if you have any suggestions :)

ruanyl:
Thanks @joshuali925 

> we want users to use query assist to add filters to everything including time field

This is actually not the intention, the major reason to introduce this change is the time filter in PPL query will conflict with the time filter on the page which confuses the users.

For example, if user asked "show all error logs in the past week" while having a different time range selected on the UI. Actually, this is not a query assist specific issue, but also a potential issue of regular ppl query, because user can write directly the time filter in the PPL query.

> Selecting or changing a time range on a calendar + quick select component is easier and faster than typing it out in natural language
Users might not change the time filter frequently in one debugging session. With the time filter they only need to set the time filter component once, but now they need to make sure to keep the time filter description in every question
LLM can make mistakes, and the query result would be useless when it uses the wrong time filter

I agree on this, as the PR is a fast fix that we agreed on with @ashwin-pc and @kgcreative 

> when users use query assist, they won't notice that there is already a time filter, and will be surprised if there are no results due to the time filter

> I have seen this before, but I'm not sure if disabling the time filter provides the best user experience. Discover already has a message to call out that no results might be caused by the time range selection. Maybe we can add some notice in query assist as well?

Thank you for sharing this. If people use ppl, I was thinking, can we have the assumption that the user is expected to be aware that PPL can contain time filter as it's valid ppl syntax?

> Although for now I don't have a strong opinion if the UX experts say it should work this way. I only have one concern: typically a disabled component just means I cannot change it and doesn't mean it's not in effect. Could the time filter be hidden instead of disabled?

Good question, I saw the current behaviour of hiding the time filter is when the selected data set doesn't support time filter, for example, it doesn't a time field. I was trying to differentiate it from such case, because in such case, the selected data set actually supports time filter, but just temporary unavailable intentionally.

joshuali925:
> I agree on this, as the PR is a fast fix that we agreed on with @ashwin-pc and @kgcreative

Got it, will there be a plan for long term fix or will this be it?

> can we have the assumption that the user is expected to be aware that PPL can contain time filter as it's valid ppl syntax?

User should know that PPL can contain the time filter, it's designed to be easy to read. I agree that `show all error logs in the past week` is a problem, just not sure if it's worth it to disable the time filter entirely to work around it.

> I was trying to differentiate it from such case, because in such case, the selected data set actually supports time filter, but just temporary unavailable intentionally.

Discover actually separated the concept of supports time field vs supports time filter. For example, an index can support a time field but not show time filter if user selected SQL. The time field is used to show the `Time` column in discover table. I raised it up because my impression from the video was that the time picker gets locked to "Last 15 hours", and we cannot query beyond that when using query assist due to some limitation.

ashwin-pc:
@ruanyl can we close this now?

PR #8986, opened 2024-11-27
Migrate query enhancement cypress tests to OSD repo

abbyhu2000:
### Description

Migrate query enhancement cypress tests to OSD repo.

## Changelog
- feat: Migrate query enhancement cypress tests to OSD repo

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/cypress/fixtures/dashboard/opensearch_dashboards/data.json.txt
+++ b/cypress/fixtures/dashboard/opensearch_dashboards/data.json.txt
@@ -1,0 +1,23 @@
+{
+  "type": "doc",
+  "value": {
+    "id": "3",
+    "index": "timestamp-nanos",
+    "source": {
+      "timestamp": "2019-01-01T12:10:30.123456789Z"
+    },
+    "type": "_doc"
+  }
+}
+
+{
+  "type": "doc",
+  "value": {
+    "id": "4",
+    "index": "timestamp-nanos",
+    "source": {
+      "timestamp": "2019-01-01T12:10:30.123498765Z"
+    },
+    "type": "_doc"
+  }
+}
--- a/cypress/fixtures/dashboard/opensearch_dashboards/data_with_index_pattern.json.txt
+++ b/cypress/fixtures/dashboard/opensearch_dashboards/data_with_index_pattern.json.txt
@@ -1,0 +1,65 @@
+{
+  "type": "doc",
+  "value": {
+    "id": "index-pattern:timestamp-*",
+    "index": ".kibana",
+    "source": {
+      "index-pattern": {
+        "fields": "[{\"name\":\"_id\",\"type\":\"string\",\"esTypes\":[\"_id\"],\"count\":1,\"scripted\":false,\"searchable\":true,\"aggregatable\":true,\"readFromDocValues\":false},{\"name\":\"_index\",\"type\":\"string\",\"esTypes\":[\"_index\"],\"count\":0,\"scripted\":false,\"searchable\":true,\"aggregatable\":true,\"readFromDocValues\":false},{\"name\":\"_score\",\"type\":\"number\",\"count\":0,\"scripted\":false,\"searchable\":false,\"aggregatable\":false,\"readFromDocValues\":false},{\"name\":\"_source\",\"type\":\"_source\",\"esTypes\":[\"_source\"],\"count\":0,\"scripted\":false,\"searchable\":false,\"aggregatable\":false,\"readFromDocValues\":false},{\"name\":\"_type\",\"type\":\"string\",\"esTypes\":[\"_type\"],\"count\":0,\"scripted\":false,\"searchable\":true,\"aggregatable\":true,\"readFromDocValues\":false},{\"name\":\"timestamp\",\"type\":\"date\",\"esTypes\":[\"date\",\"date_nanos\"],\"count\":2,\"scripted\":false,\"searchable\":true,\"aggregatable\":true,\"readFromDocValues\":true}]",
+        "timeFieldName": "timestamp",
+        "title": "timestamp-*",
+        "fieldFormatMap": "{\"timestamp\":{\"id\":\"date_nanos\"}}"
+      },
+      "type": "index-pattern"
+    },
+    "type": "_doc"
+  }
+}
+
+{
+  "type": "doc",
+  "value": {
+    "id": "2",
+    "index": "timestamp-millis",
+    "source": {
+      "timestamp": "2019-01-01T12:10:30.124Z"
+    },
+    "type": "_doc"
+  }
+}
+
+{
+  "type": "doc",
+  "value": {
+    "id": "1",
+    "index": "timestamp-millis",
+    "source": {
+      "timestamp": "2019-01-01T12:10:30.123Z"
+    },
+    "type": "_doc"
+  }
+}
+
+{
+  "type": "doc",
+  "value": {
+    "id": "3",
+    "index": "timestamp-nanos",
+    "source": {
+      "timestamp": "2019-01-01T12:10:30.123456789Z"
+    },
+    "type": "_doc"
+  }
+}
+
+{
+  "type": "doc",
+  "value": {
+    "id": "4",
+    "index": "timestamp-nanos",
+    "source": {
+      "timestamp": "2019-01-02T12:10:30.123498765Z"
+    },
+    "type": "_doc"
+  }
+}
--- a/cypress/integration/core_opensearch_dashboards/apps/query_enhancement/dataset_selector_spec.js
+++ b/cypress/integration/core_opensearch_dashboards/apps/query_enhancement/dataset_selector_spec.js
@@ -1,0 +1,180 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import {
+  MiscUtils,
+  TestFixtureHandler,
+} from '@opensearch-dashboards-test/opensearch-dashboards-test-library';
+import { CURRENT_TENANT } from '../../../../../utils/commands';
+import { clusterName, clusterConnection } from '../../../../../utils/constants';
+
+const miscUtils = new MiscUtils(cy);
+const testFixtureHandler = new TestFixtureHandler(cy, Cypress.env('openSearchUrl'));
+
+const indexSet = ['logstash-2015.09.22', 'logstash-2015.09.21', 'logstash-2015.09.20'];
+
+describe('dataset navigator', { scrollBehavior: false }, () => {
+  before(() => {
+    CURRENT_TENANT.newTenant = 'global';
+    cy.fleshTenantSettings();
+  });
+
+  describe('empty state', () => {
+    it('no index pattern', function () {
+      // Go to the Discover page
+      miscUtils.visitPage(
+        `app/data-explorer/discover#/?_g=(filters:!(),time:(from:'2015-09-19T13:31:44.000Z',to:'2015-09-24T01:31:44.000Z'))`
+      );
+
+      cy.waitForLoaderNewHeader();
+      cy.getElementByTestId('discoverNoIndexPatterns');
+    });
+  });
+
+  describe('select indices', () => {
+    before(() => {
+      testFixtureHandler.importJSONMapping(
+        'cypress/fixtures/dashboard/opensearch_dashboards/query_enhancement/mappings.json.txt'
+      );
+
+      testFixtureHandler.importJSONDoc(
+        'cypress/fixtures/dashboard/opensearch_dashboards/query_enhancement/data.json.txt'
+      );
+
+      // Since default cluster is removed, need to create a data source connection
+      miscUtils.visitPage('app/management/opensearch-dashboards/dataSources/create');
+      cy.intercept('POST', '/api/saved_objects/data-source').as('createDataSourceRequest');
+      cy.getElementByTestId(`datasource_card_opensearch`).click();
+      cy.get('[name="dataSourceTitle"]').type(clusterName);
+      cy.get('[name="endpoint"]').type(clusterConnection);
+      cy.getElementByTestId('createDataSourceFormAuthTypeSelect').click();
+      cy.get(`button[id="no_auth"]`).click();
+
+      cy.getElementByTestId('createDataSourceButton').click();
+      cy.wait('@createDataSourceRequest').then((interception) => {
+        expect(interception.response.statusCode).to.equal(200);
+      });
+      cy.location('pathname', { timeout: 6000 }).should(
+        'include',
+        'app/management/opensearch-dashboards/dataSources'
+      );
+
+      // Go to the Discover page
+      miscUtils.visitPage(`app/data-explorer/discover#/`);
+
+      cy.waitForLoaderNewHeader();
+    });
+
+    it('with SQL as default language', function () {
+      cy.getElementByTestId(`datasetSelectorButton`).click();
+      cy.getElementByTestId(`datasetSelectorAdvancedButton`).click();
+      cy.get(`[title="Indexes"]`).click();
+      cy.get(`[title=${clusterName}]`).click();
+      cy.get(`[title="timestamp-nanos"]`).click();
+      cy.getElementByTestId('datasetSelectorNext').click();
+
+      cy.get(`[class="euiModalHeader__title"]`).should('contain', 'Step 2: Configure data');
+      // should have two options: SQL and PPL
+      cy.getElementByTestId('advancedSelectorLanguageSelect')
+        .get('option')
+        .should('have.length', 2);
+
+      //select SQL
+      cy.getElementByTestId('advancedSelectorLanguageSelect').select('OpenSearch SQL');
+      cy.getElementByTestId('advancedSelectorConfirmButton').click();
+
+      cy.waitForLoaderNewHeader();
+
+      // SQL should already be selected
+      cy.getElementByTestId('queryEditorLanguageSelector').should('contain', 'OpenSearch SQL');
+      cy.waitForLoaderNewHeader();
+
+      // SQL query should be executed and sending back result
+      cy.get(`[data-test-subj="queryResultCompleteMsg"]`).should('be.visible');
+
+      // Switch language to PPL
+      cy.setQueryLanguage('PPL');
+      cy.waitForLoaderNewHeader();
+      cy.get(`[data-test-subj="queryResultCompleteMsg"]`).should('be.visible');
+    });
+
+    it('with PPL as default language', function () {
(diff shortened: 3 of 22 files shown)


cypress/utils/commands.js
```diff
@@ -2,6 +2,21 @@
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
  */
+/*
```
ananzh, line 5: duplicate license

cypress/utils/dashboards/query_enhancement/commands.js
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+Cypress.Commands.add('waitForLoaderNewHeader', () => {
+  const opts = { log: false };
```
ananzh, line 7: I think for utilities, we need to be careful. Eventually we will move the legacy discover ftr tests back to OSD as well. So the utilities should work for both legacy and is_enhancement is true. For example if could could have both of them working, that would save us a lot of future migration efforts
```
Cypress.Commands.add('waitForLoader', (isEnhancement = false) => {
  const opts = { log: false };

  Cypress.log({
    name: 'waitForPageLoad',
    displayName: 'wait',
    message: 'page load',
  });

  cy.wait(Cypress.env('WAIT_FOR_LOADER_BUFFER_MS'));

  // Use recentItemsSectionButton for query enhancement, otherwise use homeIcon
  cy.getElementByTestId(
    isEnhancement ? 'recentItemsSectionButton' : 'homeIcon',
    opts
  );
});
```

kavilla:
just to verify what is the goal for this pr? is it just to get it to this repo right? if so then i'm going to leave comments with the implication that we should fast follow these things but NOT block on anything

cypress/fixtures/dashboard/opensearch_dashboards/data.json.txt
```diff
@@ -0,0 +1,23 @@
+{
+  "type": "doc",
+  "value": {
+    "id": "3",
+    "index": "timestamp-nanos",
+    "source": {
+      "timestamp": "2019-01-01T12:10:30.123456789Z"
```
kavilla, line 7: we should make the file name of this not `data.json.txt` since that is too generic of a name and could cause confusion. 

we should also have more realistic data here as in some other fields maybe even number of other timefields. the size should be more than two. this way the test results for casting a wide net of a date range will yield 2 results the same way filtering by the jan 1st of 2019 would which could maybe lead to false positives in our filtering tests

cypress/fixtures/dashboard/opensearch_dashboards/data_with_index_pattern.json.txt
```diff
@@ -0,0 +1,65 @@
+{
+  "type": "doc",
+  "value": {
+    "id": "index-pattern:timestamp-*",
```
kavilla, line 4: is this the behavior we wanna continue? i think we did this from the legacy but if we we want to add more documents to these tests to capture the cases we are running into when regressions occur

kavilla: if not then we should just consider having stored the ndjson of saved objects and then we just upload them on test start. 

this way if we have massive test data we dont know need to open or modify that file to modify the saved objects

package.json
```diff
@@ -84,8 +84,9 @@
     "spec_to_console": "scripts/use_node scripts/spec_to_console",
     "pkg-version": "scripts/use_node -e \"console.log(require('./package.json').version)\"",
     "release_note:generate": "scripts/use_node scripts/generate_release_note",
-    "cypress:run-without-security": "env TZ=America/Los_Angeles NO_COLOR=1 cypress run --headless --env SECURITY_ENABLED=false",
-    "cypress:run-with-security": "env TZ=America/Los_Angeles NO_COLOR=1 cypress run --headless --env SECURITY_ENABLED=true,openSearchUrl=https://localhost:9200,WAIT_FOR_LOADER_BUFFER_MS=500",
+    "cypress:run-without-security": "env TZ=America/Los_Angeles NO_COLOR=1 cypress run --env SECURITY_ENABLED=false",
```
kavilla, line 87: follow up:

`run-without-security` is long to type out

we should upate the appropriate places and here so like

```
cypress:run // implies it is without security
cypress:run:security // explicit with security
```

kavilla: also might need to mention in some readme to run this function with `--headless` if people want to get the runner open

package.json
```diff
@@ -84,8 +84,9 @@
...
+    "cypress:run-with-security": "env TZ=America/Los_Angeles NO_COLOR=1 cypress run --env SECURITY_ENABLED=true,openSearchUrl=https://localhost:9200,WAIT_FOR_LOADER_BUFFER_MS=500",
+    "osd:ciGroup10": "echo \"apps/query_enhancement/*.js\"",
```
kavilla, line 89: sync with @d-rowe . and def not a blocker for any of this but we should consider while in theprocess of migrating defining a middle ground solution at least for the tests to tag themselves instead of a package.json file. because this was only suppose to be temporary and people wont know they need to update thisfor the test to run

something like the selenium stuff https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/test/functional/apps/dashboard/index.js#L83 but even a hackier one that is not using this hack would be better imo



like i said not blocking just mentioning incase it would make the work smaller increments

cypress/utils/index.d.ts
```diff
@@ -0,0 +1,153 @@
...
+ */
+
+declare namespace Cypress {
+  interface Chainable<Subject> {
+    /**
+     * Call a function when an element with a test id cannot be found
+     * @example
+     * cy.whenTestIdNotFound(['query', 'puery'], () => {...})
+     */
+    whenTestIdNotFound<S = any>(
```
kavilla, line 13: @d-rowe @ananzh so it seems we are already migrating from that package in here too right?

kavilla: also do people call this?

cypress/utils/index.d.ts
```diff
@@ -0,0 +1,153 @@
...
+      callbackFn: void,
+      options?: Partial<Loggable & Timeoutable & Withinable & Shadow>
+    ): Chainable<S>;
+    /**
+     * Get elements by their test ids
+     * @example
+     * cy.getElementsByTestIds(['query', 'puery'])
+     */
+    getElementsByTestIds<S = any>(
+      testIds: string | string[],
```
kavilla, line 24: ik carry over but too flexiable of a system is hard to keep up. i think if we have a getelementbytestid then it should just be string[] here

cypress/utils/index.d.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    /**
+     * Bulk upload NDJSON fixture data
+     * @example
+     * cy.bulkUploadDocs('plugins/test/test_data.txt')
+     */
+    bulkUploadDocs<S = any>(
+      fixturePath: string,
+      index: string
+      // options?: Partial<Loggable & Timeoutable & Withinable & Shadow>
```
kavilla, line 60: shoudl this overwrite?

cypress/utils/dashboards/query_enhancement/constants.js
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const clusterName = 'test_cluster';
+export const clusterConnection = 'http://localhost:9200';
```
kavilla, line 7: lets not port this over imo. this should pull from a config and if that config is suppose to point to an external url then only the cypress config would be able to update so easily this constant is a constant and wont be reusable in cases

cypress/utils/dashboards/query_enhancement/commands.js
```diff
@@ -0,0 +1,45 @@
...
+
+Cypress.Commands.add('setQueryLanguage', (value) => {
+  Cypress.log({
+    name: 'setQueryLanguage',
+    displayName: 'set language',
+    message: value,
+  });
+
+  cy.getElementByTestId(`queryEditorLanguageSelector`).click();
+  cy.get(`[class~="languageSelector__menuItem"]`).contains(value).click({
```
kavilla, line 42: we could either add a test id which might be easier or access this class from the parent that has a test id of queryeditor langauge selector to ensure it doesn't grab the wrong language selector

cypress/utils/helpers.js
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const apiRequest = (url, method = 'POST', body = undefined, qs = undefined) =>
```
kavilla, line 6: we should make this an object after url and method so it wont get so difficutl to update if we need to add more

cypress/utils/helpers.js
```diff
@@ -0,0 +1,31 @@
...
+      'osd-xsrf': true,
+    },
+    body: body,
+    qs: qs,
+  });
+
+export const devToolsRequest = (url, method = 'POST', body = undefined, qs = undefined) =>
+  cy.request({
+    method: 'POST',
+    form: false,
```
kavilla, line 22: nit: seems like there could be a request headers constant that we can spread here

cypress/utils/helpers.js
```diff
@@ -0,0 +1,31 @@
...
+    url: url,
+    headers: {
+      'content-type': 'application/json',
+      'osd-xsrf': true,
+    },
+    body: body,
+    qs: qs,
+  });
+
+export const devToolsRequest = (url, method = 'POST', body = undefined, qs = undefined) =>
```
kavilla, line 19: nit: this is kind of the wrong name. i think it's technically proxying to the console plugin dev tools is an application that is a child of the console plugin but it is client side.

so we aren't actually firing the request to the ui we are hitting the node server api

cypress/utils/dashboards/constants.js
```diff
@@ -0,0 +1,17 @@
...
+// STACK MANAGEMENT PATH
+export const STACK_MANAGEMENT_PATH = BASE_PATH + '/app/management';
+export const SECURITY_PLUGIN_PATH = BASE_PATH + '/app/security-dashboards-plugin#/';
+
+export const TENANTS_MANAGE_PATH = BASE_PATH + '/app/security-dashboards-plugin#/tenants';
+
+export const INDEX_PATTERN_PATH = STACK_MANAGEMENT_PATH + '/opensearch-dashboards/indexPatterns';
+export const SAVED_OBJECTS_PATH = STACK_MANAGEMENT_PATH + '/opensearch-dashboards/objects';
+
+export * from './query_enhancement/constants';
```
kavilla, line 17: we shouldn't export another folders constants here

cypress/utils/dashboards/commands.js
```diff
@@ -0,0 +1,123 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import './query_enhancement/commands';
```
kavilla, line 6: we should have global commands if needed. and then query enhancmeents and dashboards can take from it vs another plugin command

cypress/utils/commands.js
```diff
@@ -2,6 +2,21 @@
...
+ */
+
+import { BASE_PATH } from './constants';
+
+// This function does not delete all indices
+Cypress.Commands.add('deleteAllIndices', () => {
+  cy.log('Deleting all indices');
+  cy.request(
+    'DELETE',
+    `${Cypress.env('openSearchUrl')}/index*,sample*,opensearch_dashboards*,test*,cypress*`
```
kavilla, line 17: do we ahve to refresh?

cypress/utils/commands.js
```diff
@@ -2,6 +2,21 @@
...
  * SPDX-License-Identifier: Apache-2.0
  */
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { BASE_PATH } from './constants';
+
+// This function does not delete all indices
```
kavilla, line 12: can we modify this now to actually delete all indices and not a specific string?

cypress/utils/commands.js
kavilla, line 17: now that we have it can we just say `ENGINE_URL` anything besides the casing

cypress/utils/commands.js
```diff
@@ -13,3 +28,244 @@ Cypress.Commands.add('getElementsByTestIds', (testIds, options = {}) => {
...
+});
+
+Cypress.Commands.add('changeDefaultTenant', (attributes, header = {}) => {
+  const url = Cypress.env('openSearchUrl') + '/_plugins/_security/api/tenancy/config';
+
+  cy.request({
+    method: 'PUT',
+    url,
+    headers: {
+      'content-type': 'application/json;charset=UTF-8',
```
kavilla, line 216: we can modify the cy.request which i think the security plugin commands are doing to just add some of these copy and paste request headers

cypress/utils/dashboards/data_explorer/constants.js
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+export const DE_INDEX_DATA = '';
```
kavilla, line 5: empty?


also instead of us needing to import all these individual sometime we can namespace stuff like

```
DE: {
  INDEX: {
     ID: 'data-explorer',
     DATA: '',
  }
...
```
so then we would have to import 3 things we just import one and use dot notation

cypress/utils/dashboards/data_explorer/constants.js
```diff
@@ -0,0 +1,12 @@
...
+export const DE_INDEX_ID = 'data-explorer';
+export const DE_INDEX_PATTERN = DE_INDEX_ID;
+
+export const DE_PATH_FIXTURE = 'dashboard/opensearch_dashboards/data_explorer/';
+
+export const DE_DEFAULT_START_TIME = 'Sep 19, 2015 @ 06:31:44.000';
```
kavilla, line 11: this is a good candidate. for the above comment too but

why not

cypress/integration/core_opensearch_dashboards/apps/query_enhancement/dataset_selector_spec.js
kavilla, line 1: do we have to keep this folder structure?

`core_opensearch_dashboards` is redudant now since we are in the opensearch dashboards repo. we can reduce the messiness now since we should only be our tests for our plugins

kavilla, line 1: ide would be able to let us know what is a `.spec` file but `dataset_selector_spec.js` would just seem like a regular file. so please use `dataset_selector.spec.js`

also why not typescript? @d-rowe is this setup for tyupescript?

kavilla: @d-rowe @abbyhu2000 @ananzh did you guys before porting over want to list out what is the structure of the nested folders we should shoot for? otherwise we will just port of tech debt that is just really busy work. unless this was a drop in replacement why would be suprrising then probably  we are spending the effort to keep the tech debt.

we should just quickly list out what we want the end state to kind of look like

cypress/integration/core_opensearch_dashboards/apps/query_enhancement/dataset_selector_spec.js
```diff
@@ -0,0 +1,180 @@
...
+
+      cy.waitForLoaderNewHeader();
+      cy.getElementByTestId('discoverNoIndexPatterns');
+    });
+  });
+
+  describe('select indices', () => {
+    before(() => {
+      testFixtureHandler.importJSONMapping(
+        'cypress/fixtures/dashboard/opensearch_dashboards/query_enhancement/mappings.json.txt'
```
kavilla, line 38: we could consider a utils function that does this a little bit easier like we know we want to store in fixtures so then we just need to have the plugin define it data it wants to use.

but i dont really see why each plugin needs to define its own fixtures.

if we add a new plugin then the plugin will think it needs to keep adding more sample data that is increasing the resources taken for local development when this data could have been re-used

cypress/integration/core_opensearch_dashboards/apps/query_enhancement/dataset_selector_spec.js
d-rowe: Yup, this package is setup for typescript. It'd be great to have these be typescript files, but actually backfilling the types would be quite an effort during this migration it seems.

cypress/integration/core_opensearch_dashboards/apps/query_enhancement/dataset_selector_spec.js
abbyhu2000: agree on getting a utils function so importing datas can be cleaner. 

Currently changing the fixture file structure so that dashboard and plugin fixtures are not seperated, instead each folder are the data name and contain one data file and one mapping file. 

Ex: 

/fixtures
  /discover
      /data.json
      /mappings.json
  /logstash
      /data.json
      /mappings.json
  /timestamp
      /data.json
      /mappings.json
.....

cypress/integration/core_opensearch_dashboards/apps/query_enhancement/dataset_selector_spec.js
abbyhu2000: Made the file structure a little nicer by getting rid of /core-opensearch-dashboard and /plugin

EX: 
/cypress
   /integration
      /apps
         /query_enhancements
             /test files
         /another_plugin
              /test files

cypress/utils/dashboards/data_explorer/constants.js
abbyhu2000: simplify the importing using namespace, for example:

```
export const PATHS = {
  BASE: BASE_PATH,
  ENGINE: BASE_ENGINE.url,
  SECONDARY_ENGINE: SECONDARY_ENGINE.url,
  STACK_MANAGEMENT: BASE_PATH + '/app/management',
  SECURITY_PLUGIN: BASE_PATH + '/app/security-dashboards-plugin#/',
  TENANTS_MANAGE: BASE_PATH + '/app/security-dashboards-plugin#/tenants',
  INDEX_PATTERNS: BASE_PATH + '/app/management/opensearch-dashboards/indexPatterns',
  SAVED_OBJECTS: BASE_PATH + '/app/management/opensearch-dashboards/objects',
};

```

abbyhu2000:
close this in favor of https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9048

PR #8992, opened 2024-12-02
[Discover] fix PPL to not throw error if aggregation query fails

joshuali925:
### Description

The previous PR https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8743 added `handleFacetError` if date_histogram aggregation failed, because otherwise the code will process the response as data, and calling `rawAggs.data.datarows` caused NPE. 

The problem is that some PPL queries will not have a date_histogram, for example `source = table | stats count()`, and failure is expected. So it should not throw error there. This PR changes `handleFacetError` to `continue`, and renames `handleFacetError` to `throwFacetError`.

### Issues Resolved

- skip

## Testing the changes

added unit tests and locally tested

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/common/utils.test.ts
+++ b/src/plugins/query_enhancements/common/utils.test.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { handleFacetError } from './utils';
+import { throwFacetError } from './utils';
 
 describe('handleFacetError', () => {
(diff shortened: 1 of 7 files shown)


sejli (approved):
Nice catch! Thanks for the search strategy unit tests as well.

Changelog workflow failing because it's under the wrong section in the description.

ruanyl (approved):

PR #8993, opened 2024-12-02
fix: support imports without extensions in cypress webpack build

d-rowe:
### Description
When migrating to cypress v12, we saw a small number of regressions due compilation failures in cypress. This change is to resolve this class of issue. Once resolved in this branch I'll create a PR to merge the cypress 12 upgrade in main.

**Details**
By default, cypress' internal webpack preprocessor doesn't allow imports without file extensions. This makes our life a bit hard since if any file in our testing dependency graph has an import without the .js extension our cypress build will fail.

This configuration relaxes this a bit by allowing imports without file extension
     ex. import module from './module'

## Changelog
- fix: support imports without extensions in cypress webpack build

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/8993.yml
+++ b/changelogs/fragments/8993.yml
@@ -1,0 +1,2 @@
+fix:
+- Support imports without extensions in cypress webpack build ([#8993](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8993))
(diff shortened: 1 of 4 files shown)


d-rowe:
Related: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4813

LDrago27 (approved):

ananzh (approved):

ananzh:
I don't think we need AGGREGATION_VIEW: false and ML_COMMONS_DASHBOARDS_ENABLED: true but I also don't think it is hurtful. But can clean up in a follow-up if we need a more cleaned up config file.

PR #8999, opened 2024-12-03
[workspace]add 2 step loading in 'Associate data sources' modal

Qxisylolo:
### Description
This pr addresses the issue where a single data source has numerous connections, causing it to take more than 20 seconds to list the available data sources.
When a data source has numerous dqc connections, it will display the data source first with a badge 'loading' which will be replaced by '+n related' after dqc connections are completely loaded

## Screenshot(updated)
https://github.com/user-attachments/assets/cbb669d6-bf4a-437b-b622-8cd9f6ae7e43

## Changelog
- feat: Add two-steps loading for associating data sources

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/workspace/public/components/data_source_association/association_data_source_modal.test.tsx
```diff
@@ -23,42 +23,63 @@ const setupAssociationDataSourceModal = ({
...
-      type: 'OpenSearch',
-      relatedConnections: [
+
+  jest.spyOn(utilsExports, 'getOpenSearchAndDataConnections').mockResolvedValue({
+    openSearchConnections: [
+      {
+        id: 'ds1',
+        name: 'Data Source 1',
+        type: 'OpenSearch',
+        connectionType: 0,
```
raintygao, line 33: What about still using `DataSourceConnectionType` enum for this value?

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.test.tsx
```diff
@@ -23,42 +23,63 @@ const setupAssociationDataSourceModal = ({
...
+          id: 'ds1',
+          name: 'Data Source 1',
+          type: 'OpenSearch',
+          connectionType: 0,
+          relatedConnections: [
+            {
+              id: 'ds1-dqc1',
+              name: 'dqc1',
+              type: 'Amazon S3',
+              connectionType: 1,
```
raintygao, line 64: What about still using DataSourceConnectionType enum for this value?

src/plugins/workspace/public/utils.ts
```diff
@@ -539,6 +539,60 @@ export const fetchDataSourceConnections = async (
...
+    notifications?.toasts.addDanger(
+      i18n.translate('workspace.detail.dataSources.error.message', {
+        defaultMessage: 'Cannot fetch direct query connections',
+      })
+    );
+    return [];
+  }
+};
+
+export const getOpenSearchAndDataConnections = async (
```
raintygao, line 572: According to the implementation of function, it seems not an async function?

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
```diff
@@ -230,27 +253,56 @@ export const AssociationDataSourceModalContent = ({
 
   useEffect(() => {
     setIsLoading(true);
+    setIsRelatedConnectionsLoaded(false);
```
raintygao, line 256: Do we need this if the initial state is false?

src/plugins/workspace/public/utils.ts
```diff
@@ -539,6 +539,60 @@ export const fetchDataSourceConnections = async (
...
+  dataConnections: DataSourceConnection[];
+}> => {
+  try {
+    const {
+      openSearchConnections,
+      dataConnections,
+    } = convertDataSourcesToOpenSearchAndDataConnections(dataSources);
+
+    return { openSearchConnections, dataConnections };
+  } catch (error) {
```
raintygao, line 586: Why we need try catch here? `convertDataSourcesToOpenSearchAndDataConnections` seems to just do some array operations

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
```diff
@@ -230,27 +253,56 @@ export const AssociationDataSourceModalContent = ({
...
+          ({ openSearchConnections, dataConnections }) => {
+            setAllConnections([...openSearchConnections, ...dataConnections]);
+            return { openSearchConnections, dataConnections, dataSourcesList };
+          }
+        );
+      })
+      .then(({ openSearchConnections, dataConnections, dataSourcesList }) => {
+        fetchDirectQueryConnections(dataSourcesList, http, notifications).then(
+          (directQueryConnections) => {
+            const updatedOpenSearchConnections = updateFullFillRelatedConnections(
```
raintygao, line 269: looks like too much promise, can we use await to avoid callback hell to make it more clear?

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
```diff
@@ -154,18 +166,28 @@ const convertConnectionsToOptions = ({
         return [];
       }
 
-      if (showDirectQueryConnections) {
-        if (!connection.relatedConnections || connection.relatedConnections.length === 0) {
-          return [];
+      if (connection.connectionType === DataSourceConnectionType.OpenSearchConnection) {
+        if (showDirectQueryConnections) {
+          if (!connection.relatedConnections || connection.relatedConnections.length === 0) {
+            return [connection];
```
raintygao, line 172: Could you add a comment on why we return [connection] when connection is empty in this PR?

Qxisylolo: ok, Thanks for all the comments, all updated.

raintygao: Saw the new comment, if a open search connection has none child connections, will this still work if we set [connection]?

Qxisylolo: yes, if a connection doesn't have related connections, it will just return [connection]. please check the attached video

https://github.com/user-attachments/assets/f668c5d3-bea1-4636-885f-dc2d80715e5e

src/plugins/workspace/public/utils.ts
```diff
@@ -539,6 +539,50 @@ export const fetchDataSourceConnections = async (
   }
 };
 
+export const updateFullFillRelatedConnections = (
+  openSearchConnections: DataSourceConnection[],
+  directQueryConnections: DataSourceConnection[]
+): DataSourceConnection[] => {
+  return fulfillRelatedConnections(openSearchConnections, directQueryConnections);
```
wanglam, line 546: Can't get the point of adding a `updateFullFillRelatedConnections` method, can we use `fulfillRelatedConnections` directly?

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
```diff
@@ -231,26 +255,53 @@ export const AssociationDataSourceModalContent = ({
...
-        setAllConnections(connections);
+      .then((dataSourcesList) => {
+        const { openSearchConnections, dataConnections } = getOpenSearchAndDataConnections(
+          dataSourcesList
+        );
+        setAllConnections([...openSearchConnections, ...dataConnections]);
+        return { openSearchConnections, dataConnections, dataSourcesList };
+      })
+      .then(({ openSearchConnections, dataConnections, dataSourcesList }) => {
+        fetchDirectQueryConnections(dataSourcesList, http, notifications).then(
```
wanglam, line 266: Nit: I'm thinking of do we need to have a loading status for direct query connections for each data source. Since load direct query connections may take long time, but if we separate to each data source, it would be more faster.

src/plugins/workspace/public/utils.ts
```diff
@@ -539,6 +539,50 @@ export const fetchDataSourceConnections = async (
...
+): {
+  openSearchConnections: DataSourceConnection[];
+  dataConnections: DataSourceConnection[];
+} => {
+  const {
+    openSearchConnections,
+    dataConnections,
+  } = convertDataSourcesToOpenSearchAndDataConnections(dataSources);
+
+  return { openSearchConnections, dataConnections };
```
raintygao, line 583: Do we need this function? It seems not contain too much logic.

Qxisylolo: Thanks for pointing the issue, this appears redundant, I will delete it.

src/plugins/workspace/public/utils.ts
Qxisylolo: Thanks for pointing the issue, this appears redundant, I will delete it.

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
Qxisylolo: Thanks for the advice, updated and new screenshot is: 

Uploading 录屏2024-12-11 17.46.05.mov…

yubonluo: Yes, maybe you can call `getDirectQueryConnections` to get the DQCs separately and set the isRelatedConnectionsLoaded status according to the data source id.

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
```diff
@@ -25,7 +25,11 @@ import {
 import { FormattedMessage } from 'react-intl';
 import { i18n } from '@osd/i18n';
 
-import { getDataSourcesList, fetchDataSourceConnections } from '../../utils';
```
ruanyl, line 28: I think we can remove function `fetchDataSourceConnections` if it's not used anymore.

ruanyl (dismissed):

src/plugins/workspace/public/utils.ts
```diff
@@ -539,6 +539,29 @@ export const fetchDataSourceConnections = async (
   }
 };
 
+export const fetchDirectQueryConnectionsByIDs = async (
```
SuZhou-Joe, line 542: I guess we need a unit test for this function.

Qxisylolo: okok, I will add one

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
```diff
@@ -231,26 +254,68 @@ export const AssociationDataSourceModalContent = ({
...
+                });
+              });
+            })
+            .finally(() => {
+              setLoadingStatus((prev) => ({ ...prev, [ds.id]: false }));
+            });
+        });
       })
       .finally(() => {
         setIsLoading(false);
```
SuZhou-Joe, line 300: In OpenSearch Data source case, the loading state will be set to false once the saved data sources being retrieved.
In Direct query connections case, should the loading state be set to false only after all the direct query connections being fetched?

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
```diff
@@ -231,26 +254,68 @@ export const AssociationDataSourceModalContent = ({
...
+        setLoadingStatus(initialLoadingStatus);
+
+        // display data sources connections first while loading direct query connection
+        setAllConnections([...openSearchConnections, ...dataConnections]);
+        return { openSearchConnections };
+      })
+      .then(({ openSearchConnections }) => {
+        openSearchConnections.forEach((ds) => {
+          //  only fetch direct query connections for data source saved object, and set loading status accordingly
+          fetchDirectQueryConnectionsByIDs([ds.id], http, notifications)
```
SuZhou-Joe, line 280: I guess only when in direct query connections case should we fetch DQC info.

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
Qxisylolo: fetchDataSourceConnections is also required in DataSourceTable, so perhaps we should keep it.

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
Qxisylolo: Yes, please see the new update~

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
Qxisylolo: Yes, please see the new update~

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
```diff
@@ -88,42 +92,39 @@ const convertConnectionToOption = ({
...
+          })}
+        </EuiBadge>
+      ),
+    disabled: connection.connectionType === DataSourceConnectionType.DirectQueryConnection,
+    checked:
+      connection.connectionType !== DataSourceConnectionType.DirectQueryConnection &&
+      selectedConnectionIds.includes(connection.id)
+        ? ('on' as const)
+        : undefined,
+    prepend: <ConnectionIcon connection={connection} logos={logos} />,
```
Hailong-am, line 124: do we need to keep the style of indent 16px?

Hailong-am (approved):

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
Qxisylolo: yes, I think the indent is set by default in eui

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
```diff
@@ -229,28 +241,79 @@ export const AssociationDataSourceModalContent = ({
...
+          } as DataSourceConnection;
+        });
+
+      setAllConnections(allConnectionsWithDQC);
+      setIsLoading(false);
+    };
+
+    const fetchDataSourcesAndHandleRelatedConnections = async () => {
+      setIsLoading(true);
+      const { openSearchConnections, dataConnections } = await fetchDataSources();
```
SuZhou-Joe, line 296: Do we need a try catch for this method?

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
```diff
@@ -229,28 +241,79 @@ export const AssociationDataSourceModalContent = ({
...
+      await handleDirectQueryConnections(openSearchConnections, dataConnections);
```
SuZhou-Joe, line 297:
```suggestion
      const { allConnections } = await handleDirectQueryConnections(openSearchConnections, dataConnections);
      setAllConnections(allConnections);
      setIsLoading(false);
```
I feel it will be more clear that we make the function stateless, and we can move the function out of the component to increase the readability of the component and the useEffect hook.

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
```diff
@@ -229,28 +241,79 @@ export const AssociationDataSourceModalContent = ({
...
-      convertConnectionsToOptions({
-        connections: allConnections,
-        excludedConnectionIds,
-        selectedConnectionIds,
-        showDirectQueryConnections: mode === AssociationDataSourceModalMode.DirectQueryConnections,
-        logos,
-      })
-    );
-  }, [allConnections, excludedConnectionIds, selectedConnectionIds, mode, logos]);
+    if (allConnections.length > 0) {
```
SuZhou-Joe, line 304: What if we are trying to reset the options to empty array, with the if the reset may not work.

Qxisylolo: thanks for the reminder ,I will remove the if

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
Qxisylolo: Thanks for the comments, I will update it.

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
SuZhou-Joe: getDataSourcesList may throw error for connection issue, and once it throws error the modal will keeps loading.

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
```diff
@@ -148,24 +149,35 @@ const convertConnectionsToOptions = ({
...
       }
 
-      if (showDirectQueryConnections) {
-        if (!connection.relatedConnections || connection.relatedConnections.length === 0) {
-          return [];
+      if (connection.connectionType === DataSourceConnectionType.OpenSearchConnection) {
+        if (showDirectQueryConnections) {
+          if (!connection.relatedConnections || connection.relatedConnections.length === 0) {
+            // return [connection] for the case where the connnection has no direct connections for now, but it may have in the future
+            return [connection];
```
SuZhou-Joe, line 162: So many if here, can we write a unit test to cover this case?

Qxisylolo: okok, I will cover this

Qxisylolo: Here is a little change as we need to fetch all the dqc first and then perform the filtering(this functionality is already covered in the unit tests.)  I removed the outdated code that previously displayed the data source first during loading.

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
Qxisylolo: Thanks for the explanation, updated~

ruanyl (approved):

SuZhou-Joe (approved):

PR #9004, opened 2024-12-04
[Workspace] fix: Fix workspace page hanging with none collaborators for non dashboard admin

raintygao:
### Description

Fix workspace page hanging with none collaborators for non dashboard admin

Previously the permission field of a workspace with none collaborators is undefined, which will pass the workspace permission wrapper and cause page hanging. Actually this field could be an empty ACL object as workspace type object should always have this field.

## Screenshot

image
Before

image
After

## Testing the changes

1. Use OSD admin to create a workspace then delete all collaborators.
2. Use non OSD admin visit this workspace, then the page will not be hanging and throw error.

## Changelog

- fix: Fix workspace page hanging with none collaborators for non dashboard admin

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/workspace/public/components/workspace_form/utils.ts
```diff
@@ -37,6 +37,13 @@ export const isValidFormTextInput = (input?: string) => {
   return typeof input === 'string' && regex.test(input);
 };
 
+export const EMPTY_PERMISSIONS = {
+  library_read: {},
+  library_write: {},
+  read: {},
+  write: {},
+};
+
```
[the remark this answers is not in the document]
raintygao: Sure. Its type is `SavedObjectPermissions`, have updated.

d-rowe (approved):

wanglam (approved):
LGTM

src/plugins/workspace/public/components/workspace_form/utils.ts
```diff
@@ -104,7 +111,8 @@ export const convertPermissionSettingsToPermissions = (
   permissionItems: WorkspacePermissionSetting[] | undefined
 ) => {
   if (!permissionItems || permissionItems.length === 0) {
-    return undefined;
+    // Workspace object should always have permissions, set it as an empty object here instead of undefined.
+    return EMPTY_PERMISSIONS;
   }
```
Hailong-am, line 116: does it has any impact when permission control is turned off?

raintygao: I supposed it may not as this is a specific object for workspace object.

Hailong-am: when permission control is off, permission field will not initialized and add to kibana index, i am not sure it will cause any issue when try to save a object with empty permission, we can have a test

raintygao: This means the value of permission field will not be consistent between permission is on/off, right? So it seems updating client wrapper will be a better solution?

yubonluo (approved):

Hailong-am (approved):

SuZhou-Joe (approved):

PR #9011, opened 2024-12-05
[Discover Test] TestID-69: Filtering, Test Suite: Filter actions in table field

ArgusLi:
### Description

Adds a cypress test for the Discover Page.

This PR completes the "Filter actions in table field" test suite in TestID-69: Filtering.  There is still the "Filter actions in expanded table" Test Suite to complete for that ticket.

## Changelog
- skip

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


cypress/integration/core_opensearch_dashboards/filter_for_value_spec.js
```diff
@@ -0,0 +1,59 @@
...
+  beforeEach(() => {
+    cy.localLogin(Cypress.env('username'), Cypress.env('password'));
+    miscUtils.visitPage('app/data-explorer/discover');
+    cy.getNewSearchButton().click();
+  });
+  describe('filter actions in table field', () => {
+    describe('index pattern dataset', () => {
+      // filter actions should exist for DQL
+      it('DQL', () => {
+        cy.selectIndexPatternDataset('DQL');
```
d-rowe, line 20: Thanks for fully utilizing utils here across this test file.

cypress.config.ts
```diff
@@ -27,8 +27,9 @@ module.exports = defineConfig({
   },
   e2e: {
     baseUrl: 'http://localhost:5601',
+    supportFile: 'cypress/support/e2e.{js,jsx,ts,tsx}',
     specPattern: 'cypress/integration/**/*_spec.{js,jsx,ts,tsx}',
-    testIsolation: false,
+    testIsolation: true,
```
d-rowe, line 32: As a part of the Cypress I saw a couple existing tests that are not written in an isolated manner, which caused failures when I enabled this initially.

If all our tests pass with isolation on now, that's great! Let's just check the cypress tests on this PR when they run to verify that there are no new failures due to this enablement

cypress/support/e2e.js
```diff
@@ -3,4 +3,14 @@
...
+import '../utils/dashboards/data_explorer/commands.js';
+
+// Alternatively you can use CommonJS syntax:
+// require('./commands')
+
+// eslint-disable-next-line no-unused-vars
+Cypress.on('uncaught:exception', (_err) => {
+  // returning false here prevents Cypress from failing the test
+  return false;
+});
```
d-rowe, line 16: Makes sense, there are some uncaught errors that we do want to gracefully ignore. That being said, I feel there are times when bubbling up an uncaught exception would be really useful. For example, if a page fails to render it'd be nice to see the exception in the runner rather than having to run locally and dig for it.

Can we be a bit more selective here and only swallow the error you were seeing needed to be handled gracefully?

Ex.
```
const resizeObserverLoopErr = /^[^(ResizeObserver loop limit exceeded)]/;
Cypress.on('uncaught:exception', (err) => {
  /* returning false here prevents Cypress from failing the test */
  if (resizeObserverLoopErr.test(err.message)) {
    return false;
  }
});
```

cypress.config.ts
ArgusLi: I don't think that the other tests are running, I will revert the testIsolation to false and change my test to reflect the lack of test isolation.

cypress/support/e2e.js
ArgusLi: Sounds good. I added this as I was getting a specific navigation error. Keeping the ignore selective makes a lot of sense to me.

cypress.config.ts
ArgusLi: Addressed in newest commit.

ArgusLi:
Superseded by #9119

PR #9027, opened 2024-12-10
[Discover Test] TestID-69: Filtering. Add data-test-subj tags.

ArgusLi:
### Description

Adds data-test-subj tags required for TESTID-69.

## Changelog
- skip


cypress.config.ts
LDrago27, line 1: @ArgusLi Remove this file. We only want the core repository changes.

ananzh (approved):

PR #9074, opened 2024-12-18
refactor some query-enhancement utility functions, add some missing typings

angle943:
### Description

Refactors/additions for Cypress Query Enhancements

- update the custom command `waitForLoader` to take optional arg `isEnhancement: boolean`
- remove the `cy.wait()` within `waitForLoader` to speed up the command
- remove the now redundant `waitForLoaderNewHeader` command and refactor its' usage to be `waitForLoader(true)`
- add typescript typings to some of our custom commands
- refactor createWorkspaceIndexPatterns
    - conditionally call `cy.get('[type="data-source"]').contains(dataSource).click();` if `dataSource` prop is provided. This is because if your ws only has one data source, then this step is not visible in the UI, and this utility will fail
    - add optional arg `indexPatternHasTimefield: boolean`. This is needed because if the index pattern you select does not have a timefield, then the option of choosing a timestamp field does not appear, causing the test to fail
    - add the `isEnhancement: boolean` prop
 - add `deleteWorkspaceIndexPatterns` utility
 - refactor `navigateToWorkSpaceSpecificPage` to be less flaky
    - one thing to call out. I noticed a lot of flaky tests due to the nature of our left menu bar. The left menu bar's `open` state persists through navigation and refreshes. Therefore, in order to maintain consistent state before & after a function is run, if a function opens the bar, we should also add a step that closes it. This way we do not have to keep track of it's `open` state.

### Issues Resolved

-n/a

## Testing the changes

- ran cypress tests

## Changelog

- test: query-enhancements testing utility updates and additions

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/cypress/utils/dashboards/index.d.ts
+++ b/cypress/utils/dashboards/index.d.ts
@@ -1,0 +1,31 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+declare namespace Cypress {
+  interface Chainable<Subject> {
+    navigateToWorkSpaceHomePage(url: string, workspaceName: string): Chainable<any>;
+    navigateToWorkSpaceSpecificPage(opts: {
+      url: string;
+      workspaceName: string;
+      page: string;
+      isEnhancement?: boolean;
+    }): Chainable<any>;
+    createWorkspaceIndexPatterns(opts: {
+      url: string;
+      workspaceName: string;
+      indexPattern: string;
+      timefieldName?: string;
+      indexPatternHasTimefield?: boolean;
+      dataSource?: string;
+      isEnhancement?: boolean;
+    }): Chainable<any>;
+    deleteWorkspaceIndexPatterns(opts: {
+      url: string;
+      workspaceName: string;
+      indexPattern: string;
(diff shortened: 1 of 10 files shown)


cypress/utils/dashboards/index.d.ts
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+declare namespace Cypress {
```
ananzh, line 6: I like this. We should add ts support to all the utilities, which can allow proper autocomplete and type check. I think we used to have in ftr repo.

ananzh (approved):

cypress/utils/dashboards/index.d.ts
angle943: yeah i saw them more frequently in the ftr repo. I didn't add every single one here, i was mainly adding the ones that I was using just for my development benefit (IDE auto-complete) and decided to commit it to this PR. it would be great if later we can have full TS support for all of our utilities

LDrago27 (approved):
LGTM

PR #9079, opened 2024-12-19
chore: use relative paths in cypress

d-rowe:
### Description
This change is to remove absolute urls in cypress utilities in favor of relative paths. Since these paths are different for each environment, we want to be able to use dynamic base urls to unify the consumption across environments.

## Changelog
- chore: use relative paths in cypress utilities

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


angle943 (approved):

angle943:
LGTM

wanglam (approved):
LGTM!

ananzh (approved):
LGTM

sejli (approved):

cypress/utils/commands.js
```diff
@@ -104,9 +104,7 @@ Cypress.Commands.add('bulkUploadDocs', (fixturePath, index) => {
 
 Cypress.Commands.add('importSavedObjects', (fixturePath, overwrite = true) => {
   const sendImportRequest = (ndjson) => {
-    const url = `${Cypress.config().baseUrl}/api/saved_objects/_import?${
```
kavilla, line 107: does this work everytime?  for example if the base path is defined for example `_dashboards` for the app is the API also defined `_dashboards` if i remember correctly i think the base path is not added to APIs

d-rowe: For now, the net result should be the same given Cypress implicitly prefixes the url with the basePath already ([ref](https://docs.cypress.io/app/core-concepts/best-practices#Setting-a-Global-baseUrl)). Let me know if I'm missing something here though.

Issue #8945, opened 2024-11-27
[Field Display and Filtering] TESTID-69: Filtering


LDrago27:
<h2 data-pm-slice="1 1 []">Objective</h2><p>Verify filtering functionality in discover table works correctly across different data sources and query languages.</p><h2>What to test</h2><p>Test field filtering through:</p><ol class="ProsemirrorEditor-list"><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="numbered" style="margin-left: 32px"><p>Filter for/out buttons in table field</p></li><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="numbered" style="margin-left: 32px"><p>Document expansion filtering options</p></li></ol><p><br></p><p>Data type:</p><ol class="ProsemirrorEditor-list"><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>Index Patterns</p></li><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>Indexes</p></li><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>S3 Connections</p></li><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>Cloudwatch</p></li><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>Security Lake</p></li></ol><p>Language:</p><ol class="ProsemirrorEditor-list"><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>DQL</p></li><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>Lucene</p></li><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>PPL</p></li><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>SQL</p></li></ol><p><br></p><p>Test suite: filter actions in table field</p>

| Dataset       | Steps | Expected Results |
|---------------|-------|------------------|
[...]

PR #9112, opened 2024-12-21
Add tests for saved search creation and loading for query enhancement

angle943:
### Description

Adds cypress tests for query enhanced env. Creates a saved search and loads it.

### Issues Resolved

closes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8949
closes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8948

## Changelog

- test: add tests for saving search and loading it

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [x] Commits are signed per the DCO using --signoff

PR #9119, opened 2024-12-24
[Discover 2.0 Testing] Create Filtering Test (TestId-69: Filtering)

ArgusLi:
### Description

Adds a Cypress test suite for filtering in the Discover page.

### Issues Resolved

Closes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8945.

## Screenshot

https://github.com/user-attachments/assets/cece1f2f-5b25-4aab-a8de-5b776b218de4

https://github.com/user-attachments/assets/dc7161eb-75c5-4639-ba35-1dad146eb3fb

https://github.com/user-attachments/assets/5874c6bc-edcd-4608-a3f0-d0eb0dd61e40

https://github.com/user-attachments/assets/ef443dec-e1ef-428d-84e4-085dbb0eb1e5

https://github.com/user-attachments/assets/7c7fc374-409c-4b9f-abe2-84a03de8ce8b

https://github.com/user-attachments/assets/ae5a7c11-95dc-401a-af06-2334ad4418b4

https://github.com/user-attachments/assets/490de8ec-b3b2-4aaa-83ee-be3326fc156e

https://github.com/user-attachments/assets/9666d167-b164-406a-a446-1ca1bfe6b41f

## Testing the changes

With OSD running, run `yarn run cypress open`. In E2E specs, you will see a new test spec `field_display_filter.spec.js`. Select it and run it.

## Changelog

- test: Add cypress integration test for the new Data Explorer UI's Filtering.


cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/field_display_filtering.spec.js
```diff
@@ -0,0 +1,150 @@
...
+    cy.createWorkspaceIndexPatterns({
+      url: `${BASE_PATH}`,
+      workspaceName: `${WORKSPACE_NAME}`,
+      indexPattern: 'data_logs_small_time_1',
+      timefieldName: 'timestamp',
+      indexPatternHasTimefield: true,
+      dataSource: DATASOURCE_NAME,
+      isEnhancement: true,
+    });
+    cy.navigateToWorkSpaceHomePage(`${BASE_PATH}`, `${WORKSPACE_NAME}`);
```
LDrago27, line 118: You can move this to the beforeEach block. So that beforeEach test we navigate to the Discover and start from there

ArgusLi: Would it be better to keep it in before? We are creating a new search in the beforeEach block, navigating to Discover again seems to take extra time for no real benefit.

cypress/utils/commands.js
```diff
@@ -30,10 +30,49 @@ Cypress.Commands.add('getElementsByTestIds', (testIds, options = {}) => {
...
+ * @param {object} options get options. Default: {}
+ * @example
+ * // returns all DOM elements that has a data-test-subj including the string 'table'
+ * cy.findElementsByTestIdLike('table')
+ */
+Cypress.Commands.add(
+  'findElementByTestIdLike',
+  { prevSubject: true },
+  (subject, partialTestId, options = {}) => {
+    return cy.wrap(subject).find(`[data-test-subj*="${partialTestId}"]`, options);
```
angle943, line 58: why is it partialTestId? would this allow for partial match? If so what's the mechanism that makes that work? The `*`? Do you have any doc reference to this by any chance? (For my curiosity)

cypress/utils/commands.js
```diff
@@ -30,10 +30,49 @@ Cypress.Commands.add('getElementsByTestIds', (testIds, options = {}) => {
...
+ * @param {string} testId data-test-subj value.
+ * @param {object} options get options. Default: {}
+ */
+Cypress.Commands.add(
+  'findElementByTestId',
+  { prevSubject: true },
+  (subject, testId, options = {}) => {
+    return cy.wrap(subject).find(`[data-test-subj="${testId}"]`, options);
+  }
+);
```
angle943, line 74: Could we update the associated `index.d.ts` to include typings for these?

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
```diff
@@ -0,0 +1,342 @@
...
+) {
+  getExpandedDocRowValue(docTableRowNumber, expandedDocumentRowNumber).then(
+    ($expandedDocumentRowValue) => {
+      const filterFieldText = $expandedDocumentRowValue.text();
+      getExpandedDocRow(docTableRowNumber, expandedDocumentRowNumber).within(() => {
+        cy.getElementByTestId('removeInclusiveFilterButton').click();
+      });
+      // Verify pill text
+      cy.getElementByTestId('globalFilterLabelValue', {
+        timeout: 10000,
```
angle943, line 268: we have a `default` global timeout set to a ridiculous number. Did that not get applied for you?

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
```diff
@@ -0,0 +1,342 @@
...
+  });
+}
+
+/**
+ * Check the query hit text matches expectedQueryHitText.
+ * @param {string} expectedQueryHitsText expected text for query hits. Commas must be added e.g. 10,000
+ */
+export function verifyQueryHitsText(expectedQueryHitsText) {
+  cy.getElementByTestId('discoverQueryHits').should('have.text', expectedQueryHitsText);
+}
```
angle943, line 145: You should be able to use this: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/cypress/utils/apps/data_explorer/commands.js#L44 (although looking at it now, i wonder if it accounts for the `commas`. If not i think we should update that one to account for commas

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
```diff
@@ -0,0 +1,342 @@
...
+    .find('td')
+    .eq(1) // Field name is in the second column.
+    .find('span[class*="textTruncate"]');
+}
+
+/**
+ * Select a language in the Dataset Selector for Index
+ * @param {string} datasetLanguage Index supports "OpenSearch SQL" and "PPL"
+ */
+export function selectIndexDatasetLanguage(datasetLanguage) {
```
angle943, line 78: fyi lot of these i will include as global commands in my PR (ones related to setting dataset)

angle943:
Screenshot 2024-12-27 at 3 17 51 PM

i have pulled this PR locally and tried, but getting failure at first test

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/field_display_filtering.spec.js
```diff
@@ -0,0 +1,178 @@
...
+    // Load test data
+    cy.setupTestData(
+      SECONDARY_ENGINE.url,
+      ['cypress/fixtures/query_enhancements/data-logs-1/data_logs_small_time_1.mapping.json'],
+      ['cypress/fixtures/query_enhancements/data-logs-1/data_logs_small_time_1.data.ndjson']
+    );
+    // Add data source
+    cy.addDataSource({
+      name: `${DATASOURCE_NAME}`,
+      url: 'http://opensearch-node:9200',
```
LDrago27, line 128: Replace  this with   `url: `${SECONDARY_ENGINE.url}`` . 

USe `url: 'http://opensearch-node:9200'` only in local while you are testing you changes in the docker setup

ArgusLi: Fixed, thank you.

ArgusLi:
> Screenshot 2024-12-27 at 3 17 51 PM
> 
> 
> 
> i have pulled this PR locally and tried, but getting failure at first test

This should be fixed now.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/field_display_filtering.spec.js
```diff
@@ -0,0 +1,150 @@
...
+  DATASOURCE_NAME,
+  DATASET_CONFIGS,
+  INDEX_NAME,
+  INDEX_PATTERN_NAME,
+  START_TIME,
+  END_TIME,
+  WORKSPACE_NAME,
+} from './constants.js';
+import * as dataExplorer from './helpers.js';
+import { SECONDARY_ENGINE, BASE_PATH } from '../../../../../utils/constants';
```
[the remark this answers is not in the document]
ArgusLi: I haven't been able to successfully arrow functions for some reason: https://opensearch.slack.com/archives/C082BN1CPE1/p1733433369488929

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/field_display_filtering.spec.js
```diff
@@ -0,0 +1,150 @@
...
+
+function setDateRange(datasetType, language) {
+  switch (datasetType) {
+    case 'index_pattern':
+      if (language !== 'OpenSearch SQL') {
+        cy.setSearchAbsoluteDateRange(START_TIME, END_TIME);
+      }
+      break;
+  }
+}
```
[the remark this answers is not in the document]
ArgusLi: The test spec did not specify if I needed to have a time field for indices, so I left them as "I don't want to use a time filter". If we want to use time filters moving on, I can make the change.

cypress/utils/commands.js
ArgusLi: I time boxed a couple hours trying to switch to ts with no avail. I'll try again on Monday given you've successfully added typing.

cypress/utils/commands.js
ArgusLi: This is a partial match. As long as the data-test-subj contains that partialTestId it'll find it. 

This is not a Cypress specific selector, rather a CSS selector: https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors

cypress/utils/dashboards/data_explorer/commands.js
```diff
@@ -0,0 +1,122 @@
...
+ */
+
+import { DATA_EXPLORER_PAGE_ELEMENTS } from './elements.js';
+
+/**
+ * Get the New Search button.
+ */
+Cypress.Commands.add('getNewSearchButton', () => {
+  return cy.getElementByTestId(DATA_EXPLORER_PAGE_ELEMENTS.NEW_SEARCH_BUTTON, { timeout: 10000 });
+});
```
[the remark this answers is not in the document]
ArgusLi: I've removed a bunch of these after our discussion with Miki. Some must have slipped the net. Going through them again.

cypress/utils/dashboards/data_explorer/commands.js
```diff
@@ -0,0 +1,122 @@
...
+  cy.getSearchDatePickerEndDateButton().click();
+  cy.getDatePickerAbsoluteTab()
+    .should(($elements) => {
+      // This prevents there being 2 Absolute Tabs returned.
+      expect($elements).to.have.length(1);
+    })
+    .click();
+  cy.getDatePickerAbsoluteInput().clear().type(absoluteEndDate);
+  cy.getQuerySubmitButton().click();
+});
```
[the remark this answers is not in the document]
ArgusLi: Wish I saw your comment before I implemented my solution. Yours is more elegant. Will be changing to use yours.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/field_display_filtering.spec.js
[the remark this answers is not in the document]
angle943: Class method's arrow function usage is relatively newer implementation, but arrow functions for regular functions should still work. Could you give it a try?

ananzh:
can we put helpers.js into utilities `utils/apps` ?

cypress/utils/dashboards/data_explorer/commands.js
```diff
@@ -0,0 +1,122 @@
...
+ * // sets search to return results from 15 years ago to now
+ * setSearchRelativeDateRange('15', 'years ago')
+ */
+Cypress.Commands.add('setSearchRelativeDateRange', (relativeNumber, relativeUnit) => {
+  cy.getSearchDatePickerButton().click();
+  cy.getDatePickerRelativeTab().click();
+  cy.getDatePickerRelativeInput().clear().type(relativeNumber);
+  cy.getDatePickerRelativeUnitSelector().select(relativeUnit);
+  cy.getQuerySubmitButton().click();
+});
```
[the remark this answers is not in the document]
ananzh: agree. normally, we don't use relative date range due to fixed time range data.

cypress/utils/dashboards/data_explorer/commands.js
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Set search to an absolute Date using searchDatePickerButton (this button only appears if the start date is "now" and end date is a relative date, or vice versa)
+ * @param absoluteStartDate String for Absolute Datetime for start date in format 'MMM dd, yyyy @ HH:mm:ss.SSS'.
+ * @param absoluteEndDate String for Absolute Datetime for end date in format 'MMM dd, yyyy @ HH:mm:ss.SSS'.
+ * @example setSearchAbsoluteDateRangeWithSearchDatePickerButton('Dec 31, 2020 @ 16:00:00.000', 'Dec 31, 2022 @ 14:14:42.801')
+ */
+Cypress.Commands.add(
```
ananzh, line 109: Is this to set absolute time in the time range? I think the current method can set absolute time range right? why create a new one?
```
Cypress.Commands.add("setTopNavDate", (start, end, submit = true) => {
  const opts = { log: false };

  Cypress.log({
    name: "setTopNavDate",
    displayName: "set date",
    message: `Start: ${start} :: End: ${end}`,
  });

  /* Find any one of the two buttons that change/open the date picker:
   *   * if `superDatePickerShowDatesButton` is found, it will switch the mode to dates
   *      * in some versions of OUI, the switch will open the date selection dialog as well
   *   * if `superDatePickerstartDatePopoverButton` is found, it will open the date selection dialog
   */
  cy.getElementsByTestIds(
    ["superDatePickerstartDatePopoverButton", "superDatePickerShowDatesButton"],
    opts
  )
    .should("be.visible")
    .invoke("attr", "data-test-subj")
    .then((testId) => {
      cy.getElementByTestId(testId, opts).should("be.visible").click(opts);
    });

  /* While we surely are in the date selection mode, we don't know if the date selection dialog
   * is open or not. Looking for a tab and if it is missing, click on the dialog opener.
   */
  cy.whenTestIdNotFound("superDatePickerAbsoluteTab", () => {
    cy.getElementByTestId("superDatePickerstartDatePopoverButton", opts)
      .should("be.visible")
      .click(opts);
  });

  // Click absolute tab
  cy.getElementByTestId("superDatePickerAbsoluteTab", opts).click(opts);

  // Type absolute start date
  cy.getElementByTestId("superDatePickerAbsoluteDateInput", opts)
    .click(opts)
    .clear(opts)
    .type(start, {
      ...opts,
      delay: 0, // add a delay here, cypress sometimes fails to type all the content into the input.
    });

  // Click end date
  cy.getElementByTestId("superDatePickerendDatePopoverButton", opts)
    .last(opts)
    .click(opts);

  // Click absolute tab
  cy.getElementByTestId("superDatePickerAbsoluteTab", opts)
    .last(opts)
    .click(opts);

  // Type absolute end date
  cy.getElementByTestId("superDatePickerAbsoluteDateInput", opts)
    .last(opts)
    .click(opts)
    .clear(opts)
    .type(end, {
      ...opts,
      delay: 0, // add a delay here, cypress sometimes fails to type all the content into the input.
    });

  // Close popup
  cy.getElementByTestId("superDatePickerendDatePopoverButton", opts).click(
    opts
  );

  if (submit) {
    cy.updateTopNav(opts);
  }
});
```

cypress/utils/dashboards/data_explorer/elements.js
```diff
@@ -0,0 +1,41 @@
...
+  DOC_TABLE_EXPANDED_DOC_COLUMN_REMOVE_INCLUSIVE_FILTER_BUTTON: 'removeInclusiveFilterButton',
+  DOC_TABLE_EXPANDED_DOC_COLUMN_EXISTS_FILTER_BUTTON: 'addExistsFilterButton',
+  DOC_TABLE_HEADER_FIELD_PREFIX: 'docTableHeader-',
+  TABLE_FIELD_FILTER_FOR_BUTTON: 'filterForValue',
+  TABLE_FIELD_FILTER_OUT_BUTTON: 'filterOutValue',
+  SEARCH_DATE_PICKER_BUTTON: 'superDatePickerShowDatesButton',
+  SEARCH_DATE_PICKER_START_DATE_BUTTON: 'superDatePickerstartDatePopoverButton',
+  SEARCH_DATE_PICKER_END_DATE_BUTTON: 'superDatePickerendDatePopoverButton',
+  SEARCH_DATE_PICKER_RELATIVE_TAB: 'superDatePickerRelativeTab',
+  SEARCH_DATE_RELATIVE_PICKER_INPUT: 'superDatePickerRelativeDateInputNumber',
```
ananzh, line 31: should remove these relative picker constants.
for absolute ones, they are only used and wrapped in `setTopNavDate` method. so no other places will use them, we don't need constants for them.

cypress/utils/dashboards/data_explorer/elements.js
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const DATA_EXPLORER_PAGE_ELEMENTS = {
```
ananzh, line 6: if we don't use OOP any more, should we still wrap constants in a big object? we don't need to create constants for test-subjs. The benefit for constants here is that if the test-id needs to change, we only update it in one place. But we rarely update test-subjs. The existing tests and our legacy tests all just call test-subjs directly.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
```diff
@@ -0,0 +1,342 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Get specific row of DocTable.
+ * @param {number} rowNumber Integer starts from 0 for the first row
+ */
+export function getDocTableRow(rowNumber) {
```
ananzh, line 10: let me move this helpers to utils/apps/data_explorer/commands.js and move query specific to utils/apps/query_enhancements/commands.js

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
```diff
@@ -0,0 +1,342 @@
...
+ */
+export function getDocTableField(columnNumber, rowNumber) {
+  return getDocTableRow(rowNumber).findElementByTestId('docTableField').eq(columnNumber);
+}
+
+/**
+ * find all Rows in Doc Table Field Expanded Document.
+ * @param expandedDocument cypress representation of the Doc Table Field Expanded Document
+ */
+export function findDocTableExpandedDocRowsIn(expandedDocument) {
```
ananzh, line 27: should we name it to getExpandedDocTableRows?

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
```diff
@@ -0,0 +1,342 @@
...
+
+/**
+ * Get the "expandedDocumentRowNumber"th row from the expanded document from the "docTableRowNumber"th row of the DocTable.
+ * @param {number} docTableRowNumber Integer starts from 0 for the first row
+ * @param {number} expandedDocumentRowNumber Integer starts from 0 for the first row
+ * @example
+ * // returns the first row from the expanded document from the second row of the DocTable.
+ * getExpandedDocRow(1, 0);
+ */
+export function getExpandedDocRow(docTableRowNumber, expandedDocumentRowNumber) {
```
ananzh, line 39: getExpandedDocTableRow

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
ananzh: if the utilities are too many, we can create sub-files. like utils/apps/data_explorer/doc_table/commands.js and utils/apps/data_explorer/side_panel/commands.js

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
```diff
@@ -0,0 +1,342 @@
...
+
+/**
+ * Check the first expanded Doc Table Field's first row's Exists Filter button filters the correct Field.
+ * @param {number} docTableRowNumber Integer starts from 0 for the first row
+ * @param {number} expandedDocumentRowNumber Integer starts from 0 for the first row
+ * @param {string} expectedQueryHitsWithoutFilter expected number of hits in string after the filter is removed Note you should add commas when necessary e.g. 9,999
+ * @param {string} expectedQueryHitsAfterFilterApplied expected number of hits in string after the filter is applied. Note you should add commas when necessary e.g. 9,999
+ */
+export function verifyDocTableFirstExpandedFieldFirstRowExistsFilterButtonFiltersCorrectField(
+  docTableRowNumber,
```
ananzh, line 316: why we need a specific utility for the first row?

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
```diff
@@ -0,0 +1,342 @@
...
+  });
+  cy.getElementByTestId('globalFilterBar').find('[aria-label="Delete"]').click();
+  cy.getElementByTestId('discoverQueryHits').should('have.text', expectedQueryHitsWithoutFilter);
+}
+
+/**
+ * Check if the first expanded Doc Table Field's first row's Filter For, Filter Out and Exists Filter buttons are disabled.
+ * @param {boolean} isEnabled Boolean determining if these buttons are disabled
+ */
+export function verifyDocTableFirstExpandedFieldFirstRowFilterForFilterOutExistsFilterButtons(
```
ananzh, line 199: Why we need a utility to check enable or disable? We can write directly in the test, right?
```
it('verifies filter buttons state', () => {
  const shouldText = isEnabled ? 'be.enabled' : 'be.disabled';
  cy.getElementByTestId('addInclusiveFilterButton').should(shouldText);
  cy.getElementByTestId('removeInclusiveFilterButton').should(shouldText);
  cy.getElementByTestId('addExistsFilterButton').should(shouldText);
});
```

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
```diff
@@ -0,0 +1,342 @@
...
+}
+
+/**
+ * Check the Filter For button in the expandedDocumentRowNumberth field in the expanded Document filters the correct value.
+ * @param {number} docTableRowNumber Integer starts from 0 for the first row
+ * @param {number} expandedDocumentRowNumber Integer starts from 0 for the first row
+ * @param {string} expectedQueryHitsWithoutFilter expected number of hits in string after the filter is removed Note you should add commas when necessary e.g. 9,999
+ * @param {string} expectedQueryHitsAfterFilterApplied expected number of hits in string after the filter is applied. Note you should add commas when necessary e.g. 9,999
+ */
+export function verifyDocTableFirstExpandedFieldFirstRowFilterForButtonFiltersCorrectField(
```
ananzh, line 217: can we combine this filter for and filter out  into one? and simplify toggle (exists) one to see if we combine all three?

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/constants.js
```diff
@@ -5,5 +5,30 @@
...
+    ],
+  },
+  index: {
+    type: 'index',
+    name: INDEX_NAME,
+    languages: [
+      { name: 'OpenSearch SQL', isEnabled: false },
+      { name: 'PPL', isEnabled: false },
+    ],
+  },
```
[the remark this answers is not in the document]
ArgusLi: These `isEnabled` check if the filter buttons are supposed to be enabled or not. I'll add a comment.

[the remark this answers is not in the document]
ArgusLi: Actually, let me rename the variable. That would be more clear I think.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/field_display_filtering.spec.js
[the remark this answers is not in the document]
ArgusLi: Works! I've changed it.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/field_display_filtering.spec.js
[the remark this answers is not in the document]
ArgusLi: After discussions, we need to be using timestamps and time fields. Making the change.

cypress/utils/dashboards/data_explorer/commands.js
ArgusLi: This problem only surfaced due to workspaces caching the absolute start and end dates. I didn't realise that Justin's merged pr already had that solution. Will be replacing.

cypress/utils/commands.js
ArgusLi: Cypress uses the term **get** for the entire DOM and **find** for within the previously selected element. I think sticking with that convention is better.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
ArgusLi: I think findExpandedDocTableRows would be better to stick with the find naming convention used by cypress.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
ArgusLi: Ok I have combined for and out into one. The toggle is very different so I'm not combining it.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
ArgusLi: Updated to remove these timeouts.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
ArgusLi: Changed to use that. I'm not going to make that change in the current pr, but I do agree. It should be taking an integer as well, instead of a string that you have to manually add commas.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/helpers.js
ArgusLi: Moving it to the spec instead.

cypress/utils/commands.js
ArgusLi: Added typings.

cypress/utils/commands.js
angle943: ah! my css newb is showing! TIL thanks!

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/field_display_filtering.spec.js
```diff
@@ -0,0 +1,245 @@
...
+  };
+
+  /**
+   * Check the Filter For or Out buttons in the expandedDocumentRowNumberth field in the expanded Document filters the correct value.
+   * @param {string} filterButton For or Out
+   * @param {number} docTableRowNumber Integer starts from 0 for the first row
+   * @param {number} expandedDocumentRowNumber Integer starts from 0 for the first row
+   * @param {string} expectedQueryHitsWithoutFilter expected number of hits in string after the filter is removed Note you should add commas when necessary e.g. 9,999
+   * @param {string} expectedQueryHitsAfterFilterApplied expected number of hits in string after the filter is applied. Note you should add commas when necessary e.g. 9,999
+   * @example verifyDocTableFirstExpandedFieldFirstRowFilterForButtonFiltersCorrectField('for', 0, 0, '10,000', '1');
```
angle943, line 74: Screenshot 2024-12-31 at 12 58 22 PM

Nice! I did not know about @example

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/field_display_filtering.spec.js
```diff
@@ -0,0 +1,245 @@
...
+      indexPattern: INDEX_NAME,
+      timefieldName: 'timestamp',
+      indexPatternHasTimefield: true,
+      dataSource: DATASOURCE_NAME,
+      isEnhancement: true,
+    });
+  });
+
+  beforeEach(() => {
+    cy.navigateToWorkSpaceHomePage(`${BASE_PATH}`, `${workspace}`);
```
angle943, line 216: can we change this to `navigateToWorkSpaceSpecificPage` with `page: 'discover'`? See https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9112#discussion_r1899810614 for rationale

cypress/utils/apps/data_explorer/commands.js
```diff
@@ -162,3 +162,187 @@ function checkForElementVisibility() {
       }
     });
 }
+
```
angle943, line 165: - All the newly added functions here are regular functions, not cypress commands. Since this file is `commands.js`, only Cypress.Commands.add functions should be here
- So we got two options to rectify this. Either: 1) Make all these functions into cypress custom commands or 2) move these functions out to another file . I am going to vote for choice 2 for the following reason:
    - [Cypress best practices recommends it](https://docs.cypress.io/api/cypress-api/custom-commands#Best-Practices) (See `1. Don't make everything a custom command`) (I personally feel that we use too much cypress commands in the codebase already)
    
As to where we should move these... This is just a suggestion but what I did was create a `utils` directory next to the tests itself: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9112/files#diff-80ee042bf13b443df19ca0fbb20fc6242ca982791beef74d65eb1d8d74f39f89

cypress/utils/apps/data_explorer/index.d.ts
```diff
@@ -19,5 +19,33 @@ declare namespace Cypress {
...
+    selectIndexPatternDataset(indexPatternName: string, datasetLanguage: string);
+    toggleDocTableRow(rowNumber: number);
+    verifyDocTableFilterAction(
+      rowNumber: number,
+      filterElement: string,
+      expectedQueryHitsWithoutFilter: string,
+      expectedQueryHitsAfterFilterApplied: string,
+      shouldMatch: boolean
+    );
+    verifyDocTableFirstExpandedFieldFirstRowToggleColumnButtonHasIntendedBehavior();
```
angle943, line 49: Because these aren't Cypress custom commands, the types shouldn't be added here nested within `declare namespace Cypress { interface Chainable`. Because then our IDEs think we can chain these off of `cy`:



If you move these functions to a separate file, we don't need to create a typescript dec file for it, since you already JSDoc-annotated it

cypress/utils/apps/query_enhancements/constants.js
```diff
@@ -7,3 +7,29 @@ export const DATASOURCE_NAME = 'data-logs-1';
...
+  },
+  index: {
+    type: 'index',
+    name: INDEX_NAME,
+    languages: [
+      { name: 'OpenSearch SQL', isFilterButtonsEnabled: false },
+      { name: 'PPL', isFilterButtonsEnabled: false },
+    ],
+  },
+};
```
angle943, line 35: btw if my PR gets merged, you can leverage it: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9112/files#diff-d7928a5f6f39c85900fa1e53e0feef6c1c26ea04701b1eb658be5eb2e89d9bd1R104

For example, `isFilterButtonsEnabled` in my datatype is here: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9112/files#diff-d7928a5f6f39c85900fa1e53e0feef6c1c26ea04701b1eb658be5eb2e89d9bd1R54

cypress/utils/commands.js
```diff
@@ -30,10 +30,49 @@ Cypress.Commands.add('getElementsByTestIds', (testIds, options = {}) => {
   return cy.get(selectors.join(','), options);
 });
 
-Cypress.Commands.add('getElementByDataTestId', (testId) => {
-  return cy.get(`[data-testid="${testId}"]`);
+/**
```
angle943, line 33: it doesn't appear these JSDocs help with IDEs, at least on my machine. Since these aren't regular functions, I don't think JSDocs will help us here. Could we remove them and instead add the typings to here? https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/cypress/utils/index.d.ts

angle943 (changes requested):

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/field_display_filtering.spec.js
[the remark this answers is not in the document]
angle943: I'm a bit confused. how are time field for indices related to the original comment?

cypress/utils/apps/data_explorer/commands.js
ArgusLi: I agree that we should move them to their own file. I will move them to a similarly named utils file.

angle943 (approved):

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/field_display_filtering.spec.js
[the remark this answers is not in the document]
ArgusLi: Sorry, I forgot to reply to this comment. If the index time field is set to I don't want to use a time filter, there is no way to set time for that index. Does that answer your question?

LDrago27 (approved):

ananzh (approved):

PR #9120, opened 2024-12-24
[Auto Suggest] Fix Grammar Changes

paulstn:
### Description

Redos changes made to the grammar meant for fixing bugs. Fixes those same original bugs through other means.

Also rewrites the 'redo and combine' functionality to be more robust.
### The old system
The old system was simply a flag that could be set by a preferred rule to run code completion once with all the rules, and then another time without any rules, to finally combine the results from both.
### The new system
The new system requires any preferred rule to submit its name, and then goes through that cumulative list to remove only that rule from the code completion process, combining every possible result it gets. This new system allows for multiple preferred rules to be left out of the code completion process if it happens to block keyword candidates, and also for every other potentially necessary preferred rule to still be in effect while it does so.

Diagram: (change made was for the "Main Parsing Loop")
image

----
Removes grammar changes from:
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8087/files#diff-3dc92487ddb63ebd741b8639f39b9440158a1f31317978cbd84af144daf3adc0

### Each grammar change reverted / with its corresponding fix:

- [search source with logical expressions](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8087/files#diff-3dc92487ddb63ebd741b8639f39b9440158a1f31317978cbd84af144daf3adc0L56-R57), reverting required a preferred rule on logical expression to prevent code completion issues and a new redo and combine function
- [qualified name (including field names) using ident instead of ID and removing the parsing ability of DOTs](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8087/files#diff-3dc92487ddb63ebd741b8639f39b9440158a1f31317978cbd84af144daf3adc0L777-R778), fixes parsing DOTs in field names, and issues with field name appearing more than when its needed was mitigated by adding logic to the field name pref. rule to only suggest if the last token isn't ID, backticked ID, or DOT
- [table function and table source clause allowing for tables to be specified many times in the same query
](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8087/files#diff-3dc92487ddb63ebd741b8639f39b9440158a1f31317978cbd84af144daf3adc0L171-R176), handled through a pref. rule but some behavior is left as is.

## Changelog

- fix: PPL Grammar Parsing related issues

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/public/antlr/opensearch_sql/opensearch_sql_autocomplete.ts
```diff
@@ -223,6 +223,11 @@ export function processVisitedRules(
               '.' + removePotentialBackticks(nextToken?.text ?? '');
             continue;
           }
+          if (validIDToken(token)) {
+            token.text = removePotentialBackticks(token.text ?? '');
```
mengweieric, line 227: will clicking a backtick removed suggestion add backtick back to the query?  Also why is it called potential backtick?

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -244,3 +244,101 @@ export const parseQuery = <
...
+      ignoredTokens,
+      rulesToVisit: modifiedRulesToVisit,
+      getParseTree,
+      enrichAutocompleteResult,
+      query,
+      cursor,
+      context,
+    });
+
+    // combine result and nextResult
```
mengweieric, line 299: nit. good to have comments to describe the intent and any potential edge cases as I feel that the for and switch statement handles various data types and use cases, but it might not be immediately clear why certain cases (e.g., boolean or undefined) are treated a specific way.

src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
```diff
@@ -65,18 +65,10 @@ export const getSuggestions = async ({
         ...PPL_AGGREGATE_FUNCTIONS.map((af) => ({
           text: `${af}()`,
           type: monaco.languages.CompletionItemKind.Function,
-          insertText: af + '(${1:expr}) ',
+          insertText: af + ' ',
```
mengweieric, line 68: why remove?

mengweieric (approved):

src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
paulstn: there were asks to remove it, since not many people knew how to use snippets (tab + keep tabbing to get to the next cursor place). this is primarily fueled by the version of monaco we're using, where snippets are a little buggy and also don't have the full functionality that current monaco has. 
we should reintroduce them when monaco is updated.

src/plugins/data/public/antlr/shared/utils.ts
paulstn: I'll add a big comment outlining the details on combining results

src/plugins/data/public/antlr/opensearch_sql/opensearch_sql_autocomplete.ts
paulstn: The formation of this list `sigTokens` is only used by this preferred rule on `RULE_predicate` to interpret the predicate as its formed. So everything interacting with backticks is just removing them so that if we want to use value completion on a field that's already in the query that happens to have backticks, it can match the field name to the index pattern's field. 
This is needed because value completion has default behavior of NOT sending an aggregation query if it can't match the field name to the current index. 
This change simply allows for values to be suggested if there are backticks in the field within the current predicate.
Regarding your question on if it will readd the backtick, it doesn't affect the field already in the query because this pref. rule doesn't suggest things that extend beyond the current 'word', so it will only replace the current token. 
It's called potential backtick because I'm running this method on every token that could be an ID or backticked-ID, so it's just some regex that looks to remove backticks at the front and back, and doesn't require any specific type of input that has backticks for sure.

joshuali925 (approved):

mengweieric (approved):

PR #9150, opened 2025-01-06
Create cypress command namespacing util

d-buckner:
### Description

This is to enable namespaced Cypress commands. This allows us to prevent unintentional Cypress command name collisions with 3rd party Commands. I've also gone ahead and migrated 2 utils which serve as examples and verifications.

Once we've merged, we can use this to create new cypress utils and decide on what priority (if any) we want for migrating old utils.

ciGroup11 is broken currently, and has a fix pending: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9209

Example usage
```js
// commands.js
initCommandNamespace(cy, 'osd');

cy.osd.add('getBody', () => cy.get('body')); // add command to namespace

// myTest.spec.js
it('should render body', () => {
    cy.osd.getBody().should('exist');
});
```

## Testing the changes
With the two migrated utils, tests should all pass. If the s3 dataset tests fail it's a good indication that there's an issue with this PR that needs to be resolved before merging.

## Changelog
- chore: create cypress command namespacing util

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


cypress/utils/command_namespace.js
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export default function initCommandNamespace(cy, namespace) {
```
angle943, line 6: I think a comment briefly explaining what this does can be helpful as the function itself is pretty abstract

d-rowe: +1 I'll add comments here to better explain this module. It's challenging to get meaningful TS types for free here since namespaces and their commands are dynamically defined at module resolution, but I can definitely take a look at making the definitions for the two commands I migrated over.

d-rowe: Added some comments and type definitions. Let me know what you think.

angle943 (approved):

kavilla (commented):
cool stuff! some comments. one general comment is that is there anything within this repo that we wouldn't want to namespace all the functions within this repo? so instead of we manually update each line to be cy.osd. we actually just end up decorating the cypress function name if the the command was stored in a specific file. that way we wont have to know which command is cy.osd or cy it is actually always cy. so cy.

cypress/utils/dashboards/remove_workspace.js
```diff
@@ -7,7 +7,7 @@ export function removeSampleDataAndWorkspace(url, workspaceName) {
   describe('removing workspace/sampledata', () => {
     it('remove workspace', () => {
```
kavilla, line 8: not specific to this pr but how come this a utils file if it's a test ?

cypress/utils/dashboards/remove_workspace.js
```diff
@@ -7,7 +7,7 @@ export function removeSampleDataAndWorkspace(url, workspaceName) {
...
       cy.visit(`${url}/app/workspace_list`);
-      cy.openWorkspaceDashboard(workspaceName);
+      cy.osd.openWorkspaceDashboard(workspaceName);
       cy.getElementByTestId('toggleNavButton').eq(0).should('exist').click();
       cy.wait(3000);
```
kavilla, line 12: not specifc to this pr but how come we opt for the cy.wait instead of using something like `timeout`

example:
```
cy.get("#form-name", {timeout: 10000}).type("Mike Johnson")
```

https://docs.cypress.io/app/core-concepts/retry-ability#Timeouts

so we could put an increased timeout on the next line. cy.getElementByTestId (we might have to modify it if it's not available for that function). 

like 

```suggestion
  cy.getElementByTestId('collapsibleNavAppLink-workspace_detail', { timeout: 10000 }).should('exist').click();
```

then instead of an ambiguous wait time which could get flaky it will just keep retrying for the element until it sees it for 10000 ms

d-rowe:
> cool stuff! some comments. one general comment is that is there anything within this repo that we wouldn't want to namespace all the functions within this repo? so instead of we manually update each line to be cy.osd. we actually just end up decorating the cypress function name if the the command was stored in a specific file. that way we wont have to know which command is cy.osd or cy it is actually always cy. so cy.

I don't feel we want to namespace everything, as I think many of these utils are useful for 3rd party environments testing as well. Things like getting elements by test subject, etc. The ones I feel we should focus on namespacing are utils that 3rd party environments may want different implementations or for utils we don't really want to be used by other environments.

cypress/utils/dashboards/remove_workspace.js
d-rowe: Yeah, I agree we shouldn't use arbitrary waits given these are flaky and introduce race conditions. I'm a little weary of addressing in this PR since I'm not confident that I'll be able to address easily with the confidence that I won't break anything given I have somewhat limited context in our integration tests.

cypress/utils/dashboards/remove_workspace.js
d-rowe: I believe because this can be re-used across a number of tests. For example, I write a new test and I want to remove a workspace as a part of this test and get solid logs of which parts failed or succeed as a part of that process.

cypress/integration/dashboard_sanity_test.spec.ts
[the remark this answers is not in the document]
d-rowe: Yup! We shouldn't have any blockers on that now.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/dataset_selector.spec.js
```diff
@@ -37,7 +37,7 @@ describe('dataset selector', { scrollBehavior: false }, () => {
     // Create workspace
     cy.deleteWorkspaceByName(`${workspace}`);
     cy.visit('/app/home');
-    cy.createInitialWorkspaceWithDataSource(`${DATASOURCE_NAME}`, `${workspace}`);
+    cy.osd.createInitialWorkspaceWithDataSource(`${DATASOURCE_NAME}`, `${workspace}`);
```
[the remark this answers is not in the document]
d-rowe: Yup, I've updated. Good catch.

sejli (approved):
LGTM, super helpful change. Think there are some parts of our testing, as mentioned by @kavilla's comments, that we should revisit that are not part of this PR's scope, but need to be improved upon.

ananzh (approved):
Approved as 2nd reviewer as major comments are resolved.

d-rowe:
ciGroup11 is broken currently with a fix pending: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9209

ananzh (approved):

LDrago27 (approved):

PR #9152, opened 2025-01-07
feat(vis_type_vega): support reading time field

ruanyl:
### Description

This PR introduces time range filter support to Vega visualizations with PPL queries. When `"%timefield%"` is specified, the time range filter from the dashboard will be applied to the PPL query automatically.

#### Before This PR
Vega visualizations with PPL queries did not respect the time range set on the dashboard. Regardless of the selected time range, the query remained unaffected.

#### After This PR
With the addition of `"%timefield%"`, changing the time range filter on a dashboard now impacts Vega visualizations with PPL queries.

#### Example
```diff
{
    "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
    "data": {
        "url": {
+            "%timefield%": "timestamp",
            "%type%": "ppl",
            "body": {
                "query": "source=opensearch_dashboards_sample_data_logs | stats count() as request_count by extension | sort -request_count"
            }
        }
    }
}
```

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9169

## Changelog

- feat: vega visualization with ppl now supports reading time field

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/vis_type_vega/public/data_model/ppl_parser.ts
+++ b/src/plugins/vis_type_vega/public/data_model/ppl_parser.ts
@@ -28,8 +28,8 @@ export class PPLQueryParser {
     if (this.timeCache._timeRange) {
       const [source, ...others] = query.split('|');
       const bounds = this.timeCache.getTimeBounds();
-      const from = moment(bounds.min).format('YYYY-MM-DD HH:mm:ss.SSS');
-      const to = moment(bounds.max).format('YYYY-MM-DD HH:mm:ss.SSS');
+      const from = moment.utc(bounds.min).format('YYYY-MM-DD HH:mm:ss.SSS');
+      const to = moment.utc(bounds.max).format('YYYY-MM-DD HH:mm:ss.SSS');
       const timeFilter = `where \`${timefield}\` >= '${from}' and \`${timefield}\` <= '${to}'`;
       if (others.length > 0) {
         return `${source.trim()} | ${timeFilter} | ${others.map((s) => s.trim()).join(' | ')}`;
--- a/src/plugins/vis_type_vega/public/data_model/ppl_parser.ts
+++ b/src/plugins/vis_type_vega/public/data_model/ppl_parser.ts
@@ -4,8 +4,13 @@
  */
 
 import { i18n } from '@osd/i18n';
+import moment from 'moment';
+
 import { Data, UrlObject, PPLQueryRequest } from './types';
 import { SearchAPI } from './search_api';
+import { TimeCache } from './time_cache';
+
+const TIMEFIELD = '%timefield%';
 
 const getRequestName = (request: PPLQueryRequest, index: number) =>
   request.dataObject.name ||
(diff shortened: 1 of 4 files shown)


src/plugins/vis_type_vega/public/data_model/ppl_parser.ts
```diff
@@ -34,6 +55,11 @@ export class PPLQueryParser {
       );
     }
 
+    if (timefield) {
+      const query = this.injectTimeFilter(url.body.query, timefield);
+      url.body.query = query;
```
d-rowe, line 60: Is this the url that vega uses to query the cluster? I wouldn't expect the url parser to format/mutate the url, but I'm new to this area so it may be a gap in my understanding.

ruanyl: > Is this the url that vega uses to query the cluster? 

Yes, `url.body.query` is the ppl query and will be send to opensearch. We modify it here to encapsulate the time filter from the UI, for example, the time range user selected on a dashboard. The behavior is the same as vega visualization with opensearch DSL which modified the query DSL to add time filter.

> I wouldn't expect the url parser to format/mutate the url

Could you please share your concern?

d-rowe: That makes sense. No concerns, my expectation was that the parser is a pure function but it looks like this is more of a parser and formatter of the original url if I'm understanding correctly.

d-rowe (approved):

src/plugins/vis_type_vega/public/data_model/ppl_parser.ts
ruanyl: I got your point now, I think we can also make it a pure function, for example we can deep copy the `url` object and then modify the deep copied object. I don’t notice much of a difference in this particular case since the existing parser mutate the url object as well, but I’d love to hear your thoughts on it :)

src/plugins/vis_type_vega/public/data_model/ppl_parser.ts
```diff
@@ -15,13 +20,29 @@ const getRequestName = (request: PPLQueryRequest, index: number) =>
...
+        return `${source.trim()} | ${timeFilter} | ${others.map((s) => s.trim()).join(' | ')}`;
+      }
+      return `${source.trim()} | ${timeFilter}`;
+    }
+    return query;
+  }
+
   parseUrl(dataObject: Data, url: UrlObject) {
+    const timefield = url[TIMEFIELD];
+    delete url[TIMEFIELD];
```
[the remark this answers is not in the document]
ruanyl: This is not the current browser url, it's the `url` property of vega-lite config object, the TIMEFIELD defines the time field name of current index, for example, `timestamp`, and it will used as the time filter of the query.

Hailong-am (approved):

SuZhou-Joe (approved):

PR #9157, opened 2025-01-08
Disable query assistant when dataset type is not supported

SuZhou-Joe:
### Description

Discover with query enhancement now supports varies of dataset: INDEX_PATTERN / INDEXES / S3, while the PPL query generated by query assistant are only supported by opensearch engine, so that we need to disable query assistant for S3 dataset for now.

### Before

image

### After

https://github.com/user-attachments/assets/2602ae7f-0d56-45e1-894a-78587f9e831c

## Changelog

- fix: Disable query assistant when dataset type is not supported

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/public/query_assist/utils/create_extension.tsx
+++ b/src/plugins/query_enhancements/public/query_assist/utils/create_extension.tsx
@@ -89,6 +89,8 @@ const getAvailableLanguages$ = (http: HttpSetup, data: DataPublicPluginSetup) =>
     })
   );
 
+const queryAssistantSupportedDatasetTypes = ['INDEXES', 'INDEX_PATTERN'];
+
 export const createQueryAssistExtension = (
   core: CoreSetup,
   data: DataPublicPluginSetup,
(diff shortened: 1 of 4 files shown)


src/plugins/query_enhancements/public/query_assist/utils/create_extension.tsx
```diff
@@ -119,8 +121,13 @@ export const createQueryAssistExtension = (
         };
       }
     },
-    isEnabled$: () =>
-      getAvailableLanguages$(http, data).pipe(map((languages) => languages.length > 0)),
+    isEnabled$: (dependencies) => {
+      const query = dependencies.query;
+      if (!queryAssistantSupportedDatasetTypes.includes(query.dataset?.type || '')) {
+        return new BehaviorSubject(false);
```
ruanyl, line 127: Nit: what about simply do:
```suggestion
        return of(false);
```

ruanyl (dismissed):

Hailong-am (approved):

src/plugins/query_enhancements/public/query_assist/utils/create_extension.tsx
SuZhou-Joe: Sure, updated.

src/plugins/query_enhancements/public/query_assist/utils/create_extension.tsx
```diff
@@ -93,6 +93,11 @@ const getAvailableLanguages$ = (http: HttpSetup, data: DataPublicPluginSetup) =>
     })
   );
 
+const queryAssistantSupportedDatasetTypes = [
+  DEFAULT_DATA.SET_TYPES.INDEX,
+  DEFAULT_DATA.SET_TYPES.INDEX_PATTERN,
```
kavilla, line 98: thanks for doing this but should this be apart of the switch map on line 87

SuZhou-Joe: Sure, updated.

Issue #8976, opened 2024-11-27
[Saved Queries] TESTID-124: View list of Saved query 


LDrago27:
<h2 data-pm-slice="1 1 []">Objective</h2><p>This aims to test if on Navigating to Discover we can view all the Saved Queries when user clicks on Open Saved Query button.</p><h2>What to test</h2><p>Permutations to consider (only if they apply)</p><p>savedQueriesNewUI.enabled :true/false</p>

| Scenario | Steps | Expected result |
|----------|-------|-----------------|
| Shows all saved queries when users click on Open Saved Query Button | <ul><li>Navigate to Discover</li><li>Click on Open saved Query</li></ul> | Verify all the saved Queries are shown. |

<h2><strong>Pre-requisites</strong></h2><p>What do we need to make sure the test can be written</p><p><em>Steps to setting up the scenario</em></p><ol class="ProsemirrorEditor-list"><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="numbered" style="margin-left: 32px"><p>Have at least 1 existing saved query</p></li></ol><h2><strong>Known Issues</strong></h2><p><em>List of known issues to reference that could impact the results of this test case</em></p>

https://github.com/user-attachments/assets/86b163e3-3951-4faf-9560-7f32c5e4e7ac

ArgusLi:
Could I please be assigned this task?

ArgusLi:
Could you please move this into ready to review?

PR #9166, opened 2025-01-09
[Discover 2.0 Testing] Create View List of Saved Queries Test (Test-Id 124: View list of Saved query) 

ArgusLi:
### Description

Adds a Cypress test suite for filtering in the Discover page.

### Issues Resolved

Closes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8976.

## Screenshot

https://github.com/user-attachments/assets/8b5be4bd-99fc-47f0-a917-9c9b84aa2fbd

https://github.com/user-attachments/assets/2ccd98a0-6acf-4035-a92b-5ef0fc249a5c

## Testing the changes

With OSD running, run `yarn run cypress open`. In E2E specs, you will see 2 new test specs `saved_queries.spec.js` and `saved_queries_old_ui.spec.js`. Run `saved_queries.spec.js` if `data.savedQueriesNewUI.enabled: true` in the OSD config, else run `saved_queries_old_ui.spec.js`.

## Changelog

- test: Add cypress integration test for the old and new UI view saved queries.


cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/field_display_filtering.spec.js
```diff
@@ -223,7 +223,7 @@ describe('filter for value spec', () => {
   });
 
   after(() => {
-    cy.deleteWorkspaceByName(`${WORKSPACE_NAME}`);
+    cy.deleteWorkspaceByName(`${workspace}`);
     cy.deleteDataSourceByName(`${DATASOURCE_NAME}`);
```
angle943, line 227: if we delete workspace, do we need to delete datasource?

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/saved_queries_old_ui.spec.js
```diff
@@ -0,0 +1,105 @@
...
+        cy.saveQueryOldUI(config.saveName);
+      });
+    });
+    it('should see all saved queries', () => {
+      cy.getElementByTestId('saved-query-management-popover-button').click({
+        force: true,
+      });
+
+      //Wait for queries to load
+      cy.wait(2000);
```
angle943, line 94: same comment as above (instead of waiting arbitrary, can we wait for some element to load)

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/field_display_filtering.spec.js
LDrago27: Yup. We need to delete them both. Since a data source can exist outside a workspace.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/utils/saved_queries.js
```diff
@@ -0,0 +1,207 @@
...
+    case QueryLanguages.Lucene.name:
+      return 'bytes_transferred: {9950 TO *}';
+    case QueryLanguages.SQL.name:
+      return `SELECT * FROM ${dataset} WHERE bytes_transferred > 9950`;
+    case QueryLanguages.PPL.name:
+      return `source = ${dataset} | where bytes_transferred > 9950`;
+    default:
+      throw new Error(`getQueryString encountered unsupported language: ${language}`);
+  }
+};
```
angle943, line 52: (I can do this)

But it looks like this and some other functions are exactly the same as mine. I will make some refactor in the future to put them in shared.

cypress/utils/apps/data_explorer/commands.js
```diff
@@ -125,33 +125,49 @@ Cypress.Commands.add(
   }
 );
 
-Cypress.Commands.add('saveQuery', (name, description) => {
+Cypress.Commands.add('saveQueryOldUI', (name, description = ' ') => {
```
angle943, line 128: can we rename this to something else? I dont' have enough context, but what constitutes a 'new ui' vs old ui? I don't like `OldUi` because in a few months/years, what we consider now to be `newUi` will become `oldUi`

cypress/utils/apps/data_explorer/commands.js
```diff
@@ -125,33 +125,49 @@ Cypress.Commands.add(
...
   cy.whenTestIdNotFound('saved-query-management-popover', () => {
     cy.getElementByTestId('saved-query-management-popover-button').click();
   });
   cy.getElementByTestId('saved-query-management-save-button').click();
 
   cy.getElementByTestId('saveQueryFormTitle').type(name);
   cy.getElementByTestId('saveQueryFormDescription').type(description);
+
+  cy.getElementByTestId('savedQueryFormSaveButton').click({ force: true });
```
angle943, line 137: why is `force` needed? Typically, this is only needed if the element is hidden by some other element in the UI (like a popover). if so, then it is valid that it is needed but a comment should be there signifying why it's needed. Else we can just get rid of `force: true`

cypress/utils/apps/data_explorer/commands.js
```diff
@@ -125,33 +125,49 @@ Cypress.Commands.add(
...
+Cypress.Commands.add('saveQuery', (name, description = ' ') => {
+  cy.whenTestIdNotFound('saved-query-management-popover', () => {
+    cy.getElementByTestId('saved-query-management-popover-button').click();
+  });
+  cy.getElementByTestId('saved-query-management-save-button').click();
+
+  cy.getElementByTestId('saveQueryFormTitle').type(name);
+  cy.getElementByTestId('saveQueryFormDescription').type(description);
+
+  cy.getElementByTestId('savedQueryFormSaveButton').click({ force: true });
```
angle943, line 150: same `forc`

cypress/utils/apps/query_enhancements/commands.js
```diff
@@ -100,12 +100,18 @@ Cypress.Commands.add('deleteDataSourceByName', (dataSourceName) => {
...
+  cy.get('a').contains(dataSourceName).click({ force: true });
+
+  // Wait til the page is loaded (waitForLoader and just getting an element doesn't work)
+  cy.wait(2000);
 
   // Delete the dataSource connection
-  cy.getElementByTestId('editDatasourceDeleteIcon').click();
-  cy.getElementByTestId('confirmModalConfirmButton').click();
+  cy.getElementByTestId('editDatasourceDeleteIcon').click({ force: true });
+  cy.getElementByTestId('confirmModalConfirmButton').click({ force: true });
```
angle943, line 114: if we remove `deleteDataSourceByName` from the other places, these changes don't need to be here. But IF i'm wrong and `deleteDataSourceByName` needs to be there, two things:

- lets get rid of `cy.wait(2000)` and replace it with waiting for some element to exist.
- Lets get rid of `force: true` unless its really needed (and if so, add a comment why)

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/saved_queries.spec.js
```diff
@@ -0,0 +1,108 @@
...
+        setDatePickerDatesAndSearchIfRelevant(config.language);
+
+        setSearchConfigurations(config);
+        verifyDiscoverPageState(config);
+        cy.saveQuery(config.saveName);
+      });
+    });
+    it('should see all saved queries', () => {
+      cy.getElementByTestId('saved-query-management-popover-button').click({
+        force: true,
```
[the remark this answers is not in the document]
ArgusLi: I had some issues where sometimes just clicking wouldn't open the popover. I will add a comment here to explain.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/utils/saved_queries.js
ArgusLi: I've separated this function because I will be needing to make some changes to it and its associated functions for the specific queries in following saved queries tests.

cypress/utils/apps/data_explorer/commands.js
ArgusLi: The flag is `data.savedQueriesNewUI.enabled: false`, which is why I named it as such. I agree though. Would love to discuss this offline with @LDrago27.

cypress/utils/apps/data_explorer/commands.js
ArgusLi: I think this is actually left over from some testing. Will reverify and check.

cypress/utils/apps/data_explorer/commands.js
ArgusLi: I think this is actually left over from some testing. Will reverify and check.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/saved_queries.spec.js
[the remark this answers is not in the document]
angle943: if it was not working but the element not obstructed, i wonder if this works:

`cy.getElementByTestId('saved-query-management-popover-button').should('be.visible').click()`

(This will ensure it is visible in the DOM before clicking. i wonder if you are attempting to click it before it is visible)

cypress/utils/apps/query_enhancements/commands.js
ArgusLi: Let me try again. The reason why this was here is because of some really weird behaviour from Cypress. The `deleteSourceByName` was initially in the `afterEach` block, but I changed it to the `after` block instead. When it was in the `afterEach` block, when the first saved query was created, it executed perfectly, but would fail the next time it was run after the next saved query was created. 

I tried getting the element without a chained click, which would lead to Cypress' timeout and retry logic, but for some reason it still failed. I tried using waitForLoader but that didn't work either, which is when I shifted to the wait. But after that, it still failed when clicking the `editDatasourceDeleteIcon` button - the button was selected, and was clicked, but no action actually happened. Which is why I had to use `force: true`. The same thing happened for `confirmModalConfirmButton`. 

But none of these will happen now given it is only run once in the `after block` so I'll try and see if removing them will work now.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/saved_queries.spec.js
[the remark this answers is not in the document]
ArgusLi: I tried again without the force and it's working now. I think this one leftover from when I was trying to debug an issue.

cypress/utils/apps/data_explorer/commands.js
ArgusLi: Confirmed. Removing.

cypress/utils/apps/data_explorer/commands.js
ArgusLi: Confirmed, removing.

cypress/utils/apps/query_enhancements/commands.js
ArgusLi: Yup, because it's only run once now, it works without these forces and waits.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/saved_queries.spec.js
```diff
@@ -0,0 +1,108 @@
...
+      });
+
+      cy.getElementByTestId('saved-query-management-open-button').click();
+
+      //Wait for queries to load
+      cy.wait(2000);
```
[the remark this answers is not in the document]
ArgusLi: This was left over from when I copied your structure for saved search. It doesn't seem needed for saved queries. Removing.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/saved_queries_old_ui.spec.js
ArgusLi: This was left over from when I copied your structure for saved search. It doesn't seem needed for saved queries. Removing.

cypress/utils/apps/query_enhancements/commands.js
```diff
@@ -101,7 +101,7 @@ Cypress.Commands.add('deleteDataSourceByName', (dataSourceName) => {
   cy.visit('app/dataSources');
 
   // Find the anchor text corresponding to specified dataSource
-  cy.get('a').contains(dataSourceName).click();
+  cy.get('a').contains(dataSourceName).click({ force: true });
```
angle943, line 104: sorry one more force

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/saved_queries_old_ui.spec.js
```diff
@@ -0,0 +1,105 @@
+/*
```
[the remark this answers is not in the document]
angle943: i think we split up the running of tests with different configs based on folder structure, so it might be problematic for this to be in the same folder as the new-ui. but again i'll defer to @LDrago27 and @ananzh

cypress/utils/apps/query_enhancements/commands.js
angle943: did you push that change? i don't see it here :D

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/utils/saved_queries.js
LDrago27: @ArgusLi  Lets move all the utils inside to this path `/cypress/utils/apps/query_enhancements`. That will help us to consolidate the utils method in one place.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/utils/saved_queries.js
```diff
@@ -0,0 +1,207 @@
...
+          throw new Error(
+            `getSampleTableData encountered unsupported language for ${datasetType}: ${language}`
+          );
+      }
+    default:
+      throw new Error(`getSampleTableData encountered unsupported datasetType: ${datasetType}`);
+  }
+};
+
+/**
```
LDrago27, line 140: nit: Update the comments to indicate it is for `saved queries` instead of `saved search`

cypress/utils/apps/query_enhancements/commands.js
ArgusLi: I did something worse than that. I got confused with the deleteAllDataSources command and removed the force from that instead 🤦. I've retested, restored the old force and removed the one here.

cypress/utils/apps/data_explorer/commands.js
LDrago27: I agree the oldUi and newUi are n't the best descriptive definitions for the tests. How about we use savedQueryFlyout for savedQueriesNewUI and savedQueryPopover for the savedQueriesOldUi.

LDrago27, line 128: nit: Can we combine the `saveQueryOldUi` and `saveQuery` command into a single command where we can have a flag that determines which version to use.

ananzh:
ciGroup 10 failure is a known issue and @angle943  is working on it

ananzh (approved):

LDrago27 (approved):

PR #9167, opened 2025-01-09
Cleanup OsdUrlStateStorage subscription in TopNav

Maosaic:
### Description

`OsdUrlStateStorage` subscription in `TopNav` is not properly cleaned up, which is causing `ScopedHistory` throwing error and potential memory leak.

### Issues Resolved

-fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5476

## Screenshot

Before

https://github.com/user-attachments/assets/c79a6cba-4c52-45c8-8664-168923d1308a

After

https://github.com/user-attachments/assets/272f88a5-7302-4271-bff3-cc70bb75578d

## Changelog

- fix: Cleanup OsdUrlStateStorage subscription in TopNav

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/9151.yml
+++ b/changelogs/fragments/9151.yml
@@ -1,0 +1,2 @@
+test:
+- Add tests for updating a saved search ([#9151](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9151))
--- a/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/saved_search.spec.js
+++ b/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/saved_search.spec.js
@@ -19,6 +19,7 @@ import {
   verifyDiscoverPageState,
   verifySavedSearchInAssetsPage,
   postRequestSaveSearch,
+  updateSavedSearchAndSaveAndVerify,
 } from './utils/saved_search';
 
 export const runSavedSearchTests = () => {
@@ -98,6 +99,7 @@ export const runSavedSearchTests = () => {
           if (startingLanguage !== config.language) return;
 
           it(`should successfully load a saved search for ${config.testName} starting from ${startingLanguage}`, () => {
+            // using a POST request to create a saved search to load
             postRequestSaveSearch(config);
 
             cy.navigateToWorkSpaceSpecificPage({
@@ -120,6 +122,18 @@ export const runSavedSearchTests = () => {
             verifyDiscoverPageState(config);
           });
         });
+
+      it(`should successfully update a saved search for ${config.testName}`, () => {
+        // using a POST request to create a saved search to load
+        postRequestSaveSearch(config);
+        updateSavedSearchAndSaveAndVerify(config, false);
+      });
+
+      it(`should successfully save a saved search as a new saved search for ${config.testName}`, () => {
+        // using a POST request to create a saved search to load
+        postRequestSaveSearch(config);
+        updateSavedSearchAndSaveAndVerify(config, true);
+      });
     });
   });
 };
--- a/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/utils/saved_search.js
+++ b/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/query_enhancements/utils/saved_search.js
@@ -249,11 +249,9 @@ export const generateAllTestConfigurations = () => {
  * @param {QueryEnhancementLanguage} language - query language
  */
 export const setDatePickerDatesAndSearchIfRelevant = (language) => {
-  if (language === QueryLanguages.SQL.name) {
-    return;
+  if (language !== QueryLanguages.SQL.name) {
+    cy.setTopNavDate(START_TIME, END_TIME);
   }
-
-  cy.setTopNavDate(START_TIME, END_TIME);
 };
 
 /**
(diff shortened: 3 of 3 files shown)


src/plugins/data/public/query/state_sync/connect_to_query_state.ts
```diff
@@ -49,7 +49,7 @@ import { validateTimeRange } from '../timefilter';
  * @param  OsdUrlStateStorage to use for syncing and store data
  * @param syncConfig app filter and query
  */
-export const connectStorageToQueryState = async (
+export const connectStorageToQueryState = (
```
angle943, line 52: will this function ever return anything other than the unsub callback function, besides it reaching the `catch` block? Several questions regarding that:

- Based on the call pattern that you provided on the usage of the returning callback function, can't we just unsub from the end of this function instead of wrapping it with a `useEffect` by the caller?
- If some `subs` were made and then getting to the `catch` block for some error, don't we want to unsub from the `catch` as well?

Maosaic: I think the intended behaviour hear is to maintain the subscription and let the scope where invokes `connectStorageToQueryState` decide the life cycle of the subscription.

It make senses to me to unsub in the `catch` block, but currently the subscription is building at the end of the `try `block, it's unlikely to have any lingering subscription if it reach `catch` block.

src/plugins/discover/public/application/view_components/canvas/top_nav.tsx
```diff
@@ -65,7 +65,7 @@ export const TopNav = ({ opts, showSaveQuery, isEnhancementsEnabled }: TopNavPro
     ? getTopNavLinks(services, inspectorAdapters, savedSearch, isEnhancementsEnabled)
     : [];
 
-  connectStorageToQueryState(
+  const unSub = connectStorageToQueryState(
```
d-rowe, line 68: Change LGTM. One thing (I know it's outside of the scope of this PR) I wanted to discuss is that we're relying on folk's best intention and understanding of `connectStorageToQuery` to unsubscribe and avoid these memory leaks.

It may be nice to introduce a hook or HOC wrapper, which subscribes on mount and unsubscribes on unmount to abstract the subscription from the consumer.

Maosaic: +1 Good idea, any suggestion which folder to put this hook in?

d-rowe: Thanks! How about https://github.com/opensearch-project/OpenSearch-Dashboards/tree/32b0602cac0b0a5879d56b48e7d8587362e3b5eb/src/plugins/discover/public/application/view_components/utils?

SuZhou-Joe: Agree with that, should we move the `connectStorageToQueryState` into a useEffect? and we can unsubscribe the change inside that useEffect hook as well.

kavilla:
do you mind signing the commit on push?

kavilla:
do you have the changelog bot setup i think we can have a changelog for this

src/plugins/discover/public/application/view_components/canvas/top_nav.tsx
Maosaic: I have created the hook under same folder of `plugins/data`, since it's been using from `src/plugins/discover/...` and `src/plugins/vis_builder/...`

src/plugins/data/public/query/state_sync/connect_to_query_state.ts
```diff
@@ -41,27 +40,28 @@ import { QueryState, QueryStateChange } from './types';
 import { FilterStateStore, COMPARE_ALL_OPTIONS, compareFilters } from '../../../common';
 import { validateTimeRange } from '../timefilter';
 
+export interface ISyncConfig {
+  filters: FilterStateStore;
+  query: boolean;
+  dataSet?: boolean;
```
kavilla, line 46:
```suggestion
  dataset?: boolean;
```

kavilla: quick fix up in an old typo while we are in a neighbor

kavilla: hmm actuall when i think about it we no dont have a dataset outside of the query object so i think actually we could remove this field in the interface. 

wdyt? do you see it being used?

src/plugins/data/public/query/state_sync/use_connect_to_query_state.ts
kavilla, line 1: question:

do you think we should put this in a lib folder within the state_sync folder or would that be just too messy? i see some folders have it but i dont think there is a consistency within the project.

src/plugins/vis_builder/public/application/components/top_nav.tsx
```diff
@@ -56,16 +56,17 @@ export const TopNav = () => {
...
-    filters: opensearchFilters.FilterStateStore.APP_STATE,
-    query: true,
-  });
   const { selected: indexPattern } = useIndexPatterns();
   const [config, setConfig] = useState<TopNavMenuData[] | undefined>();
   const originatingApp = useTypedSelector((state) => {
     return state.metadata.originatingApp;
   });
 
+  useConnectStorageToQueryState(services.data.query, services.osdUrlStateStorage, {
```
kavilla, line 65: without diving too deep, do we know if the query state is changed and navigate to visbuilder if this gets set to after the the index patterns does it cause a difference in the selected: indexPattern?

src/plugins/discover/public/application/view_components/canvas/top_nav.tsx
```diff
@@ -65,15 +65,10 @@ export const TopNav = ({ opts, showSaveQuery, isEnhancementsEnabled }: TopNavPro
...
-  connectStorageToQueryState(
-    services.data.query,
-    osdUrlStateStorage,
-    {
-      filters: opensearchFilters.FilterStateStore.APP_STATE,
-      query: true,
-    },
-    uiSettings
-  );
+  useConnectStorageToQueryState(services.data.query, osdUrlStateStorage, {
```
wanglam, line 68: Since `syncConfig` is a dependency in  `useConnectStorageToQueryState`, this object will regenerate every render. They will unsubscribe  and subscribe again every render. Is this by design? Shall we need to add a useMemo depend on config value directly?

Issue #9143, opened 2025-01-02
[BUG] VIsuals do not resize after hiding legend


d-rowe:
https://github.com/user-attachments/assets/7766f8e8-3329-47de-8a21-bba5c29c8cb6

When hiding legends in visuals, the visuals are not resizing to the full size of their container. They require a dashboard level refresh to fill the gap left by the legend.

**To Reproduce**
1. Create a visual with a legend (right legend in my repro)
2. Click on the legend toggle at the bottom left of the visual
3. Notice empty space where the legend was prior to hiding.

**Expected behavior**
I'd expect hiding the legend to trigger a resize of the chart to fill the available space.

**Screenshots**
https://github.com/user-attachments/assets/7766f8e8-3329-47de-8a21-bba5c29c8cb6

**Host/Environment (please complete the following information):**
 - OS: MacOS 14.7.1
 - Browser and version: Chrome 131

wanglam:
Hi @d-rowe , I raised a [PR](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9170) to address this issue. Could you help me take a look and test in your local environment if this issue still exists?

d-rowe:
Thanks @wanglam. Taking a look.

krisfreedain:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain), [2](https://github.com/andrross/), [3](https://github.com/reta)

PR #9170, opened 2025-01-10
Fix visualization chart not expanded after legend toggle

wanglam:
### Description

This PR fixes an issue where the chart was not expanding after hiding the legend. The root cause was that the `VisLegend` component was using `vis.getUiState()` as the `uiState` in the [vis_controller](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/vis_type_vislib/public/vis_controller.tsx), which was different from the `uiState` in the `VisualizationChart` component. According to the code in [visualization_chart](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/visualizations/public/components/visualization_chart.tsx#L119), it was using `this.props.uiState`. Therefore, changes to `vis.legendOpen` were not triggering a re-render in `VisualizationChart`.

To resolve this issue, I added a `uiState` synchronization logic in `src/plugins/visualizations/public/expressions/visualization_renderer.tsx`. This logic compares the `visState` in `vis` and calls `setState` to ensure that the two `uiState` instances are the same. As a result, the chart will be resized correctly.

### Issues Resolved

#9143 

## Testing the changes

1. Import "Sample web logs" sample data
2. Create a new dashboard with "(Line) Avg bytes over time"
3. Update a close time range like "15 seconds ago", the chart will become no results like screenshot below
image
4. Update time range to very early like "15 years ago"
5. Click "Toggle legend" on the bottom left, the chart should expanded after legend hide
image

## Changelog

- fix: Visualization chart not expanded after legend toggle

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/visualizations/public/expressions/visualization_renderer.tsx
+++ b/src/plugins/visualizations/public/expressions/visualization_renderer.tsx
@@ -57,6 +57,10 @@ export const visualization = (): ExpressionRenderDefinition<VisRenderValue> => (
       unmountComponentAtNode(domNode);
     });
 
+    if (vis.getUiState() !== uiState) {
+      vis.setUiState(uiState);
+    }
+
     const listenOnChange = params ? params.listenOnChange : false;
     render(
       <Visualization
--- a/changelogs/fragments/9170.yml
+++ b/changelogs/fragments/9170.yml
@@ -1,0 +1,2 @@
+fix:
+- Visualization chart not expanded after legend toggle ([#9170](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9170))
--- a/src/plugins/visualizations/public/expressions/visualization_renderer.test.tsx
+++ b/src/plugins/visualizations/public/expressions/visualization_renderer.test.tsx
@@ -1,0 +1,139 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from 'react';
+import { act } from 'react-dom/test-utils';
+import { unmountComponentAtNode } from 'react-dom';
+import { setTypes } from '../services';
+import { TypesService } from '../vis_types';
+import { PersistedState } from '../persisted_state';
+import { ExprVis } from './vis';
+import { visualization } from './visualization_renderer';
+
+jest.mock('../components', () => ({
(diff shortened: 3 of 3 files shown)


d-buckner (dismissed):

raintygao (dismissed):

ashwin-pc (approved):
Nice! Thanks :)

src/plugins/visualizations/public/expressions/visualization_renderer.tsx
```diff
@@ -57,6 +57,10 @@ export const visualization = (): ExpressionRenderDefinition<VisRenderValue> => (
       unmountComponentAtNode(domNode);
     });
 
+    if (vis.getUiState() !== uiState) {
+      vis.setUiState(uiState);
```
AMoo-Miki, line 61: I am concerned that `handlers.uiState` has a type of `any` while `vis.setUiState` is expecting `PersistedState`.

wanglam: You're right, but I think it's not a big issue. According the implementation in the [Visualization component](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/visualizations/public/components/visualization.tsx#L63), the `uiState` has already been assigned to the vis object. The `vis.uiState` always be an empty since `vis` object recreated in every render. To address this `any` concern, how about add a type assertment in the compare condition. Refactor code like below:
```ts
if (uiState instanceof PersistedState && vis.getUiState() !== uiState) {
      vis.setUiState(uiState);
}
```

SuZhou-Joe: Nit: Should we do a deep compare?

SuZhou-Joe (approved):

src/plugins/visualizations/public/expressions/visualization_renderer.tsx
wanglam: `vis.getUiState()` should always be an empty object in  this place. Since we construct a new `ExpVis` object here. The `uiState` of this object should be empty too.

angle943:
This is Kiro posting.

Approved and mergeable. Discover/Visualize code owners — is this ready to merge?

Issue #8950, opened 2024-11-27
[Search Management (Managing saved searches)] TESTID-114: Update an existing Saved Search 


LDrago27:
<h2 data-pm-slice="1 1 []">Objective</h2><p>Update an existing Saved Search. </p><h2>What to test</h2><p>Permutations to consider (only if they apply)</p><p>Data type:</p><ol class="ProsemirrorEditor-list"><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>Index Patterns</p></li><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>Non Index pattern </p></li></ol><p>Language:</p><ol class="ProsemirrorEditor-list"><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>DQL</p></li><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>Lucene</p></li><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>PPL</p></li><li class="ProsemirrorEditor-listItem" data-list-indent="1" data-list-type="bulleted" style="margin-left: 32px"><p>SQL</p></li></ol><p>Query:</p>

| Scenario                                | Steps                                                                                                                                                                                                                         | Expected Result                                                        |
|-----------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------|
| **Updating an existing saved search**   | <ul><li>Navigate to Discover.</li><li>Click on "Open" and select an existing Saved Search.</li><li>Update the state by modifying any combination of Data Type, Language, and Query.</li><li>Click "Save" and uncheck the option to save it as a new Saved Search.</li><li>Click on the "New" icon in Discover to refresh the page.</li><li>Click on "Open" and load the updated Saved Search.</li></ul> | The Discover state should match that of the updated Saved Search.     |
[...]

PR #9200, opened 2025-01-17
[Fix] Saved Search Integ Test Update: update saving a new saved search test spec

angle943:
### Description

[This test case under Saved Search](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/saved_search.spec.js#L132) have been flaky and failing more often than not for the permutation: `DatasetType: INDEX_PATTERN, language: OpenSearch SQL`. This is not reproducible for me locally, however looking at the videos, the issue seem to appear that the query language is changing to SQL automatically for some reason instead of staying as PPL. I am adding a step in the test to wait for the page to be loaded before loading.

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/saved_search.spec.js
+++ b/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/saved_search.spec.js
@@ -117,7 +117,7 @@ export const runSavedSearchTests = () => {
             cy.setIndexPatternAsDataset(INDEX_PATTERN_WITH_TIME, datasourceName);
 
             cy.setQueryLanguage(startingLanguage);
-            cy.loadSaveSearch(config.saveName, false);
+            cy.loadSaveSearch(config.saveName);
             setDatePickerDatesAndSearchIfRelevant(config.language);
             verifyDiscoverPageState(config);
           });
--- a/cypress/utils/apps/data_explorer/commands.js
+++ b/cypress/utils/apps/data_explorer/commands.js
@@ -30,31 +30,19 @@ Cypress.Commands.add('saveSearch', (name, saveAsNew = false) => {
 
   cy.getElementByTestId('confirmSaveSavedObjectButton').click({ force: true });
 
-  // if saving as new save search, you need to click confirm twice;
-  if (saveAsNew) {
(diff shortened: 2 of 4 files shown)


ruanyl (dismissed):

Hailong-am:
@angle943  `saved_search.spec.js` still failing,  can you double confirm?

angle943:
Yes I am investigating now. Fortunately the original flaky test that this PR attempted to fix has passed: `should successfully save a saved search as a new saved search for OpenSearch SQL-INDEX_PATTERN`

Screenshot 2025-01-16 at 6 32 10 PM

The failiure that we are seeing now is a different test, and its a flakiness I've seen before, albeit a lot less common

ananzh (dismissed):

angle943:
After retrying, the original flaky test is failing again. I am going to try a bit of a more drastic change

angle943:
After my latest change, the past two runs were successful for the affected ciGroup10. Although we will need to run it several times to be more clear, I am having more confidence that my changes have made the test less flaky, and I think I found the root issue.

For context, the requirements on the test that was failing is [on this ticket](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8950) - "Save as a new saved search":

- Open an existing Saved Search
- Update the state by modifying any combination of data type, language, and query
- Click "Save" and check the option to save it as a new Saved Search
- Click on "New" Icon in discover to refresh page
- click on "Open" and load the new saved search.

Due to the ambiguity of the steps, I was saving the newly saved search with the same name as the previously saved search. This results two saved searches with the same name.

Screenshot 2025-01-17 at 5 18 15 AM

In order to load the "correct" saved search, I had logic that selects all elements with that same name, and by selecting `last()` one.

I believe the reason that the test was flaky was because either:

- For some reason on headless, the last element was not the newly saved search, but rather the old one
- For some reason on headless, using the `last().click()` did not select and click the last one, but rather the first one

In order to resolve this, I looked back at the test requirements on the ticket and realized there were no explicit requirements to make the newly saved search with the same name. So I am now changing it to a newly saved name, thus we no longer have to worry about possibly selecting the old saved search.

ananzh (approved):

Issue #9184, opened 2025-01-15
[enhancement] Support path alias


ruanyl:
### Problem
Currently, the path aliases `opensearch-dashboards/public` and `opensearch-dashboards/server` are defined in the tsconfig and are widely used throughout the codebase. However, these aliases are only functional for TypeScript types. They are not configured for Webpack or Babel, leading to failures during development when running dev server or production builds failed when build for release when these imports contain actual modules rather than just types.

### Expectation
The path alias should work for module imports not just only TS types, so I don't have to use relative path when importing modules like `import {...} from '../../../../src/core/public'`. Instead, I can simply type `import {...} from 'opensearch-dashboards/public'`.

### What to do
- [ ] Add path alias support for webpack to ensure browser side code can be compiled successfully during development and build
- [ ] Add path alias support for babel to ensure server side code can be transpiled  successfully
- [ ] Ensure jest unit tests also works with path alias

PR #9205, opened 2025-01-17
Support path alias for webpack and babel

ruanyl:
### Description
This PR add path alias support for webpack and babel. Previously, path alias only works for importing typescript types.

1. webpack now supports `opensearch-dashboards/public` path alias
2. Refactor the current in-house path resolution in BundleRefsPlugin to use webpack internal `compiler.resolverFactory` to work with webpack path alias
3. Introduced `babel-plugin-module-resolver` to enable path alias support for babel so that we can use path alias in server side code and jest tests.

#### Examples:
For `public` code
```
// before
import { formatUrlWithWorkspaceId } from '../../../../../src/core/public/utils';
// after
import { formatUrlWithWorkspaceId } from 'opensearch-dashboards/public/utils';
```
For `server` code
```
// before
import { fromRoot } from '../../../core/server/utils';
// after
import { fromRoot } from 'opensearch-dashboards/server/utils';
```

For jest tests:
```
// before
import { applicationServiceMock } from '../../../src/core/public/mocks';
// after
import { applicationServiceMock } from 'opensearch-dashboards/public/mocks'
```

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9184

## Changelog

- infra: add path alias support for webpack and babel

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/packages/osd-optimizer/src/worker/bundle_refs_plugin.ts
+++ b/packages/osd-optimizer/src/worker/bundle_refs_plugin.ts
@@ -20,7 +20,6 @@
  */
 
 import Path from 'path';
-import Fs from 'fs';
 
 import webpack from 'webpack';
 
@@ -28,20 +27,6 @@ import { Bundle, BundleRefs, BundleRef } from '../common';
 import { BundleRefModule } from './bundle_ref_module';
 
 const RESOLVE_EXTENSIONS = ['.js', '.ts', '.tsx'];
-
-function safeStat(path: string): Promise<Fs.Stats | undefined> {
-  return new Promise((resolve, reject) => {
-    Fs.stat(path, (error, stat) => {
-      if (error?.code === 'ENOENT') {
-        resolve(undefined);
-      } else if (error) {
-        reject(error);
-      } else {
-        resolve(stat);
-      }
-    });
-  });
-}
 
 interface RequestData {
   context: string;
@@ -80,7 +65,7 @@ export class BundleRefsPlugin {
           const context = data.context;
           const dep = data.dependencies[0];
 
-          this.maybeReplaceImport(context, dep.request).then(
+          this.maybeReplaceImport(context, dep.request, compiler).then(
             (module) => {
               if (!module) {
                 wrappedFactory(data, callback);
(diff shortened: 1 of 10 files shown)


Hailong-am:
can we have some existing code updated to new format? so that we can verify it works during build and also take it as example.

d-buckner:
Really nice QoL improvement!

packages/osd-optimizer/src/worker/bundle_refs_plugin.ts
```diff
@@ -134,64 +119,16 @@ export class BundleRefsPlugin {
...
-        const indexPath = Path.resolve(absoluteRequest, `index${ext}`);
-        const indexStats = await safeStat(indexPath);
-        if (indexStats?.isFile()) {
-          return indexPath;
+  private async resolve(request: string, context: string, compiler: webpack.Compiler) {
+    const resolver = compiler.resolverFactory.get('normal');
+    return new Promise<string | undefined>((resolve) => {
+      resolver.resolve({}, context, request, {}, (err: Error, resolvedPath: string) => {
+        if (err) {
+          resolve(undefined);
```
d-buckner, line 127: Do we want to silently fail here for all resolver errors? I'm not sure how this is handled upstream, but it may simplify build issue debugging if we reject with error so we don't swallow it.

ruanyl: The `resolve` function is called within `function maybeReplaceImport()`, even it silently fail here, webpack will still throw error when it resolve the path later. It won't swallow  the error. But I think you have a good point, we can reject here directly, because if the path cannot be resolved here, it cannot be resolved later by webpack either.

I've updated the code, please take a look again, thanks!

ruanyl:
> can we have some existing code updated to new format? so that we can verify it works during build and also take it as example.

Yes, code updated with an example.

d-buckner (approved):
Thanks for adding the error bubbling. LGTM

Hailong-am (approved):

SuZhou-Joe (dismissed):

wanglam:
@ruanyl There were some conflicts in the `yarn.lock`. Could you help to fix it?

Hailong-am (dismissed):

ruanyl:
@wanglam conflicts resolved, could you take another look please?

wanglam (dismissed):

d-buckner (dismissed):

Hailong-am (approved):

wanglam (approved):

PR #9208, opened 2025-01-17
refactor query enhancement tests

angle943:
### Description

Refactor various tests within the query enhancement
- Make a `shared.js` utils for query enhancement. Move all functions that can be used by all query enhancement tests to here, such as `getRandomizedWorkspaceName`, `generateAllTestConfigurations`, and `setDatePickerDatesAndSearchIfRelevant`
    - Update all tests to use the `getRandomizedWorkspaceName` instead of each test having that logic themselves
- Remove all redundant code related to `field_display_filtering.spec`, such as selecting a dataset, setting datepicker if relevant, etc
- Update `saved-queries-x.spec` to remove redundant functions as well as refactor certain tests to work isolated.
    - For example: `should see all saved queries` depended on previous tests to have already made the queries. So I removed this test and added the logic to the end of the previous test instead

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


cypress/utils/apps/query_enhancements/saved_queries.js
```diff
@@ -1,207 +0,0 @@
...
-          break;
-        default:
-          throw new Error(
-            `generateAllTestConfigurations encountered unsupported dataset: ${dataset.name}`
-          );
-      }
-      return generateTestConfiguration(datasetToUse, dataset.name, language);
-    })
-  );
-};
```
angle943, line 207: All of these utilities already existed in `saved_search.js` util file. So I renamed the `saved_search.js` file to `saved.js` and changed the names/comments of some of the functions so that it is clear that it is to be used by both `savedSearch` and `savedQueries`

cypress/utils/apps/query_enhancements/shared.js
```diff
@@ -0,0 +1,80 @@
...
+ * @param {string} dataset - the dataset name
+ * @param {QueryEnhancementDataset} datasetType - the type of the dataset
+ * @param {QueryEnhancementLanguageData} language - the relevant data for the query language to use
+ * @returns {object}
+ */
+
+/**
+ * Returns an array of test configurations for every query language + dataset permutation
+ * @param {GenerateTestConfigurationCallback} generateTestConfigurationCallback - cb function that generates a test case for the particular permutation
+ * @returns {object[]}
```
angle943, line 49: Note: This return type have been updated from `SavedSearchTestConfig[]` to generic `object[]`. The return type of this function is dictated by the output of the `generateTestConfigurationCallback` function.

If JSDoc had full support for generics or helpful TS utilities like `ReturnType`, I would be able to correctly type the return type of this function based on the return type of the callback, but based on my research JSDoc falls short when it comes to generics.

So unfortunately I had to resort to generic `object[]`. One another example of why we should convert our cypress files to be fully Typescript (not even mentioning just how annoying it generally is to type things using JSDoc)

ananzh (approved):

cypress/utils/apps/query_enhancements/shared.js
LDrago27, line 1: nit: Can we update the name to `common_utils.js`

cypress/utils/apps/query_enhancements/saved.js
LDrago27, line 1: nit: Can we rename it to `saved_object_utils.js` .

LDrago27 (approved):
Looks good. A couple of nits but no blockers.

cypress/utils/apps/query_enhancements/saved.js
angle943: are saved queries considered a saved object? I tried saving a saved query and it does not appear under `assets/object` page. I initially considered `saved_object` but after seeing this behavior I decided against it to avoid potential confusion.

Also in regard to the naming convention of `X_utils.js`, I am ok with that change but for context, we have several files under this directory, all without the `_utils` suffix, so if I change it for these files, I should change it for all files. (all these `utils` are under `cypres/UTILS/apps/query_enhancements` so I thought the `_utils` was not needed)

cypress/utils/apps/query_enhancements/shared.js
angle943: See my comment below for `_utils.js` suffix

cypress/utils/apps/query_enhancements/saved.js
LDrago27: saved_queries are also considered a kind of saved object. You can select type as query to filter the assets in assets page to view it. So i think the moniker of saved_objects should be good. 

That's nice point which I overlooked that all of the helpers are within the utils folder, which makes `_utils` kind of redundant. I think we should be good not using the `_utils` suffix.

PR #9223, opened 2025-01-20
Add navControlsPrimaryHeaderRight slot to header

wanglam:
### Description

This PR is for adding a new nav control slot for the primary header when the new home experience is enabled. There are many existing right nav control slots in the page header, but currently, there is no existing slot to add components to the right side of the primary header, next to the breadcrumbs bar.

Here are some screenshots to explain the existing slots in different page headers:

- Legacy page header (new home disabled):
image

- Application page header (new home enabled):
image

- Basic page header (new home enabled):
image

With this PR, a new nav control slot has been introduced to allow components to be placed in the primary header, next to the breadcrumbs bar, when the new home experience is enabled. Here's how it will look:

- Basic page header (new home enabled):
image

- Application page header (new home enabled):
image

## Changelog

- feat: Add navControlsPrimaryHeaderRight slot to header

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


SuZhou-Joe (approved):

src/core/public/chrome/ui/header/header.tsx
```diff
@@ -623,7 +631,10 @@ export function Header({
           {renderRecentItems()}
           {actionMenu}
         </EuiHeaderSection>
-        <EuiHeaderSection side="right">{rightControls}</EuiHeaderSection>
+        <EuiHeaderSection side="right">
+          {rightControls}
+          {renderNewPrimaryHeaderRight()}
```
SuZhou-Joe, line 636: A little concern here that it serves as rightControls when the header is applicationHeader, but in order to make the registration unified we seem not to have other choice.

wanglam: Registering on the same slot can avoid a component being mounted in different slots on the same page. If we register one component in `expandedRight`, maybe it will work fine in the application header. But there will be two same components in the normal header because we register them in both `expandedRight` and `newPrimaryHeaderRight`. We need to find some way to hide it. There are two solutions:

- Solution 1: Subscribe to the header variant and only register `expandedRight` when the header variant equals the application header.
In this solution, we may need to provide some way to unregister `expandedRight`. The component should be unregistered after switching to the normal page header.

- Solution 2: Subscribe to the header variant on the component side, and do not render the component if the header variant equals the normal header.
In this solution, we don't need to implement the unregister method; we only need to modify the component side.

Do you have any other thoughts about that?

SuZhou-Joe: I think current implementation looks good to me. I put the comment here for providing some context.

Solution 1 seems too much for plugin. I'd prefer solution 2 with header variant passed in props.

src/core/public/chrome/nav_controls/nav_controls_service.ts
```diff
@@ -64,6 +64,8 @@ export interface ChromeNavControls {
   registerCenter(navControl: ChromeNavControl): void;
   /** Register a nav control to be presented on the left-bottom side of the left navigation. */
   registerLeftBottom(navControl: ChromeNavControl): void;
+  /** Register a nav control to be presented on the right side of the new primary chrome header. */
```
ruanyl, line 67: what does the "new primary chrome header" refer to?

ruanyl: So we're trying to introduce an API to register a global control button which is available for all application headers and page headers. Can we simply call it `registerHeaderRight`?

wanglam: I'm open to renaming it to registerHeaderRight. I just want to add some explanations for my naming convention. I added 'New' to distinguish it from the legacy header since this slot won't be displayed in that header. I added 'Primary' to distinguish it from the secondary header, as it's positioned on the right side of the primary header. Do you think this naming convention is reasonable? Using 'headerRight' is simpler, but I'm not sure if it might confuse others into thinking that the registered elements won't be displayed.

ruanyl: Thanks for the explanation! I got the point of "Primary" now :)
For "New", since legacy header is meant to be deprecated, so I don't feel necessary to distinguish it from the legacy one, imagine in the future, the legacy header was removed, then the "New" will sound a bit strange. what about calling it `registerPrimaryHeaderRight`?

ruanyl (approved):

SuZhou-Joe (approved):

PR #9248, opened 2025-01-22
[Discover] Query Editor Shortcut Fires Incorrect Query

sejli:
### Description

* When using keyboard shortcuts to run a query in Discover 2.0, the date range gets reverted to Discover's last date range state. Since there is no date range passed into the `onSubmit()` function from the editor shortcut, the `QueryEditorTopRow` uses the props passed from the search bar as dates, which may be out of sync compared to the query service state. This PR passes in the correct date form the query service.
* Additionally, in the multiline editor, the `query` object was being used to construct the new query. This PR updates it to match the single line editor, which uses the `queryRef` defined in the same file.

## Changelog

-fix: Fix Discover query editor enter shortcuts reverting date range

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/query_editor/query_editor.tsx
+++ b/src/plugins/data/public/ui/query_editor/query_editor.tsx
@@ -80,6 +80,7 @@ export const QueryEditorUI: React.FC<Props> = (props) => {
   const editorQuery = props.query; // local query state managed by the editor. Not to be confused by the app query state.
 
   const queryString = getQueryService().queryString;
+  const timefilter = getQueryService().timefilter.timefilter;
   const languageManager = queryString.getLanguageService();
   const extensionMap = languageManager.getQueryEditorExtensionMap();
   const services = props.opensearchDashboards.services;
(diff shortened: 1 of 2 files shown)


kavilla (approved):

kavilla:
did we want to get this into 2.19?

abbyhu2000 (approved):

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -354,11 +355,11 @@ export const QueryEditorUI: React.FC<Props> = (props) => {
 
       editor.addCommand(monaco.KeyCode.Enter, () => {
         const newQuery = {
-          ...query,
+          ...queryRef.current,
```
paulstn, line 358: Regarding this change, I see that earlier in the file, `queryRef` exists for monaco 'commands' to access the latest query state. 
If `query` was a stale query state, why wasn't the single line editor failing a test that would update the query? 
If there is a test for an updated query state, and `query` has the updated query state, do we need queryRef at all? It only seems to be used here.

paulstn (approved):

PR #9290, opened 2025-01-28
[TESTID-66] Add histogram tests

silvaf-dev:
### Description

This PR incorporates all four test suites from [TESTID-66](https://github.com/orgs/opensearch-project/projects/233/views/1?filterQuery=Histogram+Interaction&pane=issue&itemId=88793695&issue=opensearch-project%7COpenSearch-Dashboards%7C8944), save for the interactions with the inner elements of the histogram.

## Screenshot

https://github.com/user-attachments/assets/508032ac-d984-42a2-8d55-429e4a8b32ad

## Changelog

- test: add histogram interaction tests

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


angle943:
@ananzh @LDrago27 is it okay if the changelog is skipped for this?

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
```diff
@@ -0,0 +1,223 @@
...
+    cy.addDataSource({
+      name: DATASOURCE_NAME,
+      url: `${SECONDARY_ENGINE.url}`,
+      authType: 'no_auth',
+    });
+    // Create workspace
+    cy.deleteWorkspaceByName(workspace);
+    cy.visit('/app/home');
+    cy.osd.createInitialWorkspaceWithDataSource(DATASOURCE_NAME, workspace);
+    cy.wait(2000);
```
angle943, line 40: Our other test don't seem to have this. can we get rid of it?

ananzh:
> @ananzh @LDrago27 is it okay if the changelog is skipped for this?

@silvaf-dev why skip changelog?

ananzh:
@silvaf-dev could you upload a video?

ananzh:
I tried to update but it seems only for author

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
```diff
@@ -0,0 +1,223 @@
...
+
+  afterEach(() => {
+    cy.deleteWorkspaceByName(workspace);
+    cy.deleteDataSourceByName(DATASOURCE_NAME);
+    // TODO: Modify deleteIndex to handle an array of index and remove hard code
+    cy.deleteIndex(INDEX_WITH_TIME_1);
+  });
+
+  generateAllTestConfigurations(generateHistogramInteractionTestConfiguration).forEach((config) => {
+    const selectLangs = () => {
```
ananzh, line 66: Instead of use all configuration and then redefine langs. I would suggest you to define your own configuration function in utils histogram.js
```
export const HistogramDatasetTypes = {
  INDEX_PATTERN: {
    name: 'INDEX_PATTERN',
    supportedLanguages: [
      DQL,
      QueryLanguages.SQL,
      QueryLanguages.PPL,
    ],
  },
   INDEXES: {
     name: 'INDEXES',
     supportedLanguages: [QueryLanguages.PPL],
   },
};

export const generateHistogramTestConfigurations = (generateTestConfigurationCallback, options = {}) => {
    const { indexPattern = INDEX_PATTERN_WITH_TIME, index = INDEX_WITH_TIME_1} = options;
    return Object.values(HistogramDatasetTypes).flatMap((dataset) =>
      dataset.supportedLanguages.map((language) => {
        let datasetToUse;
        switch (dataset.name) {
          case HistogramDatasetTypes.INDEX_PATTERN.name:
            datasetToUse = indexPattern;
            break;
          case HistogramDatasetTypes.INDEXES.name:
            datasetToUse = index;
            break;
          default:
            throw new Error(
              `generateHistogramTestConfigurations encountered unsupported dataset: ${dataset.name}`
            );
        }
        return generateTestConfigurationCallback(datasetToUse, dataset.name, language);
      })
    );
  };
```

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
```diff
@@ -0,0 +1,223 @@
...
+        DQL: 'be.visible',
+        Lucene: 'be.visible',
+        'OpenSearch SQL': 'not.exist',
+        PPL: 'be.visible',
+      };
+      const assertion = assertionByLang[config.language];
+      cy.getElementByTestId('dscChartChartheader').should(assertion);
+      cy.getElementByTestId('discoverChart').should(assertion);
+      // check interval selection persistence
+      // skipping SQL because it has no histogram, and PPL for indeces because permutations are unnecesary
```
ananzh, line 91: skipping --> skip

what does this `and PPL for indeces because permutations are unnecesary` mean? Is it because when you choose SQL as default languagle then switch to PPL then it won't show histogram? If so you can add `Switch language won't recreate the the dataset so no timefield will be selected and no histogram`

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
```diff
@@ -0,0 +1,223 @@
...
+      if (
+        config.language !== 'OpenSearch SQL' &&
```
ananzh, line 93: with redefine your configuration, you should be able to remove this or simplify this because you won't do SQL at all

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
```diff
@@ -0,0 +1,223 @@
...
+                    // TO DO: interact with the inner histogram
+                  });
+                });
+              });
+          });
+      }
+    });
+
+    it(`check the Auto interval value for ${config.testName}`, () => {
+      if (config.language !== 'OpenSearch SQL') {
```
ananzh, line 121: same issue

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
```diff
@@ -0,0 +1,223 @@
...
+        cy.setDataset(config.dataset, DATASOURCE_NAME, config.datasetType);
```
ananzh, line 122: I think we should use random data source name right?

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
```diff
@@ -0,0 +1,223 @@
...
+        cy.setQueryLanguage(config.language);
+        setDatePickerDatesAndSearchIfRelevant(config.language);
+        selectLangs().forEach((lang) => {
+          cy.setQueryLanguage(lang);
+          cy.wait(1000);
```
ananzh, line 127: do we need this wait?

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
```diff
@@ -0,0 +1,223 @@
...
+          cy.getElementByTestId('discoverIntervalSelect').should('have.value', 'auto');
+          cy.getElementByTestId('discoverIntervalDateRange').should('be.visible');
+          // TO DO: check histogram visualization
```
ananzh, line 130: I think you missed two aspects in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9290/files for 
```
Select index pattern with time fieldSet time range to cover multiple monthsVerify the default auto interval displays autoTest with different time intervals (Weekly, Day) and verify histogram displaySwitch to PPL and verify visualization remains consistent across languages
```
1) we need to switch to different intervals 
2) we need to switch language to verify the state persist like if you are using DQL - INDEX PATTERN and week, you need to switch to PPL and Lucence and check it is still week

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
```diff
@@ -0,0 +1,223 @@
...
+            });
+          });
+        }
+      } else {
+        cy.getElementByTestId('discoverIntervalSelect').should('not.exist');
+      }
+    });
+
+    it(`check collapse/expand functionality and state persistence for ${config.testName}`, () => {
+      if (config.language !== 'OpenSearch SQL') {
```
ananzh, line 188: same issue

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
```diff
@@ -0,0 +1,223 @@
...
+            cy.getElementByTestId('discoverChart').should(permutation[perm]);
+            // Uncomment after reload bug is fixed
+            //cy.reload(); // should not remove cache for the test to be meaningful
+            //cy.getElementByTestId('discoverChart').should(permutation[perm]);
+            //cy.getElementByTestId('histogramCollapseBtn').should(permutation[perm]);
+          });
+          if (perm === 'collapse') {
+            // start again from the beginning and expand again
+            cy.setQueryLanguage(config.language);
+            cy.getElementByTestId('histogramCollapseBtn').should('be.visible').click();
```
ananzh, line 217: Is this for checking collapse and expand state persist across lanuage? The button is always called histogramCollapseBtn right? It is this aria-expanded attr shown whether it is collapse or expand right?
```
<button class="euiButtonIcon euiButtonIcon--primary euiButtonIcon--empty euiButtonIcon--xSmall" type="button" aria-expanded="true" aria-label="Toggle histogram" data-test-subj="histogramCollapseBtn"><svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" class="euiIcon euiIcon--small euiIcon--inherit euiButtonIcon__icon" focusable="false" role="img" aria-hidden="true"><path fill-rule="nonzero" d="m5.157 13.069 4.611-4.685a.546.546 0 0 0 0-.768L5.158 2.93a.552.552 0 0 1 0-.771.53.53 0 0 1 .759 0l4.61 4.684c.631.641.63 1.672 0 2.312l-4.61 4.684a.53.53 0 0 1-.76 0 .552.552 0 0 1 0-.771Z"></path></svg></button>
```
So you need to check discoverChart is visible or not right?

ananzh (changes requested):

silvaf-dev:
> @silvaf-dev could you upload a video?

Done!

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
silvaf-dev: > do we need this wait?

Yes, otherwise we'll be getting false positives. Cypress will check the previous value instead of the updated value.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
silvaf-dev: I thought that the only random name is that of the workspace, not the datasource

silvaf-dev:
> @ananzh @LDrago27 is it okay if the changelog is skipped for this?

I tried to add the changelog manually but GitHub Actions is complaining, as you can see above.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
angle943: I have verified here that there's not much visual indicator that we can test against that the display has been updated. but let's add a comment above the `cy.wait`, something like `// adding a wait here to ensure the data has been updated`

cypress/utils/apps/query_enhancements/histogram_interaction.js
```diff
@@ -0,0 +1,29 @@
...
+ * @param {QueryEnhancementDataset} datasetType - the type of the dataset
+ * @param {QueryEnhancementLanguageData} language - the relevant data for the query language to use
+ * @returns {HistogramInteractionTestConfig}
+ */
+export const generateHistogramInteractionTestConfiguration = (dataset, datasetType, language) => {
+  return {
+    dataset,
+    datasetType,
+    language: language.name,
+    testName: `dataset: ${datasetType} and language: ${language.name}`,
```
angle943, line 27: i think what could help here is if you add:

`...language.supports`

and update the JSDOC to have:

```
@property {boolean} histogram - whether the histogram appears
```

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
angle943: i personally don't feel this function is needed at all, let me clarify in another comment

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
silvaf-dev: 1. Yes, this is for checking persistence across languages.
2. Yes, it's the same button.
3. Yes, because we're talking about the same button in #2.

First I check that it's visible (expanded) by default and then that it doesn't exist after I collapse it. Then I check the persistence across the remaining languages for both expand and collapse. Hope this clears things up.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
```diff
@@ -0,0 +1,223 @@
...
+      setDatePickerDatesAndSearchIfRelevant(config.language);
+      const assertionByLang = {
+        DQL: 'be.visible',
+        Lucene: 'be.visible',
+        'OpenSearch SQL': 'not.exist',
+        PPL: 'be.visible',
+      };
+      const assertion = assertionByLang[config.language];
+      cy.getElementByTestId('dscChartChartheader').should(assertion);
+      cy.getElementByTestId('discoverChart').should(assertion);
```
angle943, line 89: I would suggest refactoring lines 81-89 to:

```
      // check the visibility of histogram
      if (config.histogram) {
        cy.getElementByTestId('dscChartChartheader').should('be.visible');
        cy.getElementByTestId('discoverChart').should('be.visible');
      } else {
        cy.getElementByTestId('dscChartChartheader').should('not.exist');
        cy.getElementByTestId('discoverChart').should('not.exist');
      }
 ```

This way if the support for histogram changes in the future, we only need to change it in one area

File: cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js [hunk 13]
```diff
@@ -0,0 +1,223 @@
...
+                    cy.wrap($currentRange)
+                      .invoke('text')
+                      .then(($currentRangeTxt) => {
+                        expect($currentRangeTxt).to.equal($originalRangeTxt);
+                      });
+                    // TO DO: interact with the inner histogram
+                  });
+                });
+              });
+          });
```
angle943, line 116: I personally feel all these nesting makes it very hard to read and figure out what is happening. And because we are not checking for IDX at all, why not just skip it entirely:


```
      // check interval selection persistence across language changes. Only need to check for INDEX_PATTERNS because INDEXES
      // only supports SQL & PPL, and only one of the two supports histogram.
      if (config.histogram && config.datasetType === DatasetTypes.INDEX_PATTERN.name) {
        cy.getElementByTestId('discoverIntervalSelect').select('Week');
        cy.getElementByTestId('discoverIntervalDateRange').contains(
          `${START_TIME} - ${END_TIME} per`
        );

        for (const language of DatasetTypes.INDEX_PATTERN.supportedLanguages) {
          cy.setQueryLanguage(language.name);
          // wait a bit to ensure data is loaded
          cy.wait(1000);
          if (language.supports.histogram) {
            cy.getElementByTestId('discoverIntervalSelect').select('Week');
            cy.getElementByTestId('discoverIntervalDateRange').contains(
              `${START_TIME} - ${END_TIME} per`
            );
          }
        }
      }
  ```

Comment (angle943) [reply cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js:116]: My implementation changes it to it's own language, but not a big deal

File: cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js (hunk 8 above)
Comment (silvaf-dev) [reply cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js:130]: Should I switch to week/day only?
I'm already checking persistence for week on ``it(`check histogram visibility for ${config.testName}`)``

File: cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js (hunk 3 above)
Comment (ananzh) [reply cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js:91]: Yeah it is better to have your own config instead of using a lot of if. Since we don't need to verify SQL at all for histogram except verifying the existence. 

Meanwhile why IDX + PPL = skip?

File: cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js (hunk 6 above)
Comment (ananzh) [reply cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js:122]: currently all tests have random datasource name

File: cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js (hunk 8 above)
Comment (ananzh) [reply cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js:130]: I think it is better to switch and verify all the intervals.

Comment (ananzh) [reply cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js:130]: I don;t think you verify the interval state is persistent in that test right? That is for visibility, like show or no show histogram. 
When you change the interval, e.g. from auto to day and you switch language the interval should be day and should not go back to default auto.

**Review by ananzh — CHANGES_REQUESTED** [review]

File: cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js (hunk 3 above)
Comment (silvaf-dev) [reply cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js:91]: > Meanwhile why IDX + PPL = skip?
It's because it doesn't make sense to switch from PPL to SQL since there's no persistence to check at all.

File: cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js [hunk 14]
```diff
@@ -0,0 +1,211 @@
...
+      cy.setDataset(config.dataset, datasource, config.datasetType);
+      cy.setQueryLanguage(config.language);
+      setDatePickerDatesAndSearchIfRelevant(config.language);
+      const intervals = ['auto', 'ms', 's', 'm', 'h', 'd', 'w', 'M', 'y'];
+      cy.getElementByTestId('discoverIntervalSelect').should('have.value', intervals[0]);
+      intervals.forEach((interval) => {
+        cy.getElementByTestId('discoverIntervalSelect').select(interval);
+        cy.wait(1000); // adding a wait here to ensure the data has been updated
+        config.langPermutation.forEach((lang) => {
+          if (lang === QueryLanguages.SQL.name) return;
```
ArgusLi, line 115: nit: could you add a comment here stating why this is done?

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
```diff
@@ -0,0 +1,211 @@
...
+          cy.setQueryLanguage(lang);
+          cy.getElementByTestId('discoverIntervalSelect').should('have.value', interval);
+          cy.getElementByTestId('discoverIntervalDateRange').should('be.visible');
+        });
+      });
+      // TODO: check histogram visualization
```
ArgusLi, line 121: nit: could you add more detail why this is TODO?

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
```diff
@@ -0,0 +1,211 @@
...
+        cy.getElementByTestId('docTableExpandToggleColumn')
+          .eq(0)
+          .find('button')
+          .click({ force: true }); // reset state
+      };
+      cy.setDataset(config.dataset, datasource, config.datasetType);
+      cy.setQueryLanguage(config.language);
+      if (config.isHistogramVisible) {
+        if (config.language !== 'PPL') {
+          // remove after the bug is fixed
```
ArgusLi, line 155: nit: could you add more details about the bug here, or a bug issue link?

ArgusLi (approved):

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/histogram_interaction.spec.js
```diff
@@ -0,0 +1,218 @@
...
+          cy.getElementByTestId('docTableExpandToggleColumn')
+            .eq(0)
+            .find('button')
+            .click({ force: true }); // reset state
+        };
+        cy.setDataset(config.dataset, DATASOURCE_NAME, config.datasetType);
+        cy.setQueryLanguage(config.language);
+        if (config.isHistogramVisible) {
+          // related bug
+          // https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9294
```
ArgusLi, line 157: Could you please add `TODO:` to the front of the comment?

ananzh:
Can you add your test to package.json so one ciGroup can pick it up?

ananzh (approved):

ananzh (approved):

SuZhou-Joe (approved):

Issue #9020, opened 2024-12-06
Horizontal Scroll Bar Not Rendered Until Last Record in search query result is Reached in OpenSearch Dashboards v2.18 which was working on for few records on OSD 1.3.17  version


rakhinarveka1:
**What is the bug?**
In OpenSearch Dashboards version 2.18, the horizontal scroll bar in the table displaying search query results does not render  until the user scrolls to the last record in the table. This behavior was not observed in version 1.3.17, where the horizontal scroll bar was immediately accessible, allowing users to scroll horizontally regardless of their vertical position in the table.

This issue impacts usability, especially for tables with a significant number of columns, as users cannot view data that requires horizontal scrolling until they manually scroll down to the last record.

**How can one reproduce the bug?**

Open OpenSearch Dashboards version 2.18.
Run a search query that returns results with multiple columns and rows.
Try to use the horizontal scroll bar without scrolling down to the last record.
Observe that the horizontal scroll bar is neither rendered nor operable until the last record is reached.

**What is the expected behavior?**
The horizontal scroll bar should be rendered and operable immediately, allowing horizontal navigation irrespective of the user's vertical position in the table.

Actual Behavior
The horizontal scroll bar is only rendered or becomes operable when the user scrolls to the last record in the table.

**What is your host/environment?**
OpenSearch Dashboards Version: 2.18

**Do you have any screenshots?**
attaching screenshot as per mentioned version for the same.
Opensearch-dashboards   1.13.7
 
![image](https://github.com/user-attachments/assets/749749fd-2304-4077-9fd0-54cc0218bf25)

Opensearch Dashboard 2.18.0

 
![image](https://github.com/user-attachments/assets/21c1b5f4-5996-4baf-b13d-ddc72a7526a0)
![image](https://github.com/user-attachments/assets/3df92315-3214-46b7-8910-1cd515249a2c)

 

**Do you have any additional context?**
Add any other context about the problem.

cwperks:
Transferring this issue to OpenSearch-Dashboards repo for comment and investigation

dblock:
[[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/dblock), [2](https://github.com/andrross/), [3](https://github.com/reta), [4](https://github.com/gaiksaya), [5](https://github.com/krisfreedain), [6](https://github.com/acarbonetto)]

kavilla:
[...]

Issue #7204, opened 2024-07-09
[BUG] "Add to Dashboards after saving" not working when creating a visualization


dmoreira-sbp:
**Describe the bug**

In one of our OpenSearch environments, the "Add to Dashboards after saving" option doesn't seem to be working correctly.

**To Reproduce**
1. Go to a dashboard
2. Click on 'Edit', 'Create new', select any type of visualization and then select the source
3. Click  'Save', enter a title and keep the 'Add to Dashboards after saving' selected. 
4. It should return to the dashboard but it doesn't add it right away.

It does work if you add it after the fact. 

**Expected behavior**

**OpenSearch Version**
2.13.0 (AWS-Managed)

**Dashboards Version**
2.13.0 (Not-AWS managed. Running "opensearchproject/opensearch-dashboards:2.13.0" on AWS ECS

What I did notice is that in the GET request in which it tries to fetch the visualization id by name, it returns an empty list (adding screenshots)

**Screenshots**
EDIT: I've just noticed that the request that you see in the screenshots might be the one that the dashboard does to check if there's already another visualization with the same name, so it might not apply

![Screenshot 2024-07-09 120800](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/104890819/584c0865-cbda-4b1a-b529-1c8a79d38b35)

![Screenshot 2024-07-09 120831](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/104890819/f4633299-c348-4056-9109-36b3652a484d)
-f3d5ba04f224)
![Screenshot 2024-07-09 120845](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/104890819/c515269b-05e0-4869-8a50-e325ab83cc1e)

dmoreira-sbp:
Hi! After researching a bit more with a local environment, I see that in the "_find" call in my local environment it gets several related objects in the results, but it doesn't happen in the production environment 
![image](https://github.com/user-attachments/assets/07f99c96-b80b-422b-adf2-684415397fbf)
![image](https://github.com/user-attachments/assets/05bf9d61-6f1e-4453-a659-effc92c371e1)

neodescis:
Do any of you by chance have storeInSessionStorage turned on? It seems that with that enabled, the state data stored in location history, which is needed to add the visualization to the dashboard, doesn't survive the transition back to the dashboard page.

PR #9298, opened 2025-01-29
[Discover] chore: Improve scrolling experience

Maosaic:
### Description

Improve scrolling experience on Discover page:

1. Make vertical scrolling happen inside table rather instead of the entire right panel.(This change also bring horizontal scrolling to the bottom of the viewport instead of staying at the end of the page)
2. Make table header sticky.

### Issues Resolved

1. If user scroll down the table on Discover the entire view scrolls down. The user is not able to see their columns that they have selected nor can they even see the query that they ran.
2. There is no indication of overflow horizontally wise the user will not know that they can scroll and depending on their setup that might not even be able to scroll horizontally 

## Screenshot

https://github.com/user-attachments/assets/4e03f1c8-c589-4aa4-b3c2-8f8fbefe5917

## Changelog

- feat: Improve scrolling experience on Discover page.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/components/default_discover_table/_doc_table.scss
+++ b/src/plugins/discover/public/application/components/default_discover_table/_doc_table.scss
@@ -128,10 +128,16 @@ doc-table {
 }
 
 .osd-table {
+  border-collapse: separate;
+
+  tr:first-child td {
+    border-top: none;
+  }
+
   thead {
     position: sticky;
     top: 0;
-    background-color: white;
+    background-color: $euiColorEmptyShade;
     z-index: 1;
   }
 
--- a/src/plugins/discover/public/application/components/default_discover_table/_doc_table.scss
+++ b/src/plugins/discover/public/application/components/default_discover_table/_doc_table.scss
@@ -128,6 +128,13 @@ doc-table {
 }
 
 .osd-table {
+  thead {
+    position: sticky;
+    top: 0;
+    background-color: white;
+    z-index: 1;
+  }
+
   // sub tables should not have a leading border
   .table .table {
     margin-bottom: 0;
--- a/changelogs/fragments/9298.yml
+++ b/changelogs/fragments/9298.yml
@@ -1,0 +1,2 @@
+feat:
+- Improve scrolling experience on Discover page. ([#9298](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9298))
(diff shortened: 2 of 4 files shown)


angle943:
Hey the funcitonality looks good! I checked it out to test if another annoyance i encountered during my testing was fixed (sometimes the table would jump around a bit), but I was not able to replicate it right now (both in your PR but also in `main`).

One slight UX issue though, do we want the bottom bar for the header to be stickied too? Its a little jarring seeing it like this:

Screenshot 2025-01-30 at 9 51 20 AM

Maosaic:
No sure if there is a perfect solution for the bottom border, since adding it will also "visually breaking" row height when scrolling.

angle943:
@Maosaic here's a couple examples i found in some of the widely used tables:

1. MUI - https://github.com/opensearch-project/OpenSearch-Dashboards/issues/7204#issuecomment-2344526099
Screenshot 2025-01-30 at 10 25 26 AM

2. Cloudscape (although this example is for stickied columns)
Screenshot 2025-01-30 at 10 27 08 AM

I can find a couple more examples but i think the standard is either to have the border or a shadow

src/plugins/discover/public/application/components/default_discover_table/_doc_table.scss
```diff
@@ -128,6 +128,13 @@ doc-table {
 }
 
 .osd-table {
+  thead {
+    position: sticky;
+    top: 0;
+    background-color: white;
```
kavilla, line 134: what if toggle in darkmode

also in general we should sass variables so that if there is a color scheme change then holistically it will change if it uses the sass variables

angle943: ah good comment thanks @kavilla . something to keep in mind that iw asn't thinking about

kavilla (approved):
LGTM

src/plugins/discover/public/application/components/default_discover_table/_doc_table.scss
```diff
@@ -128,6 +128,19 @@ doc-table {
...
+  border-collapse: separate;
+
+  tr:first-child td {
+    border-top: none;
+  }
+
+  thead {
+    position: sticky;
+    top: 0;
+    background-color: $euiColorEmptyShade;
```
kavilla, line 140: makes sense just want to check with @canascar

kavilla:
@Maosaic I think we can close this issue correct? https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9020

PR #9304, opened 2025-01-30
Lighthouse performance metrics setup

ruchidh:
### Description

This workflow runs Lighthouse CI to measure the performance of OpenSearch Dashboards. It logs performance warnings without failing the CI when failing the metrics like first time load and speed of page. Additionally, it provides a hosted Lighthouse report URL in PR comments for tracking performance regressions over time.

## Screenshot

Screenshot 2025-02-20 at 12 35 07 PM

## Testing the changes

- yarn lhci autorun

## Changelog
- feat:  Lighthouse Page Performance Metrics CI workflow

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


.lighthouserc.js
```diff
@@ -0,0 +1,28 @@
...
+        'largest-contentful-paint': ['warn', { maxNumericValue: 2500 }],
+        interactive: ['warn', { maxNumericValue: 5000 }],
+        'total-blocking-time': ['warn', { maxNumericValue: 300 }],
+      },
+    },
+    upload: {
+      target: 'temporary-public-storage', // Required for GitHub integration
+    },
+  },
+};
```
angle943, line 28: which OS instance does this connect to?

.lighthouserc.js
```diff
@@ -0,0 +1,28 @@
+const baseUrl = 'http://localhost:5601';
+
+module.exports = {
+  ci: {
+    collect: {
+      url: [`${baseUrl}`, `${baseUrl}/app/data-explorer/discover`, `${baseUrl}/app/dashboards`], // Add more URLs as needed
```
angle943, line 6: Something i'm wondering about...

Without some specific actions, I don't think any results would be loaded in `discover` and no dashboards will be set up on `dashboards` page, so i'm not sure if we are getting the full benefit of having this.

Is it possible to start with some examples that'll load? perhaps the `--run-with-examples` flag (but even with that i think we need to modify the URL so we are hitting the correct page. For example `/app/dashboards` might just go to the list of dashboards you need to select from)

.lighthouserc.js
```diff
@@ -0,0 +1,28 @@
...
+      numberOfRuns: 3,
+      settings: {
+        chromePath: require('puppeteer').executablePath(),
+        chromeFlags: '--no-sandbox --disable-gpu --headless',
+      },
+    },
+    assert: {
+      preset: 'lighthouse:recommended',
+      assertions: {
+        performance: ['error', { minScore: 0.9 }],
```
angle943, line 17: oof i think we need make the minscore much lower. and it might differ between the different URLs.

For example on `dashboards` page with the example data loaded, i am getting a score of `34` (yikes)

.github/workflows/performance_testing.yml
```diff
@@ -0,0 +1,51 @@
+name: Performance Testing
```
angle943, line 1: do we need:

```
# Copyright OpenSearch Contributors
# SPDX-License-Identifier: Apache-2.0
```

.lighthouserc.js
```diff
@@ -0,0 +1,28 @@
+const baseUrl = 'http://localhost:5601';
```
angle943, line 1: do we need a license here? (like .eslintrc.js)

.lighthouserc.js
ruchidh: This is only for testing purpose, would change in final review.

.lighthouserc.js
ruchidh: 9200

package.json
```diff
@@ -464,6 +465,7 @@
     "postcss": "^8.4.5",
     "prettier": "^2.1.1",
     "prop-types": "^15.7.2",
+    "puppeteer": "^24.1.1",
```
kavilla, line 468: i would be hesitant to add this. i know some opensearch plugin uses puppeteer and they have to bundle a chrome binary so that it can not fail on some integration testing.

kavilla: i think other plugin teams have used puppeteer-core? not sure haven't looked too deep into it

.github/workflows/performance_testing.yml
```diff
@@ -0,0 +1,79 @@
...
+        shell: bash
+
+      - name: Remove security plugin
+        run: |
+          /bin/bash -c "yes | ./opensearch-${{ env.LATEST_VERSION }}/bin/opensearch-plugin remove opensearch-security"
+        shell: bash
+
+      - name: Run OpenSearch
+        run: |
+          /bin/bash -c "./opensearch-2.17.0/opensearch-tar-install.sh &"
```
kavilla, line 51: we should fix this if we have this somewhere else but if a different version then than 2.17 then this will not be right

.github/workflows/performance_testing.yml
```diff
@@ -0,0 +1,79 @@
...
+          sleep 30
+        shell: bash
+
+      - name: Install Lighthouse CI
+        run: yarn global add @lhci/cli
+
+      - name: Build OpenSearch Dashboards
+        run: yarn build
```
kavilla, line 59: build will build a platform distribution. are we sure we dont want to `yarn osd bootstrap`?

.github/workflows/performance_testing.yml
```diff
@@ -0,0 +1,79 @@
...
+          /bin/bash -c "yes | ./opensearch-${{ env.LATEST_VERSION }}/bin/opensearch-plugin remove opensearch-security"
+        shell: bash
+
+      - name: Run OpenSearch
+        run: |
+          /bin/bash -c "./opensearch-2.17.0/opensearch-tar-install.sh &"
+          sleep 30
+        shell: bash
+
+      - name: Install Lighthouse CI
```
kavilla, line 55: is there not a github action that we can pull from the marketplace and run this?

.lighthouserc.js
kavilla: +1, you can actually hit the api or bulk insert saved objects and data

.lighthouserc.js
kavilla, line 1: baseUrl technically doesn't have to be 5601. It can be configurable to be something like 443

config/opensearch_dashboards.yml
```diff
@@ -372,10 +372,10 @@
...
-# The default config is [], and no one will be dashboard admin. 
+# The default config is [], and no one will be dashboard admin.
 # If the user config is set to wildcard ["*"], anyone will be dashboard admin.
 # opensearchDashboards.dashboardAdmin.groups: ["dashboard_admin"]
 # opensearchDashboards.dashboardAdmin.users: ["dashboard_admin"]
 
 # Set the value to true to enable the new UI for savedQueries in Discover
-# data.savedQueriesNewUI.enabled: true
\ No newline at end of file
+# data.savedQueriesNewUI.enabled: true
```
kavilla, line 381: nit:  no changes observed besides white spacing

package.json
```diff
@@ -271,6 +271,7 @@
     "@elastic/filesaver": "1.1.2",
     "@elastic/makelogs": "^6.1.0",
     "@faker-js/faker": "^7.6.0",
+    "@lhci/cli": "^0.14.0",
```
kavilla, line 274: if this is a dev dependency then bootstrapping should be suffice enough correct? do we still need to add it the step to install it globally in the github workflow?

.lighthouserc.js
ruchidh: yes, have plan to do so, updating mock data

package.json
ruchidh: okay, let's discuss this

angle943:
We talked about this in private, but just for some permanence, I will write TLDR version here. Some challenges with this approach are:

- installing mock data to load the data
- Telling Lighthouse the correct URL to see the loaded data (challenge is that for dashboard, the created dashboards have randomly generated IDs in the URL. For Discover, we will need to modify the URL to have all the correct data and timestamps to load the proper data)

One way we can solve this is leveraging Cypress: https://medium.com/@mudit56p/enhancing-ui-performance-testing-with-lighthouse-and-cypress-1ad69db0a150

.lighthouserc.js
```diff
@@ -0,0 +1,50 @@
...
+    assert: {
+      // Only the key assertions
+      assertions: {
+        // performance: ['warn', { minScore: 0.2 }], // Aggregates all key metrics into a single score.
+        'first-contentful-paint': ['warn', { maxNumericValue: 1800 }], //	Indicates how fast the page starts loading.
+        'speed-index': ['warn', { maxNumericValue: 20000 }], // Lower speed index = better perceived performance.
+        // 'largest-contentful-paint': ['warn', { maxNumericValue: 2500 }], // Critical for user-perceived load speed.
+        // interactive: ['warn', { maxNumericValue: 5000 }], // Affects usability—page is responsive to user input.
+        // 'total-blocking-time': ['warn', { maxNumericValue: 300 }], // Higher TBT makes the page feel sluggish.
+        // 'cumulative-layout-shift': ['warn', { maxNumericValue: 0.1 }], // Prevents janky UI movements during load.
```
angle943, line 42: why are some of these commented out?

.lighthouserc.js
angle943: it would be good to pass this as a variable

.lighthouserc.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+const baseUrl = 'http://localhost:5601';
+
+module.exports = {
+  ci: {
+    collect: {
+      url: [
+        `${baseUrl}/app/home`,
+        `${baseUrl}/app/dashboards#/view/722b74f0-b882-11e8-a6d9-e546fe2bba5f`,
```
angle943, line 13: will this `id` always work?

lighthouse_baseline.json
```diff
@@ -0,0 +1,18 @@
...
+  },
+  "http://localhost:5601/app/dashboards": {
+    "first-contentful-paint": 1800,
+    "speed-index": 24000
+  },
+  "http://localhost:5601/app/visualize": {
+    "first-contentful-paint": 1800,
+    "speed-index": 22000
+  }
+}
```
angle943, line 18: I know JSON doesn't allow comments, but what if we switched this to YAML to allow comments? (and add comments to help readers understand how this is used and why this data structure is used)

.lighthouserc.js
ruchidh: yes, this is our fixed sample data of ecommerce , suggested by rocky. We tested it

lighthouse_baseline.json
ruchidh: the config has details actually regarding all these metrics already

.lighthouserc.js
ruchidh: So, usually 4-5 metrices we target for,  but ashwin insisted on keeping track only page speed and load one, 
over the time, we will see if we need any of them.

angle943 (approved):

package.json
```diff
@@ -257,6 +257,7 @@
     "tslib": "^2.0.0",
     "type-detect": "^4.0.8",
     "uuid": "3.3.2",
+    "wait-on": "^8.0.2",
```
kavilla, line 260: shouldn't this be a dev dependency?

package.json
```diff
@@ -468,6 +470,7 @@
     "postcss": "^8.4.5",
     "prettier": "^2.1.1",
     "prop-types": "^15.7.2",
+    "puppeteer": "^24.1.1",
```
kavilla, line 473: what was the follow up this one ?

package.json
kavilla: what was the follow up on this one?

lighthouse_baseline.json
```diff
@@ -0,0 +1,18 @@
+{
+  "http://localhost:5601/app/home": {
```
kavilla, line 2: would there be a way to config this or only baseline by path? like `app/data-explorer/discover`?

lighthouse_baseline.json
kavilla, line 1: i think this file shouldn't be here. 

for some more context: i need to look to verify, but some of the files that make it into the release distribution are files are files copied from specific locations. things in `test` and `src/dev` don't get ported over (for the most part). this ensures we keep the release artifact unpolluted. so might be better to move it to one of those folders. 

also nitpick, could it be like `baselines/lighthouse.json` ? or something if we store all these in the same folder

ruchidh: sounds good :)

lighthouse_baseline.json
ruchidh: i think with json , it would be tough, as you have to import it(http://localhost:5601) from somewhere else,  then in json wit won't work.
and as the ci/cd read this file, and json are easy to read than other format

package.json
ruchidh: makes sense, thanks for pointing

package.json
ruchidh: i checked some osd packages  using, puppeter-core, puppeter and puppeter-browser as some depenedencies, but not any plugin.

lighthouse_baseline.json
ruchidh: updates this with new method

angle943 (approved):

Maosaic:
It seems [Build and test / Lint and validate (pull_request)](https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/13908094913/job/38915726694?pr=9304) test is failing due to new packages been added, is this expected?

Also, can you check the error of [Build and test / Run functional tests on Windows (ciGroup2) (pull_request)](https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/13908094913/job/38915729710?pr=9304)?

ruchidh:
> It seems [Build and test / Lint and validate (pull_request)](https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/13908094913/job/38915726694?pr=9304) test is failing due to new packages been added, is this expected?
> 
> Also, can you check the error of [Build and test / Run functional tests on Windows (ciGroup2) (pull_request)](https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/13908094913/job/38915729710?pr=9304)?

cigroup2 error is failing for other reason, not related to the change, it gets passed earlier, i will check with anan on this if this can be avoidable

ananzh (approved):

PR #9333, opened 2025-02-05
add the ability to export the discover table current query result as CSV

angle943:
### Description

Customers have been requesting for the ability to export the current result set as CSV. This is a short term solution we have agreed upon to implement this.

## Screenshot

Screenshot 2025-02-04 at 7 08 34 PM

## Testing the changes

- Added some basic unit tests for the button, as well as doing a thorough manual testing.

## Changelog

- feat: add the ability to export current queried discover result as CSV

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+    (useSelector as jest.MockedFunction<any>).mockImplementationOnce(() => ['response']);
+    render(<DownloadCsvButton indexPattern={stubIndexPatternWithFields} rows={[{}, {}] as any} />);
+    expect(screen.getByText('Download 2 documents as CSV')).toBeInTheDocument();
+  });
+
+  it('Renders the text correctly for single row', () => {
+    (useSelector as jest.MockedFunction<any>).mockImplementationOnce(() => ['response']);
+    render(<DownloadCsvButton indexPattern={stubIndexPatternWithFields} rows={[{}] as any} />);
+    expect(screen.getByText('Download 1 document as CSV')).toBeInTheDocument();
+  });
```
joshuali925, line 39: maybe some test coverage on the download logic if possible?

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { buildColumns } from '../../../utils/columns';
+import { DOC_HIDE_TIME_COLUMN_SETTING } from '../../../../../common';
+import { getLegacyDisplayedColumns } from '../../default_discover_table/helper';
+
+export interface DownloadCsvButtonProps {
+  indexPattern: IndexPattern;
+  rows: OpenSearchSearchHit[];
+}
+
+export const DownloadCsvButton = ({ indexPattern, rows }: DownloadCsvButtonProps) => {
```
joshuali925, line 26: this seems to be entirely based on rows in memory. if we want to support downloading all rows as csv, would it need re-implementation and this logic would no longer be used?

src/plugins/discover/public/application/components/chart/chart.tsx
```diff
@@ -123,6 +134,7 @@ export const DiscoverChart = ({
       <EuiFlexItem grow={true} style={{ justifyContent: 'flex-start' }}>
         {isTimeBased && timeChartHeader}
       </EuiFlexItem>
+      {exportAsCsvButtonFlexItem}
```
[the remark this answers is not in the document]
angle943: That was what was requested by UX:

Screenshot 2025-02-04 at 2 13 04 PM

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
angle943: If we want to support downloading of all rows as CSV, that would be very possible on FE, albeit possible if we really wanted that feature.

To give context, we support the download of all rows as CSV via the reporting plugin, which handles that through a backend job. However this plugin is not available in all environments.

If we wanted that feature independent of the Reporting plugin, we would have to brainstorm on how we would implement that. Purely via FE is possible but we can end up making tons of requests to the backend to query for all the data.

But to answer your question, if the logic of querying for all of the data lives outside of this component, then the code that is written here WOULD be reusable, since once we query for all the data, the `rows` will contain all the said data

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.test.tsx
angle943: Yes I will add some more. I will also add an e2e test I think

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
```diff
@@ -0,0 +1,105 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React, { useMemo } from 'react';
+import { EuiButtonEmpty } from '@elastic/eui';
+import { FormattedMessage } from '@osd/i18n/react';
+import Papa from 'papaparse';
+// @ts-ignore
+import { saveAs } from '@elastic/filesaver';
```
zhongnansu, line 11: Can we choose an npm package that's being actively maintained? The github repo for this package is archived I believe, and the latest version is from 7 years ago. Lack of maintenance of 3rd party lib cause us trouble for things like CVE. Not to mention, moving forward beyond 3.0, we definitely want to get rid of all @elastic dependencies
https://www.npmjs.com/package/@elastic/filesaver

src/plugins/discover/public/application/components/chart/download_csv_button/index.ts
```diff
@@ -0,0 +1,31 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
```
zhongnansu, line 13: why we add a Elasticsearch license here?

angle943: ah bad copy and pasta job haha, will remove

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
angle943: @zhongnansu Hey! I appreciate this comment, as I was really hesitant on using this code too, but decided to keep with the 'norm' of the other codebases (this is used in other components).

But now that I'm not the only one thinking it, I will happily not use this. Its actually completely unnecessary IMO, it can be done with basic JS without any libraries

zhongnansu: got it. I am OK with it, since other legacy code also import this dependency. Not sure if we have a meta issue of removing all @elasctic dependency, definitely worth opening one to track IMO.

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.test.tsx
angle943: I have added some tests. The main challenge lies in the fact that, from what I can gather, we do not seem to have a proper stub/mock generator for `indexPattern` object. I've made a stub of it within my test file using a real API response I got, but that too is limited, because this data structure also contains functions that my implementation requires.

If there's a way to stub indexPatterns properly that someone knows about, I would love to hear

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
angle943: Thats a good idea, i'll leave it in for now but I have opened an issue to remove at least @elastic/filesaver globally from our codebase: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9341

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
joshuali925: I don't know if we want to do it on the frontend. Reporting does support all rows through scroll API, but we disabled it due to it causing OOM on the server and we added a 10k limit. We do have user requests to increase it (https://github.com/opensearch-project/dashboards-reporting/issues/501, https://github.com/opensearch-project/dashboards-reporting/issues/499)

> since once we query for all the data, the `rows` will contain all the said data

The `rows` contain the data only if we do this on the frontend. My concern is that it would cause OOM on client side as well, since adding all rows then converting them to CSV on the browser can be expensive. IMO the proper way is to implement stream processed CSV and `chunked` download rather than directly increasing the limit on reporting, or relying on client to do the work, but then I'm not sure if this PR can be reused

angle943: Ah i forgot how big actual data can get :D . You are completely right in that it would not be possible in FE to handle that for real-life data (unless we have a crazy process that puts every 10k rows into separate csv and start downloading them to users sequently, which is highly impractical)

Stream-processed CSV would not be possible on the FE, but if we start discussing about doing it on a BE job, then we go back to the issue of cost for our OS serverless solution. I think we will need a bigger discussion on what we want to do with downloading all hits as a team.

But you are right, the `rows` cannot be re-used

zhongnansu: given that we may still need to maintain based on thread discussion yesterday. You may want to consider move the implementation to reporting-dashboards plugin and expose an integration point for Discover in core. @angle943 

Also, I was away from reporting feature for years lol, need some refresh. Why is row a limit, the real limit would be the size right? @joshuali925

angle943: @zhongnansu one of the requirement was for this code to exist outside of the reporting plugin, because in our serverless offering the Reporting plugin is not available, yet this feature is still required

joshuali925: doesn't matter the OS architecture, OSD still has the nodejs server, which is what reporting plugin uses to generate CSV. If we stream process the scroll response from OS, it shouldn't be that expensive but needs some proof of concept and see if it's feasible

> one of the requirement was for this code to exist outside of the reporting plugin, because in our serverless offering the Reporting plugin is not available

is there any blocker to add reporting plugin there? but anyways let's keep it off from this issue

> Why is row a limitation, the real limit would be the size right?

we don't know the size of docs unless we search

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/export_as_csv.spec.js
```diff
@@ -0,0 +1,175 @@
...
+    afterEach(() => {
+      cy.deleteWorkspaceByName(workspaceName);
+      cy.osd.deleteDataSourceByName(DATASOURCE_NAME);
+      cy.osd.deleteIndex(INDEX_WITH_TIME_1);
+      cy.osd.deleteIndex(INDEX_WITHOUT_TIME_1);
+    });
+
+    generateExportAsCsvIndexPatternTestConfigurations().forEach((config) => {
+      it(`should be able to export for ${config.saveName}`, () => {
+        cy.navigateToWorkSpaceSpecificPage({
```
ananzh, line 85: I think for the test can we add two more scenarios 1) add a filter or use a query to filter then the number shown in the download can be updated? for example after adding filter we should see the number to download also updated to 127 2) In the advanced settings, if you reset the default size, then load discover, the number should be updated to the settings
Screenshot 2025-02-06 at 6 21 00 PM

Screenshot 2025-02-06 at 6 16 48 PM

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
angle943: ah i have forgotten that we have the OSD nodejs server. @joshuali925 thanks for the info how the reporting plugin works under the hood. For sure it is feasible and I can do it, but like you said I think we need better information on why the reporting plugin was not being added. Last time i talked with my manager it was due to costs (the reporting plugin also uses the server to do other tasks, so the actual export CSV might not be the reason that was causing the monetary issues)

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/export_as_csv.spec.js
angle943: sure I can add that test, that's a good idea

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
angle943: and to answer the "re-usability" aspect, the biggest challenge in this feature is making sure that:

- proper columns are added to the spreadsheet
- The cell values are properly formatted to be consistent with what is being shown in the UI

If we end up using the nodejs server to handle the entire hits download, I can move the logic that handles the above two into the server and import it from the FE to handle the current result download, so this work shouldn't be too much throwaway if we end up going that route

cypress/utils/apps/query_enhancements/export_as_csv.js
```diff
@@ -0,0 +1,161 @@
...
+  return time;
+};
+
+/**
+ * Returns what the first row should be for the downloaded file for file with selected fields
+ * @param {QueryEnhancementLanguageData} language - the language data
+ * @param {boolean} hasTime - whether the dataset has time
+ * @returns {string[]}
+ */
+export const getFirstRowForDownloadWithFields = (language, hasTime) => {
```
d-buckner, line 116: Same as above. I may be misunderstanding though if this is for testing purposes only.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/export_as_csv.spec.js
```diff
@@ -0,0 +1,224 @@
...
+        isEnhancement: true,
+      });
+      cy.createWorkspaceIndexPatterns({
+        workspaceName: workspaceName,
+        indexPattern: INDEX_WITHOUT_TIME_1,
+        timefieldName: '',
+        dataSource: DATASOURCE_NAME,
+        isEnhancement: true,
+        indexPatternHasTimefield: false,
+      });
```
d-buckner, line 72: not for this PR: it looks like this may be a good candidate for a utility as it seems like this may be a common need across tests, is that the case or is this pretty specific?

angle943: yeah this whole chunk can def be abstracted because its copy and pasted often times with minimal deviations. But yeah lets take that work to a different scope

cypress/utils/apps/query_enhancements/export_as_csv.js
```diff
@@ -0,0 +1,161 @@
...
+export const getQueriedCountForLanguage = (language) => {
+  switch (language.name) {
+    case QueryLanguages.DQL.name:
+      return 500;
+    case QueryLanguages.Lucene.name:
+      return 500;
+    case QueryLanguages.SQL.name:
+      return 10;
+    case QueryLanguages.PPL.name:
+      return 1000;
```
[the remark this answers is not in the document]
angle943: sorry i am not sure what you mean? What existing definitions are you referring to? Lets sync offline

cypress/utils/apps/query_enhancements/export_as_csv.js
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * Returns what the header should be for the downloaded file with selected fields
+ * @property {boolean} hasTime - whether the dataset has time
+ * @returns {string[]}
+ */
+export const getHeadersForDownloadWithFields = (hasTime) => {
+  if (hasTime) {
+    return ['timestamp', 'bytes_transferred', 'personal.name'];
+  } else {
+    return ['bytes_transferred', 'personal.name'];
```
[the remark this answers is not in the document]
angle943: i'm confused by this comment too haha, lets sync

cypress/utils/apps/query_enhancements/export_as_csv.js
angle943: Yeah this is purely for testing purposes. I am ensuring that the formatting of the fields and the columns are correct in CSV

cypress/utils/apps/query_enhancements/export_as_csv.js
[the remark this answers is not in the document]
d-buckner: I totally missed that this was a cypress util and not a util for the application itself. Feel free to ignore.

cypress/utils/apps/query_enhancements/export_as_csv.js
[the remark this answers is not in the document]
d-buckner: Same as above, I didn't realize this was solely for your test cases.

d-buckner (approved):

src/plugins/discover/public/application/components/chart/chart.tsx
```diff
@@ -70,18 +74,22 @@ export const DiscoverChart = ({
     [data]
   );
   const [isCollapsed, setIsCollapsed] = useState(false);
+  const histogramIsApplicable = isTimeBased && !!chartData?.values.length;
+  const showHistogram = histogramIsApplicable && (!isEnhancementsEnabled || !isCollapsed);
```
kavilla, line 78: won't using a derived `showHistogram` value (instead of a state variable) cause issues if the `isCollapsed` state changes over time? we'll want to verify that collapsing and expanding the histogram works as expected

src/plugins/discover/public/application/components/chart/chart.tsx
```diff
@@ -19,19 +19,22 @@ import { DiscoverServices } from '../../../build_services';
...
+import { OpenSearchSearchHit } from '../../doc_views/doc_views_types';
+import { DownloadCsvButton } from './download_csv_button';
 
 interface DiscoverChartProps {
   bucketInterval?: TimechartHeaderBucketInterval;
   chartData?: Chart;
   config: IUiSettingsClient;
   data: DataPublicPluginStart;
-  hits: number;
+  hits?: number;
```
kavilla, line 30: If this is an existing prop that was previously required, we should be cautious about making it optional now. 

qq:
1. Is this fixing an existing issue where hits could be undefined?
2. If not, we should keep it as a required prop to maintain strict type safety
3. If yes, we should document why hits can be undefined and ensure all consumers handle the undefined case

Generally, we want to avoid weakening type constraints on existing props unless absolutely necessary. Could you clarify this one for me?

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
kavilla, line 1: are we doing way too much in a button component. most of our components (like our filter buttons, search controls etc) just handle showing stuff and get their behavior passed in through props. but this one's handling all the csv stuff, picking columns, saving files, and showing the button all in one place

wouldn't it make more sense to break this up like our other components? maybe something like:

```typescript
interface DownloadCsvButtonProps {
  onClick: () => void;
  count: number;
  disabled?: boolean;
}

then we can re-use the logic and verify the logic from the component separate from. plus it gives us a single source of truth for the csv logic - not "which component actually handles the formatting?"

File: src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx [hunk 13]
```diff
@@ -0,0 +1,131 @@
...
+  const handleDownloadCsv = (): void => {
+    const csvRowData = rows.map(formatRowsForCsv({ displayedColumnNames, indexPattern }));
+
+    const csvData = Papa.unparse(
+      {
+        fields: displayedColumnNames,
+        data: csvRowData,
+      },
+      {
+        quotes: uiSettings.get('csv:quoteValues', true),
```
kavilla, line 106: for consistency I think we should had a constant and access it like `DOC_HIDE_TIME_COLUMN_SETTING` for these two

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    return formattedValue;
+  });
+};
+
+export const DownloadCsvButton = ({ indexPattern, rows }: DownloadCsvButtonProps) => {
+  const { uiSettings } = getServices();
+
+  const [isShortDots, hideTimeColumn] = useMemo(() => {
```
kavilla, line 86: is this useMemo actually doing anything useful here? looks like we're just getting ui settings values - nothing expensive going on. also noticed we're inconsistent with how we access ui settings in this file - as below we access it directly still

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
```diff
@@ -0,0 +1,131 @@
...
+      uiSettings.get(UI_SETTINGS.SHORT_DOTS_ENABLE),
+      uiSettings.get(DOC_HIDE_TIME_COLUMN_SETTING),
+    ];
+  }, [uiSettings]);
+
+  const displayedColumnNames = useSelector(
+    selectColumnNames({ indexPattern, hideTimeColumn, isShortDots })
+  );
+
+  const handleDownloadCsv = (): void => {
```
kavilla, line 97: - this is smomething im super nitpicky about but feel free to ignore but handleDownloadCsv  within the download csv button component feels redundant.
- this map operation could be cleaner with arrow syntax which makes it easier to test each step and see the data flow. plus we can reuse these pieces elsewhere:
```typescript
const formatData = (rows) => rows.map(formatRowsForCsv({ displayedColumnNames, indexPattern }))
const saveFile = (data) => saveAs(new Blob([data], { type: 'text/csv;charset=utf-8' }))
const handleDownload = () => saveFile(formatData(rows))
```
- why hardcode `opensearch_export` in filename? what if we support other data sources later? do we support other data sources? like s3?
- we're forcing YYYY-MM-DD format but shouldn't we use whatever date format the user has in their ui settings?
- also noticed we're locking this to csv/utf-8 in the blob type - what if we want to support other export formats later? maybe make this more flexible?

what you think about making this more future-proof?

src/plugins/discover/public/application/components/chart/download_csv_button/index.ts
```diff
@@ -0,0 +1,12 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
```
kavilla, line 2: i think wrong license header

Maosaic (approved):

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
```diff
@@ -0,0 +1,131 @@
...
+      size="s"
+      iconType="download"
+      iconSide="left"
+      onClick={handleDownloadCsv}
+      data-test-subj="dscDownloadCsvButton"
+    >
+      <FormattedMessage
+        id="discover.downloadCsv"
+        values={{ count: rows.length }}
+        defaultMessage="Download {count} {count, plural, one {document} other {documents}} as CSV"
```
kavilla, line 127: a few concerns about this download button:

1. Screen Space & Message Length
- Users have historically complained about limited space in this area
- Current message shows document count and mentions CSV specifically
- Could we simplify to just "Download Results" since CSV is currently the only option?
- If we add more formats later, we could switch to a popover pattern then

2. UX Flow Consideration
- Current one-click CSV download is simple but might be worth considering future-proofing
- If we later add multiple formats, users would need to relearn the interaction
- Maybe better to establish the popover pattern now with CSV as the only option?

3. Potential User Confusion
The current layout:
`total hits (2,245,600) ......... Download 500 documents`

This could be problematic because:
- Button is detached from the results table but shows table-specific count
- Having two different numbers (2,245,600 vs 500) on the same line could be confusing
- Users might see the 500 first and think that's all the matched hits from the query and not the page size
- The connection between the 500 and what's in their current results table isn't clear

what do you think about these ux concerns? should we consider moving this closer to the results table or rethinking how we display these numbers?

src/plugins/discover/public/application/components/chart/chart.tsx
angle943: I don't believe so. `isCollapsed` is a state, and so whenever this state updates, it will perform a re-render which will cause lines 77 and 78 to recalculate the boolean values of `histogramIsApplicable` and `showHistogram`. As a matter of fact, every time this component re-renders due to prop change or state change, these two lines will run. This might be some perf degradation due to the recalculation of these values even when unnecessary, but I don't think calculating boolean values is that expensive, and there are further perf ramifications when putting these (or one of them) into a state variable (see below).

If we put it in another state, we would start need to introduce `useEffect` to ensure that these two variables are updated, I think its overkill (and extra perf overhead).

I also verified that the button stays visible during/after collapsing

https://github.com/user-attachments/assets/4661e736-84ce-4e84-a0b8-d926ff14a1dd

src/plugins/discover/public/application/view_components/canvas/index.tsx
```diff
@@ -170,7 +170,7 @@ export default function DiscoverCanvas({ setHeaderActionMenu, history, optionalR
             (fetchState.status === ResultStatus.ERROR && !!rows?.length)) &&
             (isEnhancementsEnabled ? (
               <>
-                <MemoizedDiscoverChartContainer {...fetchState} />
+                <MemoizedDiscoverChartContainer {...fetchState} rows={rows} />
```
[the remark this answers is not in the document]
angle943: Yeah completely agreed, and that was the original location that I proposed (next to the table), but due to the vertical real estate issue of introducing another section in the table that will cause the scrollable area of the table even smaller, UX wanted this button next to the histogram (i have added you in the channel where we talked about this).

And yeah, my change introduces more prop drilling, as well as duplicating the rendering/formatting logic of the headers and cell values of the table. Its not ideal at all, and if I had a lot more time I would move that logic more upwards the React DOM tree so that there can be a single source of truth for it, but that would require a bigger refactoring effort.

I synced with UX and our manager about this, and they said this is something we want done quick. Furthermore, one of my tasks after this is to revamp the table entirely, and I'm hoping I can solve all of that issues of prop drilling & state management when I work on that, so this is moreso a temporary solution here

src/plugins/data/common/index_patterns/index_patterns/format_hit.ts
```diff
@@ -86,19 +87,26 @@ export function formatHitProvider(indexPattern: IndexPattern, defaultFormat: any
     return cache;
   }
 
-  formatHit.formatField = function (hit: Record<string, any>, fieldName: string) {
-    let partials = partialFormattedCache.get(hit);
-    if (partials && partials[fieldName] != null) {
-      return partials[fieldName];
-    }
+  formatHit.formatField = function (
```
kavilla, line 90: could you explain this update and why? we should add tests to these existing files that haven't changed in a while.

src/plugins/discover/public/application/components/chart/chart.tsx
angle943: Yeah, the answer to your questions lies in the fact that we have now placed the "Export" button next to the histogram (and hence, within this `chart.tsx` component in question).

Before my change, this entire component would not render if `!hits` due to this: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/discover/public/application/view_components/canvas/discover_chart_container.tsx#L37

My change gets rid of this code because I specifically want the `export` button rendered even when `!hits`.

So to answer your questions:

1. No, not fixing an existing issue, because previously we disallowed `!hits` and component wouldn't render without it
2.  N/A due to the fact that i need to handle `?` case
3. Yeah I can document that on a followup commit

I completely agree on maintaining strict type safety and ensuring they are valid

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
angle943: I completely agree that this component is doing a lot, but the main rationale lies in what I stated previously, namely: this is a temp effort that will most likely get revamped

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.test.tsx
```diff
@@ -0,0 +1,261 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
angle943: Oops will fix

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
angle943: do you mean the `csv:quoteValues` string?

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
angle943: I don't think its doing anything useful :D , i'll get rid of.

FYI i did it because its from this code: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/discover/public/application/components/default_discover_table/default_discover_table.tsx#L63-L70

src/plugins/discover/public/application/components/chart/download_csv_button/index.ts
angle943: oops i will update. still not 100% sure what the correct header is, is this it?

```
/*
 *
 * The OpenSearch Contributors require contributions made to
 * this file be licensed under the Apache-2.0 license or a
 * compatible open source license.
 *
 * Any modifications Copyright OpenSearch Contributors. See
 * GitHub history for details.
 */
```

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
angle943: I think those are very valid concerns :D, especially concern #3 . I brought it up a few times but this was the design we settled on

src/plugins/data/common/index_patterns/index_patterns/format_hit.ts
angle943: Yeah I can explain. This function is used to format the cell values into a readable format. This function calls `convert` method, which already has the optional parameter of `type`. which can be `html` or `text`. If `html`, it will add some HTML spice to it (such as wrapping it around `<span>`). If `text`, it will just format the string (which is what i need for export CSV).

The `formatField` was written in a way where only `html` formatted type can be outputted, which is annoying for my usecase since it will put html tags around every cell values. So I updated it to support `text` type.

There is one caveat that we do not want to leverage caching for `text` type (the cache only works for HTML type). This logic already exists in the `convert` method (I just added a comment there to make it more clear), but I also had to add that to this method.

and yeah i can add tests, but lets sync offline as I got a couple questions

src/plugins/discover/public/application/components/chart/download_csv_button/index.ts
ashwin-pc: Let es lint auto apply that. If your editor is setup correctly, it should happen automatically

ashwin-pc: Nit: Also we should make this a CI step

src/plugins/discover/public/application/components/chart/download_csv_button/download_csv_button.tsx
angle943: yeah i can handle something like this. My main reason for putting the logic specifically in this file (and scoped to CSV) was I didn't know where to put a generic utility function :P (I don't know if you remember, but some weeks ago I was asking you where I can put generic utilities)

angle943: Let me address each of these actually:

- yeah `handleDownloadCsv` is redundant, i will change in my future PR
- I willtry to implement something like what you described.
- `opensearch_export` I think is not to signify that it is an `opensearch` dataset, but rather it is from the OSD platform. Even if you are interacting with the S3 data source, and you click export, you are exporting it from OSD. So when a user sees the download in their computer, they will know they got it from OSD. But i'm up to debate this.
- The only concern with using the UI determined settings is that there are reserved keywords in file names, and so I am not sure if we want to deal with the complexities that comes with resolving those differences. This is again up for debate but I am unsure if we want to invest the time to resolve all those differences for how the file is going to be saved, which can freely be edited by the user easily.
- I think i will make a plugin dedicated to how to handle `files` in that the whole app can use, and refactor the existing areas where we use `download` functionality to use that new utility. I'll discuss this w/ the team when everyone is around

PR #9340, opened 2025-02-06
[Workspace]bugfix: remove workspace prefix when calling `get` in recent assets

SuZhou-Joe:
### Description

Recent assets is a global component sitting in page header, and it will call get to retrieve metadata of objects inside / outside workspace. As we introduce a more strict rule that will deny cross-workspace calls, we need to remove the workspace prefix inside the calls from recent assets.

### Before fix

image

### After fix

image

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/public/chrome/ui/header/recent_items.tsx
+++ b/src/core/public/chrome/ui/header/recent_items.tsx
@@ -73,7 +73,12 @@ const bulkGetDetail = (savedObjects: Array<Pick<SavedObject, 'type' | 'id'>>, ht
         .get<SavedObjectWithMetadata>(
           `/api/opensearch-dashboards/management/saved_objects/${encodeURIComponent(
             obj.type
-          )}/${encodeURIComponent(obj.id)}`
+          )}/${encodeURIComponent(obj.id)}`,
+          {
+            prependOptions: {
+              withoutClientBasePath: true,
+            },
+          }
         )
         .catch((error) => ({
           id: obj.id,


wanglam (approved):

src/core/public/chrome/ui/header/recent_items.tsx
```diff
@@ -73,7 +73,12 @@ const bulkGetDetail = (savedObjects: Array<Pick<SavedObject, 'type' | 'id'>>, ht
         .get<SavedObjectWithMetadata>(
           `/api/opensearch-dashboards/management/saved_objects/${encodeURIComponent(
             obj.type
-          )}/${encodeURIComponent(obj.id)}`
+          )}/${encodeURIComponent(obj.id)}`,
```
ruanyl, line 76: Should it filter out objects that's not belong to the current workspace before bulkGetDetails?

Hailong-am (approved):

src/core/public/chrome/ui/header/recent_items.tsx
SuZhou-Joe: According to previous decision / discussion, recent assets is a global concept so no matter in which workspace we will still show recent assets from other workspace.

ruanyl (approved):

SuZhou-Joe:
We will keep the behavior as it is today to make recent assets global, if we think it should be workspace specific, we can raise a following PR to change the behavior.

PR #9343, opened 2025-02-06
Change from `Cluster` to `Data sources` for Indexes type in dataset selector column

abbyhu2000:
### Description

Discover data selector does not distinguish between clusters and collections, so we should call it data sources instead of just clusters.

## Screenshot

Screenshot 2025-02-06 at 12 39 06 PM

## Changelog
- fix: Change from cluster to data sources for dataset selector column

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
@@ -20,9 +20,7 @@ export const DELIMITER = '::';
 
 export const indexTypeConfig: DatasetTypeConfig = {
   id: DEFAULT_DATA.SET_TYPES.INDEX,
-  title: i18n.translate('data.datasetType.indexes', {
-    defaultMessage: 'Indexes',
-  }),
+  title: 'Indexes',
   meta: {
     icon: { type: 'logoOpenSearch' },
     tooltip: 'OpenSearch Indexes',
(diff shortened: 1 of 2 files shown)


src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
```diff
@@ -68,7 +68,7 @@ export const indexTypeConfig: DatasetTypeConfig = {
         const dataSources = await fetchDataSources(services.savedObjects.client);
         return {
           ...dataStructure,
-          columnHeader: 'Clusters',
+          columnHeader: 'Data Sources',
```
Maosaic, line 71: Any i18n resource we need to update?

kavilla: good call. i can't remember if i created an issue but that is an improvement we need to do.

abbyhu2000: added i18n for the column headers, but i think there are more components in the dataset selector that needs i18n

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
```diff
@@ -20,7 +20,9 @@ export const DELIMITER = '::';
 
 export const indexTypeConfig: DatasetTypeConfig = {
   id: DEFAULT_DATA.SET_TYPES.INDEX,
-  title: 'Indexes',
+  title: i18n.translate('data.datasetType.indexes', {
```
kavilla, line 23: is this the right key? i think the ci now verifies for us (i think miki set it up) but can you verify this doesn't fail the i18n generation script?

abbyhu2000:
Removed i18n in this PR since we are tracking the lack of i18n in dataset selector as a whole

kavilla (approved):

sejli (approved):

ananzh (approved):

Issue #9318, opened 2025-01-31
[BUG] [2.15]: Nested/Object Field Arrays Display Incorrectly


rayangou:
**Describe the bug**

In OpenSearch Dashboards, the first object in a nested field array incorrectly displays all array values from adjacent objects. This issue does not occur when querying OpenSearch directly.

**To Reproduce**

1. Create an index test-nested-array with the following mapping:
```
"nested_test1": {
  "type": "nested",
  "properties": {
    "d_values": {
      "type": "double"
    },
    "s_entry": {
      "type": "keyword"
    }
  }
}
```
2. Add documents like the one below:
```
{
  "nested_test1": [
    {
      "d_values": [0.1, 0.2],
      "s_entry": "4"
    },
    {
      "d_values": [0.3, 0.4],
      "s_entry": "5"
    },
    {
      "d_values": [0.5, 0.6],
      "s_entry": "6"
    }
  ]
}

```
3. Go to OpenSearch Dashboards and create an index pattern for the test-nested-array index.
4. Expand the record and click on the JSON view.
5. Observe that the first object in the nested_test1 array displays all the array values of its adjacent objects.
6. Example of the observed behavior:
```
"nested_test1": [
  {
    "d_values": [
      [0.1, 0.2],
      [0.3, 0.4],
      [0.5, 0.6]
    ],
    "s_entry": "1"
  },
  {
    "d_values": [0.3, 0.4],
    "s_entry": "2"
  },
  {
    "d_values": [0.5, 0.6],
    "s_entry": "3"
  }
]
```

**Expected behavior**
The first object in the nested_test1 array should not include values from adjacent objects. The expected output is:

```
"nested_test1": [
  {
    "d_values": [0.1, 0.2],
    "s_entry": "1"
  },
  {
    "d_values": [0.3, 0.4],
    "s_entry": "2"
  },
  {
    "d_values": [0.5, 0.6],
    "s_entry": "3"
  }
]
```

**OpenSearch Version**
2.15

**Dashboards Version**
2.15

**Plugins**

Security, Alerting, 

**Screenshots**

![Image](https://github.com/user-attachments/assets/d233f345-0006-46d3-af0f-7f9cb1809fb6)

**Host/Environment (please complete the following information):**
 - OS: CentOS
 - Browser and version Chrome

**Additional context**

This issue appears to be specific to OpenSearch Dashboards handling of nested/Object array fields in the JSON view.

[...]

PR #9347, opened 2025-02-07
Fix flatten hit modify original array

wanglam:
### Description

This PR addresses an issue with the `flattenHit` function modifying the original hit object array. Currently, when the `flattenHit` function encounters an array value within the hit object, it flattens the array in-place, directly modifying the original hit object.

The proposed changes introduce a check to determine if the value is an array. If it is, a new array is constructed with the flattened values, preserving the original hit object array. This ensures that the `flattenHit` function no longer mutates the original hit object, preventing potential side effects or unintended modifications.

### Issues Resolved

#9318 

## Screenshot

image

## Testing the changes

Following the reproduce steps in #9318 

## Changelog

- fix: flatten hit modify original array

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/common/index_patterns/index_patterns/flatten_hit.ts
```diff
@@ -62,7 +62,7 @@ function flattenHit(indexPattern: IndexPattern, hit: Record<string, any>, deep:
 
       if (hasValidMapping || isValue) {
         if (!flat[key]) {
-          flat[key] = val;
+          flat[key] = Array.isArray(val) ? val.concat([]) : val;
```
Maosaic, line 65: Nit: imo it's more readable by using the spread operator for cloning, or maybe add some context as comment.

Maosaic (approved):

src/plugins/data/common/index_patterns/index_patterns/flatten_hit.ts
d-buckner: Just for awareness (feel free to ignore if this is known): The `Array.prototype.concat` API creates a shallow copy. If there are there modifications that are made in the array item objects themselves, they will impact the original `hits` object given they share references.

d-buckner (approved):

src/plugins/data/common/index_patterns/index_patterns/flatten_hit.ts
wanglam: Thanks for catching that. I will refactor with the spread operator instead of `Array.prototype.concat`. Do you think we should use a deep copy here? This PR is for fixing the issue of the original hit being modified inside this `flattenHit` function. The `flat[key].push` in line 67 will push elements to the original array. If we want to do a deep copy, we may need to change lines 67 and 69 too.

d-buckner: @wanglam If the shallow copy resolves the issue then I think it makes sense to avoid deep cloning as it's relatively expensive. If in the future we see cases where the items are modified as well we can always move to using deep cloning then. Thanks for taking the time to fix.

ruanyl (approved):

d-buckner (approved):

ashwin-pc (approved):

PR #8765, opened 2024-10-30
[Discover] fix error handling for sql and jobs APIs

joshuali925:
### Description

This is a followup PR for https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8743 and https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8725. There are more APIs not correctly handling errors and throws the default 500 internal server error code, and this PR fixes `/api/enhancements/jobs` and SQL search strategy to throw the correct error

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #9355, opened 2025-02-07
[Discover] Cancel S3 Queries Using SQL Plugin

sejli:
### Description

When ongoing S3 queries are aborted in Discover, the backend job should also be canceled/deleted. This PR uses the SQL plugin API to delete ongoing jobs to cancel them.

## Changelog

- feat: Deletes S3 Jobs in Backend when Original Query is Canceled

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/s3_dataset.spec.js
+++ b/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/s3_dataset.spec.js
@@ -7,6 +7,7 @@ import {
   DS_API,
   DSM_API,
   S3_CLUSTER,
+  DELETE_API,
 } from '../../../../../utils/apps/query_enhancements/constants';
 import { getRandomizedWorkspaceName } from '../../../../../utils/apps/query_enhancements/shared';
 import { prepareTestSuite } from '../../../../../utils/helpers';
@@ -90,6 +91,11 @@ const s3DatasetTestSuite = () => {
           cy.deleteWorkspaceByName(workspace);
           cy.visit('/app/home');
           cy.osd.createInitialWorkspaceWithDataSource(S3_CLUSTER.name, workspace);
+          cy.navigateToWorkSpaceSpecificPage({
+            workspaceName: workspace,
+            page: 'discover',
+            isEnhancement: true,
+          });
         });
         afterEach(() => {
           cy.deleteWorkspaceByName(workspace);
@@ -137,6 +143,40 @@ const s3DatasetTestSuite = () => {
           cy.waitForSearch();
 
           cy.getElementByTestId('queryEditorLanguageSelector').should('contain', 'PPL');
+          cy.get(`[data-test-subj="queryResultCompleteMsg"]`).should('be.visible');
+          cy.getElementByTestId('docTable').should('be.visible');
+          cy.getElementByTestId('docTable').find('tr').should('have.length', 11);
+        });
+
+        it('aborts and cancels previous query when new query is started', function () {
+          cy.getElementByTestId(`datasetSelectorButton`).click();
+          cy.getElementByTestId(`datasetSelectorAdvancedButton`).click();
+
+          cy.get(`[title="S3 Connections"]`).click();
+          cy.get(`[title="BasicS3Connection"]`).click();
+          cy.get(`[title="mys3"]`).click();
+          cy.get(`[title="default"]`).click();
+          cy.get(`[title="http_logs"]`).click();
+          cy.getElementByTestId('datasetSelectorNext').click();
+          cy.get(`[class="euiModalHeader__title"]`).should('contain', 'Step 2: Configure data');
+
+          cy.getElementByTestId('advancedSelectorLanguageSelect').select('OpenSearch SQL');
+          cy.getElementByTestId('advancedSelectorConfirmButton').click();
+
+          // Need to wait a bit for initial query to start
+          cy.wait(3000);
+
+          cy.intercept('DELETE', `**/${DELETE_API}*`).as('cancelRequest');
+          cy.getElementByTestId(`querySubmitButton`).click();
+
+          cy.wait('@cancelRequest').then((interception) => {
+            console.log('interception.request.url:', interception.request.url);
+            // Verify the request had the correct query parameters
+            expect(interception.request.url).to.include('queryId=');
+            expect(interception.request.url).to.include('id=');
+          });
+
+          cy.waitForSearch();
           cy.get(`[data-test-subj="queryResultCompleteMsg"]`).should('be.visible');
           cy.getElementByTestId('docTable').should('be.visible');
           cy.getElementByTestId('docTable').find('tr').should('have.length', 11);
--- a/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/s3_dataset.spec.js
+++ b/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/s3_dataset.spec.js
@@ -7,7 +7,7 @@ import {
   DS_API,
   DSM_API,
   S3_CLUSTER,
-  DELETE_API,
+  JOBS_API,
 } from '../../../../../utils/apps/query_enhancements/constants';
 import { getRandomizedWorkspaceName } from '../../../../../utils/apps/query_enhancements/shared';
 import { prepareTestSuite } from '../../../../../utils/helpers';
@@ -166,7 +166,7 @@ const s3DatasetTestSuite = () => {
           // Need to wait a bit for initial query to start
           cy.wait(3000);
 
-          cy.intercept('DELETE', `**/${DELETE_API}*`).as('cancelRequest');
+          cy.intercept('DELETE', `**/${JOBS_API.DELETE}*`).as('cancelRequest');
           cy.getElementByTestId(`querySubmitButton`).click();
 
           cy.wait('@cancelRequest').then((interception) => {
--- a/cypress/utils/apps/query_enhancements/constants.js
+++ b/cypress/utils/apps/query_enhancements/constants.js
@@ -21,6 +21,8 @@ export const DS_API = {
 };
 export const DSM_API = '/internal/data-source-management/fetchDataSourceMetaData';
 
+export const DELETE_API = '/api/enhancements/jobs';
+
 export const INDEX_WITH_TIME_1 = 'data_logs_small_time_1';
 export const INDEX_WITHOUT_TIME_1 = 'data_logs_small_no_time_1';
 export const INDEX_WITH_TIME_2 = 'data_logs_small_time_2';
--- a/cypress/utils/apps/query_enhancements/constants.js
+++ b/cypress/utils/apps/query_enhancements/constants.js
@@ -21,7 +21,10 @@ export const DS_API = {
 };
 export const DSM_API = '/internal/data-source-management/fetchDataSourceMetaData';
 
-export const DELETE_API = '/api/enhancements/jobs';
+export const BASE_QUERY_ENHANCEMENTS_API = '/api/enhancements';
+export const JOBS_API = {
+  DELETE: `${BASE_QUERY_ENHANCEMENTS_API}/jobs`,
+};
 
 export const INDEX_WITH_TIME_1 = 'data_logs_small_time_1';
 export const INDEX_WITHOUT_TIME_1 = 'data_logs_small_no_time_1';
--- a/src/plugins/query_enhancements/common/utils.ts
+++ b/src/plugins/query_enhancements/common/utils.ts
@@ -73,7 +73,7 @@ export const fetch = (context: EnhancedFetchContext, query: Query, aggConfig?: Q
           try {
             await http.fetch({
               method: 'DELETE',
-              path: trimEnd(API.DATA_SOURCE.ASYNC_JOBS),
+              path: API.DATA_SOURCE.ASYNC_JOBS,
               query: {
                 id: query.dataset?.dataSource?.id,
                 queryId: context.body?.pollQueryResultsParams.queryId,
--- a/src/plugins/query_enhancements/common/utils.ts
+++ b/src/plugins/query_enhancements/common/utils.ts
@@ -13,6 +13,7 @@ import {
   QueryStatusConfig,
   QueryStatusOptions,
 } from './types';
+import { API } from './constants';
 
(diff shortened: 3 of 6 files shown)


cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/s3_dataset.spec.js
kavilla, line 1: javascript file :(

cypress/utils/apps/query_enhancements/constants.js
```diff
@@ -21,6 +21,8 @@ export const DS_API = {
 };
 export const DSM_API = '/internal/data-source-management/fetchDataSourceMetaData';
 
+export const DELETE_API = '/api/enhancements/jobs';
```
kavilla, line 24: unrelated to this PR but not a big fan of the multiple constants defined instead of using nested constants for similar things like APIs, when issue is the number of modifications increasing as I will have to import each constant individually.

kavilla:
i think looks good to me. can we confirm this is wired up ? in the places we expect. or did we have fast follows to add that. as in abort query on navigation away from discover. abort query if closing the dataset selector. abort query if new query is fired (pretty sure already done)

kavilla (approved):

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/s3_dataset.spec.js
sejli: Looks like a lot of our Cypress tests are JavaScript files, is that something we want to change?

sejli:
So far, it's wired up to the point where the delete request will be sent when the query enhancement's [`fetch()` function](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/query_enhancements/common/utils.ts#L52) is aborted. So, entirely dependent on how this fetch function picks up the `AbortSignal`. 

> abort query if new query is fired

This is the main thing that this PR acheives.

> abort query if closing the dataset selector.

Do we want this? If we start loading something like databases and close the dataset selector, wouldn't we want that to load in the background? I think it's intended behavior.

> as in abort query on navigation away from discover

Think this will probably need a fast followup or some improvements on how I'm doing this.

cypress/utils/apps/query_enhancements/constants.js
joshuali925: +1 to one const, and ideally API paths are resources not actions, it should not be `DELETE_API` but something like `JOBS_API` if you want to name this specific path

src/plugins/query_enhancements/common/utils.ts
```diff
@@ -57,13 +58,34 @@ export const fetch = (context: EnhancedFetchContext, query: Query, aggConfig?: Q
...
-    })
+    http
+      .fetch({
+        method: 'POST',
+        path,
+        body,
+        signal,
+      })
+      .catch(async (error) => {
+        if (error.name === 'AbortError' && context.body?.pollQueryResultsParams?.queryId) {
```
joshuali925, line 71: i can't remember but does the current UI abort requests when user clicks Run query before the previous one finishes?

For example, 

1. user starts async query
2. query id returned
3. browser sends requests to check job status
4. server returns job status
5. browser waits a few seconds if job is pending
6. step 3-5 repeats until query finishes

If the request in step 3 is aborted, then this would work. But if user runs another query during step 5, would this abort and delete job for the previous query?

src/plugins/query_enhancements/server/utils/plugins.ts
```diff
@@ -141,7 +141,7 @@ export const OpenSearchEnhancements = (client: any, config: any, components: any
   });
 
   enhancements.deleteJob = createAction(client, components, {
-    endpoint: `${URI.ASYNC_QUERY}/<%=queryId%>`,
+    endpoint: `${URI.ASYNC_QUERY}`,
```
joshuali925, line 144: does this change break anything?

src/plugins/query_enhancements/server/routes/data_source_connection/routes.ts
```diff
@@ -99,4 +99,31 @@ export function registerDataSourceConnectionsRoutes(
       }
     }
   );
+
+  router.delete(
+    {
+      path: `${API.DATA_SOURCE.ASYNC_JOBS}`,
```
joshuali925, line 105: nit. 
```suggestion
      path: API.DATA_SOURCE.ASYNC_JOBS,
```

src/plugins/query_enhancements/server/routes/data_source_connection/routes.ts
```diff
@@ -99,4 +99,31 @@ export function registerDataSourceConnectionsRoutes(
...
+      validate: {
+        query: schema.object({
+          id: schema.string(),
```
joshuali925, line 108: nit. is it conventional to use `id` for datasource id? can this be `datasourceId` to avoid confusion since there's also `queryId` below?

src/plugins/query_enhancements/common/utils.ts
```diff
@@ -57,13 +58,34 @@ export const fetch = (context: EnhancedFetchContext, query: Query, aggConfig?: Q
...
+              method: 'DELETE',
+              path: API.DATA_SOURCE.ASYNC_JOBS,
+              query: {
+                id: query.dataset?.dataSource?.id,
+                queryId: context.body?.pollQueryResultsParams.queryId,
+              },
+            });
+          } catch (cancelError) {
+            // eslint-disable-next-line no-console
+            console.error('Failed to cancel query:', cancelError);
```
joshuali925, line 84: should it throw the cancel error?

joshuali925 (commented):
do we also want to add a cancel button on UI?

src/plugins/query_enhancements/server/routes/data_source_connection/routes.ts
```diff
@@ -99,4 +99,31 @@ export function registerDataSourceConnectionsRoutes(
...
+        const client = request.query.id
+          ? context.dataSource.opensearch.legacy.getClient(request.query.id).callAPI
+          : defaultClient.asScoped(request).callAsCurrentUser;
+
+        await client('enhancements.deleteJob', {
+          queryId: request.query.queryId,
+        });
+        return response.noContent();
+      } catch (error) {
+        const statusCode = error.statusCode === 500 ? 503 : error.statusCode || 503;
```
ruchidh, line 124: Can you describe more or may be add comment what exactly the action being taken with these error codes, 

also can we write it like to make it more readable?
const statusCode = error.statusCode && error.statusCode !== 500 ? error.statusCode : 503;

src/plugins/query_enhancements/common/utils.ts
```diff
@@ -57,13 +58,34 @@ export const fetch = (context: EnhancedFetchContext, query: Query, aggConfig?: Q
...
+        body,
+        signal,
+      })
+      .catch(async (error) => {
+        if (error.name === 'AbortError' && context.body?.pollQueryResultsParams?.queryId) {
+          // Cancel job
+          try {
+            await http.fetch({
+              method: 'DELETE',
+              path: API.DATA_SOURCE.ASYNC_JOBS,
```
joshuali925, line 76: should this allow each dataset type config to define the action to delete query job? there might be other type configs using `pollQueryResultsParams` but does not use SQL plugin async jobs

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/s3_dataset.spec.js
angle943: its something i def want to change, but we can do it later

cypress/utils/apps/query_enhancements/constants.js
sejli: Updated in latest commit

src/plugins/query_enhancements/common/utils.ts
sejli: In this scenario, yes, it will abort if another query is run in between the short polling requests.

src/plugins/query_enhancements/server/utils/plugins.ts
sejli: I don't think so, there shouldn't be anything else using this currently.

src/plugins/query_enhancements/server/routes/data_source_connection/routes.ts
sejli: Was just following the rest of the file to add in parameters, but I agree that the rest of query enhancements routes + APIs should update this. This would be out of scope for this PR though, and should be a followup item.

src/plugins/query_enhancements/common/utils.ts
sejli: Don't think it's necessary, since we don't want this thrown error to stop any other process.

src/plugins/query_enhancements/common/utils.ts
sejli: One of our followup items to discuss is how we will add the cancel button, which is why I haven't included it in this PR. However, adding a dataset type config would be a good way to do that implementation, thanks for that tip.

src/plugins/query_enhancements/server/routes/data_source_connection/routes.ts
sejli: This was added as a way to handle specific errors from SQL plugin. Was mimicking the other routes, which were added in [this PR](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8765/files#diff-aead00fbb107e2a45bef16b020de70d3e47324a3bed4c4186e4eaccbc11c098fR97).

src/plugins/query_enhancements/common/utils.ts
joshuali925: could you explain a bit why would it abort? i thought during waiting time, the last `fetch` promise will be resolved and `catch` won't trigger?

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/s3_dataset.spec.js
```diff
@@ -141,6 +147,40 @@ const s3DatasetTestSuite = () => {
...
+          cy.getElementByTestId(`datasetSelectorButton`).click();
+          cy.getElementByTestId(`datasetSelectorAdvancedButton`).click();
+
+          cy.get(`[title="S3 Connections"]`).click();
+          cy.get(`[title="BasicS3Connection"]`).click();
+          cy.get(`[title="mys3"]`).click();
+          cy.get(`[title="default"]`).click();
+          cy.get(`[title="http_logs"]`).click();
+          cy.getElementByTestId('datasetSelectorNext').click();
+          cy.get(`[class="euiModalHeader__title"]`).should('contain', 'Step 2: Configure data');
```
kavilla, line 161: nit: given that this test just verify if it aborts  verifyhing if selecting the title has this increases potential points of failure unrelated to this test.

in this test it would be better to just navigate to the page with a s3 query run it and then run it again focusing on the test name.

but this should be addressed later not now.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/s3_dataset.spec.js
```diff
@@ -141,6 +147,40 @@ const s3DatasetTestSuite = () => {
...
+
+          cy.getElementByTestId('advancedSelectorLanguageSelect').select('OpenSearch SQL');
+          cy.getElementByTestId('advancedSelectorConfirmButton').click();
+
+          // Need to wait a bit for initial query to start
+          cy.wait(3000);
```
kavilla, line 167: how come? not blocker but something we should address is avoiding abritraty wait times like a env variable that we can control globally. step 2 would be using other features that help us wait without waiting and assuming everything will happen within 3 secs

like i said tho we shouldn't address this in this pr

src/plugins/query_enhancements/common/utils.ts
joshuali925: but should we let user know if it failed? For example if I accidentally ran a `select *` on all the data, it might be very expensive. If I started another query or clicked cancel and I don't see any warning on UI, i would assume i won't be fully charged for the previous query but in fact it might still be running

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/s3_dataset.spec.js
```diff
@@ -141,6 +147,40 @@ const s3DatasetTestSuite = () => {
...
+
+          cy.intercept('DELETE', `**/${JOBS_API.DELETE}*`).as('cancelRequest');
+          cy.getElementByTestId(`querySubmitButton`).click();
+
+          cy.wait('@cancelRequest').then((interception) => {
+            console.log('interception.request.url:', interception.request.url);
```
kavilla, line 173: did we want to console log?

src/plugins/query_enhancements/common/utils.ts
joshuali925: got it, it might be better if S3 type config provides this catch logic as an error handler

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/s3_dataset.spec.js
```diff
@@ -141,6 +147,40 @@ const s3DatasetTestSuite = () => {
...
+            // Verify the request had the correct query parameters
+            expect(interception.request.url).to.include('queryId=');
+            expect(interception.request.url).to.include('id=');
+          });
+
+          cy.waitForSearch();
+          cy.get(`[data-test-subj="queryResultCompleteMsg"]`).should('be.visible');
+          cy.getElementByTestId('docTable').should('be.visible');
+          cy.getElementByTestId('docTable').find('tr').should('have.length', 11);
```
kavilla, line 182: could we potentially luck out on this being 11? for no matter what table.

would it be better if we modify the query  to be something we know like `SELECT * FROM foo LIMIT 10` then have it be like `LIMIT 20` and then we make sure it is still 10?

kavilla (approved):

joshuali925 (approved):
my main concerns are https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9355#discussion_r1962322183 and extensibility of this implementation, but i don't think they are blocking this PR

PR #9356, opened 2025-02-07
Query requests return with sample size defined in advanced setting

abbyhu2000:
### Description
The size of the SQL and PPL responses should be determined by advanced setting `discover:sampleSize`. Default value is 500.

## SQL: 
Default
```
SELECT * FROM dataset
```
* we use fetch_size param when sending SQL request to define how many hits we should get back: 

## PPL: 
Default: 
```
source = dataset | head ${advanced setting sampleSize value, default is 500}
```
* fetch_size in sending PPL is currently broken due to https://github.com/opensearch-project/sql/issues/3314; until it get resolved, we interpolate the default PPL query to add a sample size for now
* once the bug is resolved, we should change to use fetch_size just like how SQL does it. Created an issue to track this: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9385

## async SQL/PPL:
Default: 
```
SELECT * FROM dataset LIMIT ${advanced setting sampleSize value, default is 500}
source = dataset | head ${advanced setting sampleSize value, default is 500}
```
* fetch_size currently do not support for async queries; we interpolate the default queries to add a sample size

## Screenshot

https://github.com/user-attachments/assets/492440d2-184d-459a-80af-698cf011509d

## Changelog

- feat: Remove sample size in default query string but use advanced setting

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/dataset_service/types.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/types.ts
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import { EuiIconProps } from '@elastic/eui';
+import { CoreStart } from 'opensearch-dashboards/public';
 import {
   Dataset,
   DatasetField,
@@ -117,5 +118,5 @@ export interface DatasetTypeConfig {
   /**
    * Returns the initial query that is added to the query editor when a dataset is selected.
    */
-  getInitialQueryString?: (query: Query) => string | void;
+  getInitialQueryString?: (query: Query, uiSettings?: CoreStart['uiSettings']) => string | void;
 }
--- a/src/plugins/query_enhancements/public/datasets/s3_type.ts
+++ b/src/plugins/query_enhancements/public/datasets/s3_type.ts
@@ -5,7 +5,7 @@
 
 import { i18n } from '@osd/i18n';
 import { trimEnd } from 'lodash';
-import { HttpSetup, SavedObjectsClientContract } from 'opensearch-dashboards/public';
+import { CoreStart, HttpSetup, SavedObjectsClientContract } from 'opensearch-dashboards/public';
 import semver from 'semver';
 import {
   DATA_STRUCTURE_META_TYPES,
@@ -15,6 +15,7 @@ import {
   DataStructureCustomMeta,
   Dataset,
   DatasetField,
+  Query,
 } from '../../../data/common';
 import { DatasetTypeConfig, IDataPluginServices, OSD_FIELD_TYPES } from '../../../data/public';
 import {
@@ -131,7 +132,7 @@ export const s3TypeConfig: DatasetTypeConfig = {
             title: i18n.translate('queryEnhancements.s3Type.sampleQuery.basicPPLQuery', {
               defaultMessage: 'Sample query for PPL',
             }),
-            query: `source = ${dataset?.title}`,
+            query: `source = ${dataset.title} | head 500`,
           },
         ];
       case 'SQL':
@@ -140,9 +141,22 @@ export const s3TypeConfig: DatasetTypeConfig = {
             title: i18n.translate('queryEnhancements.s3Type.sampleQuery.basicSQLQuery', {
               defaultMessage: 'Sample query for SQL',
             }),
-            query: `SELECT * FROM ${dataset?.title} LIMIT 10`,
+            query: `SELECT * FROM ${dataset.title} LIMIT 500`,
           },
         ];
+    }
+  },
+
+  getInitialQueryString: (query: Query, uiSettings?: CoreStart['uiSettings']) => {
+    switch (query.language) {
+      case 'PPL':
+        return `source = ${query.dataset?.title} | head ${
+          uiSettings?.get('discover:sampleSize') ?? 500
+        }`;
+      case 'SQL':
+        return `SELECT * FROM ${query.dataset?.title} LIMIT ${
+          uiSettings?.get('discover:sampleSize') ?? 500
+        }`;
     }
   },
 };
--- a/src/plugins/query_enhancements/server/search/ppl_search_strategy.ts
+++ b/src/plugins/query_enhancements/server/search/ppl_search_strategy.ts
@@ -35,6 +35,9 @@ export const pplSearchStrategyProvider = (
   return {
     search: async (context, request: any, options) => {
       try {
+        // TODO: PPL does not support fetch_size yet due to https://github.com/opensearch-project/sql/issues/3314
+        // We can add the below back for https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9385
+        // request.body.fetch_size = await context.core.uiSettings.client.get('discover:sampleSize');
         const query: Query = request.body.query;
         const aggConfig: QueryAggConfig | undefined = request.body.aggConfig;
         const rawResponse: any = await pplFacet.describeQuery(context, request);
--- a/src/plugins/query_enhancements/server/search/sql_search_strategy.ts
+++ b/src/plugins/query_enhancements/server/search/sql_search_strategy.ts
@@ -34,7 +34,12 @@ export const sqlSearchStrategyProvider = (
     search: async (context, request: any, options) => {
       try {
         const query: Query = request.body.query;
-        const rawResponse: any = await sqlFacet.describeQuery(context, request);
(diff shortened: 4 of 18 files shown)


src/plugins/query_enhancements/public/datasets/s3_type.ts
```diff
@@ -140,7 +140,7 @@ export const s3TypeConfig: DatasetTypeConfig = {
             title: i18n.translate('queryEnhancements.s3Type.sampleQuery.basicSQLQuery', {
               defaultMessage: 'Sample query for SQL',
             }),
-            query: `SELECT * FROM ${dataset.title} LIMIT 10`,
```
abbyhu2000, line 143: should add back for async query

abbyhu2000: added back by using `getInitialQueryString` in s3 dataset type

src/plugins/query_enhancements/server/utils/facet.ts
```diff
@@ -43,6 +43,7 @@ export class Facet {
       const { format, lang } = request.body;
       const params = {
         body: {
+          fetch_size: request.body.fetch_size,
```
kavilla, line 46: if we have the context still here are we not able to just access advanced settings in this file and set the param? or was there a technical blocker that we should specify for each strategy the size if set or not.

and should we check if this value is set or not if it is then we pass it? what are implications of passing undefined to this the API? because right now i believe it will fire a request like

```
{
  body: {
    fetch_size: undefined,
    query: query.query,
    // ... other properties
  },
  // ... format property if applicable
}
```

will it be ignored?

src/plugins/query_enhancements/server/search/ppl_search_strategy.ts
```diff
@@ -35,6 +35,7 @@ export const pplSearchStrategyProvider = (
   return {
     search: async (context, request: any, options) => {
       try {
+        request.body.fetch_size = await context.core.uiSettings.client.get('discover:sampleSize');
```
kavilla, line 38: this is my bad - I get why you did it this way.

After thinking about it more, pulling from discover's advanced settings on the server side probably isn't the right move. If we want dashboards to have query editor + ppl/sql support down the road, we shouldn't make everything use discover's sample size setting.

however - until we deprecate these search strategies which is something we need to do and move to the right ones (based on data type + proper internal routes), we might be stuck. Could you check if we're already passing size params in the request/options somewhere? If we are, we should probably use that instead.

If we're not though, trying to refactor this now might be more trouble than it's worth, since we know we need to clean it up later anyway. Might make more sense to keep it on the query enhancements side for now and just add a TODO and link whatever issue we open in the code for follow up.

Let me know what you find re: the request params and we can figure out next steps.

src/plugins/query_enhancements/server/search/sql_search_strategy.ts
```diff
@@ -33,6 +33,7 @@ export const sqlSearchStrategyProvider = (
   return {
     search: async (context, request: any, options) => {
       try {
+        request.body.fetch_size = await context.core.uiSettings.client.get('discover:sampleSize');
```
kavilla, line 36: nit: instead of modifying objects directly, we should try to spread them like:
const newObj = { ...oldObj, newStuff }

It's a small thing but helps avoid unexpected side effects 👍
```

This version:
- Keeps it casual but informative
- Includes a quick example
- Gives a brief reason why
- Uses emoji to keep it friendly
- Still marks it as a "nit" to show it's a minor suggestion

File: cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/queries.spec.js [hunk 5]
```diff
@@ -98,7 +98,7 @@ const queriesTestSuite = () => {
         // Default SQL query should be set
         cy.waitForLoader(true);
         cy.getElementByTestId(`osdQueryEditor__multiLine`).contains(
```
kavilla, line 100: js file 😭

src/plugins/query_enhancements/server/utils/facet.ts
kavilla: Quick thought - was worried about async search strategies (like S3), but I think doing it here in the fetch call should actually cover everything. Since async searches eventually hit this fetch method anyway, we should be good.

Mind double-checking that though? Just want to make sure I'm not missing something with the async flow

abbyhu2000: directly passing the fetch size in the facet will cause async query to fail since it doesn't accept fetch_size as a param
```
const fetchSize = await context.core.uiSettings.client.get('discover:sampleSize');
      const params = {
        body: {
          fetch_size: fetchSize,
          query: query.query,
          ...(meta?.name && { datasource: meta.name }),
          ...(meta?.sessionId && {
            sessionId: meta.sessionId,
          }),
          ...(lang && { lang }),
        },
        ...(format !== 'jdbc' && { format }),
      };
```

Screenshot 2025-02-12 at 5 51 14 PM

abbyhu2000: i think since async search for s3 and PPL doesn't support fetch_size yet, we can interpolate the default query string for s3 dataset types and PPL queries for now. The limit number we get from advanced setting

`source = ${dataset.title} | head 500`

src/plugins/query_enhancements/server/search/ppl_search_strategy.ts
abbyhu2000: Got it. By looking at the `runSearch` and `fetch` function in data plugin, i do not see we pass in any size param here. 

Do you mind keep it on the query enhancements side for now?

src/plugins/query_enhancements/server/utils/facet.ts
abbyhu2000: Will change to only pass in fetch_size if fetch_size is defined.

src/plugins/query_enhancements/server/search/sql_search_strategy.ts
abbyhu2000: changed

src/plugins/data/public/query/query_string/dataset_service/types.ts
```diff
@@ -117,5 +117,5 @@ export interface DatasetTypeConfig {
   /**
    * Returns the initial query that is added to the query editor when a dataset is selected.
    */
-  getInitialQueryString?: (query: Query) => string | void;
+  getInitialQueryString?: (query: Query, sampleSize?: number) => string | void;
```
kavilla, line 120: woule it be worth either pass

```suggestion
  getInitialQueryString?: (query: Query, services?: Partial<IDataPluginServices>) => string | void;
```

kavilla, line 120:
```suggestion
  getInitialQueryString?: (query: Query, uiSettings?: CoreStart['uiSettings']) => string | void;
```

abbyhu2000: changed to this

ruchidh (dismissed):

src/plugins/query_enhancements/server/search/sql_search_strategy.ts
```diff
@@ -34,7 +34,12 @@ export const sqlSearchStrategyProvider = (
     search: async (context, request: any, options) => {
       try {
         const query: Query = request.body.query;
-        const rawResponse: any = await sqlFacet.describeQuery(context, request);
+        request.body.fetch_size = await context.core.uiSettings.client.get('discover:sampleSize');
```
ruchidh, line 37: I think you can,

instead of request.body.fetch_size directly, which could lead to unexpected side effects. Instead, assign it to a new variable.


const fetchSize = await uiSettings.client.get('discover:sampleSize');

const requestWithFetchSize = {
  ...request,
  body: { ...request.body, fetch_size: fetchSize ?? 50 }, // Default fallback
};
Adds a fallback (?? 50) in case fetchSize is undefined

abbyhu2000:
Issue https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9661 is blocking this PR

angle943:
This is Kiro posting.

Merge conflicts on a user-visible sample-size behavior change. @abbyhu2000 — is this still planned? If yes, please rebase. If not, please close.

Issue #2174, opened 2022-08-19
[MD] _msearch not likely to work with multiple data source


zhongnansu:
batch search will issue `_msearch` requests to OpenSearch, which is not likely to work with multiple data sources. maybe deprecate this, or simply disable it when multiple data source is supported?

kavilla:
I wouldn't disable batch requests as it appears to be a highly used setting.

zhongnansu:
> I wouldn't disable batch requests as it appears to be a highly used setting.

Got it. I saw the commit that removes the `deprecated` label of batch search setting. https://github.com/opensearch-project/OpenSearch-Dashboards/pull/735

zhongnansu:
> @zhongnansu Is this still targeting `2.5.0`? I don't see a linked PR.

This will be for later release. Removed 2.5 label

zhongnansu:
Resolved by https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9361

PR #9361, opened 2025-02-10
Add MDS to msearch

wanglam:
### Description

This PR is for adding MDS for `_internal/msearch`. A `data_source_id` can be passed to the msearch API to determine which data source should be used to execute msearch with search requests.

### Issues Resolved

#2174 

## Screenshot

No UI changes

## Testing the changes

- Clone branch code and run `yarn osd bootstrap --single-version loose`
```
data_source.enabled: true
uiSettings:
  overrides:
    'courier:batchSearches': true
```
- Login with admin user
- Visit `/app/management/opensearch-dashboards/dataSources`
- Add an OpenSearch data source connection 
- Visit `/app/home#/tutorial_directory`
- Click "Add data" on the "Sample eCommerce orders" card
- Click "View data" on the "Sample eCommerce orders" card
- Dashboards data in external data source should be showed

## Changelog

- feat: Add MDS to msearch

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/common/search/search_source/legacy/fetch_soon.ts
```diff
@@ -83,20 +83,27 @@ async function delayedFetch(
...
   }
+  const { dataSourceId, ...restRequest } = request;
+  const key = dataSourceId ?? '';
 
-  const i = requestsToFetch.length;
-  requestsToFetch = [...requestsToFetch, request];
-  requestOptions = [...requestOptions, options];
+  const i = requestsToFetch[key]?.length ?? 0;
+  requestsToFetch[key] = [...(requestsToFetch[key] ?? []), restRequest];
+  requestOptions[key] = [...(requestOptions[key] ?? []), options];
```
SuZhou-Joe, line 91: Nit: better to add some comments here to elaborate that we are grouping request by data source.

src/plugins/data/server/search/routes/msearch.ts
```diff
@@ -67,16 +67,23 @@ export function registerMsearchRoute(router: IRouter, deps: SearchRouteDependenc
             })
           ),
         }),
+        query: schema.object({
+          data_source_id: schema.maybe(schema.string()),
+        }),
```
SuZhou-Joe, line 72: Not very sure if the validation will give error if the request does not have any query.

src/plugins/data/common/search/search_source/legacy/fetch_soon.ts
```diff
@@ -83,20 +83,27 @@ async function delayedFetch(
   if (ms === 0) {
     return callClient([request], [options], fetchHandlers)[0];
   }
+  const { dataSourceId, ...restRequest } = request;
```
SuZhou-Joe, line 86: Not sure if it is straightforward to add unit test for this function but it will be great that we have one, the logic is kind of complex.

src/plugins/data/server/search/routes/msearch.ts
wanglam: I've tested with data source disabled case. The `msearch` API works as expected without the data source id query.

SuZhou-Joe (approved):

SuZhou-Joe:
I thought it would be a simple several lines fix. Thanks for this amazing PR.

src/plugins/data/server/search/routes/msearch.ts
```diff
@@ -67,16 +67,23 @@ export function registerMsearchRoute(router: IRouter, deps: SearchRouteDependenc
             })
           ),
         }),
+        query: schema.object({
+          data_source_id: schema.maybe(schema.string()),
```
kavilla, line 71: Haven't tested this yet, but I see a potential issue:

Scenario:
- Index pattern foo* saved on datasourceA
  - Matches: foo-1, foo-2, etc.
- Index pattern bar* also saved on datasourceA
  - Matches: bar-1, bar-2, etc.
- bar* indices actually exist on datasourceB

Won't this lead to 404s when querying bar* since we're looking in datasourceA but the indices are in datasourceB?

To me i almost think multisearch and multiple data sources is almost fundamentally conflicting ideas. How do we fetch the data from multiple data sources with a single query. this essentially feels like cross cluster searching. unless im missing something

zhongnansu (commented):
I have opened an issue for this before. Could you take a look and see if your PR would address the problem there? Thanks https://github.com/opensearch-project/OpenSearch-Dashboards/issues/2174 @wanglam

src/plugins/data/server/search/routes/msearch.ts
wanglam: You're right, trying to search for an index pattern in a different data source than where it exists will cause a 404 error.

To solve this, we separate the search requests by data source on the client-side before sending them. The `fetchSoon` function groups the requests based on their data source, creating multiple msearch requests, one for each data source.

This way, each msearch request is sent to the correct data source, avoiding 404 errors for indices that are in a different data source.

The internal msearch API needs to call msearch in different OpenSearch clients. They are also expensive when combined in the same request. That's why we only support one data source per request.

wanglam:
> I have opened an issue for this before. Could you take a look and see if your PR would address the problem there? Thanks #2174 @wanglam

Hi @zhongnansu , I've looked at the issue you mentioned (#2174). This PR can solve the problem described there. We will add a new parameter called "data source id" to the internal msearch API. 

The way we manage queued search requests will also be improved in this PR. We will store the search requests, parameters, and ongoing promises separately for each data source.

Feel free to review this PR. If anything needs to be changed or improved, I will update the code accordingly.

src/plugins/data/common/search/search_source/legacy/call_client.ts
```diff
@@ -36,7 +36,8 @@ import { ISearchOptions } from '../../index';
 export function callClient(
   searchRequests: SearchRequest[],
```
kavilla, line 37: Does the search request have the index pattern ? Or data source info?

wanglam: The search requests interface is not being updated. The data source ID is only present on the client side. It was added at [line 359](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/0605a80acd8b9d1dfb461bba72e26a9cc3ed94bf/src/plugins/data/common/search/search_source/search_source.ts#L359) and removed at [line 86](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/0605a80acd8b9d1dfb461bba72e26a9cc3ed94bf/src/plugins/data/common/search/search_source/legacy/fetch_soon.ts#L86). The msearch API's search requests interface remains unchanged. The only modification to the internal msearch API was the addition of a new `data_source_id` query parameter. This parameter will trigger a `msearch` call on the specified data source.

SuZhou-Joe (approved):

src/plugins/data/public/search/legacy/call_msearch.ts
```diff
@@ -39,9 +39,10 @@ import { LegacyFetchHandlers } from '../../../common/search/search_source';
  * @internal
  */
 export function getCallMsearch({ http }: { http: HttpStart }): LegacyFetchHandlers['callMsearch'] {
-  return async ({ body, signal }) => {
+  return async ({ body, signal, dataSourceId }) => {
```
kavilla, line 42: would this require modifying the  msearch search params? if we have some?

src/plugins/data/server/search/routes/msearch.ts
```diff
@@ -67,16 +67,23 @@ export function registerMsearchRoute(router: IRouter, deps: SearchRouteDependenc
             })
           ),
         }),
+        query: schema.object({
```
kavilla, line 70: could we actually just shove the data source id into each of the searches.body?  and parse out that value to get the client?

kavilla (commented):
Overall it looks good to me.

A couple thoughts:
- I'm a bit hesitant about modifying the existing `callMsearch` API. Maybe we could stuff the `dataSourceId` into each `search.body` within `searches` instead and parse it out on the server side? This could mitigate modifying existing APIs.
- Alternatively, what about adding a new `_enhanced` msearch route instead of changing the original? That could keep things more isolated.

But I can be convinced that this ok to pass by query param. Make sure we have solid testing coverage, especially for both MDS and non-MDS scenarios. Leveraging the existing Cypress tests from @ananzh and @angle943 could be a good starting point.

Otherwise, nice work! Let me know what you think about those suggestions. Looking forward to getting this merged.

kavilla (approved):

zhongnansu (approved):

zhongnansu:
Do we need to backport to 2.x?

src/plugins/data/public/search/legacy/call_msearch.ts
wanglam: It's doable to move the data source id to request body. Then we don't need to add extra data source id query.

wanglam:
> Overall it looks good to me.
> 
> A couple thoughts:
> 
> * I'm a bit hesitant about modifying the existing `callMsearch` API. Maybe we could stuff the `dataSourceId` into each `search.body` within `searches` instead and parse it out on the server side? This could mitigate modifying existing APIs.
> * Alternatively, what about adding a new `_enhanced` msearch route instead of changing the original? That could keep things more isolated.
> 
> But I can be convinced that this ok to pass by query param. Make sure we have solid testing coverage, especially for both MDS and non-MDS scenarios. Leveraging the existing Cypress tests from @ananzh and @angle943 could be a good starting point.
> 
> Otherwise, nice work! Let me know what you think about those suggestions. Looking forward to getting this merged.

Hi @kavilla, Thanks for sharing your thoughts. Moving `dataSourceId` into each `searches` inside `search.body` would be complicated on the server side. We'd need to merge `msearch` responses from different data sources, and the remote data source cluster would throw errors during the `msearch` call. The final response would look like:
```
[
    {searchResponse1},
    {error},
    {searchResponse3},
]
```
This would require refactoring the API caller to handle errors in the response body.

Our current implementation using a single data source ID per internal `msearch` call avoids this issue. The whole API request fails cleanly if errors occur.

For the `callMsearch` API on the browser side, how about moving the data source ID to the request body instead of using a query parameter? Then we don't need to change the interface of `msearch`.

PR #9371, opened 2025-02-12
feat: Add MDS support for index pattern validate script

raintygao:
### Description

feat: Add MDS support for index pattern validate script

Scripted field is a feature of index pattern that used in visualizations and display in documents. Validation is an API used for validation script when creating scripts. 

image
Before this PR, this API did not support MDS, so it would throw an error when operating the index pattern on a remote data source. After this PR, this error would be fixed.

## Screenshot

No UI change

## Testing the changes

Turn on/off the MDS, and then visit index pattern script field edit page, the API won't break down any more if editing an index pattern on remote data source.

## Changelog

- feat: Add MDS support for index pattern validate script

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/index_pattern_management/public/components/field_editor/field_editor.tsx
+++ b/src/plugins/index_pattern_management/public/components/field_editor/field_editor.tsx
@@ -136,6 +136,7 @@ export interface FieldEditorState {
   errors?: string[];
   format: any;
   spec: IndexPatternField['spec'];
+  dataSourceId?: string;
 }
 
 export interface FieldEdiorProps {
@@ -176,6 +177,7 @@ export class FieldEditor extends PureComponent<FieldEdiorProps, FieldEditorState
       isSaving: false,
       format: props.indexPattern.getFormatterForField(spec),
       spec: { ...spec },
+      dataSourceId: indexPattern.dataSourceRef?.id,
     };
     this.supportedLangs = getSupportedScriptingLanguages();
     this.deprecatedLangs = getDeprecatedScriptingLanguages();
@@ -799,6 +801,7 @@ export class FieldEditor extends PureComponent<FieldEdiorProps, FieldEditorState
         script: field.script as string,
         indexPatternTitle: indexPattern.title,
         http: this.context.services.http,
+        dataSourceId: this.state.dataSourceId,
       });
 
       if (!isValid) {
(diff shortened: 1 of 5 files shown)


zhongnansu (commented):
Thanks for improving on MDS integration with core features. Could you add more details to the PR to explain what the script feature does, what's the issue, and what benefit does MDS integration bring us? Including a video or gif might help others to better understand! @raintygao

src/plugins/index_pattern_management/server/routes/preview_scripted_field.ts
```diff
@@ -43,11 +43,18 @@ export function registerPreviewScriptedFieldRoute(router: IRouter): void {
...
+        }),
       },
     },
     async (context, request, res) => {
-      const client = context.core.opensearch.client.asCurrentUser;
       const { index, name, script, query, additionalFields } = request.body;
+      const { dataSourceId } = request.query;
+      const client =
+        dataSourceId && context.dataSource
+          ? await context.dataSource.opensearch.getClient(dataSourceId)
```
wanglam, line 56: Do we need to handle the error during `getClient`?

raintygao: I think there may be no need because if the remote client can't be got, it should throw error as expected.

src/plugins/index_pattern_management/public/components/field_editor/field_editor.tsx
```diff
@@ -799,6 +801,7 @@ export class FieldEditor extends PureComponent<FieldEdiorProps, FieldEditorState
         script: field.script as string,
         indexPatternTitle: indexPattern.title,
         http: this.context.services.http,
+        dataSourceId: this.state.dataSourceId,
```
SuZhou-Joe, line 804: Should we use `props.indexPattern.dataSourceRef?.id`? I do not think a dataSourceId state is necessary for this component.

src/plugins/index_pattern_management/server/routes/preview_scripted_field.ts
SuZhou-Joe: We may need to wrap the getClient or the http request may give statusCode of 500.

src/plugins/index_pattern_management/public/components/field_editor/field_editor.tsx
raintygao: Thanks for a nice catch, updated.

src/plugins/index_pattern_management/server/routes/preview_scripted_field.ts
raintygao: Updated.

wanglam (approved):

SuZhou-Joe (approved):

Hailong-am (approved):

PR #9379, opened 2025-02-13
Render PPL time column using the correct time zone

### LINKED ISSUE #9104: [BUG] Invalid handling of time filter in PPL
**Describe the bug**

Invalid handling of time filter in PPL

**To Reproduce**

1. Login anonymous in the future playground.
2. Select the **OTEL demo environment**.
3. Select **Discover** from the menu on the left hand side
4. In the data source selection dropdown, type **ecomm**, then select the first item `opensearch*ecomm*`
5. Using PPL query mode, set the time filter to absolute mode starting at `Oct 20, 2024 @ 00:00:00.00` and ending at `Oct 20, 2024 @23:00:00.00` then run `source = opensearch*ecomm*,`
    * Results having timestamps on both Oct 19 and Oct 20 are shown. As no timezone information is presented the assumption is that it is at least the same timezone and thus incorrect.

**Expected behavior**

 * Step 5: Results are shown in a list of entries with **Time** and **_source**, all within the selected timeframe.

**OpenSearch Version**

2.18

**Dashboards Version**

2.18

**Plugins**

Alerting, anomaly-detection, asynchronous-search, cross-cluster-replication, custom-codecs, flow-framework, geospatial, index-management, job-scheduler, knn, ml, neural-search, notifications, notifications-core, observability, performance-analyzer, reports-scheduler, security, security-analytics, skills, sql, system-templates, query-insights

**Screenshots**

N/A

**Host/Environment:**

 - OS: macOS
 - Browser and version Google Chrome, version 131.0.6778.205

**Additional context**

N/A

abbyhu2000:
### Description
When sending PPL requests, the time based fields are rendered in wrong time zone, and it looks like DQL and PPL are getting different results. We should format the PPL search response to the same format as DQL search response to indicate it is in UTC so the time fields can be formatted and rendered correctly in the discover table according to user defined time zone in advanced setting.

For example, previously 
```
// PPL search response
"order_date": "2025-02-13 00:51:50"

// DQL search response
"order_date": "2025-02-13T00:51:50+00:00”
```

### Issues Resolved

resolves #9104 

## Screenshot
Before: 
 
PPL render time field in the wrong timezone

https://github.com/user-attachments/assets/f0e24d6e-6560-4897-9d3b-5b41c780b26e

After: 

PPL and DQL return and render same results for default timezone or user defined timezone

https://github.com/user-attachments/assets/8cbc8917-3826-4c5c-81a4-0941f7af9e96

## Changelog

- fix: Make PPL time column respect time zone and date format

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/common/data_frames/utils.ts
+++ b/src/plugins/data/common/data_frames/utils.ts
@@ -15,7 +15,7 @@ import {
 } from './types';
 import { IFieldType } from './fields';
 import { IndexPatternFieldMap, IndexPatternSpec } from '../index_patterns';
-import { TimeRange } from '../types';
+import { OSD_FIELD_TYPES, TimeRange } from '../types';
 import { IDataFrame } from './types';
 import { ISearchOptions, SearchSourceFields } from '../search';
 
@@ -62,35 +62,40 @@ export const convertResult = ({
     for (let index = 0; index < data.size; index++) {
       const hit: { [key: string]: any } = {};
 
+      const processNestedField = (field: any, value: any, formatter: any): any => {
+        const nestedHit: { [key: string]: any } = value;
+        Object.entries(value).forEach(([nestedField, nestedValue]) => {
+          // Need to get the flattened field name for nested fields ex.products.created_on
+          const flattenedFieldName = `${field.name}.${nestedField}`;
+
+          // Go through search source fields to find the field type of the nested field
+          fields?.index?.fields.forEach((searchSourceField) => {
+            if (
+              searchSourceField.displayName === flattenedFieldName &&
+              searchSourceField.type === 'date'
+            ) {
+              nestedHit[nestedField] = formatter(nestedValue as string, OSD_FIELD_TYPES.DATE);
+            }
+          });
+        });
+        return nestedHit;
+      };
+
       const processField = (field: any, value: any): any => {
-        if (options && options.dateFieldsFormatter) {
-          const formatter = options.dateFieldsFormatter;
+        if (options && options.formatter) {
           // Handle date fields
           if (field.type === 'date') {
-            return formatter(value);
+            return options.formatter(value, OSD_FIELD_TYPES.DATE);
           }
           // Handle nested objects with potential date fields
           else if (field.type === 'object') {
-            const nestedHit: { [key: string]: any } = value;
-            Object.entries(value).forEach(([nestedField, nestedValue]) => {
-              // Need to get the flattened field name
-              const flattenedFieldName = `${field.name}.${nestedField}`;
-              fields?.index?.fields.forEach((indexPatternField) => {
-                if (
-                  indexPatternField.displayName === flattenedFieldName &&
-                  indexPatternField.type === 'date'
-                ) {
-                  nestedHit[nestedField] = formatter(nestedValue as string);
-                }
-              });
-            });
-            return nestedHit;
+            return processNestedField(field, value, options.formatter);
           } else {
             // Default case when the field is either a date type or object type
             return value;
           }
         } else {
-          // Default case when we don't have a dateFieldsFormatter
+          // Default case when we don't have a formatter
           return value;
         }
       };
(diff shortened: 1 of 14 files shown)


Maosaic:
Any unit test we want to add or update?

src/plugins/data/common/data_frames/utils.ts
```diff
@@ -52,9 +53,30 @@ export const convertResult = (response: IDataFrameResponse): SearchResponse<any>
   if (data && data.fields && data.fields.length > 0) {
     for (let index = 0; index < data.size; index++) {
       const hit: { [key: string]: any } = {};
+
+      const processField = (field: any, value: any): any => {
+        // Handle date fields
+        if (moment(value, ['YYYY-MM-DD HH:mm:ss'], true).isValid()) {
```
kavilla, line 59: i can't remember this is before we know it is a date field correct? or do we already know it's a date field? as in field.type === 'date'.

`YYYY-MM-DD HH:mm:ss` seems too specific to PPL, correct me if im wrong.

would it make sense to do it something like

```
const tzConfig = opensearchDashboards.services.uiSettings!.get('dateFormat:tz');
const momentParsedValue = moment(value).tz(tzConfig);
if (momentParsedValue.isValid()) return momentParsedValue?.format('YYYY-MM-DDTHH:mm:ss.SSSZ');
```

or does it need to be +00:00 because to me it feels like we would be forcing any date value of `YYYY-MM-DD HH:mm:ss` to have a value we statically append `+00:00`

kavilla:
wow. great find! 

do we have a manual test or could we add some cypress/unit test for this one. when first implementing this extensibility i had completely neglected the responses could be coming back in a different timezone format. if developers wanted to add new langauges and new data sources it might be useful for them to ensure they properly are formatting their responses or that we do it for all cases. or at least documented

src/plugins/data/common/data_frames/utils.ts
abbyhu2000: yea at first i used field.type === 'date' but then i realized this will neglect those nested field which does not have a field.type yet, therefore i switch to use moment.

abbyhu2000: We do not need to get the timezone here because opensearch backend always store time value as UTC, no matter which timezone we define in the advanced setting. So here we should treat the string we got back as in UTC time. 

We could potentially do: 
```
if (moment.utc(value).isValid()) {
          return moment.utc(value).format('YYYY-MM-DDTHH:mm:ss.SSSZ');
        }
```
but this doesnt work for some reason: when the value type is object, it will always go into the if loop as it is a date.

abbyhu2000: i changed `return value.replace(' ', 'T') + '+00:00';` to `return moment.utc(value).format('YYYY-MM-DDTHH:mm:ssZ');` so it it more general

abbyhu2000:
Added unit test for convertResult() function @Maosaic @kavilla

abbyhu2000:
Reconstruct this PR to use OSDfieldsList to get the type; also add a config in the language config so PPL will format its date fields

src/plugins/data/public/query/query_string/dataset_service/lib/index_type.ts
```diff
@@ -81,6 +81,9 @@ export const indexTypeConfig: DatasetTypeConfig = {
       pattern: dataset.title,
       dataSourceId: dataset.dataSource?.id,
     });
+    // TODO: map fields to the OSD Field type
```
kavilla, line 84: could you clean this up

abbyhu2000: removed

src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -66,7 +67,20 @@ export class QueryEnhancementsPlugin
         usageCollector: data.search.usageCollector,
       }),
       getQueryString: (currentQuery: Query) => `source = ${currentQuery.dataset?.title}`,
-      fields: { sortable: false, filterable: false, visualizable: false },
+      fields: {
+        sortable: false,
+        filterable: false,
+        visualizable: false,
+        formatter: (value: string, type: OSD_FIELD_TYPES) => {
```
abbyhu2000, line 74: make this formatter more general so in the future it can be utilized for other cases too

kavilla (approved):

sejli (approved):

PR #9404, opened 2025-02-17
Remove obsolete compose version specifier

SimonCW:
### Description

Version specifier for docker compose is obsolete. See: https://docs.docker.com/reference/compose-file/version-and-name/#version-top-level-element-obsolete

See this PR to collect changes across opensearch-project repos: https://github.com/opensearch-project/documentation-website/pull/8606

## Changelog
- skip


kavilla (approved):

Hailong-am (commented):
Can we add the note that this will need docker-compose V2

Hailong-am (approved):

SimonCW:
Hey @Hailong-am, personally I wouldn't do that. Compose V1 has stopped receiving updates in July 2023. I changed this across other libraries in the OpenSearch Org without a notice.

PR #9420, opened 2025-02-19
[Workspace] Fix find saved object within a workspace without queryparams return all the saved objects

FriedhelmWS:
### Description

Given we have two workspaces (namely `first` and `second`), and each has a saved object.

We retrieve the saved objects using the following URL (using workspace prefix but no query parameters):
`/w/first/api/opensearch-dashboards/management/saved_objects/_find?type=index-pattern`

Previously, this request would return both saved objects because when no `workspace` query parameter was provided, our endpoint would pass `workspace: undefined` to the actual query, bypassing the workspace check. As a result, the backend would return all saved objects even when the original request was made under the `first` workspace context.

This PR removes the logic of passing `workspace: undefined `when there is no `workspace` query parameter. Now, the wrapper for the `_find` endpoint automatically retrieves the current workspace context, ensuring the backend only returns saved objects associated with the current workspace.

### Issues Resolved

NA

## Screenshot

NA

## Changelog

- fix: fixing when find saved objects within a workspace returns saved objects in all the workspaces

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/saved_objects_management/server/routes/find.test.ts
+++ b/src/plugins/saved_objects_management/server/routes/find.test.ts
@@ -1,0 +1,158 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { registerFindRoute } from './find';
+import { ISavedObjectsManagement } from '../services';
+import { managementMock } from '../services/management.mock';
+import { IRouter } from 'src/core/server';
+import { httpServiceMock } from '../../../../core/server/mocks';
+
+describe('Find route', () => {
+  let router: jest.Mocked<IRouter>;
+  let managementService: jest.Mocked<ISavedObjectsManagement>;
+  let context: any;
+  let req: any;
+  let res: any;
+
+  beforeAll(() => {
+    router = httpServiceMock.createRouter();
+
+    // Mock management service
+    managementService = managementMock.create();
+
+    // Mock context with savedObjects client
+    const savedObjectsClient = {
+      find: jest.fn(),
+      get: jest.fn(),
+    };
+
+    context = {
+      core: {
+        savedObjects: {
+          client: savedObjectsClient,
+        },
+      },
+    };
+
+    // Mock request
+    req = {
+      query: {
+        perPage: 20,
+        page: 1,
+        type: 'test-type',
+        fields: ['field1', 'field2'],
+      },
+    };
+
+    // Mock response
+    res = {
+      ok: jest.fn(),
+    };
+  });
+
+  it('should register GET route with correct path', () => {
+    registerFindRoute(router, Promise.resolve(managementService));
+
+    expect(router.get).toHaveBeenCalledWith(
+      expect.objectContaining({
+        path: '/api/opensearch-dashboards/management/saved_objects/_find',
+      }),
+      expect.any(Function)
+    );
+  });
+
+  it('should handle basic find request', async () => {
+    registerFindRoute(router, Promise.resolve(managementService));
+
+    const findResponse = {
+      saved_objects: [
+        {
+          type: 'test-type',
+          id: '1',
+          attributes: {
+            field1: 'value1',
+            field2: 'value2',
+          },
+        },
+      ],
+      total: 1,
+      per_page: 20,
+      page: 1,
+    };
+
+    context.core.savedObjects.client.find.mockResolvedValue(findResponse);
+
+    const handler = router.get.mock.calls[0][1];
+    await handler(context, req, res);
+
+    expect(context.core.savedObjects.client.find).toHaveBeenCalledWith(
+      expect.objectContaining({
+        perPage: 20,
+        page: 1,
+        type: 'test-type',
+      })
+    );
+    expect(res.ok).toHaveBeenCalled();
+  });
+
+  it('should handle index-pattern type correctly', async () => {
+    registerFindRoute(router, Promise.resolve(managementService));
+
+    const findResponse = {
+      saved_objects: [
+        {
+          type: 'index-pattern',
+          id: '1',
+          attributes: {
+            title: 'test-pattern',
+          },
+          references: [],
+        },
+      ],
+      total: 1,
+      per_page: 20,
+      page: 1,
+    };
+
+    context.core.savedObjects.client.find.mockResolvedValue(findResponse);
+    context.core.savedObjects.client.get.mockResolvedValue({ attributes: {} });
+
+    const handler = router.get.mock.calls[0][1];
+    await handler(context, req, res);
+
+    expect(context.core.savedObjects.client.find).toHaveBeenCalled();
+    expect(res.ok).toHaveBeenCalled();
+  });
+
+  it('should handle workspace filtering', async () => {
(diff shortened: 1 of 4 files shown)


src/plugins/workspace/server/integration_tests/routes.test.ts
```diff
@@ -19,6 +19,27 @@ const testWorkspace: WorkspaceAttribute = {
...
+    WORKSPACE_TYPE,
+  ]);
+  await Promise.all(
+    listResult.body.result.workspaces.map((item: WorkspaceAttribute) =>
+      // this will delete reserved workspace
+      savedObjectsRepository.delete(WORKSPACE_TYPE, item.id)
+    )
+  );
+};
+
```
FriedhelmWS, line 42: Just refactoring repeated code

SuZhou-Joe (dismissed):
Thanks for the fix. LGTM

SuZhou-Joe (dismissed):

src/plugins/saved_objects_management/server/routes/find.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    // Mock management service
+    managementService = {
+      isImportAndExportable: jest.fn().mockReturnValue(true),
+      getDefaultSearchField: jest.fn().mockReturnValue('title'),
+      getIcon: jest.fn().mockReturnValue('icon'),
+      getTitle: jest.fn().mockReturnValue('title'),
+      getEditUrl: jest.fn().mockReturnValue('/edit/mock'),
+      getInAppUrl: jest.fn().mockReturnValue('/api/mock'),
+      getNamespaceType: jest.fn().mockReturnValue('mockType'),
+    };
```
SuZhou-Joe, line 38: We can leverage existing mocks from:
- https://github.com/opensearch-project/OpenSearch-Dashboards/blob/8a9666444b49c39a9c03aaf7c0af3a7c111564d8/src/core/server/http/http_service.mock.ts
- https://github.com/opensearch-project/OpenSearch-Dashboards/blob/d7004dc5b0392477fdd54ac66b29d231975a173b/src/plugins/saved_objects_management/server/services/management.mock.ts

to construct these mocks.

FriedhelmWS: Done

src/plugins/saved_objects_management/server/routes/find.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  it('should register GET route with correct path', () => {
+    expect(router.get).toHaveBeenCalledWith(
+      expect.objectContaining({
+        path: '/api/opensearch-dashboards/management/saved_objects/_find',
+      }),
+      expect.any(Function)
+    );
+  });
```
SuZhou-Joe, line 72: This test case looks weird because it only has assertions. I am thinking if we should move `registerFindRoute(router, Promise.resolve(managementService))` into each test case or we move the assertions into `beforeEach`

src/plugins/saved_objects_management/server/routes/find.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  let managementService: jest.Mocked<ISavedObjectsManagement>;
+  let context: any;
+  let req: any;
+  let res: any;
+
+  beforeAll(() => {
+    router = {
+      get: jest.fn(),
+      handleLegacyErrors: jest.fn((handler) => handler),
+    } as any;
```
SuZhou-Joe, line 28: We can use the mock from http service as well I guess and we do not need a type casting here.

src/plugins/saved_objects_management/server/routes/find.test.ts
FriedhelmWS: Done, move `registerFindRoute(router, Promise.resolve(managementService))` to each individual test cases to avoid confusion and improve isolation.

src/plugins/saved_objects_management/server/routes/find.test.ts
FriedhelmWS: Done, previously got some issue on import module but now solved

src/plugins/workspace/server/integration_tests/routes.test.ts
```diff
@@ -696,4 +671,70 @@ describe('workspace service api integration test when savedObjects.permission.en
       ).toEqual({ write: { users: ['foo'] } });
     });
   });
+
+  describe('Saved Object Management APIs', () => {
```
wanglam, line 675: Shall we move these integration tests under `src/plugins/saved_objects_management` folder? It was testing API in saved objects management instead of workspace routes.

wanglam, line 675: Shall we move these integration tests under `src/plugins/saved_objects_management` folder? It was testing saved object management APIs instead of workspace related API. I'm prefer to creating integration tests in that file.

FriedhelmWS: This is a bit tricky I guess, because here we only test the saved object API under path `w/${workspaceId}/saved_object...`, which means we only test the API under workspace context. This is to ensure the behaviour of the API is expected when we already entered a workspace and trying to find saved objects. Therefore I think it is more relevant to the workspace scope instead of saved object management context.

wanglam (approved):

SuZhou-Joe (dismissed):

SuZhou-Joe (approved):

src/plugins/saved_objects_management/server/routes/find.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    await handler(context, req, res);
+
+    expect(context.core.savedObjects.client.find).toHaveBeenCalled();
+    expect(res.ok).toHaveBeenCalled();
+  });
+
+  it('should handle workspace filtering', async () => {
+    registerFindRoute(router, Promise.resolve(managementService));
+
+    req.query.workspaces = ['workspace1'];
```
Hailong-am, line 138: nit: should we have a case for workspace is empty?

Hailong-am (approved):

src/plugins/saved_objects_management/server/routes/find.ts
```diff
@@ -98,7 +98,11 @@ export const registerFindRoute = (
         ...req.query,
         fields: undefined,
         searchFields: [...searchFields],
-        workspaces: req.query.workspaces ? Array<string>().concat(req.query.workspaces) : undefined,
+        ...(req.query.workspaces
+          ? {
+              workspaces: Array<string>().concat(req.query.workspaces),
+            }
+          : {}),
```
Hailong-am, line 105: I think we need to filter out `*` here, if we pass workspaces as `*`, it will able to query all workspace.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/f52b5d2e9f0f1b93f3f22246783e992820e1213c/src/plugins/workspace/server/saved_objects/workspace_id_consumer_wrapper.ts#L48-L58

SuZhou-Joe: We have another PR to fix the * issue: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9409. I think it should be fine for this PR to leave *.

src/plugins/saved_objects_management/server/routes/find.test.ts
FriedhelmWS: I think the no workspace case is already been covered by the above "basic find" test case. For a find request without workspace filtering then the saved object is always opt-in so that it somehow pointless to explicitly test this case.

PR #9433, opened 2025-02-21
Fix flakieness in cypress tests

ananzh:
### Description

Fix flakieness in cypress tests

## Changelog

- test: Fix flakieness in cypress tests

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/cypress/utils/apps/query_enhancements/saved_queries.js
+++ b/cypress/utils/apps/query_enhancements/saved_queries.js
@@ -10,6 +10,20 @@ import { APPLIED_FILTERS } from './saved';
 import { setDatePickerDatesAndSearchIfRelevant } from './shared';
 
 /**
+ * Ensures the saved query management popover is open by checking its state and clicking if needed
+ */
+export const ensurePopoverIsOpen = () => {
+  cy.getElementByTestId('saved-query-management-popover-button')
+    .parent()
+    .parent()
+    .then(($popover) => {
+      if (!$popover.hasClass('euiPopover-isOpen')) {
+        cy.getElementByTestId('saved-query-management-popover-button').click();
+      }
+    });
+};
+
+/**
  * Error text when there is a name conflict when saving a query.
  * @constant
  * @type {string}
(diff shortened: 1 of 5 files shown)


cypress/utils/apps/query_enhancements/saved_queries.js
```diff
@@ -9,6 +9,20 @@ import { APPLIED_FILTERS } from './saved';
...
+ * Ensures the saved query management popover is open by checking its state and clicking if needed
+ */
+export const ensurePopoverIsOpen = () => {
+  cy.getElementByTestId('saved-query-management-popover-button')
+    .parent()
+    .parent()
+    .then(($popover) => {
+      if (!$popover.hasClass('euiPopover-isOpen')) {
+        cy.getElementByTestId('saved-query-management-popover-button').click();
+      }
```
angle943, line 22: The `.parent().parent()` will start failing if any change is introduced that causes a reshuffling of DOM elements (for ex, someone makes a change to the `ouiPopover`). 

Because DOM elements are not guaranteed to be static, I'm worried about this change. There's two solutions I can think of:

1. (ideal) add a data-test-subj to the actual element that you are needing, and do `.getElementByTestId()` on that element directly (this way we are not relying on the actual HTML DOM structure
2. if for some reason you can't add the data-test-subj to that element, then is it okay if we add an `else{}` clause here and throw an error? Because if the DOM structure changes and the `.parent().parent()` no longer works, this test will start silently failing and it'll be hard to debug why

ananzh: will try

angle943 (approved):

cypress/utils/apps/query_enhancements/saved_queries.js
ananzh: I will create an issue to assign myself. This is to make diff test env to have same tests.

abbyhu2000 (approved):

Issue #8463, opened 2024-10-03
[BUG] EOL of Node.js version 18.x used in Opensearch-dashboards


Chandana-EP:
Node.js version 18.x will reach its end of life on April 30th, 2025.

**Expected behavior**
Opensearch-dashboards to not bundle dependencies/libraries that have reached EOL and aren't supported anymore.

**Dashboards Version**
2.14.0

When can we expect the NodeJS Non-EOL version update for Opensearch Dashboards ??

ananzh:
Node 18 ends security support on 4/30/2025.

<img width="764" alt="Screenshot 2024-10-14 at 6 59 20 PM" src="https://github.com/user-attachments/assets/67b0562b-2df4-495c-976a-ebb98ffe40f9">

Should we plan to bump to 20 or 22? 

CC: @ashwin-pc @AMoo-Miki

AMoo-Miki:
Due to its commitment to backward compatibility within a major version, OSD 2.x will need to support Node.js versions 14, 16, and 18. It is likely that OSD will need to upgrade or replace numerous libraries to ensure compatibility with Node.js 20 or 22 during both the development and runtime execution phases. While some libraries may run correctly with an older version of Node.js at runtime, developing with them on older Node.js versions may not be possible. If these upgrades impose any breaking changes, it is highly probable that Node.js upgrades will need to be deferred until the release of OSD 3.x.

In my opinion, there should be no issues using OSD 2.x runtime on Node.js 20 or 22, and [the documentation](
https://opensearch.org/docs/latest/install-and-configure/install-dashboards/index/#nodejs-compatibility) provides instructions for utilizing these versions in your runtime environment. Maybe post back if you give it a shot.

mrudrego:
@AMoo-Miki , any idea when OSD 3.x is planned to be released ?

Thanks,

AMoo-Miki:
> @AMoo-Miki , any idea when OSD 3.x is planned to be released ?

According to [this](https://opensearch.org/releases.html), not in the near future.

coralsundy:
Hi @AMoo-Miki with 3.0.0 in the plan for 2025 now, are we in the process of merging #9171 ?

Build repo might need to update images as well.

Thanks.

peterzhuamazon:
Hi @AMoo-Miki with 3.0.0 in the plan for 2025 now, are we in the process of merging #9171 ?

Build repo might need to update images as well.

Thanks.

[...]

PR #9466, opened 2025-02-28
[chore] upgrade to Node.js 20

kavilla:
### Description

Upgrading the Node.js version used from 18.x to the latest LTS version 20.x.

### Issues Resolved

closes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/8463

## Screenshot

n/a

## Testing the changes

local and ci

## Changelog

- fix: upgrade to Node.js v20

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


kavilla:
what to see how far the automation gets

package.json
```diff
@@ -78,20 +78,21 @@
...
-    "osd:ciGroup10": "echo \"cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/01/*.spec.js\"",
+    "osd:ciGroup10": "echo \"cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/01/*.spec.js,!cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/01/saved_search.spec.js\"",
     "osd:ciGroup11": "echo \"cypress/integration/dashboard_sanity_test.spec.ts\"",
     "osd:ciGroup12": "echo \"cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/02/*.spec.js\"",
     "osd:ciGroup13": "echo \"cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/03/*.spec.js\"",
     "osd:ciGroup14": "echo \"cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/04/*.spec.js\"",
     "osd:ciGroup15": "echo \"cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/05/*.spec.js\"",
+    "osd:ciGroup16": "echo \"cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/01/saved_search.spec.js\"",
```
angle943, line 95: is this and above change for `osd:ciGroup10` a temporary thing that you will revert, or you think it should be permanent?

zhongnansu (approved):

package.json
```diff
@@ -519,7 +529,7 @@
     "zlib": "^1.0.5"
   },
   "engines": {
-    "node": ">=14.20.1 <19",
+    "node": ">=14.20.1 <21",
```
virajsanghvi, line 532: Are we still supporting < 20?

cypress.config.ts
```diff
@@ -16,8 +16,9 @@ module.exports = defineConfig({
     runMode: 2,
     openMode: 0,
   },
-  viewportWidth: 2000,
-  viewportHeight: 1320,
+  numTestsKeptInMemory: 5,
+  viewportWidth: 1920,
+  viewportHeight: 1080,
```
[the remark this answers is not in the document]
kavilla: more so 2000 x 1320 seems non standard aspect ratio

where as 1920 x 1080 being 16x9 to potentially avoid any visual oddities

package.json
kavilla: some light testing seems like it technically can. at least no worse than it was before.

however, i believe for 20 < it will become not compatible

virajsanghvi: @ashwin-pc We're good to drop compatibility for node < 20, right? I think its more preferrable to remove and add back in later if necessary vs keep and have to maintain for all of 3.x (even if we can validate its fine through node 14 for now).

virajsanghvi (commented):
approving as node version question can be decided on later, but let's make sure we close it before 4/1

virajsanghvi (approved):

package.json
ashwin-pc: I agree. Lets remove it.

PR #9469, opened 2025-03-01
Update formatHit to accept type as param

angle943:
### Description

- Update `formatHit.formatField()` to accept `type` as an argument.
- This is needed because when working on the "export to CSV" functionality, we want to export it as type "text", not type "html", which was the default (and unconfigurable)
- `formatHit.formatField()` calls `convert()`, which already accepts `type`, so my code is just leveraging that
- But since `formatHit.formatField()` assumed that it was only for HTML type, the code was written to always leverage cache. But we should only be leveraging cache when its for HTML type, so I added an `if` condition.
- I also added unit tests for the entire `format_hit.ts` file
- This is part of a larger work on export to CSV functionality. You can see the draft PR here: #9333 . 

## Changelog
- feat: update formatHit.formatField() to accept `type` as an argument

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/common/index_patterns/index_patterns/format_hit.ts
+++ b/src/plugins/data/common/index_patterns/index_patterns/format_hit.ts
@@ -42,7 +42,7 @@ export function formatHitProvider(indexPattern: IndexPattern, defaultFormat: any
     hit: Record<string, any>,
     val: any,
     fieldName: string,
-    type: FieldFormatsContentType = 'html'
+    type: FieldFormatsContentType
   ) {
     const field = indexPattern.fields.getByName(fieldName);
     const format = field ? indexPattern.getFormatterForField(field) : defaultFormat;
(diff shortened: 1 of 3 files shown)


src/plugins/data/common/index_patterns/index_patterns/format_hit.ts
```diff
@@ -42,15 +42,16 @@ export function formatHitProvider(indexPattern: IndexPattern, defaultFormat: any
...
     val: any,
     fieldName: string,
-    type: FieldFormatsContentType = 'html'
+    type: FieldFormatsContentType
   ) {
     const field = indexPattern.fields.getByName(fieldName);
     const format = field ? indexPattern.getFormatterForField(field) : defaultFormat;
 
     return format.convert(val, type, { field, hit, indexPattern });
   }
```
angle943, line 51: This function `convert` is only used by this file, and since the only two usage of it passes `type` as an argument now, there is no more need to make this an optional param

src/plugins/data/common/index_patterns/index_patterns/index_pattern.ts
```diff
@@ -251,7 +255,6 @@ export class IndexPattern implements IIndexPattern {
    * @param name field name
    * @param script script code
    * @param fieldType
-   * @param lang
```
angle943, line 254: my IDE was complaining that this is not a param, so might as well remove it even though its not in scope of my work

d-buckner (approved):

LDrago27 (approved):

ananzh (approved):

Issue #353, opened 2021-05-13
[PROPOSAL]  Dashboard Rows in OpenSearch Dashboards


jkowall:
## Requirements - What kind of business use case are you trying to solve?
*   Dashboarding today allows for long and complex dashboards, but they are hard to scroll and manage on screen
*   Allow dashboards to link to other dashboards or copy configurations from them
*   Create centralized views of data into a single dashboard for use cases in a NOC/operations center for example. 

## Problem - What blocks you from solving this today?
There is a missing feature to create rows or separators between sections of a long dashboard. This would allow for separations of parts of a dashboard and the ability to collapse sections when not in use.

Everything in a dashboard is singular to a specific dashboard. We would need to create the notion of the ability to link dashboards or copy dashboard configurations into another dashboard. 

## Proposal - what do you suggest to solve the problem or improve the existing situation?
There are two ways we can solve this:

1. Adding a “Row container” that will contain tiles (Visualizations )
    1. Every row will have a name and an arrangement of tiles
    2. The user can minimize or collapse rows
    3. Instead of placing visualizations one by one, Rows could clone (or reference?) all visualizations of an existing dashboard (later?)
2. Adding a new object that will contain dashboards
    4. Same as creating dashboards and adding visualizations, but adding dashboards

We would like to create a way to add a row and drag existing panels into and between rows for example this is possible in Grafana.

![image](https://user-images.githubusercontent.com/1859948/118139323-3a276b80-b3d5-11eb-9e06-46001d4adf6d.png)

![image](https://user-images.githubusercontent.com/1859948/118139346-3f84b600-b3d5-11eb-834c-20ceb366e6aa.png)

When adding a new panel in OpenSearch Dashboards, you should also have an option to add a row.

![image](https://user-images.githubusercontent.com/1859948/118139399-4d3a3b80-b3d5-11eb-8b52-965dafc214c2.png)

The row could be renamed, or deleted for example this in Grafana.

![image](https://user-images.githubusercontent.com/1859948/118139438-57f4d080-b3d5-11eb-887b-5e12d12215c3.png)

![image](https://user-images.githubusercontent.com/1859948/118139466-5deab180-b3d5-11eb-94d1-aac7f5c45a6d.png)

The user may drag panels into rows or between rows.

## Assumptions
The user can create dashboards as new, or editing existing dashboards or linking existing dashboards, or copied into existing dashboards.

## Any open questions to address
[...]

PR #9478, opened 2025-03-03
[Workspace] Display error handling page when the enter the workspace in stale state

FriedhelmWS:
### Description

Add a error handler to display the fatal error page when visit the workspace in a stale state (i.e. `currentWorkspaceId && !currentWorkspace?.id` is true) so that users could navigate back to the home page.

### Issues Resolved

NA

## Screenshot

Before: page stuck in the loading state
Screenshot 2025-03-04 at 13 39 46

After: display the fatal error page
Screenshot 2025-03-04 at 13 44 05

## Changelog

- fix: Add error handling page for workspace stale state to prevent getting stuck in loading

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/workspace/public/plugin.test.ts
```diff
@@ -367,6 +368,45 @@ describe('Workspace plugin', () => {
...
+    });
+    setupMock.workspaces.workspaceError$.error(workspaceError);
+
+    await workspacePlugin.setup(setupMock, {});
+    await waitFor(
+      () => {
+        expect(applicationStartMock.navigateToApp).toBeCalledWith(WORKSPACE_FATAL_ERROR_APP_ID, {
+          replace: true,
+          state: {
+            error: 'Cannot find current workspace since it is stale',
```
FriedhelmWS, line 400: Question: is this message appropriate?

src/core/public/workspace/workspaces_service.mock.ts
```diff
@@ -14,11 +14,14 @@ const createWorkspacesSetupContractMock = () => {
   const workspaceList$ = new BehaviorSubject<WorkspaceObject[]>([]);
   const currentWorkspace$ = new BehaviorSubject<WorkspaceObject | null>(null);
   const initialized$ = new BehaviorSubject<boolean>(false);
+  const workspaceError$ = new BehaviorSubject<boolean>(false);
```
SuZhou-Joe, line 17:
```suggestion
  const workspaceError$ = new BehaviorSubject<string>('');
```
Should we make it an observable of string or error?

FriedhelmWS: Done

src/core/public/workspace/workspaces_service.ts
```diff
@@ -74,11 +76,8 @@ export class WorkspacesService implements CoreService<WorkspacesSetup, Workspace
             /**
              * Current workspace is stale
              */
-            this.currentWorkspaceId$.error({
-              reason: WORKSPACE_ERROR.WORKSPACE_IS_STALE,
-            });
-            this.currentWorkspace$.error({
-              reason: WORKSPACE_ERROR.WORKSPACE_IS_STALE,
+            this.workspaceError$.error({
```
SuZhou-Joe, line 79: As there is already an error observable, should we just
```
workspaceError$.next('{some error reason}')
```
?

src/plugins/workspace/public/plugin.ts
```diff
@@ -298,23 +314,25 @@ export class WorkspacePlugin
       core.http.basePath.getBasePath()
     );
 
-    if (workspaceId) {
+    const workspaceError = core.workspaces.workspaceError$;
```
SuZhou-Joe, line 317:
```suggestion
    const workspaceError$ = core.workspaces.workspaceError$;
```
We'd better suffix with a $ to indicate it is a observable.

src/plugins/workspace/public/plugin.ts
```diff
@@ -298,23 +314,25 @@ export class WorkspacePlugin
...
+            core,
+            i18n.translate('workspace.error.workspaceIsStale', {
+              defaultMessage: 'Cannot find current workspace since it is stale',
+            })
+          );
+        }
+      },
+    });
+
+    if (!workspaceError.hasError && workspaceId) {
```
SuZhou-Joe, line 332: What about we move line #332 - #353 into a subscription of workspaceError$? And if there is error populated from the observable, we should the error page, or we show the normal page.

src/core/public/workspace/workspaces_service.ts
FriedhelmWS: I think we still need this to be a error because it will be easier for us to do the `workspaceError$.hasError` check. I think for some security reason use `workspaceError$.hasError.value` may cause some issue.

src/plugins/workspace/public/plugin.ts
FriedhelmWS: I think if we still use workspaceError to trigger the error then probably we should only use it as a indication of error. If we do the "checking for error and if there is no error, then do ..." inside the handler for error, I feel it is a bit out-of-scope for a error observer.

src/plugins/workspace/public/plugin.ts
FriedhelmWS: Done

src/core/public/workspace/workspaces_service.ts
```diff
@@ -56,6 +57,7 @@ export class WorkspacesService implements CoreService<WorkspacesSetup, Workspace
   private currentWorkspace$ = new BehaviorSubject<WorkspaceObject | null>(null);
   private initialized$ = new BehaviorSubject<boolean>(false);
   private client$ = new BehaviorSubject<IWorkspaceClient | null>(null);
+  private workspaceError$ = new BehaviorSubject<string>('');
```
wanglam, line 60: Why we need to add a new `workspaceError$` property? Can we reuse existing `currentWorkspaceId$` and `currentWorkspace$`?

FriedhelmWS: Previously I did this but this causes the `WORKSPACE_IS_STALE` error to be throw out and we unable to catch it, so the error can only be caught by the `FatalErrorsService`. In this way, we cannot use the workspace fatal error page anymore. 

In addition, I think use a additional error subject also make sense, so we can handle any workspace specific error any do the error handling only toward this error subject.

src/core/public/workspace/workspaces_service.ts
FriedhelmWS: Done

src/plugins/workspace/public/plugin.ts
```diff
@@ -686,9 +652,10 @@ export class WorkspacePlugin
     navigation: NavigationPublicPluginStart
   ) {
     if (contentManagement) {
+      const workspaceClient = coreStart.workspaces.client$.getValue() as WorkspaceClient;
```
SuZhou-Joe, line 655: Is the type cast here needed?

src/plugins/workspace/public/plugin.ts
```diff
@@ -298,44 +293,11 @@ export class WorkspacePlugin
...
-          // Add workspace id to recent workspaces.
-          recentWorkspaceManager.addRecentWorkspace(workspaceId);
-        })();
-      }
-    }
+    this.workspaceValidationService.setup(core, workspaceId);
 
     const mountWorkspaceApp = async (params: AppMountParameters, renderApp: WorkspaceAppType) => {
       const [coreStart, { navigation }] = await core.getStartServices();
+      const workspaceClient = coreStart.workspaces.client$.getValue() as WorkspaceClient;
```
SuZhou-Joe, line 300: I am thinking we can get the client from `this.workspaceValidationService.setup`.

src/plugins/workspace/public/services/workspace_validation_service.ts
```diff
@@ -0,0 +1,90 @@
...
+  ChromeStart,
+  CoreSetup,
+  CoreStart,
+  WorkspaceError,
+} from '../../../../core/public';
+import { WorkspaceClient } from '../workspace_client';
+
+export class WorkspaceValidationService {
+  private workspaceClient?: WorkspaceClient;
+  private workspaceId: string | undefined;
```
SuZhou-Joe, line 22: This property seems useless.

src/plugins/workspace/public/plugin.ts
```diff
@@ -298,44 +293,11 @@ export class WorkspacePlugin
...
-              application.navigateToApp(WORKSPACE_DETAIL_APP_ID);
-            }
-            currentAppIdSubscription.unsubscribe();
-          });
-          // Add workspace id to recent workspaces.
-          recentWorkspaceManager.addRecentWorkspace(workspaceId);
-        })();
-      }
-    }
+    this.workspaceValidationService.setup(core, workspaceId);
```
SuZhou-Joe, line 296: should we await the setup?

src/plugins/workspace/public/workspace_client.ts
```diff
@@ -150,6 +156,7 @@ export class WorkspaceClient implements IWorkspaceClient {
         result: null,
       };
     } else {
+      workspaceError$.next(workspaceResp.error!);
```
SuZhou-Joe, line 159: Can we just call `this.workspaces.workspaceError$.next(workspaceResp.error!)`? We do not need to change the public method type by doing so.

src/plugins/workspace/public/services/workspace_validation_service.ts
FriedhelmWS: I think we still need this as we need `recentWorkspaceManager.addRecentWorkspace(this.workspaceId!);`.

src/plugins/workspace/public/plugin.ts
FriedhelmWS: We will need to make `workspaceClient` public so I wonder is it appropriate to directly expose it  from `WorkspaceValidationService`?

src/plugins/workspace/public/plugin.ts
FriedhelmWS: I think so because `getValue()` can possibly return null, but since we do initiate `WorkspaceClient` inside `WorkspaceValidationService` so I think the  type assertion is necessary to tell TypeScript that `workspaceClient` can never be null

src/plugins/workspace/public/services/workspace_validation_service.ts
```diff
@@ -0,0 +1,98 @@
...
+      workspaces.workspaceError$,
+      workspaces.initialized$,
+    ]).subscribe({
+      next: ([reason, initialized]) => {
+        if (!reason && initialized) {
+          /**
+           * If the workspace id is valid and user is currently on workspace_fatal_error page,
+           * we should redirect user to overview page of workspace.
+           */
+          const currentAppIdSubscription = application.currentAppId$.subscribe((currentAppId) => {
```
SuZhou-Joe, line 71: The subscription here looks weird to me, should we add application.currentAppId$ into the combineLatest?

SuZhou-Joe (approved):

wanglam (approved):

SuZhou-Joe (approved):

PR #9483, opened 2025-03-04
[chore] update caniuse and fix some things in cypress

kavilla:
### Description

Caniuse out of date cause one of our tests to fail because it does a raw compare and sees the caniuse complaint.

Also, returning back a promise and then doing stuff in the save search spec.

Also just bumping up the timeout for the get elements just to give it more of a chance.

### Issues Resolved

n/a

## Screenshot

n/a

## Testing the changes

https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9466

## Changelog
- test: update caniuse and return promise and properly chain the async operations

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -16,8 +16,8 @@ module.exports = defineConfig({
     runMode: 2,
     openMode: 0,
   },
-  viewportWidth: 2000,
-  viewportHeight: 1320,
+  viewportWidth: 1920,
+  viewportHeight: 1080,
   env: {
     ENGINE: {
       name: 'default',
@@ -50,7 +50,7 @@ module.exports = defineConfig({
   e2e: {
     baseUrl: 'http://localhost:5601',
     specPattern: 'cypress/integration/**/*.spec.{js,jsx,ts,tsx}',
-    testIsolation: false,
+    testIsolation: true,
     setupNodeEvents,
   },
 });
--- a/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/01/saved_search.spec.js
+++ b/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/01/saved_search.spec.js
@@ -90,14 +90,20 @@ const runSavedSearchTests = () => {
 
       it(`should successfully update a saved search for ${config.testName}`, () => {
         // using a POST request to create a saved search to load
-        postRequestSaveSearch(config);
-        updateSavedSearchAndSaveAndVerify(config, workspaceName, DATASOURCE_NAME, false);
+        cy.wrap(null).then(() => {
+          return postRequestSaveSearch(config).then(() => {
+            updateSavedSearchAndSaveAndVerify(config, workspaceName, DATASOURCE_NAME, false);
+          });
+        });
       });
 
       it(`should successfully save a saved search as a new saved search for ${config.testName}`, () => {
         // using a POST request to create a saved search to load
-        postRequestSaveSearch(config);
-        updateSavedSearchAndSaveAndVerify(config, workspaceName, DATASOURCE_NAME, true);
+        cy.wrap(null).then(() => {
+          return postRequestSaveSearch(config).then(() => {
+            updateSavedSearchAndSaveAndVerify(config, workspaceName, DATASOURCE_NAME, true);
+          });
+        });
       });
     });
   });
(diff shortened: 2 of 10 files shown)


cypress.config.ts
```diff
@@ -50,7 +50,7 @@ module.exports = defineConfig({
   e2e: {
     baseUrl: 'http://localhost:5601',
     specPattern: 'cypress/integration/**/*.spec.{js,jsx,ts,tsx}',
-    testIsolation: false,
+    testIsolation: true,
```
angle943, line 53: i was wondering why this was set to false. If tests pass then yeah this should be set to true.

But to give some background knowledge. In other environments we set this to false. This is because some of our tests require authentication and that does not persist if this is set to true

cypress/utils/apps/query_enhancements/saved.js
```diff
@@ -433,11 +433,11 @@ const getSavedObjectPostBody = (config, workspaceId, datasourceId, indexPatternI
...
 export const postRequestSaveSearch = (config) => {
-  cy.get('@WORKSPACE_ID').then((workspaceId) => {
-    cy.get('@DATASOURCE_ID').then((datasourceId) => {
-      cy.get('@INDEX_PATTERN_ID').then((indexPatternId) => {
+  return cy.get('@WORKSPACE_ID').then((workspaceId) => {
+    return cy.get('@DATASOURCE_ID').then((datasourceId) => {
+      return cy.get('@INDEX_PATTERN_ID').then((indexPatternId) => {
         // POST a saved search
-        cy.request({
+        return cy.request({
```
angle943, line 440: why `returns`?

cypress/utils/commands.js
```diff
@@ -5,13 +5,19 @@
...
+    return cy.get(`[data-test-subj="${testId}"]`, options);
+  }
+);
 
-Cypress.Commands.add('getElementByTestIdLike', (testId, options = {}) => {
-  return cy.get(`[data-test-subj*="${testId}"]`, options);
-});
+Cypress.Commands.add(
+  'getElementByTestIdLike',
+  (testId, options = { timeout: Cypress.config('defaultCommandTimeout') }) => {
```
angle943, line 17: there should be a way to just type the `options` so that it has all the things that the usual `options` for `cy.get()` has. This way if we ever need to add more stuff in the future, we don't need to explictely update the typing

kavilla: i think you already set this value at the cypress config level so i was kind of shooting in the dark with this one and can prolly remove not positive if the option will be passed down nested levels would make sense to me. but basically anyone could override it. unless you are trying to say it would be better to have a set an options that we would let people set and not let them configure it past that

cypress/utils/apps/query_enhancements/saved.js
kavilla: i wasn't originally seeing things wait for the saved search to load which should be async so i'm returning the promise chain so that then we can chain off the response and then proceed to validate. but i could be missing a wait from intercepting somehting.

cypress.config.ts
```diff
@@ -16,8 +16,8 @@ module.exports = defineConfig({
     runMode: 2,
     openMode: 0,
   },
-  viewportWidth: 2000,
-  viewportHeight: 1320,
+  viewportWidth: 1920,
+  viewportHeight: 1080,
```
d-buckner, line 20: Do we have any screenshot comparisons we have to worry about breaking here?

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/01/saved_search.spec.js
```diff
@@ -90,14 +92,20 @@ const runSavedSearchTests = () => {
 
       it(`should successfully update a saved search for ${config.testName}`, () => {
         // using a POST request to create a saved search to load
-        postRequestSaveSearch(config);
-        updateSavedSearchAndSaveAndVerify(config, workspaceName, DATASOURCE_NAME, false);
+        cy.wrap(null).then(() => {
```
d-buckner, line 95: Is this to ensure that the async operations are completed before continuing to the assertions?

angle943:
This is Kiro posting.

Chore PR with merge conflicts. @kavilla — is this still needed given the CI pipeline has evolved since? Please rebase or close.

PR #9490, opened 2025-03-04
[CI] Update Discover Integration Tests S3 Workflow

sejli:
### Description

* Updates paths for running tests in workflow
* Miscellaneous workflow improvements (better naming, removing unnecessary lines)
* Adding session storage clear for reruns

## Changelog

- test: Clear session storage in S3 integ test and update workflow

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


.github/workflows/cypress_workflow_with_s3.yml
```diff
@@ -71,18 +64,11 @@ jobs:
...
-          path: ${{ env.FTR_PATH }}
-          repository: ${{ env.TEST_REPO }}
-          ref: '${{ env.TEST_BRANCH }}'
-
       - name: Setup spec files
         if: ${{ inputs.specs == '' }}
         shell: bash
         run: |
-          DASHBOARDS_SPEC="$(yarn --silent osd:ciGroup10)"
+          DASHBOARDS_SPEC="cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/04/s3_dataset.spec.js"
```
angle943, line 71: should this just be in a separate folder?

angle943 (approved):

.github/workflows/cypress_workflow_with_s3.yml
sejli: I think that would be ideal, since then it wouldn't get run with the other tests during PRs and speed up some of the CI downtime. I don't have too much context on how the current directories are set up for our tests, do you know where exactly would be a good place to add a new directory just for this test? I could then update the `package.json` to have a command for this.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/04/s3_dataset.spec.js
```diff
@@ -100,6 +100,10 @@ const s3DatasetTestSuite = () => {
         });
         afterEach(() => {
           cy.deleteWorkspaceByName(workspace);
+          cy.window().then((win) => {
+            win.localStorage.clear();
```
kavilla, line 104: is `cy.clearLocalStorage()` and `cy.clearAllSessionStorage()`
source: https://docs.cypress.io/api/commands/clearlocalstorage

not available?

also should we consider the clearing out in the `beforeEach`? in case a different test file doesn't follow best practices and clears out this info. so ensuring that this test run is clean. regardless if the previous test cleared out local storage and session storage or not

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/04/s3_dataset.spec.js
```diff
@@ -100,6 +100,10 @@ const s3DatasetTestSuite = () => {
...
+            win.sessionStorage.clear();
+          });
```
kavilla, line 106:
```suggestion
          cy.clearLocalStorage();
          cy.clearAllSessionStorage();
```

are these not available? source: https://docs.cypress.io/api/commands/clearlocalstorage

also should we consider the clearing out in the `beforeEach`? in case a different test file doesn't follow best practices and clears out this info. so ensuring that this test run is clean. regardless if the previous test cleared out local storage and session storage or not

sejli: Was following other tests to see how they were doing it: https://github.com/search?q=repo%3Aopensearch-project%2FOpenSearch-Dashboards%20cy.window()&type=code

.github/workflows/cypress_workflow_with_s3.yml
ananzh: Yeah would prefer another ciGroup for s3 tests.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/04/s3_dataset.spec.js
kavilla: > Other tests' caches shouldn't affect this one.

True as of today, but if another dev decides to add another test file that touches the same stuff this spec tests but dosn't clear it then this test could be potentially impacted

.github/workflows/cypress_workflow_with_s3.yml
ananzh: or `"osd:s3Tests": "echo \"cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/s3_tests/*.spec.js\""`

kavilla (approved):

angle943 (approved):

PR #9498, opened 2025-03-06
move hits location and show number of rows alongside it

angle943:
### Description

- Move the location of HITS to be closer to the table
- Show the number of rows alongside it. Now it will say `Results (X/Y)` where X=Number of rows, Y=Hits
- If Hits does not exist (for ex, SQL), then it will just show `Results (X)`

#### Background

- Ultimately there will be an "export to CSV" button next to this new results section:

Screenshot 2025-02-28 at 9 44 22 AM

- This PR is a prelim PR that just updates the location of the text, and adds the `ResultsActionBar` component that an house all these things (including extensions)
- I will take care of the "Export to CSV" functionality after this PR is merged. And then after that make the right side of the ActionBar be able to receive extensions.

## Screenshot

Before (non-query-enhanced):

Screenshot 2025-03-05 at 6 25 30 PM

- This IMO looks weird. HITS is in the middle, and the timerange text is too close to the border to the left. According to @ashwin-pc , the border should not exist to begin with.

After (non-query-enhanced):

Screenshot 2025-03-07 at 3 59 20 PM

Before (query-enhanced):

Screenshot 2025-03-05 at 6 25 57 PM

After (query-enhanced):

Screenshot 2025-03-07 at 4 00 02 PM

With no hits:

Screenshot 2025-03-07 at 4 00 21 PM

Before (with no time data):

Screenshot 2025-03-07 at 12 59 10 PM

- Notice we are incorrectly showing the toggle for the histogram

After (with no time data):

Screenshot 2025-03-07 at 4 00 11 PM

## Testing the changes

- Unit tests & updated cypress tests

## Changelog

- feat: move HITs counter to be closer to table & show results count

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/discover/public/application/components/results_action_bar/results_action_bar.scss
```diff
@@ -0,0 +1,4 @@
+.dscResultsActionBar {
+  border-bottom: 1px solid $euiColorLightShade;
```
kavilla, line 2: I think we probably want to use the sass variable over pixels as well. Can't remember tho.

And the text size looks like it should be compressed

angle943: I found `$ouiBorderThin` that encompasses everything. i can use this. What do you mean the text size should be compressed? Are you talking about the horizontal spacing between text, or are you talking about the font size itself?

canascar: 1. I'm fine with defining the text as `bold`
2. Advanced settings - can you elaborate on this? Advanced settings for what?

angle943: we discussed offline about advanced settings. We decided not to go with that. i have updated the text to be bold

src/plugins/discover/public/application/view_components/canvas/discover_chart_container.tsx
```diff
@@ -3,99 +3,35 @@
...
-                    window.location.reload();
-                  }}
-                />
-              </EuiPanel>
-            ),
-          },
-        ]}
-      />
-    </EuiPopover>
-  );
```
angle943, line 82: The `DiscoverChart` below now only contains the histogram and related things. We previously had `Hits` in there too, as well as `DiscoverOptions` which is only for non-query-enhanced mode.

Both Hits and DiscoverOptions I moved it to the new `ResultsActionBar`, and now this `DiscoverChart` below is purely for histogram-related components. Therefore, I simplified the logic such that if `!isTimeBased`, we don't render it at all, and I removed all `isTimeBased` conditional logic within `DiscoverChart`

src/plugins/discover/public/application/components/discover_options/discover_options.tsx
```diff
@@ -0,0 +1,69 @@
...
+                  }}
+                />
+              </EuiPanel>
+            ),
+          },
+        ]}
+      />
+    </EuiPopover>
+  );
+};
```
angle943, line 69: I moved this to a separate file. Not adding tests for this because it is going to be deprecated soon anyways

src/plugins/discover/public/application/components/results_action_bar/results_action_bar.tsx
```diff
@@ -0,0 +1,61 @@
...
+              showResetButton={showResetButton}
+              onResetQuery={resetQuery}
+              rows={rows}
+            />
+          </EuiFlexItem>
+        </EuiFlexGroup>
+      </EuiFlexItem>
+      <EuiFlexItem grow={false}>
+        <EuiFlexGroup direction="row" gutterSize="none" justifyContent="flexStart">
+          {/* TODO: We will allow extensions to be added here */}
```
ananzh, line 51: Could you elaborate on what types of extensions are planned for this area?

src/plugins/discover/public/application/components/chart/hits_counter/hits_counter.tsx
```diff
@@ -61,14 +69,20 @@ export function HitsCounter({ hits, showResetButton, onResetQuery }: HitsCounter
...
-            <strong data-test-subj="discoverQueryHits">{hits.toLocaleString()}</strong>{' '}
-            <FormattedMessage
-              id="discover.hitsPluralTitle"
-              defaultMessage="{hits, plural, one {hit} other {hits}}"
-              values={{
-                hits,
-              }}
-            />
+            <strong>
+              <FormattedMessage
```
ananzh, line 73: nit: Do we want to add a aria-label for the entire results display to help screen readers understand the relationship between these numbers?

src/plugins/discover/public/application/components/results_action_bar/results_action_bar.tsx
```diff
@@ -0,0 +1,61 @@
...
+import React from 'react';
+import { EuiFlexGroup, EuiFlexItem } from '@elastic/eui';
+import { HitsCounter } from '../chart/hits_counter';
+import { OpenSearchSearchHit } from '../../doc_views/doc_views_types';
+import { DiscoverOptions } from '../discover_options/discover_options';
+
+export interface DiscoverResultsActionBarProps {
+  hits?: number;
+  showResetButton?: boolean;
+  resetQuery(): void;
```
ananzh, line 17: nit: should we make this optional since it is only used when showResetButton is true

src/plugins/discover/public/application/components/results_action_bar/results_action_bar.tsx
angle943: it can be just about anything, but here is an example:

abc

src/plugins/discover/public/application/components/chart/hits_counter/hits_counter.tsx
angle943: yeah that's a good idea, i'll add

src/plugins/discover/public/application/components/results_action_bar/results_action_bar.tsx
angle943: While i agree with your statement, I wouldn't suggest we go that route for the following reasons:

- It was previously implemented as a required prop (see `HitsCounter` component)
- `ResultsActionBar` only is used from `canvas/index.tsx`, and from there we always pass the prop
- Having both `showResetButton` and `resetQuery` as optionals now adds unwanted complexities. For example, what if `showResetButton` is true while `resetQuery` is not defined? In an ideal situtation, `resetQuery()` should be optional and `showResetButton` shouldn't exist. The component should simply render a reset button if `resetQuery` is defined. But I feel that is unnecessary work and current implementation works fine

src/plugins/discover/public/application/components/chart/chart.tsx
```diff
@@ -146,13 +122,16 @@ export const DiscoverChart = ({
...
+      {chartData && showHistogram && (
         <EuiFlexItem grow={false}>
           <section
             aria-label={i18n.translate('discover.histogramOfFoundDocumentsAriaLabel', {
               defaultMessage: 'Histogram of found documents',
             })}
-            className="dscTimechart"
+            className={classNames('dscTimechart', {
+              // eslint-disable-next-line @typescript-eslint/naming-convention
+              dscTimechart__nonEnhanced: !isEnhancementsEnabled,
```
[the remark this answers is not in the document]
angle943: yeah but hmm, now that you pointed this out, I think I can handle this differently. For context the `query-enhanced` one already has `border-bottom` via this:

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/discover/public/application/components/chart/chart.tsx#L145

which applies `border-bottom` along with other stylings i don't want in the non-query-enhanced view:

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/discover/public/application/components/chart/_histogram.scss#L13-L18

but maybe I can make this part be the one in charge of applying the border. i'll play around with it and see

src/plugins/discover/public/application/components/results_action_bar/results_action_bar.tsx
angle943: actually i change my mind, lets do my last bullet point :D

ananzh (approved):

ananzh (approved):

LDrago27 (approved):

PR #9501, opened 2025-03-06
Initial logic flow for building visualization expressions in discover

abbyhu2000:
### Description
Basic discover visualization for metrics using vega expressions

https://github.com/user-attachments/assets/dc7c906c-c1d8-4077-bd8d-580c772bc3f0

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/discover/public/application/view_components/canvas/index.tsx
```diff
@@ -171,6 +172,7 @@ export default function DiscoverCanvas({ setHeaderActionMenu, history, optionalR
             (isEnhancementsEnabled ? (
               <>
                 <MemoizedDiscoverChartContainer {...fetchState} />
+                <MemoizedDiscoverVisualization {...fetchState} />
```
abbyhu2000, line 175: @d-buckner i can expect that the fetchState contain all the information for a aggregated PromQL returned data right? especially `rows: OpenSearchHit[]` of the `fetchState` here is the data i will use to feed into the visualization render logics

Screenshot 2025-03-06 at 2 11 32 PM

abbyhu2000: also i can expect to be able to find the Prometheus data source information(id, schemas of the fields) through the data frame services, correct? 

```
/**
 * A data frame is a two-dimensional labeled data structure with columns of potentially different types.
 */
export interface IDataFrame {
  type?: DATA_FRAME_TYPES.DEFAULT;
  name?: string;
  schema?: Array<Partial<IFieldType>>;
  meta?: Record<string, any>;
  fields: IFieldType[];
  size: number;
}
```

src/plugins/data/public/search/expressions/discoverSearch.ts
```diff
@@ -0,0 +1,60 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
```
kavilla, line 4: nit: wrong license

src/plugins/discover/public/application/view_components/canvas/index.tsx
kavilla: that schema isn't really used anymore. depends on the prom said if they will returning back the schema with the row/col where the schema is the information about the source regardless of the query or resource.

i would propose still relying on the index pattern fields. the temporary index pattern should fetch the fields in the dataset config. 

that index pattern could point to a metric it's abstracted out enough

src/plugins/discover/public/application/components/visualizations/vislib/line/line_vis_type.ts
```diff
@@ -0,0 +1,22 @@
...
+import { i18n } from '@osd/i18n';
+
+import { VisualizationType } from '../../../../view_components/utils/use_visualization_type';
+import { toExpression } from './to_expression';
+
+export interface LineOptionsDefaults {
+  type: 'line';
+}
+
+export const createLineConfig = (): VisualizationType<LineOptionsDefaults> => ({
```
kavilla, line 15: we have create_histogram_configs  chart/utils/ would it make more sense to add it to that folder?

src/plugins/discover/public/application/view_components/canvas/discover_chart_container.tsx
```diff
@@ -22,7 +22,7 @@ import { DiscoverChart } from '../../components/chart/chart';
 import { QUERY_ENHANCEMENT_ENABLED_SETTING } from '../../../../common';
 import { getNewDiscoverSetting, setNewDiscoverSetting } from '../../components/utils/local_storage';
 
-export const DiscoverChartContainer = ({ hits, bucketInterval, chartData }: SearchData) => {
```
kavilla, line 25: is rows and chartData different?

src/plugins/discover/public/application/view_components/canvas/index.tsx
abbyhu2000: make sense. @d-buckner can we make sure that the metrics are created and mapped using the temporary index pattern flow like PPL?

src/plugins/discover/public/application/components/visualizations/vislib/line/line_vis_type.ts
abbyhu2000: sure good idea

src/plugins/discover/public/application/view_components/canvas/discover_chart_container.tsx
abbyhu2000: This is an example of chart data, which is used to render the discover histogram. 

Screenshot 2025-03-07 at 4 08 31 PM

abbyhu2000: chart data will only be produced if discover histogram exists --> which means chart data will not be produced by sending any PPL query other than the default one

rows will always contain information(with an aggregated value or not) since this is the information we will output in the discover table. Therefore, i think it makes sense to use the rows data here to render visualizations in discover.

abbyhu2000: I also see histogram in discover is rendered by directing using [<Chart>](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/2a9cb948bf42e7424bdd68e83aca28e7a49e8b2a/src/plugins/discover/public/application/components/chart/histogram/histogram.tsx) instead of using expression renderer. But since for discover visualization we might wish to onboard more visualization type in the future, it might be more beneficial to take the expressions renderer approach.

src/plugins/query_enhancements/server/search/promql_search_strategy.ts
```diff
@@ -16,7 +16,7 @@ import {
 } from '../../../data/common';
 
 // Query 500 samples by default
-const TARGET_SAMPLES = 500;
+const TARGET_SAMPLES = 9;
```
d-buckner, line 19: Are we safe to revert this line now?

abbyhu2000: yes

PR #9512, opened 2025-03-10
Use `depth.limit` instead of `nested_objects.limit` as the validation for deeply nested objects

huyaboo:
### Description
This PR changes `nested_objects.limit` to `depth.limit` to ensure that object layers do not exceed the cluster settings. This check also returns a default value even if the call to get the `_cluster/settings` fails.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_importer/server/utils/util.ts
+++ b/src/plugins/data_importer/server/utils/util.ts
@@ -42,7 +42,7 @@ export const determineMapping = (
   };
 };
 
-export const determineType = (
+const determineType = (
   value: any,
   currentNestedCount: number,
   nestedObjectsLimit: number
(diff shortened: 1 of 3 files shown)


src/plugins/data_importer/server/utils/util.ts
```diff
@@ -92,8 +92,35 @@ export const determineNumberType = (value: number) => {
...
+    const defaultSettings = !isNaN(defaultSettingsRaw) ? defaultSettingsRaw : defaultLimit;
+
+    const persistentSettingsRaw = Number(clusterSettings.persistent.indices?.mapping?.depth?.limit);
+    const persistentSettings = !isNaN(persistentSettingsRaw) ? persistentSettingsRaw : defaultLimit;
+
+    const transientSettingsRaw = Number(clusterSettings.transient.indices?.mapping?.depth?.limit);
+    const transientSettings = !isNaN(transientSettingsRaw) ? transientSettingsRaw : defaultLimit;
+
+    // To ensure maximum compatibility, we're only considering the minimum of all 3 settings
+    return min([defaultSettings, persistentSettings, transientSettings]) ?? defaultLimit;
```
ruchidh, line 122: nit

, Math.min , js native method, would have worked fine too, we would not loadash, as lodash export are usaully very expensive.

ruchidh (approved):

PR #9522, opened 2025-03-11
Prometheus manager and search strategy

d-buckner:
### Description
This is to add the prometheus manager and search strategy. With this change you can query Prometheus (after creating a prometheus connection through the datasources API).

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/query_enhancements/common/constants.ts
```diff
@@ -48,6 +48,7 @@ export const URI = {
   PROMQL: '/_plugins/_promql',
   SQL: '/_plugins/_sql',
   ASYNC_QUERY: '/_plugins/_async_query',
+  DIRECT_QUERY_SYNC: '/_plugins/_directquery/_sync',
```
joshuali925, line 51:
```suggestion
  DIRECT_QUERY: {
    QUERY:  '/_plugins/_directquery/_query',
    RESOURCES:  '/_plugins/_directquery/_resources',
  },
```

if you pull the latest change, _sync should be renamed to _query.

currently resources API is actually
```
/_plugins/_directquery/{dataSources}/resources
```

should we make it `/_plugins/_directquery/_resources/{dataSource}` for consistency?

src/plugins/data_source_management/framework/utils/shared.ts
```diff
@@ -57,6 +57,7 @@ export const formatError = (name: string, message: string, details: string) => {
 export const DIRECT_QUERY_BASE = '/api/directquery';
 export const PPL_BASE = `${DIRECT_QUERY_BASE}/ppl`;
 export const PPL_SEARCH = '/search';
+export const PROMQL_SEARCH = '/search';
```
joshuali925, line 60: unused?

src/plugins/data_source_management/framework/utils/shared.ts
```diff
@@ -78,6 +79,7 @@ export const SECURITY_PLUGIN_ACCOUNT_API = '/api/v1/configuration/account';
 
 // Server route
 export const PPL_ENDPOINT = '/_plugins/_ppl';
+export const PROMQL_ENDPOINT = '/_plugins/_promql';
```
joshuali925, line 82: unused?

src/plugins/query_enhancements/server/connections/clients/base_connection_client.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export interface GetResourcesResponse {
+  nextToken?: string;
+  status: string;
+  data: ResourceData[];
+  type: string;
+}
+
+export abstract class BaseConnectionClient {
+  protected abstract client: OpenSearchClient;
```
joshuali925, line 32: this might not be true for all concrete connection clients, for `SqlConnectionClient` we'll use `OpenSearchClient` to make the request, but it can be other underlying client for other concrete connection clients

src/plugins/query_enhancements/server/plugin.ts
```diff
@@ -103,6 +104,7 @@ export class QueryEnhancementsPlugin
     this.logger.info('queryEnhancements: Setup complete');
     return {
       defineSearchStrategyRoute: defineSearchStrategyRouteProvider(this.logger, router),
+      prometheusManager: new PrometheusManager(),
```
joshuali925, line 107: not needed right now but it can be convenient to add `prometheusManager` to route handler context, since query enhancement will also define routes that will eventually get resources using `prometheusManager`

src/plugins/query_enhancements/server/utils/plugins.ts
```diff
@@ -60,9 +60,10 @@ export const OpenSearchEnhancements = (client: any, config: any, components: any
     needBody: true,
   });
   enhancements.promqlQuery = createAction(client, components, {
-    endpoint: URI.PROMQL,
+    endpoint: `${URI.DIRECT_QUERY_SYNC}`,
```
joshuali925, line 63: nit
```suggestion
    endpoint: URI.DIRECT_QUERY_SYNC,
```

also this is used by `core.opensearch.legacy.createClient` to create the legacy client. I heard that legacy clients are deprecating, it might be better to directly use `client.transport.request` to query, like here
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/b93d6fad6843040413ac4b7ce1c43ed299867352/src/plugins/query_enhancements/server/connections/clients/sql_connection_client.ts#L26-L30

src/plugins/query_enhancements/server/search/promql_search_strategy.ts
```diff
@@ -80,3 +81,49 @@ export const promqlSearchStrategyProvider = (
     },
   };
 };
+
+function createDataFrame(rawResponse: PrometheusResponse) {
+  const { result: series } = JSON.parse(rawResponse.result) as PrometheusResult;
```
joshuali925, line 86: i thought opensearch API shouldn't return a JSON string? it should be an already parsed object

d-buckner: I may need to pull the latest changes in the sql plugin. Currently the `results` I'm getting from the opensearch api is a json string.

d-buckner: Would be great to avoid parsing this on the frontend

src/plugins/query_enhancements/server/plugin.ts
d-buckner: Gotcha. I'll make sure to include this in an upcoming PR.

src/plugins/data_source_management/framework/utils/shared.ts
d-buckner: Thanks for catching, removing

src/plugins/data_source_management/framework/utils/shared.ts
d-buckner: Thanks for catching, removing

src/plugins/query_enhancements/server/connections/clients/base_connection_client.ts
d-buckner: Makes sense. I can have this client interface be a generic that the specific connection clients can provide.

src/plugins/query_enhancements/common/constants.ts
d-buckner: > currently resources API is actually
>/_plugins/_directquery/{dataSources}/resources
>should we make it /_plugins/_directquery/_resources/{dataSource} for consistency?

Yeah `/_plugins/_directquery/_resources/{dataSource}` would be great here so it can match the endpoint structure.

joshuali925: updated to make it `/_plugins/_directquery/_resources/{dataSource}`, you can pull and try it later. sample usage
```
GET /_plugins/_directquery/_resources/my_prometheus/api/v1/series?match[]={name=~'.+'}
```

src/plugins/query_enhancements/server/search/promql_search_strategy.ts
joshuali925: i see in the code currently it indeed returns query results as a string. @goyamegh is it intentional or just temporary code to make it work?

src/plugins/query_enhancements/common/constants.ts
d-buckner: Thanks, just pulled the latest sql plugin changes and updated.

PR #9524, opened 2025-03-11
[Auto Suggest] PromQL Autocomplete

paulstn:
### Description

Going into the feature branch, still WIP:
- tests for newest commits
- label value taking in metrics to reduce suggestions
- investigate label functions (e.g. `label_replace()`)

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/public/antlr/promql/code_completion.ts
```diff
@@ -0,0 +1,128 @@
...
+  services,
+}: QuerySuggestionGetFnArgs): Promise<QuerySuggestion[]> => {
+  if (!services || !services.appName || !indexPattern || !dataset) return [];
+  try {
+    const { lineNumber, column } = position || {};
+    const suggestions = getOpenSearchPromQLAutoCompleteSuggestions(query, {
+      line: lineNumber || selectionStart,
+      column: column || selectionEnd,
+    });
+    const prometheusResourceClient = services.data.resourceClientFactory.create('prometheus');
```
d-buckner, line 40: I'd suggest passing an instance of the `PrometheusResourceClient` into `getSuggetions`. That way we can re-use the instance across multiple suggestions.

src/plugins/data/public/antlr/promql/code_completion.ts
```diff
@@ -0,0 +1,128 @@
...
+        ...labels.map((af: string) => ({
+          text: `${af}`,
+          type: monaco.languages.CompletionItemKind.File,
+          detail: SuggestionItemDetailsTags.Table,
+        }))
+      );
+    }
+
+    if (suggestions.suggestLabelValues && suggestions.suggestLabelValues.label) {
+      // TODO: update when we can get metric name passed in
```
d-buckner, line 70: For my context, in the meantime is this suggesting the same labels for all metrics?

paulstn: This is for suggesting the values of a label and right now it's suggesting all values as if we had a label matcher without any metric.
In the conditional above, we're able to suggest the only relevant labels for a metric.

src/plugins/data/public/antlr/promql/code_completion.ts
paulstn: Thanks for the catch! I'll make this change

PR #9528, opened 2025-03-12
[Autocomplete Improvement] Allow query editor to resize

ruchidh:
### Description

This PR introduces a new resizable prop to the CodeEditor component, allowing conditional resizing based on the use case. The SingleLineInput component now explicitly disables resizing, ensuring a fixed-height editor for single-line inputs while keeping the resizing functionality for multi-line editors.

Changes:

- Added a resizable prop to CodeEditor (default: true for backward compatibility).
- Wrapped the MonacoEditor in a Resizable container only when resizable is true.
- Updated SingleLineInput to pass resizable={false} to CodeEditor, preventing unintended resizing.
- Ensured multi-line editors retain the existing resizable behavior.

Why?
Previously, resizing was not working. With this change, the resizing behavior is now configurable per component usage.

Testing:
Verified that multi-line editors remain resizable.
Confirmed that SingleLineInput remains a fixed height and does not show resizing controls.
Ensured no regressions in existing behavior.

## Screenshot

https://github.com/user-attachments/assets/d220ccde-773c-4b6c-a723-7e067ff6dbff

## Changelog
- feat: [Autocomplete Improvement] Allow query editor to resize

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
```diff
@@ -113,11 +115,20 @@ export interface Props {
...
    */
   triggerSuggestOnFocus?: boolean;
-}
 
-export class CodeEditor extends React.Component<Props, {}> {
+  /**
+   * Whether the editor should be manually resizable or not
+   */
+  resizable?: boolean;
+}
```
kavilla, line 123: missing new line character

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
```diff
@@ -175,29 +201,47 @@ export class CodeEditor extends React.Component<Props, {}> {
       }
     });
 
+    const editorComponent = (
```
kavilla, line 204: moving the editor component outside of the return statement in theory shouldn't matter but given the amount of state issues we have and what is being rendered i'm a little bit worried about doing this. have we validated switching languages, have we validated the collapse feature, datat\ that are from a different language, etc

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
```diff
@@ -175,29 +201,47 @@ export class CodeEditor extends React.Component<Props, {}> {
...
+      <MonacoEditor
+        theme="euiColors"
+        language={languageId}
+        value={value}
+        onChange={onChange}
+        editorWillMount={this._editorWillMount}
+        editorDidMount={this._editorDidMount}
+        width="100%"
```
kavilla, line 212: i think this might be a breaking change. the code editor technically is globally accessible. if anyone is added

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
```diff
@@ -154,8 +165,23 @@ export class CodeEditor extends React.Component<Props, {}> {
...
+  // Handles user manual resizing
+  onResize = (_event: React.SyntheticEvent, { size }: { size: { height: number } }) => {
+    this.setState({ editorHeight: size.height }, () => {
+      if (this._editor) this._editor.layout();
+    });
+  };
+
+  // Handles auto-resizing when the parent container changes
+  _updateDimensions = () => {
+    if (this._editor) {
```
kavilla, line 177: will this be impacted by taking this out of the render logic? is this editor state being properly updated ?

changelogs/fragments/9528.yml
```diff
@@ -0,0 +1,2 @@
+fix:
+- Fix the resize issue in monaco editor for discover ([#9528](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9528))
```
kavilla, line 2: this isn't really for discover, it would be for the anyone using the code editor component.

if it is targetted for discover does it make sense that we should just add the react resizable around the query editor logic within the data plugin . it would be safer.

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
ruchidh: it is not inside render logic, the posittion of this method was at bottom, i just moved it to above render so that all class methods present in a single place

changelogs/fragments/9528.yml
ruchidh: it should be everywhere actually, it's a global feature to be used in editor , i will remove discover keyword

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
ruchidh: sorry i didn't get this one, can i get more context on this ?

_someone changing this value and not knowing where the original value came from._

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
ruchidh: I have tested across all languages switch and with single line too. it works fine.
This is just code abstraction,  in a separate variable to make it more readable and optimized.

it will not impact, as it is not a change in logic. This was needed to avoid duplicate code, for multiple scenario, like for single inline we don't need resize but for multiline only we need resize, so abstracted the common code in a variable,

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
ananzh: The width prop is passed to both the Resizable component and the inner div style. The Resizable component expects a numeric width value (width as number), but the width prop in the CodeEditor interface is defined as width?: string | number, meaning it could be either a string (like "100%") or a number. If someone changes the width prop to a string value (like "100%"), it might cause issues with the Resizable component, which expects a number. Additionally, there's no clear documentation or indication that the width prop needs to be a number when resizable is true, which could lead to confusion.

ananzh (commented):
I don't think we need to bring in another dependency for resize. So CodeEditor is wrapped in the `<div className="defaultEditor" ` container. `.defaultEditor` is the CSS class for the container div that wraps everything, including the Monaco editor. The `<CodeEditor>` component is a child of this container and is a React wrapper around the actual Monaco editor. The DefaultEditor container could be resizable. In monaco-editor, there is a prop called automaticLayout. If it is true, then the Monaco editor detects this change and adjusts its internal layout according to the container change.

Can we pull the monaco-editor bump PR and research on top of that? I think we could utilize native browser capabilities without additional dependencies and limit the change to DefaultEditor for query only without affecting other components?

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
ananzh: This is a more valid concern. The _updateDimensions method only calls this._editor.layout() to tell Monaco to recalculate its layout. It doesn't update the component's state (editorHeight).

If the parent container changes size, the ReactResizeDetector will call _updateDimensions, which will tell Monaco to recalculate its layout, but it won't update the editorHeight state. This could potentially lead to inconsistencies between the actual height of the editor and the height stored in the component's state.

ananzh: in practice this might not be an issue. but I think this is a valid concern. why not utilize monaco-editor prop? so we don't need to maintain our own state for the editor height, which eliminates the potential for state synchronization issues.

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
angle943: I think what Ruchi did is fine. Like she said, whether `resizable: true or false`, the parent `div` in the code below gets the passed `width` prop. So that div will get the passed prop, and the code editor will just take `100%` width of whatever was set of the parent

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
angle943: I agree to Ruchi's comments. Since we are now rendering this in two conditionals, i think it makes sense to put it out like what she did. But yeah as long as she has tested thoroughly i think we good

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
```diff
@@ -175,29 +202,47 @@ export class CodeEditor extends React.Component<Props, {}> {
...
+            onResize={this.onResize}
+          >
+            <div style={{ width, height: editorHeight, overflow: 'hidden' }}>{editorComponent}</div>
+          </Resizable>
+
+          {/* Auto Resize */}
+          <ReactResizeDetector handleWidth handleHeight onResize={this._updateDimensions} />
+        </React.Fragment>
+      );
+    }
```
angle943, line 237: - I see a `code_editor.test.tsx`. Could we add some unit tests in there that acommpanies what you have added/changed?
- Does this warrant some extra cypress tests to ensure that things are not broken? ie when changing it and going to back to another language, or changing editor, etc etc

angle943 (changes requested):

ruchidh:
> I don't think we need to bring in another dependency for resize. So CodeEditor is wrapped in the `<div className="defaultEditor" ` container. `.defaultEditor` is the CSS class for the container div that wraps everything, including the Monaco editor. The `<CodeEditor>` component is a child of this container and is a React wrapper around the actual Monaco editor. The DefaultEditor container could be resizable. In monaco-editor, there is a prop called automaticLayout. If it is true, then the Monaco editor detects this change and adjusts its internal layout according to the container change.
> 
> Can we pull the monaco-editor bump PR and research on top of that? I think we could utilize native browser capabilities without additional dependencies and limit the change to DefaultEditor for query only without affecting other components?

So, current version of monaco editor does not support resizable, the default editor is not resizable, i will check in new version of monaco if they have fixed this.

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
ruchidh: Updates the test with new changes.

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
```diff
@@ -113,11 +115,21 @@ export interface Props {
...
+  resizable?: boolean;
 }
 
-export class CodeEditor extends React.Component<Props, {}> {
+export class CodeEditor extends React.Component<Props, { editorHeight: number }> {
   _editor: monaco.editor.IStandaloneCodeEditor | null = null;
 
+  constructor(props: Props) {
+    super(props);
+    this.state = { editorHeight: (props.height as number) || 20 };
```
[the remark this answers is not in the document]
ruchidh: So 20 is the default for one line height of editor that we have been using everywhere. I will assign this in a const

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
angle943: are you referring to the update on the snapshot test? I dont' think that tests the new functionality you added

ananzh:
@ruchidh I checked the automaticLayout property of Monaco editor, which ensures the editor content adjusts properly when the container is resized. It is enabled by default.

I think the most built-in method is to use EuiResizableContainer. Instead of implementing a custom resize solution, we could use this built-in component without additional dependencies. 

It provides vertical and horizontal resize. If we need a side resize drag, we could customize one. But I think we should confirm what is the proper layout.

https://github.com/user-attachments/assets/057569fd-08fc-42dc-8884-bf6824c1e1b1

ruchidh:
> automaticLayout

i think this is the best, we can discard creating external wrapper and instead use the `automaticLayout` property

ruchidh:
this will be added into new editor

PR #9529, opened 2025-03-12
[Discover] fix: Correctly load saved search from snapshot URL

Maosaic:
### Description

Previously, when loading a snapshot URL that contains a saved search with temporary editing, the temporary editing content will not be loaded. This is because when the app getting initialize, status store in URL snapshot will be loaded first, later the async fetch result from saved search will override those temporary editing.

Ideally when loading a saved search, if it’s loaded by user click on saved search panel UI, we want to wipe current state and use the saved query states, if the saved search is loaded from user directly paste a url in browser, we want use the url snapshot state instead.

To build the ideal experience, this PR contains changes below:

1. When user click a saved search from UI panel, clean filter and query
2. When loading saved search, saved filters will apply on top of existing filters 
3. When loading saved search, only apply query from saved search if there is no existing query

## Screenshot

Before

https://github.com/user-attachments/assets/4b826da5-78a8-4741-b78d-9edce1c951c6

After

https://github.com/user-attachments/assets/b12092e4-7f70-49f3-a6a9-52e827d0b36a

## Changelog

- fix: Correctly load saved search from snapshot URL

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/view_components/utils/use_search.ts
+++ b/src/plugins/discover/public/application/view_components/utils/use_search.ts
@@ -398,11 +398,14 @@ export const useSearch = (services: DiscoverViewServices) => {
 
   // Get savedSearch if it exists
   useEffect(() => {
-    (async () => {
+    const loadSavedSearch = async () => {
       const savedSearchInstance = await getSavedSearchById(savedSearchId);
 
-      const query =
-        savedSearchInstance.searchSource.getField('query') || data.query.queryString.getQuery();
+      const savedSearchQuery = savedSearchInstance.searchSource.getField('query');
+      const dataQuery = data.query.queryString.getQuery();
+
+      // Use eixisting query, if not, use query from saved search
+      const query = dataQuery.query ? dataQuery : savedSearchQuery ?? dataQuery;
 
       const isEnhancementsEnabled = await uiSettings.get('query:enhancements:enabled');
       if (isEnhancementsEnabled && query.dataset) {
@@ -429,10 +432,11 @@ export const useSearch = (services: DiscoverViewServices) => {
       // sync initial app filters from savedObject to filterManager
       const filters = cloneDeep(savedSearchInstance.searchSource.getOwnField('filter'));
 
-      let actualFilters: any[] = [];
+      // merge filters in saved search with exisiting filters in filterManager
+      const actualFilters = cloneDeep(filterManager.getAppFilters());
 
       if (savedQuery) {
-        actualFilters = data.query.filterManager.getFilters();
+        actualFilters.push.apply(actualFilters, data.query.filterManager.getFilters());
       } else if (filters !== undefined) {
         const result = typeof filters === 'function' ? filters() : filters;
         if (result !== undefined) {
@@ -454,7 +458,9 @@ export const useSearch = (services: DiscoverViewServices) => {
           }
         );
       }
-    })();
+    };
+
+    loadSavedSearch();
     // This effect will only run when getSavedSearchById is called, which is
     // only called when the component is first mounted.
     // eslint-disable-next-line react-hooks/exhaustive-deps
--- a/src/plugins/discover/public/application/view_components/utils/use_search.ts
+++ b/src/plugins/discover/public/application/view_components/utils/use_search.ts
@@ -400,12 +400,13 @@ export const useSearch = (services: DiscoverViewServices) => {
   useEffect(() => {
     const loadSavedSearch = async () => {
       const savedSearchInstance = await getSavedSearchById(savedSearchId);
-
-      const savedSearchQuery = savedSearchInstance.searchSource.getField('query');
       const dataQuery = data.query.queryString.getQuery();
+      const defaultQuery = data.query.queryString.getDefaultQuery();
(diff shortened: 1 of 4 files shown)


src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -398,11 +398,15 @@ export const useSearch = (services: DiscoverViewServices) => {
 
   // Get savedSearch if it exists
   useEffect(() => {
-    (async () => {
+    const loadSavedSearch = async () => {
       const savedSearchInstance = await getSavedSearchById(savedSearchId);
+      const dataQuery = data.query.queryString.getQuery();
+      const defaultQuery = data.query.queryString.getDefaultQuery();
```
kavilla, line 404: oh so you are saying it's always the default query string? which i think would be empty?

kavilla: this led me into a rabbit but still not able to fix it. perhaps you can look into these potential issues that could be causing this problem to be solved to be give difficult.

https://github.com/kavilla/OpenSearch-Dashboards-1/commit/d2fecdf13e82800791afe6b3dd8f5b2f5861e051

one thing this is weird  is that we are pushing dataset to url state no matter if query enhancements enabled or not

kavilla (approved):

angle943 (approved):

src/plugins/discover/public/application/view_components/utils/use_search.ts
Maosaic: I'm doing this because, whenever we clear the query or initialize QueryStringManager, we set the query to default.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/query/query_string/query_string_manager.ts#L56

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/query/query_string/query_string_manager.ts#L213

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -429,10 +433,11 @@ export const useSearch = (services: DiscoverViewServices) => {
       // sync initial app filters from savedObject to filterManager
       const filters = cloneDeep(savedSearchInstance.searchSource.getOwnField('filter'));
 
-      let actualFilters: any[] = [];
+      // merge filters in saved search with exisiting filters in filterManager
+      const actualFilters = cloneDeep(filterManager.getAppFilters());
```
virajsanghvi, line 437: does this always result in an array so you can push later?

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -429,10 +433,11 @@ export const useSearch = (services: DiscoverViewServices) => {
...
 
       if (savedQuery) {
-        actualFilters = data.query.filterManager.getFilters();
+        actualFilters.push.apply(actualFilters, data.query.filterManager.getFilters());
```
virajsanghvi, line 440: are duplicate filters a potential concern here?

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -398,11 +398,15 @@ export const useSearch = (services: DiscoverViewServices) => {
...
+    const loadSavedSearch = async () => {
       const savedSearchInstance = await getSavedSearchById(savedSearchId);
+      const dataQuery = data.query.queryString.getQuery();
+      const defaultQuery = data.query.queryString.getDefaultQuery();
+      const isDataQueryDefault = dataQuery.query === defaultQuery.query;
+      const savedSearchQuery = savedSearchInstance.searchSource.getField('query');
 
-      const query =
-        savedSearchInstance.searchSource.getField('query') || data.query.queryString.getQuery();
+      // Use eixisting query, if eixisting query match default, use query from saved search
```
virajsanghvi, line 408: nit: existing

What if the user just happened to enter the default query? I guess this is okay because we explicitly reset the querystring? Is dataQuery.query always defined? Do we have any state that determines when the query is user edited?

virajsanghvi (commented):
Know this is closed, left some questions to better understand

src/plugins/discover/public/application/view_components/utils/use_search.ts
Maosaic: `getAppFilters` should return empty array when no filters exist

src/plugins/discover/public/application/view_components/utils/use_search.ts
Maosaic: we remove duplicate filter when merge in new filter 

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/query/filter_manager/filter_manager.ts#L85-L87

src/plugins/discover/public/application/view_components/utils/use_search.ts
Maosaic: Yes, you're right, we don't have any flag state if a query has been user edited after saved. if a saved search url happened to be the default query with "unsaved" user edit, those "unsaved" user edit won't load with the url.

PR #9530, opened 2025-03-12
download CSV from discover page

angle943:
### Description

- adds the ability to export to CSV from the discover page

## Screenshot

- When hits are available

Screenshot 2025-03-11 at 8 24 47 PM

- when hits are not available

Screenshot 2025-03-11 at 8 24 55 PM

## Testing the changes

- I have added both unit tests and cypress test

## Changelog

- feat: add the ability to export to CSV from the discover page

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -352,6 +350,52 @@ export const useSearch = (services: DiscoverViewServices) => {
...
+      const dataset = indexPattern;
+      if (!dataset) {
+        throw new Error('Dataset not found');
+      }
+
+      if (!validateTimeRange(timefilter.getTime(), toastNotifications)) {
+        throw new Error('Invalid time range');
+      }
+
+      const searchSource = await updateSearchSource({
```
kavilla, line 365: am i able to click the export to csv button 10 times and it's not aborting the previous query?

angle943: The "Export to csv" button is disabled while it is loading, so you won't be able to do that
Screenshot 2025-03-12 at 3 12 45 PM

angle943: but do we want to abort if while it is loading, and the user navigates away to a different page? I am not sure how we want to handle this. But the current code will finish the loading & download even from a different page

src/plugins/discover/public/application/components/download_csv/constants.ts
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * The upper limit option of the "Max" download option
+ */
+export const MAX_DOWNLOAD_CSV_COUNT = 10000;
```
kavilla, line 9: this is a little nested for a constants comparable to other constants file. is it worth bringing this up to a higher location to avoid situations where if we do add more constants all the constants are accessible at from the same location.

kavilla: the other thing is that `10,000` is the default page size from OpenSearch if the user did not configure their page size in the index settings.

Technically users could configure that value to be 20,000 however that is remarked to be risky as that could impact the stability of the cluster.

is exporting to csv avaialbe for all data types? is there a limitation set by different data types? like can i export to csv S3 10,000 records at once? will there be any performance concern?

src/plugins/discover/public/application/components/download_csv/use_download_csv.ts
```diff
@@ -0,0 +1,118 @@
...
+    const displayedColumns = getLegacyDisplayedColumns(
+      state.discover.columns,
+      indexPattern,
+      uiSettings.get(DOC_HIDE_TIME_COLUMN_SETTING),
+      uiSettings.get(UI_SETTINGS.SHORT_DOTS_ENABLE)
+    );
+    return displayedColumns.map((column) => column.name);
+  });
+
+  const downloadCsvForOption = async (option: DownloadCsvFormId) => {
```
ruchidh, line 84: nit: can we try to wrap it inside useCallback() ?

src/plugins/discover/public/application/components/download_csv/constants.ts
kavilla: the other proposal that would probably empower the users to be in control, without an arbitrary 10k limit would be adding a new ui setting for this

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/discover/server/ui_settings.ts#L61

we could add under the sample size setting like `MAX_CSV_SIZE_SETTING` and set that value to be `10000` by default.

but then if the user wants to increase it to get more or a admin wants to reduce the amount to avoid potential throttle issues they could

angle943: - Sure I could move it. The general approach i took was it should be contained in the deepest level that is accessible by all consumers, which for my use case is this `/download_csv/` area, but if there is a better approach that aligns with how things are architected in OSD, i'm all ears. Do you have a place in mind?

- For your second comment, I think a context that can help this discussion is that the "Max Available" option will not appear if the dataset does not return back HITS number. So currently it only exists for DQL and Lucene (that I am aware). I have reached out to @sejli about getting me an s3 connection, and so I will test that once that is ready, but I think S3 dataset type does not return hits? In which case the `Max Available` option will not be available, and only the "Visible" option

Also the `Max Available` option will not appear if the currently queried result is bigger than 10,000. So say they configured to query for 20,000. Then when the user clicks the "Export as CSV" option, only `Visible (20,000)` option will exist.

The 10000 is just the upper limit the product team decided that we will allow the "download" button to query for. And if the hits count is below that, then the "Max available" will show that hit count instead of 10000:

Screenshot 2025-03-12 at 4 21 08 PM

src/plugins/discover/public/application/view_components/utils/use_search.ts
kavilla: we dont really handle the navigate away and aborting calls well as of right now either. 

but at least in the context of discover, if for example i typed an "expensive" query that we often see users doing:
- `*` as an index pattern = OpenSearch must check every index in cluster (thousands, each index has millions or 10s of millions of documents)
    - we actually already run into issues today with the sample size of 500 
- `-10y time range` = Must scan years of historical data (possibly in cold storage)
- `personal.user_id: *e49*` wildcard = Forces full scan, can't use index optimizations
    - High cardinality field = Lots of unique values to check
 
 which will result in CPU spikes, memory pressure, slow response, possible timeouts
 so the i dont think we should disable it but we should let change into a button that will allow the user to abort the current query. if they realized it's causing too much pressure on the system.

src/plugins/discover/public/application/components/results_action_bar/results_action_bar.tsx
```diff
@@ -44,6 +53,11 @@ export const DiscoverResultsActionBar = ({
               rows={rows}
             />
           </EuiFlexItem>
+          {indexPattern && rows?.length ? (
+            <EuiFlexItem grow={false}>
```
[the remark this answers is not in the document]
angle943: I'm personally not a big fan of that as that type of logic sometimes led to bugs. I like being explicit. And I also see that that type of logic led to bugs in OSD too: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9153/files

I could change it it `!!indexPattern && (rows?.length ?? 0) > 0) && ()}` to avoid that but at that point what is wrong with what i have?

src/plugins/discover/public/application/components/download_csv/use_download_csv.ts
```diff
@@ -0,0 +1,118 @@
...
+  rows,
+  indexPattern,
+  onLoading,
+  onSuccess,
+  onError,
+}: UseDiscoverDownloadCsvProps) => {
+  const { fetchForMaxCsvOption } = useDiscoverContext();
+  const { uiSettings } = getServices();
+  const [isLoading, setIsLoading] = useState(false);
+  const displayedColumnNames = useSelector((state) => {
```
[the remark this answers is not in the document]
angle943: The `useSelector` is itself a hook and we are not able to call it within a `useMemo`. I also think it is memoized though under the hood? I could be wrong about this tho

src/plugins/discover/public/application/components/download_csv/download_csv_callout.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    <EuiCallOut
+      className="dscDownloadCsv__callout"
+      data-test-subj="dscDownloadCsvCallout"
+      color="warning"
+    >
+      <EuiText size="s" className="dscDownloadCsv__calloutText">
+        <FormattedMessage
+          id="discover.downloadCsvCallout"
+          defaultMessage="There is a limit of 10,000 total result downloads."
```
kavilla, line 20: should this pull from the constant?

src/plugins/discover/public/application/components/download_csv/use_download_csv.ts
angle943: I initially started out with this wrapped under a callback, but the dependency for the callback became too long. I just tried on my editor and this is what I get:

```
[displayedColumnNames, fetchForMaxCsvOption, hits, indexPattern, onError, onLoading, onSuccess, rows, uiSettings]
```

At this point, is it worth even wrapping it on callback? and memoization is not free, so I personally only memoize when it makes sense to

src/plugins/discover/public/application/components/download_csv/download_csv_callout.tsx
angle943: ah good point, yeah i will update

src/plugins/discover/public/application/components/download_csv/use_download_csv.ts
```diff
@@ -0,0 +1,118 @@
...
+      return '';
+    }
+
+    return formattedValue;
+  });
+};
+
+export const saveDataAsCsv = (csvData: string) => {
+  const blob = new Blob([csvData], { type: 'text/csv;charset=utf-8' });
+  const fileName = `opensearch_export_${moment().format('YYYY-MM-DD')}`;
```
kavilla, line 59: should we be pulling some of this from advanced settings?

Screenshot 2025-03-12 at 4 33 22 PM

src/plugins/discover/public/application/components/download_csv/download_csv.tsx
```diff
@@ -0,0 +1,64 @@
...
+        closePopover={closePopover}
+        panelPaddingSize="none"
+      >
+        <DiscoverDownloadCsvPopoverContent
+          rowsCount={rows?.length || 0}
+          hitsCount={hits || 0}
+          downloadForOption={handleDownloadCsvForOption}
+        />
+      </EuiPopover>
+      <EuiGlobalToastList toasts={toasts} dismissToast={onDismiss} toastLifeTimeMs={6000} />
```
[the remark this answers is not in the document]
angle943: ah i think i found it: notifications:lifetime:info . Let me just make sure its the proper one to use and i'll update if it is so. Thanks!

src/plugins/discover/public/application/components/download_csv/use_download_csv.ts
angle943: I think you are talking about the formatting of the filename? I'm a bit reluctant on that, but open for debate. But here's my rationale:

Because we are saving this to a file, we have to respect the allowed characters that a filename can have. So things like `/` and such are not possible. So if we want to respect the user configured date format, I feel like we have to consider all these exceptions, and I didn't want to deal with it. Especially since its just a file name and they can rename it as they choose.

But open to suggestions!

if you were talking about something else, let me know!

src/plugins/discover/public/application/components/download_csv/constants.ts
angle943: And @kavilla i just saw your point about making 10000 a UI setting. I like that idea, but a few concerns:

```
the other thing is that 10,000 is the default page size from OpenSearch if the user did not configure their page size in the index settings.

Technically users could configure that value to be 20,000 however that is remarked to be risky as that could impact the stability of the cluster.
```

The EXPORT_CSV_LIMIT setting should respect this above setting you are talking about, right? Where does this above default 10,000 limit live? And if it is set to 10,000, and user updates EXPORT_CSV_LIMIT to 20,000, how do we make it respect the upper limit of 10,000 default page size (do we even want the csv limit to respect this?)

angle943: Screenshot 2025-03-12 at 5 08 21 PM

Just tested with S3, yeah no option to download max available. I noticed that the query is very slow for S3, but once it has loaded, the ability to download the "visible" will be pretty much instant because no additional api calls are being made for that option

abbyhu2000:
Nice feature! Just one small comment: I see in the download_csv folders have a lot of files and it seems like some files like <[download_csv_button.tsx](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9530/files#diff-df2928dba81208e35a723a58d6a0bddcadcb0204fb3cefdae31878cd303f8ded)> and <[download_csv_callout.tsx](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9530/files#diff-5d1ea96d11589258218113d065c9b6e0c3ccce334ed16589b9ef91750378b71e)> are fairly simple single component file without any extra logics, imo is it worth it to combine some of them to increase some readability and faster navigation?

angle943:
@abbyhu2000 hey! this is a contentious topic, but love to open up the debate.

You are right, `DiscoverDownloadCsvButton` and `DownloadCsvCallout` is pretty simple. But I would argue these should be separate. For context to everyone:

- `DiscoverDownloadCsvButton`, it is passed as a prop to the `EuiPopover`'s button prop in `download_csv.tsx`.
- `DownloadCsvCallout` is rendered in `DownloadcsvPopoverContent` component and is hidden behind a conditional `{showMaxOption && <DiscoverDownloadCsvCallout />}`.

I would argue that these should be separate for the following reason:

- Within the context of `<DownloadCsv />`, you whouldn't worry about the implementation details of how the button looks like and what type of props it has. Similarly, within context of `DownloadCsvPopoverContent`, you shouldn't worry about the UI and text and stuff of the PopoverContent. Just whehther or not it should render it or not.
- Both the parent components `<DownloadCsv />` and `<DownloadCsvPopoverContent />` can blow up and get big over time, so I think it is worth abstracting these things out.
- The unit tests can be more easily scoped. For example, the parent's unit test doesn't have to worry about the inner workings of these sub components, and the sub components can have their own unit tests
- Due to how React renders the DOM tree using a virtual DOM tree, this can potentially be an optimization technique as the whole of the parent components don't need to re-render, if nothing within the child components haven't re-rendered.


Due to the above, I would personally vote that we split it up where we can without over doing it. Of course it can be argued that I might be over doing it here, and it is up for debate, but for the above reasons I think keeping them separate makes more sense

src/plugins/discover/public/application/components/download_csv/download_csv_callout.tsx
angle943: I updated this to use the constant

src/plugins/discover/public/application/components/download_csv/download_csv.tsx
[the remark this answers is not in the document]
angle943: I ended up removing this `<EuiGlobalToastList />` and leveraging the `notificationService` instead so that toasts can still exist in the DOm when you navigate away from the discover page

src/plugins/discover/public/application/view_components/utils/use_search.ts
angle943: I added the ability to abort a request:


https://github.com/user-attachments/assets/7297f78a-e24b-4e9d-a8bc-11d53404b77b

src/plugins/discover/public/application/components/download_csv/constants.ts
angle943: About putting the 10,000 into user settrings, it is tied to the settings per index: https://opensearch.org/docs/latest/install-and-configure/configuring-opensearch/index-settings/#dynamic-index-level-index-settings

We discussed with the team and decided that we will keep this hardcoded for now, and if users request this to be configurable we will fast follow

LDrago27 (approved):

src/plugins/discover/public/application/components/download_csv/use_download_csv.ts
Maosaic: I see the `quoteValues` and `separator` setting being read from L100-L101
```
      const csvData = unparse(
        { fields: displayedColumnNames, data: csvRowData },
        {
          quotes: uiSettings.get('csv:quoteValues', true),
          delimiter: uiSettings.get('csv:separator', ','),
          header: true,
        }
      );

Maosaic (approved):

d-buckner (approved):

PR #9532, opened 2025-03-12
[Query enhancement]feat: Append prompt for query assistant in request payload

raintygao:
### Description

Append prompt for query assistant in request payload

## Screenshot

No UI change.

## Testing the changes

Use query assistant and try it with `fail` or `error` in question then see request payload

## Changelog

- feat: Append prompt for query assistant in request payload

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
+++ b/src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
@@ -15,7 +15,7 @@ import { useOpenSearchDashboards } from '../../../../opensearch_dashboards_react
 import { QueryAssistParameters } from '../../../common/query_assist';
 import { getStorage } from '../../services';
 import { useGenerateQuery } from '../hooks';
-import { getPersistedLog, AgentError, ProhibitedQueryError } from '../utils';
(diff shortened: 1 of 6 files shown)


SuZhou-Joe (approved):

src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
```diff
@@ -67,7 +67,8 @@ export const QueryAssistBar: React.FC<QueryAssistInputProps> = (props) => {
     previousQuestionRef.current = inputRef.current.value;
     persistedLog.add(inputRef.current.value);
     const params: QueryAssistParameters = {
-      question: inputRef.current.value,
+      // Only append query prompt in request payload
+      question: appendQueryPrompt(inputRef.current.value),
```
ruanyl, line 71: Can we append it at server side? So we just need to do it once if the API get called in different place.

raintygao: For this case I would prefer to limit it in frontend because we may not want to append this prompt in each place, i'm concerned it may add maintain cost if we have different strategy in each place in future. What's your opinion?

ruanyl (approved):

PR #9533, opened 2025-03-12
Add better error handling and messages

huyaboo:
#### Before
- Errors were not really handled well. Parsing completely stops after the first error
- Mappings were not used when ingesting
- Several issues with predicting the mapping for `""`

#### After
- Specify the error message in the toast
- Specify which rows in the file/text were invalid
- Throw handled exceptions for malformed file inputs
- Moved around some logic in `util.ts`

## Changelog

- fix: Fix the mapper and enhance the error handling

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_importer/server/utils/util.ts
```diff
@@ -31,70 +29,61 @@ export const validateEnabledFileTypes = (fileTypes: string[], fileParsers: FileP
...
   }
 };
 
-const isValidDate = (date: string) => {
-  return (
-    moment(date, moment.ISO_8601, true).isValid() || moment(date, moment.RFC_2822, true).isValid()
+const flattenObject = (obj: Record<string, any>) => {
+  const result: Record<string, any> = {};
+  for (const key of Object.keys(obj)) {
+    if (obj[key] && typeof obj[key] === 'object') {
```
ruchidh, line 57: may be optimise the code here, like Object.assign instead of looping over keys manually

ruchidh (approved):

src/plugins/data_importer/server/utils/util.ts
huyaboo: i need to change the keys before adding them to the new object so assign might not help

PR #9550, opened 2025-03-14
Complete prometheus manager wiring/modeling

d-buckner:
### Description
This is to update the prometheus manager to use the new resource API routes and better model the query and response types

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/query_enhancements/server/routes/resources/routes.ts
```diff
@@ -0,0 +1,49 @@
...
+          ]),
+          resourceName: schema.maybe(schema.string()),
+        }),
+        query: schema.object({}, { unknowns: 'allow' }),
+      },
+    },
+    async (context, request, response) => {
+      const { dataConnectionId, dataConnectionType, resourceType, resourceName } = request.params;
+      if (dataConnectionType === 'prometheus') {
+        const resources = await new PrometheusManager().getResources(context, request, {
```
d-buckner, line 39: @joshuali925 I know you mentioned this could be added to the route handler context. Can you share where that's defined? Happy to use a shared instance in context in follow up.

joshuali925: you can define route context like this https://github.com/opensearch-project/OpenSearch-Dashboards/blob/e8cf0ca4527f55b451e4f9d0af02c82451e0a73e/src/plugins/query_enhancements/server/plugin.ts#L78-L85

then it's accessible in the route handler's `context.<context-name>` variable

It actually must be a shared instance, because the overridden client from other plugins won't work if every time the manager is newly created. might be better if you can make it a singleton

d-buckner: Yeah, that's what I was worried about as well. Will export a singleton, register in this context, and then consume from handler context. I'll send out a commit shortly to address.

d-buckner: I've updated the prometheus manager so it's only vended as a singleton. I didn't get a chance to update the router context, but will include in next change.

PR #9566, opened 2025-03-18
[Cross-Cluster-Search] Show cross cluster connections in Data Sources and Workspaces page

LDrago27:
### Description

This PR aims at showing the cross cluster connections for Opensearch Connections in Data Sources and Workspaces Page. 

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9578

## Screenshot

Data Sources pages:
image
image

Workspace Creation page:
image
image

## Testing the changes

- Create a Domain with remote connections (https://opensearch.org/docs/latest/search-plugins/cross-cluster-search/, https://docs.aws.amazon.com/opensearch-service/latest/developerguide/cross-cluster-search.html)
- Run the application in query enhancements turned on. (`yarn start:enhancements --no-base-path` & with following options in opensearh_dashboards.yml `
workspace.enabled: true
opensearch.ignoreVersionMismatch: true
data_source.enabled: true
`)
- Add the domain with remote connections as a Data Source to the Application
- Navigate to the data sources and workspaces page to validate that the remote cluster connections are being shown

## Changelog

- feat: Show cross cluster connections in Data Sources and Workspaces page
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source/common/data_sources/types.ts
+++ b/src/plugins/data_source/common/data_sources/types.ts
@@ -60,4 +60,5 @@ export enum DataSourceEngineType {
   OpenSearchServerless = 'OpenSearch Serverless',
   Elasticsearch = 'Elasticsearch',
   NA = 'No Engine Type Available',
+  OpenSearchCrossCluster = 'OpenSearch(Cross-cluster search)',
 }
--- a/src/plugins/data_source_management/public/components/data_source_table/data_source_table.scss
+++ b/src/plugins/data_source_management/public/components/data_source_table/data_source_table.scss
@@ -1,0 +1,24 @@
+// Hide the table header for the expanded table
+.data-source-expanded-table thead {
+  display: none;
+}
+
+// Remove padding for the first cell in expanded rows
+.data-source-table [id^="row"][id$="expansion"] > td:first-child > div:first-child {
+  padding: 0;
+}
+
+// Remove top border for the first expanded row
+.data-source-table-expanded-row:first-child > td {
+  border-top: 0;
+}
+
+// Remove bottom border for the last expanded row
+.data-source-table-expanded-row:last-child > td {
+  border-bottom: 0;
+}
+
+// Add left margin to the title in expanded rows
+.data-source-table-expanded-row_title {
+  margin-left: 20px;
+}
--- a/src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
+++ b/src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
@@ -41,6 +41,8 @@ import {
 } from '../utils';
 import { LoadingMask } from '../loading_mask';
 import { DEFAULT_DATA_SOURCE_UI_SETTINGS_ID } from '../constants';
+import './data_source_table.scss';
+import { DataSourceEngineType } from '../../../../data_source/common/data_sources';
 
 /* Table config */
 const pagination = {
@@ -86,6 +88,9 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
   const currentWorkspace = useObservable(workspaces ? workspaces.currentWorkspace$ : of(null));
   const isDashboardAdmin = !!application?.capabilities?.dashboards?.isDashboardAdmin;
   const canAssociateDataSource = !!currentWorkspace && isDashboardAdmin;
+  const [itemIdToExpandedRowMap, setItemIdToExpandedRowMap] = useState<
+    Record<string, React.ReactNode>
+  >({});
 
   /* useEffectOnce hook to avoid these methods called multiple times when state is updated. */
   useEffectOnce(() => {
@@ -124,11 +129,38 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
     [notifications.toasts]
   );
 
+  const toggleDetails = (item: DataSourceTableItem) => {
+    const itemIdToExpandedRowMapValues = { ...itemIdToExpandedRowMap };
+    if (itemIdToExpandedRowMapValues[item.id]) {
+      delete itemIdToExpandedRowMapValues[item.id];
+    } else {
+      itemIdToExpandedRowMapValues[item.id] = (
+        <EuiInMemoryTable
+          items={item?.relatedConnections ?? []}
+          itemId="id"
+          columns={columns}
+          className="data-source-expanded-table"
+          rowProps={{
+            className: 'data-source-table-expanded-row',
+          }}
+        />
+      );
+    }
+    setItemIdToExpandedRowMap(itemIdToExpandedRowMapValues);
+  };
+
   const fetchDataSources = useCallback(async () => {
     setIsLoading(true);
     try {
       const response = await getDataSources(savedObjects.client);
-      const finalData = await fetchDataSourceConnections(response, http, notifications, false);
+      const finalData = await fetchDataSourceConnections(
+        response,
+        http,
+        notifications,
+        false,
+        false,
+        true
+      );
       setDataSources(finalData);
     } catch (error) {
       setDataSources([]);
@@ -250,31 +282,47 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
   /* Table columns */
   const columns: Array<EuiBasicTableColumn<DataSourceTableItem>> = [
     {
+      align: 'left',
+      width: '40px',
+      isExpander: true,
+      render: (item: DataSourceTableItem) =>
+        item?.relatedConnections?.length ? (
+          <EuiButtonIcon
+            onClick={() => toggleDetails(item)}
+            data-test-subj="expandCollapseButton"
+            aria-label={itemIdToExpandedRowMap[item.id] ? 'Collapse' : 'Expand'}
+            iconType={itemIdToExpandedRowMap[item.id] ? 'arrowUp' : 'arrowDown'}
+          />
+        ) : null,
+    },
+    {
       field: 'title',
       name: i18n.translate('dataSourcesManagement.dataSourcesTable.dataSourceField', {
         defaultMessage: 'Data source',
       }),
-      render: (
-        name: string,
-        index: {
-          id: string;
-          tags?: Array<{
-            key: string;
-            name: string;
-          }>;
-        }
-      ) => (
-        <>
-          <EuiButtonEmpty size="xs" {...reactRouterNavigate(history, `${index.id}`)} flush="left">
-            {name}
-          </EuiButtonEmpty>
-          {index.id === defaultDataSourceId ? (
-            <EuiBadge iconType="starFilled" iconSide="left">
-              Default
-            </EuiBadge>
-          ) : null}
-        </>
-      ),
+      render: (name: string, item: DataSourceTableItem) => {
+        return (
+          <>
+            <EuiButtonEmpty
+              size="xs"
+              {...reactRouterNavigate(history, `${item.id}`)}
+              className={
+                item.type === DataSourceEngineType.OpenSearchCrossCluster
+                  ? 'data-source-table-expanded-row_title'
+                  : ''
+              }
+              flush="left"
+            >
+              {name}
+            </EuiButtonEmpty>
+            {item.id === defaultDataSourceId ? (
+              <EuiBadge iconType="starFilled" iconSide="left">
+                Default
+              </EuiBadge>
+            ) : null}
+          </>
+        );
+      },
       dataType: 'string' as const,
       sortable: ({ title }: { title: string }) => title,
     },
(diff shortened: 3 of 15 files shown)


src/plugins/data_source/common/data_sources/types.ts
```diff
@@ -60,4 +60,5 @@ export enum DataSourceEngineType {
   OpenSearchServerless = 'OpenSearch Serverless',
   Elasticsearch = 'Elasticsearch',
   NA = 'No Engine Type Available',
+  OpenSearchCrossCluster = 'OpensSearch(Cross-cluster search)',
```
Hailong-am, line 63: what's the difference between `OpenSearchCrossCluster` and `OpenSearch`?

LDrago27: There is no inherent difference between OpenSearchCrossCluster & OpenSearch. They both refer to a Opensearch cluster. However with the above change we want to show users all the remote Opensearch clusters that the current Opensearch cluster is connected with. hence the distinction.

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
```diff
@@ -249,32 +281,45 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
 
   /* Table columns */
   const columns: Array<EuiBasicTableColumn<DataSourceTableItem>> = [
+    {
+      align: 'left',
```
kavilla, line 285: we should add these to the scss file and ref the class name

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
```diff
@@ -249,32 +281,45 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
...
-          {index.id === defaultDataSourceId ? (
-            <EuiBadge iconType="starFilled" iconSide="left">
-              Default
-            </EuiBadge>
-          ) : null}
-        </>
-      ),
+      render: (name: string, item: DataSourceTableItem) => {
+        const indentStyle =
+          item.type === DataSourceEngineType.OpenSearchCrossCluster ? { marginLeft: '20px' } : {};
```
kavilla, line 304: same as above

src/plugins/workspace/public/utils.ts
```diff
@@ -539,14 +596,32 @@ export const fetchDataSourceConnections = async (
...
+      const remoteClusterConnectionsPromises = dataSources.map((ds) => {
+        if (
+          ds.dataSourceEngineType === DataSourceEngineType.OpenSearch ||
+          ds.dataSourceEngineType === DataSourceEngineType.Elasticsearch
+        ) {
+          return getRemoteClusterConnections(ds.id, http!).catch(() => []);
+        }
+        return [];
+      });
+      const remoteClusterConnectionsResult = await Promise.all(remoteClusterConnectionsPromises);
```
kavilla, line 616: this might have the same effect if the underlying call doesn't timeout out these promises. on line 612, we are swallowing all errors so if it fails it will fail silently. and no error will be thrown.

then with promise all it will on be fulfilled when all of the input's promises fulfill. if we have 10 data sources it a single one of those data sources calls fails we will be left with dangling promises. for node 20 the garbage collection is extremely aggressive. so it will see these dangling promises left open and constantly take the time to prioritize gc and other parts of the application will get slowed down.

we should really avoid this pattern

src/plugins/data_source_management/server/routes/remote_cluster_router.ts
```diff
@@ -0,0 +1,75 @@
...
+        });
+
+        const remoteClusters = Object.entries(result.body).map(([key, value]) => ({
+          connectionAlias: key,
+          ...value,
+        }));
+        return response.ok({ body: remoteClusters });
+      } catch (error) {
+        return response.custom({
+          statusCode: 503,
```
kavilla, line 38: i think this might get flagged as a security concern. if i try to hit this api without the proper authentication will i get a `503` instead of something like a `401`

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
```diff
@@ -94,32 +96,67 @@ const convertConnectionToOption = ({
...
+      connection?.parentId
+    ),
+    checked:
+      connection.connectionType !== DataSourceConnectionType.DirectQueryConnection &&
+      selectedConnectionIds.includes(connection.id) &&
+      !connection?.parentId
+        ? ('on' as const)
+        : undefined,
+    prepend: connection?.parentId ? (
+      <div style={{ marginLeft: '20px' }}>
```
kavilla, line 151: same as above with avoiding in line styling

src/plugins/data_source_management/server/routes/remote_cluster_router.ts
```diff
@@ -0,0 +1,75 @@
...
+      validate: {
+        query: schema.object({
+          dataSourceId: schema.string(),
+        }),
+      },
+    },
+    async (context, request, response) => {
+      const client = request.query.dataSourceId
+        ? await context.dataSource.opensearch.getClient(request.query.dataSourceId)
+        : context.core.opensearch.client.asCurrentUser;
```
kavilla, line 23: i dont think this will be possible. if we are validating that the query always has a dataSourceId defined then we wouldn't be able to use the local cluster.

I think the schema should include a `maybe`

src/plugins/data_source_management/server/routes/remote_cluster_router.ts
```diff
@@ -0,0 +1,75 @@
...
+      }
+    }
+  );
+
+  router.get(
+    {
+      path: `${REMOTE_CLUSTER}/indexes`,
+      validate: {
+        query: schema.object({
+          dataSourceId: schema.string(),
```
kavilla, line 50: same as above we are expecting that the dataSourceId is always passed here

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -120,7 +157,8 @@ export const fetchDataSourceConnections = async (
   http: HttpSetup | undefined,
   notifications: NotificationsStart | undefined,
   directQueryTable: boolean,
-  hideLocalCluster: boolean = false
+  hideLocalCluster: boolean,
```
kavilla, line 160: i think the node should be complaining at you in your logs, i had to add this default value because at some places some people didn't pass this param.

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -84,9 +106,10 @@ export const getLocalClusterConnections = async (http: HttpSetup) => {
 export const mergeDataSourcesWithConnections = (
   dataSources: DataSourceTableItem[],
   directQueryConnections: DataSourceTableItem[],
+  remoteClusterConnections: DataSourceTableItem[],
   localClusterConnections?: DataSourceTableItem[]
 ): DataSourceTableItem[] => {
-  const dataSourcesList: DataSourceTableItem[] = [];
+  let dataSourcesList: DataSourceTableItem[] = [];
```
kavilla, line 112: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/DEVELOPER_GUIDE.md#avoid-mutability-and-state

src/plugins/data_source/common/data_sources/types.ts
kavilla, line 63:
```suggestion
  OpenSearchCrossCluster = 'OpenSearch(Cross-cluster search)',
```

kavilla: tbh it always hurt me why we didn't go the path of saying `DEFAULT = 'DEFAULT'`

because now we are in this world where we we have to type out the word `OpenSearch` anytime when we add a new flavor of OpenSearch. to highlight what @Hailong-am we now in this situation, it would have been easier to understand. if it was like DEFAULT, SERVERLESS, CROSS_CLUSTER. 

 plus spaces in ENUMs that are really difficult in practice. if this is stored as a saved object and i want to write a body via some swaggerlike tool. typing out long strings with white spaces.

src/plugins/data_source_management/framework/utils/shared.ts
```diff
@@ -70,6 +70,7 @@ export const EDIT = '/edit';
 export const DATACONNECTIONS_UPDATE_STATUS = '/status';
 export const SECURITY_ROLES = '/api/v1/configuration/roles';
 export const EVENT_ANALYTICS = '/event_analytics';
+export const REMOTE_CLUSTER = '/remote_cluster';
```
kavilla, line 73: unrelated to this pr, but this looks like these arejust constants which should be top level to avoid redfining this everytime

src/plugins/data_source_management/server/routes/remote_cluster_router.ts
LDrago27: @kavilla Do we want to expose the exact error code to user . I was worried we might be exposing too much information hence was thinking of using a rather generic 503 error message.

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
LDrago27: Fixed it

src/plugins/workspace/public/utils.ts
LDrago27: I have updated the code to address your concerns regarding unfulfilled promises.

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
LDrago27: Fixed it

src/plugins/data_source_management/server/routes/remote_cluster_router.ts
LDrago27: Fixed it

src/plugins/data_source_management/server/routes/remote_cluster_router.ts
LDrago27: Fixed it

src/plugins/data_source_management/public/components/utils.ts
LDrago27: Updated the code to follow the guidelines.

src/plugins/data_source/common/data_sources/types.ts
LDrago27: Thanks. Fixed the typo

src/plugins/data_source_management/public/components/data_source_table/data_source_table.scss
```diff
@@ -0,0 +1,15 @@
+.data-source-expanded-table thead {
+  display: none;
+}
+
+.data-source-table [id^="row"][id$="expansion"] > td:first-child > div:first-child {
```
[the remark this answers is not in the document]
LDrago27: Totally agree with the sentiment. Have added comments to delineate the css.

src/plugins/workspace/public/utils.ts
```diff
@@ -557,6 +627,33 @@ export const fetchDataSourceConnections = async (
...
+
+  const remoteClusterConnections: DataSourceConnection[] = [];
+
+  for (const ds of dataSources) {
+    if (
+      ds.dataSourceEngineType === DataSourceEngineType.OpenSearch ||
+      ds.dataSourceEngineType === DataSourceEngineType.Elasticsearch
+    ) {
+      try {
+        const remoteClusterConnectionsResult = await getRemoteClusterConnections(ds.id, http);
```
angle943, line 646: can we do this in parallel?

src/plugins/workspace/public/utils.ts
```diff
@@ -557,6 +627,33 @@ export const fetchDataSourceConnections = async (
...
+        remoteClusterConnections.push(...remoteClusterConnectionsResult);
+      } catch {
+        // Ignore errors when fetching remote cluster connections
+      }
+    }
+  }
+
+  return remoteClusterConnections;
+};
+
```
angle943, line 656: can we get unit tests for these utilities?

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
```diff
@@ -94,32 +97,67 @@ const convertConnectionToOption = ({
...
+    prepend: connection?.parentId ? (
+      <div className="aligned-child-logo">
+        <ConnectionIcon connection={connection} logos={logos} />
+      </div>
+    ) : (
+      <ConnectionIcon connection={connection} logos={logos} />
+    ),
+    parentId: connection.parentId,
+  };
+};
```
angle943, line 160: i think a lot of the complicated logic here can be extracted into util functions which will also make it easier to unit test in isolation

src/plugins/workspace/public/utils.ts
LDrago27: Yup. Planning on raising a follow up PRs to add unit test specific to the new changes that we have introduced.

src/plugins/workspace/public/components/data_source_association/association_data_source_modal.tsx
LDrago27: Fixed it

src/plugins/workspace/public/utils.ts
LDrago27: Made the calls in parallel.

angle943 (dismissed):

angle943 (approved):

ananzh (approved):

PR #9569, opened 2025-03-18
Update visualization when there is change

abbyhu2000:
### Description
Update visualization when there is change

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/view_components/canvas/discover_visualization.tsx
+++ b/src/plugins/discover/public/application/view_components/canvas/discover_visualization.tsx
@@ -41,7 +41,6 @@ export const DiscoverVisualization = ({ hits, bucketInterval, chartData, rows }:
         return;
       }
       const exp = await toExpression(services, searchContext, rows, indexPattern);
-      console.log('expression for vis in discover', exp);
       setExpression(exp);
     }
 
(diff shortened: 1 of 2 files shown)


src/plugins/discover/public/application/view_components/canvas/discover_visualization.tsx
```diff
@@ -65,7 +64,11 @@ export const DiscoverVisualization = ({ hits, bucketInterval, chartData, rows }:
   return (
     <EuiPanel className="discoverVisualization" data-test-subj="visualizationLoader">
       {expression ? (
-        <ReactExpressionRenderer expression={expression} searchContext={searchContext} />
+        <ReactExpressionRenderer
+          key={JSON.stringify(searchContext) + expression}
```
d-buckner, line 68: I don't have major concerns with this in the wip branch, but I do feel we should avoid "forcing" updates with non-functional prop updates. We can look at why the ReactExpressionRenderer isn't re-rendering properly with the change in `searchContext`, but my hunch is that it's the same object reference which wouldn't get picked up if we have shallow memoization. Happy to take a deeper look later on when we prepare these changes for main.

d-buckner (approved):

PR #9571, opened 2025-03-19
Add showVisualization setting in language config

abbyhu2000:
### Description

Add `showVisualization` setting in the language config, so currently discover will only render visualization for promQl based queries. 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/view_components/canvas/discover_visualization.tsx
+++ b/src/plugins/discover/public/application/view_components/canvas/discover_visualization.tsx
@@ -34,6 +34,10 @@ export const DiscoverVisualization = ({ hits, bucketInterval, chartData, rows }:
     filters: filterManager.getFilters(),
     timeRange: timefilter.timefilter.getTime(),
   });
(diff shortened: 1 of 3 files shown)


src/plugins/discover/public/application/view_components/canvas/discover_visualization.tsx
```diff
@@ -47,36 +51,33 @@ export const DiscoverVisualization = ({ hits, bucketInterval, chartData, rows }:
...
-          className="discoverVisualization__empty"
-          data-test-subj="emptyDiscoverVisualization"
-        >
-          <>No data to display</>
-        </EuiFlexItem>
-      )}
+      <ReactExpressionRenderer
+        key={JSON.stringify(searchContext) + expression}
+        expression={expression}
+        searchContext={searchContext}
```
d-buckner, line 77: Just mirroring what I mentioned previously. It looks like we may want to create a new `searchContext` object when it gets updated so that React can do non-forced re-renders based on the shallow comparision.

PR #9574, opened 2025-03-19
create and integrate the clientside resource client factory

d-buckner:
### Description
This is to create a clientside API to surface data connection resources. With this PR I've implemented the prometheus resource API which will be used for dynamic query suggestions. 

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/resources/base_resource_client.ts
+++ b/src/plugins/data/public/resources/base_resource_client.ts
@@ -1,0 +1,29 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { HttpSetup } from 'opensearch-dashboards/public';
+
+const BASE_API = 'api/enhancements';
+
+export class BaseResourceClient {
+  private http: HttpSetup;
+  private dataConnectionType: string;
+
+  constructor(http: HttpSetup, dataConnectionType: string) {
+    this.http = http;
+    this.dataConnectionType = dataConnectionType;
+  }
+
+  protected async get<T>(
+    dataConnectionId: string,
+    resourceType: string,
+    resourceName?: string
+  ): Promise<T> {
+    const resourceNameSuffix = resourceName ? `/${resourceName}` : '';
+    const path = `/${BASE_API}/${this.dataConnectionType}/${dataConnectionId}/resources/${resourceType}${resourceNameSuffix}`;
+    const response = await this.http.get(path);
+    return response.data;
+  }
+}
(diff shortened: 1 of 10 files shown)


src/plugins/data/public/resources/resource_client_factory.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { HttpSetup } from 'opensearch-dashboards/public';
+import { PrometheusResourceClient } from './prometheus_resource_client';
+
+type DataConnectionType = 'prometheus';
+
+export class ResourceClientFactory {
```
kavilla, line 11: ik just feature branch but would this be essentially be accomplished with a route on the server side

src/plugins/data/public/resources/base_resource_client.ts
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { HttpSetup } from 'opensearch-dashboards/public';
+
+const BASE_API = 'api/enhancements';
```
kavilla, line 8: prolly for feature branch purposes but data plugin doesn't depend on the enhancements plugin so might have to invert it

src/plugins/data/public/resources/resource_client_factory.ts
d-buckner: Yeah, this hits the `data_enhancements` resource API route in OSD server which proxies the request to the sql plugin in OS.

src/plugins/discover/public/application/components/visualizations/vislib/line/to_expression.ts
```diff
@@ -80,14 +83,11 @@ const createVegaSpec = (rows: OpenSearchSearchHit[], indexPattern: IndexPattern)
...
   for (let i = 1; i < columns.length; i++) {
-    const colorIndex = (i - 1) % colorPalette.length;
-    const color = colorPalette[colorIndex];
-
     const yAxisParam = {
       field: columns[i].name,
       type: 'quantitative',
       scale: { zero: false },
-      axis: { titleColor: color, title: columns[i].name },
+      axis: { title: '' },
```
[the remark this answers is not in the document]
d-buckner: This file is only used by the demo

src/plugins/data/public/resources/base_resource_client.ts
d-buckner: Yeah, I felt bad about this tbh. I'd like to move the query enhancement languages (`sql`, `ppl`, `promql`) autocompletion definitions to the query enhancement to avoid having this coupling in the data plugin. Thoughts?

Issue #9401, opened 2025-02-17
[BUG] Discover Page contents table doesn't resize always once filtered.


varghese1508:
**Describe the bug**

In the Discover Page, sometimes if we filter the contents such that only one item is present, if we then add a field, when we then remove/disable the filter, the table doesn't resize to accommodate the original contents, i.e., it we can only see one row at a time or so.

**To Reproduce**
1. Go to Discover Tab
2. Ensure there is quite a bit of data available
3. Add a filter such that everything but one entry is filtered out (Eg. filter by id)
4. Add a field from the list of available fields
5. Temporarily disable the filter

These steps may not work consistently. Sometimes, repetition of this activity will cause this bug. Or even narrowing down the search window.

**Expected behavior**
The table of entries should resize to correctly display the 100 rows per page. Instead, it shows one row and a scroll bar to scroll through the 100 items.

**OpenSearch Version**
v2.11

**Dashboards Version**
v2.11

**Plugins**

This issue was re-produced with basic plugins as well.

**Screenshots**

<img width="1505" alt="Image" src="https://github.com/user-attachments/assets/93b45477-56f0-416c-b0ad-9bc5fac5a33c" />

**Additional context**

Hard resetting this table, or going back to the main page and returning generally resets this.

tungsheng:
@kavilla @ashwin-pc We also observed this issue in 2.19.0 playground. Want to check to see the current statue and timeline on resolving this issue. Thank you for your support!! cc @sulabhjain-dev

![Image](https://github.com/user-attachments/assets/47f0ab6b-0f7e-4053-8d97-72c6a0bd36ba)

neodescis:
The data grid is causing this. It is not correctly resetting its height when the row count changes. One fix is to force the component to remount entirely, but it would be better to fix the grid itself.

tungsheng:
Related issue in oui: [Auto fit row heights based on cell context in OuiDataGrid](https://github.com/opensearch-project/oui/issues/1218).

neodescis:
Yes, a re-render is not sufficient. Triggering a remount of the grid using key={rowCount} on the grid does work though.

ashwin-pc:
In the new Discover experience, we have reverted back to using the old table for now, given that this new table component is not very stable. We will look into implementing the table later, since we are prioritizing, improving the querying experience in the new Discover better first. This table component is currently in KTLO. We just don't remove it since there are users who do like it, even though it has its flaws.

[...]

PR #9582, opened 2025-03-20
fix issue in discover grid table on resize when filtering out

ruchidh:
### Description

In the Discover Page, sometimes if we filter the contents such that only one item is present, if we then add a field, when we then remove/disable the filter, the table doesn't resize to accommodate the original contents, i.e., it we can only see one row at a time or so.

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9401

## Screenshot

https://drive.google.com/file/d/1zw3iitCa9jL0kmx_UOHk5WZvS-xjkRjC/view?usp=sharing

## Changelog
- fix: Discover table not updating with row size when filtering out 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/components/data_grid/data_grid_table.tsx
+++ b/src/plugins/discover/public/application/components/data_grid/data_grid_table.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import React, { useState, useMemo, useCallback } from 'react';
+import React, { useState, useMemo, useCallback, useEffect } from 'react';
 import { EuiDataGrid, EuiDataGridSorting, EuiPanel } from '@elastic/eui';
 import { IndexPattern } from '../../../opensearch_dashboards_services';
 import { fetchTableDataCell } from './data_grid_table_cell_value';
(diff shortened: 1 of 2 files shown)


angle943 (dismissed):

src/plugins/discover/public/application/components/data_grid/data_grid_table.tsx
```diff
@@ -131,9 +131,12 @@ export const DataGridTable = ({
     ];
   }, []);
 
-  const table = useMemo(
-    () => (
+  const tableKey = useMemo(() => `${columns.join(',')}-${rows.length}`, [columns, rows.length]);
+
+  const table = useMemo(() => {
```
Maosaic, line 136: I don't see a reason to memorize the table

ruchidh: useMemo is necessary here because rowHeightsOptions were created inline without useMemo, it would be a new object on every render, causing the EuiDataGrid to re-render even when the values remain the same. also

PR #9587, opened 2025-03-21
[Prometheus] Add alertmanager and alert group resources

mengweieric:
### Description

Add alertmanager and alert group resources to query enhancement endpoints for retrieving alert group resources from prometheus.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/server/connections/clients/base_connection_client.ts
+++ b/src/plugins/query_enhancements/server/connections/clients/base_connection_client.ts
@@ -12,11 +12,7 @@ export interface ResourcesQuery {
   query: Record<string, string>;
 }
 
-export interface ClientRequest {
-  query: ResourcesQuery;
-  path: string;
-  params: Partial<TransportRequestParams>;
-}
+export type ClientRequest = Omit<TransportRequestParams, 'method'>;
 
 export interface GetResourcesResponse<R> {
   nextToken?: string;
--- a/src/plugins/query_enhancements/server/connections/managers/prometheus_manager.ts
+++ b/src/plugins/query_enhancements/server/connections/managers/prometheus_manager.ts
@@ -14,12 +14,15 @@ import { URI } from '../../../common/constants';
 import { PromQLConnectionClient } from '../clients/promql_connection_client';
 
 const BASE_RESOURCE_API = 'api/v1';
+const BASE_ALERT_MANAGER_API = 'alertmanager/api/v2';
 
 const PROMETHEUS_RESOURCE_TYPES = {
   LABELS: 'labels',
   LABEL_VALUES: 'label_values',
   METRIC_METADATA: 'metric_metadata',
   ALERTS: 'alerts',
+  ALERTS_GROUPS: 'alert_manager_alert_groups',
+  RULES: 'rules',
 } as const;
 
 // docs: https://prometheus.io/docs/concepts/metric_types/#metric-types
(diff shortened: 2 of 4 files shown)


src/plugins/query_enhancements/server/connections/managers/prometheus_manager.ts
```diff
@@ -14,12 +14,14 @@ import { URI } from '../../../common/constants';
 import { PromQLConnectionClient } from '../clients/promql_connection_client';
 
 const BASE_RESOURCE_API = 'api/v1';
+const BASE_ALERT_MANAGER_API = 'alertmanager/api/v2';
```
d-buckner, line 17: Looks like this route depends on the alert manager plugin. The intention here was to be a proxy to the Prometheus resource APIs. This doesn't look to align with that and seems better served in alert manager given these aren't concepts of Prometheus itself, but rather the alert manager plugin.

mengweieric: This doesn’t create a dependency on Alertmanager but aligns with our design where Alertmanager data is part of the resources. This API serves as the central place for all Prometheus resource data. Keeping it separate in the alerting plugin would lead to duplication and unnecessary dependencies, requiring other modules to rely on the alerting plugin if they also need any data from Alertmanager outside of alerting plugin.

d-buckner: Gotcha makes sense that this isn't coming from the actual alerts plugin. But why are we hitting a different base api than our existing ${BASE_RESOURCE_API}/alerts then?
[https://github.com/opensearch-project/OpenSearch-Dashboards/blob/feature/prometheu[…]_enhancements/server/connections/managers/prometheus_manager.ts](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/feature/prometheus/src/plugins/query_enhancements/server/connections/managers/prometheus_manager.ts#L78)
I guess why conceptually do we want an different base api only for /alerts/groups than all the other resources?

d-buckner: Followup: This is actually based on prometheus's api structure that we directly proxy through to the prometheus server in the sql plugin. Makes sense then.

d-buckner (approved):

joshuali925 (approved):

src/plugins/query_enhancements/server/connections/clients/base_connection_client.ts
```diff
@@ -12,11 +12,7 @@ export interface ResourcesQuery {
   query: Record<string, string>;
 }
 
-export interface ClientRequest {
-  query: ResourcesQuery;
-  path: string;
-  params: Partial<TransportRequestParams>;
-}
+export type ClientRequest = Omit<TransportRequestParams, 'method'>;
```
d-buckner, line 15: It looks like this will no longer vend our ResourceQuery types. Can you do a union of these types so we can still have strong typing for consumers?

ex.
```
export type ClientRequest = Omit<TransportRequestParams, 'method'> & {
  query: ResourceQuery;
}
```

src/plugins/query_enhancements/server/connections/managers/prometheus_manager.ts
```diff
@@ -94,11 +115,8 @@ class PrometheusManager extends BaseConnectionManager<OpenSearchClient> {
     query: PrometheusResourceQuery
   ): Promise<GetResourcesResponse<R>> {
     return this.getClient(context, request).getResources<R>({
-      query,
       path: `${URI.DIRECT_QUERY.RESOURCES}/${query.dataSourceName}/${this.getResourceURI(query)}`,
-      params: {
-        querystring: new URLSearchParams(query.query).toString(),
-      },
+      querystring: new URLSearchParams(query.query).toString(),
```
d-buckner, line 119: Can you verify this doesn't break the existing usage in Discover (for fetching labels and metrics)?

src/plugins/query_enhancements/server/connections/clients/base_connection_client.ts
joshuali925: i think the issue is `ClientRequest.params` is supposed to be passed to `transport.request()` call, but i missed the `.params` in design

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/6fa0a60eea464322c5118940ed529cc5c06ae388/src/plugins/query_enhancements/server/connections/clients/sql_connection_client.ts#L28
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/6fa0a60eea464322c5118940ed529cc5c06ae388/src/plugins/query_enhancements/server/connections/clients/promql_connection_client.ts#L28

if these two are `...request.params` instead of `...request` then we shouldn't need to change prometheus_manager nor the type definition for `ClientRequest`

d-buckner (approved):

src/plugins/query_enhancements/server/connections/managers/prometheus_manager.ts
mengweieric: tested with /resources/metric_metadata and /resources/labels that all worked

PR #9594, opened 2025-03-24
[workspace]refactor workspace_ui_setting_wrapper

Qxisylolo:
### Description
This pr refactors  workspace_ui_setting_wrapper, now it will just return workspace ui setting config if in a workspace. And it will merge with global, user settings in ui_setting_client.

### Issue

## Screenshot

 get global config
截屏2025-03-24 19 47 59

get workspace config
截屏2025-03-24 19 48 05

## Changelog
- feat: add workspace scope and refactor workspace_ui_setting_wrapper

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -108,16 +87,18 @@ export class WorkspaceUiSettingsClientWrapper {
...
+
         const configObject = await wrapperOptions.client.get<Record<string, any>>(
           'config',
           id,
           options
         );
+
         const savedObjectsClient = this.getWorkspaceTypeEnabledClient(wrapperOptions.request);
 
         const workspaceObject = await savedObjectsClient.get<WorkspaceAttribute>(
```
wanglam, line 104: Will the `attributes` of `workspaceObject` includes non workspace scope attributes?

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -55,45 +51,28 @@ export class WorkspaceUiSettingsClientWrapper {
...
-          // Workspace level default data source value should not extend global UIsettings value.
-          [DEFAULT_DATA_SOURCE_UI_SETTINGS_ID]: workspaceLevelDefaultDS,
-          // Workspace level default index pattern value should not extend global UIsettings value.
-          [DEFAULT_INDEX_PATTERN_UI_SETTINGS_ID]: workspaceLevelDefaultIndex,
-        };
-
-        return configObject as SavedObject<T>;
+      // since in UI_settings, we will get from all scope and merge
+      // if the requestWorkspaceId is undefined, we need to manually remove the placeholder to get global config
+      if (id.startsWith(CURRENT_WORKSPACE_PLACEHOLDER)) {
```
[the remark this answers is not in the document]
Qxisylolo: Yes, it may includes non workspace scope attributes

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
Qxisylolo: Yes, it may includes non workspace scope attributes

wanglam:
Hi @Qxisylolo ,Thanks for refactoring the UI setting feature to add workspace scope UI settings feature. I have the following questions about the current implementation:

1. It seems the workspace UI setting will be stored in the `uiSettings` of workspace meta saved object. I suppose only workspace scope UI settings will be stored in this field, right? Do we need to validate the UI setting keys before getting or updating the `uiSettings` in workspace? Do we need to protect the `uiSettings` field to avoid adding illegal settings when updating workspace meta object?
2. For workspace UI settings operations:
   1. How can we update the workspace UI settings? By using saved objects client or uiSettings client? If we want to use saved object client, what is the interface for this operation? 
   2. Can we update a workspace scope UI settings outside a specific workspace by using uiSettings client? If we can do this, how can we pass a specific workspace ID? If we can't do this, do we need any code change to protect this?
   3. Can we update another workspace's UI settings if already in a specific workspace by using saved objects client?
   4. Can we update specific workspace UI settings when outside a workspace by using saved objects client?

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
wanglam: Based on my understanding, it will return non-workspace scope UI settings plus global UI settings if we call `getUserProvided` with workspace scope. Is this by design?

Qxisylolo: if we get ui setting within a workspace, it will return global config attribute + user config attribute+ workspace config attribute

src/plugins/data/server/ui_settings.ts
```diff
@@ -208,6 +208,18 @@ export function getUiSettings(
       schema: schema.nullable(schema.string()),
       scope: workspaceEnabled ? UiSettingScope.WORKSPACE : UiSettingScope.GLOBAL,
     },
+    defaultDataSource: {
```
SuZhou-Joe, line 211: A little bit weird where is this setting before workspace?

src/core/server/ui_settings/ui_settings_client.ts
```diff
@@ -273,28 +303,45 @@ export class UiSettingsClient implements IUiSettingsClient {
...
    */
-  private groupChanges(changes: Record<string, any>) {
-    const userLevelKeys = [] as string[];
-    Object.entries(this.defaults).forEach(([key, value]) => {
+
+  private groupSettingsKeys(defaults: NonNullable<UiSettingsServiceOptions['defaults']>) {
+    Object.entries(defaults).forEach(([key, value]) => {
       if (
         value.scope === UiSettingScope.USER ||
         (Array.isArray(value.scope) && value.scope.includes(UiSettingScope.USER))
```
SuZhou-Joe, line 311: Nit: extract a util to check scope.

src/core/server/ui_settings/ui_settings_client.ts
```diff
@@ -168,15 +196,17 @@ export class UiSettingsClient implements IUiSettingsClient {
...
     } else {
       // group changes into different scope
-      const [global, personal] = this.groupChanges(changes);
+      const [global, personal, workspace] = this.groupChanges(changes);
       if (global && Object.keys(global).length > 0) {
         await this.write({ changes: global });
       }
+      if (workspace && Object.keys(workspace).length > 0) {
+        await this.write({ changes: workspace, scope: UiSettingScope.WORKSPACE });
+      }
```
SuZhou-Joe, line 209: User may not be able to update workspace setting due to permission, shall we put workspace setting as the first write so that it won't partial success.

src/core/server/ui_settings/ui_settings_client.ts
```diff
@@ -152,10 +162,28 @@ export class UiSettingsClient implements IUiSettingsClient {
       // default will get from all scope and merge
       // loop UiSettingScopeReadOptions
       for (const readOptions of UiSettingScopeReadOptions) {
-        userProvided = { ...userProvided, ...this.onReadHook<T>(await this.read(readOptions)) };
+        if (readOptions.scope !== UiSettingScope.WORKSPACE) {
```
SuZhou-Joe, line 165: I do not think there will be difference between read user level setting and workspace level setting.

What about
- read config with scope
- manually set the value as undefined if the key can not be found from the response
- trim keys that is not in the scope
- loop the steps above.

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -141,6 +122,12 @@ export class WorkspaceUiSettingsClientWrapper {
 
         return configObject as SavedObjectsUpdateResponse<T>;
       }
+      // for the workspace scope settings
+      // if the requestWorkspaceId is not defined
+      // then we need to skip updating workspace level setting by manually removing the placeholder
+      if (id.startsWith(CURRENT_WORKSPACE_PLACEHOLDER)) {
+        id = id.replace(`${CURRENT_WORKSPACE_PLACEHOLDER}_`, '');
```
SuZhou-Joe, line 129: Though `id` is a copy, it is not a good practice to mutate function arguments.

SuZhou-Joe: Shall we through error here? It does not make sense that there will be a request trying to update workspace level setting but without workspace id context.

src/plugins/data/server/ui_settings.ts
SuZhou-Joe: Default data source is a global config and workspace level config.

PR #9601, opened 2025-03-25
fix: Query assistant doesn't refresh generated ppl

Qxisylolo:
### Description
This pr fix the issue that query assistant doesn't refresh generated ppl

## Screenshot
https://github.com/user-attachments/assets/454d88d2-067f-4b73-b3bc-822c9f470ef4

## Changelog
- fix: Query assistant doesn't refresh generated ppl

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


raintygao (approved):

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -177,10 +177,12 @@ export class QueryStringManager {
...
    * @param {Query} query
    */
-  public setQuery = (query: Partial<Query>) => {
+  public setQuery = (query: Partial<Query>, force: boolean = false) => {
     const curQuery = this.query$.getValue();
     let newQuery = { ...curQuery, ...query };
-    if (!isEqual(curQuery, newQuery)) {
+    // If the current query is different from the new query, or the user explicitly set force to true,
+    // then proceed with updating the query.
+    if (!isEqual(curQuery, newQuery) || force) {
```
raintygao, line 185: Nit: Im thinking ideally we should update the modified condition criteria(aka the compare logic in isEqual) instead of adding a variable for this directly to bypass.

wanglam (approved):

wanglam:
I did some investigation of this bug. The query state inside `SideBar` still has not been updated due to stringify comparison in [use_query_string_manager](https://github.com/Qxisylolo/OpenSearch-Dashboards/blob/fix/query_assistant_doesn't_refresh_generated_ppl/src/plugins/data/public/ui/search_bar/lib/use_query_string_manager.ts#L53). The reason this bug can be fixed was because [data fetch was subscribed](https://github.com/Qxisylolo/OpenSearch-Dashboards/blob/fix/query_assistant_doesn't_refresh_generated_ppl/src/plugins/discover/public/application/view_components/utils/use_search.ts#L417) to the `data.query.queryString.getUpdate$()`. The `data$` will emit new values during the fetch function execution. Then the `SideBar` component will re-render. It will compare the current query string with the previous query string. Then it will update the text value on the text input.

ruanyl (approved):

SuZhou-Joe:
The `setQuery` itself is coupled with too many conditions check and even a UI toast, that definitely not a data service should do I think. `setQuery` itself should simply update the query$ observable and there may be another util function which will do the conditions check before update the observable.

PR #9602, opened 2025-03-25
Feature/data importer

### LINKED ISSUE #9199: [RFC] Integrate a Data Importer Plugin into Dashboards
# Overview
Currently in OpenSearch Dashboards (OSD), users can ingest documents into OpenSearch and visualize their data in OpenSearch Dashboards. However, there is no existing mechanism to easily import custom static data through the Dashboards UI. Sample data exists but there is no way for custom data to be imported. This is similar to #1791, which articulates the issue as well. In short, there are several use cases for enabling data import through Dashboards:
1. For users:
	1. They can quickly try out Dashboards with their data
	2. They do not need to write one-off scripts to ingest documents into OpenSearch, especially if they don't have the OpenSearch endpoint on hand or they cannot install NPM scripts, like [OpenSearch-CSV](https://github.com/ashwin-pc/Opensearch-CSV)
2. For developers:
	1. Developers can quickly test OSD features with datasets that are more relevant to their use case
	2. Developers can write custom file parsers (elaborated later)

# Requirements
This list is by no means exhaustive but there are several basic features that must be a part of this feature.

As a user:
- I should be able to type my data in the desired format and ingest it into dashboards
- I should be able to upload a file of choice in the desired format and ingest it into dashboards
- I should be able to select which index to ingest it into
- I should be able to select which datasource to ingest it into

As a developer/admin:
- I should have the ability to register my own custom file parsing logic
- I should be able to configure exactly what file types are exposed to end users
- I should be able to configure the file size limit and text character limit

# Out of scope
- A way to create the index and any mappings/settings necessary before import. This is out of scope because the user flow hasn't been finalized. The main purpose of this RFC is to establish a method of ingesting data via OSD directly. However this plugin should be extensible enough that this requirement can be accommodated 
- A custom setup page/data analysis page. Like the bullet point above, it's better suited as a separate issue but should be considered
- Bulk ingesting multiple datasets/ingesting large datasets. This feature focuses on importing smaller static datasets. With larger datasets (large in this context means datasets in the GB), users are more suited to ingest directly into OpenSearch as it was designed for that specific purpose.

# Approach
To integrate support, we must split our approach into UI and server components

## UI
OUI component library should contain the necessary components for the user to execute the actions specified in the requirements:
- `OuiFilePicker`: For uploading data
- `OuiCodeEditor`/`monaco`: For inputting text
- `OuiSelect`: For choosing file type and import type (file/text upload)
- `OuiSelectable`: For choosing the index name (Fetching can be done with the `IndexPatternsService`, or as a last resort, expose an API to query for index names)
- `DataSourceSelectable`: This is exposed via the `dataSourceManagement` plugin and should handle the datasource fetching for us
- `OuiButton`: For executing the import process

## Server
In OSD core, we have a similar method to import files via Saved Objects import.

For client side:
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/33f7ba6e7b37cff0141724c04147b2ca46e16d9b/src/plugins/saved_objects_management/public/management_section/objects_table/components/flyout.tsx#L193-L226

For server side:
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/33f7ba6e7b37cff0141724c04147b2ca46e16d9b/src/core/server/saved_objects/routes/import.ts#L48-L81

We can follow a similar approach and expose two routes (tentatively named):
```
/api/data_importer_plugin/_import_text
/api/data_importer_plugin/_import_file
```

As the names suggest, we split up importing text and file inputs into two routes. There are two reasons why they're split:
1. It makes the schemas smaller, easier to write, and easier to validate
2. Some file types can only be upload-able (Excel sheets for instance) and cannot be textual input. There can exist some data formats that can only be textual and cannot be uploaded to a file, but I cannot think of any formats. The point is, handling a file should be separate from handling text, especially as file uploads will most definitely involve larger datasets than typed ones.

For text input, the flow is as follows:
1. Call `validateText()` to ensure data is well formed
2. Call `ingestText()` to ingest that document into OpenSearch

For file input, the flow is as follows:
1. Call `ingestFile()` and use the stream to validate and ingest to OpenSearch in chunks

Because the underlying `FileStream` may be arbitrarily large (indeed, we can add a config to limit the max size in bytes but that limit is set by the user), we cannot store the entire contents in memory. We must process this as a `stream`. This means there will be no pre-validation step and there's a possibility that only some documents can be ingested. The issue of how to handle these failed records is an implementation detail but we can specify which documents succeeded/failed to ingest in the response body.

To accommodate the many types of files, there needs to be a parser dedicated to each type of file called `IFileParser`. The structure of this parser is as follows:
```Typescript
/**
 * Parser that handles a particular file type
 */
export interface IFileParser {
  /**
   * Can this file type support text input? If true, validateText() and ingestText() MUST be supplied
   * @returns
   */
  supportsText: () => boolean;

  /**
   * Given text input, validate that it is in the expected format; required if supportsText() returns true
   * @param text
   * @param options
   * @returns
   * @throws Can throw an error if text doesn't match expected format
   */
  validateText?: (text: string, options: ValidationOptions) => Promise<boolean>;

  /**
   * Assuming valid text input, handle the ingestion into OpenSearch; required if supportsText() returns true
   * @param text
   * @param options
   * @returns
   * @throws Can throw server errors when attempting to ingest into OpenSearch
   */
  ingestText?: (text: string, options: IngestOptions) => Promise<IngestResponse>;

  /**
   * Can this file type support text input? If true, validateFile() MUST be supplied
   * @returns
   */
  supportsFile: () => boolean;

  /**
   * Given an arbitrary file stream, handle the validation and ingestion into OpenSearch; required if supportsFile() returns true
   * @param file
   * @param options
   * @returns
   * @throws Can throw server errors when attempting to ingest into OpenSearch
   */
  ingestFile?: (file: Readable, options: IngestOptions) => Promise<IngestResponse>;
}
```

### Registering Custom file formats
By default, the `DataImporterPlugin` will supply three parsers: `.ndjson, .csv, and .json`. Parsers have to be registered in the `FileParserService`, which will register the `IFileParsers` generated by the `DataImporterPlugin` as well as any other plugins. For the latter, `DataImporterPluginSetup` will expose a function `registerFileParser()` to other plugins to register a custom `IFileParser` (like for example `.xlsx, .geojson, .gltf, .biojson`, etc.):
```Typescript
export interface DataImporterPluginSetup {
  /**
   * Register custom file type parsers to ingest into OpenSearch
   * @param fileType The file type to register a parser for (should NOT be csv, ndjson, or json filetypes)
   * @param fileParser
   * @throws errors if a filetype is already registered in this plugin or another plugin or if the IFileParser does not implement some methods
   */
  registerFileParser: (fileType: string, fileParser: IFileParser) => void;
}

export class DataImporterPlugin
  implements Plugin<DataImporterPluginSetup, DataImporterPluginStart> {
  private readonly fileParsers: FileParserService = new FileParserService();

  constructor(private readonly initializerContext: PluginInitializerContext) {}

  public async setup(
    core: CoreSetup,
    deps: DataImporterPluginSetupDeps
  ): Promise<DataImporterPluginSetup> {
    // Register default file parsers
    this.fileParsers.registerFileParser(CSV_FILE_TYPE, new CSVParser());
    this.fileParsers.registerFileParser(NDJSON_FILE_TYPE, new NDJSONParser());
    this.fileParsers.registerFileParser(JSON_FILE_TYPE, new JSONParser());

    // Handle other setup() logic

    return {
	  // Exposed function for other plugins to consume
      registerFileParser: (fileType, fileParser) => {
        this.fileParsers.registerFileParser(fileType, fileParser);
      },
    };
  }
}
```

# POC
A PoC plugin [data-importer-plugin](https://github.com/huyaboo/data-importer-plugin) is introduced which helps capture this vision. It doesn't implement everything stated in this RFC, but it provides the core feature set outlined in the requirements

huyaboo:
### Description
This PR adds a new `Data Importer Plugin`, a plugin that enables users to ingest their CSV/NDJSON/JSON static data files into OpenSearch directly from OpenSearch Dashboards UI.

Features:
- MDS compatible
- Supports CSV/NDJSON/JSON files
- Text box or file upload

### Issues Resolved
closes #9199

## Screenshot

image

https://github.com/user-attachments/assets/7a08c541-258b-4fe5-a4d8-2072137d1a20

## Testing the changes

- New + existing unit tests pass for the plugin

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


changelogs/fragments/9462.yml
```diff
@@ -0,0 +1,2 @@
+fix:
+- Remove system indexes from the index name selector ([#9462](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9462))
```
zhyuanqi, line 2: Should we update the changelog?

src/plugins/data_importer/server/utils/util.ts
```diff
@@ -0,0 +1,115 @@
...
+  dataSourceEnabled: boolean,
+  context: RequestHandlerContext,
+  dataSourceId?: string
+): Promise<OpenSearchClient> => {
+  return dataSourceEnabled && dataSourceId
+    ? await context.dataSource.opensearch.getClient(dataSourceId)
+    : context.core.opensearch.client.asCurrentUser;
+};
+
+export const validateEnabledFileTypes = (fileTypes: string[], fileParsers: FileParserService) => {
```
xinruiba, line 21: nit: rename to ```validateFileTypes```

src/plugins/data_importer/server/parsers/json_parser.ts
```diff
@@ -0,0 +1,104 @@
...
+        resolve(true);
+      } catch (e) {
+        resolve(false);
+      }
+    });
+
+    const total = isSuccessful ? 1 : 0;
+
+    return {
+      total: 1,
```
xinruiba, line 41: ```total: total```?

xinruiba, line 41:
```suggestion
      total: total,
```

src/plugins/data_importer/server/parsers/json_parser.ts
```diff
@@ -0,0 +1,104 @@
...
+            });
+            resolve(1);
+          } catch (_) {
+            resolve(0);
+          }
+        });
+    });
+
+    return {
+      total: 1,
```
xinruiba, line 73: I believe you want to depend on this total attribute to calculate the number of successful ingested files when ingest multiple files together right?

In that case, I'll suggest to use the resolve number here instead of hardcoding 1.

src/plugins/data_importer/server/parsers/json_parser.ts
```diff
@@ -0,0 +1,104 @@
...
+    };
+  }
+
+  public async parseFile(file: Readable, limit: number, _: ParseOptions) {
+    const documents: Array<Record<string, any>> = [];
+    await new Promise<void>((resolve, reject) => {
+      let rawData = '';
+      file
+        .on('data', (chunk) => (rawData += chunk))
+        .on('error', (e) => reject(e))
```
xinruiba, line 85: Any reason we use different error handling format here instead of ```resolve(0)```?

huyaboo: `parseFile()` is used when previewing a file. If the parsing fails on preview, we should reject invalid formats before proceeding (and surface that to the user).

src/plugins/data_importer/server/parsers/json_parser.ts
huyaboo: Actually `total` is the total number of documents detected, regardless of ingestion status. The `failedRows` records not only the failed rows but also the total count of failed documents (by getting the length of the array)

src/plugins/data_importer/server/utils/util.ts
huyaboo: Renamed

src/plugins/data_importer/server/types.ts
```diff
@@ -0,0 +1,112 @@
...
+   * Used for CSV file types, indicates the delimiter to use when parsing the CSV file
+   */
+  delimiter?: string;
+}
+export type ParseOptions = ValidationOptions;
+
+/**
+ * Parser that handles a particular file type
+ */
+export interface IFileParser {
```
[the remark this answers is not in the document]
huyaboo: Renamed to `IFileProcessor`

changelogs/fragments/9462.yml
huyaboo: Updated

src/plugins/data_importer/server/processors/json_processor.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    return {
+      total: 1,
+      message: `Indexed ${numSucessfulDocs} document`,
+      failedRows: numSucessfulDocs === 1 ? [] : [1],
+    };
+  }
+
+  public async parseFile(file: Readable, limit: number, _: ParseOptions) {
+    const documents: Array<Record<string, any>> = [];
```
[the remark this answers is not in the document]
huyaboo: Yes so `limit` is defined in the `FileProcessor` interface since we only need to preview the first `N` entries. For a file with many entries, we don't want to process the entire file since the UI will get cluttered.

src/plugins/data_importer/server/processors/csv_processor.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    return {
+      total: numDocuments,
+      message: `Indexed ${numDocuments - failedRows.length} documents`,
+      failedRows: failedRows.sort((n1, n2) => n1 - n2),
+    };
+  }
+
+  public async parseFile(file: Readable, limit: number, options: ParseOptions) {
+    const { delimiter } = options;
```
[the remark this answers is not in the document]
huyaboo: Yes, `limit` helps limit how many entries we want to process. We want to parse only the first `N` (10 by default) entries to keep preview quick

zhongnansu (approved):
lgtm, could you add an github issue for adding cypress test to cover this new feature?

xinruiba (approved):

ZilongX (approved):

zhyuanqi (approved):

src/plugins/data_importer/server/utils/determine_mapping.ts
```diff
@@ -0,0 +1,75 @@
...
+    case value == null:
+      return { type: DYNAMIC_MAPPING_TYPES.NULL };
+    case typeof value === 'string' && value.length === 0:
+      return defaultType;
+    case isBooleanType(value):
+      return { type: DYNAMIC_MAPPING_TYPES.BOOLEAN };
+    case isNumericType(value):
+      return { type: determineExactNumberType(Number(value)) };
+    case isValidDate(_.toString(value)):
+      // TODO Dates need further parsing since OpenSearch expects a certain timestamp format
```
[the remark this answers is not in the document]
huyaboo: Not really, Date handling in `OpenSearch` requires a separate parser of its own. OpenSearch expects a specific timestamp format should the `date` type be used and is a common source of ingestion errors.

huyaboo:
> lgtm, could you add an github issue for adding cypress test to cover this new feature?

Yes, @ruchidh can pick up this task

PR #9603, opened 2025-03-25
Fix: Skip appending time range to ppl when not querying with source

Maosaic:
### Description

Update PPL query builder to skip appending time range to ppl when not querying with source.

### Issues Resolved

Resolve a issue that cause PPL query to failed unexpectedly

## Testing the changes

Before

Screenshot 2025-03-25 at 1 34 54 PM

After

Screenshot 2025-03-25 at 1 34 31 PM

## Changelog

-fix: Skip appending time range to ppl when not querying with source

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/common/utils.ts
+++ b/src/plugins/query_enhancements/common/utils.ts
@@ -121,3 +121,21 @@ export const buildQueryStatusConfig = (response: any) => {
     sessionId: response.data.sessionId,
   } as QueryStatusConfig;
 };
+
+/**
+ * Test if a PPL query is using search command
+ * https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/cmd/search.rst
+ */
+export const isPPLSearchQuery = (query: Query) => {
+  if (query.language !== 'PPL') {
+    return false;
+  }
+
+  if (typeof query.query !== 'string') {
+    return false;
+  }
+
+  const string = query.query.toLowerCase().replace(/\s/g, '');
+
+  return string.startsWith('source=') || string.startsWith('searchsource=');
+};
--- a/src/plugins/query_enhancements/public/search/ppl_search_interceptor.ts
+++ b/src/plugins/query_enhancements/public/search/ppl_search_interceptor.ts
@@ -20,6 +20,7 @@ import {
   EnhancedFetchContext,
   fetch,
   formatDate,
+  isPPLSearchQuery,
   QueryAggConfig,
   SEARCH_STRATEGY,
 } from '../../common';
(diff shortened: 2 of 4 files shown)


src/plugins/query_enhancements/public/search/ppl_search_interceptor.ts
```diff
@@ -95,6 +95,11 @@ export class PPLSearchInterceptor extends SearchInterceptor {
     const datasetService = queryString.getDatasetService();
     if (datasetService.getType(dataset.type)?.languageOverrides?.PPL?.hideDatePicker === false)
       return query;
+    if (
```
ruanyl, line 98: would be nice to have a unit test

ruanyl (approved):

Hailong-am (approved):

sejli (commented):
Nit: We might want to include something for `search source = ...` since `search source =` is technically the [default starting command](https://opensearch.org/docs/latest/search-plugins/sql/ppl/syntax/), but can be omitted from the query.

src/plugins/query_enhancements/public/search/ppl_search_interceptor.ts
```diff
@@ -95,6 +95,11 @@ export class PPLSearchInterceptor extends SearchInterceptor {
...
+      typeof query.query === 'string' &&
+      !query.query.toLowerCase().replace(/\s/g, '').startsWith('source=')
```
angle943, line 100: i was thinking why you used `replace(/\s\g, '')` instead of just `trim()`, and then i realized that there can be spaces in between `source` and `=`. Can we extract this to a util function, like `queryStartsWithSource`? There's a lot going on especially in this already convoluted file that makes it difficult to read. (we can also unit test that util function, which supports the other comment by @ruanyl )

src/plugins/query_enhancements/common/utils.ts
```diff
@@ -121,3 +121,21 @@ export const buildQueryStatusConfig = (response: any) => {
     sessionId: response.data.sessionId,
   } as QueryStatusConfig;
 };
+
+/**
+ * Test if a PPL query is using search command
+ * https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/cmd/search.rst
+ */
+export const isSearchQuery = (query: Query) => {
```
angle943, line 129: nit: `isPplSearchQuery`?

angle943 (dismissed):

src/plugins/query_enhancements/common/utils.ts
sejli: +1, think having some indicator to specify this function's usecase (like `isPPLSearchQuery` would be good. Not blocking though, could have a fast follow up/add in future PR

sejli (dismissed):

src/plugins/query_enhancements/common/utils.ts
Maosaic: I will update the function name

> aren't all PPL queries a search?

I think the search here means a specific PPL command, not the generic "search" concept. There are some commands that are mutually exclusive with `search`, for example `describe`.

https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/cmd/describe.rst

angle943 (approved):

Hailong-am (approved):

PR #9615, opened 2025-03-27
Define upper bound for prometheus datapoints sent to the frontend

d-buckner:
### Description
While testing with larger prometheus datasets, we found considerable performance issues. This resulted in hangs for up to 15s due to backpressure on the garbage collector. This highlights that our `to_expression` and `parse_expression` have a lot of room to optimize for memory, but mainly that we were overwhelming the frontend by sheer amount of datapoints. In the example I found over 160,000 datapoints were sent to the frontend, which is not renderable by virtually any html based renderer. This sets an upper bound of 1000 datapoints we send to the frontend (20 series max, 50 samples per series).

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/query_enhancements/server/search/promql_search_strategy.ts
```diff
@@ -16,7 +16,9 @@ import {
 } from '../../../data/common';
 
 // Query 500 samples by default
```
joshuali925, line 18: outdated comment?

d-buckner: Thanks for the catch, I've updated

PR #9620, opened 2025-03-28
feat: make plugin setup/start order stable and consistent

Hailong-am:
### Description

Add a sort before doing topological sort, that will eliminate the [uncertain order](https://utcc.utoronto.ca/%7Ecks/space/blog/unix/ReaddirOrder) of `readdir`. The topological sort algorithm is stable, once we have a ordered input, the sorting result will become stable too.

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9629

## Changelog

- fix: Add sort to make plugins setup/start order stable

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/10/queries.spec.js
+++ b/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/10/queries.spec.js
@@ -140,8 +140,10 @@ const queriesTestSuite = () => {
         cy.osd.waitForLoader(true);
         cy.wait(1000);
 
-        // Logs table is displayed by default
-        cy.getElementByTestId('docTable').should('be.visible');
+        // Statistic tab is selected
+        cy.getElementByTestId('exploreTab-explore_statistics')
+          .should('be.visible')
+          .and('have.attr', 'aria-selected', 'true');
         // Navigate to visualization
         cy.get('#explore_visualization_tab').click();
         cy.getElementByTestId('exploreVisualizationLoader').should('be.visible');
--- a/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/23/auto_tab_detect.spec.js
+++ b/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/23/auto_tab_detect.spec.js
@@ -1,0 +1,307 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+  DATASOURCE_NAME,
+  INDEX_WITH_TIME_1,
+  START_TIME,
+  END_TIME,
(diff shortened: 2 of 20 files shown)


src/core/server/plugins/plugins_system.test.ts
```diff
@@ -408,8 +408,8 @@ test('`uiPlugins` returns ordered Maps of all plugin manifests', async () => {
       "order-0",
       "order-1",
       "order-2",
-      "order-3",
       "order-4",
+      "order-3",
```
SuZhou-Joe, line 412: Is the order change here expected?

Hailong-am: Yes, with this dependency graph, the order with sorted input
`[order0, order1, order2, order3, order4]`
and result will be 
`order0, order1, order2, order4, order3`


image

src/core/server/plugins/plugins_system.ts
```diff
@@ -244,8 +244,14 @@ export class PluginsSystem {
     // topological ordering. If the cloned graph is _not_ empty at the end, we
     // know we were not able to topologically order the graph. We exclude optional
     // dependencies that are not present in the plugins graph.
+
+    // plugins order is determined by the order in which they were added to the system,
+    // that depends on `readdir` to read the plugins directory which returns the files in a non-deterministic order.
```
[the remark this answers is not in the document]
wanglam: As my understanding, the plugins search paths were declared in the `Env` like the code below:
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/2eac413da726cdfe92d4d25a460ff80f57d82780/packages/osd-config/src/env.ts#L133-L138

In the old implementation, the core plugins will try to setup before third party plugins. Although, they still follow the dependency order. I'm not sure if it's safe to sort all kinds of plugins below.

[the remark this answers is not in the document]
Hailong-am: The contract defined in manifest file only make sure they are sort into topologic order, and we don't break it. There may have trick behavior depends on the discovery order that's the pr want to fixed.

src/core/server/plugins/plugins_system.test.ts
SuZhou-Joe: As long as the order remains consistent, it should be fine. I'm just wondering why the order changed in this case, since localeCompare should place order3 before order4.

SuZhou-Joe (approved):

src/core/server/plugins/plugins_system.test.ts
Hailong-am: If the two plugins are independent, the order of their setup will be the reverse of the order in the array. It's kind of similar with stack based on code here

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/436d54d55e9ebfeaa0e2a6382ff8462bd167adf4/src/core/server/plugins/plugins_system.ts#L266-L267

Issue #9264, opened 2025-01-23
[BUG][Query enhancements] Discover state partially resets when switching between plugins


joshuali925:
**Describe the bug**

When query enhancements is enabled, switching from Discover to Dashboard would set query language back to DQL, and switching back to Discover would reset the selected index pattern to the default one. This becomes problematic when there are no index patterns in the current workspace. It would show the previously selected dataset with an empty "SQL" query bar, but it is actually sending DSL requests to `/internal/search/opensearch-with-long-numerals`

**To Reproduce**
1. Create a new workspace with a opensearch connection. Ensure there are no index patterns in this workspace
2. Click on discover and select an index from the opensearch connection with SQL language
3. Run some query like `SELECT request,response FROM apachelog where response = '501' limit 2`
4. Click on Dashboards and click back to Discover
5. See index is still selected, language says "OpenSearch SQL", and query bar is empty
6. Now try to run the same SQL query, discover shows different results
7. In network tab, it calls `POST /w/czLj6e/internal/search/opensearch-with-long-numerals` instead of `/w/czLj6e/api/enhancements/search/sql`

**Expected behavior**
UI appearance and underlying state should be consistent

**OpenSearch Version**
main

**Dashboards Version**
main

**Plugins**

none

**Screenshots**

Notice SQL query is inside DSL
![Image](https://github.com/user-attachments/assets/af41f86c-8c38-4421-bd59-2b5d6297a18b)

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

One root cause is in `setQuery`, 
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/0be70a72634e1d78f7b4f0e70bd21ff618ff1116/src/plugins/data/public/query/query_string/query_string_manager.ts#L168-L171

Since there's no index pattern, when resetting the query to default on switching from Dashboard to Discover, the current logic sees the default query's dataset is empty and will keep the original dataset, but language got reset to DQL.

I've tried small effort fixes like 
```ts
    if (!query.dataset) {
      newQuery.dataset = undefined;
    }
```

and the index is no longer selected, but clicking on "Run" still sends DSL queries to the `apachelog` index. Apparently there are more places causing issues and need to be fixed

andrross:
[...]

PR #9625, opened 2025-03-28
Fix a issue that can cause incorrect query to fire when switch plugin

Maosaic:
### Description

Fix a issue that can cause `Discover` fire incorrect query when switch from other plugin. There are a few problems add up to this unexpected user experience:

1. When user navigate to `Discover` using the left side navigation panel, it will always try to do a "fresh start" by loading a empty dataset with the query and language store in local storage. But because `setQuery` always try to merge the next query state into the current query state, if there is a existing dataset, it will never be replace by the empty dataset.
2. Currently, dataset is store in `queryStringManager` as part of the `query` state, and because `queryStringManager` is global, dataset state persist when user navigate to other plugin after visiting Discover.
3. Whenever the selected language changed, we need to update the query and language store in local storage, this update was missing when:

    - Switching language between "DQL" and "Lucene" using the legacy language switcher
    - Loading a saved search
    - Selecting dataset

To addrees these problems, this PR introduces the changes below:

1. Update `clearQuery` to allow empty dataset replace existing dataset
2. Call `clearQuery` if url does not contain `_q` state("fresh start")
3. Add missing local storage updates

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9264

## Screenshot

Before

https://github.com/user-attachments/assets/ab8a77fa-eb98-48b8-8755-a80e8620a4ab

After

https://github.com/user-attachments/assets/9d973688-52c5-448f-97be-07a6e57590d7

## Changelog

- fix: fix a issue that can cause incorrect query to fire when switch plugin

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/query_string_manager.ts
+++ b/src/plugins/data/public/query/query_string/query_string_manager.ts
@@ -225,7 +225,28 @@ export class QueryStringManager {
    * Resets the query to the default one.
    */
   public clearQuery = () => {
-    this.setQuery(this.getDefaultQuery());
+    const curQuery = this.query$.getValue();
+    let newQuery = this.getDefaultQuery();
+    // Check if dataset changed and if new dataset has language restrictions
+    if (newQuery.dataset && !isEqual(curQuery.dataset, newQuery.dataset)) {
+      // Get supported languages for the dataset
+      const supportedLanguages = this.datasetService
+        .getType(newQuery.dataset.type)
+        ?.supportedLanguages(newQuery.dataset);
+
+      // If we have supported languages and current language isn't supported
+      if (supportedLanguages && !supportedLanguages.includes(newQuery.language)) {
+        // Get initial query with first supported language and new dataset
+        newQuery = this.getInitialQuery({
+          language: supportedLanguages[0],
+          dataset: newQuery.dataset,
+        });
+      }
+
+      // Add to recent datasets
+      this.datasetService.addRecentDataset(newQuery.dataset);
+    }
+    this.query$.next(newQuery);
   };
 
   // Todo: update this function to use the Query object when it is udpated, Query object should include time range and dataset
--- a/src/plugins/data/public/query/query_string/query_string_manager.ts
+++ b/src/plugins/data/public/query/query_string/query_string_manager.ts
@@ -177,9 +177,13 @@ export class QueryStringManager {
    * Updates the query.
    * @param {Query} query
    */
-  public setQuery = (query: Partial<Query>, force: boolean = false) => {
+  public setQuery = (
+    query: Partial<Query>,
+    force: boolean = false,
+    mergeCurrentQuery: boolean = true
+  ) => {
     const curQuery = this.query$.getValue();
-    let newQuery = { ...curQuery, ...query };
+    let newQuery = mergeCurrentQuery ? { ...curQuery, ...query } : (query as Query);
     // If the current query is different from the new query, or the user explicitly set force to true,
     // then proceed with updating the query.
     if (!isEqual(curQuery, newQuery) || force) {
@@ -225,28 +229,9 @@ export class QueryStringManager {
    * Resets the query to the default one.
    */
   public clearQuery = () => {
-    const curQuery = this.query$.getValue();
-    let newQuery = this.getDefaultQuery();
-    // Check if dataset changed and if new dataset has language restrictions
-    if (newQuery.dataset && !isEqual(curQuery.dataset, newQuery.dataset)) {
-      // Get supported languages for the dataset
-      const supportedLanguages = this.datasetService
-        .getType(newQuery.dataset.type)
-        ?.supportedLanguages(newQuery.dataset);
-
-      // If we have supported languages and current language isn't supported
-      if (supportedLanguages && !supportedLanguages.includes(newQuery.language)) {
-        // Get initial query with first supported language and new dataset
-        newQuery = this.getInitialQuery({
-          language: supportedLanguages[0],
-          dataset: newQuery.dataset,
-        });
-      }
-
-      // Add to recent datasets
-      this.datasetService.addRecentDataset(newQuery.dataset);
-    }
-    this.query$.next(newQuery);
+    const force = false;
+    const mergeCurrentQuery = false;
+    this.setQuery(this.getDefaultQuery(), force, mergeCurrentQuery);
   };
 
   // Todo: update this function to use the Query object when it is udpated, Query object should include time range and dataset
--- a/src/plugins/data/public/query/state_sync/connect_to_query_state.ts
+++ b/src/plugins/data/public/query/state_sync/connect_to_query_state.ts
@@ -77,11 +77,13 @@ export const connectStorageToQueryState = (
       filters: filterManager.getAppFilters(),
     };
 
-    // set up initial '_q' flag in the URL to sync query and filter changes
     if (!osdUrlStateStorage.get('_q')) {
+      // set up initial '_q' flag in the URL to sync query and filter changes
       osdUrlStateStorage.set('_q', initialStateFromURL, {
         replace: true,
       });
+      // clear existing query and apply default query
+      queryString.clearQuery();
     }
 
     if (syncConfig.query && !_.isEqual(initialStateFromURL.query, queryString.getQuery())) {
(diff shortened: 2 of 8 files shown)


src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -502,6 +502,9 @@ export const useSearch = (services: DiscoverViewServices) => {
 
       filterManager.setAppFilters(actualFilters);
       data.query.queryString.setQuery(query);
+      // Update local storage after loading saved search
```
kavilla, line 505: imo we should feature gate this. or feature gate the localstorage usage specifically when query enhancements is disabled.

The locals torage for query language from what i gathered kind of acts as a user preference - it updates when users manually submit queries or change languages, assuming they'll want that same language in future sessions.

if we do this where the local storage updates every time a saved search loads, probably doesn't make sense to have a user preference. it makes more sense to either remove the localstorage functionality or only keep it when query enhancements is disabled.

src/plugins/data/public/query/state_sync/connect_to_query_state.ts
```diff
@@ -77,11 +77,13 @@ export const connectStorageToQueryState = (
       filters: filterManager.getAppFilters(),
     };
 
-    // set up initial '_q' flag in the URL to sync query and filter changes
     if (!osdUrlStateStorage.get('_q')) {
+      // set up initial '_q' flag in the URL to sync query and filter changes
```
kavilla, line 81: is line 46 used in here?

src/plugins/data/public/query/state_sync/connect_to_query_state.ts
```diff
@@ -77,11 +77,13 @@ export const connectStorageToQueryState = (
...
       osdUrlStateStorage.set('_q', initialStateFromURL, {
         replace: true,
       });
+      // clear existing query and apply default query
+      queryString.clearQuery();
```
kavilla, line 86: is there any impact if we dont feature gate it?

Maosaic: I think it's relatively save here, anyway we are applying the default query(L75-L76 + L91)

src/plugins/data/public/query/state_sync/connect_to_query_state.ts
Maosaic: Are we suppose to read the `syncConfig` here? I'm not quite sure how it's set and use

src/plugins/discover/public/application/view_components/utils/use_search.ts
Maosaic: I'm seeing we are setting local storage in query editor language picker when query enhancement enabled, imo if we are using it in language picker, we should use it with save search

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -225,7 +225,28 @@ export class QueryStringManager {
...
+        .getType(newQuery.dataset.type)
+        ?.supportedLanguages(newQuery.dataset);
+
+      // If we have supported languages and current language isn't supported
+      if (supportedLanguages && !supportedLanguages.includes(newQuery.language)) {
+        // Get initial query with first supported language and new dataset
+        newQuery = this.getInitialQuery({
+          language: supportedLanguages[0],
+          dataset: newQuery.dataset,
+        });
```
virajsanghvi, line 243: Why not always just return this? Its effectively what getDefaultQuery is doing (or pass the new Query to getDefaultQuery to reuse the same logic).

I guess main difference is the recentDataset?

src/plugins/data/public/query/state_sync/connect_to_query_state.ts
virajsanghvi: Should we be mutating state in this method? I don't think callers would expect arguments to be mutated? Is it better to do this as part of whatever calls this?

src/plugins/data/public/ui/query_string_input/query_string_input.tsx
```diff
@@ -469,7 +469,9 @@ export default class QueryStringInputUI extends Component<Props, State> {
       body: JSON.stringify({ opt_in: language === 'kuery' }),
     });
 
+    // Update local storage
     this.services.storage.set('userQueryLanguage', language);
+    this.services.data.query.queryString.getInitialQueryByLanguage(language);
```
virajsanghvi, line 474: nit: It is odd that we are calling a getter and relying on a side effect - can we just call the setter we expect to be calling here?

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -502,6 +502,9 @@ export const useSearch = (services: DiscoverViewServices) => {
...
+      data.query.queryString.getLanguageService().setUserQueryLanguage(query.language);
+      data.query.queryString.getInitialQueryByLanguage(query.language);
```
virajsanghvi, line 507: same comment as above

src/plugins/data/public/query/state_sync/connect_to_query_state.ts
Maosaic: This only happen if the url state does not contain any query state, technically it does not change any state since the query is reset at L75-L76 anyway, here it just apply the it to `queryStringManager`

src/plugins/data/public/query/query_string/query_string_manager.ts
Maosaic: The main difference here compare to the old `clearQuery`, is to allow empty dataset, and I kept rest of code branch from `setQuery` to avoid regressions.

src/plugins/data/public/ui/query_string_input/query_string_input.tsx
Maosaic: I agree, but calling the setter will again need to call the same getter(`getInitialQueryByLanguage `) first to get the query string, then call the setter which has already been called in getter, this also feels weird to me.

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -225,7 +225,28 @@ export class QueryStringManager {
...
+      }
+
+      // Add to recent datasets
+      this.datasetService.addRecentDataset(newQuery.dataset);
+    }
+    this.query$.next(newQuery);
```
angle943, line 249: looks like this file has an associated `query_string_manager.test.ts`. is it possible to add unit tests for the additions you've added?

src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -225,7 +225,28 @@ export class QueryStringManager {
    * Resets the query to the default one.
    */
   public clearQuery = () => {
-    this.setQuery(this.getDefaultQuery());
+    const curQuery = this.query$.getValue();
```
angle943, line 228: this is doing awful lot for a method called `clearQuery()`. Looking at the code, it looks like we gotrta handle the scenario where dataset has changed. would it be feasible to have this code in like a new method called `handleDatasetChange` or something?

angle943 (approved):

ruchidh (approved):

src/plugins/data/public/ui/query_string_input/query_string_input.tsx
kavilla: correct me if im wrong, this will clear out the query whenever someone switches languages (between DQL/Lucene) - for the non-enhanced version.

In the query string input component (default experience), the query sticks around during language switches since DQL/Lucene have similar syntax.

we should def ensure feature gating this one if possible

Maosaic: If I understand it correctly, `getInitialQueryByLanguage` should only change the query save in local storage without changing the query in search bar.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/query/query_string/query_string_manager.ts#L301-L315

kavilla: would that make it out of sync?

Maosaic: I think the query saved in local storage never meant to be sync with the searching one, the local storage query serve as a fallback mechanism when initializing.

LDrago27 (approved):

ananzh (approved):

PR #9627, opened 2025-03-29
update promql autocomplete with better representation of suggestions

paulstn:
### Description

Mostly small changes:

- updated resource factory to only have a single instance of prometheus resource client, so that calling it won't recreate a new instance
- added the type of a metric (counter, gauge, etc) to its description
- updated all prometheus suggestions to be more accurate in its icons and descriptions
- started suggesting functions and aggregation operators through `code_completion`, so that the suggestions can be differentiated and later also include its own help
- added help to metrics, a more in depth description of a given metric from how it was set in prometheus

https://github.com/user-attachments/assets/68ddd506-b121-43ae-9645-9220ebb5e40b

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/discover/public/application/view_components/utils/use_prometheus.ts
```diff
@@ -28,7 +29,7 @@ export const usePrometheus = (): PrometheusContext => {
     services: { data },
   } = useOpenSearchDashboards<DiscoverViewServices>();
   const prometheusResourceClient = useMemo(() => {
-    return data.resourceClientFactory.create('prometheus');
+    return data.resourceClientFactory.get('prometheus') as PrometheusResourceClient;
```
d-buckner, line 32: Moving forward we can figure make sure the Prometheus client factory gets vended properly in this situation. Happy to sync with you on suggestions here.

PR #9628, opened 2025-03-29
[TEST] Reenable saved search test and convert it to use before and after

angle943:
### Description

- Re-enable the skipped saved search tests
- Convert the test to use `before` and `after`

## Changelog

- test: reenable saved search cypress tests

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/01/saved_search.spec.js
```diff
@@ -60,6 +59,7 @@ const runSavedSearchTests = () => {
         });
 
         cy.setDataset(config.dataset, DATASOURCE_NAME, config.datasetType);
+        cy.osd.grabIdsFromDiscoverPageUrl();
```
angle943, line 62: before this change, we were grabbing the aliases for WORKSPACE_ID, DATASOURCE_ID, and INDEX_PATTERN_ID when creating a workspace, data source, etc.

We need these aliases to make an API call in these tests to created a saved search.

But the way cypress aliases work is that they do not persist between tests. That is why this test could not use `before` instead of `beforeEach`. But with my change, the aliases are grabbed at the beginning of each tests, and so we can successfully move to using `before/after`

angle943:
Screenshot 2025-03-31 at 2 15 49 PM

Ran it with Node 20. all passes. Also tested on neo

cypress/utils/commands.osd.js
```diff
@@ -291,6 +291,41 @@ cy.osd.add('grabDataSourceId', (workspaceName, dataSourceName) => {
...
+  cy.wait(2000);
+  cy.get('button.euiFilterButton').click();
+  cy.get('.euiFilterSelect__items')
+    .contains(/search/)
+    .click();
+  cy.wait(2000);
+  cy.getElementByTestId('checkboxSelectAll').click();
+  cy.getElementByTestId('savedObjectsManagementDelete').click();
+  cy.getElementByTestId('confirmModalConfirmButton').click();
+});
```
angle943, line 327: i am adding this because before, when using `beforeEach`, we didn't need to worry about deleting the created saved searches since we are starting clean for each test. But with the usage of `before`, this is no longer the case and there are conflicts in the name saving

cypress/utils/apps/query_enhancements/saved.js
```diff
@@ -540,7 +535,7 @@ export const updateSavedSearchAndSaveAndVerify = (
   // Load updated saved search and verify
   cy.getElementByTestId('discoverNewButton').click();
   // wait for the new tab to load
-  cy.getElementByTestId('docTableHeader').should('be.visible');
+  cy.wait(2000);
```
sejli, line 538: Don't remember off the top of my head, but isn't there a command we can use to verify that the tab has finished loading instead of using a `cy.wait()`? Any reason why we're removing the `docTableHeader` check?

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/01/saved_search.spec.js
```diff
@@ -29,7 +29,7 @@ const workspaceName = getRandomizedWorkspaceName();
 const runSavedSearchTests = () => {
   describe('saved search', () => {
     // TODO: Currently we cannot convert this into a "before" and "after" due to us grabbing several aliases that are required by postRequestSaveSearch()
```
Maosaic, line 31: No longer need this comment?

cypress/utils/commands.osd.js
```diff
@@ -291,6 +291,41 @@ cy.osd.add('grabDataSourceId', (workspaceName, dataSourceName) => {
...
+    if (datasourceIdMatch && datasourceIdMatch[1]) {
+      cy.wrap(datasourceIdMatch[1]).as('DATASOURCE_ID');
+    }
+    if (indexPatternIdMatch && indexPatternIdMatch[1]) {
+      cy.wrap(indexPatternIdMatch[1]).as('INDEX_PATTERN_ID');
+    }
+  });
+});
+
+cy.osd.add('deleteAllSavedSearches', (workspaceName) => {
```
sejli, line 312: Could we skip the UI here and just use the saved objects API?

sejli, line 312: Could we just use the saved objects API to delete objects?

kavilla, line 312: instead of going through the ui, we should use the saved objects api and just make it generic for other saved objects

like this

```
cy.osd.add('deleteSavedObject', (type, id, options = {}) => {
  const url = `${Cypress.config().baseUrl}/api/saved_objects/${type}/${id}`;

  return cy.request({
    method: 'DELETE',
    url,
    headers: {
      'osd-xsrf': true,
    },
    failOnStatusCode: false,
    ...options,
  });
});

cy.osd.add('deleteSavedObjectByType', (type, search) => {
  const searchParams = new URLSearchParams({
    fields: 'id',
    type,
  });

  if (search) {
    searchParams.set('search', search);
  }

  const url = `${
    Cypress.config().baseUrl
  }/api/opensearch-dashboards/management/saved_objects/_find?${searchParams.toString()}`;

  return cy.request(url).then((response) => {
    response.body.saved_objects.map(({ type, id }) => {
      cy.deleteSavedObject(type, id);
    });
  });
});
```

workspace should just be a reference and not really a permission modal so i dont think it is needed if we just hit the api directly

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/01/saved_search.spec.js
angle943: oops will remove

cypress/utils/apps/query_enhancements/saved.js
angle943: Ah so just a bit of context, the docTableHeader is the header of the table when there are results. Previously when you click on "new", it would open up a new state but with previous config applied. It no longer does that, so when clicking on new, there are no results available. So the 'docTableHeader' began failing.

There is [cy.waitForLoader](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/cypress/utils/commands.osd.js#L261-L274), but this would not work because it is checking for something that already exists even before you click on the 'new' button.

There is also [cy.waitForSearch](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/38a0d9f9bbeea7dcadba6d71f8a7a43133a89cec/cypress/utils/apps/data_explorer/commands.js#L45-L53) but this also wouldn't work because the docTable will not exist.

but the `cy.wait(2000)` is lazy, let me come up with something better

cypress/utils/commands.osd.js
angle943: oh shoot that is genius, thank you! I will look into implementing something like this

cypress/utils/apps/query_enhancements/saved.js
angle943: i updated with `cy.getElementByTestId('loadingSpinnerText').should('not.exist');` (this waits until the spinner is gone)

cypress/utils/commands.osd.js
angle943: implemented it below using @kavilla 's recommendation

angle943: implemented. I slightly altered the code though, fort `deleteSavedObjectByType`, I made it also accept a workspace id as a required prop. Without the workspace ID, it will delete all saved objects for that type, irregardless of workspace. This can be dangerous when tests are running in parallel. But scoping it to workspace id should mitigate the risk.

However, I didn't make workspace ID required for `deleteSavedObject`, since it accepts an `id`, and that'll be unique

cypress/utils/apps/query_enhancements/commands.js
```diff
@@ -101,6 +101,7 @@ Cypress.Commands.add('setQueryEditor', (value, options = {}) => {
       .should('be.visible')
       .wait(200)
       .type(escape ? `${value}{esc}` : value, {
+        delay: 40,
```
angle943, line 104: i noticed that the tests were sometimes failing locally because it wouldn't fully type out the query string it is supposed to type. Sometimes it would skip a letter or so on. It might be related to the monaco update? The default delay is 10, i upped it to 40ms.

Maosaic (approved):

cypress/utils/apps/query_enhancements/commands.js
sejli: I remember this was a common issue with Cypress typing. Usually, playing around with the delay settings work, but I used to rerun the tests a few hundred times to ensure that it won't be flaky. I'll monitor the Cypress tests on our CI; if they continue to be flaky on typing, I think we should revisit this to figure out a consistent solution.

cypress/utils/commands.osd.js
```diff
@@ -291,6 +291,57 @@ cy.osd.add('grabDataSourceId', (workspaceName, dataSourceName) => {
...
+    url,
+    headers: {
+      'osd-xsrf': true,
+    },
+    failOnStatusCode: false,
+    ...options,
+  });
+});
+
+cy.osd.add('deleteSavedObjectsByType', (workspaceId, type, search) => {
```
sejli, line 326: Nice! We can probably use this for all of our other tests as well.

sejli (approved):

PR #9637, opened 2025-04-01
fix: trim the url for creating data source

s-zx:
### Description

Error may occur if spaces exist in data source url. To solve this problem, this PR trims the data source url when user is connecting to the data source. 

### Issues Resolved

None

## Screenshot

https://github.com/user-attachments/assets/73469f9d-4b33-40aa-ba92-7a6b670328d7

## Testing the changes

In the data source URL field, if multiple spaces are entered before or after the URL, these leading or trailing spaces will be automatically removed when the field loses focus.

## Changelog
- fix: trim the url for creating data source

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_source_management/public/components/create_data_source_wizard/components/create_form/create_data_source_form.tsx
```diff
@@ -373,6 +373,13 @@ export class CreateDataSourceForm extends React.Component<
     return null;
   };
 
+  handleUrlBlur = () => {
+    this.setState((prevState) => ({
+      endpoint: prevState.endpoint.trim(),
```
wanglam, line 378: Can we add some unit tests to cover this endpoint trimming?

src/plugins/data_source_management/public/components/create_data_source_wizard/components/create_form/create_data_source_form.tsx
```diff
@@ -373,6 +373,13 @@ export class CreateDataSourceForm extends React.Component<
...
+    }));
+    this.validateEndpoint();
```
wanglam, line 380: The `setState` operation is not a synchronous method, which means the state won't be updated immediately. Do we need to move this validateEndpoint to the callback parameter of `this.setState`?

src/plugins/data_source_management/public/components/create_data_source_wizard/components/create_form/create_data_source_form.tsx
```diff
@@ -629,7 +636,7 @@ export class CreateDataSourceForm extends React.Component<
                 )}
                 isInvalid={!!this.state.formErrorsByField.endpoint.length}
                 onChange={this.onChangeEndpoint}
-                onBlur={this.validateEndpoint}
+                onBlur={this.handleUrlBlur}
```
wanglam, line 639: What about renaming to `handleEndpointBlur` to align with the state key?

s-zx: Yes, I think this is better.

src/plugins/data_source_management/public/components/create_data_source_wizard/components/create_form/create_data_source_form.tsx
s-zx: OK, I will do that.

src/plugins/data_source_management/public/components/create_data_source_wizard/components/create_form/create_data_source_form.tsx
s-zx: You are right, although trimming has no impact on validation, it's better to validate in the callback.

src/plugins/data_source_management/public/components/create_data_source_wizard/components/create_form/create_data_source_form.tsx
ruanyl, line 378: I feel it's better to do the `trim()` in `onClickCreateNewDataSource` before submitting to have better separate of concerns. We can leave the `validateEndpoint` as it is.

s-zx: Yes, I think this is a good way if user don't need to know about the trimming of the url.

ruanyl (dismissed):

src/plugins/data_source_management/public/components/create_data_source_wizard/components/create_form/create_data_source_form.test.tsx
```diff
@@ -160,6 +160,24 @@ describe('Datasource Management: Create Datasource form', () => {
...
+    /* set form fields */
+    setAuthTypeValue(authTypeIdentifier, AuthType.UsernamePasswordType);
+    changeTextFieldValue(titleIdentifier, 'test');
+    changeTextFieldValue(descriptionIdentifier, 'test');
+    changeTextFieldValue(endpointIdentifier, '       https://testTrimEndpoint.com       ');
+    changeTextFieldValue(usernameIdentifier, 'test123');
+    changeTextFieldValue(passwordIdentifier, 'test123');
+    findTestSubject(component, 'createDataSourceTestConnectionButton').simulate('click');
+    findTestSubject(component, 'createDataSourceButton').simulate('click');
+    const testEndpoint =
```
wanglam, line 174: What about replace these with code below?
```
    expect(mockTestConnectionHandler).toHaveBeenLastCalledWith(
      expect.objectContaining({
        endpoint: 'https://testTrimEndpoint.com',
      })
    );
```
Current implementation looks a little bit complicated to me.

s-zx: Cool! This is much more better.

Hailong-am (approved):

wanglam (approved):

angle943 (approved):

Issue #9588, opened 2025-03-21
[BUG] Dashboards' 'Add panels' popover does not dismiss when you navigate to other areas of the app


jhenley-icr-team:
**Describe the bug**

When in edit mode for a dashboard on the Add panels popover, if one selects the hamburger menu the panel doesn't close allowing one to navigate to other parts of the ui while the popover continues to be present. 

**To Reproduce**
1. Go to a dashboard, click edit, then click add
2. Now with the Add panels popover open click the hamburger menu and select 
3. Click Discover
4. See the discover page but with the add panels popover still present.

**Expected behavior**
Navigating away from the dashboard one was editing should close the add panels popover.

**OpenSearch Version**
2.18.0

**Dashboards Version**
2.18.0

**Plugins**
None

**Screenshots**

![Image](https://github.com/user-attachments/assets/71f7ed4f-d2b9-479f-9ed4-f75910be8131)

![Image](https://github.com/user-attachments/assets/b1a68d7b-f80d-4f47-9afa-6b6f2adb81ad)

![Image](https://github.com/user-attachments/assets/0550ee5e-2f2f-48d3-ad6a-635e6a4103e5)

![Image](https://github.com/user-attachments/assets/c6c0dd39-580a-4873-9d44-e44a57328969)

![Image](https://github.com/user-attachments/assets/126751f0-95bf-4d84-822f-c31db2bf541d)

krisfreedain:
Catch all [triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain)

PR #9643, opened 2025-04-02
fix: close flyout when clicking outside

s-zx:
### Description

This PR solves the BUG where the web page changes while the flyout is not closed. 

The source of this BUG is that the EuiFlyout is not aware of clicking events happen outside of it (header element in this case), hence will not call the `onClose` callback.

 Adding a `outsideClickCloses` paramenter to `EuiFlyout`  can solve this issue.

### Issues Resolved

fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9588

## Screenshot

https://github.com/user-attachments/assets/2fb4c279-c53e-4a3c-919c-21cfe0f5b27a

### Dashboard
1. Go to a dashboard, click edit, then click add.
2. Now with the Add panels popover.
3. Click three buttons on the top left nav, the flyout should disappear.

### Discover
1. Go to Discover.
2. Click "Open" or "Inspect" in the top right nav.
4. A flyout will appear on the right.
5. Click three buttons on the top left nav, the flyout should disappear.

## Changelog

- fix: Close flyout when clicking outside

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


d-buckner:
>The source of this BUG is that the EuiFlyout is not aware of clicking events happen outside of it (header element in this case), hence will not call the onClose callback.

Thanks for description, makes a lot of sense. Can we address the bug in EUI itself so everyone using EuiFlyout component can benefit?

s-zx:
> > The source of this BUG is that the EuiFlyout is not aware of clicking events happen outside of it (header element in this case), hence will not call the onClose callback.
> 
> Thanks for description, makes a lot of sense. Can we address the bug in EUI itself so everyone using EuiFlyout component can benefit?

Thanks for your suggestion. I am not sure what do you mean by "address the bug in EUI itself". If you mean address the bug by passing `outsideClickCloses=true` directly to the `EuiFlyout` in `src/core/public/overlays/flyout/flyout_service.tsx`, I am concerned about bringing other unexpected bugs by doing so. For this reason, I want to solve this bug by simply passing a `outsideClickCloses` parameter individually, which will have no impact on other existing flyouts. 
Additionally, after testing most of the flyouts, I only found this BUG exists in the three flyout components changed in this PR. So this BUG is not widely existed.

image

d-buckner:
> > > The source of this BUG is that the EuiFlyout is not aware of clicking events happen outside of it (header element in this case), hence will not call the onClose callback.
> > 
> > 
> > Thanks for description, makes a lot of sense. Can we address the bug in EUI itself so everyone using EuiFlyout component can benefit?
> 
> Thanks for your suggestion. I am not sure what do you mean by "address the bug in EUI itself". If you mean address the bug by passing `outsideClickCloses=true` directly to the `EuiFlyout` in `src/core/public/overlays/flyout/flyout_service.tsx`, I am concerned about bringing other unexpected bugs by doing so. For this reason, I want to solve this bug by simply passing a `outsideClickCloses` parameter individually, which will have no impact on other existing flyouts. Additionally, after testing most of the flyouts, I only found this BUG exists in the three flyout components changed in this PR. So this BUG is not widely existed.
> 
> image

Totally valid concern. Sounds like this is a good way to proceed given the low surface area of the bug. Thanks for the context.

d-buckner (approved):

wanglam (approved):
LGTM

SuZhou-Joe:
The flows keep failing(after retried 4 times), @s-zx could you please confirm if that is flaky or some regression introduced after this change?

s-zx:
> The flows keep failing(after retried 4 times), @s-zx could you please confirm if that is flaky or some regression introduced after this change?

Thank you for letting me know, I will look at it.

PR #9647, opened 2025-04-02
feat: support streaming when content type is event stream

### LINKED ISSUE #9604: [RFC] Support streaming / non-JSON request
## Background

OpenSearch Dashboards (OSD) currently handles HTTP requests with a JSON-centric approach, assuming short-lived interactions with immediate full responses. This works for traditional dashboards but limits scenarios requiring real-time streaming (e.g., chatbots, live data pipelines, or AI/ML integrations). Streaming protocols like Server-Sent Events (SSE) or chunked responses are increasingly critical for these use cases. Today, OSD’s http module parses responses as JSON by default, making streaming impossible. This RFC proposes solutions to enable streaming support.

## Proposal
Extend OSD’s HTTP capabilities to support streaming responses. Three options are under consideration:

---

### **Option 1: Add `pureFetch` Flag to HTTP Module**
Introduce a `pureFetch` option in the `http` service to bypass response parsing and return the raw `Response` object.

**Implementation**:
```typescript
// Example usage in a plugin
// Using standard AbortController
const controller = new AbortController();
const response = await http.fetch('/path/to/stream', {
  pureFetch: true, // Disables JSON parsing
  headers: { Accept: 'text/event-stream' },
  signal: controller.signal
});

// Handle stream via reader.read()
const reader = response.body.getReader();

// Terminate stream
controller.abort();
```

**Pros**:
- Minimal changes to existing code.
- Flexible: Plugins handle streams as needed (SSE, custom protocols, etc.).
- Uses existing Fetch API standards.

**Cons**:
- Requires plugins to manually manage streams (e.g., error handling, cleanup).
- No built-in support for SSE-specific features (automatic reconnection, event types).

---

### **Option 2: Expose `getRequest` Private Method**
Make the private `getRequest` method in `src/core/public/http/fetch.ts` public, allowing plugins to construct custom requests.

**Implementation**:
```typescript
// Example usage
const request = http.getRequest('/stream', { method: 'GET' });
// Using standard AbortController
const controller = new AbortController();
const response = await fetch(request, {
  signal: controller.signal
});
// Handle raw response

// Terminate stream
controller.abort();
```

**Pros**:
- Maximum flexibility for plugins.
- Avoids modifying existing `http.fetch` logic.

**Cons**:
- Exposes internal implementation details (risk of breaking changes in future OSD updates).
- Bypasses OSD’s centralized request/response handling (security, interceptors, etc.).
- Violates encapsulation principles.

---

### **Option 3: Implement SSE Support**
Add a dedicated SSE method to the HTTP module (e.g., `http.sse()`), ~~leveraging the `EventSource` API(not possible because EventSource API does not support customized headers.)~~.

**Implementation**:
```typescript
// Example usage
const eventSource: EventSource = http.sse('/sse-endpoint');

// Plugin can choose to close the eventSource.
eventSource.onMessage();
eventSource.close();
```

**Pros**:
- Standardized SSE support out-of-the-box.
- Handles reconnection, event parsing, and errors automatically.
- Cleaner API for SSE-specific use cases.

**Cons**:
- Limited to SSE (not usable for non-SSE streaming).
- Requires server-side SSE implementation.
- Adds maintenance overhead for a protocol-specific solution.
- Native SSE does not support customized headers, we need to include `@microsoft/fetch-event-source` to support those customized headers in OpenSearch Dashboards request, which is using fetch to mimic SSE actually.
- SSE only support `GET` method.

---

## **Recommendation**
**Combine Option 1 and Option 3**:
1. **Implement `pureFetch`** for general streaming use cases (e.g., raw NDJSON, custom protocols).
2. **Add an SSE-specific helper** (e.g., `http.sse()`) to simplify SSE adoption.

**Why This Hybrid Approach?**  
- **Flexibility**: `pureFetch` supports non-SSE streaming (e.g., large file downloads, custom binary protocols).
- **Convenience**: Dedicated SSE handling avoids redundant code in plugins.
- **Future-Proof**: Covers both generic streaming and the most common streaming paradigm (SSE).

---

### **Comparison of Options**
| **Criteria**               | Option 1 (`pureFetch`) | Option 3 (SSE) | Hybrid (1 + 3) |
|-----------------------------|------------------------|----------------|----------------|
| Supports raw streams        | ✅                     | ❌             | ✅             |
| Built-in SSE parsing        | ❌                     | ✅             | ✅             |
| Reconnection handling       | ❌                     | ✅             | ✅             |
| Custom protocols            | ✅                     | ❌             | ✅             |
| Maintenance overhead        | Low                    | Moderate       | Moderate       |
| Plugin developer effort     | High (DIY SSE)         | Low            | Low            |

SuZhou-Joe:
### Description

This PR supports returning pure fetch response when content-type in response header is `text/event-stream`. And I searched for the whole project that no API is using `text/event-stream` as its response content type. https://github.com/search?q=org%3Aopensearch-project%20text%2Fevent-stream&type=code

### Issues Resolved

closes #9604 

## Screenshot

https://github.com/opensearch-project/dashboards-assistant/pull/493

## Testing the changes

Unit test has been added, and you can find the consumer in: https://github.com/opensearch-project/dashboards-assistant/pull/493

## Changelog

- feat: support streaming when content type is event stream

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/public/http/fetch.test.ts
+++ b/src/core/public/http/fetch.test.ts
@@ -481,6 +481,21 @@ describe('Fetch', () => {
 
       expect(ndjson).toEqual(content);
     });
+
+    it('should make requests for text/event-stream content', async () => {
+      const body = 'data: ';
+      fetchMock.post('*', () => ({
+        body,
+        headers: { 'Content-Type': 'text/event-stream' },
+      }));
+
+      const data = await fetchInstance.post('/my/path', {
+        body,
+        asResponse: true,
+      });
+
+      expect(await data.response?.text()).toEqual(body);
+    });
   });
 
   describe('interception', () => {
--- a/src/core/public/http/fetch.test.ts
+++ b/src/core/public/http/fetch.test.ts
@@ -495,6 +495,22 @@ describe('Fetch', () => {
       });
 
       expect(await data.response?.text()).toEqual(body);
+      expect(data.body === data.response?.body).toEqual(true);
+    });
+
+    it('should return pure body for text/event-stream;charset=UTF-8 content', async () => {
+      const body = 'data: ';
+      fetchMock.post('*', () => ({
+        body,
+        headers: { 'Content-Type': 'text/event-stream;charset=UTF-8' },
+      }));
+
+      const data = await fetchInstance.post('/my/path', {
+        body,
+        asResponse: true,
+      });
+
+      expect(data.body === data.response?.body).toEqual(true);
     });
   });
 
--- a/src/core/public/http/fetch.ts
+++ b/src/core/public/http/fetch.ts
@@ -53,7 +53,7 @@ interface Params {
 
 const JSON_CONTENT = /^(application\/(json|x-javascript)|text\/(x-)?javascript|x-json)(;.*)?$/;
 const NDJSON_CONTENT = /^(application\/ndjson)(;.*)?$/;
-const EVENT_STREAM_CONTENT = /text\/event-stream/;
+const EVENT_STREAM_CONTENT = /^(text\/event-stream)(;.*)?$/;
 
 const removedUndefined = (obj: Record<string, any> | undefined) => {
   return omitBy(obj, (v) => v === undefined);
--- a/src/core/public/http/fetch.ts
+++ b/src/core/public/http/fetch.ts
@@ -53,6 +53,7 @@ interface Params {
 
 const JSON_CONTENT = /^(application\/(json|x-javascript)|text\/(x-)?javascript|x-json)(;.*)?$/;
 const NDJSON_CONTENT = /^(application\/ndjson)(;.*)?$/;
+const EVENT_STREAM_CONTENT = /text\/event-stream/;
 
 const removedUndefined = (obj: Record<string, any> | undefined) => {
   return omitBy(obj, (v) => v === undefined);
@@ -195,6 +196,11 @@ export class Fetch {
         body = fetchOptions.withLongNumeralsSupport
           ? parse(await response.text())
           : await response.json();
+      } else if (EVENT_STREAM_CONTENT.test(contentType)) {
+        // Browsers often need to buffer the entire response before decompressing, which defeats the purpose of streaming.
+        // So for plugins who want to use streaming experience,
+        // please remember to set 'Content-Encoding' as 'identity' or ''
+        body = response.body;
       } else {
         const text = await response.text();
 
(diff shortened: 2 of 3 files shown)


src/core/public/http/fetch.ts
```diff
@@ -53,6 +53,7 @@ interface Params {
 
 const JSON_CONTENT = /^(application\/(json|x-javascript)|text\/(x-)?javascript|x-json)(;.*)?$/;
 const NDJSON_CONTENT = /^(application\/ndjson)(;.*)?$/;
+const EVENT_STREAM_CONTENT = /text\/event-stream/;
```
Hailong-am, line 56: nit: better to align with json and ndjson defined above? `/^(text\/event-stream)(;.*)?$/;`

src/core/public/http/fetch.test.ts
```diff
@@ -481,6 +481,21 @@ describe('Fetch', () => {
 
       expect(ndjson).toEqual(content);
     });
+
+    it('should make requests for text/event-stream content', async () => {
+      const body = 'data: ';
+      fetchMock.post('*', () => ({
+        body,
+        headers: { 'Content-Type': 'text/event-stream' },
```
Hailong-am, line 489: can we add a case which headers is `{ 'Content-Type': 'text/event-stream;charset=UTF-8'},`

src/core/public/http/fetch.test.ts
```diff
@@ -481,6 +481,21 @@ describe('Fetch', () => {
...
+      }));
+
+      const data = await fetchInstance.post('/my/path', {
+        body,
+        asResponse: true,
+      });
+
+      expect(await data.response?.text()).toEqual(body);
```
wanglam, line 497: Nit: Do we need to check if `data.response.body` was returned as expected here?

wanglam (approved):

src/core/public/http/fetch.test.ts
SuZhou-Joe: Sure, added one more assertion data.response.body === data.body.

src/core/public/http/fetch.test.ts
SuZhou-Joe: Sure, added.

src/core/public/http/fetch.ts
SuZhou-Joe: Sure, updated.

Hailong-am (approved):

wanglam (approved):

Issue #9650, opened 2025-04-03
[BUG] Data source client should honor the timeout settings in opensearch_dashboards.yml


gaobinlong:
**Describe the bug**

For now, the data source opensearch client doesn't honor the `opensearch.requestTimeout` settings in the yaml file, so the [timeout is 30s by default](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/0fed5eb1a95ba4c143bd48ae1f9508dc21c739bf/src/plugins/data_source/server/client/client_config.ts#L62) even though a bigger value is set in the yaml file, this makes users confused.

We should make the data source opensearch client honor the timeout settings specified in the yaml file.

**To Reproduce**
1. Set `opensearch.requestTimeout` to 120s in the yaml file.
2. Trigger a search request in Discover page against a OpenSearch cluster which runs in debug mode an will sleep 65s when searching.
3. After 2 minutes, you can see that there's a frontend error 'Failed to fetch', which is not as expected, the search request should get response from OpenSearch after 65s. So why we wait 2 minutes, that's because the node server will retry the search request 3 times if the first request didn't get response after the default timeout 30s, so after 4*30s=120s, the frontend still cannot get any response, an error 'Failed to fetch' is thrown.

**Expected behavior**
 The data source opensearch client honor the timeout settings specified in the yaml file.

**OpenSearch Version**
3.0.0

**Dashboards Version**
3.0.0

**Plugins**

Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

krisfreedain:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/andrross)

PR #9651, opened 2025-04-03
Data source opensearch client honors the timeout settings in yaml file

gaobinlong:
### Description

The data source opensearch client does not honor the timeout settings specified in the yaml file, this PR fixes the bug and adds some unit tests.

### Issues Resolved

Fix https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9650.

## Screenshot

No UI change, only change the node server code.

Before this bug fix, we get `Failed to fetch` error even the OpenSearch cluster returns search result after 30s, see below:

image

, and after this change, the Discover page can show the search result normally if OpenSearch can return the search result after 30s and before the specified opensearch.requestTimeout setting in the yaml file elapses.

Another case, if OpenSearch returns some error after 30s, then the Discover page shows the error directly rather than showing the frontend error `Failed to fetch` after retries:

Pasted Graphic 2

## Testing the changes

1. Set opensearch.requestTimeout to 120s and opensearch.shardTimeout to 80s in the yaml file.
2. Trigger a search request in Discover page against a OpenSearch cluster which runs in debug mode an will sleep 65s when searching.
3. Discover page can show the search result normally after 65s when it receives the response.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_source/server/client/client_config.test.ts
```diff
@@ -17,9 +18,13 @@ describe('parseClientOptions', () => {
...
+        requestTimeout: duration(1, 'seconds'),
+        pingTimeout: duration(2, 'seconds'),
+      },
     } as DataSourcePluginConfigType;
 
-    expect(parseClientOptions(config, TEST_DATA_SOURCE_ENDPOINT)).toEqual(
+    expect(parseClientOptions(config, TEST_DATA_SOURCE_ENDPOINT, [])).toEqual(
       expect.objectContaining({
         node: TEST_DATA_SOURCE_ENDPOINT,
         ssl: {
```
Hailong-am, line 30: can we verify `requestTimeout` and `pingTimeout` too?

src/plugins/data_source/server/client/client_config.test.ts
```diff
@@ -39,8 +44,12 @@ describe('parseClientOptions', () => {
       clientPool: {
         size: 5,
       },
+      globalOpenSearchConfig: {
+        requestTimeout: duration(1, 'seconds'),
+        pingTimeout: duration(2, 'seconds'),
+      },
     } as DataSourcePluginConfigType;
-    expect(parseClientOptions(config, TEST_DATA_SOURCE_ENDPOINT)).toEqual(
+    expect(parseClientOptions(config, TEST_DATA_SOURCE_ENDPOINT, [])).toEqual(
```
Hailong-am, line 52: same here

Hailong-am (dismissed):

src/plugins/data_source/server/client/client_config.test.ts
gaobinlong: Sure, added yet, thank you!

gaobinlong:
@kristenTian @zhongnansu could you help to review this PR, thanks!

gaobinlong:
Hi @Flyingliuhub, mind taking a look at this PR, thank you!

Hailong-am (approved):

src/plugins/data_source/config.ts
```diff
@@ -5,6 +5,8 @@
 
 import { schema, TypeOf } from '@osd/config-schema';
 import { fileAppenderSchema } from './audit_config';
+// eslint-disable-next-line @osd/eslint/no-restricted-paths
```
d-buckner, line 8: What's the reason for overriding and using restricted paths?

gaobinlong: Using restricted path is because the interface `DataSourcePluginConfigType` is also referenced by the public module, so eslint throws `Server modules cannot be imported into client modules or shared modules`, but I see the same issue in audit_config.ts: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source/audit_config.ts#L10, so maybe it's OK.

d-buckner: Can you take a look at how others have done this in the past? Ref: https://github.com/search?q=repo%3Aopensearch-project%2FOpenSearch-Dashboards+SharedGlobalConfig&type=code

These eslint rules are a really valuable mechanism used by the platform to codify supported usages/patterns, we want to be really intentional about when and why we suppress these.

SuZhou-Joe:
> The data source opensearch client honor the timeout settings specified in the yaml file

Guess a typo, should be `does not honor`? And another question is that should we create configs for data source plugin instead of reusing the opensearch client config for local cluster?

gaobinlong:
> > The data source opensearch client honor the timeout settings specified in the yaml file
> 
> Guess a typo, should be `does not honor`? And another question is that should we create configs for data source plugin instead of reusing the opensearch client config for local cluster?

Thanks, changed the description. Add some independent timeout settings for data source plugin make sense, data source plugin should honor the plugin-level timeout settings firstly and then honor the global setting secondarily, we may implement this in another PR.

src/plugins/data_source/config.ts
gaobinlong: Moving `SharedGlobalConfig` to common will change many files referencing it, I've changed it by defining the timeout fields directly, so `SharedGlobalConfig` is not needed to import, please help to take a look, thanks!

d-buckner (approved):

src/plugins/data_source/server/plugin.ts
```diff
@@ -57,7 +60,15 @@ export class DataSourcePlugin implements Plugin<DataSourcePluginSetup, DataSourc
...
-    const config: DataSourcePluginConfigType = await this.config$.pipe(first()).toPromise();
+    const pluginConfig: ConfigSchema = await this.config$.pipe(first()).toPromise();
+    const globalConfig = await this.globalConfig$.pipe(first()).toPromise();
+    const config = {
+      ...pluginConfig,
+      globalOpenSearchConfig: {
+        requestTimeout: globalConfig.opensearch.requestTimeout,
+        pingTimeout: globalConfig.opensearch.pingTimeout,
+      },
+    } as DataSourcePluginConfigType;
```
SuZhou-Joe, line 71: Seems the type cast here is not needed.

SuZhou-Joe (approved):

src/plugins/data_source/config.ts
```diff
@@ -61,4 +61,11 @@ export const configSchema = schema.object({
...
 
-export type DataSourcePluginConfigType = TypeOf<typeof configSchema>;
+export type ConfigSchema = TypeOf<typeof configSchema>;
+
+export type DataSourcePluginConfigType = ConfigSchema & {
+  globalOpenSearchConfig: {
+    requestTimeout: Duration;
+    pingTimeout: Duration;
+  };
+};
```
Hailong-am, line 71: nit: Not related to this change, maybe we should add datasource own request timeout

Hailong-am (approved):

PR #9655, opened 2025-04-04
Extend the query editor extension to include action bar button area

angle943:
### Description

- Extend the Data plugin's `__enhance` to include an area to put buttons for the ResultsActionBar
- update `__enhance` to be experimental
- Rename `getSearchBarButton` to `getQueryControlButtons`
- Add documentation on how to use it

## Changelog
- feat: enable experimental data plugin's __enhance
- feat: add resultsActionBar into data plugin's __enhance

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extension.tsx
```diff
@@ -70,27 +76,43 @@ export interface QueryEditorExtensionConfig {
...
   /**
    * A function that returns the query editor extension banner. The banner is a
    * component that will be displayed on top of the search bar.
    * @param dependencies - The dependencies required for the extension.
-   * @returns The component the query editor extension.
+   * @returns The query editor extension component.
    */
   getBanner?: (dependencies: QueryEditorExtensionDependencies) => React.ReactElement | null;
-
-  getSearchBarButton?: (
```
angle943, line 84: i ended up changing this function name to `getQueryControlButtons`. Maybe long time ago this placed stuff near. search bar, but now it puts it in the `queryControls`. If you also look at the code below, this is renamed to `queryContrtols` even before this change. 

This is a breaking change, but this feature was marked as `@internal` before my change so I think we are safe to change it

src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extension.tsx
```diff
@@ -10,15 +10,20 @@ import { Observable } from 'rxjs';
 import { DataStructureMeta, Query } from '../../../../common';
 import { ResultStatus } from '../../../query/query_string/language_service/lib';
 
+// ID for the container that will house the action bar buttons. Should be used by Discover Plugin's ResultsActionBar component
+export const ACTION_BAR_BUTTONS_CONTAINER_ID = 'query-editor-action-bar-buttons-container';
```
angle943, line 14: this may appear not to be the best place to put this in, but since its being read by the discover plugin's `ResultActionBar`, I wanted devs to be able to cmd+f where it comes from (namely, this file)

src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extensions.tsx
```diff
@@ -33,11 +34,14 @@ const QueryEditorExtensions: React.FC<QueryEditorExtensionsProps> = React.memo((
...
   }, [configMap]);
 
+  const actionBarContainer = document.getElementById(ACTION_BAR_BUTTONS_CONTAINER_ID);
+
   return (
     <>
       {sortedConfigs.map((config) => {
         const extensionComponentId = `osdQueryEditorExtensionComponent-${config.id}`;
         const extensionQueryControlsId = `osdQueryEditorExtensionQueryControls-${config.id}`;
+        const extensionActionBarContainerId = `osdQueryEditorExtensionActionBarContainer-${config.id}`;
```
ruchidh, line 44: nit:
Is config.id guaranteed to be unique and safe for use in class names?

src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extensions.tsx
```diff
@@ -57,6 +61,16 @@ const QueryEditorExtensions: React.FC<QueryEditorExtensionsProps> = React.memo((
           queryControlsContainer.appendChild(extensionQueryControlsContainer);
         }
 
+        // Make sure action bar buttons are rendered in order
+        let extensionActionBarContainer = document.getElementById(extensionActionBarContainerId);
+        if (!extensionActionBarContainer) {
+          extensionActionBarContainer = document.createElement('div');
```
ruchidh, line 67: There is a code repeatation here seems for same logic of getting, then create and append child
may be we can make a common method and just pass the dynamic arugments .
eg.
   const extensionActionBarContainer = actionBarContainer
          ? createOrGetElement(
              `osdQueryEditorExtensionActionBarContainer-${config.id}`,
              ['osdQueryEditorExtensionActionBarContainer', `osdQueryEditorExtensionActionBarContainer__${config.id}`],
              actionBarContainer
            )
          : undefined;
          
          
         and define method 
         
      
    const createOrGetElement = (
  id:,
  classNames:
  parent: 
): {
// move the common logic of create and get here }
};

src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extensions.tsx
angle943: Yes, although perhaps unintentionally (unintentionally by us)?

Plugin owners can pass multiple configs into the data plugin, and they dictate what they want the ID to be. Nothing is stopping anyone though from passing a config twice with the same `id`. However, if this is done, due to the nature of how we store these config, the newest config will overwrite the old config with the same id (this is because we store these configs in a key-value object with the `key` being the id). So in the end it'll end up being unique.

Due to the type of customers that will use this __enhance features (namely plugin developers), I think this is fine, and they should know not to overwrite any existing `id`.

src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extensions.tsx
angle943: good idea, let me see if i can do something like this

src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extension.tsx
Maosaic: I'm more leaning towards letting discover expose the ele id, since we are trying to locate a element hosting on other plugin. Unless I'm missing something, if we have the ele id imported here, search for id string and variable reference in IDE should still work.

angle943: in my head, Discover Plugin consumes Data Plugin, so I thought it more appropriate here. Thoughts @kavilla ?

docs/plugins/data/query-editor-enhancements.md
```diff
@@ -0,0 +1,174 @@
+# Enhancing the Data Plugin
+
+# 1. Introduction
+
+One of the core functionalities of the Data plugin is the query editor, which is a key component in several places such as the Discover page.
+
+We have dedicated a few areas around the query editor to be customized by plugin developers, and this guide will walk through how to do so.
+
+**Note:** this is an experimental feature, and is subject to change at any time in the future.
```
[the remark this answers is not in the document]
sejli: +1 This is awesome

src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extension.tsx
sejli: If I remember correctly, this was only used by the same file, so should be safe to update.

src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extension.tsx
```diff
@@ -46,6 +51,7 @@ export interface QueryEditorExtensionDependencies {
   fetchStatus?: ResultStatus;
 }
 
+// When updating this please update docs/plugins/data/query-editor-enhancements.md
 export interface QueryEditorExtensionConfig {
```
sejli, line 55: Nit: Is there a reason why all of the `@returns` are the same?

angle943: they all return a react component. i could've reworded it but i think the `getX` name should already make it clear what type of component it is referencing. I just reworded it because `The component the query editor extension` doesn't make sense :D

LDrago27 (approved):

src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extension.tsx
Hailong-am: dashboard-assistant use this too, I have raised a pr to update the name as well. 
https://github.com/opensearch-project/dashboards-assistant/blob/main/public/plugin.tsx#L299-L306

Issue #9474, opened 2025-03-03
[BUG] Flashing scrollbars in Discover select data modal


brucejxz:
**Describe the bug**

In Discover, with `query:enhancements:enabled`, go to the index pattern picker, then `View all available data`, then select `Index patterns`. The scrollbars start to appear and disappear rapidly.

**To Reproduce**

https://github.com/user-attachments/assets/c1e141f2-17e1-4e3a-a45f-93ba223d3861

**Expected behavior**

Nothing

**OpenSearch Version**
2.19.0

**Dashboards Version**
2.19.0

**Plugins**

N/A

**Screenshots**

See above

**Host/Environment (please complete the following information):**
 - OS: MacOS
 - Chrome 133.0.6943.126

**Additional context**

N/A

andrross:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/andrross)

PR #9668, opened 2025-04-09
[Discover]Fix dataset selector flashing

Hailong-am:
### Description

Hide the y scroll bar to avoid the dataset selector flashing

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9474

## Screenshot

**Before**

https://github.com/user-attachments/assets/df1d755b-d808-4547-b452-1e330a8b8f30

**After**

https://github.com/user-attachments/assets/2e1a22ce-4b0a-43bc-9ede-cbeefdd63592

## Changelog

- fix: [Discover]Fix dataset selector flashing

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/dataset_selector/_dataset_explorer.scss
+++ b/src/plugins/data/public/ui/dataset_selector/_dataset_explorer.scss
@@ -8,7 +8,8 @@
   grid-template-columns: repeat(auto-fit, minmax(200px, 240px)) minmax(300px, 1fr);
   height: 100%;
   max-height: calc(100vh - 200px);
-  overflow-x: hidden;
+  overflow-x: auto;
+  scrollbar-gutter: stable;
   border: $euiBorderThin;
 
   &__column {
--- a/src/plugins/data/public/ui/dataset_selector/_dataset_explorer.scss
+++ b/src/plugins/data/public/ui/dataset_selector/_dataset_explorer.scss
@@ -9,7 +9,7 @@
   height: 100%;
   max-height: calc(100vh - 200px);
   overflow-x: auto;
-  scrollbar-gutter: stable;
+  overflow-y: hidden;
   border: $euiBorderThin;
 
   &__column {
--- a/src/plugins/data/public/ui/dataset_selector/_dataset_explorer.scss
+++ b/src/plugins/data/public/ui/dataset_selector/_dataset_explorer.scss
@@ -8,7 +8,7 @@
   grid-template-columns: repeat(auto-fit, minmax(200px, 240px)) minmax(300px, 1fr);
   height: 100%;
   max-height: calc(100vh - 200px);
-  overflow-x: auto;
+  overflow-x: hidden;
   border: $euiBorderThin;
 
   &__column {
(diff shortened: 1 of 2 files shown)


Hailong-am:
@kavilla would you mind to take a look?

src/plugins/data/public/ui/dataset_selector/_dataset_explorer.scss
```diff
@@ -8,7 +8,7 @@
   grid-template-columns: repeat(auto-fit, minmax(200px, 240px)) minmax(300px, 1fr);
   height: 100%;
   max-height: calc(100vh - 200px);
-  overflow-x: auto;
+  overflow-x: hidden;
```
SuZhou-Joe, line 11: I saw this issue several times but now not able to reproduce the issue but this setting may prevent it from scrolling. will `scrollbar-gutter: stable;` work?

Hailong-am: that works as well, but the scroll bar gutter will always present like this
image

Hailong-am: To double check the selectable items, it has css for `text-overflow: eclipses`, so this probably will not have scrollbar for x-axis

image

d-buckner: +1 to @SuZhou-Joe's suggestion. Disabling x scrollbars to address these reflow issues present an accessibility issue as the rightmost content can be no longer accessible for users with page zoom or smaller screen sizes.

SuZhou-Joe: > text-overflow: eclipses
The 3rd column'd width is auto so text-overflow may not work for the 3rd column.

> the scroll bar gutter will always present like this
Yeah gutter will preserve some width for the scrollbar and I think we may need to make some trade-off to address the flashing issue.

Hailong-am: make sense, considering accessibility which is more important, let's me switch it to `scrollbar-gutter: stable;`

Hailong-am: when switch to `scrollbar-gutter: stable;` the issue still happened, let me convert it to draft first

Hailong-am: @SuZhou-Joe @d-buckner after more research, the issue happens when both x and y scrollbar present, to hide x scrollbar may have accessibility issue for small screen size. Instead, we can hide y scrollbar to avoid they both present causing flashing. 

Each column in the grid has its own y scrollbar, the remove outer one could remove double y scrollbar too.
image

SuZhou-Joe (approved):

src/plugins/data/public/ui/dataset_selector/_dataset_explorer.scss
SuZhou-Joe: Thanks @Hailong-am for trying on those options, looks good to me.

d-buckner: Thanks for taking the time @Hailong-am! Changes look good to me as well.

d-buckner (approved):

wanglam (approved):

joshuali925:
This PR breaks the dataset table scrollbar. fix in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9804

PR #9692, opened 2025-04-18
Update default index pattern logic

lunalzm:
### Description

Currently, when a default index pattern is associated with a deleted data source, the system continues to use it as the default pattern. This PR enhances the existing logic that handles non-existent default index patterns to also check for non-existing data sources.

### Issues Resolved

fixes: Remove default index pattern when associated data source is deleted.

## Changelog

- fix: Update default index pattern logic

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/common/index_patterns/index_patterns/ensure_default_index_pattern.ts
+++ b/src/plugins/data/common/index_patterns/index_patterns/ensure_default_index_pattern.ts
@@ -71,7 +71,7 @@ export const createEnsureDefaultIndexPattern = (
               });
             }
           } catch (e) {
-            // console.error(e);
+            console.error(e);
           }
         } else {
           return;
@@ -82,7 +82,6 @@ export const createEnsureDefaultIndexPattern = (
     }
 
     if (patterns.length >= 1) {
-      // await this.setDefault(patterns[0]);
       await uiSettings.set('defaultIndex', patterns[0]);
     } else {
       const isEnhancementsEnabled = await uiSettings.get('query:enhancements:enabled');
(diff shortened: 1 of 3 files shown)


src/plugins/data/common/index_patterns/index_patterns/ensure_default_index_pattern.ts
```diff
@@ -44,26 +42,33 @@ export const createEnsureDefaultIndexPattern = (
    * one otherwise.
    */
   return async function ensureDefaultIndexPattern(this: IndexPatternsContract) {
+    const datasources = await savedObjectsClient.find({ type: 'data-source' });
```
SuZhou-Joe, line 45: data sources plugin are not always enabled, this call may fail I think.

src/plugins/data/common/index_patterns/index_patterns/ensure_default_index_pattern.ts
```diff
@@ -44,26 +42,33 @@ export const createEnsureDefaultIndexPattern = (
...
+      const refIdBool = !!refId;
+      if (!refIdBool || existDataSources.includes(refId)) {
+        availableArray.push(item.id);
+      }
+    });
     if (canUpdateUiSetting === false) {
       return;
     }
-    const patterns = await this.getIds();
     let defaultId = await uiSettings.get('defaultIndex');
```
SuZhou-Joe, line 59: What about optimizing the check flow to:

1. Get the data source from references of default index pattern
2. If there is no data source, we are good, if there is data source, try to get the data source info by using savedObjectClient.get
3. If the call fails(the data source deleted / dissociated from the workspace), get index patterns / data sources list to see if there is other index patterns can be a good candidate of default index pattern
4. Set the one as default index pattern.

By doing so, no additional call is required if the index pattern and its underlying data source is good.

PR #9694, opened 2025-04-18
Fix Ci failures

LDrago27:
### Description

The branding URL's for Opensearch were not working. This was causing a failure in the `isUrlValid()` as a result it was setting URL to default value instead of URL that was being passed. This resulted in failures of CI group 2 and Ci group 9.

## Changelog

- fix: Fix Branding test urls
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


d-buckner (approved):
Thanks for fixing! No concerns with this PR, but curious on your thoughts if it makes sense to use stubs here in the future so we're not dependent on external http requests in our tests.

LDrago27:
> Thanks for fixing! No concerns with this PR, but curious on your thoughts if it makes sense to use stubs here in the future so we're not dependent on external http requests in our tests.

Yeah . I like the idea we can stub the http response for the URLs so we that we return a valid response for these config urls.

ruchidh (approved):

Maosaic (approved):

PR #9702, opened 2025-04-23
Add default index pattern validation in Discover page

lunalzm:
### Description

This change improves the validation mechanism for default index pattern in the Discover page. It adds the checking logic to ensure proper validation of default index patterns, enhancing the user experience and preventing potential issues related to invalid index pattern selections.

### Screenshots
- Before:

https://github.com/user-attachments/assets/f785697a-1875-4b68-b1cf-f6108c44ca44

- After:

https://github.com/user-attachments/assets/af3a3950-577b-4c81-b754-0fadec695cc0

## Changelog
- feat: Add default index pattern validation in Discover page

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/discover/public/plugin.ts
```diff
@@ -277,11 +277,10 @@ export class DiscoverPlugin
...
         setHeaderActionMenuMounter(params.setHeaderActionMenu);
         syncHistoryLocations();
-        const {
-          core: {
-            application: { navigateToApp },
-          },
-        } = await this.initializeServices();
+        const { core: coreStart, plugins: pluginsStart } = await this.initializeServices();
+
+        pluginsStart.data.indexPatterns.clearCache();
```
SuZhou-Joe, line 282: Do we have to clearCache?

SuZhou-Joe:
@lunalzm Could you please post a video to show the difference before & after the fix? That should help to explain what we are trying to fix.

lunalzm:
> @lunalzm Could you please post a video to show the difference before & after the fix? That should help to explain what we are trying to fix.
OK👌

src/plugins/discover/public/plugin.ts
lunalzm: No, you're right.

src/plugins/discover/public/plugin.ts
```diff
@@ -277,11 +277,9 @@ export class DiscoverPlugin
...
         setHeaderActionMenuMounter(params.setHeaderActionMenu);
         syncHistoryLocations();
-        const {
-          core: {
-            application: { navigateToApp },
-          },
-        } = await this.initializeServices();
+        const { core: coreStart, plugins: pluginsStart } = await this.initializeServices();
+
+        await pluginsStart.data.indexPatterns.ensureDefaultIndexPattern();
```
wanglam, line 282: Seems this `ensureDefaultIndexPattern` has already been called in 
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/discover/public/application/view_components/canvas/top_nav.tsx#L105

Do we need to ensure twice in this discover page? Can we remove one of the `ensureDefaultIndexPattern` logic?

wanglam: I think we should figure out why the old ensureDefaultIndexPattern logic is not working as expected. And we need to check if there are any other places using this TopNav component before removing this ensure logic.

lunalzm: It seems that the TopNav component is only utilized in the OpenSearch-Dashboards/src/plugins/discover/public/application/view_components/canvas/index.tsx file, where it is exported as part of the DiscoverCanvas function. However, upon reviewing the codebase, it appears that there are no other files or components invoking DiscoverCanvas, which leads to the conclusion that the old ensureDefaultIndexPattern logic is not being executed as expected.

angle943:
This is Kiro posting.

No author activity in ~12 months and has merge conflicts. @lunalzm — is this still active? If yes, please rebase. If not, this should be closed.

PR #9726, opened 2025-04-25
[workspace] support multiple-scopes ui settings

Qxisylolo:
### Description
This pr refactors ui settings.

1. When saving the local cache in public side, we will
    1. Merge the configs following default priority(User level > Workspace level > Global level) if the config does not define scope or it has more than 1 scope.
    2. If the config is declared with only 1 scope, the value from specific scope will have the top priority.

2. ReadWithScope will always fetch data from specific scope / setting without merge, if the target scope does not comply with the scope defined in settings’ metadata, the method should throw error.
3. WriteWithScope will always write value into the passed in scope, if the target scope does not comply with the scope defined in settings’ metadata, the method should throw error.
4. Write should
    1. If the setting defines a single scope in its metadata, honor the scope defined.
    2. If the setting does not define a scope, we will use global to keep backward compatibility.
    3. If the setting defines multiple scopes in its metadata, throw error because UI setting does not know which scope it should use.

### Issue

### get
1. get with specific scope（out of a workspace）
in this case, `workspace_ui_setting_wrapper` should return not found error which will be catch and return {}. 
截屏2025-04-25 16 54 14

2. get with specific scope（in a workspace）
in this case, workspace_ui_setting_wrapper will return workspace meta data
截屏2025-04-25 16 57 03

3. get without scope（in a workspace）
We will retrieve all level settings and merge them based on scope priority: global < workspace < user. If a setting exists in multiple scopes and each scope has a value defined, the value from the higher-priority scope will override the lower ones.

### set
Take default data source as example，it has 2 scopes, global and workspace

1. set with specific scope（in of a workspace）
should successfully update workspace meta data
截屏2025-04-25 17 03 39

* set workspace level default data source
截屏2025-04-25 17 22 27

* set global level default data source
截屏2025-04-25 17 23 02

2.  set with specific scope（out of a workspace）
`workspace_ui_setting_wrapper`  will return error as It is forbidden to update workspace level uiSettings when out of a workspace.

截屏2025-04-25 17 04 09

3. update multi-scope settings without scope
 if we are trying to update default data source, it will throw error,  because it has multiple scopes
截屏2025-04-25 17 07 44

4. update single-scope settings without scope
we honor the scope defined. for example, default workspace is user level
截屏2025-04-25 17 08 36

## Changelog
- feat: support multiple-scopes ui settings 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/core/public/ui_settings/ui_settings_api.ts
```diff
@@ -32,14 +32,17 @@ import { BehaviorSubject } from 'rxjs';
...
 
 export interface UiSettingsApiResponse {
   settings: UiSettingsState;
 }
 
 interface Changes {
   values: {
-    [key: string]: any;
+    [scope: UiSettingScope | string]: {
+      [key: string]: any;
```
FriedhelmWS, line 44: Minor: is there anyway to not use `any` here?

src/core/public/ui_settings/ui_settings_api.ts
```diff
@@ -130,16 +141,39 @@ export class UiSettingsApi {
...
+            '/api/opensearch-dashboards/settings',
+            {
+              changes: settings,
+            },
+            { scope }
+          );
+        }
+        return this.sendRequest('POST', '/api/opensearch-dashboards/settings', {
+          changes: settings,
+        });
```
FriedhelmWS, line 161: Minor: can we do something like
```
const options = scope !== 'undefined' && scope !== undefined ? { scope } : undefined;

return this.sendRequest(
  'POST',
  '/api/opensearch-dashboards/settings',
  {
    changes: settings,
  },
  options
);
```
to remove duplicate code?

src/core/server/ui_settings/ui_settings_client.ts
```diff
@@ -268,33 +282,56 @@ export class UiSettingsClient implements IUiSettingsClient {
...
+  private groupChanges(changes: Record<string, any>) {
     const userChanges = {} as Record<string, any>;
     const globalChanges = {} as Record<string, any>;
+    const workspaceChanges = {} as Record<string, any>;
 
     Object.entries(changes).forEach(([key, val]) => {
-      if (userLevelKeys.includes(key)) {
+      if (this.uiSettingsKeysWithMoreThanOneScope.includes(key)) {
+        // if this setting has more than one scope, we should not update this setting without specifying scope
+        throw new Error(`Unable to update "${key}", because it has multiple scopes`);
```
FriedhelmWS, line 324: Minor: do we have to create a dedicated `uiSettingsKeysWithMoreThanOneScope` for duplicate scope checking? Can we checking the scope directly when processing changes?

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -110,15 +94,15 @@ export class WorkspaceUiSettingsClientWrapper {
...
-        const configObject = await wrapperOptions.client.get<Record<string, any>>(
-          'config',
-          id,
-          options
-        );
+      if (type === 'config' && id.startsWith(CURRENT_WORKSPACE_PLACEHOLDER)) {
         const savedObjectsClient = this.getWorkspaceTypeEnabledClient(wrapperOptions.request);
+        if (!requestWorkspaceId) {
+          throw new Error(
+            ' It is forbidden to update workspace level uiSettings when out of a workspace'
```
FriedhelmWS, line 103: Very minor: there's a space 😆

src/core/public/ui_settings/ui_settings_api.ts
```diff
@@ -130,16 +141,39 @@ export class UiSettingsApi {
...
+        if (result.status === 'fulfilled') {
+          // merge settings
+          Object.assign(results.settings, result.value.settings);
+        } else {
+          changes.callback(result.reason);
+          return;
+        }
+      }
+
+      changes.callback(undefined, results);
```
SuZhou-Joe, line 176: The behavior seems different, previously we emit changes.values[scope] to callback instead of the merged `results`.

src/core/public/ui_settings/ui_settings_api.ts
```diff
@@ -130,16 +141,39 @@ export class UiSettingsApi {
 
     const changes = this.pendingChanges;
     this.pendingChanges = undefined;
-
+    const results: UiSettingsApiResponse = { settings: {} };
     try {
       this.sendInProgress = true;
+      const scopeEntries = Object.entries(changes.values);
+      const requestPromises = scopeEntries.map(([scope, settings]) => {
+        if (scope !== 'undefined' && scope !== undefined) {
```
SuZhou-Joe, line 149: Do we need to check if there is pending changes inside the scope? Now if there is no pending changes, we will still send a request to server.

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -272,25 +273,36 @@ export async function getDataSourcesWithFields(
...
+    );
   }
 }
 
 export async function setFirstDataSourceAsDefault(
   savedObjectsClient: SavedObjectsClientContract,
   uiSettings: IUiSettingsClient,
-  exists: boolean
+  exists: boolean,
+  isWorkspaceLevelUpdate: boolean
```
SuZhou-Joe, line 293: Nit: what about using scope here?

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -78,22 +75,9 @@ export class WorkspaceUiSettingsClientWrapper {
...
-          ...(workspaceObject ? workspaceObject.attributes.uiSettings : {}),
-          // Workspace level default data source value should not extend global UIsettings value.
-          [DEFAULT_DATA_SOURCE_UI_SETTINGS_ID]: workspaceLevelDefaultDS,
-          // Workspace level default index pattern value should not extend global UIsettings value.
-          [DEFAULT_INDEX_PATTERN_UI_SETTINGS_ID]: workspaceLevelDefaultIndex,
-        };
-
-        return configObject as SavedObject<T>;
+        return {
+          attributes: workspaceObject?.attributes?.uiSettings || {},
```
SuZhou-Joe, line 79: Let's return id and other essential properties as well.

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -110,15 +94,15 @@ export class WorkspaceUiSettingsClientWrapper {
...
-      if (type === 'config' && requestWorkspaceId && !id.startsWith(CURRENT_USER_PLACEHOLDER)) {
-        const configObject = await wrapperOptions.client.get<Record<string, any>>(
-          'config',
-          id,
-          options
-        );
+      if (type === 'config' && id.startsWith(CURRENT_WORKSPACE_PLACEHOLDER)) {
         const savedObjectsClient = this.getWorkspaceTypeEnabledClient(wrapperOptions.request);
+        if (!requestWorkspaceId) {
+          throw new Error(
```
SuZhou-Joe, line 102: We should throw 400 error instead of a error that will be handled as 500 error eventually.

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -135,11 +119,7 @@ export class WorkspaceUiSettingsClientWrapper {
           options
         );
 
-        if (workspaceUpdateResult.attributes.uiSettings) {
-          configObject.attributes = workspaceUpdateResult.attributes.uiSettings;
-        }
-
-        return configObject as SavedObjectsUpdateResponse<T>;
+        return workspaceUpdateResult.attributes.uiSettings as SavedObjectsUpdateResponse<T>;
```
SuZhou-Joe, line 122: Still, is it able to remove the type cast here? And the value seems not consistent, previous `configObject: { attributes: finalSettings }`, not the result is the finalSettings itself.

src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -124,8 +129,8 @@ You can use \`IUiSettingsClient.get("${key}", defaultValue)\`, which will just r
     );
   }
 
-  async set(key: string, value: any) {
-    return await this.update(key, value);
+  async set(key: string, value: any, scope?: UiSettingScope) {
```
SuZhou-Joe, line 132: What about get? Do not find a place consuming `getWithScope` method of ui_settings_api.

src/core/server/ui_settings/ui_settings_client.ts
SuZhou-Joe: Agree, check that from `this.defaults` should e good enough.

src/core/server/ui_settings/ui_settings_client.ts
```diff
@@ -268,33 +282,56 @@ export class UiSettingsClient implements IUiSettingsClient {
...
+      ) {
+        this.globalLevelSettingsKeys.push(key);
       }
     });
+  }
+
+  /**
+   * group change into different scopes
+   * @param changes ui setting changes
+   * @returns [global, user]
```
SuZhou-Joe, line 314: The comments need to update accordingly.

src/plugins/data_source_management/public/components/edit_data_source/edit_data_source.tsx
```diff
@@ -146,7 +153,12 @@ export const EditDataSource: React.FunctionComponent<RouteComponentProps<{ id: s
   const setDefaultDataSource = async () => {
     try {
       if (getDefaultDataSourceId(uiSettings) === dataSourceID) {
```
SuZhou-Joe, line 155: We need to update `getDefaultDataSourceId` as well I think.

src/core/public/ui_settings/ui_settings_api.ts
Qxisylolo: Using any here is intentional, as the keys and values can vary widely.

src/core/public/ui_settings/ui_settings_api.ts
Qxisylolo: In previous changes, the ui_settings_api didn't support scope. It just batched setting changes and flushed changes. And use prev.callback(error, resp) to pass results.

in new implementation, we support scope.
When calling batchSet, we can pass settings with specific scopes (or no scope):
```
batchSet(key1, value1, scope1);
batchSet(key2, value2);  // Default scope (undefined)
batchSet(key3, value3, scope3);
batchSet(key4, value4, scope4);
```

In batchSet, settings are grouped based on their scope, when flushPendingChanges is triggered, we send multiple requests to the server, one for each scope group. After the requests are completed, the results are merged into a single response as they are completed in one flush.(This is similar to the previous implementation, but with the added step of categorizing the requests by scope)

src/core/public/ui_settings/ui_settings_api.ts
Qxisylolo: actually, we don't need to. When flushPendingChanges is triggered, multiple requests are sent in one as part of a single pendingChanges batch. And each request corresponds to a different scope. Whether or not a request is sent is controlled by the sendInProgress flag

changelogs/fragments/9726.yml
```diff
@@ -0,0 +1,2 @@
+feat:
+- Workspace scope ([#9726](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9726))
```
wanglam, line 2: Could you add more description about Workspace scope? I prefer to use something like `Add Workspace scope to ui setting` for the changelog.

src/core/public/ui_settings/ui_settings_api.ts
```diff
@@ -65,16 +68,17 @@ export class UiSettingsApi {
    * already in progress it will wait until the previous request is complete
    * before sending the next request
    */
-  public batchSet(key: string, value: any) {
+  public batchSet(key: string, value: any, scope?: UiSettingScope) {
     return new Promise<UiSettingsApiResponse | undefined>((resolve, reject) => {
       const prev = this.pendingChanges || NOOP_CHANGES;
-
+      const prevValues = { ...prev.values };
+      const scopedKey = scope ?? 'undefined';
```
wanglam, line 75: Seems this `undefined` string will be shared in many places, can we shared with a constant variable?

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
Qxisylolo: we can't remove the type cast here, because in this wrapper，we will handle 2 scenarios：
1. we request to get workspace level config.
2. we simply bypass and delegate to the underlying client, which returns SavedObject<T>

In the first scenario, the returned type is inherently SavedObject<Record<string, any>>, while in the second scenario, it must remain SavedObject<T> to align with the original saved object client's behavior

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
```diff
@@ -193,7 +194,7 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
           await fetchDataSources();
           setSelectedDataSources([]);
           if (payload.some((p) => p.id === defaultDataSourceId)) {
-            setFirstDataSourceAsDefault(savedObjects.client, uiSettings, true);
+            setFirstDataSourceAsDefault(savedObjects.client, uiSettings, true, !!currentWorkspace);
```
wanglam, line 197: Nit: we can pass `true` here directly.

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -272,25 +273,36 @@ export async function getDataSourcesWithFields(
...
 ) {
   if (exists) {
     uiSettings.remove(DEFAULT_DATA_SOURCE_UI_SETTINGS_ID);
```
wanglam, line 296: Do we need to pass the scope option when remove this setting?

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
wanglam: I think we still need to fill other required fields here. The saved object client wrapper should follow the interface of the `get` method. It should return a completed saved object. There will be some potential issues when returning an incomplete saved object. For example, the `references` is a required property in `SavedObject`, if our result doesn't have this property. It will throw an error when trying to execute something like `obj.references.includes`.

Qxisylolo: yes, I will fill other required fields here.

src/plugins/data_source_management/public/components/utils.ts
Qxisylolo: thanks for the comments, all updated

src/core/public/ui_settings/ui_settings_api.ts
Qxisylolo: thanks for the comments, update.

src/core/public/ui_settings/ui_settings_client.ts
Qxisylolo: thanks for the comments, updated

src/plugins/data_source_management/public/components/edit_data_source/edit_data_source.tsx
Qxisylolo: yes, updated

changelogs/fragments/9726.yml
Qxisylolo: updated, thank you

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
Qxisylolo: updated, thanks

src/core/public/ui_settings/types.ts
```diff
@@ -54,6 +54,11 @@ export interface IUiSettingsClient {
    */
   get: <T = any>(key: string, defaultOverride?: T) => T;
 
+  /**
+   * Gets the value for a specific uiSetting and scope and sent to the server
+   */
+  getUserProvided: <T = any>(key: string, scope: UiSettingScope) => Promise<T>;
```
SuZhou-Joe, line 60: Nit: `getUserProvidedWithScope`?

src/core/public/ui_settings/ui_settings_api.ts
SuZhou-Joe: Thanks, I misunderstood the intention here. Forget it.

src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -124,12 +129,20 @@ You can use \`IUiSettingsClient.get("${key}", defaultValue)\`, which will just r
     );
   }
 
-  async set(key: string, value: any) {
-    return await this.update(key, value);
+  async getUserProvided<T = any>(key: string, scope: UiSettingScope) {
+    return await this.api.getWithScope(scope).then((response: any) => {
```
SuZhou-Joe, line 133: Better to add a validation here to ensure the setting supports the scope passed in.

src/core/public/ui_settings/ui_settings_api.ts
```diff
@@ -65,16 +70,17 @@ export class UiSettingsApi {
...
    * before sending the next request
    */
-  public batchSet(key: string, value: any) {
+  public batchSet(key: string, value: any, scope?: UiSettingScope) {
     return new Promise<UiSettingsApiResponse | undefined>((resolve, reject) => {
       const prev = this.pendingChanges || NOOP_CHANGES;
-
+      const prevValues = { ...prev.values };
+      const scopedKey = scope ?? NO_SCOPE;
+      if (!prevValues[scopedKey]) {
```
SuZhou-Joe, line 78: Shall we do some validation here before adding the change into pendingChanges?

src/core/server/ui_settings/ui_settings_client.ts
```diff
@@ -98,6 +104,10 @@ export class UiSettingsClient implements IUiSettingsClient {
   private readonly overrides: NonNullable<UiSettingsServiceOptions['overrides']>;
   private readonly defaults: NonNullable<UiSettingsServiceOptions['defaults']>;
   private readonly log: Logger;
+  private readonly userLevelSettingsKeys: string[] = [];
+  private readonly workspaceLevelSettingsKeys: string[] = [];
+  private readonly globalLevelSettingsKeys: string[] = [];
+  private readonly uiSettingsKeysWithMoreThanOneScope: string[] = [];
```
SuZhou-Joe, line 110: Seems this property is useless.

src/plugins/data_source_management/public/components/data_source_selector/data_source_selector.tsx
```diff
@@ -169,6 +169,7 @@ export class DataSourceSelector extends React.Component<
       }
 
       const defaultDataSource = getDefaultDataSourceId(this.props.uiSettings) ?? null;
+
```
SuZhou-Joe, line 172: Nit: remove this line.

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -335,9 +338,13 @@ export function getFilteredDataSources(
...
 }
 
-export function getDefaultDataSourceId(uiSettings?: IUiSettingsClient) {
+export function getDefaultDataSourceId(uiSettings?: IUiSettingsClient, scope?: UiSettingScope) {
   if (!uiSettings) return null;
-  return uiSettings.get<string | null>(DEFAULT_DATA_SOURCE_UI_SETTINGS_ID, null);
+  // if specify the scope, then we will call getUserProvided to request from server
+  // otherwise, we will call get to get defaultDataSource stored in cache
+  if (scope)
+    return uiSettings.getUserProvided<string | null>(DEFAULT_DATA_SOURCE_UI_SETTINGS_ID, scope);
```
SuZhou-Joe, line 346: getUserProvided is a async call, we need to change all the places that are using this function.

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -61,10 +58,18 @@ export class WorkspaceUiSettingsClientWrapper {
...
+      if (type === 'config' && id.startsWith(CURRENT_WORKSPACE_PLACEHOLDER)) {
+        // if not in a workspace and try to get workspace level settings
+        // it should return NotFoundError
+        if (!requestWorkspaceId) {
+          throw SavedObjectsErrorHelpers.createGenericNotFoundError();
+        }
+
+        const normalizeDocId = id.replace(`${CURRENT_WORKSPACE_PLACEHOLDER}_`, '');
+
         const configObject = await wrapperOptions.client.get<Record<string, any>>(
```
SuZhou-Joe, line 70: Seems no need to get the global config here anymore.

Qxisylolo: we have to get global config to fill other required fields https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9726#discussion_r2062516330.

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
```diff
@@ -61,10 +58,18 @@ export class WorkspaceUiSettingsClientWrapper {
        * the global ui settings and workspace ui settings have higher priority if the same setting
        * was defined in both places
        */
-      if (type === 'config' && requestWorkspaceId) {
+      if (type === 'config' && id.startsWith(CURRENT_WORKSPACE_PLACEHOLDER)) {
+        // if not in a workspace and try to get workspace level settings
+        // it should return NotFoundError
+        if (!requestWorkspaceId) {
+          throw SavedObjectsErrorHelpers.createGenericNotFoundError();
```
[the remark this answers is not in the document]
Qxisylolo: createGenericNotFoundError will create 400 error. 
```
function decorateBadRequestError(error: Error, reason?: string) {
  return decorate(error, CODE_BAD_REQUEST, 400, reason);
}
```

src/core/public/ui_settings/ui_settings_api.ts
Qxisylolo: do you mean to add some validation for scope? if so, we already make validation in https://github.com/opensearch-project/OpenSearch-Dashboards/blob/412e11b0d97786c7b304fe1206040edb1402b848/src/core/server/ui_settings/ui_settings_client.ts#L225

src/core/public/ui_settings/ui_settings_api.ts
```diff
@@ -65,16 +70,17 @@ export class UiSettingsApi {
...
+        prevValues[scopedKey] = {};
+      }
+      prevValues[scopedKey][key] = value;
```
raintygao, line 81: Here looks a bit confused to me.

```
preValues['undefined'] = xxxx
```

The scope could be a string `'undefined'` , do we need to do this?

src/core/public/ui_settings/ui_settings_api.ts
```diff
@@ -32,14 +32,19 @@ import { BehaviorSubject } from 'rxjs';
...
 export interface UiSettingsApiResponse {
   settings: UiSettingsState;
 }
 
+const NO_SCOPE = 'undefined';
+
 interface Changes {
   values: {
-    [key: string]: any;
+    [scope: UiSettingScope | string]: {
```
raintygao, line 45: Since `UiSettingScope` has already used an enum to limit the scope, `[scope: UiSettingScope | string]:` would be enough and more safe?

src/core/public/ui_settings/ui_settings_api.ts
raintygao, line 81: This operation looks not to `prevValues`, it looks more like to `newValues`.

Qxisylolo: I think it's fine, since we are only using 'undefined' temporarily to identify settings without a scope passed in. Later, we will recognize them during flushChanges and treat them as setMany without a scope.

src/core/public/ui_settings/ui_settings_api.ts
Qxisylolo: we just use string to store settings without a scope passed in.

src/core/public/ui_settings/ui_settings_api.ts
Qxisylolo: thanks for pointing out the confusion, update~

src/plugins/data_source_management/public/components/utils.ts
Qxisylolo: thanks for remaindering, updated

src/core/public/ui_settings/ui_settings_api.ts
```diff
@@ -65,16 +70,18 @@ export class UiSettingsApi {
    * already in progress it will wait until the previous request is complete
    * before sending the next request
    */
-  public batchSet(key: string, value: any) {
+  public batchSet(key: string, value: any, scope?: UiSettingScope) {
     return new Promise<UiSettingsApiResponse | undefined>((resolve, reject) => {
       const prev = this.pendingChanges || NOOP_CHANGES;
-
+      const newValues = { ...prev.values };
+      // if there isn;t a scope passed in, we will use NO_SCOPE as a identifier temporarily
```
SuZhou-Joe, line 77:
```suggestion
      // if there isn't a scope passed in, we will use NO_SCOPE as a identifier temporarily
```

src/plugins/data_source_management/public/components/data_source_aggregated_view/data_source_aggregated_view.tsx
```diff
@@ -135,11 +135,11 @@ export class DataSourceAggregatedView extends React.Component<
           });
           return;
         }
-
         this.setState({
           ...this.state,
           allDataSourcesIdToTitleMap,
-          defaultDataSource: getDefaultDataSourceId(this.props.uiSettings) ?? null,
+          // for data source aggregated view, get default data source from cache
+          defaultDataSource: (await getDefaultDataSourceId(this.props.uiSettings)) ?? null,
```
SuZhou-Joe, line 142: Seems we should pass in scope.

src/plugins/data_source_management/public/components/data_source_multi_selectable/data_source_multi_selectable.tsx
```diff
@@ -77,7 +77,8 @@ export class DataSourceMultiSelectable extends React.Component<
   async componentDidMount() {
     this._isMounted = true;
     try {
-      const defaultDataSource = getDefaultDataSourceId(this.props.uiSettings) ?? null;
+      // for data source selectable, get default data source from cache
+      const defaultDataSource = (await getDefaultDataSourceId(this.props.uiSettings)) ?? null;
```
SuZhou-Joe, line 81: Shall we pass in scope for this call?

src/plugins/data_source_management/public/components/data_source_selectable/data_source_selectable.tsx
```diff
@@ -213,8 +213,8 @@ export class DataSourceSelectable extends React.Component<
         });
         return;
       }
-
-      const defaultDataSource = getDefaultDataSourceId(this.props.uiSettings) ?? null;
+      // for data source selectable, get default data source from cache
```
SuZhou-Joe, line 216: Could you please elaborate more why for this component we can get the data source from cache?

src/plugins/data_source_management/public/components/data_source_selector/data_source_selector.tsx
```diff
@@ -168,7 +168,7 @@ export class DataSourceSelector extends React.Component<
         return;
       }
 
-      const defaultDataSource = getDefaultDataSourceId(this.props.uiSettings) ?? null;
+      const defaultDataSource = (await getDefaultDataSourceId(this.props.uiSettings)) ?? null;
```
SuZhou-Joe, line 171: Is `scope` required here?

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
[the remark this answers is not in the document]
SuZhou-Joe: Yeah, I mean if we should apply 400 statusCode for `get` as well?

src/plugins/data_source_management/public/components/data_source_selectable/data_source_selectable.tsx
Qxisylolo: For these cases, we can retrieve the data from the cache, as the default data source is always set at this point. Each time the data is set with a scope, we will retrieve it with the same scope and refresh the cache. Therefore, in such cases, fetching from the cache should consistently return the correct result

Unlike other cases — for example, when dealing with the data source table or editing a data source — we always need to set the default data source manually or set first data source as default data source. Therefore, in most of those cases, we should pass the scope.

src/plugins/workspace/server/saved_objects/workspace_ui_settings_client_wrapper.ts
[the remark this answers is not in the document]
Qxisylolo: sorry, I looked at the wrong line. I think it's better to use a 'not found' error, since the UI client will handle it accordingly.
```
// ignore 404 and return an empty object
if (ignore404Errors && SavedObjectsErrorHelpers.isNotFoundError(error)) {
  return {};
}
```
Returning {} seems more appropriate when it's a 'not found'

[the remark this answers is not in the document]
SuZhou-Joe: I see, thanks.

src/plugins/data_source_management/public/components/data_source_selector/data_source_selector.tsx
Qxisylolo: data_source_selector, data_source_selectable, and data_source_aggregated_view share similar handling — we don’t need to pass a scope, we can directly retrieve the data from the cache, as the default data source is always set at this point.

src/core/public/ui_settings/ui_settings_api.ts
```diff
@@ -65,16 +70,18 @@ export class UiSettingsApi {
    * already in progress it will wait until the previous request is complete
    * before sending the next request
    */
-  public batchSet(key: string, value: any) {
+  public batchSet(key: string, value: any, scope?: UiSettingScope) {
```
kavilla, line 73: With some of the regressions related to putting conditional logic within existing stable code that hasnt changed but is shared across multiple plugins. Would it be safer to implement just new functions.

Like `batchSetWithScope`

src/core/public/ui_settings/ui_settings_client.test.ts
```diff
@@ -40,24 +40,33 @@ function setup(options: { defaults?: any; initialSettings?: any } = {}) {
     defaults = {
       dateFormat: { value: 'Browser' },
       aLongNumeral: { value: `${BigInt(Number.MAX_SAFE_INTEGER) + 11n}`, type: 'number' },
+      defaultDatasource: { value: 'Browser-ds', scope: UiSettingScope.WORKSPACE },
```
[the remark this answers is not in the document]
Qxisylolo: defaults represents a collection of UI settings, and defaultDatasource is simply the key for one of those settings

src/core/public/ui_settings/ui_settings_service.mock.ts
```diff
@@ -39,6 +39,7 @@ const createSetupContractMock = () => {
     getDefault: jest.fn(),
     get: jest.fn(),
     get$: jest.fn(),
+    getUserProvidedWithScope: jest.fn(),
```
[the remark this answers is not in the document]
Qxisylolo: in the previous implementation, none of the APIs(public side) supported scoped access — the get function simply retrieved data from the cache. (Each time we load a page, the UI settings are fetched from the server and stored in the cache.)  `getUserProvidedWithScope` now allows us to pass in a scope explicitly.

src/core/public/ui_settings/ui_settings_api.ts
```diff
@@ -130,16 +144,34 @@ export class UiSettingsApi {
 
     const changes = this.pendingChanges;
     this.pendingChanges = undefined;
-
+    const results: UiSettingsApiResponse = { settings: {} };
     try {
       this.sendInProgress = true;
+      const scopeEntries = Object.entries(changes.values);
```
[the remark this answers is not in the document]
Qxisylolo: Perhaps we can send the request immediately if a scope is passed. This way, we can ensure the atomicity

src/core/public/ui_settings/ui_settings_api.ts
```diff
@@ -90,6 +88,18 @@ export class UiSettingsApi {
     });
   }
 
+  public async getAll() {
+    return this.sendRequest('GET', '/api/opensearch-dashboards/settings', undefined);
+  }
+
+  public async getWithScope() {
```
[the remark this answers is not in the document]
Qxisylolo: getAll refers to get all settings and merge.

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -335,9 +338,21 @@ export function getFilteredDataSources(
     .sort((a, b) => a.label.toLowerCase().localeCompare(b.label.toLowerCase()));
 }
 
-export function getDefaultDataSourceId(uiSettings?: IUiSettingsClient) {
+export async function getDefaultDataSourceId(
```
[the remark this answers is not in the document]
Qxisylolo: we can keep getDefaultDataSourceId, as in the future, it will only get default data source within scope and we won't read it from cache

src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -248,11 +295,18 @@ You can use \`IUiSettingsClient.get("${key}", defaultValue)\`, which will just r
       ) {
         const { settings } = this.cache[key]?.preferBrowserSetting
           ? this.setBrowserStoredSettings(key, newVal)
-          : (await this.api.batchSet(key, newVal)) || {};
+          : (await this.selectedApi(scope).batchSet(key, newVal)) || {};
```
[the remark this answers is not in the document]
Qxisylolo: fixed, thanks

src/core/public/ui_settings/ui_settings_api.ts
```diff
@@ -90,6 +88,21 @@ export class UiSettingsApi {
...
```
SuZhou-Joe, line 95: No need for getWithScope I think if scope is already a property of the UISettingAPI instance.

src/core/public/ui_settings/ui_settings_api.ts
```diff
@@ -55,10 +56,9 @@ const NOOP_CHANGES = {
 export class UiSettingsApi {
   private pendingChanges?: Changes;
   private sendInProgress = false;
-
   private readonly loadingCount$ = new BehaviorSubject(0);
 
-  constructor(private readonly http: HttpSetup) {}
+  constructor(private readonly http: HttpSetup, private readonly scope?: UiSettingScope) {}
```
SuZhou-Joe, line 61:
```suggestion
  constructor(private readonly http: HttpSetup, private readonly scope: UiSettingScope = Scope.Global) {}
```
So that we do not add check if this.scope exist in other method.

src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -42,6 +47,9 @@ interface UiSettingsClientParams {
   defaults: Record<string, PublicUiSettingsParams>;
   initialSettings?: UiSettingsState;
   done$: Observable<unknown>;
+  apiForWorkspace: UiSettingsApi;
+  apiForUser: UiSettingsApi;
+  apiForGlobal: UiSettingsApi;
```
SuZhou-Joe, line 52:
```suggestion
  uiSettingApis: Record<string, UiSettingsApi>
```
Let's make it a map for extensibility, especially we may introduce scopeService in the future.

src/core/public/ui_settings/ui_settings_api.ts
Qxisylolo: We can't default the scope to global when it's not provided, because user-level settings also omit the scope.
In such cases, we should let the server side to decide which scope it belongs

src/core/public/ui_settings/ui_settings_service.ts
```diff
@@ -45,12 +46,35 @@ export interface UiSettingsServiceDeps {
...
+     * if the scope is not clarified, it will remains the previous logic, leave the handling to server to decide the destinanted scope
+     */
     this.uiSettingsApi = new UiSettingsApi(http);
-    http.addLoadingCountSource(this.uiSettingsApi.getLoadingCount$());
+    this.uiSettingsApiForWorkspace = new UiSettingsApi(http, UiSettingScope.WORKSPACE);
+    this.uiSettingsApiForUser = new UiSettingsApi(http, UiSettingScope.USER);
+    this.uiSettingsApiForGlobal = new UiSettingsApi(http, UiSettingScope.GLOBAL);
+
+    const combinedLoadingCount$ = combineLatest([
+      this.uiSettingsApi.getLoadingCount$(),
```
SuZhou-Joe, line 67: this.uiSettingsApi = this.uiSettingsApiForGlobal I think, do we need to listen for both?

Qxisylolo: yes, We need to create four API instances because:

uiSettingsApiForGlobal is specifically used to set the default data source outside of a workspace.
uiSettingsApi is used for other settings where no scope is explicitly passed — it follows the default behavior and relies on the server to determine the target scope.

Qxisylolo: yes, We need to create four API instances because:

uiSettingsApiForGlobal is specifically used to set the default data source outside of a workspace.
uiSettingsApi is used for other settings where no scope is explicitly passed — it follows the default behavior and relies on the server to determine the target scope.

src/core/public/ui_settings/ui_settings_api.ts
Qxisylolo: We can't default the scope to global when it's not explicitly provided, because user-level settings also omit the scope.
In such cases, we should delegate the decision to the server, allowing it to determine the appropriate scope based on the context.

src/core/public/ui_settings/ui_settings_client.ts
Qxisylolo: updated, please have a look. thanks

src/core/public/ui_settings/ui_settings_api.ts
SuZhou-Joe: I see, thanks.

src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -248,11 +323,12 @@ You can use \`IUiSettingsClient.get("${key}", defaultValue)\`, which will just r
       ) {
         const { settings } = this.cache[key]?.preferBrowserSetting
           ? this.setBrowserStoredSettings(key, newVal)
-          : (await this.api.batchSet(key, newVal)) || {};
-        this.cache = defaultsDeep({}, defaults, this.getBrowserStoredSettings(), settings);
+          : (await this.selectedApi(scope).batchSet(key, newVal)) || {};
+
+        this.mergeSettingsIntoCache(key, defaults, true, settings, scope);
```
SuZhou-Joe, line 328: where is `this.getBrowserStoredSettings()`?

src/plugins/data_source_management/public/components/edit_data_source/edit_data_source.tsx
```diff
@@ -72,6 +77,23 @@ export const EditDataSource: React.FunctionComponent<RouteComponentProps<{ id: s
     })();
   });
 
+  const loadDefaultDataSourceId = useCallback(async () => {
+    try {
+      setIsLoading(true);
```
wanglam, line 82: I prefer to use a separate loading state for default data source ID loading. The fetchDataSourceDetailsByID will set isLoading state too. The loading states will conflict if one resolves but another one is still fetching.

src/plugins/workspace/server/saved_objects/integration_tests/workspace_ui_settings_wrapper.test.ts
```diff
@@ -56,38 +57,59 @@ describe('workspace ui settings saved object client wrapper', () => {
   }, 30000);
 
   beforeEach(async () => {
-    await globalUiSettingsClient.set('defaultIndex', 'global-index');
```
wanglam, line 59: Nit: do we need to add some integration tests for the defaultIndex ui setting?

src/core/server/ui_settings/ui_settings_client.test.ts
```diff
@@ -673,6 +809,88 @@ describe('ui settings', () => {
...
+        },
+      };
+
+      const { uiSettings, savedObjectsClient } = setup({ defaults });
+
+      savedObjectsClient.get.mockImplementation((_type, id, _options) => {
+        if (id === `${CURRENT_WORKSPACE_PLACEHOLDER}_${ID}`) {
+          return Promise.resolve({
+            attributes: {
+              bar: 'my personal value',
```
wanglam, line 831: Nit: prefer use workspace-value here.

src/plugins/data_source_management/public/components/edit_data_source/edit_data_source.tsx
Qxisylolo: updated, please have a look, thanks

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -335,9 +338,21 @@ export function getFilteredDataSources(
...
+  // if specify the scope, then we will call getUserProvidedWithScope to request from server
+  // otherwise, we will get defaultDataSource stored in cache
+  if (scope) {
+    const result = await uiSettings.getUserProvidedWithScope<string | null>(
+      DEFAULT_DATA_SOURCE_UI_SETTINGS_ID,
+      scope
+    );
+    return result;
+  }
+  return uiSettings.get<string | null>(DEFAULT_DATA_SOURCE_UI_SETTINGS_ID);
```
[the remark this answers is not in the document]
Qxisylolo: sorry, i may delete it by mistake. updated

src/plugins/workspace/server/saved_objects/integration_tests/workspace_ui_settings_wrapper.test.ts
Qxisylolo: I think it is fine, as defaultDataSource tests has cover the part we did for defaultIndex

wanglam (approved):
LGTM~ Thanks for this refactoring.

SuZhou-Joe (approved):

PR #9734, opened 2025-04-28
Fix collaborators displays under custom tab on left navigation when saved object permission is disabled

FriedhelmWS:
### Description

Before:
```
savedObjects.permission.enabled: false
```
Screenshot 2025-04-28 at 10 26 38

After:
Screenshot 2025-04-28 at 10 29 23

## Changelog
- fix: Fix collaborators displays under custom tab on navigation when saved object permission is disabled

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/public/plugin.ts
+++ b/src/plugins/workspace/public/plugin.ts
@@ -570,6 +570,7 @@ export class WorkspacePlugin
     this.collaboratorsAppUpdater$.next(() => {
       return {
         status: isPermissionEnabled ? AppStatus.accessible : AppStatus.inaccessible,
+        chromeless: !isPermissionEnabled,
         navLinkStatus: core.chrome.navGroup.getNavGroupEnabled()
           ? AppNavLinkStatus.visible
           : AppNavLinkStatus.hidden,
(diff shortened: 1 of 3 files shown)


src/plugins/workspace/public/plugin.ts
```diff
@@ -570,6 +570,7 @@ export class WorkspacePlugin
     this.collaboratorsAppUpdater$.next(() => {
       return {
         status: isPermissionEnabled ? AppStatus.accessible : AppStatus.inaccessible,
```
wanglam, line 572: The application status will be "inaccessible" when permission is disabled. I think the inaccessible application should not be displayed in the left navigation. Could you deep dive into the implementation details to figure out why status update does not work here?

yubonluo: By default, navLinkStatus updates its value based on the status (https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/application/utils/get_app_info.ts#L35), but assigning a value to navLinkStatus again during the plugin's start phase causes the original value to be overwritten. I think we just comment this line is enough.
```
 navLinkStatus: core.chrome.navGroup.getNavGroupEnabled()
          ? AppNavLinkStatus.visible
          : AppNavLinkStatus.hidden,
```

yubonluo (approved):
LGTM

ruanyl (approved):

SuZhou-Joe (approved):

PR #9735, opened 2025-04-28
Fix chatbot flyout cannot be resized beyond the window size

FriedhelmWS:
### Description

Previously, the implementation of the resize button allowed users to drag it and resize the chatbot window outside of the browser window. This meant that if users dragged the button outside of the browser window and released the mouse press, they could never resize the window back to its original size since the button was now outside of the window and could not be reached.

This PR consists of the following changes:

1. When resizing the chatbot window, prevent users from resizing the chatbot window beyond the browser window size
2. When resizing the window, if the chatbot window exceeds the window size, automatically shrink the chatbot window accordingly

## Screenshot

Before:

https://github.com/user-attachments/assets/423fb273-7541-4835-94f1-a80de25e5d7b

After:

https://github.com/user-attachments/assets/2b77d6dd-ab2d-4eb8-8c5f-8e1ae0e117f0

## Changelog
- fix: Chatbot flyout cannot be resized beyond the window size

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/public/overlays/sidecar/components/resizable_button.test.tsx
+++ b/src/core/public/overlays/sidecar/components/resizable_button.test.tsx
@@ -6,8 +6,15 @@
 import React from 'react';
 
 import { ResizableButton, MIN_SIDECAR_SIZE } from './resizable_button';
-import { shallow } from 'enzyme';
+import { mount, shallow } from 'enzyme';
 import { SIDECAR_DOCKED_MODE } from '../sidecar_service';
+import { act } from 'react-dom/test-utils';
+
+const originalAddEventListener = window.addEventListener;
+
+const restoreWindowEvents = () => {
+  window.addEventListener = originalAddEventListener;
+};
 
 const storeWindowEvents = () => {
   const map: Record<string, Function> = {};
@@ -47,6 +54,17 @@ test('it should be vertical when docked mode is takeover', () => {
 test('it should emit onResize with new flyout size when docked right and drag and horizontal', () => {
   const windowEvents = storeWindowEvents();
 
+  Object.defineProperty(window, 'innerHeight', {
+    writable: true,
+    configurable: true,
+    value: 2000,
+  });
+  Object.defineProperty(window, 'innerWidth', {
+    writable: true,
+    configurable: true,
+    value: 2000,
+  });
+
   const onResize = jest.fn();
   const newProps = { ...props, onResize };
   const component = shallow(<ResizableButton {...newProps} />);
@@ -62,6 +80,17 @@ test('it should emit onResize with new flyout size when docked right and drag an
 
 test('it should emit onResize with new flyout size when docked left and drag and horizontal', () => {
   const windowEvents = storeWindowEvents();
+
+  Object.defineProperty(window, 'innerHeight', {
+    writable: true,
+    configurable: true,
+    value: 2000,
+  });
+  Object.defineProperty(window, 'innerWidth', {
+    writable: true,
+    configurable: true,
+    value: 2000,
+  });
 
   const onResize = jest.fn();
   const newProps = { ...props, onResize, dockedMode: SIDECAR_DOCKED_MODE.LEFT };
@@ -79,6 +108,17 @@ test('it should emit onResize with new flyout size when docked left and drag and
 test('it should emit onResize with new flyout size when drag and vertical', () => {
   const windowEvents = storeWindowEvents();
 
+  Object.defineProperty(window, 'innerHeight', {
+    writable: true,
+    configurable: true,
+    value: 2000,
+  });
+  Object.defineProperty(window, 'innerWidth', {
+    writable: true,
+    configurable: true,
+    value: 2000,
+  });
+
   const onResize = jest.fn();
   const newProps = { ...props, onResize, dockedMode: SIDECAR_DOCKED_MODE.TAKEOVER };
   const component = shallow(<ResizableButton {...newProps} />);
@@ -94,6 +134,7 @@ test('it should emit onResize with new flyout size when drag and vertical', () =
 
 test('it should emit onResize with min size when drag if new size is below the minimum', () => {
   const windowEvents = storeWindowEvents();
+
   const onResize = jest.fn();
   const newProps = { ...props, onResize };
   const component = shallow(<ResizableButton {...newProps} />);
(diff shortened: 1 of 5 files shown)


src/core/public/overlays/sidecar/components/resizable_button.tsx
```diff
@@ -29,6 +29,30 @@ export const ResizableButton = ({ dockedMode, onResize, flyoutSize }: Props) =>
   const initialFlyoutSize = useRef(flyoutSize);
   const setFocus = (e: React.MouseEvent<HTMLButtonElement>) => e.currentTarget.focus();
 
+  useEffect(() => {
+    const handleWindowResize = () => {
+      if (flyoutSize > MIN_SIDECAR_SIZE) {
+        // Make sure flyout never below min size even if the window goes below the size
+        if (dockedMode === SIDECAR_DOCKED_MODE.TAKEOVER && flyoutSize > window.innerHeight) {
```
wanglam, line 36: I'm prefer to separate these size calculation logics to a function. Then it can be shared in the `setSidecarConfig` function.

src/core/public/overlays/sidecar/components/resizable_button.tsx
```diff
@@ -29,6 +29,30 @@ export const ResizableButton = ({ dockedMode, onResize, flyoutSize }: Props) =>
...
+          (dockedMode === SIDECAR_DOCKED_MODE.LEFT || dockedMode === SIDECAR_DOCKED_MODE.RIGHT) &&
+          flyoutSize > window.innerWidth
+        ) {
+          // Automatically reduce the width in left or right docked mode when resize the window
+          onResize(window.innerWidth);
+        }
+      }
+    };
+
+    window.addEventListener('resize', handleWindowResize);
```
wanglam, line 49: Can we optimize this listener registration logic? It will be call many times after `flyoutSize` updated.

wanglam:
@FriedhelmWS Thanks for fixing this resizing issue. I still have one open question. Do we need to add some validation logic on the `setSidecarConfig` function?

src/core/public/overlays/sidecar/components/resizable_button.tsx
raintygao: I think there has already been a similar limitation when calling `setSidecarConfig`

src/core/public/overlays/sidecar/components/resizable_button.tsx
FriedhelmWS: cool, I will add debouncing here

FriedhelmWS:
> @FriedhelmWS Thanks for fixing this resizing issue. I still have one open question. Do we need to add some validation logic on the `setSidecarConfig` function?

Done, add the check in `setSidecarConfig`

Demo:

https://github.com/user-attachments/assets/c16007c9-4328-4c37-ada8-c86f134056eb

src/core/public/overlays/sidecar/sidecar_service.tsx
```diff
@@ -133,7 +134,12 @@ export class SidecarService {
         // init
         (!sidecarConfig$.value && config.dockedMode && config.paddingSize)
       ) {
-        sidecarConfig$.next({ ...sidecarConfig$.value, ...config } as ISidecarConfig);
+        const paddingSize = calculateNewPaddingSize(config.dockedMode, config.paddingSize);
```
wanglam, line 137: The config object is a partial `ISidecarConfig`, do we need to check if `config.paddingSize` be undefined here? I think we don't update padding size to MIN_SIZE for this case.

src/core/public/overlays/sidecar/helper.ts
```diff
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { MIN_SIDECAR_SIZE } from './components/resizable_button';
```
wanglam, line 6: This is a circual import references here, can we avoid this?

src/core/public/overlays/sidecar/components/resizable_button.test.tsx
```diff
@@ -106,3 +147,110 @@ test('it should emit onResize with min size when drag if new size is below the m
...
+  component = mount(<ResizableButton {...rightProps} />);
+
+  await act(async () => {
+    Object.defineProperty(window, 'innerWidth', { value: 800 });
+    window.dispatchEvent(new Event('resize'));
+
+    await new Promise((resolve) => setTimeout(resolve, 50)); // wait for debounce
+  });
+
+  expect(onResize).toHaveBeenCalledWith(800);
```
wanglam, line 253: Can we use `await waitFor` wrap this line and remove the debounce wait in line 231 and 250?

wanglam (approved):

src/core/public/overlays/sidecar/components/resizable_button.tsx
```diff
@@ -29,6 +32,33 @@ export const ResizableButton = ({ dockedMode, onResize, flyoutSize }: Props) =>
...
+          onResize(newPaddingSize);
+        }
+      }
+    }, RESIZE_DEBOUNCE_DELAY);
+
+    const handleWindowResize = () => {
+      debouncedResize();
+    };
+
+    window.addEventListener('resize', handleWindowResize);
```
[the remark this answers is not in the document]
SuZhou-Joe: +1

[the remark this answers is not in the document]
FriedhelmWS: Done

src/core/public/overlays/sidecar/helper.ts
FriedhelmWS: Done, moved the `MIN_SIDECAR_SIZE` into `helper.ts` reasonably

src/core/public/overlays/sidecar/components/resizable_button.test.tsx
FriedhelmWS: Done

src/core/public/overlays/sidecar/helper.ts
```diff
@@ -44,3 +46,27 @@ export const getSidecarLeftNavStyle = (config: ISidecarConfig | undefined) => {
   }
   return {};
 };
+
+export const calculateNewPaddingSize = (
+  dockedMode: SIDECAR_DOCKED_MODE | undefined,
+  currentSize: number | undefined
```
wanglam, line 52: Nit: Based current implementation, the `currentSize` won't be `undefined` anymore. Can we remove the `| undefined` and the `currentSize` empty check below.

wanglam (approved):
Looks good to me, thanks for fixing this issue~

src/core/public/overlays/sidecar/sidecar_service.tsx
```diff
@@ -133,7 +134,13 @@ export class SidecarService {
         // init
         (!sidecarConfig$.value && config.dockedMode && config.paddingSize)
       ) {
-        sidecarConfig$.next({ ...sidecarConfig$.value, ...config } as ISidecarConfig);
+        sidecarConfig$.next({
+          ...sidecarConfig$.value,
+          ...config,
+          ...(config.paddingSize
```
SuZhou-Joe, line 140: Nit: I would prefer to do nothing or throw error if the passed in config is invalid. cc @raintygao

wanglam: Do you mean we should move this padding size validation to the plugin side? We may need to write duplicated logic in different plugins if it is used in many plugins. So I think we'd better add this logic in the sidecar service.

raintygao: Im okay to both ways, but actually I think we could do params fallback in calling side, for this common service, we may not need too much extra handling logic.

SuZhou-Joe: I was thinking we just use the value and it is component's duty to decide how to render, but still sounds tricky. I am OK with the solution now. Forget it.

PR #9767, opened 2025-05-03
Add a CI job to prevent introducing new TypeScripts errors

Maosaic:
### Description

For a long time, OSD did not enforce any TypeScript errors check and we have a long list of existing TypeScript errors. The goal here is to prevent introducing any new TypeScript errors so we can focus on fixing existing ones. Currently, I added this check script to CI test job, but moving forward I hope we can add this to build script.

This PR adds two scripts:
1. A script that generates/updates the baseline error list(`ts_error_baseline.json`)
2. A script that checks TypeScripts error against baseline error list, which will fail when identify new TypeScripts error.

How to use:
1. Run `yarn typecheck` anytime you want to check TypeScripts error.
2. Run `yarn update-ts-baseline` **ONLY** if you fixed existing TypeScripts errors, commit the updated `ts_error_baseline.json` file with your code change.
3. Raise your PR normally, CI will automatically run `typecheck` script.
4. TODO: Add a CI workflow to automatically update baseline error file when existing error got fixed.

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/1660

## Screenshot

Check succeeded 
Screenshot 2025-05-02 at 3 06 55 PM

Check failed
Screenshot 2025-05-02 at 3 07 15 PM

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ts_error_baseline.json
wanglam, line 1: Nit: There are some type errors from node_modules, do we need to exclude these type errors?

scripts/ts_error_checker.js
```diff
@@ -0,0 +1,199 @@
...
+      return [];
+    }
+  } catch (error) {
+    console.error(`${colors.red}✖ ${colors.reset}Error loading baseline: ${error.message}`);
+    process.exit(1);
+  }
+}
+
+// Helper to run TypeScript compiler and get structured output
+function getCurrentErrors() {
```
wanglam, line 62: Can we separate a unified type check logic to share between `ts_error_checker.js` and `update_ts_baseline.js`?

scripts/ts_error_checker.js
```diff
@@ -0,0 +1,199 @@
...
+          code: match[4],
+          message: match[5],
+        });
+      }
+    });
+
+    console.log(
+      `${colors.blue}ℹ ${colors.reset}Found ${colors.bold}${errors.length}${colors.reset} TypeScript errors in current codebase`
+    );
+    return errors;
```
wanglam, line 110: Do we need to sort this error array before returning the result? Will the output order be stable between different machines?

scripts/ts_error_checker.js
```diff
@@ -0,0 +1,199 @@
...
+
+  console.log(`${colors.cyan}▶ ${colors.reset}Executing: ${TSC_BIN} ${args.join(' ')}\n`);
+
+  const result = spawnSync(TSC_BIN, args, {
+    encoding: 'utf8',
+    shell: true,
+    stdio: 'pipe',
+  });
+
+  // Process doesn't throw an exception when there are type errors
```
raintygao, line 81: Thanks for the change. Im thinking whether it would be more practical if we could also add a check in the pre-commit stage, then we only run `tsc` for changed files.

scripts/ts_error_checker.js
```diff
@@ -0,0 +1,199 @@
...
+
+// Check if an error is new or already in the baseline
+function isNewError(error, baselineErrors) {
+  // An error is considered the same if it has the same file, code, line, and column
+  // If you touch it, you should fix it!
+  return !baselineErrors.some(
+    (baselineError) =>
+      baselineError.file === error.file &&
+      baselineError.code === error.code &&
+      baselineError.line === error.line &&
```
wanglam, line 122: I'm not sure if we need to compare the line and column here. The error will be treated as a new error if we just remove/add some lines in the source file. We may need to address all the type errors even if I didn't touch the type error logic.

angle943: is that a problem @wanglam ? Ideally we should be removing all typescript errors, and should have mechanisms that can progress to that goal. I am not sure we want to spend time/effort to ensure that existing typescript errors can live in the codebase without any alerts.

The purpose of this PR is to prevent new typescript errors, without hammering us down with the existing ones.

wanglam: I agree with you. We need to fix all the TypeScript errors step by step. However, since we are comparing line & column numbers here, developers might get confused when new errors occur. The new errors might not be related to their code changes. It would take significant effort to address this if there are many existing type errors that need to be fixed across multiple files. I'm open to comparing line & column numbers here, but it would be better if we have some separate PRs to fix these TypeScript errors in the future. I'm happy to join this effort and help improve the TypeScript type coverage.

Maosaic: They always run `update-ts-baseline` to generate a new baseline file and bypass testing

ashwin-pc (approved):
Lets start with this. Id rather us have some type checking than none. We can always improve how we check the code later.

scripts/ts_error_checker.js
virajsanghvi: Think it'll be important to communicate that if people are intentionally fixing existing TS errors, to run update-ts-baseline to prevent those fixes from being lost.

Thanks for making this change!

PR #9731, opened 2025-04-25
Add resource API pattern

kathyguo18:
### Description
feat. Add resource API pattern in query_enhancements

## Testing the changes
yarn test

## Changelog
- Add resource API pattern in query_enhancements

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #9770, opened 2025-05-05
Add resource API pattern

kathyguo18:
### Description
feat. Add resource API pattern in query_enhancements

for details see https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9770#issuecomment-2863807507

### Issues Resolved
[RFC] Add A Resource API in Query Enhancements [#9779](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9779)

## Testing the changes
yarn test

## Changelog
- feat: Add resource API pattern in query_enhancements

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/query_enhancements/server/routes/resources/routes.ts
```diff
@@ -0,0 +1,49 @@
...
+      },
+    },
+    async (context, request, response) => {
+      const { type } = request.body.connection;
+      const manager = resourceManagerService.getManager(type);
+      if (!manager) {
+        return response.notFound();
+      }
+      try {
+        const resourcesResponse = await manager.handleRequest(context, request);
```
joshuali925, line 37: i would prefer this to be more explicit, rather than handling all CRUD operations in a single `handleRequest` function. so similar question as https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9731#discussion_r2072047823: consider for example, when we add prometheus delete alerts API, should it be routing to `manager.handleRequest`? Prometheus APIs are restful, where API paths are resources, how does handleRequest know how to distinguish the operation?

joshuali925 (commented):
you had an older PR https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9731. if one is outdated, could you close it?

src/plugins/query_enhancements/server/routes/resources/routes.ts
kathyguo18: Thanks for the comment! Updated to handlePostRequest.

joshuali925 (approved):

ashwin-pc:
Thanks for the RFC @kathyguo18. Can you mark this feature as experimental? we might be moving the query API to core and in which case we might not want to keep this here as well. you can see other examples of this in the code.

src/plugins/query_enhancements/server/connections/resource_manager_service.ts
```diff
@@ -0,0 +1,29 @@
...
+        `Manager for dataConnectionType ${dataConnectionType} is already registered. Unable to register another manager.`
+      );
+    }
+    this.managers[dataConnectionType] = manager;
+  }
+
+  getManager(dataConnectionType: string): BaseConnectionManager | undefined {
+    return this.managers[dataConnectionType];
+  }
+}
```
angle943, line 26: is there a use case where we want to delete a manager? If so should we expose that as a method?

src/plugins/query_enhancements/server/connections/resource_manager_service.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { BaseConnectionManager } from './managers/base_connection_manager';
+
+class ResourceManagerService {
+  private readonly managers: { [key: string]: BaseConnectionManager };
+  constructor() {
+    this.managers = {};
```
angle943, line 11: super nitpick: i think a Map would be better suited here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

src/plugins/query_enhancements/server/routes/resources/routes.ts
```diff
@@ -0,0 +1,49 @@
...
+import { schema } from '@osd/config-schema';
+import { IRouter } from 'opensearch-dashboards/server';
+import { BASE_API } from '../../../common';
+import { resourceManagerService } from '../../connections/resource_manager_service';
+import { coerceStatusCode } from '..';
+
+export function registerResourceRoutes(router: IRouter) {
+  router.post(
+    {
+      path: `${BASE_API}/resources`,
```
angle943, line 15: nitpick: do we always use hardcoded string for the route? IE the `resources` part.

angle943 (changes requested):
Looks good, i think just the addition of `@experimental` flag is needed as @ashwin-pc suggested and i'll approve

src/plugins/query_enhancements/server/connections/resource_manager_service.ts
joshuali925: i don't think that would be needed, registering manager happens one time during plugin setup, i don't see a reasonable trigger on when would a plugin want to delete a manager

src/plugins/query_enhancements/server/routes/resources/routes.ts
kathyguo18: Yes. This new API is make to be a generic API to retrieve resources. So it's always the hardcoded string "resources". Besides POST, GET/DELETE/PUT may be added in the future.

joshuali925: I personally prefer to define APIs like
```ts
export const API = {
  RESOURCES: '/api/query_enhancements/v1/resources',
  RESOURCE: '/api/query_enhancements/v1/resources',
} as const;
```

this makes it easy to grep when debugging, and I remember joined string literals is no longer a string literal due to our old typescript version, it causes lsp hover to show `string` when hovering on `API.RESOURCES` rather than the actual string literal

But I don't think there's a documented standard, i see all different things in OSD

src/plugins/query_enhancements/server/connections/resource_manager_service.ts
kathyguo18: Thanks for the comment! Updated to use Map.

kathyguo18:
> Looks good, i think just the addition of `@experimental` flag is needed as @ashwin-pc suggested and i'll approve

Thanks for reviewing! Added `@experimental` flag in registerResourceRoutes function, BaseConnectionManager class, and ResourceManagerService class.

joshuali925 (dismissed):

angle943 (dismissed):

src/plugins/query_enhancements/server/connections/resource_manager_service.ts
```diff
@@ -0,0 +1,32 @@
...
+    this.managers.set(dataConnectionType, manager);
+  }
+
+  getManager(dataConnectionType: string): BaseConnectionManager | undefined {
+    return this.managers.get(dataConnectionType);
+  }
+}
+
+// export as singleton
+export const resourceManagerService = new ResourceManagerService();
```
kavilla, line 32: nit: the way core does this is via getters and setters. https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/services.ts#L37 

just a heads up placing this in query enhancements could be a little risky. query enhancements is optionally enabled. technically we can disable this plugin and now the resource manager service will no longer exist. 

typically we try to rely on key plugins with registration points that if disabled OSD does not function at all.

perhaps that could be a follow up. more so in protection of consumer plugins not relying on an optional plugin

kavilla (dismissed):

angle943:
@kavilla do you know why this is failing? https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/15150059713/job/42594604840?pr=9770

joshuali925 (dismissed):

kavilla (approved):

joshuali925 (approved):

ZilongX:
is there a target release tag for this change ?

PR #9724, opened 2025-04-24
[explore] add plugin

kavilla:
### Description

This PR introduces the Explore plugin (experimental), a new data exploration experience in OpenSearch Dashboards. The Explore plugin builds upon the foundation of Discover while focusing on a query-driven approach with enhanced multi-dataset support. It exists alongside the current Discover plugin to allow for incremental adoption without disrupting existing workflows.

### Issues Resolved

n/a

## Screenshot

n/a

## Testing the changes
By default is not enabled
```
yarn start:explore
```

## Changelog
- feat: Add new Explore plugin

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #9773, opened 2025-05-06
[Explore] support explore feature flag and workspace types

joshuali925:
### Description

This PR continues #9724 and makes the following changes

- disable explore if `explore.enabled` is not `true`
- register nav groups to observability workspace
- redirect to old discover if not in an observability workspace

Currently we only want this explore UI to be available in observability workspace, under the feature flag.

## Changelog
- feat: support explore only in the observability workspace type

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


angle943 (dismissed):

src/plugins/explore/opensearch_dashboards.json
```diff
@@ -17,6 +17,7 @@
     "visualizations",
     "usageCollection"
   ],
+  "requiredBundles": ["workspace"],
```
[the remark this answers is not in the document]
joshuali925: this PR uses `isNavGroupInFeatureConfigs` from workspace https://github.com/opensearch-project/OpenSearch-Dashboards/blob/07da1178ec5a01f0c479e3c579c83a3304a92df8/src/plugins/explore/public/plugin.ts#L52-L56

Maosaic (approved):

src/plugins/explore/public/plugin.ts
```diff
@@ -30,12 +42,22 @@ export class ExplorePlugin implements Plugin<ExplorePluginSetup, ExplorePluginSt
...
-        // Get start services as specified in opensearch_dashboards.json
         const [coreStart, depsStart] = await core.getStartServices();
-        // Render the application
+        const features = await core.workspaces.currentWorkspace$
+          .pipe(take(1))
+          .toPromise()
+          .then((workspace) => workspace?.features);
+        if (
+          !features ||
+          (!isNavGroupInFeatureConfigs(DEFAULT_NAV_GROUPS.observability.id, features) &&
```
ruanyl, line 54: Would you leave a comment here to describe the intention here? Why it should navigate to discover app?

kavilla: +1 I think should it just not display on the workspaces. The current discover plugin should already lead to the classic discover experience

src/plugins/workspace/public/index.ts
```diff
@@ -11,3 +11,4 @@ export function plugin() {
 
 export { WorkspacePluginSetup, WorkspaceCollaborator } from './types';
 export { WorkspaceCollaboratorType } from './services';
+export { isNavGroupInFeatureConfigs } from './utils';
```
kavilla, line 14: I’m not sure if the required bundling will guarantee that the workspace plugin is always loaded before the explore plugin. Even if it does, directly importing functions this way can create a tight coupling between the two plugins, which can become a maintenance challenge.

It might be cleaner to expose this utility through the start lifecycle method of the workspace plugin, ensuring a more explicit API contract. This approach aligns better with the intended OpenSearch Dashboards plugin architecture, where plugins communicate through well-defined interfaces.

Additionally, tightly coupling these plugins means that future changes to the workspace plugin would require testing against the explore plugin, potentially complicating development.

kavilla (changes requested):
To keep the architecture clean, it’s better to avoid directly importing utils from the workspaces plugin. While the explore plugin depends heavily on workspaces, the reverse shouldn’t be true to prevent unnecessary coupling.

src/plugins/workspace/public/index.ts
wanglam: What do you think about moving `isNavGroupInFeatureConfigs` to `src/core`? I think we can move this function to `src/plugins/workspace/public/utils.ts`. Then the workspace plugin and explore plugin can use the same function from core module.

src/plugins/explore/public/plugin.ts
```diff
@@ -3,25 +3,37 @@
...
 
 export class ExplorePlugin implements Plugin<ExplorePluginSetup, ExplorePluginStart> {
+  private config: ConfigSchema;
+
+  constructor(private readonly initializerContext: PluginInitializerContext) {
+    this.config = initializerContext.config.get<ConfigSchema>();
+  }
+
   public setup(core: CoreSetup, plugins: ExploreSetupPlugins): ExplorePluginSetup {
+    if (!this.config.enabled) return {};
```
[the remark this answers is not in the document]
joshuali925: good point, i forgot about that, will remove

src/plugins/explore/public/plugin.ts
```diff
@@ -48,14 +70,6 @@ export class ExplorePlugin implements Plugin<ExplorePluginSetup, ExplorePluginSt
       },
     ]);
 
-    core.chrome.navGroup.addNavLinksToGroup(DEFAULT_NAV_GROUPS.observability, [
```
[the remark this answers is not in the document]
joshuali925: this removes duplicated code. observability is still there
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/07da1178ec5a01f0c479e3c579c83a3304a92df8/src/plugins/explore/public/plugin.ts#L65-L71

src/plugins/explore/common/types/config.ts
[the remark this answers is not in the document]
joshuali925: according to plugin architecture, it didn't show custom files in the top level directory
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/1b10e502ed2bbdbcea7f8c5a17efc76e4d83d92b/src/core/CONVENTIONS.md#L33-L65

for newer plugins i thought we are doing that, like https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/query_enhancements/common/config.ts

If not should it be documented to be in `plugins/xxx/config.ts`?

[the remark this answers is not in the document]
kavilla: That's what I thought for the common so makes sense could we put it not in types tho?

src/plugins/explore/public/plugin.ts
kavilla: Did we decide that is something we care about ? Like what if I try to update an index pattern id that doesn't exist in my current workspace. But try to update the URL directly. Imo we shouldn't automatically switch to a workspace that has that index pattern.  Same for this. 

This also will require us to implement new code to validate that saved explores can't be loaded into a non observability or security analytics workspace type. Otherwise they will be able to load it into their dashboard an explore saved object but not be able to modify it.

src/plugins/workspace/public/index.ts
wanglam: There are some workspace utils functions in `src/core/utils/workspace.ts`, and we already introduced `features` as a property of workspace object in `src/core/types/workspace.ts`. The nav group is a concept of core too. So I think the `isNavGroupInFeatureConfigs` can be moved to the core module.

joshuali925: ah i didn't notice it's already part of core. refactored it in d918051 to move to core and remove dependency of `workspace` for `explore`. thanks

src/plugins/explore/public/plugin.ts
joshuali925: > Imo we shouldn't automatically switch to a workspace that has that index pattern

So this won't switch the workspace, it only redirects to the existing discover.

I don't think we need to handle when users modify the URL incorrectly. This logic mostly prevents for example when a plugin adapted to `explore`, when user clicks the button, they get redirected to the `app/explore` URL, but they could be in a different workspace and we don't want them to see 404

I remember having the offline discussion on whether we should just show 404 but can't remember the outcome. @angle943 do you remember?

> This also will require us to implement new code to validate that saved explores can't be loaded into a non observability or security analytics workspace type

Users can only create saved explores in the explore app, and the objects will be isolated by workspace. If user is in a non observability or security analytics workspace type, they won't have any saved explores.

angle943 (dismissed):

src/plugins/explore/public/plugin.ts
```diff
@@ -30,12 +32,24 @@ export class ExplorePlugin implements Plugin<ExplorePluginSetup, ExplorePluginSt
...
+          .toPromise()
+          .then((workspace) => workspace?.features);
+        // We want to limit explore UI to only show up under the observability
+        // workspace. If user lands in the explore plugin URL in a different
+        // workspace, we will redirect them to classic discover.
+        if (
+          !features ||
+          !isNavGroupInFeatureConfigs(DEFAULT_NAV_GROUPS.observability.id, features)
+        ) {
+          coreStart.application.navigateToApp('discover', { replace: true });
```
kavilla, line 49: personally, i think it's might be easier to just not render the link in the side nav. and just let people who modify the url directly to be explore.

but i would be fine if we plan on failing  to load a saved explore into a non-observability workspace. otherwise it would be weird for saved explore to load into a dashboard but they wouldn't be able to edit it.

src/plugins/explore/public/plugin.ts
kavilla: more context https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9773#discussion_r2088394380

kavilla (commented):
removed the block but prefer if a workspaces member could verify and i could get give a second approval.

but do you mind responding to this https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9773#discussion_r2088394380

also one more nitpick: i think this could use a changelogging fragment since we are iterating and merging into main which could get released and used

src/plugins/explore/public/plugin.ts
joshuali925: > i suggest removing the explore link from the side nav for non-observability workspaces

There's no explore link from the side nav for non-observability workspaces. The end result we want is user sees a single Discover link in the side nav (not implemented yet), and clicking it leads to the correct app without redirection based on the current workspace.

This logic in this PR is not for the side nav, it is for when a user accidentally lands on `/app/explore` in a non-observability workspace, due to redirection from other plugins like "Trace analytics", we don't want the page to break and we don't want to show explore, so we'll redirect them to discover.

If you'd argue we should let the page break and ask plugins to fix their redirection, I'm ok with that too, although it means the plugins needs to check current workspace before redirecting, which we decided offline that it's not something we want plugins to do.

> i'd be fine if we prevent loading saved explores in non-observability workspaces entirely

I think preventing saved explores to load in non-observability workspaces makes more sense. We anyways don't want to support saved explores to load in non-observability workspaces.

Given these two points, maybe instead of redirection, we should show an error page saying "Explore app is only available in Observability workspaces, use Discover instead"?

---

> the embeddable mount won't cause a redirect on dashboard page load if it exsits on a dashboard that isn't observability workspace

On dashboard we'd only want to show a subset of the components in explore, so I think it won't go through this mount function. But I'll double check when we have the embeddable functionality.

joshuali925:
@kavilla thanks, replied.

@wanglam could you help to review the workspace related changes? thanks

joshuali925:
The failures in typescript errors seem incorrect

```
✖ FAILURE: Found 5 new TypeScript errors:

  1. src/plugins/workspace/public/utils.ts:313:6
     Error TS2349: This expression is not callable.

  2. src/plugins/workspace/public/utils.ts:313:14
     Error TS7006: Parameter 'ds' implicitly has an 'any' type.

  3. src/plugins/workspace/public/utils.ts:325:6
     Error TS2349: This expression is not callable.

  4. src/plugins/workspace/public/utils.ts:325:14
     Error TS7006: Parameter 'ds' implicitly has an 'any' type.

  5. src/plugins/workspace/public/utils.ts:326:11
     Error TS7006: Parameter 'ds' implicitly has an 'any' type.
```

I do not see these errors in my editor with tsserver, `ds` also has a proper type. This is a new CI, how do I debug it?

kavilla (approved):

wanglam:
> The failures in typescript errors seem incorrect
> 
> ```
> ✖ FAILURE: Found 5 new TypeScript errors:
> 
>   1. src/plugins/workspace/public/utils.ts:313:6
>      Error TS2349: This expression is not callable.
> 
>   2. src/plugins/workspace/public/utils.ts:313:14
>      Error TS7006: Parameter 'ds' implicitly has an 'any' type.
> 
>   3. src/plugins/workspace/public/utils.ts:325:6
>      Error TS2349: This expression is not callable.
> 
>   4. src/plugins/workspace/public/utils.ts:325:14
>      Error TS7006: Parameter 'ds' implicitly has an 'any' type.
> 
>   5. src/plugins/workspace/public/utils.ts:326:11
>      Error TS7006: Parameter 'ds' implicitly has an 'any' type.
> ```
> 
> I do not see these errors in my editor with tsserver, `ds` also has a proper type. It doesn't make sense if `ds` cannot infer type from `dataSources.filter` This is a new CI, how do I debug it?
> 
> ```tsx
> export const convertDataSourcesToOpenSearchAndDataConnections = (
>   dataSources: DataConnection[] | DataSource[]
> ): Record<'openSearchConnections' | 'dataConnections', DataSourceConnection[]> => {
>   const openSearchConnections = dataSources
>     .filter((ds) => ds.type === DATA_SOURCE_SAVED_OBJECT_TYPE)
> ```
> 
> image

Hi @joshuali925, I think these new type errors are not related to your changes. They are existing errors that are being treated as new because the error line number and column number are different.
image
The type check works as expected in my local environment, as shown in the screenshot above. @s-zx is working on a PR to fix all type errors inside the workspace plugin.

wanglam (approved):

angle943 (approved):

joshuali925:
@wanglam thanks for explaining, although I still don't understand why it's an error. `filter` and `map` are properly typed in typescript with generics, given `dataSources: DataConnection[] | DataSource[]`, tsserver should be able to infer the type. I tried in vscode as well:
image

anyways I merged the PR since it's not something new, thanks

Maosaic:
Hi Team, if you changed error line/column and don't want to fix the existing type error please update the baseline error list(`ts_error_baseline.json`) by running `yarn update-ts-baseline` or it will block CI for other PRs. 

@joshuali925 to have the error show up in vscode, please config your IDE typeScript to use the project typeScript version.

Screenshot 2025-05-19 at 2 40 18 PM

joshuali925:
@Maosaic got it, thanks. will send a PR to update, it would be good to document this somewhere in developer guide

PR #9777, opened 2025-05-07
[State Management Framework] Add Base Implementation, Utils, Hooks & Examples

LDrago27:
## Background

Over the last few releases we have had several regressions due to use of multiple useEffect hooks and multiple state updates. This PR aims at implementing a uni-directional state management framework built on top of our existing rxjs-based plugins. This initial framework establishes the patterns and architecture necessary for a clean Redux migration while remaining fully backwards compatible with our current codebase.

## Design
image

So currently in the Service we are already returning the rxjs observables and this extends the pattern and adds in the selectors and action, which will be exclusively used to read and write values to the state going forward.

**Sample Service Interface** 
```
Service<A, S> {
  selectors: serviceSelector<S>,
  actions: serviceAction<A>,
  // @deprecated selectors and actions are now recommended to interact w/ the service
  data$: BehaviorSubject
}
```

### Where do we store the new baseClasses ? 

* Extend the functionality of opensearch_dashboard_utils & opensearch_dashboard_react 

It already contains some of the state management utils that are being used in OSD. 
**Advantage:** All of the state management utils can be co-located. This plugin is already being used by most of the plugins so adding the state management utils won’t introduce a new dependency across the plugins. 

#### For plugins (Check the example Plugin for reference)

* Each service in the plugin will extend from BaseSelectors and BaseActions to create their own implementation of Selectors and Actions and expose the ServiceSelector and ServiceAction
* Each service will register the Selector and Action with the globalStore using `globalStoreServiceRegister` 
* Now any plugin can access and update the state using the Selector and Action , by using the various hooks defined. 
(Namely usePluginKeys, useActions, useSelector)

image

### useStore
`useStore` provides access to the current store.

#### Example:
```typescript
import { useStore } from '../../../../src/plugins/opensearch_dashboards_react/public';
const store = useStore();
```

### usePluginKeys

`usePluginKeys` retrieves all registered plugin keys. This is useful for dynamically listing available plugins.

#### Example:
```typescript
import { usePluginKeys } from '../../../../src/plugins/opensearch_dashboards_react/public';

const PluginList = () => {
  const pluginKeys = usePluginKeys();

  return (
    <ul>
      {pluginKeys.map((key) => (
        <li key={key}>{key}</li>
      ))}
    </ul>
  );
};
```

###  useSelector

`useSelector` subscribes to specific parts of the state and re-renders the component only when the selected state changes.

#### Example:
```typescript
import { useSelector } from '../../../../src/plugins/opensearch_dashboards_react/public';

const CounterDisplay = () => {
  const count = useSelector('counterPlugin', (state) => state.count);

  return <div>Count: {count}</div>;
};
```

### useActions

`useActions` provides access to actions for a specific plugin.

#### Example:
```typescript
import { useActions } from '../../../../src/plugins/opensearch_dashboards_react/public';

const CounterControls = () => {
  const actions = useActions('counterPlugin');

  return (
    <div>
      <button onClick={actions.increment}>Increment</button>
      <button onClick={actions.decrement}>Decrement</button>
      <button onClick={actions.reset}>Reset</button>
    </div>
  );
};
```
### globalStoreServiceRegister
`globalStoreServiceRegister` is an utility function to register a service in the global store.

#### Example:
```typescript
import { globalStoreServiceRegister } from '../../../src/plugins/opensearch_dashboards_react/public';

  const counterService = new CounterService();
  const counterServiceStart = counterService.start();

  globalStoreServiceRegister(
    COUNTER_SERVICE_PLUGIN_KEY,
    counterServiceStart.selectors,
    counterServiceStart.actions
  );
```

## State Adapters

The BaseSelectors and BaseActions have been enhanced to support both observables and Redux slices as state sources via a State Adapter. The common utils module offers BehaviorSubjectAdapter and ReduxStoreAdapter, enabling services to inject either a Redux store or an observable based on their needs. This abstraction unifies state access across implementations through a consistent set of hooks and utilities.

### Observable Based Selectors, Actions
```typescript
    const state$ = new BehaviorSubject<CounterState>({ count: initialCount });
    const stateAdapter = BehaviorSubjectAdapter<CounterState>(state$);
    const actions = new CounterActions(state$);
    const selectors = new CounterSelectors(state$);
```

### Redux based Slices
```typescript
  const counterSlice = createSlice({
    name: 'counter',
    initialState: { count: 0 } as CounterState,
    reducers: {
      updateState: (state, action: PayloadAction<CounterState>) => {
        return action.payload;
      },
    },
  });
  
  // Create the Redux store
  const store = configureStore({
    reducer: {
      counter: counterSlice.reducer,
    },
  });
  
  const stateAdapter = new ReduxStoreAdapter(
    store,
    (state) => state.counter,
    (updater) =>
      counterSlice.actions.updateState(updater(store.getState().counter))
  );
  
  const actions = new CounterActions(stateAdapter);
  const selectors = new CounterSelectors(stateAdapter);

```
## Testing the changes

**Running the Example App using the above State Management Framework**
- Run opensearch using `yarn opensearch snapshot`
- Run opensearch with examples using `yarn start --run-examples`
- On home page Navigate to Developer examples using the side panel
- Click on `State_Mangament_Example` Card
## Changelog

- feat: Add new state management base implementations and hooks 
### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


ruanyl:
@LDrago27 Are there any context for this PR?

LDrago27:
> @LDrago27 Are there any context for this PR?

Thanks. Added more context.

examples/state_management_example/.gitignore
kavilla, line 1: is this actually needed?

examples/state_management_example/.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  root: true,
+  extends: ['@elastic/eslint-config-kibana', 'plugin:@elastic/eui/recommended'],
+  rules: {
+    '@osd/eslint/require-license-header': 'off',
```
kavilla, line 5: imo i think this just turns off the require icense header which i think we are better of on. so perhaps we can just remoe this file

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_plugin_keys.ts
```diff
@@ -0,0 +1,37 @@
...
+import { useStore } from '../context';
+
+/**
+ * React hook to get all registered plugin keys from the central state store.
+ *
+ * @returns An array of all currently registered plugin keys.
+ *
+ * @example
+ * const pluginKeys = usePluginKeys();
+ * return <ul>{pluginKeys.map(key => <li key={key}>{key}</li>)}</ul>;
```
kavilla, line 16: does it actually return like this?

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_selector.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @param pluginKey - The unique key for the plugin whose state you want to select from.
+ * @param selectorFn - Function to select a part of the plugin's state.
+ * @returns The selected state value, or undefined if not available.
+ *
+ * @example
+ * const value = useSelector('myPlugin', state => state.someValue);
+ *
+ * @throws Error if used outside of PluginStoreContext.Provider
+ */
+export function useSelector<TState, TResult>(
```
kavilla, line 22: i'd mark all the stuff that is exported out as experimental

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_plugin_keys.ts
LDrago27: Yup. it returns the plugin keys as a list.

examples/state_management_example/.gitignore
LDrago27: Thanks. Will remove it . These are some of the artifacts left behind from the plugin creation script.

examples/state_management_example/.eslintrc.js
LDrago27: Thanks. Will remove it . These are some of the artifacts left behind from the plugin creation script.

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_selector.ts
LDrago27: Cool. Added experimental flag to all of the new utils.

LDrago27:
> not blocking this merge but do you think this could benefit with a fast follow of a swagger like api doc. the read me is helpful but something i can just fast look up how what the api is, the input and the output back
> 
> https://editor.swagger.io/
> 
> i do like this approach. it really enables developers to use whatever store they would ike right?

@kavilla 
Yeah, Having a swagger UI will help. I will create a follow up PR to add the Swagger UI for the new hooks. 

As per your question regarding store. Currently I have a basic version of global store , which is provided by default. But Each plugin can choose to provide its own version of Store if they want. Most of the `BaseSelector` , `BaseAction` and `Store` are kept as basic as possible so if we want to switch to redux and do away with observables we can keep the hooks more or less same and just swap out the Base Classes and implemented  Actions and Selectors.

SuZhou-Joe:
Curious how `CounterService` will look like, and actually Assistant has similar pain points that we need to share state across different plugins. Previously we use exposed observable + UIActions, the approach is kind of complex but works inside or outside React render cycle.

For the state management approach:
1. Is the state accessible outside React in a plugin and how to use it?
2. What's the relationship between UIActions and this useActions hook? Should we add deprecate tag to UIActions?

examples/state_management_example/.i18nrc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "prefix": "stateManagementExample",
+  "paths": {
+    "stateManagementExample": "."
+  },
+  "translations": ["translations/ja-JP.json"]
```
angle943, line 6: why japenese?

ananzh (dismissed):

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_selector.ts
```diff
@@ -0,0 +1,46 @@
...
+): TResult | undefined {
+  const store = useStore();
+  if (!store) {
+    throw new Error('useSelector must be used inside PluginStoreContext.Provider');
+  }
+
+  const selector = store.getSelector(pluginKey);
+
+  // Memoize selector function to prevent unnecessary re-renders
+  const select = useMemo(() => (state: TState) => selectorFn(state), [selectorFn]);
```
angle943, line 36: This will just add additional memory, it won't do anything. If the passed selectorFn is memoized by the caller, then we are just adding additional memoization here, which is unnecessary. If the passed selectorFn is not memoized by the caller, then this will re-memoize every time anyways because `selectorFn` is an argument

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_selector.ts
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { useMemo } from 'react';
+import { useSyncExternalStore } from 'use-sync-external-store/shim';
+import { useStore } from '../context';
+/**
+ * @experimental
```
angle943, line 10: should this @experimental flag be inside the /** */ block below?

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_plugin_keys.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  const [pluginKeys, setPluginKeys] = useState(store.getAllRegisteredPluginKeys());
+
+  useEffect(() => {
+    const unsubscribe = store.subscribeToKeysChange(() => {
+      setPluginKeys(store.getAllRegisteredPluginKeys());
+    });
+
+    return () => unsubscribe();
```
angle943, line 35: nitpick: `return unsubscribe;`

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_plugin_keys.ts
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { useEffect, useState } from 'react';
+import { useStore } from '../context';
+/**
+ * @experimental
+ */
```
angle943, line 10: should this @experimental be inside /** */ below?

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_actions.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { useStore } from '../context';
+import { BaseActions } from '../../../../opensearch_dashboards_utils/public';
+
+/**
+ * @experimental
+ */
```
angle943, line 11: should this be inside below /** */ ?

LDrago27: Thanks fixed it.

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_plugin_keys.ts
LDrago27: Thanks fixed it.

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_selector.ts
LDrago27: Thanks fixed it.

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_selector.ts
LDrago27: I have removed the memoization from the selector itself. Now it is upto the caller to actually remember it.

LDrago27:
> > Curious how `CounterService` will look like, and actually Assistant has similar pain points that we need to share state across different plugins. Previously we use exposed observable + UIActions, the approach is kind of complex but works inside or outside React render cycle.
> > For the state management approach:
> > 
> > 1. Is the state accessible outside React in a plugin and how to use it?
> > 2. What's the relationship between UIActions and this useActions hook? Should we add deprecate tag to UIActions?
> 
> @LDrago27 similar question. So we create a BehaviorSubject in the service. Then pass this BehaviorSubject to actions and selectors when creating them right? Then register these actions and selectors with the global store.
> 
> * UIActions: BehaviorSubject lives in service, actions have reference to it. The service exposes the UIActions instance through its setup/start method
> * New system: BehaviorSubject lives in service, actions and selectors have reference to it. The service registers actions and selectors with the global store using a unique key
> 
> Seems the main difference is that in the new system, there's a standardized way (the global store) to access these actions and selectors from anywhere in the application using a string key? So we basically just change the way to access the state?

@ananzh and @SuZhou-Joe 

Sorry for getting back to you guys late. Initially, the plan was to introduce a standardized approach for accessing actions and selectors specifically for observables. However, after discussions with the team, I realized that our system uses a mix of both Redux stores and observables.

To address this, the latest commit refactors the Selectors and Actions to be more generic, allowing them to work with either Redux stores or observables. This ensures that, regardless of the underlying state management strategy, the rest of the abstractions—particularly at the component level—can remain consistent and unchanged.

angle943 (approved):

ananzh (approved):

examples/state_management_example/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+The adapter pattern provides a unified interface for different state backends:
+
+```typescript
+// For RxJS Observables
+const observableAdapter = new BehaviorSubjectAdapter(subject);
+
+// For Redux stores
+const reduxAdapter = new ReduxStateAdapter(store, state => state.mySlice);
+```
```
virajsanghvi, line 26: Is this something we want to promote over using redux directly?

examples/state_management_example/README.md
```diff
@@ -0,0 +1,160 @@
...
+Access actions from a registered service:
+
+```typescript
+// In a React component
+const actions = useAction<CounterState>('counterService');
+// Then use actions
+<button onClick={() => actions.increment()}>+</button>
+```
+
+### useSelector
```
virajsanghvi, line 79: how does one get predefined selectors?

examples/state_management_example/README.md
```diff
@@ -0,0 +1,160 @@
...
+globalStoreServiceRegister('observableCounter', selectors, actions);
+```
+
+### Redux-Based Counter
+
+```typescript
+// Create with Redux pattern
+const store = configureStore({
+  reducer: { counter: counterReducer }
+});
```
virajsanghvi, line 111: due we want usage based around a slice?

examples/state_management_example/README.md
```diff
@@ -0,0 +1,160 @@
...
+const selectors = new CounterSelectors(adapter);
+
+// Register with global store
+globalStoreServiceRegister('reduxCounter', selectors, actions);
+```
+
+### Component Usage
+
+```tsx
+// The same component works with both backends
```
virajsanghvi, line 123: may want to call out this is to demonstrate a single interface with different backends vs expected usage

examples/state_management_example/README.md
```diff
@@ -0,0 +1,160 @@
...
+<CounterDisplay serviceKey="reduxCounter" />
+```
+
+## Key Benefits
+
+1. **Backend Flexibility**: Choose the state management pattern that best fits your needs
+2. **Consistent API**: Components use the same hooks regardless of backend
+3. **Easy Migration**: Switch backends without changing component code
+4. **Type Safety**: Full TypeScript support throughout
+
```
virajsanghvi, line 147: another benefit is easier to contribute state across plugins if there's a global registry

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_selector.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @param pluginKey - The unique key for the plugin whose state you want to select from.
+ * @param selectorFn - Function to select a part of the plugin's state.
+ * @returns The selected state value, or undefined if not available.
+ *
+ * @example
+ * const value = useSelector('myPlugin', state => state.someValue);
+ *
+ * @throws Error if used outside of PluginStoreContext.Provider
+ */
+export function useSelector<TState, TResult>(
```
virajsanghvi, line 21: For redux, isn't this less capable given its useSelector optimizations?

src/plugins/opensearch_dashboards_react/public/state_management/store/action_selector_store.ts
```diff
@@ -0,0 +1,83 @@
...
+   */
+  public getAllRegisteredPluginKeys(): string[] {
+    return this.pluginKeys$.getValue();
+  }
+
+  /**
+   * Retrieves the selector instance for a given plugin key.
+   * @param pluginKey - The plugin key
+   */
+  public getSelector<T = any>(pluginKey: string): BaseSelectors<T> {
```
virajsanghvi, line 62: If we provide this for selectors and actions, but useActions to get to actions, why don't we have a useSelectors?

src/plugins/opensearch_dashboards_react/public/state_management/store/action_selector_store.ts
```diff
@@ -0,0 +1,83 @@
...
+   */
+  private readonly selectors: Record<string, BaseSelectors<any>> = {};
+  /**
+   * Holds actions for each registered plugin, keyed by pluginKey.
+   */
+  private readonly actions: Record<string, BaseActions<any>> = {};
+  /**
+   * Emits the list of all registered plugin keys whenever it changes.
+   */
+  private readonly pluginKeys$ = new BehaviorSubject<string[]>([]);
```
virajsanghvi, line 28: Is there a reason we're adding a new BehaviorSubject?

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_actions.ts
```diff
@@ -0,0 +1,28 @@
...
+ * @example
+ * const actions = useAction<MyState>('myPlugin');
+ * actions.increment();
+ *
+ * @throws Error if used outside of PluginStoreContext.Provider
+ */
+export function useAction<T>(pluginKey: string): BaseActions<T> {
+  const store = useStore(); // Access the central store
+  if (!store) {
+    throw new Error('useAction must be used inside PluginStoreContext.Provider');
```
virajsanghvi, line 24: can we cover this?

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_plugin_keys.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @example
+ * const pluginKeys = usePluginKeys();
+ * return <ul>{pluginKeys.map(key => <li key={key}>{key}</li>)}</ul>;
+ *
+ * @throws Error if used outside of PluginStoreContext.Provider
+ */
+export function usePluginKeys(): string[] {
+  const store = useStore();
+  if (!store) {
+    throw new Error('usePluginKeys must be used inside PluginStoreContext.Provider');
```
virajsanghvi, line 23: can we cover this?

src/plugins/opensearch_dashboards_react/public/state_management/hooks/use_selector.ts
```diff
@@ -0,0 +1,40 @@
...
+  pluginKey: string,
+  selectorFn: (state: TState) => TResult
+): TResult | undefined {
+  const store = useStore();
+  if (!store) {
+    throw new Error('useSelector must be used inside PluginStoreContext.Provider');
```
virajsanghvi, line 27: can we cover this?

src/plugins/opensearch_dashboards_react/public/state_management/store/global_store.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @param pluginKey - The unique key for the plugin.
+ * @param selector - The selector instance for the plugin.
+ * @param action - The action instance for the plugin.
+ */
+export const globalStoreServiceRegister = <T>(
+  pluginKey: string,
+  selector: Store['selectors'][string],
+  action: Store['actions'][string]
+): void => {
+  globalStore.registerService(pluginKey, selector, action);
```
virajsanghvi, line 26: can we cover this?

src/plugins/opensearch_dashboards_utils/common/state_management/state_adapter.ts
```diff
@@ -0,0 +1,102 @@
...
+    private stateSelector: (state: S) => T,
+    private createAction: (updater: (state: T) => T) => any
+  ) {}
+
+  getState(): T {
+    return this.stateSelector(this.store.getState());
+  }
+
+  setState(updater: (prevState: T) => T): void {
+    this.store.dispatch(this.createAction(updater));
```
virajsanghvi, line 74: can we cover this?

virajsanghvi (commented):
Sorry for late feedback - main questions are around accessing predefined selectors, intent related to useSelector and redux (as we lose some functionality), and making sure to cover error cases in tests

PR #9783, opened 2025-05-10
Add doc viewer component to explore plugin

Maosaic:
### Description

Add doc viewer component to explore plugin

## Screenshot

Screenshot 2025-05-09 at 6 50 05 PM

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/packages/osd-stylelint-config/config/restricted_properties.json
+++ b/packages/osd-stylelint-config/config/restricted_properties.json
@@ -11,7 +11,8 @@
       "src/plugins/data/public/ui/typeahead/_suggestion.scss",
       "src/plugins/vis_type_timeseries/public/application/components/_error.scss",
       "packages/osd-ui-framework/src/components/form/check_box/_check_box.scss",
-      "src/plugins/discover/public/application/components/doc_viewer/doc_viewer.scss"
+      "src/plugins/discover/public/application/components/doc_viewer/doc_viewer.scss",
+      "src/plugins/explore/public/components/doc_viewer/doc_viewer.scss"
     ]
   }
 }
--- a/src/plugins/explore/public/components/doc_viewer/doc_viewer.scss
+++ b/src/plugins/explore/public/components/doc_viewer/doc_viewer.scss
@@ -1,0 +1,72 @@
+.osdDocViewerTable {
+  margin-top: $euiSizeS;
+}
+
+.osdDocViewer {
+  pre,
+  .osdDocViewer__value {
+    display: inline-block;
+    word-break: break-all;
+    word-wrap: break-word;
+    white-space: pre-wrap;
+    color: $euiColorFullShade;
+    vertical-align: top;
+    padding-top: 2px;
+  }
+
+  .osdDocViewer__field {
+    padding-top: 8px;
+  }
+
+  .dscFieldName {
+    color: $euiColorDarkShade;
+  }
+
+  td,
+  pre {
+    font-family: $euiCodeFontFamily;
+  }
+
+  tr:first-child td {
+    border-top-color: transparent;
+  }
+
+  tr:hover {
+    .osdDocViewer__actionButton {
+      opacity: 1;
+    }
+  }
+}
+
+.osdDocViewer__buttons,
+.osdDocViewer__field {
+  white-space: nowrap;
+}
+
+.osdDocViewer__buttons {
+  width: 60px;
+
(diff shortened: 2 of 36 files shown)


src/plugins/explore/public/components/doc_viewer/doc_viewer.scss
```diff
@@ -0,0 +1,72 @@
+.osdDocViewerTable {
```
kavilla, line 1: will this cause conflicts since these are the same as the discover table?

so if we style this it will have impact to classic discover?

src/plugins/explore/public/components/doc_viewer/doc_viewer.scss
```diff
@@ -0,0 +1,72 @@
...
+    word-break: break-all;
+    word-wrap: break-word;
+    white-space: pre-wrap;
+    color: $euiColorFullShade;
+    vertical-align: top;
+    padding-top: 2px;
+  }
+
+  .osdDocViewer__field {
+    padding-top: 8px;
```
kavilla, line 18: has this always been hardcoded values?

packages/osd-stylelint-config/config/restricted_properties.json
```diff
@@ -11,7 +11,8 @@
       "src/plugins/data/public/ui/typeahead/_suggestion.scss",
       "src/plugins/vis_type_timeseries/public/application/components/_error.scss",
       "packages/osd-ui-framework/src/components/form/check_box/_check_box.scss",
-      "src/plugins/discover/public/application/components/doc_viewer/doc_viewer.scss"
+      "src/plugins/discover/public/application/components/doc_viewer/doc_viewer.scss",
```
kavilla, line 14: this check was originally put in place to guide us toward using proper plugin patterns for style modifications, rather than directly changing global styles. while the current approach works, it bypasses that intentional guardrail. i think we could explore handling these styles through the plugin system as originally intended with specific class names

kavilla (dismissed):
indirectly could modify the classic discover styling

kavilla:
nit i think this could use a changelog since it's getting into main and shipped at anytime

src/plugins/explore/public/components/doc_viewer/doc_viewer.scss
Maosaic: yes, these are copy pasted

src/plugins/explore/public/components/doc_viewer/doc_viewer.scss
Maosaic: Yes, it will have impact to classic discover if we make any change, but I don't want to rename every classname, since we didn't made any css change. I prefer to change class name when we introduce new style to explore. But we need to document this.

Maosaic:
> nit i think this could use a changelog since it's getting into main and shipped at anytime

@kavilla is it ok if a new version get released before we enable explore?

packages/osd-stylelint-config/config/restricted_properties.json
Maosaic: Do you have any reference how to do this properly? FYI the `scss` file is copy pasted from discover

kavilla: i would just copy and paste and then change the class names i think as yhou suggested as above and then feel free to add  this here still

src/plugins/explore/public/components/doc_viewer/doc_viewer.scss
kavilla: we should clean that up

src/plugins/explore/public/components/doc_viewer/doc_viewer.scss
kavilla: i think that might be too risky ☹️

packages/osd-stylelint-config/config/restricted_properties.json
Maosaic: FYI, I double checked the rule applying to this file, it's for detecting `font-family` change, not global styles.

sejli (approved):
LGTM. I agree with @kavilla's points, but I think the scope of this PR is just to bring over the doc viewer component to the new plugin.

angle943 (approved):

PR #9784, opened 2025-05-12
fix path alias not working for distribution builds

ruanyl:
### Description

When run `opensearch-build` to build OSD distributions, the source codes are copied to `build/opensearch-dashboards`, so the build dir is NOT the project root, the current babel path alias plugin config `babel-plugin-module-resolver` assume OSD is build from project root dir, so the resolved path for alias are incorrect at the moment if it's build from `build/opensearch-dashboards`.

To fix the issue, we will pass the build dir to `babel-plugin-module-resolver` when building OSD distributions.

## Changelog

- fix: path alias not properly resolved if the source code was copied and built from another directory

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/packages/osd-babel-preset/node_preset.js
+++ b/packages/osd-babel-preset/node_preset.js
@@ -66,8 +66,8 @@ module.exports = (_, options = {}) => {
       [
         require.resolve('babel-plugin-module-resolver'),
         {
-          root: [path.resolve(__dirname, '../..')],
-          cwd: path.resolve(__dirname, '../..'),
+          root: options['babel-plugin-module-resolver']?.root ?? [path.resolve(__dirname, '../..')],
+          cwd: options['babel-plugin-module-resolver']?.cwd ?? path.resolve(__dirname, '../..'),
           alias: {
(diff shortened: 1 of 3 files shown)


packages/osd-babel-preset/node_preset.js
```diff
@@ -66,8 +66,8 @@ module.exports = (_, options = {}) => {
       [
         require.resolve('babel-plugin-module-resolver'),
         {
-          root: [path.resolve(__dirname, '../..')],
-          cwd: path.resolve(__dirname, '../..'),
+          root: options['babel-plugin-module-resolver']?.root ?? [path.resolve(__dirname, '../..')],
```
wanglam, line 69: Do we need to use `[]` wrap `options['babel-plugin-module-resolver']?.root` if exists?

ruanyl: May not necessary, root can be an array or string, the plugin will normalize the option

wanglam (approved):

SuZhou-Joe (approved):

PR #9785, opened 2025-05-12
fix: fail to delete multiple workspaces with the same data source

s-zx:
### Description

Change workspace deletion from parallel to serial to avoid operation conflicts on the same data source, which could lead to deletion failures.

## Testing the changes

Delete multiple workspaces with the same data source.

## Changelog

- fix: fail to delete multiple workspaces with the same data source

### Check List

- [ ] All tests pass
- [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/public/components/delete_workspace_modal/delete_workspace_modal.tsx
+++ b/src/plugins/workspace/public/components/delete_workspace_modal/delete_workspace_modal.tsx
@@ -38,7 +38,7 @@ export function DeleteWorkspaceModal(props: DeleteWorkspaceModalProps) {
 
   const deleteWorkspaces = async () => {
     if (selectedWorkspaces && selectedWorkspaces.length > 0) {
-      selectedWorkspaces.forEach(async (selectedWorkspace) => {
+      for (const selectedWorkspace of selectedWorkspaces) {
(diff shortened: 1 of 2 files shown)


src/plugins/workspace/public/components/delete_workspace_modal/delete_workspace_modal.tsx
```diff
@@ -38,7 +38,7 @@ export function DeleteWorkspaceModal(props: DeleteWorkspaceModalProps) {
 
   const deleteWorkspaces = async () => {
     if (selectedWorkspaces && selectedWorkspaces.length > 0) {
-      selectedWorkspaces.forEach(async (selectedWorkspace) => {
+      for (const selectedWorkspace of selectedWorkspaces) {
```
ruanyl, line 41: After changing to `for...of`, I think it can return directly at line 53 when catch error

ruanyl: I think then L55-L72 should be moved after L45

s-zx: Yes, you are right. Thanks for pointing out this error case!

PR #9805, opened 2025-05-21
[workspace] fix default data source not set when re-associating after deletion

Qxisylolo:
### Description
This pr fixes default data source not set when re-associating after deletion
In this implementation, if a default data source is not set, automatically set the first **selected** data source as default

delete set as default action in manage direct query data connection table

### To reproduce

      1. deletes all existing data sources.
      2. Then, re-associates a data source.

## Screenshot
before:

https://github.com/user-attachments/assets/77ca648e-f2be-42d4-8683-337312cf6ddf

after:

https://github.com/user-attachments/assets/560fd387-0f9c-4050-969a-14c81e987575

manage direct query data connection table：

before:

截屏2025-08-06 10 35 40
after

截屏2025-08-06 10 54 45

## Changelog
- fix: fix default data source not set when re-associating after deletion

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
```diff
@@ -163,6 +164,18 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
         true
       );
       setDataSources(finalData);
+
+      // If all data sources have been deleted and new ones are being associated,
+      // we need to ensure a default data source is set.
+      if (previousDataSourcesRef.current === 0 && finalData.length >= 1) {
```
SuZhou-Joe, line 170: Should we check if there is already a default data source? And I do not think it is good to update something in a fetch method.

Qxisylolo: Yes, if there is no default data source and the users try to associate some, we will set the first selected data source as the default. I tested it for non-admin users and it works fine.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_data_connection/manage_direct_query_data_connections_table.tsx
```diff
@@ -110,6 +108,27 @@ export const ManageDirectQueryDataConnectionsTable = ({
...
+      setDefaultDataSourceId(id);
+    } catch (error) {
+      notifications.toasts.addWarning(error.message);
+    }
+  }, [uiSettings, currentWorkspace, notifications.toasts]);
+  useEffectOnce(() => {
+    loadDefaultDataSourceId();
+  });
+
+  useEffectOnce(() => {
```
yubonluo: Nit: Duplicated code

src/plugins/workspace/public/components/data_source_association/data_source_association.tsx
```diff
@@ -72,6 +80,10 @@ export const DataSourceAssociation = ({ excludedDataSourceIds, onComplete, onErr
 
           if (res.success) {
             failedCount = res.result.filter((r) => !!r.error).length;
+            if (!defaultDataSourceId && uiSettings) {
+              // If a default data source is not already set, automatically set the first selected data source as default.
+              await uiSettings.set('defaultDataSource', objects[0].id, UiSettingScope.WORKSPACE);
```
yubonluo: Nit: use `DEFAULT_DATA_SOURCE_UI_SETTINGS_ID` instead of `defaultDataSource`

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_data_connection/manage_direct_query_data_connections_table.tsx
Qxisylolo: Thanks for pointing that out—I'm sorry I just see these comments and have already addressed them.

src/plugins/workspace/public/components/data_source_association/data_source_association.tsx
```diff
@@ -68,6 +77,14 @@ export const DataSourceAssociation = ({ excludedDataSourceIds, onComplete, onErr
...
           if (res.success) {
             failedCount = res.result.filter((r) => !!r.error).length;
+            if (!defaultDataSourceId && uiSettings) {
+              // If a default data source is not already set, automatically set the first selected data source as default.
+              await uiSettings.set(
+                DEFAULT_DATA_SOURCE_UI_SETTINGS_ID,
+                objects[0].id,
+                UiSettingScope.WORKSPACE
+              );
+            }
```
SuZhou-Joe: Seems we will trigger onComplete callback, can we set default data source in the callback, which should be inside data source plugin?

Qxisylolo: Perhaps we could leave this logic in data_source_association.tsx, since data source association is managed in multiple places. If we put it in onComplete, we would need to modify quite a few places for different scenarios, which could make things harder to maintain. Keeping it centralized in one place would be more manageable.

I also think it's fine to pass a default data source, since we're already passing excludedDataSourceIds.

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
```diff
@@ -73,10 +73,6 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
   const { HeaderControl } = navigation.ui;
   const workspaceClient = useObservable(workspaces.client$);
   const DataSourceAssociation = workspaceClient?.ui().DataSourceAssociation;
-  const defaultDataSourceIdRef = useRef(
-    uiSettings.get$<string | null>(DEFAULT_DATA_SOURCE_UI_SETTINGS_ID)
-  );
-  const defaultDataSourceId = useObservable(defaultDataSourceIdRef.current);
```
wanglam, line 79: Just curious about why the `DEFAULT_DATA_SOURCE_UI_SETTINGS_ID` observable doesn't work for this case.

Qxisylolo: get$ is getting observable from cache. and now we will get default ds directly through api

src/plugins/workspace/public/components/data_source_association/data_source_association.tsx
SuZhou-Joe: excludedDataSourceIds is used to hide some data sources, which is required for the component. While defaultDataSourceId is not information that this component should be aware of.

src/plugins/data_source_management/public/components/data_source_table/data_source_table.test.tsx
```diff
@@ -167,8 +168,7 @@ describe('DataSourceTable', () => {
     });
 
     it('should delete datasources & fail', async () => {
-      spyOn(utils, 'deleteMultipleDataSources').and.returnValue(Promise.reject({}));
-      spyOn(utils, 'setFirstDataSourceAsDefault').and.returnValue({});
+      spyOn(utils, 'deleteMultipleDataSources').and.returnValue(Promise.resolve({}));
```
SuZhou-Joe: We should create a new test case instead of update the content, you can see the description of the test case is for testing the failure case.

src/plugins/data_source_management/public/components/data_source_table/data_source_table.test.tsx
```diff
@@ -531,4 +566,36 @@ describe('DataSourceTable', () => {
       expect(component.text()).toContain('connectionAlias1');
     });
   });
+
+  describe('should get default datasources failed', () => {
```
SuZhou-Joe:
```suggestion
  describe('uiSettings APIs throw failure', () => {
```

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
```diff
@@ -93,6 +89,22 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
     Record<string, React.ReactNode>
   >({});
 
+  const [defaultDataSourceId, setDefaultDataSourceId] = useState<string | null>();
+
+  const loadDefaultDataSourceId = useCallback(async () => {
+    try {
+      const scope = workspaces.currentWorkspace$ ? UiSettingScope.WORKSPACE : UiSettingScope.GLOBAL;
```
SuZhou-Joe:
```suggestion
      const scope = !!workspaces.currentWorkspace$.getValue() ? UiSettingScope.WORKSPACE : UiSettingScope.GLOBAL;
```
The observable itself will always be present.

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
```diff
@@ -163,18 +165,57 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
...
+    try {
+      const res = await fetchDataSources();
+
+      if (!defaultDataSourceId && res.length > 0) {
+        // If a default data source is not already set, automatically set the first selected data source as default.
+        // since already fetchDataSources, there is no need to call setFirstDataSourceAsDefault
+        await uiSettings.set(
+          DEFAULT_DATA_SOURCE_UI_SETTINGS_ID,
+          res[0].id,
+          UiSettingScope.WORKSPACE
```
SuZhou-Joe: It can be GLOBAL right? Do we need to check if inside a workspace?

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
```diff
@@ -163,18 +165,57 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
...
+    {
+      renderComponent: (
+        <DataSourceAssociation
+          excludedDataSourceIds={dataSources.map((ds) => ds.id)}
+          onComplete={handleDataSourceUpdated}
+        />
+      ),
+    } as TopNavControlComponentData,
+  ];
+
```
SuZhou-Joe: What about move this section up so that it won't be detected as a change?

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
```diff
@@ -418,7 +461,7 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
     setIsDeleting(true);
 
     deleteMultipleDataSources(savedObjects.client, selectedDataSources)
-      .then(() => {
+      .then(async () => {
         setSelectedDataSources([]);
         // Fetch data sources
         fetchDataSources();
```
SuZhou-Joe: Do we need to await this?

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_data_connection/manage_direct_query_data_connections_table.tsx
```diff
@@ -113,6 +111,24 @@ export const ManageDirectQueryDataConnectionsTable = ({
     },
   };
 
+  const loadDefaultDataSourceId = useCallback(async () => {
+    try {
+      const scope = workspaces.currentWorkspace$ ? UiSettingScope.WORKSPACE : UiSettingScope.GLOBAL;
```
SuZhou-Joe: !!workspaces.currentWorkspace$.getValue()

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_data_connection/manage_direct_query_data_connections_table.tsx
```diff
@@ -113,6 +111,24 @@ export const ManageDirectQueryDataConnectionsTable = ({
...
+    } catch (error) {
+      notifications.toasts.addWarning(error.message);
+    }
+  }, [uiSettings, workspaces.currentWorkspace$, notifications.toasts]);
+
+  useEffectOnce(() => {
+    loadDefaultDataSourceId();
+  });
+
   const setDefaultDataSource = async () => {
```
SuZhou-Joe: Let's remove useless code changes in this file.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_data_connection/manage_direct_query_data_connections_table.tsx
```diff
@@ -566,7 +589,11 @@ export const ManageDirectQueryDataConnectionsTable = ({
             })}
             iconType="flag"
             onClick={async () => {
-              await uiSettings.set(DEFAULT_DATA_SOURCE_UI_SETTINGS_ID, item.id);
+              await uiSettings.set(
+                DEFAULT_DATA_SOURCE_UI_SETTINGS_ID,
+                item.id,
+                UiSettingScope.WORKSPACE
```
SuZhou-Joe: Do we need to detect if inside a workspace or not?

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
Qxisylolo: I'm so sorry I had a commit that I forgot to push. It includes this fix and for a few other mistakes in the comments

src/plugins/data_source_management/public/components/data_source_table/data_source_table.test.tsx
Qxisylolo: sorry I paste the wrong line

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
Qxisylolo: yes it can be GLOBAL. this one is fixed

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
Qxisylolo: We don't need to, this is fixed

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_data_connection/manage_direct_query_data_connections_table.tsx
Qxisylolo: we don't need to detect here, as it locates in a if condition: 
```
  if (currentWorkspace) {
    actionColumn.actions.push({
      render: (item) => {
        return (
          <EuiButtonIcon
            isDisabled={defaultDataSourceId === item.id}
            aria-label="Set as default data source"
            title={i18n.translate('dataSourcesManagement.dataSourcesTable.setAsDefault.label', {
              defaultMessage: 'Set as default',
            })}
            iconType="flag"
            onClick={async () => {
              await uiSettings.set(
                DEFAULT_DATA_SOURCE_UI_SETTINGS_ID,
                item.id,
                UiSettingScope.WORKSPACE
              );
            }}
          />
        );
      },
    });
  }
```

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_data_connection/manage_direct_query_data_connections_table.tsx
Qxisylolo: ok

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
Qxisylolo: ok

src/plugins/data_source_management/public/components/data_source_table/use_data_source_updater.ts
```diff
@@ -0,0 +1,56 @@
...
+import { DEFAULT_DATA_SOURCE_UI_SETTINGS_ID } from '../constants';
+import { UiSettingScope } from '../../../../../core/public';
+import { DataSourceTableItem } from '../../types';
+import { WorkspaceObject, IUiSettingsClient, NotificationsStart } from '../../../../../core/public';
+
+interface UseDataSourceUpdaterDeps {
+  fetchDataSources: () => Promise<DataSourceTableItem[]>;
+  defaultDataSourceId: string | null | undefined;
+  uiSettings: IUiSettingsClient;
+  loadDefaultDataSourceId: () => Promise<void>;
```
wanglam, line 15: Nit: From the hook perspective, I prefer renaming this props to something like `onDefaultDataSourceIdSet`. `loadDefaultDataSourceId` is a little bit strange to this hook.

src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx
```diff
@@ -93,6 +90,24 @@ export const DataSourceTable = ({ history }: RouteComponentProps) => {
     Record<string, React.ReactNode>
   >({});
 
+  const [defaultDataSourceId, setDefaultDataSourceId] = useState<string | null>();
```
wanglam, line 93: Nit: we can separate a hook something like `useScopedDefaultDataSource` to share the loading logic with `edit_data_source.tsx` in the future.

wanglam (approved):

PR #9812, opened 2025-05-22
fix(nav)!: hide item in `all` use-case if `showInAllNavGroup` is false

joshuali925:
### Description

There can be 3 cases for a plugin to register its nav links to the nav group

1. Registered to `all` use-case and other nav groups
2. Registered to `all` use-case only
3. Registered to other nav groups but not `all` use-case

The code is

```ts
    // registers to other nav groups
    core.chrome.navGroup.addNavLinksToGroup(DEFAULT_NAV_GROUPS.observability, [
      {
        id: 'my-plugin',
        category: undefined,
        order: 300,
        // optionally, plugin can set "showInAllNavGroup: boolean" here
      },
    ]);

    // registers to `all` nav group
    core.chrome.navGroup.addNavLinksToGroup(DEFAULT_NAV_GROUPS.all, [
      {
        id: 'my-plugin',
        category: undefined,
        order: 300,
      },
    ]);
```

Previously the `showInAllNavGroup` does not seem to make a difference. In all 3 cases, regardless of `showInAllNavGroup` is undefined, false, or true, the plugin's nav link shows up in the nav group in Analytics workspace even in case 3 where it is not registered to the `all` nav group. The nav link will show under the registered nav group as category, or if the provided `category` is undefined, it will come up under the "Custom" category. It does not show under the provided `category`.

This PR makes it so that
1. When registering to a non-`all` nav group,
   1. if `showInAllNavGroup` is undefined or false, then it will not add the nav link to the nav group in Analytics workspaces
   2. if `showInAllNavGroup` is true, then it adds the nav link under the registered nav group category (for example if registered to `DEFAULT_NAV_GROUPS.observability`, then it shows up under "Observability". Note that it's not the `category` field passed in and this is existing behavior)
2. When registering to `DEFAULT_NAV_GROUPS.all` nav group, `showInAllNavGroup` has no effect, and it will add a nav link to the given `category` in Analytics workspaces

Also note that now 1 and 2 do not interfere with each other. For example, if a nav link is added to `DEFAULT_NAV_GROUPS.all` and `DEFAULT_NAV_GROUPS.observability` with `showInAllNavGroup: true`, then it shows up twice in Analytics workspaces.

This fix **will be a breaking change** for plugins who did not register to the `all` nav group, or they registered to the `all` nav group but also other groups with `showInAllNavGroup: true`. If we are going ahead with this, I'll send PRs to fix plugins.

I looked into the code and seems logic is very complicated, and I don't know if this is the best fix. My requirement is to show a plugin in the Observability workspace but **not** in the Analytics workspace, and it didn't seem possible without this PR.

Question for workspaces team: why not rely on registering to `all` use-case nav group explicitly, rather than implicit control using `showInAllNavGroup`? @SuZhou-Joe do you have some context on the background?

## Changelog
- fix: [nav] hide item in `all` use-case if showInAllNavGroup is false

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


SuZhou-Joe:
Thanks @joshuali925 , `showInAllNavGroup` should be deprecated. The context is that:
Previously left navigation is a nested structure and all use case is a derived use case, and later on the design changed a little bit and we forgot to adjust the logic here. Thanks for making the PR and I will raise a PR to deprecate `showInAllNavGroup` and rely on `addNavLinksToGroup`.

SuZhou-Joe (dismissed):

SuZhou-Joe:
@joshuali925 I raised a PR to deprecate the showInAllNavGroup property and optimize the logic to populate `custom` category, please take a look: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9818

PR #9816, opened 2025-05-23
[Feature] Vended Dashboard Synchronization

RyanL1997:
### Description
Add a synchronization flow for vended dashboards (dashboards that is created by integration workflow)

**Implementation Overview**
Editor _ Mermaid Chart-2025-05-21-183430

Displays:
- Last refresh time
- Polling status
- Manual sync button (triggers REFRESH MATERIALIZED VIEW)

**Feature Controls**
- Feature Flag: `data_source_management.directQuerySyncEnabled` (off by default)
- URL Parameter Override: `&directQuerySync=true|false`
  - Enables per-dashboard or iframe-based opt-in/out behavior
  - Takes precedence over feature flag

**Sync Behavior**
- Only enabled when:
  - All visualizations share the same index-pattern
  - mdsId (remote cluster reference) is consistent across visualizations
- Backed by metadata extraction + validation
- Skips logic entirely when disabled (for performance)

**Multi-Cluster (Neo) Support**
- Extracts mdsId from visualization references
- Enables refreshes on remote Flint materialized views

## Demo

https://github.com/user-attachments/assets/18821622-b43a-40c3-b720-616bdbd5b591

## Changelog
- feat: [Feature][Integration] Vended Dashboards Synchronization #9816

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_source_management/public/plugin.ts
```diff
@@ -303,10 +319,85 @@ export class DataSourceManagementPlugin
...
+
     return {
       getAuthenticationMethodRegistry: () => this.authMethodsRegistry,
     };
   }
 
-  public stop() {}
+  public stop() {
+    if (this.bannerId && this.core) {
+      this.core.overlays.banners.remove(this.bannerId);
```
joshuali925, line 399: is coreStart still available when stop is triggered?

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync.scss
```diff
@@ -0,0 +1,3 @@
+.direct-query-sync {
+  margin-bottom: -16px;
```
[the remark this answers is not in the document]
RyanL1997: fixed.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+interface DirectQuerySyncProps {
+  http: HttpStart;
+  notifications: NotificationsStart;
+  savedObjectsClient: SavedObjectsClientContract;
+  dashboardId: string;
+  removeBanner: () => void;
+}
+
+export const DashboardDirectQuerySync: React.FC<DirectQuerySyncProps> = ({
```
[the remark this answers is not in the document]
RyanL1997: nice suggestion, fixed

src/plugins/data_source_management/framework/types.tsx
```diff
@@ -9,10 +9,12 @@ export enum DirectQueryLoadingStatus {
   SUCCESS = 'success',
   FAILED = 'failed',
   RUNNING = 'running',
+  SUBMITTED = 'submitted',
   SCHEDULED = 'scheduled',
   CANCELED = 'canceled',
   WAITING = 'waiting',
   INITIAL = 'initial',
+  FRESH = 'fresh',
```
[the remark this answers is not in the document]
RyanL1997: good point. I also had a chat with @Swiddis , and we all agree that the state `INITIAL` should handle the case we are dealing with here - for just getting a initial states when page load. I have also reverted this change and re-use the `INITIAL` status for now.

src/plugins/data_source_management/public/constants.ts
```diff
@@ -29,3 +30,34 @@ export const EDIT = '/edit';
 export const DATACONNECTIONS_UPDATE_STATUS = '/status';
 export const INTEGRATIONS_BASE = '/api/integrations';
 export const observabilityMetricsID = 'observability-metrics';
+
+// Module for handling EMR states for Dashboards Progress Bar. All of these except "fresh" are
+// directly from the EMR job run states. "ord" is used to approximate progress (eyeballed relative
+// stage times), and "terminal" indicates whether a job is in progress at all.
+export const EMR_STATES = new Map<string, { ord: number; terminal: boolean }>([
```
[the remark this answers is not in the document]
RyanL1997: Fixed, and I have also did a clean up of these states to get rid of the unused states.

src/plugins/data_source_management/public/plugin.ts
```diff
@@ -185,7 +198,7 @@ export class DataSourceManagementPlugin
 
     // when the feature flag is disabled, we don't need to register any of the mds components
     if (!this.featureFlagStatus) {
-      return undefined;
+      return undefined as any;
```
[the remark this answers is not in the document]
RyanL1997: reverted, these were just some changes that is outside of the scope of this PR. This was coming from the some of the experiments I did for trying to fix the pre-existing ts error.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync.tsx
```diff
@@ -0,0 +1,137 @@
...
+          {i18n.translate('dataSourcesManagement.directQuerySync.dataScheduledToSync', {
+            defaultMessage: 'Data scheduled to sync every {interval}. Last sync: {lastSyncTime}.',
+            values: {
+              interval: syncInfo.refreshInterval
+                ? intervalAsMinutes(1000 * syncInfo.refreshInterval)
+                : '--',
+              lastSyncTime: syncInfo.lastRefreshTime
+                ? `${new Date(syncInfo.lastRefreshTime).toLocaleString()} (${intervalAsMinutes(
+                    new Date().getTime() - syncInfo.lastRefreshTime
+                  )} ago)`
```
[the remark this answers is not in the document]
RyanL1997: fixed, good catch

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
```diff
@@ -0,0 +1,248 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { HttpStart, SavedObjectsClientContract } from 'opensearch-dashboards/public';
+
+export interface SavedObject {
```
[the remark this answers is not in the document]
RyanL1997: fixed.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
```diff
@@ -0,0 +1,248 @@
...
+      if (!panelObj) continue;
+
+      const indexPatternRef = panelObj.references.find((ref) => ref.type === 'index-pattern');
+      if (indexPatternRef) {
+        indexPatternIds.push(indexPatternRef.id);
+      }
+    }
+
+    const uniqueIndexPatternIds = Array.from(new Set(indexPatternIds));
+    const isConsistent = uniqueIndexPatternIds.length <= 1;
```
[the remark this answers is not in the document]
RyanL1997: fixed.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
```diff
@@ -0,0 +1,248 @@
...
+      return null;
+    }
+
+    // Step 6: Generate the refresh query
+    const refreshQuery = generateRefreshQuery(parts);
+
+    return { refreshQuery, refreshInterval, lastRefreshTime, mappingName, mdsId: localMdsId };
+  } catch (err) {
+    onError('Failed to fetch dashboard information');
+    return null;
```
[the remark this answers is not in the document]
RyanL1997: no, we don't need that, and i have removed them from both the class and tests.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
```diff
@@ -0,0 +1,248 @@
...
+      const dataSourceRef = indexPattern.references.find((ref) => ref.type === 'data-source');
+      localMdsId = dataSourceRef?.id; // Can be undefined if no data-source reference
+      indexTitle = indexPattern.attributes.title || null;
+    } else if (uniqueIndexPatternIds.length === 0) {
+      localMdsId = undefined;
+      return null;
+    }
+
+    if (!indexTitle) {
+      throw new Error('Failed to fetch index pattern title');
```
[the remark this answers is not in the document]
RyanL1997: we should always return null, and these were coming from the debug I did. fixed.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
```diff
@@ -0,0 +1,248 @@
...
+}
+
+export async function fetchIndexMapping(
+  index: string,
+  httpClient: HttpStart,
+  mdsId?: string
+): Promise<Record<string, any> | null> {
+  try {
+    const baseUrl = `${DSL_BASE}${DSL_MAPPING}`;
+    const url = mdsId ? `${baseUrl}/dataSourceMDSId=${encodeURIComponent(mdsId)}` : baseUrl;
```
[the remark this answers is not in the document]
RyanL1997: currently it is like this - we have separate routes for regular and mds enable. But yes, I agree that, from the first point, we should just have one route to handle both instead of 2.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/72fa1239edc09780bdb854bef3c9ac70537ffd39/src/plugins/data_source_management/server/routes/dsl.ts#L133

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
```diff
@@ -0,0 +1,248 @@
...
+    const indexPatternIds: string[] = [];
+    for (const panelRef of panelRefs) {
+      const panelObj = response.objects.find(
+        (obj) => obj.id === panelRef.id && obj.type === panelRef.type
+      );
+      if (!panelObj) continue;
+
+      const indexPatternRef = panelObj.references.find((ref) => ref.type === 'index-pattern');
+      if (indexPatternRef) {
+        indexPatternIds.push(indexPatternRef.id);
```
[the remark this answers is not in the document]
RyanL1997: good point, fixed

src/plugins/data_source_management/config.ts
```diff
@@ -15,6 +15,9 @@ export const configSchema = schema.object({
       defaultValue: [],
     }),
   }),
+  dashboardDirectQuerySyncEnabled: schema.boolean({
```
kavilla: nitpick: do we want to future proof this in-case we want to add direct query sync to other pages?

could be something like this? https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/query_enhancements/public/plugin.tsx#L94 and this will not be present in other pages correct? haven't gotten to that part of the code yet so ignore this if this already handled.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.scss
```diff
@@ -0,0 +1,8 @@
+// Adjusts the banner's bottom margin to reduce the gap between the banner and the dashboard content below it.
+.direct-query-sync-banner {
```
kavilla: nit: although not explicit [we use camelCase for class names in sass files ](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/DEVELOPER_GUIDE.md#sass-files)

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.scss
```diff
@@ -0,0 +1,8 @@
...
+  margin-bottom: -16px;
```
kavilla: for consistency we should use the sass variable like `-$ouiSize` https://oui.opensearch.org/1.19/#/guidelines/sass

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.scss
```diff
@@ -0,0 +1,8 @@
...
+
+  .direct-query-sync-link {
```
kavilla: nit i think it could be something like

```suggestion
.directQuerySync {
   &__banner: {
     margin-bottom: -euiSize;
   }
   
   &__link: {
     margin-left: $euiSizeXS;
   }
 }
```
}

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.scss
```diff
@@ -0,0 +1,8 @@
+// Adjusts the banner's bottom margin to reduce the gap between the banner and the dashboard content below it.
```
kavilla, line 6: i think sass files should still include the license header

```
/*
 */
 ```

File: src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.tsx [hunk 19]
```diff
@@ -0,0 +1,138 @@
...
+
+  // Refresh the window when loadStatus becomes 'success'
+  useEffect(() => {
+    if (loadStatus === 'success') {
+      window.location.reload();
+    }
+  }, [loadStatus]);
+
+  // Handle the "Sync Now" action
+  const handleSynchronize = () => {
```
kavilla, line 66: should we wrap this in a useCallback?

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.tsx
```diff
@@ -0,0 +1,138 @@
...
+    if (!syncInfo || !syncInfo.refreshQuery) {
+      return;
+    }
+
+    const match = syncInfo.refreshQuery.match(
+      /REFRESH MATERIALIZED VIEW `(.+?)`\.`(.+?)`\.`(.+?)`/
+    );
+    if (!match || match.length < 4) {
```
kavilla, line 74: whats this 4?

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.tsx
```diff
@@ -0,0 +1,138 @@
...
+  };
+
+  if (!syncInfo) {
+    return null;
+  }
+
+  const state = EMR_STATES.get(loadStatus)!;
+
+  return (
+    <div className="direct-query-sync-banner" data-test-subj="directQuerySyncBar">
```
kavilla: should this respect the advanced settings:
Screenshot 2025-06-03 at 4 57 50 PM

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+    const datasource = match[1];
+
+    const requestPayload = {
+      datasource,
+      query: syncInfo.refreshQuery,
+      lang: 'sql' as const,
+    };
+
+    startLoading(requestPayload);
```
kavilla, line 86: what if this fails? do we have a catch that stops loading

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
kavilla, line 1: these file names are getting a little bit long. some users run the stuff on windows which could hit the max path length and won't be able to open these files.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
```diff
@@ -0,0 +1,236 @@
...
+  database: string | null;
+  index: string | null;
+}
+
+export interface DirectQuerySyncInfo {
+  refreshQuery: string;
+  refreshInterval: number | null;
+  lastRefreshTime: number | null;
+  mappingName: string | null;
+  mdsId?: string;
```
kavilla, line 25: is there a different in the datasource id and the mdsId?

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
```diff
@@ -0,0 +1,236 @@
...
+}
+
+const DIRECT_QUERY_BASE = '/api/directquery';
```
kavilla: is this not available from a common folder of constants?

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
```diff
@@ -0,0 +1,248 @@
...
+    );
+
+    const indexPatternIds: string[] = [];
+    for (const panelRef of panelRefs) {
+      const panelObj = response.objects.find(
+        (obj) => obj.id === panelRef.id && obj.type === panelRef.type
+      );
+      if (!panelObj) continue;
+
+      const indexPatternRef = panelObj.references.find((ref) => ref.type === 'index-pattern');
```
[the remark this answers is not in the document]
kavilla: yeah im surprised too maybe there is but only a quick glance and it seems like it does something similar https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/dashboard/public/application/utils/create_dashboard_container.tsx#L341

src/plugins/data_source_management/public/plugin.ts
```diff
@@ -303,10 +319,85 @@ export class DataSourceManagementPlugin
...
+      if (urlFlag !== null) {
+        return urlFlag;
+      }
+      return this.config.dashboardDirectQuerySyncEnabled;
+    };
+
+    const updateBanner = () => {
+      const appId = this.currentAppId;
+
+      if (appId === 'dashboards') {
```
kavilla, line 344: yeah we could maybe future proof this and make this configurable

src/plugins/data_source_management/public/plugin.ts
```diff
@@ -303,10 +319,85 @@ export class DataSourceManagementPlugin
...
+        const hash = window.location.hash;
+        const isDashboardViewMatch = hash.match(/#\/view\/([^\/?]+)(\?.*)?$/);
```
kavilla: you might be able to use the application utils https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/application/utils/parse_app_url.ts#L49

kavilla: theres other url utils too if it doesn't fit your needs

src/plugins/data_source_management/public/plugin.ts
```diff
@@ -303,10 +319,85 @@ export class DataSourceManagementPlugin
...
+                        this.bannerId = null;
+                      }
+                    },
+                  })
+                )
+              );
+            }
+          } else if (this.bannerId) {
+            core.overlays.banners.remove(this.bannerId);
+            this.bannerId = null;
```
kavilla, line 373: nit: could this logic be simplified  to a single or statement?

src/plugins/data_source_management/public/plugin.ts
```diff
@@ -303,10 +319,85 @@ export class DataSourceManagementPlugin
...
+      }
+    };
+
+    core.application.currentAppId$.subscribe((appId: string | undefined) => {
+      this.currentAppId = appId;
+      updateBanner();
+    });
+
+    window.addEventListener('hashchange', () => {
+      updateBanner();
```
kavilla, line 391: are we guranteed to not have the banner in other parts of the app?

src/plugins/data_source_management/public/plugin.ts
kavilla: oh i see you export the those helpers. should we use it here then?

src/plugins/data_source_management/public/constants.ts
```diff
@@ -3,7 +3,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { i18n } from '@osd/i18n';
 import { DirectQueryDatasourceType } from './types';
+import { DirectQueryLoadingStatus } from '../framework/types';
```
kavilla, line 8: how come the two types files but they both seem to output relatively the same level of typing. meaning it is not `DirectQueryFrameworkLoadingStatus` could it just be apart of the types files above?

src/plugins/data_source_management/public/constants.ts
```diff
@@ -29,3 +31,30 @@ export const EDIT = '/edit';
...
+  [DirectQueryLoadingStatus.RUNNING, { ord: 70, terminal: false }],
+  [DirectQueryLoadingStatus.SUCCESS, { ord: 100, terminal: true }],
+  [DirectQueryLoadingStatus.FAILED, { ord: 100, terminal: true }],
+  [DirectQueryLoadingStatus.CANCELED, { ord: 100, terminal: true }],
+  [DirectQueryLoadingStatus.INITIAL, { ord: 100, terminal: true }],
+]);
+
+export const MAX_ORD = 100;
+
+export function intervalAsMinutes(interval: number): string {
```
kavilla, line 50: how is this used? if this is for non client facing

would it be more precise to go with nano or ms?

src/plugins/opensearch_dashboards_utils/public/index.ts
```diff
@@ -71,6 +71,8 @@ export {
   getStatesFromOsdUrl,
   setStateToOsdUrl,
   withNotifyOnErrors,
+  parseUrl,
```
kavilla, line 74: can we verify that these have tests behind them. just in case. if we expose them and people want to use them but somehow unknowingly change them then it could be a bad time.

src/plugins/data_source_management/public/plugin.ts
```diff
@@ -303,10 +319,85 @@ export class DataSourceManagementPlugin
     };
     setRenderAssociatedObjectsDetailsFlyout(renderAssociatedObjectsDetailsFlyout);
 
+    const getDirectQuerySyncFromUrl = (): boolean | null => {
+      const parsedHash = parseUrlHash(window.location.href);
+      if (parsedHash && parsedHash.query) {
+        const directQuerySync = parsedHash.query.directQuerySync;
+        if (directQuerySync === 'true') return true;
```
kavilla, line 327: qq: instead of having a value it's either the url has it or not. is it possible that it could have neither of these values? might it be easier to assume the presence fo the para is true. and you can explicitly set it to false

also we should do something similar to how the `?embed` does it 

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/dashboard/public/application/components/dashboard_top_nav/dashboard_top_nav.tsx#L62

src/plugins/data_source_management/public/plugin.ts
```diff
@@ -303,10 +319,85 @@ export class DataSourceManagementPlugin
...
+        if (directQuerySync === 'false') return false;
+      }
+      return null;
+    };
+
+    const shouldMountBanner = (): boolean => {
+      const urlFlag = getDirectQuerySyncFromUrl();
+      if (urlFlag !== null) {
+        return urlFlag;
```
kavilla, line 336: is this what we want.

should it not be like this config is to ensure that if we need to shut it off then no matter what the url passes in it will be ignored?

like should it not be something like

```typescript
return this.config.dashboardDirectQuerySyncEnabled && getDirectQuerySyncFromUrl()
```

(I dont think that is accurate return but somethjing similar to that)

kavilla (commented):
overall it's nice! awesome.

just could i get some verification before approving.
1. what happens in the case the dashboard loads and the user goes back. will it start throwing errors because it's trying to mount something that doesn't exist
2. in the same vain, does the banner service encapsulate any error or failures  that could occur. so the dashboard will be unimpacted if there is a problem loading the component?
3. verify the exported utility fucntions have tests. just to ensure we protect ourselves from anyone changing it.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.scss
RyanL1997: Good suggestion, fixed.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.scss
RyanL1997: Fixed, using `-$euiSize` now.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.scss
RyanL1997: Took the suggestion down below and changed to `.directQuerySync`.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.scss
RyanL1997: Added

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
RyanL1997: yes,  I think the windows CI should catch the failure tho. But yes, and I think that `direct_query_data_source_components` played a major part of it 😅 . Any suggestions for fixing the files we introduce in this PR for now? Or maybe we should do a follow up to rename the component dir into ` direct_query_components`..

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
RyanL1997: I think they should be the same thing right? to reference to the correct identifier of remote opensearch connections

src/plugins/data_source_management/public/plugin.ts
RyanL1997: nice catch, I have also applied the `parseUrlHash()` to here. In addition, I'm actually discovered a cleaner way to get the hash + id by just checking `parsedHash.pathname` it should be something like `/view/411eaf12-3e8f-4e70-b8d9-c1f2f00178e2` which is exactly what we need from the url.

src/plugins/data_source_management/public/constants.ts
RyanL1997: The original intention for this framework type is to use this as a place for contain some EMR related types, for example the loading status you are seeing here. But yes, we need to use this for the progress bar which is relying on the EMR status from the EMR jobs too - thats the reason why we are importing them here.

src/plugins/opensearch_dashboards_utils/public/index.ts
RyanL1997: Is this sufficient? https://github.com/opensearch-project/OpenSearch-Dashboards/blob/79744fbbbc91fe1801c81c05a9d0854f271d957b/src/plugins/opensearch_dashboards_utils/public/state_management/url/parse.test.ts#L33

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
RyanL1997: good catch. This should be imported instead of re-create the new ones here. I just fixed.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.tsx
RyanL1997: I believe the current implementation is sufficient and performs well as-is: 
- `DashboardDirectQuerySyncBanner` doesn’t re-render frequently—it renders initially, once when `syncInfo` is set, and potentially when `loadStatus` changes (which triggers a page reload). Since re-renders are infrequent, the cost of recreating `handleSynchronize` is negligible. Additionally, `EuiLink` is a lightweight component that isn’t memoized, so a new `onClick` handler doesn’t cause unnecessary re-renders.
- Also, `handleSynchronize` isn’t used in a dependency array (e.g., in `useEffect`), and there are no memoized child components that would benefit from a stable function reference.

src/plugins/data_source_management/public/constants.ts
RyanL1997: I'd say minute should be the only use case, and here is why:
- for integration, the mv index refresh cycle is 15 mins. According to this, the minutes we can have here should be less than 15mins because when it hits 15mins, it will just initiate a new refresh job in spark, so that this time should be updated again if user reload the page
- here is the reference of what we are having in production for CW console:
Screenshot 2025-06-04 at 4 51 45 PM

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.tsx
RyanL1997: Could you elaborate more on this? Which specific setting this should respect to?

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.tsx
RyanL1997: The `match.length < 4` check ensures that the regex `/REFRESH MATERIALIZED VIEW `(.+?)`\.`(.+?)`\.`(.+?)`/` captures all three expected groups (`datasource`, `database`, `index`) plus the full match. If `match.length < 4`, it means the query is incomplete (e.g., missing `index`), which would result in an invalid `requestPayload` for `startLoading`. This check prevents proceeding with an incomplete query, ensuring robustness.

src/plugins/data_source_management/public/plugin.ts
RyanL1997: I was trying to change it to:
```
          this.bannerId && (core.overlays.banners.remove(this.bannerId), (this.bannerId = null));
```
However, the above gives me the linting error:
```
        361:21  error  Expected an assignment or function call and instead saw an expression  @typescript-eslint/no-unused-expressions
```

src/plugins/data_source_management/public/plugin.ts
RyanL1997: Considering the safe deployment, we would like to have the feature flag in config `off` by default. And for CW console to use it, it will just pass that url flag to `true` to enabled it in their iframe embeddable.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.tsx
RyanL1997: if the query fails, it will received the query job run status as `failed` which will stop the polling. source code: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/70e10115e3075858ff145ad4e37a67534b3cce4f/src/plugins/data_source_management/framework/hooks/direct_query_hook.tsx#L54

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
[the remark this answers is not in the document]
RyanL1997: Transferring the offline conversation with @joshuali925 here:
- I haven't discovered any of the existing api to use yet for this.
- We were trying to discover the implementation of dashboard filter, because there is a usage that when a dashboard has multiple panels that comes from different source of index pattern, there will be a drop down to listed the all the index patterns
- However, after trace all the way down to here:
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/70e10115e3075858ff145ad4e37a67534b3cce4f/src/plugins/dashboard/public/application/utils/create_dashboard_container.tsx#L189
We found out that it is not possible to directly re-use any of the logic here

RyanL1997:
@kavilla here is a video for me testing through the mounting of this feature when it gets enabled to further proof this should be only mounted on dashboard view page:

https://github.com/user-attachments/assets/1708973f-470c-4d93-a4ce-9e3e7df94681

src/plugins/data_source_management/public/plugin.ts
RyanL1997: Transferring some of the discussion with @kavilla , I think for the current usage we may only consider the dashboard as the only scenerio for the only place to take this feature. Since the high level purpose of this feature is targeting on the dashboard only. But yes, I agree that I re name the feature flag to just `data_source_management.directQuerySyncEnabled`. So that we still leave the space for future configurations if we would like to have more use cases for this feature.

src/plugins/data_source_management/public/plugin.ts
RyanL1997: I think for now if we do not have these values, the feature will be considered off by default. I think the current behavior is aligned with the `&embed`, except for url param can actually take over the control of the default feature flag - as I mentioned in this [comment](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9816#discussion_r2130537443), so that the CW console can use this feature even when the feature flag in config file is off by default (safe deployment on NEO and AOS)

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_utils.ts
```diff
@@ -0,0 +1,248 @@
...
+export function extractIndexInfo(
+  mapping: Record<string, any>,
+  concreteTitle: string
+): {
+  parts: IndexExtractionResult | null;
+  refreshInterval: number | null;
+  lastRefreshTime: number | null;
+  mappingName: string | null;
+} {
+  const mappingValues = Object.values(mapping)[0] as any;
```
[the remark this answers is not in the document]
RyanL1997: No NPE occurs if `Object.values(mapping)[0]` is `{}`. Since `{}` is truthy, `if (!mappingValues)` is false, and we proceed to access properties like `mappingValues?.mappings?._meta?.name`. Optional chaining ensures no error when mappings is `undefined`, defaulting to `null`, so the function safely returns `{ parts: null, refreshInterval: null, lastRefreshTime: null, mappingName: null }`.

src/plugins/data_source_management/public/plugin.ts
RyanL1997: Yes, `this.core` (a `CoreStart` instance) is available when `stop` is triggered. I think in the lifecycle, `start` is called before `stop`, and `start` sets `this.core` to the provided `CoreStart` instance. Since `stop` is only invoked on a started plugin, `this.core` will always be non-`null` at that point. 

Additionally, the `if (this.bannerId && this.core)` check provides extra safety, though it’s not strictly necessary given the lifecycle guarantee.

RyanL1997:
For CI failure on both Linux and Windows for `ciGroup10`:
```
   │54 passing (15.0m)
     │1 failing
     │
     │1)    visualize app
     │       
     │         metric chart
     │           should show Percentiles:
     │
     │      retry.try timeout: Error: expected [ '2,147,483,648',
     │   '1st percentile of machine.ram',
     │   '3,221,225,472',
     │   '5th percentile of machine.ram',
     │   '7,516,192,768' ] to sort of equal [ '2,147,483,648',
     │   '1st percentile of machine.ram',
     │   '3,218,227,994.654',
     │   '5th percentile of machine.ram',
     │   '7,338,816,147.36' ]
     │     at Assertion.assert (/home/runner/work/OpenSearch-Dashboards/OpenSearch-Dashboards/packages/osd-expect/expect.js:111:11)
```
The above is not relevant to this change and the same failure is observed on the other open PRs for now.

kavilla (approved):

kavilla:
Existing test failures. Lgtm.

PR #9834, opened 2025-05-30
[Explore] Line chart auto visualization 

abbyhu2000:
### Description

This PR introduces a new visualization system for the Explore plugin, located at `src/plugins/explore/public/components/visualizations`. This component provides a flexible, rule-based approach to visualization.

# Visualization System File Structure and Data Flow

## File Structure

```
src/plugins/explore/public/components/visualizations/
├── line/
│   ├── line_chart_rules.ts         # Defines line chart specific rules
│   ├── line_chart_utils.ts         # Utility functions for line charts
│   ├── line_vis_config.ts          # Line chart configuration and UI
│   └── to_expression.ts            # Line chart specific expression generation (DEPRECATED)
├── utils/
│   ├── to_expression.ts            # Generic expression generation utility
│   └── use_visualization_types.ts   # Hooks and utilities for visualization type detection
├── rule_repository.ts              # Central repository of all visualization rules
├── types.ts                        # Type definitions for the visualization system
├── visualization.tsx               # Visualization component
├── visualization_container.tsx     # Container component that manages visualization state
└── visualization_registry.ts       # Registry that matches data to visualization rules
```

## Data Flow Diagram

```
┌─────────────────────┐
│ Query Execution     │
│ (Data Collection)   │
└──────────┬──────────┘
           │
           ▼
┌─────────────────────┐
│ visualization_      │
│ container.tsx       │◄───────────────────┐
└──────────┬──────────┘                    │
           │                               │
           ▼                               │
┌─────────────────────┐                    │
│ use_visualization_  │                    │
│ types.ts            │                    │
└──────────┬──────────┘                    │
           │                               │
           ▼                               │
┌─────────────────────┐                    │
│ visualization_      │                    │
│ registry.ts         │                    │
└──────────┬──────────┘                    │
           │                               │
           ▼                               │
┌─────────────────────┐                    │
│ rule_repository.ts  │                    │
│ (Rule Selection)    │                    │
└──────────┬──────────┘                    │
           │                               │
           ▼                               │
┌─────────────────────┐                    │
│ utils/              │                    │
│ to_expression.ts    │                    │
└──────────┬──────────┘                    │
           │                               │
           ▼                               │
┌─────────────────────┐                    │
│ Expression Renderer │                    │
└──────────┬──────────┘                    │
           │                               │
           ▼                               │
┌─────────────────────┐                    │
│ visualization.tsx   │────────────────────┘
│ (UI Rendering)      │
└─────────────────────┘
```

## Visualization Rendering Process

The visualization rendering process in OpenSearch Dashboards follows these steps:

1. **Data Collection**: When a query is executed, the system collects data and field schema information.

2. **Rule Matching**: The `getVisualizationType` function in `visualization_registry.ts` analyzes the data structure by categorizing columns into numerical, categorical, and date types.

3. **Rule Selection**: The system finds the best matching visualization rule by:
   - Iterating through all rules defined in `rule_repository.ts`
   - Checking if each rule's `matches` function returns true for the current data structure
   - Selecting the rule with the highest priority chart type

4. **Expression Generation**: Once a rule is selected:
   - The rule ID is stored in the visualization data
   - The visualization container finds the rule by ID
   - The rule's `toExpression` function is called with the appropriate parameters, including the selected chart type
   - This function generates a Vega specification based on the data structure and chart type

5. **Rendering**: The generated expression is passed to the expression renderer, which renders the visualization.

## Rule Selection Logic

Rules are defined in `rule_repository.ts` and follow this pattern:

1. **Rule Definition**: Each rule has:
   - A unique ID
   - A `matches` function that determines if the rule applies to the current data structure
   - A list of supported chart types with priorities
   - A `toExpression` function that generates the appropriate visualization specification

2. **Matching Criteria**: Rules match based on the number and types of columns. Currently this PR introduces four rules. 
   - Rule 1: 1 numerical column & 1 date column → Simple line chart
Screenshot 2025-06-03 at 2 35 02 PM
   - Rule 2: 2 numerical columns & 1 date column → Line-bar chart
Screenshot 2025-06-03 at 2 35 23 PM
   - Rule 3: 1 numerical column, 1 categorical column & 1 date column → Multi-line chart
Screenshot 2025-06-03 at 2 35 36 PM
   - Rule 4: 1 numerical column, 2 categorical columns & 1 date column → Faceted multi-line chart
Screenshot 2025-06-03 at 2 35 46 PM

3. **Priority System**: When multiple rules match, the one with the highest priority chart type is selected.

4. **Chart Type Selection**: Each rule supports multiple chart types (line, area, bar), and the appropriate rendering function is selected based on the chosen chart type.

### Screen recording

https://github.com/user-attachments/assets/dc646953-579b-459f-b60a-5aa6d0c68325

### Storybook integration
Done

### Unit test
Done

### Cypress test
TODO

### Bugs need to be fixed:
* for rule 4, if graphs are too much, some of the graphs will be hidden by the style panel. TODO: make it horizontally scrollable

### Enhancements
* for rule 2, line & bar combo chart, we might need to add mechanism to allow switching between line or bar for the two axis
* chart type persistence when user change query. We first see if the previous chart type has a rule that satisfy the new query result, if yes, we persist the previous chart type; if not, we do a full rule evaluation to find the highest priority chart rule
* move the field type mapping out of explore and to data plugin osdFieldType factory so these info can be shared

## Changelog
- feat: Auto visualization for explore

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/public/query/query_string/language_service/types.ts
```diff
@@ -58,6 +58,7 @@ export interface LanguageConfig {
     formatter?: (value: any, type: OSD_FIELD_TYPES) => any;
   };
   showDocLinks?: boolean;
+  showVisualization?: boolean;
```
ruanyl: Q: Are we going to disable visualization for specific languages?

src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -341,6 +345,7 @@ export const useSearch = (services: DiscoverViewServices) => {
       data$.next({
         status: rows.length > 0 ? ResultStatus.READY : ResultStatus.NO_RESULTS,
         fieldCounts: fetchStateRef.current.fieldCounts,
+        fieldSchema: searchSource.getDataFrame()?.schema,
```
ruanyl: I assume the `useSearch` is used by discover to retrieve data? But for logs type of data, the `fieldSchema` here is not the same the data schema to be used for the visualization, because the `fieldSchema` doesn't contain the `stats` parts in the query. 
@ashwin-pc please correct me if I'm wrong.

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization.tsx
```diff
@@ -0,0 +1,60 @@
...
+            <ReactExpressionRenderer
+              key={JSON.stringify(searchContext) + expression}
+              expression={expression}
+              searchContext={searchContext}
+            />
+          </div>
+        </EuiPanel>
+      </EuiFlexItem>
+      <EuiFlexItem grow={false} style={{ width: '300px' }}>
+        <EuiPanel className="stylePanel" data-test-subj="stylePanel">
```
ruanyl: Prefer a more specific name for `data-test-subj` to avoid duplication.

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization_container.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const {
+    data: {
+      query: { filterManager, queryString, timefilter },
+    },
+    expressions: { ReactExpressionRenderer },
+  } = services;
+  const { indexPattern } = useDiscoverContext();
+
+  // Register all rules
+  visualizationRegistry.registerRule(lineChartRule);
```
ruanyl: Should the rule registration happening at plugin setup() phase?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/line_chart_utils.ts
```diff
@@ -0,0 +1,274 @@
...
+/**
+ * Build proper Vega-Lite mark configuration
+ * @param styles The style options
+ * @param markType The mark type ('line' or 'bar')
+ * @returns The mark configuration object
+ */
+export const buildMarkConfig = (
+  styles: Partial<LineChartStyleControls> | undefined,
+  markType: 'line' | 'bar' = 'line'
+): any => {
```
ruanyl: Can we not specify `any` but let typescript to infer the type?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/line_chart_utils.ts
```diff
@@ -0,0 +1,274 @@
...
+      tooltip: addTooltip,
+      strokeWidth: lineWidth,
+      interpolate: getVegaInterpolation(lineMode),
+    };
+  } else {
+    // Neither line nor dots - fallback to points
+    return {
+      type: 'point',
+      tooltip: addTooltip,
+      size: 0, // Make points invisible
```
ruanyl: Why the type is `point` but make the points invisible? Could you perhaps add a comment?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/line_chart_utils.ts
```diff
@@ -0,0 +1,274 @@
...
+          value: 'Current Time',
+        },
+      }),
+    },
+  };
+};
+
+/**
+ * Apply grid and axis styling
+ * @param axis The base axis configuration
```
ruanyl: Nit: 
```suggestion
 * @param baseAxis The base axis configuration
```

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/line_chart_utils.ts
```diff
@@ -0,0 +1,274 @@
...
+ * @returns The styled axis configuration
+ */
+export const applyAxisStyling = (
+  baseAxis: any,
+  styles: Partial<LineChartStyleControls>,
+  axisType: 'category' | 'value',
+  numericalColumns?: ExploreVisColumn[],
+  categoricalColumns?: ExploreVisColumn[],
+  dateColumns?: ExploreVisColumn[],
+  axisIndex: number = 0 // Add axis index parameter with default value
```
ruanyl: Enum instead of arbitrary number?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/to_expression.ts
```diff
@@ -0,0 +1,575 @@
...
+ * @returns The Vega spec for a combined line and bar chart
+ */
+const createLineBarChart = (
+  transformedData: Array<Record<string, any>>,
+  numericalColumns: ExploreVisColumn[],
+  dateColumns: ExploreVisColumn[],
+  styles: Partial<LineChartStyleControls>
+): any => {
+  const metric1Field = numericalColumns[0].column;
+  const metric2Field = numericalColumns[1].column;
```
ruanyl: For future enhancement: we may need to allow user to switch `metric1Field` and `metric2Field` between `line` and `bar` via `styles`.

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/to_expression.ts
```diff
@@ -0,0 +1,575 @@
...
+      transformedData,
+      numericalColumns!,
+      categoricalColumns!,
+      dateColumns!,
+      styles
+    );
+  }
+
+  // Rule 4: 1 Metric & 1 Date & 2 Categorical → Faceted Multi-line Chart
+  else if (numMetrics === 1 && numDates === 1 && numCategories === 2) {
```
ruanyl: Nit: `else` is unnecessary

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/use_debounced_value.ts
```diff
@@ -0,0 +1,98 @@
...
+ * @param delay The debounce delay in milliseconds
+ * @returns [localValue, setLocalValue] - The local value and setter for immediate UI updates
+ */
+export function useDebouncedValue<T>(
+  value: T,
+  onChange: (value: T) => void,
+  delay: number = 500
+): [T, (value: T) => void] {
+  const [localValue, setLocalValue] = useState<T>(value);
+  const timeoutRef = useRef<NodeJS.Timeout | null>(null);
```
ruanyl: I think it should be `number` not `NodeJS.Timeout` in browser environment:
```suggestion
  const timeoutRef = useRef<number | null>(null);
```

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/use_visualization_types.ts
```diff
@@ -0,0 +1,109 @@
...
+export interface VisualizationTypeResult {
+  visualizationType?: VisualizationType;
+  numericalColumns?: ExploreVisColumn[];
+  categoricalColumns?: ExploreVisColumn[];
+  dateColumns?: ExploreVisColumn[];
+  transformedData?: Array<Record<string, any>>;
+}
+
+const getFieldTypeFromSchema = (schema?: string): ExploreVisFieldType =>
+  FIELD_TYPE_MAP[schema || ''] || 'unknown';
```
ruanyl: Should it fallback to `typeof columnValue` instead of `unknown`?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/use_visualization_types.ts
```diff
@@ -0,0 +1,109 @@
...
+
+// Implement this function to return the visualization type based on the query based on the returned data
```
ruanyl: Is this comment still up-to-date? Sounds like a TODO item

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/line_chart_rules.ts
```diff
@@ -0,0 +1,22 @@
+/*
```
joshuali925: should these not be in legacy? the legacy directory was created for the duplicated files, new components or refactored ones can be outside and we'll make legacy empty eventually

kavilla:
do we find ourselves only needing this within the explore plugin? if someone wants to add another visualization do they have to create a new folder within the plugin or can they register to something.

src/plugins/data/public/query/query_string/language_service/types.ts
angle943: +1. Are we ever going to have languages in the new explore plugin that does not have visualizations?

src/plugins/discover/public/application/view_components/utils/use_search.ts
kavilla: this is just for demo purposes right? should fetchstateref.current i see you add the ref there. so i think we can use this

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization.stories.tsx
```diff
@@ -0,0 +1,274 @@
...
+            <line x1="50" y1="150" x2="250" y2="150" stroke="#69707D" strokeWidth="1" />
+            <line x1="50" y1="150" x2="50" y2="50" stroke="#69707D" strokeWidth="1" />
+          </svg>
+        </div>
+      </div>
+    </div>
+  );
+};
+
+const mockStyleOptions = {
```
angle943: is there a type we can append to this? That way if interface changes, there will be a TS error here

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export interface ExploreVisualizationProps {
+  expression: string;
+  searchContext: IExpressionLoaderParams['searchContext'];
+  styleOptions: LineChartStyleControls;
+  visualizationData: VisualizationTypeResult;
+  onStyleChange: (newOptions: Partial<LineChartStyleControls>) => void;
+  ReactExpressionRenderer: React.ComponentType<{
+    expression: string;
+    searchContext: IExpressionLoaderParams['searchContext'];
```
kavilla: search Context here vs the search context at the root

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization.tsx
angle943: fyi we have been using `explore` prefix to namespace it. We should also be more specific to classNames

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization_container.scss
```diff
@@ -0,0 +1,32 @@
...
+  display: flex;
+  flex-direction: column;
+  width: 300px;
+}
+
+.stylePanel__inner {
+  flex: 1 1 auto;
+  overflow: auto;
+  padding-right: 8px;
+}
```
angle943: can we prefix all of these with `explore`?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization.tsx
```diff
@@ -0,0 +1,60 @@
...
+          <div className="visualizationPanel__inner">
+            <ReactExpressionRenderer
+              key={JSON.stringify(searchContext) + expression}
+              expression={expression}
+              searchContext={searchContext}
+            />
+          </div>
+        </EuiPanel>
+      </EuiFlexItem>
+      <EuiFlexItem grow={false} style={{ width: '300px' }}>
```
kavilla: inline styles

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization_container.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  // Hook to get the visualization type based on the rows and field schema
+  // This will be called every time the rows or fieldSchema changes
+  useEffect(() => {
+    if (fieldSchema) {
+      const result = getVisualizationType(rows, fieldSchema);
+      if (result) {
+        setVisualizationData({ ...result });
+
+        // Todo: everytime the fields change, do we reset the style options?
```
angle943: nitpick: TODO instead of Todo for more easier find

angle943:
can we move everything out of the `legacy` folder?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/line_chart_rules.ts
```diff
@@ -0,0 +1,22 @@
...
+      (numericalColumns.length === 1 && dateColumns.length === 1) || // Rule 1
+      (numericalColumns.length === 2 && dateColumns.length === 1) || // Rule 2
+      (numericalColumns.length === 1 &&
+        categoricalColumns.length === 1 &&
+        dateColumns.length === 1) || // Rule 3
+      (numericalColumns.length === 1 && categoricalColumns.length === 2 && dateColumns.length === 1) // Rule 4
+    );
+  },
+  createConfig: () => createLineConfig(),
+};
```
angle943: 1. Can we have unit tests for this (and other files) that have logic?

2. (Nitpick? Open for discussion) When i see all these `something.length === X || ...`, its hard to see what it is actually doing. Would it make sense to encapsulate these different rules in a variable with a meaninful name?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/line_chart_utils.ts
```diff
@@ -0,0 +1,274 @@
...
+    encoding: {
+      y: {
+        datum: styles.thresholdLine.value,
+        type: 'quantitative',
+      },
+    },
+  };
+
+  // Add tooltip content if enabled
+  if (styles.addTooltip !== false) {
```
angle943: nitpick: `if (styles.addTooltip)`

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/line_chart_utils.ts
```diff
@@ -0,0 +1,274 @@
...
+    thresholdLayer.encoding.tooltip = {
+      value: `Threshold: ${styles.thresholdLine.value}`,
```
angle943: should we translate this?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/use_visualization_types.ts
```diff
@@ -0,0 +1,109 @@
...
+  ) => Promise<string | undefined>;
+}
+
+export type ExploreVisFieldType = 'numerical' | 'categorical' | 'date' | 'unknown';
+
+// Map both OSD_FIELD_TYPES and OPENSEARCH_FIELD_TYPES to ExploreVisFieldType
+// We also need to handle the case where a new field is created with a opensearch type
+const FIELD_TYPE_MAP: Partial<Record<string, ExploreVisFieldType>> = {
+  // Map OSD_FIELD_TYPES to ExploreVisFieldType
+  [OSD_FIELD_TYPES.BOOLEAN]: 'categorical',
```
kavilla: we should prolly add a new optional field to the osdfieldtypes that has this info. and set it with the osd field type factory set these values

ps48:
We'll need a way to prioritize visualization types if there are rule conflicts. Example, line and metric type have the same rule. The a priority order should choose one over the other.

src/plugins/explore/public/application/legacy/discover/application/view_components/canvas/index.tsx
```diff
@@ -186,3 +188,4 @@ export default function DiscoverCanvas({ setHeaderActionMenu, optionalRef }: Vie
 
 const MemoizedDiscoverTable = React.memo(DiscoverTable);
 const MemoizedDiscoverChartContainer = React.memo(DiscoverChartContainer);
+const MemoizedExploreVisualizationContainer = React.memo(ExploreVisualizationContainer);
```
angle943: lets not memoize this unless we have data: https://kentcdodds.com/blog/usememo-and-usecallback#reactmemo-and-friends

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/visualization_registry.ts
```diff
@@ -0,0 +1,43 @@
...
+          categoricalColumns,
+          dateColumns,
+        };
+      }
+    }
+
+    // Todo: Handle the case where no rule matches, render empty state or default table visualization
+    console.log('No matching visualization type found');
+    return;
+  }
```
angle943: what if two rules satisfies this? Do we want to somehow prioritize one rule over another?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/visualization_registry.ts
```diff
@@ -0,0 +1,43 @@
...
+    this.rules.push(rule);
+  }
+
+  /**
+   * Get the matching visualization type based on the columns.
+   */
+  getVisualizationType(columns: ExploreVisColumn[]) {
+    const numericalColumns = columns.filter((column) => column.schema === 'numerical');
+    const categoricalColumns = columns.filter((column) => column.schema === 'categorical');
+    const dateColumns = columns.filter((column) => column.schema === 'date');
```
angle943: can we have these hard-coded strings as enum values for refactorability?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/use_debounced_value.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+    if (max !== undefined && numValue > max) {
+      numValue = max;
+    }
+
+    setDebouncedValue(numValue);
+  };
+
+  return [localValue, handleNumericChange];
+}
```
angle943: can we have unit test for this file?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/collections.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+import { i18n } from '@osd/i18n';
+import { $Values } from '@osd/utility-types';
+export const Positions = Object.freeze({
+  RIGHT: 'right' as 'right',
+  LEFT: 'left' as 'left',
+  TOP: 'top' as 'top',
+  BOTTOM: 'bottom' as 'bottom',
+});
```
angle943: can you just use enum for this?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/types.ts
```diff
@@ -0,0 +1,74 @@
...
+import { VisualizationType } from './utils/use_visualization_types';
+
+export interface ExploreVisColumn {
+  id: number;
+  name: string;
+  schema: ExploreVisFieldType;
+  column: string;
+}
+
+export type ExploreVisFieldType = 'numerical' | 'categorical' | 'date' | 'unknown';
```
angle943: i think enum for this would be better for refactoribility

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/use_visualization_types.ts
```diff
@@ -0,0 +1,109 @@
...
+    indexPattern: IndexPattern,
+    transformedData?: Array<Record<string, any>>,
+    numericalColumns?: ExploreVisColumn[],
+    categoricalColumns?: ExploreVisColumn[],
+    dateColumns?: ExploreVisColumn[],
+    styleOptions?: Partial<LineChartStyleControls>
+  ) => Promise<string | undefined>;
+}
+
+export type ExploreVisFieldType = 'numerical' | 'categorical' | 'date' | 'unknown';
```
angle943: enum for refactoribility

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/types.ts
```diff
@@ -0,0 +1,74 @@
...
+    dateColumns: ExploreVisColumn[]
+  ) => boolean;
+  createConfig: () => VisualizationType;
+}
+
+// Styling: Threshold line configuration
+export interface ThresholdLine {
+  color: string;
+  show: boolean;
+  style: 'full' | 'dashed' | 'dot-dashed';
```
angle943: enum?

Maosaic:
I think we should move the new components out of the `/legacy` folder

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/style_panel/threshold_options.tsx
```diff
@@ -0,0 +1,146 @@
...
+                  data-test-subj="thresholdWidth"
+                />
+              </EuiFlexItem>
+            </EuiFlexGroup>
+          </EuiFormRow>
+        </>
+      )}
+    </EuiPanel>
+  );
+};
```
angle943: can we have unit tests for all these components?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/line_chart_rules.ts
abbyhu2000: moved out of legacy

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/types.ts
abbyhu2000: changed to enum

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/use_visualization_types.ts
abbyhu2000: changed to enum

src/plugins/data/public/query/query_string/language_service/types.ts
abbyhu2000: yes i think we could remove this from the language config since we will always have visualization enabled for explore plugin. and since the eligibility of visualization is solely based on the resulted fields from query result, the rules are the same no matter which query language it is.

abbyhu2000: removed

abbyhu2000:
> do we find ourselves only needing this within the explore plugin? if someone wants to add another visualization do they have to create a new folder within the plugin or can they register to something.

Created an VisualizationRegistry service so external plugin can register to and use it to add additional rules

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization_container.tsx
abbyhu2000: added and exposed a visualization registry service, and during setup() will initiate the service: 
`private visualizationRegistryService = new VisualizationRegistryService();`

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/use_visualization_types.ts
abbyhu2000: outdated

abbyhu2000:
> We'll need a way to prioritize visualization types if there are rule conflicts. Example, line and metric type have the same rule. The a priority order should choose one over the other.

changed the rule mapping logic: 
* previous: mapped to a specific chart type's rule
* now: mapped to a specific rule, each rule will have multiple available chart types with different priorities, default is to take the highest priority chart type

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/visualization_registry.ts
abbyhu2000: responded as previous comment, changed the rule mapping logic

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization.tsx
abbyhu2000: added explore as prefix to all `data-test-subj`

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/line_chart_utils.ts
abbyhu2000: defined a type for the Vega-Lite mark config

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/line_chart_utils.ts
abbyhu2000: if user toggle off both dots and line, then essentially the graph should show nothing, thus we just assign a type as 'point' but the size is 0 so they are invisible. Added comment: 
```
// Toggle off both line and dots - show empty
    return {
      type: 'point',
      tooltip: addTooltip,
      size: 0, // Make points invisible
    };
```

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/to_expression.ts
abbyhu2000: i could add a TODO issue to track this enhancement

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/line_chart_utils.ts
abbyhu2000: updated to enum

```
export enum ValueAxisPosition {
  Left = 0,
  Right = 1,
}

File: src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/to_expression.ts (hunk 10 above)
Comment (abbyhu2000) [reply src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/to_expression.ts]: outdated. Changed the structure so each rule has its own function to_expression function at src/plugins/explore/public/components/visualizations/rule_repository.ts

File: src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/use_debounced_value.ts (hunk 11 above)
Comment (abbyhu2000) [reply src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/use_debounced_value.ts]: in package node_modules/@types/node/timers.d.ts, setTimeout type is NodeJS.Timeout. Changed it to number would give typescript error

```
function setTimeout<TArgs extends any[]>(
            callback: (...args: TArgs) => void,
            ms?: number,
            ...args: TArgs
        ): NodeJS.Timeout;

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/use_visualization_types.ts
abbyhu2000: i fallback to unknown here to indicate this is a new type that the explore visualization system do not understand thus it will not be used as a valid field in the resulted visualization

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization.stories.tsx
abbyhu2000: yes, its `LineChartStyleControls` and added to the story file

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization_container.scss
abbyhu2000: done

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization.tsx
abbyhu2000: removed

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization_container.tsx
abbyhu2000: fixed

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/line_chart_utils.ts
abbyhu2000: i only check against false since we want default behavior is to have tooltip. so if addTooltip is undefined, we enable it

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/line/line_chart_utils.ts
abbyhu2000: added i18n

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/collections.ts
abbyhu2000: use enum for positions

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/types.ts
abbyhu2000: fixed to enum

src/plugins/explore/public/application/legacy/discover/application/view_components/canvas/index.tsx
abbyhu2000: i think use memoize here is beneficial because it helps prevent unnecessary re-renders of the VisualizationContainer component when its props (rows and fieldSchema) haven't changed when DiscoverCanvas might re-render frequently due to other state changes so we do not repeat expensive operations such as data transformation and expression generation

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/use_debounced_value.ts
ruanyl: Actually we can do `window.setTimeout()` which is should return a `number`

ruanyl (approved):

src/plugins/data/common/osd_field_types/types.ts
```diff
@@ -62,7 +62,7 @@ export enum OPENSEARCH_FIELD_TYPES {
   HALF_FLOAT = 'half_float',
   SCALED_FLOAT = 'scaled_float',
   DOUBLE = 'double',
-  INTEGER = 'integer',
+  INTEGER = 'int', // opensearch has changed type from 'integer' to 'int'
```
ruanyl: @abbyhu2000 I approved this PR, feel free to update it in your next PR:
```suggestion
  INTEGER = 'integer',
  INT = 'int',
```

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/utils/use_visualization_types.ts
abbyhu2000: yes we should move this mapping out of the explore plugin and just build the visFieldType as part of `osdFieldType`. Refactoring this change involves making changes to multiple functions in data plugin such as use_search(), getDataFrames(), are you okay with making this change as part of a follow-up PR? @kavilla

src/plugins/discover/public/application/view_components/utils/use_search.ts
abbyhu2000: @ruanyl i am not quite sure that i understand what your concern here is?

src/plugins/explore/public/application/legacy/discover/application/components/visualizations/explore_visualization.tsx
abbyhu2000: what do you mean by this? @kavilla

ruanyl (approved):

src/plugins/explore/public/services/visualization_registry_service.ts
```diff
@@ -0,0 +1,84 @@
...
+    };
+  }
+
+  /**
+   * Get the visualization registry instance
+   */
+  public getRegistry(): VisualizationRegistry {
+    return this.registry;
+  }
+}
```
angle943, line 84: sorry for being a unit test stickler, but this file should be easy to unit test, can we get one for it?

abbyhu2000: added

abbyhu2000:
Cigroup 10 and the link checker failure is unrelated to this PR

angle943 (dismissed):
Amazing work, thank you!

abbyhu2000:
investigating on ciGroup 5 and 6 failures, they do not seem to relate to functional failures. Only fail on windows test with below error messaged in build plugins stage, linux tests pass.
Screenshot 2025-06-09 at 2 27 29 PM

ruanyl (approved):

angle943 (approved):

PR #9842, opened 2025-06-03
[Explore] create new explore/discover experience banners

angle943:
### Description

- Removes the temporary "WIP" banner in Explore plugin
- Creates a new experience banner for both new and classic experience.

- The banners will only appear if explore-plugin is enabled and you are in observability workspace.
- The new experience banner will only appear if you are in the explore app
- The classic experience banner will only appear if you are in classic app

## Changelog
- feat: add new experience banners for explore for both new and classic
- feat: update data plugin's __enhance type to include promises

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


angle943:
ciGroup10 are nown failures

src/plugins/explore/public/components/experience_banners/create_query_editor_extension_config.tsx
```diff
@@ -0,0 +1,49 @@
...
+      const [coreStart] = await core.getStartServices();
+      const currentAppId = await coreStart.application.currentAppId$.pipe(take(1)).toPromise();
+      return currentAppId === 'explore' ? (
+        <NewExperienceBanner />
+      ) : (
+        <ClassicExperienceBanner
+          navigateToExplore={() =>
+            coreStart.application.navigateToApp('explore', { replace: true })
+          }
+        />
```
joshuali925: maybe this can go into a wrapper ExperienceBanner, inside it determines which actual banner to render using async logic, so getBanner doesn't need to be async?

src/plugins/explore/public/components/experience_banners/create_query_editor_extension_config.tsx
```diff
@@ -0,0 +1,49 @@
...
+    order: 1,
+    isEnabled$: async () => {
+      const features = await core.workspaces.currentWorkspace$
+        .pipe(take(1))
+        .toPromise()
+        .then((workspace) => workspace?.features);
+      const isExploreEnabledWorkspace =
+        (features && isNavGroupInFeatureConfigs(DEFAULT_NAV_GROUPS.observability.id, features)) ??
+        false;
+      return new BehaviorSubject(isExploreEnabledWorkspace);
```
joshuali925: i believe if you use the observable subscribe to get `features`, then `isEnabled$` doesn't need to be async. I'm not very familiar with observables but it might look like this according to claude
```tsx
isEnabled$: () => core.workspaces.currentWorkspace$.pipe(
  map((workspace) => workspace?.features),
  map((features) => 
    (features && isNavGroupInFeatureConfigs(DEFAULT_NAV_GROUPS.observability.id, features)) ?? false
  ),
  take(1),
  shareReplay(1)
);
```

src/plugins/explore/public/components/experience_banners/create_query_editor_extension_config.tsx
angle943: That is possible, and admittedly the easier path to solve this problem, but I was trying to avoid that. One of the fundamental tenets we were trying to achieve with this refactor was to make components only receive props it really needs, so that it is truly re-usable and maintainble. 

Passing around heavy stuff like `core` goes against that principle, because now the banner needs to know the business logic of `core` just to perform a simple thing such as rerouting

But since that logic needs to exist someplace, the most appropriate place IMO is near the `pluginSetup` level (for which this file is jsut a helper function for).

This requires me to refactor all the `getX` within the data plugin to handle async code, but i think that is overall a flexibility we do want

src/plugins/explore/public/components/experience_banners/create_query_editor_extension_config.tsx
angle943: nice that worked. i updated the PR to have that.

But what is not nice is how freaking complicated that is just to do a simple thing. I dislike rsjx very much.

I still want to keep the async support for `isEnabled$` though, so that other plugin devs don't need to be restricted to making that function synchrounous

src/plugins/explore/public/components/experience_banners/new_experience_banner/new_experience_banner.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <EuiLink
+        href={NEW_DISCOVER_INFO_URL}
+        target="_blank"
+        data-test-subj="exploreNewExperienceBanner__learnMore"
+      >
+        new discover experience
+      </EuiLink>
+      , or{' '}
+      <EuiLink onClick={handleSwitch} data-test-subj="exploreNewExperienceBanner__switch">
+        switch back to classic discover
```
joshuali925, line 45: these texts need i18n?

src/plugins/explore/public/components/experience_banners/new_experience_banner/new_experience_banner.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+import React, { useState } from 'react';
+import { EuiCallOut, EuiLink } from '@elastic/eui';
+import './new_experience_banner.scss';
+import { NEW_DISCOVER_INFO_URL, SHOW_CLASSIC_DISCOVER_LOCAL_STORAGE_KEY } from '../constants';
+
+export const NewExperienceBanner = () => {
+  const [isVisible, setIsVisible] = useState<boolean>(true);
+
+  // TODO: This should open a modal to add more friction
```
joshuali925, line 14: i don't really think a modal would add friction if user has decided to switch, but maybe we can get some feedback

docs/plugins/data/query-editor-enhancements.md
```diff
@@ -33,7 +33,7 @@ export interface QueryEditorExtensionConfig {
    * A function that determines if the query editor extension is enabled and should be rendered on UI.
    * @returns whether the extension is enabled.
    */
-  isEnabled$: (dependencies: QueryEditorExtensionDependencies) => Observable<boolean>;
+  isEnabled$: (dependencies: QueryEditorExtensionDependencies) => Promise<Observable<boolean>> | Observable<boolean>;
```
joshuali925: do we still need `Promise<Observable<boolean>>`? i feel it's a bit anti pattern, it should either be promise or observable, but an observable promise feels weird

docs/plugins/data/query-editor-enhancements.md
```diff
@@ -46,14 +46,14 @@ export interface QueryEditorExtensionConfig {
    * @param dependencies - The dependencies required for the extension.
    * @returns The query editor extension component.
    */
-  getComponent?: (dependencies: QueryEditorExtensionDependencies) => React.ReactElement | null;
+  getComponent?: (dependencies: QueryEditorExtensionDependencies) => Promise<React.ReactElement> | React.ReactElement | null;
```
joshuali925: one reason I don't want to have async components is that it does not explicitly require consumers to define loading states (like fallback components with Suspense). I think they are important for user experience, and consumer should decide if they should be null or should be something else

src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extension.tsx
```diff
@@ -145,10 +147,41 @@ export const QueryEditorExtension: React.FC<QueryEditorExtensionProps> = (props)
...
+    };
+
+    const setComponents = async () => {
+      const [
+        bannerResult,
+        componentResult,
+        queryControlButtonsResult,
+        bottomPanelResult,
+        actionBarButtonsResult,
+      ] = await Promise.all([
```
joshuali925: if any component getter fails, this would cause all things to not load. should this be allSettled?

src/plugins/explore/public/components/experience_banners/classic_experience_banner/classic_experience_banner.tsx
```diff
@@ -0,0 +1,53 @@
...
+        >
+          Learn more
+        </EuiLink>
+      </div>
+      <EuiSmallButton
+        className="exploreClassicExperienceBanner__newExperienceButton"
+        data-test-subj="exploreClassicExperienceBanner__newExperienceButton"
+        onClick={switchToNewExperience}
+      >
+        Try the new Discover
```
joshuali925, line 49: these need i18n?

src/plugins/explore/public/components/experience_banners/constants.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const NEW_DISCOVER_INFO_URL =
+  'https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9826';
```
joshuali925, line 9: i remember there's a URL service, does this need to go in that?

src/plugins/explore/public/components/experience_banners/create_query_editor_extension_config.tsx
joshuali925: commented here https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9842#discussion_r2130007052

I feel it might be anti pattern to have an observable promise

src/plugins/explore/public/components/experience_banners/create_query_editor_extension_config.tsx
joshuali925: > components only receive props it really needs

the wrapper can receive a promise that resolves to which version to render, i don't think it needs to know `core` or any actual logic. It only needs to await that promise and actually render the right banner

related comment here https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9842#discussion_r2130029796

src/plugins/explore/public/components/experience_banners/new_experience_banner/new_experience_banner.tsx
angle943: I talked about it with Ashwin in our team channel. I had issues with putting i18n in this text because of the string interpolation of the links (not all languages follow the same grammatical order).

We decided to skip i18n for this temporary banners

src/plugins/explore/public/components/experience_banners/new_experience_banner/new_experience_banner.tsx
angle943: Product decision, the rationale is to convey to the customer what they are losing by switching (like visualization, log pattern, etc)

docs/plugins/data/query-editor-enhancements.md
angle943: hmm i see your point, let me remove that here

src/plugins/explore/public/components/experience_banners/classic_experience_banner/classic_experience_banner.tsx
angle943: same comment as above

src/plugins/explore/public/components/experience_banners/create_query_editor_extension_config.tsx
angle943: true.... but i would argue a promise still does not make sense in this context. But your other comment is a super valid point and IMO should override my stubbornness in props pattern. will change

src/plugins/explore/public/components/experience_banners/create_query_editor_extension_config.tsx
angle943: yeah agreed its weird. will change

docs/plugins/data/query-editor-enhancements.md
angle943: yeah that is a valid point, i agree and will remove

src/plugins/data/public/ui/query_editor/query_editor_extensions/query_editor_extension.tsx
angle943: good point. the other point to bring up is what to do if any fails, but since that was not handled prior, i will mark it as out of scope in my current pr. but based on our other comments, i think i can just undo this change

src/plugins/explore/public/components/experience_banners/constants.ts
angle943: Thank you for this comment, this is good info. I did look at the `doc_links_service.ts`, and looks like most of the links are pertaining to opensearch doc. Because my changes currently link to a github issue, and it is local to my banners, i will keep it there. But if reqs change and we have official doc that we want to link it to instead, will make the update

src/plugins/explore/public/components/experience_banners/create_query_editor_extension_config.tsx
```diff
@@ -0,0 +1,51 @@
...
+        map((workspace) => workspace?.features),
+        map(
+          (features) =>
+            (features &&
+              isNavGroupInFeatureConfigs(DEFAULT_NAV_GROUPS.observability.id, features)) ??
+            false
+        ),
+        take(1),
+        shareReplay(1)
+      ),
```
ruanyl: Seems we can remove `take(1)` and `shareReplay(1)`
```suggestion
      core.workspaces.currentWorkspace$.pipe(
        map((workspace) => workspace?.features),
        map(
          (features) =>
            (features &&
              isNavGroupInFeatureConfigs(DEFAULT_NAV_GROUPS.observability.id, features)) ??
            false
        ),
      ),
```

src/plugins/explore/public/components/experience_banners/experience_banner_wrapper/experience_banner_wrapper.tsx
```diff
@@ -0,0 +1,44 @@
...
+    callInitializeBannerWrapper();
+  }, [initializeBannerWrapper]);
+
+  if (!isMounted) {
+    return null;
+  }
+
+  // there is no reason for !!handleSwitchToExplore to be false here, but in case it is, then
+  // we will not show the banner
+  return !!handleSwitchToExplore && showClassic ? (
```
ruanyl, line 39: I'm a bit confuse here, we don't need to check `!!handleSwitchToExplore` here and store it as a state right? Can be simply like:
```suggestion
  return showClassic ? (
```

angle943: (Typing from phone. OOTO today)

Yeah that would work, but typescript would not like it. I can make the callback optional in the banner component but philosophically speaking that does not make sense, as that callback is critical to the banner. Another thing I can do is just coeorce the type to be non nullable. 

I guess the most "correct" way to handle this is to have one state object for both the Boolean and callback as an object and type the state to be null | {flag: false} | {flag: true, callback: fn}

And if we do that we don't even need the isMounted Boolean. 

Tagging @joshuali as he's taking over this while I'm out

src/plugins/explore/public/components/experience_banners/create_query_editor_extension_config.tsx
angle943: Thank you! My rsjx knowledge is very lacking. @joshuali925

src/plugins/explore/public/components/experience_banners/experience_banner_wrapper/experience_banner_wrapper.tsx
joshuali925: `handleSwitchToExplore` is nullable but a required prop for the classic banner, so i think the current code is reasonable

joshuali925 (approved):

abbyhu2000 (approved):

joshuali925:
merging in, the ciGroup10 is failing for all PRs

PR #9844, opened 2025-06-03
Remove Migrations for Explore App

LDrago27:
### Description

If you try to run Opensearch Dashboards against an existing Opensearch Domain using the command `yarn start:explore` command. The Application fails to boot up and gets stuck showing the following error message. 

```
server    log   [22:36:41.623] [info][savedobjects-service] Detected mapping change in "properties.explore"
```

This was being caused by the migration script present within the Explore App. This PR aims at removing the migration script for Explore App to resolve the issue. 

## Changelog

- fix: Remove Migrations for Explore App

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


kavilla (commented):
I think we should delete the file or empty out the contents of the file. If we retain the file we should rename it to like explore

virajsanghvi:
> I think we should delete the file or empty out the contents of the file. If we retain the file we should rename it to like explore

Agree with this, what's the value of keeping the file around?

LDrago27:
Thanks. @virajsanghvi @kavilla Removed the unused migration files.

joshuali925 (approved):
agree that we don't need this migration, thanks

angle943:
i just want to clarify. in classic discover plugin, we have this migration script, which i think is valid (because otherwise why would we have it). So having it again in explore was doing some sort of a double migration?

So if we remove it from explore, we are still doing the necessary migration for the user, right? (via the old discover plugin)

angle943 (approved):

LDrago27:
> i just want to clarify. in classic discover plugin, we have this migration script, which i think is valid (because otherwise why would we have it). So having it again in explore was doing some sort of a double migration?
> 
> So if we remove it from explore, we are still doing the necessary migration for the user, right? (via the old discover plugin)

@angle943 Yup, even if we remove the migrations in explore. The migration is Discover will still work as intended.

PR #9846, opened 2025-06-04
[CVE-2024-4068] add package resolution for micromatch

angle943:
### Description

- Remove `jest-specific-snapshot` from osd-storybook plugin. It was not being used and it had transitive dependency on micromatch@^3.1.10, which has a [dependency on braces@^2.3.1](https://github.com/micromatch/micromatch/blob/10aacd747e9bfe3cef5b2132b085dd8f34be1582/package.json#L41), which is a known [CVE](https://advisories.opensearch.org/advisory/CVE-2024-4068)
- `@storybook/core-common@6.5.16`, `@storybook/core-server@6.5.16`, and `@storybook/manager-webpack4@6.5.16` all had dependency on `webpack@4`. Webpack4 has a dependency on `micromatch "^3.1.10"`, which has a bad braces dependency like above. Therefore I added package resolution on `**/@storybook/**/webpack` to make it to miki's webpack
- After the above two changes, the three dependencies that causes the CVE due to micromatch v3 is:
    - anymatch@^2.0.0
    - fork-ts-checker-webpack-plugin@^4.1.6
    - sane@^4.0.3

- `anymatch` and `sane` are related. The dependency tree looks like `@storybook/addon-essentials@~6.3.12 > @storybook/addon-docs@6.3.13 > @jest/transform@^26.6.2 > jest-haste-map@^26.6.2 > sane@^4.0.3 > anymatch@^2.0.0`
- `anymatch` needs to be at least v3: https://github.com/micromatch/anymatch/blob/d474c820e61a85efcfc90e5b039d8ea14a4b4fae/package.json (it removes `micromatch` as a dependency)
- `sane` needs to be v5.0.1: https://github.com/amasad/sane/blob/master/package.json#L32
-  `jest-haste-map` has to be v27: https://github.com/jestjs/jest/commit/be16e47afcc9f64653b9a47782cb48a5ca243e65#diff-798ca62684058ef217e8631ed75a87d8c2f249a0155d2649a49f1eef900c746f
- jest upgrade to v27 seems safe?: https://jestjs.io/blog/2021/05/25/jest-27
    - One way to resolve this is to add `"**/@storybook/**/@jest/transform": "^27.0.0"`. But will not do so, reasoning will be below
 - For `fork-ts-checker-webpack-plugin`, we must get to v5.0.0. This has [several breaking changes](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/releases?page=11), one being "Changed options format". @storybook/builder-webpack4@6.5.16 uses these options: https://github.com/storybookjs/storybook/blob/7db13fface8f05b870a16b90694a840785830f7e/lib/builder-webpack4/src/preview/iframe-webpack.config.ts#L219
 - Therefore, I think the safest route to solve the above two issues is to add a package resolution to `"**/@storybook/**/micromatch": "^4.0.8"`. Since looking at the [v4 changelog](https://github.com/micromatch/micromatch/blob/master/CHANGELOG.md), it is safe to do so:
     - sane@4.0.3: https://github.com/amasad/sane/blob/5c9f0e02ac5ecebe397ede1b05f37df50d6e32aa/src/common.js#L62-L63
     - fork-ts-checker-webpack-plugin@4.1.6: https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/77effc8e7168d063abae4027bf9d13b31c9551d3/src/index.ts#L575
     - anymatch@2.0.0: https://github.com/micromatch/anymatch/blob/c906c0234c5fb676426da90e429c909566106028/index.js

## Changelog
- chore: add package resolutions to fix braces CVE-2024-4068

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


angle943:
ciGroup10 has been consistently failing. we are looking into a fix

AMoo-Miki:
I did a superficial look at the libs dependent on `braces@2` and saw that a lot of `makeRe` were used which was removed in `braces@3`. As such, we cannot simply override the older version. We will need to get rid of or bump all the libs or their parents who end up bringing in `micromatch@3` or `braces@2`.

angle943:
Thank you so much @AMoo-Miki . i see now fixing CVE is not as easy :D .

I have found what i think is the correct way to solve this CVE. I have updated the description. Please let me know if this is sufficient.

kavilla:
does this have an associated issue that this can close?

do we know if this only impacts 3.x? do we need to backport it?

kavilla (dismissed):

changelogs/fragments/9846.yml
```diff
@@ -0,0 +1,2 @@
+chore:
+- Add package resolutions to fix braces CVE ([#9846](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9846))
```
kavilla: nit: including the cve id might be preferable here incase any developer is looking at the changelog only to see if a cve was handled

angle943: good call, updated!

angle943:
The associated issue is already closed: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6792

I think its because we re-introduced it?

This is only for 3.0, since we did not backport storybook to 2.x

LDrago27 (approved):

PR #9853, opened 2025-06-06
[Chore] Remove alias in permission controlled UI settings wrapper

FriedhelmWS:
### Description

Change from using alias to relative path. 

This is to avoid plugins that does not modify babel config due to recent change in aliasing from failing the test cases.

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ruanyl:
@FriedhelmWS which plugin was failed?

PR #5429, opened 2023-11-03
[OSCI][FIX] Discover: Fix the Subfields Display Issue

MadaniKK:
### Description

Fix the subfields display issue in Discover left nav bar.

### Issues Resolved
Issue #5400

## Screenshot
Before:
image

After:
image

### Check List

- [☑️ ] All tests pass
  - [ ☑️ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [☑️  ] Commits are signed per the DCO using --signoff

Issue #5400, opened 2023-10-29
[BUG]Discover: Some Fields Aren't Displayed in the Left Navigation Bar


MadaniKK:
**Describe the bug**
In Discover some fields are not being displayed in the left Field Nav              
              
**To Reproduce**
* step 1: go to dev tools. create an index and add nested field.
```
PUT test-2/_doc/1
{
  "products": [
    {
      "base_price": 28.99,
      "discount_percentage": 0,
      "quantity": 1,
      "manufacturer": "Tigress Enterprises"
    },
    {
      "base_price": 32.99,
      "discount_percentage": 0,
      "quantity": 1,
      "manufacturer": "Oceanavigations"
    }
  ]
}

PUT test-2/_doc/2
{
  "products": [
    {
      "base_price": 31.14,
      "discount_percentage": 5,
      "quantity": 1,
      "manufacturer": "Virora"
    },
    {
      "base_price": 35.69,
      "discount_percentage": 10,
      "quantity": 3,
      "manufacturer": "Oceanavigations"
    }
  ]
}

```
* step 2: go to `Dashboards management` --> `Index patterns` --> `create index pattern`
<img width="1612" alt="Screenshot 2023-10-25 at 12 49 21" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/79961084/f50862e1-5ee1-4935-a4cb-7d246cf21c8c">

* step 3: enter `test*` to match `test-2`
*  go back to discover. in the index pattern selector (on top of left nav), choose `test*`.

So in new Discover (2.11), if you check the left field nav, you could see we miss the nested field`products`. but the flyout and display are fine. It is just that there is no way customer could select nested field. 

<img width="1788" alt="Screenshot 2023-10-25 at 12 37 29" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/79961084/83b9775d-3eb4-4e0e-bb85-047951f00d85">

But if we do the above steps in 2.9 OSD with legacy discover, we could see `products` in left nav and once click `+` button to select it, you could see multiple lines in each row as customer shown.

<img width="1779" alt="Screenshot 2023-10-25 at 12 13 16" src="https://github.com/opensearch-project/OpenSearch-Dashboards/assets/79961084/31624488-c4f8-4059-b0cd-b7df75a898a4">
[...]

Issue #5208, opened 2023-10-04
[BUG] Multiline content in [message] field is not displayed correctly 


vinylen:
**Describe the bug**

With the new discover UI we are unable to see multiline messages in a coherent way. Multiline messages just gets "stretched" out, never word wrapped. When switching back to the old discover UI, multiline messages gets displayed in a better way.

**To Reproduce**
Have multiline content in a field, toggle that field so that it is shown in the new discover UI

**Expected behavior**
Multiline messages should get displayed in the same way as before, showing a few lines.

**OpenSearch Version**
2.10

**Dashboards Version**
2.10

**Plugins**

opensearch-prometheus-exporter
s3-repository

**Screenshots**
New discover with multiline content, everything is just in a single row.
![image](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/98466471/4e36a220-43fa-4636-bc02-f839a3a00f71)

Old discover with multiline content:
![image](https://github.com/opensearch-project/OpenSearch-Dashboards/assets/98466471/0f5c3605-2100-4cf7-b466-36e7ea6370ea)

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

vinylen:
I can confirm the bug is also present in 2.11 - https://playground.opensearch.org/app/data-explorer/discover#?_a=(discover:(columns:!(products),isDirty:!f,sort:!()),metadata:(indexPattern:'3c72c3e0-6fa3-11ed-a8ec-958f39714b93',view:discover))&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-1y,to:now))&_q=(filters:!(),query:(language:kuery,query:''))

MadaniKK:
Hi, can I take this issue? Thank you

joshuarrrr:
This may need to be fixed in the OUIDataGrid component.

ananzh:
I think the bug is that in new Discover nested field is not correctly shown in left field nav. Here is how to reproduce it:

* step 1: go to dev tools. create an index and add nested field.
```
PUT test-2/_doc/1
{
  "products": [
    {
      "base_price": 28.99,
      "discount_percentage": 0,
      "quantity": 1,
      "manufacturer": "Tigress Enterprises"
    },
    {
      "base_price": 32.99,
      "discount_percentage": 0,
      "quantity": 1,
      "manufacturer": "Oceanavigations"
    }
  ]
}

[...]

PR #4875, opened 2023-08-31
[Data Explorer][Discover 2.0] Fix issues when change index pattern

ananzh:
### Description
* side nav only shows the selected index pattern fields after switch index pattern
Screenshot 2023-08-31 at 09 23 24

* allow reset column state when switch index pattern

There are two conditions: 1) If there is common field, common field will be saved. 
Screenshot 2023-08-31 at 09 23 51

Screenshot 2023-08-31 at 09 23 59

2) if no common field, will show `_source` after switch
Screenshot 2023-08-31 at 09 23 34

Screenshot 2023-08-31 at 09 23 43

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4840 https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4846

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
  - [ ] `yarn test:ftr`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #9856, opened 2025-06-06
[Discover] Adding Back Unknown Fields

sejli:
#### Context
[Object fields](https://docs.opensearch.org/docs/latest/field-types/supported-field-types/object/) are the default way of storing JSON objects in an OpenSearch index. When indexing a document with an object field, the `object` type does not need to be specified. Internally, the objects are flattened and stored as key/value pairs with no relation to other fields within the object.
```
PUT testindex1/_mappings
{
    "properties": {
      "patient": { 
        "properties" :
          {
            "name" : {
              "type" : "text"
            },
            "id" : {
              "type" : "keyword"
            }
          }   
      }
    }
}
```

On the other hand, [nested fields](https://docs.opensearch.org/docs/latest/field-types/supported-field-types/nested/) maintain the relationship between parent and children by storing the nested objects as separate documents. The mapping also has a defined `nested` field type.
```
PUT testindex1
{
  "mappings" : {
    "properties": {
      "patients": { 
        "type" : "nested"
      }
    }
  }
}
```

#### Issues
A bug was noticed in OpenSearch Dashboard's Discover when several object fields would not be displayed. For example, the `opensearch_dashboards_sample_data_ecommerce` index contains an object field out of the box called `products`. By nature of objet fields in OpenSearch, this field does not have a defined type in the mappings. Thus, upon index pattern creation, there is no field created for the object field.

Pasted image 20250606023303

In older versions of Discover, these fields would still be displayed because Discover would treat them as unknown fields. Since the index patterns themselves would not have knowledge of object fields in their fields list, the field counts (a map of the number of times a field appears in a search result) would the existence of these object fields and Discover would display them as unknown. The latest working version of this was in 2.9, where products can be added as a column and seen in the `_source`.

Pasted image 20250606023500

image

Pasted image 20250606025012

Following the development of Discover 2.0 and the de-angular project in the later versions, the inclusion of unknown fields was removed in [this PR](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/4875/files#diff-76d2d253b467eaf62b95207cf6d4125175b1d266fc18963977c0504d0e8b17d5L52) due to a bug where field counts were not being reset before firing a new search. The [newly added use_search](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/discover/public/application/view_components/utils/use_search.ts#L321) from the de-angular project would continue to add field counts to the fetch state, regardless of if the index pattern was switched. In this PR, I reset the field counts prior to firing the search request, so the field counts can be built from a fresh state, as well as add the unknown fields back to the `getIndexPatternFieldList()` function.
Before:
Pasted image 20250606024906
After:
Pasted image 20250606025150
Pasted image 20250606025115

#### Flattened versus Object
Because we removed the unknown fields, issues such as [this (1)](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5400) and [this (2)](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5208) were created to call out that the unknown fields were not being displayed in the left nav. To resolve this, the `flattenHit()` function was updated in [this PR](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/5429) to always set the `deep` parameter to true, which would flatten the entire object field down to individual key/value pairs. This is why in the current Discover, object fields are only shown as a collection of the children fields. To resolve this, I included the original object field in the final output such that both the flattened fields and the original object can be viewed in Discover.

## Changelog

- fix: Adds the displaying of unknown fields back to Discover

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/discover/public/application/components/default_discover_table/table_row.tsx
+++ b/src/plugins/discover/public/application/components/default_discover_table/table_row.tsx
@@ -106,7 +106,7 @@ const TableRowUI = ({
 
         const sanitizedCellValue = dompurify.sanitize(formattedValue);
 
-        if (!fieldInfo?.filterable) {
+        if (fieldInfo?.filterable === false) {
           return (
             <td
               key={colName}
--- a/src/plugins/discover/public/application/components/sidebar/lib/get_index_pattern_field_list.ts
+++ b/src/plugins/discover/public/application/components/sidebar/lib/get_index_pattern_field_list.ts
@@ -28,12 +28,26 @@
  * under the License.
  */
 
-import { IndexPattern } from 'src/plugins/data/public';
+import { difference } from 'lodash';
+import { IndexPattern, IndexPatternField } from 'src/plugins/data/public';
 
 export function getIndexPatternFieldList(
   indexPattern?: IndexPattern,
   fieldCounts?: Record<string, number>
 ) {
   if (!indexPattern || !fieldCounts) return [];
-  return [...indexPattern.fields.getAll()];
+
+  const fieldNamesInDocs = Object.keys(fieldCounts);
+  const fieldNamesInIndexPattern = indexPattern.fields.getAll().map((fld) => fld.name);
+  const unknownTypes: IndexPatternField[] = [];
+
+  difference(fieldNamesInDocs, fieldNamesInIndexPattern).forEach((unknownFieldName) => {
+    unknownTypes.push({
+      displayName: String(unknownFieldName),
+      name: String(unknownFieldName),
+      type: 'unknown',
+    } as IndexPatternField);
+  });
+
+  return [...indexPattern.fields.getAll(), ...unknownTypes];
 }
--- a/src/plugins/discover/public/application/view_components/canvas/discover_table.tsx
+++ b/src/plugins/discover/public/application/view_components/canvas/discover_table.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useMemo } from 'react';
 import { DEFAULT_COLUMNS_SETTING, MODIFY_COLUMNS_ON_SWITCH } from '../../../../common';
 import { DiscoverViewServices } from '../../../build_services';
 import { useOpenSearchDashboards } from '../../../../../opensearch_dashboards_react/public';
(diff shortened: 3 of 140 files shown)


src/plugins/discover/public/application/components/default_discover_table/table_row.tsx
```diff
@@ -106,7 +106,7 @@ const TableRowUI = ({
 
         const sanitizedCellValue = dompurify.sanitize(formattedValue);
 
-        if (!fieldInfo?.filterable) {
+        if (fieldInfo?.filterable === false) {
```
ruanyl, line 109: This changed the behavior which requires `fieldInfo?.filterable` to be strictly `false`, is it by intention?

src/plugins/discover/public/application/components/sidebar/lib/get_index_pattern_field_list.ts
```diff
@@ -28,12 +28,26 @@
...
 export function getIndexPatternFieldList(
   indexPattern?: IndexPattern,
   fieldCounts?: Record<string, number>
 ) {
   if (!indexPattern || !fieldCounts) return [];
-  return [...indexPattern.fields.getAll()];
+
+  const fieldNamesInDocs = Object.keys(fieldCounts);
+  const fieldNamesInIndexPattern = indexPattern.fields.getAll().map((fld) => fld.name);
+  const unknownTypes: IndexPatternField[] = [];
```
ruanyl, line 42: Would `unknownFields` sound a bit more appropriate?

src/plugins/discover/public/application/components/default_discover_table/table_row.tsx
sejli: Yeah, this was by intention. In the case that `!fieldInfo` was undefined, it would load the table row with the incorrect formatting. Specifically, object fields would be displayed as plaintext rather than formatted objects.

src/plugins/discover/public/application/components/sidebar/lib/get_index_pattern_field_list.ts
sejli: I agree, this is just taken from the previous version of `get_index_pattern_field_list.ts`, prior to when this block was removed.

src/plugins/discover/public/application/view_components/canvas/discover_table.tsx
```diff
@@ -65,6 +68,17 @@ export const DiscoverTable = ({ rows, scrollToTop }: Props) => {
     };
   });
   const dispatch = useDispatch();
+
+  useEffect(() => {
```
Maosaic: Can we pass `fieldCounts` from `canvas`? To aviod introduce new side effect subscription.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/discover/public/application/view_components/canvas/index.tsx#L170

Maosaic:
Can you also add this fix to `explore`?

angle943 (changes requested):
because this has Explore related changes, please sync up with @Maosaic before this gets merged!

Maosaic (approved):

PR #9862, opened 2025-06-09
Implement polling for index state in Vended Dashboard progress

Swiddis:
### Description

In addition to the existing progress flow for queries started in OSD, this PR detects already-running queries with a 10s polling interval. Polling only runs if the V.D. banner is rendered, so no perf hit for other dashboards. 

### Issues Resolved

N/A

## Screenshot

N/A

## Testing the changes

Creating a vended dashboard with an auto-refresh process happening in the background. To see the state, you'll need a version of Spark running this PR: https://github.com/opensearch-project/opensearch-spark/pull/1195. If that PR isn't present, the feature doesn't have any effect on the UI.

## Changelog
- feat: Implement polling for index state in Vended Dashboard progress

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


Swiddis:
Test failures from the last run seem unrelated, new tests don't introduce new failures, so probably ready as-is.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.tsx
```diff
@@ -119,13 +136,12 @@ export const DashboardDirectQuerySyncBanner: React.FC<DirectQuerySyncProps> = ({
         </EuiText>
       ) : (
         <EuiCallOut size="s">
-          <EuiLoadingSpinner size="s" />
-
+          <EuiLoadingSpinner size="s" />{' '}
```
RyanL1997: same as the above

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.tsx
```diff
@@ -109,8 +127,7 @@ export const DashboardDirectQuerySyncBanner: React.FC<DirectQuerySyncProps> = ({
                   )}`
                 : '--',
             },
-          })}
-
+          })}{' '}
```
RyanL1997: I think we don't need this `' '` since the scss should handle the margin here:
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.scss

src/plugins/data_source_management/public/constants.ts
```diff
@@ -32,20 +32,65 @@ export const DATACONNECTIONS_UPDATE_STATUS = '/status';
...
-export const MAX_ORD = 100;
+export const ExternalIndexState = {
+  CREATING: 'creating',
+  ACTIVE: 'active',
+  REFRESHING: 'refreshing',
+  RECOVERING: 'recovering',
+  CANCELLING: 'cancelling',
+};
+
+export type Progress = { in_progress: true; percentage: number } | { in_progress: false };
```
RyanL1997: nit: `SyncProgress` probably?

RyanL1997 (commented):
Just dropped some comments on this:

RyanL1997 (approved):
LGTM

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.tsx
```diff
@@ -53,14 +59,28 @@ export const DashboardDirectQuerySyncBanner: React.FC<DirectQuerySyncProps> = ({
     };
 
     loadSyncInfo();
+    const interval = setInterval(loadSyncInfo, 10000);
```
kavilla: nitpick: is 10k the default length it takes? https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/DEVELOPER_GUIDE.md#avoid-magic-numbersstrings

kavilla (dismissed):

kavilla:
lgtm. will verify why the test failures as yeah this seems pretty isolated from all that. is there a label for vended dashboards work that y'all would like for me to create and apply? for easier look ups in the future

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner.tsx
Swiddis: 10 seconds is a convenient-enough refresh, exact value doesn't matter -- happy to do a larger number if it's better, but 10s seems lightweight enough while being decently responsive. I'll push a commit to remove the magic.

Swiddis:
> is there a label for vended dashboards work that y'all would like for me to create and apply? for easier look ups in the future

If it's convenient/preferred for y'all -- don't have a strong preference either way

kavilla (dismissed):

src/plugins/data_source_management/public/constants.ts
```diff
@@ -32,20 +32,65 @@ export const DATACONNECTIONS_UPDATE_STATUS = '/status';
...
-]);
-
-export const MAX_ORD = 100;
+export const ExternalIndexState = {
+  CREATING: 'creating',
+  ACTIVE: 'active',
+  REFRESHING: 'refreshing',
+  RECOVERING: 'recovering',
+  CANCELLING: 'cancelling',
+};
```
joshuali925: nit for better type and guard
```suggestion
} as const;
```

src/plugins/data_source_management/public/constants.ts
```diff
@@ -32,20 +32,65 @@ export const DATACONNECTIONS_UPDATE_STATUS = '/status';
...
+ * "active" state doesn't distinguish between a fresh-activated or older activated index, we also
+ * need to know if the index has been refreshed before.
+ *
+ * @param state An index state, as defined by the Flint State Machine.
+ * @param queryStatus the Direct Query status for any running queries.
+ * @param hasLastRefresh Whether the index has been refreshed before.
+ * @returns A `Progress` value
+ */
+export const asSyncProgress = (
+  state: string | null,
```
joshuali925:
```suggestion
  state: typeof ExternalIndexState[keyof typeof ExternalIndexState] | null,
```

maybe first extract the type somewhere else. using `string` may be hard to maintain later

src/plugins/data_source_management/public/constants.ts
```diff
@@ -32,20 +32,65 @@ export const DATACONNECTIONS_UPDATE_STATUS = '/status';
...
+      return { in_progress: true, percentage: 60 };
+    case ExternalIndexState.RECOVERING:
+      return { in_progress: true, percentage: 60 };
+    case ExternalIndexState.CANCELLING:
+      return { in_progress: true, percentage: 90 };
+    default:
+      // Null state, or other error states
+      return { in_progress: false };
+  }
+};
```
joshuali925: this is not really a constant, can this move to some other file? and does this have unit tests?

src/plugins/data_source_management/public/components/frame_work_test/hooks/drect_query_hook.test.tsx
```diff
@@ -91,7 +98,11 @@ describe('useDirectQuery', () => {
...
     await act(async () => {
-      await result.current.startLoading({ datasource: 'test_source' });
+      await result.current.startLoading({
+        datasource: 'test_source',
+        query: 'SELECT 1',
+        lang: 'sql',
+      });
     });
 
     expect(result.current.loadStatus).toBe(DirectQueryLoadingStatus.FAILED);
```
joshuali925, line 108: i realize it's not from this PR but this file name has a typo, it should be direct_query_hook.test.tsx

src/plugins/data_source_management/public/constants.ts
```diff
@@ -32,20 +32,65 @@ export const DATACONNECTIONS_UPDATE_STATUS = '/status';
...
+  // Query loading status takes precedence if in a processing state, otherwise fallback to state
+  switch (queryStatus) {
+    case DirectQueryLoadingStatus.SUBMITTED:
+      return { in_progress: true, percentage: 0 };
+    case DirectQueryLoadingStatus.SCHEDULED:
+      return { in_progress: true, percentage: 25 };
+    case DirectQueryLoadingStatus.WAITING:
+      return { in_progress: true, percentage: 50 };
+    case DirectQueryLoadingStatus.RUNNING:
+      return { in_progress: true, percentage: 75 };
```
joshuali925: curious I think the running state takes the most time, will it stuck at 75% to the user?

src/plugins/data_source_management/public/constants.ts
```diff
@@ -32,20 +32,65 @@ export const DATACONNECTIONS_UPDATE_STATUS = '/status';
...
+ * @param queryStatus the Direct Query status for any running queries.
+ * @param hasLastRefresh Whether the index has been refreshed before.
+ * @returns A `Progress` value
+ */
+export const asSyncProgress = (
+  state: string | null,
+  queryStatus: DirectQueryLoadingStatus | null,
+  hasLastRefresh: boolean
+): SyncProgress => {
+  // Query loading status takes precedence if in a processing state, otherwise fallback to state
```
joshuali925: why is there a status plus a state? can they be combined?

src/plugins/data_source_management/public/constants.ts
Swiddis: Depends on the query, but yeah -- adding more granular progress tracking for mid-running was out of scope. Originally I wanted to just show the state names themselves instead of a percentage, but it was decided that percentage was easier to grok even if less descriptive/accurate.

src/plugins/data_source_management/public/constants.ts
Swiddis: https://github.com/opensearch-project/opensearch-spark/blob/main/docs/index.md#index-state-transition vs https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/job-states.html, we prefer the second when it's available (since it's more granular) and fallback to the first if it's not.

We only have access to the EMR job when we've started the query ourselves, there's currently no infrastructure to recover the job ID if a job starts in the background (e.g. due to auto-schedule). Index state is always available but has less information, it essentially equates to "is EMR `running` or not?".

src/plugins/data_source_management/public/constants.ts
Swiddis: Index states here aren't guaranteed to be comprehensive longer-term (we have plans to add more), handling a `default` case is important. Does Typescript have anything similar to a [`non_exhaustive` attribute](https://doc.rust-lang.org/reference/attributes/type_system.html) for cases like this?

Swiddis: I moved ExternalIndexState to be an enum type and represented this as `ExternalIndexState | string`.

kavilla (approved):

Issue #3008, opened 2022-12-02
Questions about CVEs, dependencies, and overall project prioritization


tmarkley:
### CVEs/Dependencies
I've noticed that recent CVEs are addressed by lazily adding a manual dependency resolution to the `package.json`.

Examples: 
* https://github.com/opensearch-project/OpenSearch-Dashboards/pull/2640
* https://github.com/opensearch-project/OpenSearch-Dashboards/pull/2454
* https://github.com/opensearch-project/OpenSearch-Dashboards/pull/1552

The "guidelines" introduced in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/2674 are insufficient and are contributing to these poor practices. In some cases (i.e. the resolution for `d3-color`), there is no path forward without significant refactoring or rewriting. However, I don't see issues being opened to address that work. That work is currently being punted until it causes more problems.

This is unsustainable and jeopardizes the future velocity of this project. There are already too many dependencies scattered across Dashboards, sub-packages, and plugins, and temporary band-aids make the problems worse. See https://github.com/opensearch-project/OpenSearch-Dashboards/issues/1298 for more details. It's often the case that if you dive into the upstream dependencies of a library that needs to be upgraded you'll find a path to success.

### Technical Debt and Prioritization
Diving into these dependency problems (as well as a fresh perspective after taking a few months off) led me to take a step back and look at the broader approach to development and changes in the OpenSearch Dashboards project.

From a very high level I think new features and improvements (e.g. Multiple Data Sources, Visualization Builder, Extensions) are being prioritized over important technical debt projects such as #2875, #1660, #1560, #1060, https://github.com/opensearch-project/OpenSearch-Dashboards/issues/1625, https://github.com/opensearch-project/OpenSearch-Dashboards/issues/2032, and [many more](https://github.com/opensearch-project/OpenSearch-Dashboards/labels/technical%20debt) while the group of active maintainers and community members is not large (about a dozen by my count?).

Important upcoming initiatives such as #2929 are going to require a lot of this technical debt to be addressed. As we've learned in the past, these projects always require more effort than initially estimated.

### Summary 
Overall I would say that the state of the code base, lack of sufficient documentation, and potential security concerns caused by the current dependency tree should lead the majority of contributors on OpenSearch Dashboards to focus on improving the health of the project. The community might not initially understand the importance of this work, but I'm sure you fine folks at AWS can write up a tidy document to explain that. :)

[...]

PR #9866, opened 2025-06-11
Add dependabot workflow to monitor and upgrade outdated dependencies

cwperks:
### Description

Taken from https://github.com/opensearch-project/opensearch-js/pull/905, this PR adds a dependabot workflow that runs weekly to automatically update any outdated dependencies in the package.json file

**Note**: This only monitors the root package.json file. This should be extended in another PR to monitor all of the packages in the `packages` folder.

### Issues Resolved

Related to https://github.com/opensearch-project/OpenSearch-Dashboards/issues/3008

## Changelog

chore: Add dependabot workflow to automatically monitor and upgrade outdated dependencies

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: /
+    open-pull-requests-limit: 3
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependabot"
+      - "dependencies"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


ruanyl:
Agree on keeping the dependencies up-to-date, but we run into plugin build issues several times even when bumping a single package version in core, did you consider such case? cc @AMoo-Miki

cwperks:
> Agree on keeping the dependencies up-to-date, but we run into plugin build issues several times even when bumping a single package version in core, did you consider such case? cc @AMoo-Miki

Would you be able to point to any examples?

Ideally we have a version catalog for common versions used by OSD Core + all plugins. My understanding is that the root level of the `package.json` is that version catalog.

For plugin (or package) specific dependencies those should be only in the set of plugins that use that particular dependency.

In OpenSearch we also have a version catalog to keep versions consistent across plugins even if OS Core does not have a direct dependency.

We can also configure dependabot to skip major updates as those often need other considerations.

PR #9874, opened 2025-06-12
[Explore] add pie, scatter, heatmap, single metric auto vis

Qxisylolo:
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+  const textEncoding: any = {
+    field: color,
+    format: '.1f',
+  };
+
+  if (!isRegular && styles.label.type !== LabelAggregationType.NONE && styles.label.type) {
+    textEncoding.aggregate = styles.label.type;
+  }
+  const labelLayer: any = {
```
FriedhelmWS: Nit: could we remove the use of any in the production code but define proper type for these objects?

src/plugins/explore/public/components/visualizations/heatmap/heatmap_exclusive_vis_options.tsx
```diff
@@ -0,0 +1,237 @@
...
+            {shouldShowType && (
+              <EuiFlexItem>
+                <EuiFormRow
+                  label={i18n.translate('explore.stylePanel.heatmap.label.type', {
+                    defaultMessage: 'Type',
+                  })}
+                >
+                  <EuiSelect
+                    value={styles.type}
+                    onChange={(e) => updateLabelOption('type', e.target.value as any)}
```
FriedhelmWS: Nit: do we need to the cast to any here?

src/plugins/explore/public/components/visualizations/heatmap/heatmap_vis_options.tsx
```diff
@@ -0,0 +1,159 @@
...
+}) => {
+  const shouldShowType = numericalColumns.length === 3;
+  const updateStyleOption = <K extends keyof HeatmapChartStyleControls>(
+    key: K,
+    value: HeatmapChartStyleControls[K]
+  ) => {
+    onStyleChange({ [key]: value });
+  };
+
+  function interAxesFromColumns(
```
FriedhelmWS: Minor question: why not use arrow function here?

src/plugins/explore/public/components/visualizations/scatter/scatter_vis_options.tsx
```diff
@@ -0,0 +1,150 @@
...
+      }
+      if (axis.axisRole === AxisRole.Y) {
+        return { ...axis, field: y };
+      }
+      return axis;
+    });
+
+    updateStyleOption('StandardAxes', axesWithFields);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [numericalColumns, categoricalColumns, dateColumns]);
```
FriedhelmWS, line 56: This useEffect seems to be exactly the same one we used in `heatmap_vis_options.tsx`. Is that possible for us to create a custom hook base on this useEffect to reduce duplicate code?

src/plugins/explore/public/components/visualizations/heatmap/to_expression.ts
```diff
@@ -0,0 +1,154 @@
...
+      },
+      y: {
+        field: yAxis?.field?.default.column,
+        type: 'nominal',
+        axis: applyAxisStyling(yAxis),
+      },
+      color: {
+        field: numericalColor.column,
+        type: 'quantitative',
+        // log doesn't support bin, delete log or disable bin when scale type is log?
```
FriedhelmWS: Is this a open question for later?

src/plugins/explore/public/components/visualizations/metric/metric_vis_options.tsx
```diff
@@ -0,0 +1,154 @@
...
+                label={i18n.translate('explore.vis.metric.fontSize', {
+                  defaultMessage: 'Font Size',
+                })}
+              >
+                <EuiFieldNumber value={fontSize} onChange={(e) => handleFontSize(e.target.value)} />
+              </EuiFormRow>
+            </EuiFlexItem>
+            <EuiFlexItem>
+              <EuiSwitch
+                label="Use color for font color"
```
FriedhelmWS: Do we still need i18n for this label?

src/plugins/explore/public/components/visualizations/pie/pie_vis_config.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import { PieVisStyleControls, PieChartStyleControlsProps } from './pie_vis_options';
+import { Positions } from '../types';
+
+export interface PieExclusiveStyleControl {
+  donut: boolean;
+  showValues: boolean;
+  showLabels: boolean;
+  truncate: number;
+  // Todo:  shall we support 2 layers pie chart?
```
FriedhelmWS, line 17: Will we support this later?

src/plugins/explore/public/components/visualizations/pie/to_expression.ts
```diff
@@ -0,0 +1,85 @@
...
+export const createPieSpec = (
+  transformedData: Array<Record<string, any>>,
+  numericalColumns: VisColumn[],
+  categoricalColumns: VisColumn[],
+  dateColumns: VisColumn[],
+  styleOptions?: Partial<PieChartStyleControls>
+) => {
+  const numericFields = numericalColumns.map((item) => item.column)[0];
+  const numericNames = numericalColumns.map((item) => item.name)[0];
+  const categoryField = categoricalColumns.map((item) => item.column)[0];
```
FriedhelmWS: Is the mapping here necessary? Could we just simply do:
```
const numericFields = numericalColumns[0]?.column;
const numericNames = numericalColumns[0]?.name;
const categoryField = categoricalColumns[0]?.column;
```

src/plugins/explore/public/components/visualizations/scatter/scatter_exclusive_vis_options.tsx
```diff
@@ -0,0 +1,83 @@
...
+                { value: 'diamond', text: 'Diamond' },
+              ]}
+              value={styles.pointShape}
+              onChange={(e) => updateStyle('pointShape', e.target.value as PointShape)}
+            />
+          </EuiFormRow>
+        </EuiFlexItem>
+        <EuiFlexItem>
+          <EuiSwitch
+            label="Filled"
```
FriedhelmWS: We may need i18n here

src/plugins/explore/public/components/visualizations/style_panel/standard_axes_options.tsx
```diff
@@ -0,0 +1,292 @@
...
+  standardAxes,
+  onStandardAxesChange,
+  disableGrid,
+}) => {
+  const updateAxis = (index: number, updates: Partial<StandardAxes>) => {
+    const updatedAxes = [...standardAxes];
+    updatedAxes[index] = {
+      ...updatedAxes[index],
+      ...updates,
+    };
```
FriedhelmWS, line 51: Do we have to shallow copy the entire array and then do mutation? Can we do something like:
```
const updatedAxes = standardAxes.map((axis, i) =>
  i === index ? { ...axis, ...updates } : axis
);
```

src/plugins/explore/public/components/visualizations/style_panel/standard_axes_options.tsx
```diff
@@ -0,0 +1,292 @@
...
+                        );
+                        if (selectedColumn) {
+                          updateAxis(index, {
+                            field: {
+                              ...axis.field,
+                              default: selectedColumn,
+                            },
+                          });
+                        }
+                      }}
```
FriedhelmWS, line 94: Nit: I think it would be nice to have a standalone function for these logic.

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
```diff
@@ -0,0 +1,125 @@
...
+import { HeatmapChartStyleControls } from './heatmap_vis_config';
+import { generateColorBySchema } from '../utils/utils';
+
+// isRegular=== true refers to 2 dimension and 1 metric heatmap.
+export const createlabelLayer = (
+  styles: Partial<HeatmapChartStyleControls>,
+  isRegular: boolean,
+  xAxis?: VisColumn,
+  yAxis?: VisColumn,
+  color?: string
```
ruanyl: 1. Is the name `color` appropriate? should be `field`? 
2. Is this field mandatory?

src/plugins/explore/public/components/visualizations/style_panel/standard_axes_options.tsx
```diff
@@ -0,0 +1,292 @@
...
+                              ]
+                            : [
+                                { value: 'bottom', text: 'Bottom' },
+                                { value: 'top', text: 'Top' },
+                              ]
+                        }
+                      />
+                    </EuiFormRow>
+                  </EuiFlexItem>
+                </EuiFlexGroup>{' '}
```
FriedhelmWS: Empty space here 👀

src/plugins/explore/public/components/visualizations/style_panel/standard_axes_options.tsx
```diff
@@ -0,0 +1,292 @@
...
+          <EuiTitle size="xs">
+            <h3>
+              {i18n.translate('explore.vis.standardAxes.title', {
+                defaultMessage: 'Axes',
+              })}
+            </h3>
+          </EuiTitle>
+        </EuiFlexItem>
+        <EuiFlexItem grow={false}>
+          <EuiButton onClick={() => onChangeSwitchAxes(standardAxes)}>Switch X and Y</EuiButton>
```
FriedhelmWS: i18n 👀

src/plugins/explore/public/components/visualizations/types.ts
```diff
@@ -3,8 +3,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { Positions } from './utils/collections';
+// import { Positions } from './utils/collections';
```
FriedhelmWS: Code commented out 👀

src/plugins/explore/public/components/visualizations/style_panel/standard_axes_options.tsx
```diff
@@ -0,0 +1,292 @@
...
+      ...updates,
+    };
+
+    onStandardAxesChange(updatedAxes);
+  };
+
+  return (
+    <>
+      {standardAxes.map((axis, index) => {
+        const isYAxis = axis.axisRole === 'y';
```
FriedhelmWS: Minor: can we use the defined enum AxisRole.Y instead of string? Also for some other place that uses string 'x'/'y' in this PR.

src/plugins/explore/public/components/visualizations/visualization_registry.ts
```diff
@@ -61,6 +65,7 @@ export class VisualizationRegistry {
 
     for (const rule of this.rules) {
       if (rule.matches(numericalColumns, categoricalColumns, dateColumns)) {
+        console.log('rule', numericalColumns, categoricalColumns, dateColumns);
```
FriedhelmWS: Is this for logging purpose?

FriedhelmWS:
There are also a few lint error and test failure, could you please have a look?

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
```diff
@@ -0,0 +1,125 @@
...
+) => {
+  if (!styles.label?.show) {
+    return null;
+  }
+  const textEncoding: any = {
```
ruanyl:
```typescript
import type { Encoding } from 'vega-lite/build/src/encoding';

  const textEncoding: Encoding<string>['text'] = {
    field: color,
    format: '.1f',
  };
```

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
ruanyl: In this case, I think we can simply remove `any` and let typescript infer the type

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
```diff
@@ -0,0 +1,125 @@
...
+    },
+  };
+  return labelLayer;
+};
+
+export const getDataBound = (
+  transformedData: Array<Record<string, any>>,
+  numericalColor: string
+): number[] => {
+  if (!transformedData) return [];
```
ruanyl: Seems `transformedData` is nullable, should you update the type definition of `transformedData` on line 53

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
```diff
@@ -0,0 +1,125 @@
...
+      text: textEncoding,
+      angle: { value: styles.label.rotate ? 45 : 0 },
+    },
+  };
+  return labelLayer;
+};
+
+export const getDataBound = (
+  transformedData: Array<Record<string, any>>,
+  numericalColor: string
```
ruanyl: Did you mean `numericalField`?
```suggestion
  numericalField: string
```

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
```diff
@@ -0,0 +1,125 @@
...
+        as: numericFields,
+      },
+    ];
+  }
+  return [];
+};
+
+export const setRange = (styles: Partial<HeatmapChartStyleControls>) => {
+  const ranges = styles?.exclusive?.customRanges ?? [];
+  // always make sure the range lenghth >=2 to interpolate
```
ruanyl:
```suggestion
  // always make sure the range length >=2 to interpolate
```

ruanyl: According to the implementation, it only makes sure the value `>=1`, isn't it?

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
```diff
@@ -0,0 +1,125 @@
...
+  const colors = generateColorBySchema(ranges.length + 1, styles?.exclusive?.colorSchema!);
```
ruanyl: Instead of non-null assertion, prefer this

```suggestion
  const colors = generateColorBySchema(ranges.length + 1, styles?.exclusive?.colorSchema ?? ColorSchemas.BLUES);
```

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
```diff
@@ -0,0 +1,125 @@
...
+  return colors;
+};
+
+export const enhanceStyle = (
+  markLayer: any,
```
ruanyl, line 92: Avoid `any`, but I think we can fix the vega-lite spec TS types in the next PR

src/plugins/explore/public/components/visualizations/scatter/scatter_vis_options.tsx
Qxisylolo: I think is fine here as rules is changing with different charts type. so it is better for each charts to handle its own logic of inferring axes.

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
Qxisylolo: you can see vege is just a plain object , so it is unnecessary to  define a separate type

src/plugins/explore/public/components/visualizations/heatmap/heatmap_exclusive_vis_options.tsx
ruanyl:
```suggestion
                    onChange={(e) => updateLabelOption('type', e.target.value as LabelAggregationType)}
```

src/plugins/explore/public/components/visualizations/heatmap/to_expression.ts
Qxisylolo: open question

src/plugins/explore/public/components/visualizations/pie/pie_vis_config.ts
Qxisylolo: not sure

src/plugins/explore/public/components/visualizations/pie/to_expression.ts
Qxisylolo: I don’t see much of a difference here as numericalColumns here should be just an array with one item(But length will be 2 if we need to implement 2 layers pie). 
And I think the previous approach makes more sense, because we consistently treat the first field as the default when displaying data

src/plugins/explore/public/components/visualizations/heatmap/heatmap_exclusive_vis_options.tsx
```diff
@@ -0,0 +1,237 @@
...
+    key: K,
+    value: HeatmapChartStyleControls['exclusive'][K]
+  ) => {
+    onChange({
+      ...styles,
+      [key]: value,
+    });
+  };
+  const colorSchemas = getColorSchemas();
+  const scaleTypes = getScaleType();
```
ruanyl, line 48: Nit: does it necessarily to be a function? Can simply be a constant? 
```
const COLOR_SCHEMA = Object.freeze([...])
```

src/plugins/explore/public/components/visualizations/heatmap/heatmap_vis_options.tsx
```diff
@@ -0,0 +1,159 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React, { useEffect } from 'react';
+import { EuiTabbedContent, EuiTabbedContentTab } from '@elastic/eui';
+import { i18n } from '@osd/i18n';
+import { HeatmapChartStyleControls } from './heatmap_vis_config';
+import { GeneralVisOptions } from '../style_panel/general_vis_options';
+import { VisColumn, FieldSetting, StandardAxes, AxisRole, Positions } from '../types';
```
ruanyl:
```suggestion
import { VisColumn, FieldSetting, StandardAxes, AxisRole } from '../types';
```

src/plugins/explore/public/components/visualizations/heatmap/heatmap_vis_options.tsx
ruanyl: I suggest to move this function out of the component

src/plugins/explore/public/components/visualizations/heatmap/heatmap_vis_options.tsx
```diff
@@ -0,0 +1,159 @@
...
+        return { ...axis, field: y };
+      }
+      return axis;
+    });
+
+    updateStyleOption('StandardAxes', axesWithFields);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [numericalColumns, categoricalColumns, dateColumns]);
+
+  const handleChangeSwitchAxes = (axes: StandardAxes[]) => {
```
ruanyl:
```suggestion
  const handleSwitchAxes = (axes: StandardAxes[]) => {
```

src/plugins/explore/public/components/visualizations/heatmap/heatmap_vis_options.tsx
```diff
@@ -0,0 +1,159 @@
...
+    {
+      id: 'axes',
+      name: i18n.translate('explore.vis.heatmapChart.tabs.axes', {
+        defaultMessage: 'Axes',
+      }),
+      content: (
+        <AllAxesOptions
+          disableGrid={true}
+          standardAxes={styleOptions.StandardAxes}
+          onChangeSwitchAxes={handleChangeSwitchAxes}
```
ruanyl:
```suggestion
          onSwitchAxes={handleChangeSwitchAxes}
```

src/plugins/explore/public/components/visualizations/metric/metric_vis_options.tsx
Qxisylolo: added and address other i18n comments, thanks for pointing  out

src/plugins/explore/public/components/visualizations/pie/to_expression.ts
FriedhelmWS: Ok, then maybe we should change the name to singular word, otherwise it seems it bit weird if we call a single value numericFields/numericNames.

src/plugins/explore/public/components/visualizations/heatmap/to_expression.ts
FriedhelmWS: Should we keep open question in the comment?

src/plugins/explore/public/components/visualizations/pie/to_expression.ts
```diff
@@ -0,0 +1,85 @@
...
+    encoding: {
+      text: {
+        field: numericFields,
+        type: 'nominal',
+      },
+    },
+  };
+
+  const baseSpec = {
+    $schema: 'https://vega.github.io/schema/vega-lite/v5.json',
```
ruanyl: Nit: use a constant variable for the schema link?

src/plugins/explore/public/components/visualizations/scatter/scatter_vis_options.tsx
```diff
@@ -0,0 +1,150 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React, { useEffect } from 'react';
+import { EuiTabbedContent, EuiTabbedContentTab } from '@elastic/eui';
+import { i18n } from '@osd/i18n';
+import { ScatterChartStyleControls } from './scatter_vis_config';
+import { GeneralVisOptions } from '../style_panel/general_vis_options';
+import { VisColumn, FieldSetting, AxisRole, StandardAxes, Positions } from '../types';
```
ruanyl:
```suggestion
import { VisColumn, FieldSetting, AxisRole, StandardAxes } from '../types';
```

src/plugins/explore/public/components/visualizations/scatter/scatter_vis_options.tsx
```diff
@@ -0,0 +1,150 @@
...
+  dateColumns = [],
+}) => {
+  const updateStyleOption = <K extends keyof ScatterChartStyleControls>(
+    key: K,
+    value: ScatterChartStyleControls[K]
+  ) => {
+    onStyleChange({ [key]: value });
+  };
+
+  function interAxesFromColumns(
```
ruanyl: Extract this function out of the component

ruanyl: btw, what does the function name stands for?

src/plugins/explore/public/components/visualizations/types.ts
```diff
@@ -98,3 +102,74 @@ export interface ValueAxis {
...
+  default: VisColumn;
+  options?: VisColumn[];
+}
+
+export enum AxisRole {
+  X = 'x',
+  Y = 'y',
+}
+
+export enum AxisPosition {
```
ruanyl: Seems this is not used?

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
Qxisylolo: yes, I will update it to colorField as there can be 3 numerical fields.

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
Qxisylolo: 1. yes, I will update it to colorField
2. yes, I will remove ?

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
Qxisylolo: yes. but we will only consider the case when user actually adds a range, so the length should be >=2. I will update the comments so it won't be confusing.
```
  if (
    styles.exclusive?.useCustomRanges &&
    styles.exclusive?.customRanges?.length &&
    styles.exclusive?.customRanges?.length >= 1
  ) {
    const customRanges = styles.exclusive?.customRanges;

    const [min, max] = getDataBound(transformedData, colorField);

    // identify min and max to set domain
    const domainMin = Math.min(...customRanges.map((r) => r.min ?? min));
    const domainMax = Math.max(...customRanges?.map((r) => r.max ?? max));
    // overwrite color scale type to quantize to map continuous domains to discrete output ranges
    markLayer.encoding.color.scale.type = 'quantize';
    markLayer.encoding.color.scale.domain = [domainMin, domainMax];

    const range = setRange(styles);

    markLayer.encoding.color.scale.range = range;
  }
```

src/plugins/explore/public/components/visualizations/types.ts
Qxisylolo: yes, i delete it in the upcoming commits

src/plugins/explore/public/components/visualizations/scatter/scatter_vis_options.tsx
Qxisylolo: sorry this name is misspelled. it should be inferAxesFromColumns.

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
Qxisylolo: ok

src/plugins/explore/public/components/visualizations/style_panel/standard_axes_options.tsx
Qxisylolo: updated, thanks

src/plugins/explore/public/components/visualizations/pie/to_expression.ts
Qxisylolo: ok updated

abbyhu2000:
for rule 2, i see legend vis controls on the style panel, but i do not see any legends in the graph.
Screenshot 2025-06-12 at 5 41 43 PM

src/plugins/explore/public/components/visualizations/heatmap/to_expression.ts
abbyhu2000: we have been using // TODO: to indicate open question for later in the code base

abbyhu2000:
i see for oneMetricOneCateRule, in the code we map to pie, bar, line, area. But on the mapping document, the order is bar, pie, point, line, area. 

which one is the final order?

src/plugins/explore/public/components/visualizations/types.ts
abbyhu2000, line 1: should we add a chart type enum since now we are adding a lot more chart types?
```
export enum ChartType {
  HEATMAP = 'heatmap',
  PIE = 'pie',
  BAR = 'bar',
  LINE = 'line',
  AREA = 'area',
  METRIC = 'metric',
  SCATTER = 'scatter',
  // Add more as needed
}
```

src/plugins/explore/public/components/visualizations/style_panel/custom_ranges.tsx
abbyhu2000, line 1: Should we add unit tests for the new added files in the style panel folder?

abbyhu2000:
I think we missing unit tests for quite a few new files, but otherwise look good to me!

ruanyl (dismissed):

Qxisylolo:
> for rule 2, i see legend vis controls on the style panel, but i do not see any legends in the graph. Screenshot 2025-06-12 at 5 41 43 PM

updated, thanks
截屏2025-06-16 14 11 06

src/plugins/explore/public/components/visualizations/heatmap/to_expression.ts
Qxisylolo: I think it was a random event, and I've already deleted the comment

Qxisylolo:
> i see for oneMetricOneCateRule, in the code we map to pie, bar, line, area. But on the mapping document, the order is bar, pie, point, line, area.
> 
> which one is the final order?

the final order should be bar pie line area, when bar is implemented, it will take the highest priority

src/plugins/explore/public/components/visualizations/style_panel/custom_ranges.tsx
Qxisylolo: yes, I have added some, thanks

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
```diff
@@ -0,0 +1,155 @@
...
+    },
+  };
+  return labelLayer;
+};
+
+export const getDataBound = (
+  transformedData: Array<Record<string, any>>,
+  colorField: string
+): number[] => {
+  const values = transformedData.map((d) => Number(d[colorField])).filter((v) => !isNaN(v));
```
yubonluo, line 60: Question: will there be an infinite / -infinite result?

FriedhelmWS (approved):

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
```diff
@@ -0,0 +1,155 @@
...
+};
+
+export const addTransform = (styles: Partial<HeatmapChartStyleControls>, numericFields: string) => {
+  if (styles?.exclusive?.percentageMode) {
+    return [
+      {
+        joinaggregate: [{ op: 'max', field: numericFields, as: 'max_value' }],
+      },
+      {
+        calculate: `(datum["${numericFields}"] / datum.max_value)`,
```
yubonluo: Question:  Whether to avoid the situation of `datum.max_value` === 0?

src/plugins/explore/public/components/visualizations/types.ts
Qxisylolo: I can include it in following pr

abbyhu2000 (dismissed):

abbyhu2000:
For metric vis, i think we only render a metric visualization when there is only 1 numerical field which results in a single value.
For the case: only 1 numerical field but not a single value, i don’t think metric viz makes sense. For example, 
`source = opensearch_dashboards_sample_data_ecommerce | fields products.base_price` which only has 1 numerical field but resulted in a series of values, the rendered vis looks like all those values rendering on top of one another. 
i think table might make more sense in this case?

src/plugins/explore/public/components/visualizations/rule_repository.ts
```diff
@@ -183,6 +192,197 @@ const oneMetricTwoCateOneDateRule: VisualizationRule = {
...
+    chartType = 'heatmap'
+  ) => {
+    return createRegularHeatmap(transformedData, numericalColumns, styleOptions);
+  },
+};
+
+const oneMetricOneCateRule: VisualizationRule = {
+  id: 'one-metric-one-category',
+  name: 'one metric and one category',
+  description: 'Heatmap for one metric and one category',
```
abbyhu2000: wrong description?

src/plugins/explore/public/components/visualizations/rule_repository.ts
```diff
@@ -183,6 +192,197 @@ const oneMetricTwoCateOneDateRule: VisualizationRule = {
...
+  chartTypes: [{ type: 'heatmap', priority: 100, name: 'Heatmap' }],
+  toExpression: (
+    transformedData,
+    numericalColumns,
+    categoricalColumns,
+    dateColumns,
+    styleOptions,
+    chartType = 'heatmap'
+  ) => {
+    return createRegularHeatmap(transformedData, numericalColumns, styleOptions);
```
abbyhu2000, line 230: does this call the wrong function? this is for one numerical column and two categorical columns, but the called function is createRegularHeatmap which only accepts numerical columns as props

Qxisylolo: no, it doesn't, this is the right function to call. In createRegularHeatmap, we only need numericalColumns since we infer categorical fields with categorical columns in vis control options. they will be stored in standardAxes.
```
export function inferAxesFromColumns(
  numerical?: VisColumn[],
  categorical?: VisColumn[]
): { x: FieldSetting | undefined; y: FieldSetting | undefined } {
  if (numerical?.length === 3) {
    return {
      x: {
        default: numerical[0],
        options: numerical,
      },
      y: {
        default: numerical[1],
        options: numerical,
      },
    };
  }
  if (numerical?.length === 1 && categorical?.length === 2) {
    return {
      x: {
        default: categorical[0],
      },
      y: { default: categorical[1] },
    };
  }
  return { x: undefined, y: undefined };
}

```

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
Qxisylolo: I don't think such a case will happen, as transformedData is the result searched with a query.  In real-world scenarios, NaN is more likely to appear

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
Qxisylolo: Yes, `datum.max_value === 0` is the case we should pay attention to, thanks for pointing out. We can retain the origin value by:
```
calculate: `(datum.max_value === 0 ? datum["${numericFields}"] : datum["${numericFields}"] / datum.max_value)`,
```
Thanks!

Qxisylolo:
> products.base_price

Thanks for pointing that out. We will count validValues and uniqueValues.
 validValuesCount===1:
截屏2025-06-17 10 54 42

 validValuesCount>1:
截屏2025-06-17 10 54 22

src/plugins/explore/public/components/visualizations/rule_repository.ts
Qxisylolo: fixed, thanks

ruanyl (approved):

src/plugins/explore/public/components/visualizations/heatmap/heatmap_chart_utils.ts
```diff
@@ -0,0 +1,155 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { Encoding } from 'vega-lite/build/src/encoding';
+import { LabelAggregationType, VisColumn, ColorSchemas, FieldSetting } from '../types';
+import { HeatmapChartStyleControls } from './heatmap_vis_config';
+import { generateColorBySchema } from '../utils/utils';
+
+// isRegular=== true refers to 2 dimension and 1 metric heatmap.
+export const createlabelLayer = (
```
SuZhou-Joe, line 12: Nit: createLabelLayer

ruanyl (approved):

SuZhou-Joe (approved):

PR #9891, opened 2025-06-13
feat(docs): ensure `.md` files are git tracked when generating sidebar

joshuali925:
### Description

This PR makes the sidebar docs generation to ignore markdown files that are not git tracked. Additionally it ignores lighthouse reports in `/.lighthouseci/` directory.

### Issues Resolved

I might have local markdown files that are not tracked by git, it won't make sense to put those in [docs/_sidebar](https://github.com/opensearch-project/opensearch-dashboards/blob/main/docs/_sidebar.md).

The sidebar generation only includes `.md` and it does not pick up `.MD`, `.markdown`, `.MARKDOWN`, etc. This is existing behavior.

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/scripts/generate_docs_sidebar.js
+++ b/scripts/generate_docs_sidebar.js
@@ -6,6 +6,7 @@
 
 const fs = require('fs');
 const path = require('path');
+const { execSync } = require('child_process');
 
 const startDirs = [
   {
@@ -27,6 +28,18 @@ const startDirs = [
 
 const sidebarFile = './docs/_sidebar.md'; // Location to save the generated sidebar
 const excludeDirs = ['node_modules', '.git']; // Directories to exclude from the search
+
+const isGitTrackedMarkdownFile = (() => {
+  let tracked;
+  return (filePath) => {
+    if (tracked === undefined) {
+      tracked = new Set(
+        execSync(`git ls-files '*.md'`).toString().trim().split('\n').filter(Boolean)
+      );
+    }
+    return tracked.has(filePath);
+  };
+})();
 
 // Function to recursively find Markdown files and return a nested structure
 function findMarkdownFiles(dir, prefix = '', baseDir = '', recursively = true) {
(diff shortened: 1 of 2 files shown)


scripts/generate_docs_sidebar.js
```diff
@@ -28,6 +29,18 @@ const startDirs = [
 const sidebarFile = './docs/_sidebar.md'; // Location to save the generated sidebar
 const excludeDirs = ['node_modules', '.git']; // Directories to exclude from the search
 
+const isGitTrackedMarkdownFile = (() => {
+  let tracked;
+  return (filePath) => {
+    if (tracked === undefined) {
+      tracked = new Set(
+        execSync(`git ls-files '*.md'`).toString().trim().split('\n').filter(Boolean)
```
kavilla, line 37: Should this check the status? Like 


```suggestion
        execSync(`git status -s '*.md'`)
```

And if the file is unstaged it will have a space as the first character. In other words:
A file is staged if the first letter is not a space.

That way once we stage the markdown file and trying to commit then it will generate

kavilla (commented):
Thank you so much for doing this!!!! I keep ignoring this commit hook. So this will make life better 

Just a comment tho. If we should use git status instead

scripts/generate_docs_sidebar.js
joshuali925: The script builds the sidebar using all md files, not just the modified ones in the current commit. `git status -s` only returns information in the git working tree and staging area, so if I see a `a.md` file on disk but it's not part of the `git status` output, I won't be able to tell if it's already tracked by git and should be included in the sidebar, or if it's part of `.gitignore` and should be excluded.

If you do `git add new.md; git rm existing.md`, `ls-files` will include `new.md` but not `existing.md`, so this should work properly. I think we can directly use `git ls-files` results to build the sidebar instead of manual traversal, but there is some logic for path prefix and not really any benefit to refactor this

(BTW `git status --porcelain` works better if you ever need it in scripts, and `git diff --name-only --cached --diff-filter=A` returns the files added in staging area)

PR #9902, opened 2025-06-16
[Explore] Support logs, traces, metrics flavor for explore plugin

joshuali925:
### Description

This PR
- duplicates `app.tsx` into 3 pages: logs, traces, metrics. Currently they are identical, but as features are built they will diverge.
- Registers the 3 flavors to side nav

## Changelog
- feat: support logs, traces, metrics flavor for explore plugin

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/9920.yml
+++ b/changelogs/fragments/9920.yml
@@ -1,0 +1,2 @@
+feat:
+- Add bar chart with style panel ([#9920](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9920))
--- a/changelogs/fragments/9932.yml
+++ b/changelogs/fragments/9932.yml
@@ -1,0 +1,2 @@
+fix:
+- Fix histogram UI in Explore ([#9932](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9932))
--- a/src/plugins/explore/public/application/components/header_dataset_selector.tsx
+++ b/src/plugins/explore/public/application/components/header_dataset_selector.tsx
@@ -14,7 +14,7 @@ import {
   beginTransaction,
   finishTransaction,
 } from '../utils/state_management/actions/transaction_actions';
-import { setDataset, setQuery } from '../utils/state_management/slices/query_slice';
+import { setQuery } from '../utils/state_management/slices/query_slice';
 
 export interface HeaderDatasetSelectorProps {
   datasetSelectorRef: React.RefObject<HTMLDivElement>;
@@ -46,13 +46,8 @@ export const HeaderDatasetSelector: React.FC<HeaderDatasetSelectorProps> = ({
 
       dispatch(beginTransaction());
       try {
-        // EXPLICIT cache clear - separate cache logic
         dispatch(clearResults());
-
-        // Update dataset
         dispatch(setQuery(queryStringState));
-
-        // Execute queries - cache already cleared
         dispatch(executeQueries({ services }) as any);
       } finally {
         dispatch(finishTransaction());
--- a/src/plugins/explore/public/application/components/query_panel.tsx
+++ b/src/plugins/explore/public/application/components/query_panel.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import React, { useCallback, useEffect, useRef, useState } from 'react';
+import React, { useCallback, useRef, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { EuiPanel, EuiButton, EuiSuperDatePicker } from '@elastic/eui';
 import { monaco } from '@osd/monaco';
@@ -14,7 +14,7 @@ import {
   getEffectiveLanguageForAutoComplete,
 } from '../../../../data/public';
 import { IndexPattern } from '../../../../data/common/index_patterns';
-import { setQueryString } from '../utils/state_management/slices/query_slice';
+import { setQuery } from '../utils/state_management/slices/query_slice';
 
 import { RecentQuerySelector } from './recent_query_selector';
 import {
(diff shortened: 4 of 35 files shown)


angle943:
QQ: you know how the left nav takes a plugin and adds a link to it? Does that architecture allow us to add multiple names (with different names) to it from the same plugin? Because Explore, Metrics, and Traces will all currently live under one plugin

joshuali925:
> QQ: you know how the left nav takes a plugin and adds a link to it? Does that architecture allow us to add multiple names (with different names) to it from the same plugin? Because Explore, Metrics, and Traces will all currently live under one plugin

We can add multiple names for the same plugin but they will need to have different plugin ids. The plugin id is the app id that will appear in URL after `/app/`.

This PR adds them as `explore/logs`, `explore/metrics`, `explore/traces` and this is what it looks like
image

src/plugins/explore/common/index.ts
```diff
@@ -5,7 +5,11 @@
 
 export const PLUGIN_ID = 'explore';
 export const PLUGIN_NAME = 'Discover';
-export const LOGS_VIEW_ID = 'logs';
+export const EXPLORE_FLAVOR_ID = {
+  LOGS: 'logs',
+  TRACES: 'traces',
+  METRICS: 'metrics',
+} as const;
```
angle943: nitpick: why not enum?

src/plugins/explore/public/application/index.tsx
```diff
@@ -10,30 +10,46 @@ import { Provider as ReduxProvider } from 'react-redux';
...
+  switch (flavorId) {
+    case 'logs':
+      return <LogsPage setHeaderActionMenu={props.setHeaderActionMenu} />;
+    case 'traces':
+      return <TracesPage setHeaderActionMenu={props.setHeaderActionMenu} />;
+    case 'metrics':
+      return <MetricsPage setHeaderActionMenu={props.setHeaderActionMenu} />;
+    default:
+      const invalidId: never = flavorId;
+      return `Unexpected explore flavor id: ${invalidId}`;
```
angle943, line 42: is it better to throw an error? why a string?

angle943: also, should we have it default to `logs`? (or have a rendering logic that reroute to logs)

src/plugins/explore/public/plugin.ts
```diff
@@ -248,6 +248,14 @@ export class ExplorePlugin
           return () => {};
         }
 
+        // If there's no flavor id, by default redirect to the logs flavor.
+        if (!flavorId) {
+          coreStart.application.navigateToApp(`${PLUGIN_ID}/${EXPLORE_FLAVOR_ID.LOGS}`, {
+            replace: true,
+          });
+          return () => {};
+        }
```
angle943, line 259: ah i left a comment about this above

src/plugins/explore/common/index.ts
joshuali925: just preference, i find it hard to understand enum when they are transpiled to javascript. But if OSD uses enum everywhere i can change it

src/plugins/explore/public/application/index.tsx
joshuali925: This is actually dead code, it exists for type check to make the switch cases exhaustive. it might be better to render an error page if somehow it is triggered. I can update it

`logs` is default, the redirection happens in plugin.ts

angle943: i think a comment explaining its actually not possible to reach it (and a note to the file that has it) would be good

joshuali925: makes sense, added a comment

angle943 (approved):

Maosaic:
LGTM, but need rebase on top of recent commits

angle943 (approved):

Maosaic (approved):

src/plugins/explore/public/plugin.ts
```diff
@@ -250,6 +250,14 @@ export class ExplorePlugin
           return () => {};
         }
 
+        // If there's no flavor id, by default redirect to the logs flavor.
+        if (!flavor) {
+          coreStart.application.navigateToApp(`${PLUGIN_ID}/${ExploreFlavor.Logs}`, {
+            replace: true,
+          });
+          return () => {};
+        }
```
raintygao, line 259: @joshuali925 This means we don't need view route any more, right?  
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/query_explore/src/plugins/explore/public/application/index.tsx#L72 

If user enters with a view url(http://localhost:6601/w/RfEuis/app/explore/view/c5bcfc10-5191-11f0-8fb3-277e1ca2b963), it seems that this redirect logic will always be triggered and navigated with replace before view route is matched, then explore id on url will be cleared?

joshuali925: @raintygao right, I don't think we should use `app/explore/view`, because the URL structure should be `app/explore/<flavor-id>`.

During redirection on opening the saved object, https://github.com/opensearch-project/OpenSearch-Dashboards/blob/cb842c0ef1fa46ee64fd353c5d904ece249c816f/src/plugins/explore/public/application/legacy/discover/application/components/top_nav/open_search_panel.tsx#L109-L125
the code should determine which flavor it is, and redirect to the right URL directly. This part isn't implemented, last time I checked the saved explores didn't store the flavor id yet

> If user enters with a view url(http://localhost:6601/w/RfEuis/app/explore/view/c5bcfc10-5191-11f0-8fb3-277e1ca2b963), it seems that this redirect logic will always be triggered and navigated with replace before view route is matched, then explore id on url will be cleared?

This sounds like a race condition on path matching, do you have steps to reproduce? Also it should be something like `/app/explore/logs#/view/c5bcfc10-5191-11f0-8fb3-277e1ca2b963` as mentioned above

raintygao: > last time I checked the saved explores didn't store the flavor id yet

Most attributes have been stored in the object right now exclude flavor id, I'm working on other part right now and I will let you know if that is done.

PR #9908, opened 2025-06-17
[Explore] feat: add explore embeddable

raintygao:
### Description

Add embeddable for explore plugin

## Screenshot

https://github.com/user-attachments/assets/01b86419-81cf-43bb-bf08-bec986caf79e

## Changelog

- feat: Add explore embeddable

## Note
1. Support table render if no visualization is matched once table is finalized(in later PR)
2. When a PPL embeddable is added to dashboard, other embeddables using different language query may fail due to conflicts. This occurs because search service https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/search/search_service.ts#L160 only honors global query language instead of in request level. Will address this issue offline.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/application/legacy/discover/application/view_components/utils/index_pattern_helper.ts
```diff
@@ -38,7 +38,7 @@ export const fetchSavedSearch = async (
   toastNotifications
 ) => {
   try {
-    const savedSearch = await services.getSavedSearchById(savedSearchId);
+    const savedSearch = await services.getSavedExploreById(savedSearchId);
     return savedSearch;
   } catch (error) {
```
raintygao, line 43: Just fix the import, this function is actually not used, not sure if it will be deprecated later.

wanglam (approved):

PR #9925, opened 2025-06-18
fix save explore button not working

ruanyl:
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/build_services.ts
```diff
@@ -44,7 +43,7 @@ export function buildServices(
     filterManager: plugins.data.query.filterManager,
     indexPatterns: plugins.data.indexPatterns, // Direct access for convenience
     getSavedExploreById: async (id?: string) => {
-      return getSavedExploreByIdFromClient(services.savedObjectsClient, savedObjectService, id);
+      return savedObjectService.get(id);
```
raintygao, line 46: ` return savedObjectService.get(id);`This is what discover did in the past, but don't know why it was replaced with others.

ruanyl: The previous `getSavedExploreByIdFromClient` looks suspicious to me, I think it should be using `SavedExploreLoader`. Don't why it was using `savedObjectsClient` directly, maybe @ananzh can double confirm.

raintygao (approved):

SuZhou-Joe (approved):

src/plugins/explore/public/build_services.ts
ananzh: @ruanyl we haven't fixed any logic related to saved objects. I will send you a message to understand the requirement.

PR #9927, opened 2025-06-18
[Bugfix]: make UI setting client more robust when the setting key not exists

SuZhou-Joe:
## Changelog

- fix: make UI setting client more robust when the setting key not exists

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/public/ui_settings/ui_settings_client.test.ts
+++ b/src/core/public/ui_settings/ui_settings_client.test.ts
@@ -261,6 +261,12 @@ describe('#set', () => {
     await client.set('defaultIndex', 'index', UiSettingScope.USER);
     expect(getAll).toHaveBeenCalled();
   });
+
+  it('should not throw error if the key does not exist', async () => {
+    const { client } = setup();
+
+    await expect(client.set('not_exist', UiSettingScope.GLOBAL)).resolves.not.toThrowError();
+  });
 });
 
 describe('#getUserProvidedWithScope', () => {
@@ -286,6 +292,13 @@ describe('#getUserProvidedWithScope', () => {
     await expect(
       client.getUserProvidedWithScope('defaultDatasource', UiSettingScope.WORKSPACE)
     ).resolves.toBe('ds');
+  });
+
+  it('should not throw error if the key does not exist', async () => {
+    const { client } = setup();
+    const notExistSettingKey = 'not_exist';
+
+    await client.getUserProvidedWithScope(notExistSettingKey, UiSettingScope.GLOBAL);
   });
 });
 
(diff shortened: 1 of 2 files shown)


src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -277,8 +277,7 @@ You can use \`IUiSettingsClient.get("${key}", defaultValue)\`, which will just r
     key: string,
     defaults: Record<string, any>,
     enableUserControl: boolean,
-    settings: Record<string, any> = {},
-    scope: UiSettingScope | undefined
```
raintygao, line 281: Is the scope remove related to setting key fallback?

SuZhou-Joe: Nope, scope parameters is useless and safe to remove.

src/core/public/ui_settings/ui_settings_client.ts
```diff
@@ -141,7 +141,7 @@ You can use \`IUiSettingsClient.get("${key}", defaultValue)\`, which will just r
   }
 
   private validateScope(key: string, scope: UiSettingScope) {
-    const definition = this.cache[key];
+    const definition = this.cache[key] || {};
```
FriedhelmWS, line 144: Nit: could we add a logging to warn the user in this case?

SuZhou-Joe: Good idea. Will do that in a following PR. Let's focus on addressing the fix first. https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9928 Let's use this issue for tracking.

src/core/public/ui_settings/ui_settings_client.test.ts
```diff
@@ -261,6 +261,12 @@ describe('#set', () => {
     await client.set('defaultIndex', 'index', UiSettingScope.USER);
     expect(getAll).toHaveBeenCalled();
   });
+
+  it('should not throw error if the key does not exist', async () => {
+    const { client } = setup();
+
+    await expect(client.set('not_exist', UiSettingScope.GLOBAL)).resolves.not.toThrowError();
+  });
```
Hailong-am, line 269: How about set non exist key without scope

src/core/public/ui_settings/ui_settings_client.test.ts
```diff
@@ -287,6 +293,13 @@ describe('#getUserProvidedWithScope', () => {
       client.getUserProvidedWithScope('defaultDatasource', UiSettingScope.WORKSPACE)
     ).resolves.toBe('ds');
   });
+
+  it('should not throw error if the key does not exist', async () => {
+    const { client } = setup();
+    const notExistSettingKey = 'not_exist';
+
+    await client.getUserProvidedWithScope(notExistSettingKey, UiSettingScope.GLOBAL);
```
Hailong-am, line 301: Same here without scope

src/core/public/ui_settings/ui_settings_client.ts
Hailong-am: > Nit: could we add a logging to warn the user in this case?

Why we need to add warning for this case?

Hailong-am (approved):

SuZhou-Joe:
@Hailong-am Thanks for the comment, will add those enhancement in a following PR.

PR #9922, opened 2025-06-18
docs: add guidelines on testing and development

joshuali925:
### Description

This PR adds some standards and guidelines for testing and development. It provides guidelines as well as requirements on these types of checks:
- unit tests
- integration tests
- functional tests (selenium and cypress)
- performance tests
- linting and type errors

@AMoo-Miki has done some work to improve cypress functional tests and backward compatibility tests, but they are not in scope for this PR.

If this PR is approved, I'll work with infra to turn on Codecov status check to fail on PRs that do not meet code coverage standards, and block merge on PRs with failed CI.

### Issues Resolved

This enforces some mechanisms to hopefully ensure better code quality, maintainability, and less bugs.

## Changelog
- docs: add guidelines on testing and development

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #9933, opened 2025-06-18
[Patterns] Basic table UI + tests

paulstn:
### Description

Implements the UI component of Log Patterns. Creates the patterns tab. Adds storybook tests.

image

TODO:
Patterns container logic, which would append a patterns clause and pull those results into the patterns tab. 
No Results view, which would contain further description to help the user use the patterns tab.
Fix for pagination page click where clicking the numerical page (e.g. 2) does not call the onChange handler, issue stemmed from integration of Patterns within new Explore components

## Changelog

- feat: Created patterns tab with patterns table UI

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/patterns_table/constants.tsx
```diff
@@ -0,0 +1,81 @@
...
+        return '-';
+      }
+      return val;
+    },
+    align: HorizontalAlignment.Right,
+  },
+];
+
+// Mock data for patterns
+export const mockPatternItems: PatternItem[] = [
```
joshuali925: can these be defined in storybook rather than mixed with code that will run?

src/plugins/explore/public/components/patterns_table/constants.tsx
```diff
@@ -0,0 +1,81 @@
...
+    name: 'Pattern',
+    render: (pattern: string) => pattern || '-',
+  },
+  {
+    field: 'count',
+    name: 'Event count',
+    render: (val: number) => {
+      // Check if value is a valid, finite number
+      if (typeof val !== 'number' || isNaN(val) || !isFinite(val)) {
+        return '-';
```
joshuali925: i remember that we should use the longer dash `–`

src/plugins/explore/public/components/patterns_table/constants.tsx
```diff
@@ -0,0 +1,81 @@
...
+    width: '15%',
+  },
+  {
+    field: 'pattern',
+    name: 'Pattern',
+    render: (pattern: string) => pattern || '-',
+  },
+  {
+    field: 'count',
+    name: 'Event count',
```
joshuali925: would these need traslation?

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -32,6 +33,29 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
 
   tabRegistry.registerTab(logsTabDefinition);
 
+  // Register Visualizations Tab
+  tabRegistry.registerTab({
+    id: 'explore_patterns_tab',
+    label: 'Patterns',
```
joshuali925, line 41: does this need translation?

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -32,6 +33,29 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
...
+    flavor: [],
+    order: 15,
+    supportedLanguages: ['PPL'],
+
+    prepareQuery: (query) => {
+      return query;
```
joshuali925, line 47: this should add the patterns command?

src/plugins/explore/public/components/patterns_table/patterns_container.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { COUNT_FIELD, PATTERNS_FIELD } from './constants';
+
+export const PatternsContainer = () => {
+  const rows = useSelector(selectRows); // TODO: use rows from patterns-appended query
+  const totalHits = useSelector(selectTotalHits);
+
+  // Convert rows to pattern items or use default if rows is undefined
+  const items: PatternItem[] = rows?.map((row) => ({
+    pattern: row._source[PATTERNS_FIELD],
+    ratio: row._source[COUNT_FIELD] / 2096, // TODO: pull from total hits
```
joshuali925, line 19: q: rows should automatically be the new response from patterns query? does this need to wait for that query to return?

joshuali925 (commented):
missing tests?

src/plugins/explore/public/components/patterns_table/constants.tsx
paulstn: sure, i was writing my jest tests which reuse this function, however.

src/plugins/explore/public/components/patterns_table/constants.tsx
paulstn: should i wrap the text with i18n to do that?

src/plugins/explore/public/application/register_tabs.ts
paulstn: I'm planning on including that in the next PR w/ container logic. However, when I tried using this to modify the query, it seemed to modify the query in the query editor. Is that not the expected behavior?

src/plugins/explore/public/components/patterns_table/patterns_container.tsx
paulstn: I believe I'm going to pull a different redux selector than row which will be defined specifically for the patterns query. It would need to wait for that query to finish executing.

src/plugins/explore/public/components/patterns_table/constants.tsx
paulstn: is that just two dashes? how should i write it in, copy the character directly?

src/plugins/explore/public/components/patterns_table/constants.tsx
angle943: you could have like `mocks.ts` or something within this folder that could be used by both the unit tests and storybook. that way its clear that it is purely for dev purposes, `constants.tsx` sounds like it should be used in prod.

Also why is this file `.tsx`? i think `.ts` is sufficient?

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -32,6 +33,29 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
 
   tabRegistry.registerTab(logsTabDefinition);
 
+  // Register Visualizations Tab
+  tabRegistry.registerTab({
+    id: 'explore_patterns_tab',
```
angle943, line 40: how did other tabs make their ID? `explore` and `tab` seem redundant.

EDIT: ah that is how other tabs are registered.... we should change this later :|

src/plugins/explore/public/components/patterns_table/constants.tsx
```diff
@@ -0,0 +1,81 @@
...
+export const PATTERNS_FIELD = 'patterns_field';
+export const COUNT_FIELD = 'count';
+
+export const patternsTableColumns = [
+  {
+    field: 'ratio',
+    name: 'Event ratio',
+    render: (val: number) => {
+      // Check if value is a valid, finite number before formatting
+      if (typeof val !== 'number' || isNaN(val) || !isFinite(val)) {
```
angle943: can you extract this to a util on top of this file? Like:

```
const isValidFiniteNumber = (val: number) => {
   return typeof val === 'number && !isNan(val) && isFinite(val);
}
```

this seems to be used some times here, so might as well extract it as well as unit test the crap out of that function

src/plugins/explore/public/components/patterns_table/constants.tsx
angle943: yeah you can use i18n. search within explore for `i18n` and you'll see several ways to do it.

all texts that user sees should be translated

src/plugins/explore/public/components/patterns_table/constants.tsx
```diff
@@ -0,0 +1,81 @@
...
+    render: (val: number) => {
+      // Check if value is a valid, finite number
+      if (typeof val !== 'number' || isNaN(val) || !isFinite(val)) {
+        return '-';
+      }
+      return val;
+    },
+    align: HorizontalAlignment.Right,
+  },
+];
```
angle943: lets have unit tests for these funcions. should be easy to test

src/plugins/explore/public/components/patterns_table/patterns_container.tsx
```diff
@@ -0,0 +1,24 @@
...
+    count: row._source[COUNT_FIELD],
+  }));
+
+  return <PatternsTable items={items} />;
```
angle943, line 23: lets unit test this (even if its WIP i think you can have some base test herte and refactor it later)

src/plugins/explore/public/components/patterns_table/patterns_table.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const result: PatternItem[] = [...baseItems];
+
+  for (let i = 0; i < count; i++) {
+    const baseItem = baseItems[i % baseItems.length];
+    result.push({
+      pattern: `${baseItem.pattern} (variant ${Math.floor(i / baseItems.length) + 1})`,
+      ratio: baseItem.ratio / (Math.floor(i / baseItems.length) + 2), // divides the canon num by 2 every iter.
+      count: Math.floor(baseItem.count / (Math.floor(i / baseItems.length) + 2)), // same as above
+    });
+  }
```
angle943: if you make `mocks.ts` like my other suggestion suggests, this function can be there.

But maybe then `mocks` is not the right name, i'll leave it up to you to come up with a clever name that makes it clear that its just for tests.

@joshuali925 Going forward if we mandate unit tests and storybook for react components, we will often times find ourselves in a situation where mocks and helper functions need to be shared by both files. I think we need a standard we can follow, such as naming that third file something like:

```
something.mocks.ts
```


That if `.mocks.ts` is what we agreed upon for test data/functions, we can easily exclude it from our release build using regex

src/plugins/explore/public/components/patterns_table/patterns_table.tsx
```diff
@@ -0,0 +1,39 @@
...
+  items: PatternItem[];
+}
+
+export interface PatternItem {
+  pattern: string;
+  ratio: number;
+  count: number;
+}
+
+export const PatternsTable = ({ items }: PatternsTableProps) => {
```
angle943, line 20: excellent separation of business & rendering logic!

src/plugins/explore/public/components/tabs/patterns_tab.tsx
```diff
@@ -0,0 +1,19 @@
...
+ * Logs tab component for displaying log entries
+ */
+export const PatternsTab = () => {
+  return (
+    <div className="explore-logs-tab tab-container">
+      <ActionBar />
+      <PatternsContainer />
+    </div>
+  );
+};
```
angle943, line 19: lets unit test this (along with other ones).

To make it easy if you want, Just mock `ActionBar` and `PatternsContainer` with some data-test-id and just see that it is rendered. Easy coverage

src/plugins/explore/public/components/patterns_table/patterns_table.stories.tsx
joshuali925: there's a oneline recommendation in testing guideline to add mocks for services https://github.com/opensearch-project/OpenSearch-Dashboards/blob/7223dd967103ef7b11b7bff0bc77c934af25dd9f/TESTING.md?plain=1#L62

But in this case should they be `xxx.stubs.ts`?

src/plugins/explore/public/components/patterns_table/constants.tsx
joshuali925: it's actually `—`, you can copy it, it's emdash. example https://github.com/opensearch-project/OpenSearch-Dashboards/blob/1d3dd3119bc620138c6e1700b322b77ac6aae7a2/src/plugins/workspace/public/components/workspace_form/data_source_connection_table.tsx#L209-L209

src/plugins/explore/public/application/register_tabs.ts
joshuali925: i thought we don't want to modify the visible query @ananzh?

src/plugins/explore/public/components/patterns_table/patterns_table.stories.tsx
angle943: some (whether it be `.stubs` or `.mocks` or `.dev` or whatever) prefix i think we should have reserved so that we can exclude from final build. In the example you show, another way we can do it is `mocks.tsx?` or `stubs.tsx?` or whatever too, and exclude these names, but i personally would prefer using a prefix instead, since i think that opens more flexbility down the road (say for a given component you got some crazy mocks, and you need more than one file for mocks. You can have like `someNameA.mocks.ts`, `someOtherName.mocks.ts`, and so on), and clearer way to exclude files from final build.

angle943: but the most important thing is I believe some sort of a consistent way of knowing where mocks go, and having it be consistent enough so that we can target mock files specifically (such as exlcluding it from final build, adding/removing specific lint rules for it, etc)

@paulstn 's PR is not the first PR where i saw mocks in a prod file, so I do think there is ambiguity currently on where mocks go

joshuali925: from what i see in the repo it's not consistent, we have `mock.ts`, `mocks.ts`, `xxx.mocks.ts`, `xxx.test.mocks.ts`, `__mocks__/xxx.ts`. You are right we should have some standards, i'll update #9922 

what about this?
```
xxx.ts
xxx.test.ts
xxx.mocks.ts
xxx.stubs.ts
```

there can also be top level `mocks.ts` and `stubs.ts` that exports or uses multiple individual mocks and stubs files

src/plugins/explore/public/components/patterns_table/constants.tsx
paulstn: I kept it `.tsx` to make it general in case I'd need to add further upon it, but `.ts` is sufficient, I'll change it to that. I'll make a `mocks.ts` file for this mock.

src/plugins/explore/public/components/patterns_table/patterns_table.stories.tsx
angle943: great, but we can also add `....stories.tsx`. But question, what is the difference between `mocks` and `stubs`? in my head they are the same thing?

joshuali925: my understanding is that stubs only provide canned responses, mocks are additionally used to verify interaction calls

joshuali925: also i've not really used storybook, if i have multiple nested components, should i have a `xxx.stories.tsx` for each child as well as the parent?

angle943: @joshuali925 ah i see about stubs, i get it now.

2. Its up to the dev really, but at my previous team we just made storybook for every react component, whether it be child or parent. the value of that really depends on the particular component, but many times it does help.

With storybook the props are configurable, but when you only make a storybook for the parent component, it is much harder to control the children's props, as you then only have control over the `props` of the parent.

Being a less flexible person myself I vote just to make stories for everythign unless there's a good reason not to


But what is interesting about what we are doing is the fact that we are making this container vs view component distinction. In that case I don't see much value in adding a storybook for the container component.

src/plugins/explore/public/components/patterns_table/constants.tsx
paulstn: sounds good

src/plugins/explore/public/application/register_tabs.ts
paulstn: there's still some discussion about this, but for now i've just put it as a placeholder and i'm going to focus on this in another pr

src/plugins/explore/public/components/patterns_table/patterns_table.stories.tsx
paulstn: would i be right in assuming that my mockPatternItems and generateLargeDataset provides a canned response, so it needs to go to a `.stubs.ts`? and anything that's essentially a subset of a test with an `expect` inside of it should go into `.mocks.ts`?

joshuali925: > But what is interesting about what we are doing is the fact that we are making this container vs view component distinction. In that case I don't see much value in adding a storybook for the container component.

@angle943 makes sense but I still don't know what's a good rule on when to add storybook, I merged #9922, feel free to raise PR and update 🤪 

regarding mocks vs stubs (also fixtures) i don't see a lot of value to separate them, but i'll update docs later to mention recommendations on file names

joshuali925 (approved):

joshuali925 (approved):

angle943 (approved):

PR #9953, opened 2025-06-20
[Explore] Support filters for Explore PPL visualizations

joshuali925:
### Description

This PR adds Dashboard filters support to PPL. The basic filters are converted to PPL, respecting `courier:ignoreFilterIfFieldNotInIndex` setting. The DQL query and DSL filters are ignored.

Now the PPL search interceptor reads the query from the search request rather than global query service in `PPLSearchInterceptor.buildQuery()`. This removes workaround in explore embeddable to set global query introduced by https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9908.

## Screenshot

image

## Testing the changes

added unit test

## Changelog
- feat: Support filters for Explore PPL visualizations

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/public/search/search_service.ts
```diff
@@ -149,6 +149,15 @@ export class SearchService implements Plugin<ISearchSetup, ISearchStart> {
     };
   }
 
+  // Get language id from the request if available. QueryStringManager returns
+  // the global language id, which won't work in dashboard with different
+  // visualizations.
+  private getLanguageId(request: Parameters<ISearchStart['search']>[0]): string | undefined {
+    if (request.params?.body?.query?.queries?.[0]?.language)
+      return request.params.body.query.queries[0].language;
```
kavilla, line 157: i think this might have an issue. can you verify.

when the user types in a query and is enhancements is enabled then firing a query on a dashboard should be queries[1] where query[0] is the viz query.

my understanding of queries is like a stack. you want the top (highest index) query be used

src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -70,7 +70,7 @@ export class QueryEnhancementsPlugin
       getQueryString: (currentQuery: Query) => `source = ${currentQuery.dataset?.title}`,
       fields: {
         sortable: false,
-        filterable: false,
+        filterable: true,
```
kavilla: we might have to revisit this as not every field can be filterable (non indexed fields)

src/plugins/query_enhancements/public/search/sql_search_interceptor.ts
```diff
@@ -46,7 +46,11 @@ export class SQLSearchInterceptor extends SearchInterceptor {
       },
     };
 
-    return fetch(context, this.queryService.queryString.getQuery()).pipe(
+    // Use query from request if available, otherwise fall back to queryStringManager
+    const query =
+      request.params?.body?.query?.queries?.[0] || this.queryService.queryString.getQuery();
```
kavilla: tbh. i dont think there should be a case where the search interceptor shouldnt get the query object.

src/plugins/query_enhancements/public/search/ppl_search_interceptor.ts
```diff
@@ -88,21 +95,43 @@ export class PPLSearchInterceptor extends SearchInterceptor {
...
+      UI_SETTINGS.COURIER_IGNORE_FILTER_IF_FIELD_NOT_IN_INDEX
+    );
+
+    const filterClause = convertFiltersToWhereClause(filters, index, ignoreFilterIfFieldNotInIndex);
+    const commands = query.query.split('|');
+    if (filterClause) {
+      commands.splice(1, 0, filterClause);
+    }
+
+    const dataset = this.queryService.queryString.getQuery().dataset;
```
kavilla: shouldn't we be accessing the dataset from the query object in the request

src/plugins/query_enhancements/public/search/ppl_search_interceptor.ts
```diff
@@ -88,21 +95,43 @@ export class PPLSearchInterceptor extends SearchInterceptor {
...
+    // Only append filters if query is running search command (e.g. not describe command)
     if (!isPPLSearchQuery(query)) return query;
 
-    const [baseQuery, ...afterPipeParts] = query.query.split('|');
-    const afterPipe = afterPipeParts.length > 0 ? ` | ${afterPipeParts.join('|').trim()}` : '';
-    const timeFilter = this.getTimeFilter(dataset.timeFieldName);
-    return { ...query, query: baseQuery + timeFilter + afterPipe };
+    const filters = this.queryService.filterManager.getFilters();
+    const index = request.params?.index
+      ? this.indexPatterns.getByTitle(request.params.index)
```
kavilla: for getByTitle do we need to ignore errors if this index doesn't exist.

nit: the other thing is that this should usually be fired by the search source. so not sure if we want to consider building the request to get the filterClause and pass it to the search request. but i'm cool with this logic here

kavilla: however if we do want to move to a data source specific search interceptor and strategy. then this might influence where this should exist

src/plugins/data/public/search/search_service.ts
joshuali925: for DSL visualizations there won't be a `queries` field here, the user query is added as a filter in `request.params.body.query.bool.filter`. For Explore visualizations I actually don't see `queries` being more than one element, even if I type a query in the search bar

src/plugins/query_enhancements/public/search/sql_search_interceptor.ts
joshuali925: right, this is picked from https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9965, it's no longer in this PR

src/plugins/query_enhancements/public/search/ppl_search_interceptor.ts
joshuali925: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/3a5245b94c3a5fcd8eb3b4c178f983c8c35e67df/src/plugins/query_enhancements/public/search/ppl_search_interceptor.ts#L119-L121

correct, i left a fixme comment, didn't finish this

edit: fixed in 21caacb

src/plugins/query_enhancements/public/search/ppl_search_interceptor.ts
joshuali925: yeah the search source contains build DSL query logic, but i'm not sure about adding every language there. the interceptor and strategy can be defined by plugins, but seems search source isn't extensible currently

for data source specific search interceptor and strategy, i think the query building logic could be different per data source even for the same language, making the logic per language might not be enough. One example is 
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/3a5245b94c3a5fcd8eb3b4c178f983c8c35e67df/src/plugins/query_enhancements/public/search/ppl_search_interceptor.ts#L80-L93

src/plugins/query_enhancements/public/plugin.tsx
joshuali925: i think it takes the intersection, both language and the actual field need to be filterable in order to add filters
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/762eb4e5ca471f35f7194bd496820f26120b1e23/src/plugins/data/common/index_patterns/fields/index_pattern_field.ts#L135-L142

the field filterable behavior isn't modified

src/plugins/query_enhancements/public/search/ppl_search_interceptor.ts
joshuali925: > for getByTitle do we need to ignore errors if this index doesn't exist.

i think so, will add. ~~but when does it not exist? i can't remember does it exist if user selected the index dataset rather than creating index patterns in discover, will do some testing~~

edit: fixed in 21caacb. I did some testing, the `getByTitle` works as expected for `indexes` datasets as well

angle943 (dismissed):

src/plugins/query_enhancements/public/plugin.tsx
joshuali925: actually you are right this is `getOsdFieldOverrides().filterable !== undefined` not `!getOsdFieldOverrides().filterable`, if setting true here overrides the check below. I removed `filterable: true,` in 0ee6f7e

also i pasted the wrong code, the logic to check whether fields are filterable for filter UI is in https://github.com/opensearch-project/OpenSearch-Dashboards/blob/0daad39467004cd06bd784ec154bcbb1b2e6ea63/src/plugins/data/common/index_patterns/fields/utils.ts#L48-L55

src/plugins/data/public/search/search_service.ts
joshuali925: checked with @kavilla offline. For DSL visualizations, the queries passed to buildOpenSearchQuery is indeed `['<global-query>', '<vis-query>']`. But `queries` is not returned for DSL, it returns the built DSL JSON directly.
image

Currently `queries` for PPL always contains 1 element regardless of user query in search bar. https://github.com/opensearch-project/OpenSearch-Dashboards/blob/0daad39467004cd06bd784ec154bcbb1b2e6ea63/src/plugins/data/common/opensearch_query/opensearch_query/build_opensearch_query.ts#L54-L78

If we add PPL support in the search bar, then this may change, but so far there is no plan

kavilla (approved):

ananzh (approved):

PR #9956, opened 2025-06-23
[Explore] Refactor: Improve Explore Plugin Result Access Patterns & Reduce Code Duplication

ananzh:
###  **ExecutionCacheKeys Structure** 
Note: ExecutionCacheKeys is a temp solution for access result. It is set to a fix length of 2 for easy tmp use. Tabs always use ExecutionCacheKeys[1] while side panel, histogram use ExecutionCacheKeys[0]. This temp solution will be cleaned out if we don't need to save timefilter or modified if needed. 
- **Always length 2**: `[defaultCacheKey, tabCacheKey]` for consistent access 
- **Clear separation**: Index 0 for non-tab components, Index 1 for tab components 
- **Simplified logic**: No more length checking needed

### **Removed Problematic Selectors** 
- Deleted `selectRows`, `selectTotalHits`, `selectFieldCounts`, `selectCurrentCacheKey` 
- Components now use direct cache access for better control 
- Fixed `selectTabData` to remove dependencies on deleted selectors  

### **Component Updates** 
- **Updated all tab components** to use `executionCacheKeys[1]` consistently 
- **Added TODO comments** for custom processor implementation guidance 

### **CSV Export Improvements** 
- **Renamed** `createSearchSourceWithQuery` → `updateSearchSource` (Discover consistency) 
- **Fixed cache usage**: Uses existing `executionCacheKeys[1]` instead of recreating keys 
- **Reduced duplication**: `exportMaxSizeCsv` now reuses `updateSearchSource`  

### **Documentation** 
- **Complete rewrite** of docs and clean out

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/application/pages/logs/logs_page.tsx
```diff
@@ -66,9 +66,10 @@ export const LogsPage: React.FC<Partial<Pick<AppMountParameters, 'setHeaderActio
       return [];
     }
 
-    // Use default query cacheKey
-    const cacheKey = executionCacheKeys[0];
-    const results = state.results[cacheKey];
+    // Use default query cacheKey with safety check
+    const cacheKey =
+      executionCacheKeys && executionCacheKeys.length >= 1 ? executionCacheKeys[0] : null;
```
angle943: 1. lets get rid of `with safety check` (all our code should have safety check so no need to have that in our comments
2. `executionCacheKeys` is typed as an array so we don't need the `executionCacheKeys &&` part
3. To make it easy to read, how about `executionCacheKeys[0] ?? null`
4. but honestly, i think `executionCacheKeys[0]` is enough. If its an empty string or undefined then it will work the same as `null`

src/plugins/explore/public/application/pages/metrics/metrics_page.tsx
```diff
@@ -66,9 +66,9 @@ export const MetricsPage: React.FC<Partial<Pick<AppMountParameters, 'setHeaderAc
       return [];
     }
 
-    // Use default query cacheKey
-    const cacheKey = executionCacheKeys[0];
-    const results = state.results[cacheKey];
+    const cacheKey =
+      executionCacheKeys && executionCacheKeys.length >= 1 ? executionCacheKeys[0] : null;
```
angle943: Same as above

src/plugins/explore/public/application/pages/traces/traces_page.tsx
```diff
@@ -66,9 +66,10 @@ export const TracesPage: React.FC<Partial<Pick<AppMountParameters, 'setHeaderAct
       return [];
     }
 
-    // Use default query cacheKey
-    const cacheKey = executionCacheKeys[0];
-    const results = state.results[cacheKey];
+    // Use default query cacheKey with safety check
+    const cacheKey =
+      executionCacheKeys && executionCacheKeys.length >= 1 ? executionCacheKeys[0] : null;
```
angle943: same as above

src/plugins/explore/public/application/pages/logs/logs_page.tsx
angle943: Can we also update the `executionCacheKeys` in the `UIState` interface to be of tuple type:

```
executionCacheKeys: [string, string]
```

(then you prob need to replace the initialState to be ['', ''])

src/plugins/explore/public/components/data_table/explore_data_table.tsx
```diff
@@ -47,10 +48,17 @@ const ExploreDataTableComponent = () => {
...
   const savedSearch = useSelector(selectSavedSearch);
-  const rows = useSelector(selectRows);
   const columns = useSelector(selectColumns);
   const { indexPattern } = useIndexPatternContext();
 
+  const executionCacheKeys = useSelector((state: RootState) => state.ui.executionCacheKeys);
+  const results = useSelector((state: RootState) => state.results);
+
+  const cacheKey =
+    executionCacheKeys && executionCacheKeys.length >= 2 ? executionCacheKeys[1] : null;
```
angle943:
```
const cacheKey = executionCacheKeys[1] ?? null
```

or

```
const cacheKey = executionCacheKeys[1];
```

src/plugins/explore/public/components/patterns_table/patterns_container.tsx
```diff
@@ -5,13 +5,26 @@
...
+import { RootState } from '../../application/utils/state_management/store';
 
 export const PatternsContainer = () => {
-  const rows = useSelector(selectRows); // TODO: use rows from patterns-appended query
-  const totalHits = useSelector(selectTotalHits);
+  const executionCacheKeys = useSelector((state: RootState) => state.ui.executionCacheKeys);
+  const results = useSelector((state: RootState) => state.results);
+
+  const cacheKey =
+    executionCacheKeys && executionCacheKeys.length >= 2 ? executionCacheKeys[1] : null;
```
angle943: same here

src/plugins/explore/public/components/patterns_table/patterns_container.tsx
```diff
@@ -5,13 +5,26 @@
...
+  const rawResults = cacheKey ? results[cacheKey] : null;
+
+  // TODO: Register custom processor for patterns tab if needed
+  //       If no need, feel free to remove this comment
+  // const tabDefinition = services.tabRegistry?.getTab?.('patterns');
+  // const processor = tabDefinition?.resultsProcessor || defaultResultsProcessor;
+  // const processedResults = processor(rawResults, indexPattern);
+
+  const rows = rawResults?.hits?.hits || [];
+  const totalHits = (rawResults?.hits?.total as any)?.value || rawResults?.hits?.total || 0;
```
angle943: why is `total` typed as `number` in the interface `SearchResponse`? What scenarios would it be `total.value`? is our typing for that wrong?

src/plugins/explore/public/components/tabs/action_bar/action_bar.tsx
```diff
@@ -25,8 +22,15 @@ const ActionBarComponent = () => {
   const { core } = services;
 
   const savedSearch = useSelector(selectSavedSearch);
-  const rows = useSelector(selectRows);
-  const totalHits = useSelector(selectTotalHits);
+
+  const executionCacheKeys = useSelector((state: RootState) => state.ui.executionCacheKeys);
+  const results = useSelector((state: RootState) => state.results);
+  const cacheKey =
+    executionCacheKeys && executionCacheKeys.length >= 2 ? executionCacheKeys[1] : null;
```
angle943: same here

src/plugins/explore/public/components/visualizations/visualization_container.tsx
```diff
@@ -36,22 +34,19 @@ export const VisualizationContainer = () => {
...
-  const rows = useSelector(selectRows);
-  const fieldSchema = useSelector((state: RootState) => {
-    const executionCacheKeys = state.ui?.executionCacheKeys || [];
-    if (executionCacheKeys.length === 0) {
-      return [];
-    }
+  // Use tab-specific cache key consistently
+  const executionCacheKeys = useSelector((state: RootState) => state.ui.executionCacheKeys);
+  const results = useSelector((state: RootState) => state.results);
+  const cacheKey =
```
angle943: same here

src/plugins/explore/public/components/visualizations/visualization_container.tsx
```diff
@@ -36,22 +34,19 @@ export const VisualizationContainer = () => {
...
-      return results.fieldSchema;
-    }
+  // TODO: Register custom processor for visualization tab
+  // const tabDefinition = services.tabRegistry?.getTab?.('explore_visualization_tab');
+  // const processor = tabDefinition?.resultsProcessor || defaultResultsProcessor;
 
-    return [];
-  });
+  const rows = useMemo(() => rawResults?.hits?.hits || [], [rawResults]);
+  const fieldSchema = useMemo(() => rawResults?.fieldSchema || [], [rawResults]);
```
angle943, line 66: rawResults use data from `results`, which itself is gotten from the memoized `useSelector`. These above two memoizations won't do much, it'll just waste memory. Lets remove it

src/plugins/explore/common/index.ts
```diff
@@ -5,6 +5,7 @@
 
 export const PLUGIN_ID = 'explore';
 export const PLUGIN_NAME = 'Discover';
+export const LOGS_VIEW_ID = 'logs';
```
joshuali925: remove this, components should use `ExploreFlavor.Logs`

src/plugins/explore/public/application/pages/logs/logs_page.tsx
```diff
@@ -66,9 +66,10 @@ export const LogsPage: React.FC<Partial<Pick<AppMountParameters, 'setHeaderActio
...
+    const results = cacheKey ? state.results[cacheKey] : null;
```
joshuali925: the null check already happened above
```suggestion
    const cacheKey = executionCacheKeys[0];
    const results = cacheKey ? state.results[cacheKey] : null;
```

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -139,10 +139,9 @@ export const executeQueries = createAsyncThunk<
   const query = state.query;
   const activeTabId = state.ui.activeTabId || 'logs';
   const results = state.results;
-  const cacheKeys = state.ui.executionCacheKeys ? [...state.ui.executionCacheKeys] : [];
 
   if (!services) {
-    return { cacheKeys };
+    return { cacheKeys: ['', ''] };
```
joshuali925: why does it need to be two strings? also if it's two strings tuple, the `cacheKeys: string[]` is inaccurate

src/plugins/explore/public/application/utils/state_management/docs/query_processing.md
```diff
@@ -0,0 +1,174 @@
...
+#### Histogram Query (`executeHistogramQuery`)
+- **Purpose**: Provides data for histogram charts and sidebar components
+- **Features**: Includes aggregations for time-based visualization
+- **Cache Key**: Always stored at `executionCacheKeys[0]`
+- **Configuration**: Uses interval from Redux state
+
+#### Tab Query (`executeTabQuery`)
+- **Purpose**: Provides data for active tab components
+- **Features**: Raw search results without aggregations
+- **Cache Key**: Always stored at `executionCacheKeys[1]`
```
joshuali925: why not make it a explicit hash map, rather than depending on index? what if there are more query types added?

src/plugins/explore/public/application/utils/state_management/docs/query_processing.md
```diff
@@ -0,0 +1,174 @@
...
+### Default Query Preparation
+
+All queries go through a default preparation step that ensures histogram compatibility:
+
+```typescript
+export const defaultPrepareQuery = (query: any) => {
+  return {
+    ...query,
+    query: typeof query.query === 'string' 
+      ? query.query.replace(/\s*\|\s*stats.*$/i, '')
```
joshuali925:
```suggestion
      ? query.query.replace(/\s*\|\s*stats\s+.+$/i, '') 
```

also what if it's a multi-line query where `stats` spans multiple lines? i don't think JS supports dot all, you'd need some other regex

src/plugins/explore/public/components/visualizations/visualization_container.tsx
ananzh: good point. will update.

src/plugins/explore/common/index.ts
ananzh: will clean

src/plugins/explore/public/components/patterns_table/patterns_container.tsx
ananzh: let me clean it up here as it is not used

src/plugins/explore/public/components/visualizations/visualization_container.tsx
ananzh: got some errors 
```
Please only submit bug reports when using the officially supported version.

=============
ERROR 
      /home/ubuntu/work/OpenSearch-Dashboards/src/plugins/explore/public/components/visualizations/visualization_container.tsx
        64:9  error  The 'rows' logical expression could make the dependencies of useMemo Hook (at line 75) change on every render. To fix this, wrap the initialization of 'rows' in its own useMemo() Hook   react-hooks/exhaustive-deps
        64:9  error  The 'rows' logical expression could make the dependencies of useMemo Hook (at line 148) change on every render. To fix this, wrap the initialization of 'rows' in its own useMemo() Hook  react-hooks/exhaustive-deps
```
when removing. I think rows are fieldSchema are memoized in a lot of places. WIll have @abbyhu2000 to clean out this file.

angle943 (approved):

src/plugins/explore/public/components/visualizations/visualization_container.tsx
joshuali925: that's because you are creating a new array object on every re-render if hits is undefined. you can use a constant empty array

src/plugins/explore/public/components/visualizations/visualization_container.tsx
```diff
@@ -45,25 +44,26 @@ export const VisualizationContainer = () => {
...
-    if (results && results.fieldSchema) {
-      return results.fieldSchema;
-    }
+  // TODO: Register custom processor for visualization tab
+  // const tabDefinition = services.tabRegistry?.getTab?.('explore_visualization_tab');
+  // const processor = tabDefinition?.resultsProcessor || defaultResultsProcessor;
 
-    return [];
-  });
+  const rows = useMemo(() => rawResults?.hits?.hits || [], [rawResults]);
```
joshuali925, line 63: rows is never undefined, the null check at line 103 should be removed

src/plugins/explore/public/application/utils/state_management/docs/query_processing.md
ananzh: This file is outdated and the logic of executionCacheKeys has been completely cleaned out . It was a temp solution.
```
### Cache Key Strategy

Components compute cache keys directly using prepared query strings:

- **Default Components**: Use `defaultPrepareQuery(queryString)` for histogram/sidebar components
- **Tab Components**: Use tab-specific `prepareQuery(queryString)` or fall back to `defaultPrepareQuery`
- **Direct Computation**: Components compute their own cache keys when needed
- **No Shared State**: Cache keys computed on-demand, not stored in UI state
```

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
ananzh: this file is outdated. We have cleaned out cacheKeys. Now we just return 
```
if (!services) {
    return;
  }
```

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -28,15 +27,13 @@ import { RootState } from '../store';
...
-  return {
-    ...query,
-    query:
-      typeof query.query === 'string' ? query.query.replace(/\s*\|\s*stats.*$/i, '') : query.query,
-  };
+export const defaultPrepareQuery = (queryString: string): string => {
+  // Remove stats pipe for histogram compatibility
+  return typeof queryString === 'string'
+    ? queryString.replace(/\s*\|\s*stats.*$/i, '')
+    : queryString;
```
joshuali925, line 36: the typeof string check is unnecessary? and this comment still applies here https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9956#discussion_r2167467331

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -499,6 +297,104 @@ export const executeTabQuery = createAsyncThunk<
...
     throw error;
   }
+};
+
+/**
+ * Helper function to create SearchSource with common configuration
+ */
+const createSearchSourceWithQuery = async (
+  preparedQuery: any,
+  indexPattern: any,
```
joshuali925, line 307: avoid any

PR #9975, opened 2025-06-25
[Explore] Add AI summary component in explore plugin

FriedhelmWS:
### Description

Screenshot 2025-06-25 at 15 05 45

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/index.tsx
+++ b/src/plugins/explore/public/application/index.tsx
@@ -9,13 +9,14 @@ import { Router, Route, Switch } from 'react-router-dom';
 import { Provider as ReduxProvider } from 'react-redux';
 import { Store } from 'redux';
 import { AppMountParameters } from '../../../../core/public';
-import { ExploreServices } from '../types';
+import { ExploreServices, ExploreSetupDependencies } from '../types';
 import { LogsPage } from './pages/logs';
 import { OpenSearchDashboardsContextProvider } from '../../../opensearch_dashboards_react/public';
 import { IndexPatternProvider } from './components/index_pattern_context';
 import { ExploreFlavor } from '../../common';
 import { TracesPage } from './pages/traces';
 import { MetricsPage } from './pages/metrics';
+import { DataPublicPluginSetup } from '../../../../plugins/data/public';
 
 // Route component props interface
 interface ExploreRouteProps {
@@ -24,16 +25,22 @@ interface ExploreRouteProps {
 }
 
 type ExploreComponentProps = ExploreRouteProps &
-  Partial<Pick<AppMountParameters, 'setHeaderActionMenu'>>;
+  Partial<Pick<AppMountParameters, 'setHeaderActionMenu'>> & { dataSetup: DataPublicPluginSetup };
 
 const renderExploreFlavor = (flavor: ExploreFlavor, props: ExploreComponentProps) => {
   switch (flavor) {
     case ExploreFlavor.Logs:
-      return <LogsPage setHeaderActionMenu={props.setHeaderActionMenu} />;
+      return (
+        <LogsPage setHeaderActionMenu={props.setHeaderActionMenu} dataSetup={props.dataSetup} />
+      );
     case ExploreFlavor.Traces:
-      return <TracesPage setHeaderActionMenu={props.setHeaderActionMenu} />;
+      return (
+        <TracesPage setHeaderActionMenu={props.setHeaderActionMenu} dataSetup={props.dataSetup} />
+      );
     case ExploreFlavor.Metrics:
-      return <MetricsPage setHeaderActionMenu={props.setHeaderActionMenu} />;
+      return (
+        <MetricsPage setHeaderActionMenu={props.setHeaderActionMenu} dataSetup={props.dataSetup} />
+      );
     default:
       // This code should never be reached at runtime, it exists to make the
       // switch cases exhaustive
@@ -44,20 +51,22 @@ const renderExploreFlavor = (flavor: ExploreFlavor, props: ExploreComponentProps
 
 // View route for saved searches
 const ViewRoute = (props: ExploreComponentProps) => (
-  <LogsPage setHeaderActionMenu={props.setHeaderActionMenu} />
+  <LogsPage setHeaderActionMenu={props.setHeaderActionMenu} dataSetup={props.dataSetup} />
 );
 
 export const renderApp = (
   { element, history, setHeaderActionMenu }: AppMountParameters,
   services: ExploreServices,
   store: Store,
-  flavor: ExploreFlavor
+  flavor: ExploreFlavor,
+  setupDeps: ExploreSetupDependencies
 ) => {
   // Create main route props
   const mainRouteProps = {
     services,
     history,
     setHeaderActionMenu,
+    dataSetup: setupDeps.data,
   };
   ReactDOM.render(
     <Router history={history}>
@@ -69,7 +78,7 @@ export const renderApp = (
                 {/* View route for saved searches */}
                 {/* TODO: Do we need this? We might not need to, please revisit */}
                 <Route path="/view/:id" exact>
-                  <ViewRoute {...mainRouteProps} />
+                  <ViewRoute {...mainRouteProps} dataSetup={setupDeps.data} />
                 </Route>
 
                 <Route path={[`/`]} exact={false}>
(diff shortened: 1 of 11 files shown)


src/plugins/explore/public/components/results_summary/results_summary_panel.scss
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/* stylelint-disable @osd/stylelint/no_modifying_global_selectors */
```
kavilla: please avoid doing this if possible. this is so we don't accidentally modify styles globally

if any plugin uses the class `resultSummary` then we would be modifying them which might be likely.

src/plugins/explore/public/components/results_summary/results_summary_panel.tsx
```diff
@@ -0,0 +1,302 @@
...
+  data: DataPublicPluginSetup;
+  usageCollection?: UsageCollectionSetup;
+  brandingLabel?: string;
+}
+
+type AccordionState = 'closed' | 'open';
+
+const ACCORDION_STATE_LOCAL_STORAGE_KEY = 'resultsSummary.summaryAccordionState';
+
+export const convertResult = (body: IDataFrame) => {
```
kavilla, line 39: how come we don't use the proper search functionality?

src/plugins/explore/public/components/results_summary/results_summary.tsx
```diff
@@ -0,0 +1,181 @@
...
+  EuiCopy,
+  EuiSmallButtonEmpty,
+  EuiAccordion,
+  EuiPanel,
+  EuiIcon,
+} from '@elastic/eui';
+import React from 'react';
+import { i18n } from '@osd/i18n';
+
+export enum FeedbackStatus {
```
kavilla, line 21: ive seen this defined multiple times in this project we shouldn't redefine it if i t can be available at a different place

src/plugins/explore/public/components/results_summary/results_summary_panel.scss
FriedhelmWS: Hi, I personally think this would be intentional as we want to custom the accordion component (i.e. add buttons to the header). Could you please suggest a better approach without modifying the global style if possible? Thanks!

src/plugins/explore/public/application/index.tsx
```diff
@@ -25,16 +26,22 @@ interface ExploreRouteProps {
...
 type ExploreComponentProps = ExploreRouteProps &
-  Partial<Pick<AppMountParameters, 'setHeaderActionMenu'>>;
+  Partial<Pick<AppMountParameters, 'setHeaderActionMenu'>> & { dataSetup: DataPublicPluginSetup };
 
 const renderExploreFlavor = (flavor: ExploreFlavor, props: ExploreComponentProps) => {
   switch (flavor) {
     case ExploreFlavor.Logs:
-      return <LogsPage setHeaderActionMenu={props.setHeaderActionMenu} />;
+      return (
+        <LogsPage setHeaderActionMenu={props.setHeaderActionMenu} dataSetup={props.dataSetup} />
```
kavilla: eventually if any access to services and plugins. it would make more sense to use the OpenSearch Dashboards react context

src/plugins/explore/public/components/results_summary/results_summary.tsx
```diff
@@ -0,0 +1,181 @@
...
+  actionButtonVisible: boolean;
+  feedback: FeedbackStatus;
+  afterFeedbackTip: string;
+  onFeedback: (satisfied: boolean) => void;
+  summary: string;
+  canGenerateSummary: boolean;
+  loading: boolean;
+  onGenerateSummary: () => void;
+  brandingLabel?: string;
+  sampleSize: number;
```
kavilla, line 41: are we not respecting the advanced settings sample size

FriedhelmWS: This is the sample size for AI agent so I think it should not be the same sample size we defined in advanced settings. In the future, we can add another setting option for AI-related sample size.

src/plugins/explore/public/application/index.tsx
FriedhelmWS: Would it be appropriate to put a setup dependency in the context of explore service?

src/plugins/explore/public/components/results_summary/results_summary.tsx
FriedhelmWS: Could we leave this to a later PR for refactoring? Since this one is only add the component to the page, there would be another PR for fully integrate with t2ppl feature by @ruchidh, add unit test, remove mocks, clean up the code, etc.

src/plugins/explore/public/components/results_summary/results_summary_panel.tsx
FriedhelmWS: This is not the search but convert the result data frame to proper JSON object, I think this is only relevant to t2ppl API.

src/plugins/explore/public/components/results_summary/results_summary.tsx
kavilla: apologies. do you mind explaining more. i assumed the results summary should be the summary on the results provided to the user. the summary utilizes the query from the user right? if so then that should be the sample size from advanced settings

if not regardless of the returned results we are just generating a summary of a static number? if the user gets 10 results, wouldn't it be confusing to display a summary on 100 results because we statically defined it for 100 results.

src/plugins/explore/public/components/results_summary/results_summary_panel.scss
kavilla: hey. thanks for the quick response.

for some context the stylelinter was added to prevent visual regression bugs. and resultsSummary seems like it has a high likelihood to change the styling of a component unrelated to this component without the plugin owners knowledge.

we could potentially make this component class name more specific but my rule of thumb has been if its failing the stylelinter then we should revisit the actual ux design.

oui component supports the button in the header if im not mistaken but do you mean the styling? looking at the oui props https://oui.opensearch.org/1.19/#/layout/accordion there should be ability to pass in classNames to the component to style without using global selectors

src/plugins/explore/public/components/results_summary/results_summary_panel.scss
```diff
@@ -0,0 +1,77 @@
...
+
+.resultsSummary {
+  &.resultsSummary_accordion {
+    .euiAccordion__triggerWrapper {
+      /* stylelint-disable  @osd/stylelint/no_restricted_values */
+      background: lightOrDarkTheme(linear-gradient(to left, #edf7ff, #f9f4ff), $euiColorEmptyShade);
```
kavilla, line 11: iif this is absolutely required and there is no SASS variable already then we should define them. as this shouldn't be the only place this is used

src/plugins/explore/public/components/results_summary/results_summary_panel.tsx
kavilla: for more context, PPL search essentially does the same. and it's structured like search via
`hits`.

so to me it looks like we generally treat this feature as search where we could just leverage existing code rather no copy and pasting.

FriedhelmWS: For discover, we retrieve the raw data from an subscribed observable, we actually not doing a search again as all results are relevant, no result will be filter out in this function and it is just simply a utils function to convert result from one format to another.

src/plugins/explore/public/components/results_summary/results_summary_panel.scss
FriedhelmWS: I’ve updated some of the CSS class names to be more distinctive and to use custom className passing. There are still a few global styles that we can’t work around, but I hope these changes help reduce the risk of future style collisions.

src/plugins/explore/public/application/index.tsx
ruanyl: @FriedhelmWS I think you can use `data` from the context which is `DataPublicPluginStart`

FriedhelmWS: Done, sorry for didn't noticed that QueryStart and SearchStart are also valid to be used.

ruanyl (dismissed):

raintygao (approved):

PR #9979, opened 2025-06-25
[Explore] Add fields selection when there is no rule matched

FriedhelmWS:
### Description
Screenshot 2025-06-25 at 19 07 34

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/rule_repository.ts
+++ b/src/plugins/explore/public/components/visualizations/rule_repository.ts
@@ -48,6 +48,7 @@ const oneMetricOneDateRule: VisualizationRule = {
   description: 'Time series visualization for single metric',
   matches: (numerical, categorical, date) =>
     numerical.length === 1 && date.length === 1 && categorical.length === 0,
+  matchIndex: [1, 0, 1],
   chartTypes: [
     { ...CHART_METADATA.line, priority: 100 },
     { ...CHART_METADATA.area, priority: 80 },
@@ -83,6 +84,7 @@ const twoMetricOneDateRule: VisualizationRule = {
   matches: (numerical, categorical, date) =>
     numerical.length === 2 && categorical.length === 0 && date.length === 1,
   chartTypes: [{ ...CHART_METADATA.line, priority: 100 }],
+  matchIndex: [2, 0, 1],
   toExpression: (
     transformedData,
     numericalColumns,
@@ -108,6 +110,7 @@ const oneMetricOneCateOneDateRule: VisualizationRule = {
   description: 'Time series visualization with one metric and one category',
   matches: (numerical, categorical, date) =>
     numerical.length === 1 && categorical.length === 1 && date.length === 1,
+  matchIndex: [1, 1, 1],
   chartTypes: [
     { ...CHART_METADATA.line, priority: 100 },
     { ...CHART_METADATA.area, priority: 80 },
@@ -166,6 +169,7 @@ const oneMetricTwoCateOneDateRule: VisualizationRule = {
   description: 'Multiple time series visualizations',
   matches: (numerical, categorical, date) =>
     numerical.length === 1 && categorical.length === 2 && date.length === 1,
+  matchIndex: [1, 2, 1],
   chartTypes: [
     { ...CHART_METADATA.line, priority: 100 },
     { ...CHART_METADATA.area, priority: 80 },
@@ -224,6 +228,7 @@ const threeMetricsRule: VisualizationRule = {
   matches: (numerical, categorical, date) =>
     numerical.length === 3 && date.length === 0 && categorical.length === 0,
   chartTypes: [{ ...CHART_METADATA.heatmap, priority: 100 }],
+  matchIndex: [3, 0, 0],
   toExpression: (
     transformedData,
     numericalColumns,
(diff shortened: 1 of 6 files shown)


src/plugins/explore/public/components/visualizations/visualization_empty_state.tsx
ruanyl, line 1: Shall we name the file as `field_selector.tsx`? Also the component name

src/plugins/explore/public/components/visualizations/visualization_empty_state.tsx
```diff
@@ -0,0 +1,474 @@
...
+        }
+        initialIsOpen={initialIsOpen}
+      >
+        <EuiSpacer size="xs" />
+        <EuiPanel paddingSize="s" hasBorder={false} color="subdued">
+          {children}
+        </EuiPanel>
+      </EuiAccordion>
+      <EuiSpacer size="s" />
+      <div style={{ width: '100%', height: 1, backgroundColor: '#D4DAE5' }} />
```
ruanyl: what's this for?

ruanyl: OuiHorizontalRule?

src/plugins/query_enhancements/public/query_assist/_index.scss
```diff
@@ -3,15 +3,53 @@
...
 .queryAssist {
   &.queryAssist__summary {
     margin-top: $euiSizeS;
   }
 
-  &.queryAssist__summary_banner {
-    /* stylelint-disable  @osd/stylelint/no_restricted_values */
-    background: lightOrDarkTheme(linear-gradient(to left, #edf7ff, #f9f4ff), $euiColorEmptyShade);
-    padding: $euiSizeXS $euiSizeXS $euiSizeXS $euiSizeS;
+  &.queryAssist__summary_accordion {
```
ruanyl: These changes not belong to this PR?

src/plugins/explore/public/components/visualizations/visualization_empty_state.tsx
FriedhelmWS: Thanks!

src/plugins/query_enhancements/public/query_assist/_index.scss
FriedhelmWS: Sorry, removed

src/plugins/explore/public/components/visualizations/visualization_empty_state.tsx
FriedhelmWS: How about call it `visualization_selector`? Since it is not only for field but also chart type selections.

ruanyl (approved):

src/plugins/explore/public/components/visualizations/visualization_container.tsx
```diff
@@ -65,12 +65,15 @@ export const VisualizationContainer = () => {
   const selectedChartType = useSelector(selectChartType);
   const fieldSchema = useMemo(() => rawResults?.fieldSchema || [], [rawResults]);
 
-  const visualizationData = useMemo(() => {
+  const [visualizationData, setVisualizationData] = useState<
```
SuZhou-Joe, line 68: Do not get it why we can't keep the memo.

src/plugins/explore/public/components/visualizations/visualization_registry.ts
```diff
@@ -57,7 +57,15 @@ export class VisualizationRegistry {
 
     // TODO: Handle the case where no rule matches, render empty state or default table visualization
```
SuZhou-Joe, line 58: Not sure if we can remove this TODO since we already handle the empty case.

src/plugins/explore/public/components/visualizations/rule_repository.ts
```diff
@@ -48,6 +48,7 @@ const oneMetricOneDateRule: VisualizationRule = {
   description: 'Time series visualization for single metric',
   matches: (numerical, categorical, date) =>
     numerical.length === 1 && date.length === 1 && categorical.length === 0,
+  matchIndex: [1, 0, 1],
```
SuZhou-Joe, line 51: Seems there is already `matches`, do we still need matchIndex, or what's the difference?

src/plugins/explore/public/components/visualizations/visualization_container.tsx
FriedhelmWS: I think the actual visualization empty state component is design to requiring update the `visualizationData ` so seems it will not work with `useMemo`.

src/plugins/explore/public/components/visualizations/visualization_registry.ts
FriedhelmWS: Sorry for missing this, removed this in the next PR.

src/plugins/explore/public/components/visualizations/rule_repository.ts
FriedhelmWS: For the visualization empty state component, we need to identify which visualization rules are applicable based on available data columns. The matching logic should use a "greater than or equal to" comparison rather than exact matching, meaning a rule is valid if the available columns meet or exceed the rule's minimum requirements. To support this, we may introduce a matchIndex parameter that indicates the degree of compatibility, making it easier for the field selection component to prioritize and present options to users.

PR #9985, opened 2025-06-25
[Explore] feat: update table action bar and field selector

Maosaic:
### Description

1. Allow query editor control field selector hide/show
2. Update result text in table action bar 
3. Update download csv button

## Screenshot

Screenshot 2025-06-25 at 1 15 15 PM

## Changelog

- feat: Update table action bar and field selector in Explore

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/application/pages/logs/logs_page.tsx
```diff
@@ -199,15 +202,19 @@ export const LogsPage: React.FC<Partial<Pick<AppMountParameters, 'setHeaderActio
             minSize="260px"
             mode={['collapsible', { position: 'top' }]}
             paddingSize="none"
+            style={{ display: showDataSetFields ? 'block' : 'none' }}
```
joshuali925, line 208: nit. avoid inline styles

src/plugins/explore/public/application/utils/state_management/selectors/index.ts
```diff
@@ -113,10 +113,11 @@ export const selectRows = createSelector([selectResults], (results) => {
 });
 
 export const selectTotalHits = createSelector([selectResults], (results) => {
-  if ((results as any)?.hits?.total?.value !== undefined) {
-    return (results as any).hits.total.value;
-  }
-  return 0;
+  return results?.hits?.total ?? 0;
```
joshuali925: would there be a `total.value`? I know `total` is a number in type, but I've seen `total.value` appearing in some cases before

src/plugins/explore/public/application/legacy/discover/application/components/chart/hits_counter/hits_counter.tsx
```diff
@@ -69,38 +79,25 @@ export function HitsCounter({ hits, showResetButton, onResetQuery, rows }: HitsC
       >
         <EuiFlexItem grow={false}>
           <EuiText size="s">
-            <strong>
+            {hits && (
```
angle943: if we don't have hit count, then none of this shows up? For esxample, in SQL, we don't have hits count. I think we should account for the scenarios when we don't have hits count

src/plugins/explore/public/application/utils/state_management/selectors/index.ts
Maosaic: Base on the `ISearchResult` type and the real state snapshot for PPL, there should be no `total.value`

Screenshot 2025-06-25 at 3 08 47 PM

src/plugins/explore/public/application/pages/logs/logs_page.tsx
angle943: i tried getting joey to change it but he said it shouldn't stop the bug fix. lets catch all these when merging to main

angle943 (dismissed):

LDrago27 (dismissed):

angle943 (approved):

ananzh (approved):

PR #9987, opened 2025-06-25
update title of explore pages

angle943:
- add the flavor to the title:
Screenshot 2025-06-25 at 4 10 24 PM


src/plugins/explore/public/helpers/get_flavor_id/get_flavor_id.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { ExploreFlavor } from '../../../common';
+
+export const getFlavorId = (): ExploreFlavor | null => {
+  const url = window.location.href;
+  const match = url.match(/explore\/(\w+)\/#/);
```
joshuali925: i think `/explore/logs#/` is also valid URL. should we just pass flavor as a prop to each page?

LDrago27: Are we not storing the flavour in the URL storage, which we can query to get it, rather than using a regex

src/plugins/explore/public/application/index.tsx
```diff
@@ -16,7 +16,6 @@ import { IndexPatternProvider } from './components/index_pattern_context';
 import { ExploreFlavor } from '../../common';
 import { TracesPage } from './pages/traces';
 import { MetricsPage } from './pages/metrics';
-import { NoFlavor } from './components/no_view';
```
joshuali925, line 19: thanks for this

src/plugins/explore/public/helpers/get_flavor_id/get_flavor_id.ts
angle943: i was initially against prop drilling because `getFlavorId` can potentially be a util that is used a lot by any component, and i didn't want to prop drill it all the way down to the components that needs it, but URL regex parsing might become problematic. We synced offline, i will prop drill it one level down for now

angle943: i don't think the url storage has the stuff before the hash? doesn't that only have _a, _g, _q?

kavilla: regardless we should use the utils https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/src/core/public/application/utils

angle943: ok nvm, we synced offline again and i changed to using `appId` as that is more reliable. I didn't want consumers tho to worry about passing proper services and whatnot, so converted it to a hook

joshuali925 (approved):

LDrago27 (approved):

src/plugins/explore/public/helpers/get_flavor_id/get_flavor_id.ts
kavilla: @angle943 @joshuali925 , could you provide us the offline sync information for others to have context. i agree with the original stance. since i think i see a lot of prop drilling increasing a lot within this plugin and it might become too unmaintainable

we have https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/src/core/public/application/utils that should be relatively stable.

but to be honest i dont understand this flavor concept. i thought the flavor would be different plugins essential if it was an app path. kind of like the explore plugin was data explorer.

how do users add metrics to a dashboard? will they be modifying our saved object if needed?

does state not clear when navigating between different flavors?

joshuali925: > i thought the flavor would be different plugins essential if it was an app path

they are in the same plugin but different applications

> how do users add metrics to a dashboard? will they be modifying our saved object if needed?

you mean adding the saved explore of metrics flavor to a dashboard? user should just save it like saved search then add to a dashboard

> does state not clear when navigating between different flavors?

state should clear

Issue #9861, opened 2025-06-09
[RFC] OpenSearch Dashboards Banner Plugin


sidwang42:
Placeholder for [RFC] OpenSearch Dashboards Banner Plugin

sidwang42:
cc: @eemijuntunen @seraphjiang @zhongnansu

ashwin-pc:
Can we add more details to this RFC before opening it? Having a placeholder here serves no purpose.

ruanyl:
Instead of introducing another plugin, can we extend the current core banner service by introducing `location` so that user can choose where the banner should be displayed(current location or on the top)?

sidwang42:
> Instead of introducing another plugin, can we extend the current core banner service by introducing `location` so that user can choose where the banner should be displayed(current location or on the top)?

Thanks for the thoughtful suggestion, @ruanyl!

Here are two main reasons why we choose to introduce another plugin:

### 1. Separation of Concerns

The core banner service was intentionally designed with a single, focused responsibility: displaying lists of global system level notifications at the top of the page body. Its implementation reflects this constraint:

```typescript
// Current implementation is tightly coupled to top-of-body rendering
export const BannersList: React.FunctionComponent<Props> = ({ banners$ }) => {
  // ...
  return (
    <div className="osdGlobalBannerList">
      {banners.map((banner) => (
        <BannerItem key={banner.id} banner={banner} />
      ))}
    </div>
  );
};
```
While our goal here is to display a banner at the top of the page—even above the global headers—reusing the core banner service would require repositioning existing layout elements (like shifting the header content down). This creates visual and structural confusion. Extending this service to support additional locations, text URLs, and style settings would blur its purpose and introduce branching logic that increases complexity and maintenance overhead.

### 2. Widespread Usage
The core banner service is used by multiple plugins across the codebase. For example, the data source management plugin uses it extensively. Changing the core would require updates to all these consumers, creating a high risk of regressions.

ruanyl:
Hi @sidwang42 

To address your concerns (1) and (2), I propose extending the existing OverlayBannersService in a backward-compatible way, without introducing a new plugin. Here's the approach:

1. Extend `OverlayBannersService` to support banner `type`
We can add a type parameter to the add method, with a default value of 'sub-hero-banner' to preserve existing behavior:
```typescript
    const service: OverlayBannersStart = {
[...]

PR #9989, opened 2025-06-26
[Banner Plugin] Working static banner with no overlapping

sidwang42:
## Description

This PR is a working static banner with no overlapping and feature flag supported.
It creates the basic structure for the banner plugin.

Banner Plugin Meta Issue:
- [ ]  #9990 

## Issues Resolved

Related Issue:

- [ ] #9861 

## Screenshot

image

## Changelog

- feat: Add static banner plugin and feature flag

## Testing the changes

1. Uncomment the line `banner.enabled: true ` in file `OpenSearch-Dashboards/config/opensearch_dashboards.yml`
2. Run OSD and the banner will be shown on the home page

## Future PRs will add:
1. Advanced settings for customizing the banner
2. Banner Content Configuration via YML files
3. Server-side integration
4. Dynamic CSS to avoid overlapping with other UI elements

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,4 @@ snapshots.js
 # Ignore the lighthousereport generatated
 .lighthouserc.js
 .lhci_output.json
+/.lighthouseci/
--- a/scripts/generate_docs_sidebar.js
+++ b/scripts/generate_docs_sidebar.js
@@ -6,6 +6,7 @@
 
 const fs = require('fs');
 const path = require('path');
+const { execSync } = require('child_process');
 
 const startDirs = [
   {
@@ -27,6 +28,18 @@ const startDirs = [
 
 const sidebarFile = './docs/_sidebar.md'; // Location to save the generated sidebar
 const excludeDirs = ['node_modules', '.git']; // Directories to exclude from the search
+
+const isGitTrackedMarkdownFile = (() => {
+  let tracked;
+  return (filePath) => {
+    if (tracked === undefined) {
+      tracked = new Set(
+        execSync(`git ls-files '*.md'`).toString().trim().split('\n').filter(Boolean)
+      );
+    }
+    return tracked.has(filePath);
+  };
+})();
 
 // Function to recursively find Markdown files and return a nested structure
 function findMarkdownFiles(dir, prefix = '', baseDir = '', recursively = true) {
@@ -61,7 +74,11 @@ function findMarkdownFiles(dir, prefix = '', baseDir = '', recursively = true) {
           results.push({ type: 'directory', name: entry.name, children: nestedResults });
         }
       }
-    } else if (entry.name.endsWith('.md') && entry.name !== '_sidebar.md') {
+    } else if (
+      entry.name.endsWith('.md') &&
+      entry.name !== '_sidebar.md' &&
+      isGitTrackedMarkdownFile(path.join(entry.path, entry.name))
+    ) {
       const docPath = `${prefix}${entry.name}`;
       // Adjust the path based on its base directory ('docs' or 'src')
       const linkPath =


config/opensearch_dashboards.yml
```diff
@@ -379,3 +379,5 @@
 
 # Set the value to true to enable the new UI for savedQueries in Discover
 # data.savedQueriesNewUI.enabled: true
+
```
zhongnansu, line 382: Let's add a comment to explain this settings

src/core/public/chrome/ui/header/header.tsx
```diff
@@ -657,6 +657,7 @@ export function Header({
 
   return (
     <>
+      <div id="globalBanner" />
```
zhongnansu: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/b5767804c131dfecb6f588584ca6f6801b503beb/src/core/public/overlays/banners/banners_list.tsx#L49

Let's consider use other naming to not being mixed with current banners.

src/plugins/banner/common/index.ts
```diff
@@ -0,0 +1,10 @@
+export const PLUGIN_ID = 'banner';
```
zhongnansu, line 12: We need to add Apache 2.0 license to all new code files(no need for the elastic liscense). Could you reference to other core plugins to update yours?
```
/*
 *
 * The OpenSearch Contributors require contributions made to
 * this file be licensed under the Apache-2.0 license or a
 * compatible open source license.
 *
 * Any modifications Copyright OpenSearch Contributors. See
 * GitHub history for details.
 */
```

src/plugins/banner/.eslintrc.js
zhongnansu, line 1: Why do we need this file? Why can't we just depend on the `.eslintrc.js` in the root directory?

src/plugins/banner/.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  root: true,
+  extends: ['@elastic/eslint-config-kibana', 'plugin:@elastic/eui/recommended'],
+  rules: {
+    '@osd/eslint/require-license-header': 'off',
```
zhongnansu: we need liscense headers

src/plugins/banner/.i18nrc.json
zhongnansu, line 1: I don't think we need configure this at plugin level, all OSD components should use the same

src/plugins/banner/common/index.ts
```diff
@@ -0,0 +1,10 @@
...
+export const PLUGIN_NAME = 'banner';
+
+export interface BannerConfig {
+  text: string;
+  color: 'primary' | 'success' | 'warning' | 'danger';
```
zhongnansu: Can we get rid of `danger` color? As UX mentioned, It doesn't make any sense to display a purely user-defined error message.

src/plugins/banner/public/components/global_banner.tsx
```diff
@@ -0,0 +1,157 @@
+import React, { Fragment, useEffect, useState, Suspense, useRef } from 'react';
+import { EuiCallOut, EuiLoadingSpinner } from '@elastic/eui';
```
zhongnansu, line 13: Can we switch to OUI?

src/plugins/banner/public/plugin.ts
```diff
@@ -0,0 +1,81 @@
...
+import { BannerPluginSetup, BannerPluginStart } from './types';
+import { bannerServiceInstance } from './services/banner_service_instance';
+import { GlobalBanner } from './components/global_banner';
+
+export class BannerPlugin implements Plugin<BannerPluginSetup, BannerPluginStart> {
+  constructor(_initializerContext: PluginInitializerContext) {}
+
+  public setup(core: CoreSetup): BannerPluginSetup {
+    // Setup banner with default values
+    bannerServiceInstance.setup({
```
zhongnansu: default values from come from yml. Are you planning to implement this in the future, if that's the case, can we add a TODO here in the code

src/plugins/banner/public/plugin.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    return {};
+  }
+
+  public start(core: CoreStart): BannerPluginStart {
+    // Clear any stored banner height to prevent incorrect values from persisting
+    try {
+      localStorage.removeItem('opensearch-dashboards-banner-height');
+    } catch (e) {
+      // Ignore localStorage errors
```
zhongnansu: error handling is needed, add TODO instead of ignoring

src/plugins/banner/public/services/banner_service_instance.ts
```diff
@@ -0,0 +1,4 @@
+import { BannerService } from './banner_service';
+
+// Create singleton instance of the banner service
```
zhongnansu: Singleton is not necessary. OSD’s plugin lifecycle itself manages service instances for you in a way that is functionally singleton-scoped within your plugin context. Each plugin in OSD is initialized once per process.The setup() and start() lifecycle methods are called once for each plugin instance.

src/plugins/banner/server/config.ts
```diff
@@ -0,0 +1,19 @@
...
+});
+
+export const config: PluginConfigDescriptor<BannerPluginConfig> = {
+  schema: configSchema,
+  exposeToBrowser: {
+    enabled: true,
+  },
+};
+
+export function createConfig(initializerContext: PluginInitializerContext) {
```
zhongnansu: why do we need this? I didn't find any usage of this

src/plugins/banner/server/config.ts
```diff
@@ -0,0 +1,19 @@
+import { schema, TypeOf } from '@osd/config-schema';
+import { PluginConfigDescriptor, PluginInitializerContext } from '../../../core/server';
+
+export type BannerPluginConfig = TypeOf<typeof configSchema>;
+
+export const configSchema = schema.object({
+  enabled: schema.boolean({ defaultValue: false }),
+});
+
+export const config: PluginConfigDescriptor<BannerPluginConfig> = {
```
zhongnansu: what's this used for?

src/plugins/banner/server/config.ts
```diff
@@ -0,0 +1,19 @@
+import { schema, TypeOf } from '@osd/config-schema';
+import { PluginConfigDescriptor, PluginInitializerContext } from '../../../core/server';
+
+export type BannerPluginConfig = TypeOf<typeof configSchema>;
```
zhongnansu: `BannerPluginConfigType`

src/plugins/banner/common/index.ts
sidwang42: Sure

src/plugins/banner/server/config.ts
sidwang42: Deleted for now. Should only be added when needed later.

src/plugins/banner/.i18nrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "translations": [
+    "translations/ja-JP.json",
+    "translations/zh-CN.json"
```
huyaboo: Will we add more translations?

src/core/public/chrome/ui/header/header.tsx
```diff
@@ -657,6 +657,7 @@ export function Header({
 
   return (
     <>
+      <div id="pluginGlobalBanner" />
```
huyaboo: Shouldn't this fall under a feature flag?

src/plugins/banner/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Basic Banner
+
+This initial implementation includes:
+- A basic banner with a default hard-coded message
+- Proper positioning and layout adjustments
+- Dismiss functionality
+- Markdown support
+
+### Future Enhancements
```
huyaboo: Nit: It's probably better to just leave these as part of the GH issues rather than committing them in the `README.md`

src/plugins/banner/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Future PRs will add:
+1. Advanced settings for customizing the banner
+2. Configuration via YML files
+3. Server-side integration
+4. Dynamic CSS to avoid overlapping with other UI elements
+
+## Technical Details
+
+The banner uses a CSS variable (`--global-banner-height`) to dynamically adjust the layout based on the banner's height. This ensures that the OpenSearch Dashboards header and content are properly positioned below the banner.
```
huyaboo: Could we extract these details under the `/docs` directory instead?

src/plugins/banner/README.md
```diff
@@ -0,0 +1,37 @@
+# Banner Plugin
+
+This plugin adds a global banner to OpenSearch Dashboards that can display important announcements or notifications to users.
+
```
huyaboo, line 4:
````suggestion
## Configuration
```yaml
# Enables the plugin
banner.enabled: true

# Add other configurations here 
banner.foo: "bar"
```
````

(non-blocking) IMO, it would be nice to have configurations in the `README.md` to make it easy to use

File: src/plugins/banner/common/index.ts (hunk 3 above)
Comment (huyaboo) [reply src/plugins/banner/common/index.ts:12]: I believe running the lint script should automatically add these headers

File: src/plugins/banner/server/config.ts [hunk 18]
```diff
@@ -0,0 +1,18 @@
...
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { schema, TypeOf } from '@osd/config-schema';
+
+export type BannerPluginConfigType = TypeOf<typeof configSchema>;
+
+export const configSchema = schema.object({
+  enabled: schema.boolean({ defaultValue: false }),
```
huyaboo, line 17: Do we not have a setting to change the text from the yaml file?

src/plugins/banner/server/types.ts
```diff
@@ -0,0 +1,17 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+export interface BannerPluginSetup {
+  bannerEnabled: () => boolean;
```
huyaboo, line 13: Why are we exposing the `bannerEnabled()` flag here?

src/plugins/banner/public/plugin.ts
```diff
@@ -0,0 +1,93 @@
...
+import { bannerServiceInstance } from './services/banner_service_instance';
+import { GlobalBanner } from './components/global_banner';
+
+export class BannerPlugin implements Plugin<BannerPluginSetup, BannerPluginStart> {
+  constructor(_initializerContext: PluginInitializerContext) {}
+
+  public setup(core: CoreSetup): BannerPluginSetup {
+    // Setup banner with default values
+    bannerServiceInstance.setup({
+      text: 'This is an important announcement for all users.',
```
huyaboo: So is this line committed everytime we need to make new announcements?

src/plugins/banner/public/plugin.ts
```diff
@@ -0,0 +1,93 @@
...
+    return {};
+  }
+
+  public start(core: CoreStart): BannerPluginStart {
+    // Clear any stored banner height to prevent incorrect values from persisting
+    try {
+      localStorage.removeItem('opensearch-dashboards-banner-height');
+    } catch (e) {
+      // TODO
+      // Ignore localStorage errors for now
```
huyaboo: Can you elaborate on this? Why should we ignore errors?

src/plugins/banner/public/plugin.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    // Get the current banner config to determine initial height
+    const currentConfig = bannerServiceInstance.getCurrentConfig();
+
+    // Set initial height based on banner visibility
+    const initialHeight = currentConfig.isVisible ? '40px' : '0px';
```
huyaboo: Nit: I'm seeing `40` and `0` get used a lot. Can we extract this in `common/constants.ts` to give more meaning to these values?

src/plugins/banner/public/plugin.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    // Set an initial banner height to ensure headers are positioned correctly
+    // This prevents layout shifts when the banner is enabled/disabled
+    document.documentElement.style.setProperty('--global-banner-height', initialHeight);
+
+    // Add to the globalBanner container immediately and retry if needed
+    const renderBanner = () => {
```
huyaboo: nit: is there a way we can extract this function into a separate file/util?

src/plugins/banner/public/components/global_banner.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { BannerConfig } from '../../common';
+import { BannerService } from '../services/banner_service';
+import { bannerServiceInstance } from '../services/banner_service_instance';
+
+const ReactMarkdownLazy = React.lazy(() => import('react-markdown'));
+
+// Custom link renderer for markdown links to open in new tab
+const LinkRenderer = (props: any) => {
```
huyaboo: Nit: can this be extracted into a separate components file like `link_renderer.tsx`? There seems to be multiple components when it should just contain the `GlobalBanner`

src/plugins/banner/public/components/global_banner.tsx
```diff
@@ -0,0 +1,167 @@
...
+        setTimeout(() => {
+          document.documentElement.style.transition = '';
+        }, 300);
+      }
+    };
+  }, [bannerConfig]);
+
+  // Handle close button click
+  const handleDismiss = () => {
+    // Update the banner config to hide the banner
```
huyaboo: Nit: instead of leaving multiple comments, could we rename this to something concise like `hideBanner()`?

src/plugins/banner/public/plugin.ts
sidwang42: > So is this line committed everytime we need to make new announcements?

Will support yml and advnaced settings in the next pr. Add TODO here for now

src/plugins/banner/.i18nrc.json
sidwang42: Will add the i18n support later, delete related files for now.

src/plugins/banner/server/types.ts
sidwang42: > Why are we exposing the `bannerEnabled()` flag here?

The flag is used now in header.tsx as feature flag. Not sure if that's the right way to do this

src/plugins/banner/public/plugin.ts
sidwang42: > Can you elaborate on this? Why should we ignore errors?

Realized that the localStorage is not needed for now.

src/core/public/chrome/ui/header/header.tsx
sidwang42: Fixed. Thanks for pointing out.

src/plugins/banner/server/config.ts
sidwang42: > Do we not have a setting to change the text from the yaml file?

Will be added in the next pr.

src/plugins/banner/public/plugin.ts
sidwang42: > nit: is there a way we can extract this function into a separate file/util?

Leave it like this for now. Will consider extract it into a separate file in the future as the filer grows longer : )

config/opensearch_dashboards.yml
```diff
@@ -379,3 +379,6 @@
...
+# Set the value to true to enable the banner plugin
+banner.enabled: true
```
zhongnansu: this should by default be false, and commented out in yml.

src/plugins/banner/public/services/banner_service_instance.ts
```diff
@@ -0,0 +1,15 @@
...
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { BannerService } from './banner_service';
+
+// Export a default instance
+export const bannerServiceInstance = new BannerService();
```
zhongnansu: Why export here? Can we create the instance within the plugin lifecycle to make sure it's only created once

src/plugins/banner/public/services/banner_service.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { BehaviorSubject } from 'rxjs';
+import { BannerConfig } from '../../common';
+
+export class BannerService {
+  private bannerConfig$ = new BehaviorSubject<BannerConfig>({
+    text: 'Default banner text',
```
zhongnansu: can we move the default values to config schema, similar to `enabled`? Since all defaults should be coming from config.ts or yml.

src/plugins/banner/public/plugin.ts
```diff
@@ -0,0 +1,88 @@
...
+      } else {
+        // If the container isn't available yet, retry after a short delay
+        setTimeout(renderBanner, 50);
+      }
+    };
+
+    // Start the rendering process
+    renderBanner();
+
+    // Make the banner service available to other plugins
```
zhongnansu: why this needs to be exposed to other plugins?

src/plugins/banner/public/plugin.ts
```diff
@@ -0,0 +1,88 @@
...
+    return {
+      bannerService: bannerServiceInstance,
+    };
+  }
+
+  public stop() {
+    // Clean up the banner when the plugin stops
+    const globalBannerContainer = document.getElementById('globalBanner');
```
zhongnansu: do you need to update the banner id as well? Better make it a constant?

src/plugins/banner/public/components/global_banner.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+interface GlobalBannerProps {
+  bannerService?: BannerService;
+}
+
+export const GlobalBanner: React.FC<GlobalBannerProps> = ({ bannerService }) => {
+  const [bannerConfig, setBannerConfig] = useState<BannerConfig | null>(null);
+  const bannerRef = useRef<HTMLDivElement>(null);
+
+  const actualBannerService = bannerServiceInstance;
```
zhongnansu: can we have better naming `actualBannerService` is confusing

src/plugins/banner/public/plugin.ts
zhongnansu: let's extract it. Plugin.ts file is critical as it's the entry point, and should only contain high level initiation.

src/plugins/banner/public/components/global_banner.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+  useEffect(() => {
+    const subscription = actualBannerService.getBannerConfig$().subscribe((config) => {
```
zhongnansu: you may want to add TODOs, if in the future those settings are fetched from API

src/plugins/banner/public/components/global_banner.tsx
sidwang42: noted

src/core/public/chrome/ui/header/header.tsx
```diff
@@ -655,8 +658,15 @@ export function Header({
     return useApplicationHeader ? renderApplicationHeader() : renderPageHeader();
   };
 
+  // Get the banner plugin configuration
+  const bannerPluginConfig = injectedMetadata
+    ?.getPlugins()
+    ?.find((plugin: { id: string }) => plugin.id === 'banner')?.config;
```
zhongnansu, line 664: can we make `banner` a constant and defined it in plugin and import here. Let's avoid hardcoded text as much as possible.

src/plugins/banner/public/components/global_banner.tsx
```diff
@@ -0,0 +1,152 @@
...
+  // Hide banner when close button is clicked
+  const hideBanner = () => {
+    bannerService.updateBannerConfig({
+      isVisible: false,
+    });
+  };
+
+  if (!bannerConfig || !bannerConfig.isVisible) {
+    return null;
+  }
```
xinruiba, line 115: Why do we need to check banner config here?

src/plugins/banner/public/components/global_banner.tsx
```diff
@@ -0,0 +1,152 @@
...
+  }, [bannerService]);
+
+  // Update the CSS variable with the banner's height
+  useEffect(() => {
+    // Add a smooth transition when changing banner visibility
+    document.documentElement.style.transition = 'padding-top 0.3s ease';
+
+    // If banner is not visible, set height to 0
+    if (!bannerConfig?.isVisible) {
+      document.documentElement.style.setProperty('--global-banner-height', HIDDEN_BANNER_HEIGHT);
```
xinruiba, line 49: We're currently dismissing the banner by setting its height to 0, which feels a bit hacky. Is there a cleaner way to remove or unmount the banner from the DOM directly?

sidwang42: ~~Thanks for pointing out. I'm currently working on a solution which can disable the .css of the plugin and also unmount the banner, which would be more elegant than this approach.~~

After exploring various alternatives, the current approach appears to be the most optimal as it minimizes changes to header.tsx. Trying to unmount the banner from the DOM will require more modification.

Also, the line here is only setting the --global-banner-height to zero to push the other headers back to the top. The actual dismissing of the banner is:
`  
if (!bannerConfig || !bannerConfig.isVisible) {
    return null;
  }
`

src/plugins/banner/public/components/global_banner.tsx
sidwang42: > Why do we need to check banner config here?

The bannerConfig might be null here until the async subscription provides data.

src/core/public/chrome/ui/header/header.tsx
sidwang42: That would make the header dependent on the plugin by importing a constant from it, effectively treating the plugin as part of the core.

The current approach is more flexible—if the plugin name changes or is removed, we simply avoid injecting this part into the DOM.

ZilongX:
is there a targeting release for the banner plugin ?

zhongnansu:
@sidwang42 Let's add a task as github issue to track adding test for the plugin itself

sidwang42:
> @sidwang42 Let's add a task as github issue to track adding test for the plugin itself

Here is the one for tracking all the banner plugin related PRs and future PR plans: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9990

sidwang42:
> is there a targeting release for the banner plugin ?

It is targeting 3.2.0

src/plugins/banner/public/plugin.ts
```diff
@@ -0,0 +1,55 @@
...
+import { renderBanner, unmountBanner, setInitialBannerHeight } from './services/render_banner';
+
+export class BannerPlugin implements Plugin<BannerPluginSetup, BannerPluginStart> {
+  private readonly bannerService = new BannerService();
+
+  constructor(private readonly initializerContext: PluginInitializerContext) {}
+
+  public setup(core: CoreSetup): BannerPluginSetup {
+    // Get configuration from server
+    const config = this.initializerContext.config.get() as any;
```
huyaboo, line 24: Minor nit: this could be typed as the configtype right? We should avoid `any`

xinruiba (approved):

src/plugins/banner/public/plugin.ts
sidwang42: Noted.

kavilla:
wouldn't this collide with the banner service from the overlay service? https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/overlays/overlay_service.ts#L82

have we verified this interaction?

src/core/public/chrome/ui/header/header.tsx
kavilla: how come we didn't utilize the banner overlay https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/overlays/overlay_service.ts#L82

kavilla:
did this pass the ci? im not able to see the run. it should have failed the linter ci step

src/plugins/banner/docs/technical_details.md
kavilla, line 1: we should verify im pretty sure this will bundle markdown file with the release distribution. we should put this in a readme or just in the docs folder

src/plugins/banner/public/components/global_banner.tsx
```diff
@@ -0,0 +1,152 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
```
kavilla, line 4: here is the right license header

```
/*
 */
 ```

**Comment by sidwang42** [comment]
> wouldn't this collide with the banner service from the overlay service? https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/overlays/overlay_service.ts#L82
> 
> have we verified this interaction?

Good question — this potential overlap was discussed in the RFC here: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9861. Since the new banner plugin is focused solely on header-level display and doesn't register with the overlay service, there shouldn't be any direct conflict. But happy to double-check the interaction during integration testing to be safe.

File: src/plugins/banner/public/index.scss [hunk 38]
```diff
@@ -0,0 +1,65 @@
...
+  width: 100%;
+  min-height: var(--global-banner-height);
+}
+
+/**
+ * 3. Smooth transitions - Prevents jarring layout shifts when banner height changes
+ */
+/* stylelint-disable @osd/stylelint/no_modifying_global_selectors */
+.euiHeader {
+  transition: top 0.3s ease;
```
kavilla, line 32: we should revert this PR. this will modify everything once loaded

sidwang42:
@kavilla 
I’ve opened a new PR #10125 to address the issue you pointed out — would love to get your feedback when you have a moment. Thanks!

PR #9991, opened 2025-06-26
[Explore]add visualization to dashboard from Vis tab

Qxisylolo:
### Description
This pr adds visualization and log table to dashboard from Vis tab.

## Screenshot

1. in a new save explore and try to add visualization to new dashboard from Vis tab

https://github.com/user-attachments/assets/1e6455be-92ad-4fc0-8ff6-04a3d7c5d18b

2. in an existing save explore and try to add visualization to new dashboard from Vis tab
a note will be displayed to notify user that new query will update existing save explore

https://github.com/user-attachments/assets/65960cd4-f256-405d-8132-f05bc8e74a9a

4.  in an existing save explore and try to add visualization to existing dashboard from Vis tab

https://github.com/user-attachments/assets/36d494eb-c699-4646-8764-2e9d8f718fd1

## Changelog
- feat: add visualization to dashboard from Vis tab
 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/application/utils/hooks/use_saved_explore.ts
```diff
@@ -53,12 +56,8 @@ export const useSavedExplore = (exploreIdFromUrl?: string) => {
           }
         }
 
-        // Update savedSearch to store just the ID (like discover)
-        // TODO: remove this once legacy state is not consumed any more
-        const legacyStateWithId = {
-          savedSearch: exploreIdFromUrl, // Store the ID, not the object
-        };
-        dispatch(setState(legacyStateWithId));
```
raintygao: Any reasons that we removed this? `legacyStateWithId` could be consumed by others.

src/plugins/explore/public/application/utils/hooks/use_saved_explore.ts
```diff
@@ -53,12 +56,8 @@ export const useSavedExplore = (exploreIdFromUrl?: string) => {
...
+        // Store the name to ensure state consistency for both top nav and add to dashboard button
+        dispatch(setSavedSearchName(title));
```
raintygao: Why ` dispatch(setSavedSearchName(title));` is not along with `dispatch(setSavedSearch(exploreIdFromUrl));`? They look like doing a same thing, put them together will have more readability.

src/plugins/explore/public/application/utils/state_management/slices/legacy_slice.ts
```diff
@@ -50,6 +53,9 @@ const legacySlice = createSlice({
     setSavedSearch: (state, action: PayloadAction<string | undefined>) => {
       state.savedSearch = action.payload;
     },
+    setSavedSearchName: (state, action: PayloadAction<string | undefined>) => {
+      state.savedSearchName = action.payload;
+    },
```
raintygao: Please stop adding new property to legacyState, I think you could maintain this in new state of explore.

src/plugins/explore/public/components/visualizations/visualization_container.tsx
```diff
@@ -27,23 +28,42 @@ import {
...
 
+export type SaveResult = { id?: string } | { error: Error };
+
 export const VisualizationContainer = () => {
   const { services } = useOpenSearchDashboards<ExploreServices>();
   const dispatch = useDispatch();
   const {
-    data: {
-      query: { filterManager, queryString, timefilter },
-    },
```
raintygao, line 42: Any reasons that we remove this?

src/plugins/explore/public/components/visualizations/visualization_container.tsx
```diff
@@ -27,23 +28,42 @@ import {
...
+import { useSyncQueryStateWithUrl } from '../../../../data/public';
+import { ExecutionContextSearch } from '../../../../expressions/common/';
+import { SaveAndAddButtonWithModal } from './add_to_dashboard_button';
 import {
   setStyleOptions,
   setChartType as setSelectedChartType,
 } from '../../application/utils/state_management/slices/ui_slice';
 import { defaultPrepareQuery } from '../../application/utils/state_management/actions/query_actions';
 
+export type SaveResult = { id?: string } | { error: Error };
```
raintygao: Don't export this if this is not imported by others.

src/plugins/explore/public/components/visualizations/add_to_dashboard_button.tsx
```diff
@@ -0,0 +1,248 @@
...
+     * - The component mounts,and the saved explore ID is already retrieved from the URL.
+     * - savedExploreId changes
+     * - savedExploreName changes externally (e.g., renamed via top nav)
+     */
+    const loadSavedExplore = async () => {
+      const savedObject = await getSavedExploreById(savedExploreId);
+      setSavedExplore(savedObject);
+    };
+    loadSavedExplore();
+  }, [savedExploreId, getSavedExploreById, savedExploreName]);
```
raintygao: Do we really need this?

src/plugins/explore/public/application/utils/hooks/use_saved_explore.ts
Qxisylolo: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10002. in this new pr, they remove setState. so I remove it in this pr as well.

src/plugins/explore/public/components/visualizations/add_to_dashboard_button.tsx
```diff
@@ -0,0 +1,248 @@
...
+          dashboardId = await addToDashboard(dashboard, { id, type: 'explore' }, mode, {
+            newDashboardName,
+            createDashboardOptions,
+          });
+        } else {
+          dashboardId = await addToDashboard(dashboard, { id, type: 'explore' }, mode, {
+            existingDashboardId: selectDashboard!.id,
+          });
+        }
+      }
```
raintygao:
```suggestion
      if (id) {
      let props;
        if (mode === 'new') {
          props =  {
            newDashboardName,
            createDashboardOptions,
          });
        } else {
         props = {
            existingDashboardId: selectDashboard!.id,
          });
        }
        dashboardId = await addToDashboard(dashboard, { id, type: 'explore' }, mode, props)
      }
```

src/plugins/explore/public/components/visualizations/add_to_dashboard_modal.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+  const handleTitleDuplicate = () => {
+    setisLoading(false);
+    setIsTitleDuplicate(true);
+  };
+
+  const handleDashboardDuplicate = () => {
+    setisLoading(false);
+    setIsDashboardDuplicate(true);
+  };
```
raintygao, line 99: handleTitleorDashboardDuplicate?

src/plugins/explore/public/application/utils/hooks/use_saved_explore.ts
Qxisylolo: no actually. setSavedSearchName is title and setSavedSearch  is id. There is a case where the user can rename a saved Explore in the top navigation. And after clicking "Add to Dashboard" button, it should display a comment notifying the user that the new query will update the existing saved Explore (the name will be shown). Therefore, we need to reload the saved Explore when the "Add to Dashboard" button is clicked if title changes

src/plugins/explore/public/components/visualizations/visualization_container.tsx
Qxisylolo: I haven't removed it — it just stopped destruction because we also need data.query for:
```
  const { startSyncingQueryStateWithUrl } = useSyncQueryStateWithUrl(
    data.query,
    osdUrlStateStorage
  );
  ```. 
If we destruct the state into query, it will conflict with:
```
const query = useSelector((state: RootState) => state.query);
```

File: src/plugins/explore/public/application/legacy/discover/application/components/top_nav/get_top_nav_links.tsx [hunk 9]
```diff
@@ -66,7 +67,7 @@ export const getTopNavLinks = (
       });
       // TODO this behavior is different from Discover. Clicking New in Explore
       // only closes the saved search and does not change the query.
-      store.dispatch(setSavedSearch(undefined));
+      store!.dispatch(setSavedSearch(undefined));
```
wanglam: Is this for fixing the type error? I prefer to check if store is not undefined first instead of adding "!" here.

src/plugins/explore/public/components/visualizations/visualization_container.tsx
Qxisylolo: oh, sorry this is an useless line. delete it

src/plugins/explore/public/embeddable/explore_embeddable.tsx
```diff
@@ -227,7 +227,7 @@ export class ExploreEmbeddable
         dateColumns: VisColumn[],
         styleOpts: any
       ) => {
-        return rule.toExpression!(
+        return rule?.toExpression!(
```
wanglam: Can we change to `rule?. toExpression?.(`

src/plugins/explore/public/components/visualizations/add_to_dashboard_modal.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+  useEffect(() => {
+    const loadAllDashboards = async () => {
+      const res = await savedObjectsClient.find({
+        type: 'dashboard',
+      });
+      const dashboards = res.savedObjects;
+      setExistingDashboard(dashboards as DashboardInterface[]);
+      setSelectDashboard(
+        dashboards.length > 0 ? (res.savedObjects[0] as DashboardInterface) : null
```
wanglam: Nit: Prefer use `dashboards[0]` instead of `res.savedObjects[0]` here.

src/plugins/explore/public/components/visualizations/add_to_dashboard_button.tsx
```diff
@@ -0,0 +1,248 @@
...
+    <>
+      <EuiButton size="s" onClick={() => setShowAddToDashboardModal(true)}>
+        {i18n.translate('explore.addtoDashboardButton.name', {
+          defaultMessage: 'Add to dashboard',
+        })}
+      </EuiButton>
+      {showAddToDashboardModal && savedExplore && (
+        <AddToDashboardModal
+          savedExplore={
+            savedExplore ? saveStateToSavedObject(savedExplore, uiState, indexPattern) : undefined
```
wanglam: Why we need to check if `savedExplore` exists here? I think it should always been exists here..

src/plugins/explore/public/components/visualizations/add_to_dashboard_button.tsx
Qxisylolo: yes, i think we need it. unless there is a global state for savedExplore

src/plugins/explore/public/components/visualizations/add_to_dashboard_button.tsx
Qxisylolo: fixed ,thanks!

src/plugins/explore/public/application/utils/hooks/use_saved_explore.ts
angle943: i removed it because the `setState` name is too generic. In my PRT i replace this with `dispatch(setSavedSearch(legacyStateWithId))` (which essentially is the same thing as above)

Do we no longer need that logic?

src/plugins/explore/public/application/utils/state_management/slices/legacy_slice.ts
angle943: piggy backing off this comment, couple question:

- has the new "saved explore" schema been merged to this branch?
- If saved expore is ready, can we all savedSearch/savedExplore property out of the `legacySlice` and move it to another slice? Now what slice is a question, perhaps saved search should have its own slice? Feel free to start a discussion with this in slack, @ananzh has the most context in how the slices were set up

src/plugins/explore/public/components/visualizations/add_to_dashboard_button.tsx
angle943: i'm hoping some work can be done as part of saved explore/search dev on cleaning out the saved explore service. Currently its mostly copied and pasted from old discover.

Like why is this `getSavedExploreById` an async function?

Another question, do do we even need this `savedExploreService`? We now have saved explore data being retrieved from that service as well as have the saved explore in redux. This is a bug waiting to happen.

Perhaps we can think a bit about this?

src/plugins/explore/public/application/utils/hooks/use_saved_explore.ts
Qxisylolo: we still need this logic,  I just ahead it right after retrieving savedExploreObject
```
     const savedExploreObject = await getSavedExploreById(exploreIdFromUrl);
      // store the id(if it is undefined, then new save explore)
      dispatch(setSavedSearch(exploreIdFromUrl));
      if (savedExploreObject?.id) {
 ....
      }
   ```
This ensures we can identify new saved explore(id is undefined), or context will always remember the latest saved explore. it may result in an inconsistent state between the button and the top navigation

File: src/plugins/explore/public/components/visualizations/add_to_dashboard_button.tsx (hunk 6 above)
Comment (Qxisylolo) [reply src/plugins/explore/public/components/visualizations/add_to_dashboard_button.tsx]: I have extracted the save logic in a new function and perhaps we could use this to handle the logic of saving(if it is ok, i will replace the save logic in top nav)? 
Now, we only need savedExploreService[getSavedExploreIdFromUrl] to load the save explore from url then we can save explore in redux.

File: src/plugins/explore/public/application/utils/state_management/slices/legacy_slice.ts (hunk 3 above)
Comment (Qxisylolo) [reply src/plugins/explore/public/application/utils/state_management/slices/legacy_slice.ts]: Thanks for the comments
* yes, in pr: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9964/files#diff-c8593b3bf58363de9925258795094a62b741dcccf4e25158b2c14e5014568ebf.  cc @raintygao 
*  currently I temporarily save it in ui slice. But I think there will be a better choice.

**Review by ruanyl — DISMISSED** [review]

File: src/plugins/explore/public/application/utils/state_management/slices/ui/ui_slice.ts [hunk 13]
```diff
@@ -4,6 +4,10 @@
  */
 
 import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import {
```
ruanyl: Nit: these imports are not used

ruanyl (dismissed):

src/plugins/explore/public/application/utils/state_management/slices/ui/ui_slice.ts
Qxisylolo: thanks for the comments, fixed it and move save explore file to helper folder

ruanyl (approved):

raintygao (approved):

PR #10000, opened 2025-06-26
Fix Autocomplete Bugs 

LDrago27:
### Description

- Devalue fields in the Autosuggestion Wizard that begin with _ (Mostly are meta fields)
- Close the Autocomplete Suggestion Wizard when using CMD + Enter to run the query 
- Switching Datasets using the Dataset picker was still showing the fields from the previous dataset
( Although the React Component was being updated the provideCompletionItem Method was using the older reference for indexPattern and dataset that was causing data from previous dataset to show up. Currently Modified the method to fetch the current dataset and indexPattern from queryString manager before firing the actual autocomplete query. This ensures that we are always getting the latest updates)
- Minor Cleanups

## Changelog

- fix : Resolve Autocomplete bugs 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
+++ b/src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
@@ -150,7 +150,7 @@ export const getSimplifiedPPLSuggestions = async ({
           availableFields,
           (f: string) => `${f} `,
           (f: string) => {
-            return f.startsWith('_') ? `9` : `3`;
+            return f.startsWith('_') ? `9` : `3`; // This devalues all the Field Names that start _ so that appear further down the autosuggest wizard
           }
         )
       );
--- a/src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
+++ b/src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
@@ -146,7 +146,13 @@ export const getSimplifiedPPLSuggestions = async ({
       );
 
       finalSuggestions.push(
-        ...formatAvailableFieldsToSuggestions(availableFields, (f: string) => `${f} `, '3')
+        ...formatAvailableFieldsToSuggestions(
+          availableFields,
+          (f: string) => `${f} `,
+          (f: string) => {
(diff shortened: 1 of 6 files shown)


src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
```diff
@@ -146,7 +146,13 @@ export const getSimplifiedPPLSuggestions = async ({
       );
 
       finalSuggestions.push(
-        ...formatAvailableFieldsToSuggestions(availableFields, (f: string) => `${f} `, '3')
+        ...formatAvailableFieldsToSuggestions(
+          availableFields,
+          (f: string) => `${f} `,
+          (f: string) => {
+            return f.startsWith('_') ? `9` : `3`;
```
angle943: lets add a comment here why we're doing this, this is very hard to understand from looking at the code.

LDrago27: Sure. Added a comment to explain the rational

angle943 (approved):

src/plugins/explore/public/components/query_panel/query_panel.tsx
```diff
@@ -131,14 +129,31 @@ const QueryPanel: React.FC<QueryPanelProps> = ({ datePickerRef, services, indexP
...
 
-        // Use centralized IndexPattern from context
+        // Get the current dataset from Query Service to avoid stale closure values
+        const currentDataset = services?.data?.query?.queryString?.getQuery().dataset;
+
+        // Get the current indexPattern from services to avoid stale closure values
+        let currentIndexPattern = indexPattern;
+        if (currentDataset) {
+          try {
+            currentIndexPattern = await services?.indexPatterns?.get(
```
ananzh, line 139: This is a temp solution right? I think we still need to ensure index pattern is updated right?

LDrago27: Actually I checked the indexPattern gets updated within the React Component. However the updated indexPattern is not getting passed into the method.

PR #10007, opened 2025-06-27
top nav refactor

angle943:
- this is essentially a refactoring, not much logic change other than some organizations, such as moving `get_top_nav_links` out of `legacy` and splitting it up into smaller pieces, and adding unit tests and storybook to where i deem the components are in "final" state. The ones without unit tests I have added a TODO comment.


src/plugins/explore/public/components/top_nav/top_nav_links/get_top_nav_links.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  if (share) {
+    topNavLinks.push({
+      ...shareTopNavData,
+      run: getShareButtonRun(services, savedExplore),
+    });
+  }
+
+  return topNavLinks;
+};
```
angle943, line 57: before this file used a `Map` instead of an array as i am doing now. I thought it was very unnecessary to use a Map for this simple functionality

src/plugins/explore/public/types.ts
```diff
@@ -154,7 +154,7 @@ export interface ExploreServices {
   overlays: CoreStart['overlays'];
 
   // From DataExplorerServices (since Explore incorporates DataExplorer functionality)
-  store: ReduxStore; // Redux store
+  store: AppStore; // Redux store
```
angle943, line 157: This AppStore fixes type errors. ReduxStore is very generic and doe not know the context of our reducers and state

Maosaic (approved):

joshuali925 (approved):

PR #10015, opened 2025-06-28
ppl reference

ruchidh:
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/query_panel/components/footer/index.scss
+++ b/src/plugins/explore/public/components/query_panel/components/footer/index.scss
@@ -6,7 +6,7 @@
 .queryPanel__footer {
 
   // stylelint-disable-next-line @osd/stylelint/no_modifying_global_selectors
-  .euiFlexGroup--gutterExtraSmall>.euiFlexItem {
+  .euiFlexGroup--gutterExtraSmall > .euiFlexItem {
     margin: 0;
   }
 


src/plugins/explore/public/components/query_panel/components/footer/language_reference/ppl_language.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { LanguageReference } from '../../../../../../../data/public';
+import { useOpenSearchDashboards } from '../../../../../../../opensearch_dashboards_react/public';
+
+const PPLReference = () => {
+  const opensearchDashboards = useOpenSearchDashboards<IDataPluginServices>();
+  const pplDocs = opensearchDashboards.services.docLinks?.links.noDocumentation.ppl.base;
+  const limitationDocs =
+    opensearchDashboards.services.docLinks?.links.noDocumentation.sqlPplLimitation.base;
+  const pplFullName = (
+    <FormattedMessage id="queryEnhancements.queryBar.pplFullLanguageName" defaultMessage="PPL" />
```
angle943, line 18: i think "PPL" shouldn't be translated

src/plugins/explore/public/components/query_panel/components/footer/language_reference/ppl_language.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <FormattedMessage id="queryEnhancements.queryBar.pplLimitationDoc" defaultMessage="here" />
+  );
+
+  return (
+    <>
+      <EuiText size="s">
+        <p>
+          <FormattedMessage
+            id="queryEnhancements.queryBar.pplSyntaxOptionsDescription"
+            defaultMessage="Piped Processing Language ({pplDocsLink}) is a query language that focuses on processing data in a sequential, step-by-step manner. OpenSearch SQL/PPL language limitations can be found {limitationDocsLink}."
```
angle943, line 30: I think you got this from the previous implementation right? If so it is out of scope, but this translation is problematic as it assumes that "OpenSearch SQL/PPL language limitations can be found" + "here" in that order is grammatically correct in all languages

src/plugins/explore/public/components/query_panel/components/footer/show_input_type.tsx
```diff
@@ -41,7 +43,8 @@ export const ShowInputType: React.FC<ShowInputTypeProps> = ({
       className="queryPanel__footer__showInputType"
       data-test-subj="queryPanelFooterShowInputType"
     >
-      {getDisplayText}{' '}
+      {getDisplayText}
+      {getDisplayText.includes(LanguageType.PPL) && pplLanguageReference(LanguageType.PPL)}
```
angle943, line 47: the function `pplLanguageReference` gets `PPL` as an argument? Can we either do:

- make `pplLanguageReference()` generic (by removing `ppl` in the name) and having a switch statement that switches based on the `selectedLanguage` argument and displays the correct language reference (recommended)
- Remove the `selectedLanguage` argument since it already assumes that it is on PPL (via the name)

src/plugins/explore/public/components/query_panel/components/footer/language_reference/ppl_language.tsx
joshuali925: actually i don't think it requires that order, translators can still do something like

`See {limitationDocsLink} for OpenSearch SQL/PPL language limitations` (in another language)

angle943: ah you are rigtht for this use case, my bad

angle943:
implemented here: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10066

PR #10024, opened 2025-06-30
[Explore] Add execute dispatch to histogram and quick select

sejli:
### Description

Adds the execute query logic to the histogram and date picker quick select menu

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/legacy/discover/application/components/chart/chart.tsx
+++ b/src/plugins/explore/public/application/legacy/discover/application/components/chart/chart.tsx
@@ -61,8 +61,14 @@ export const DiscoverChart = ({
         to: moment(ranges.to).toISOString(),
         mode: 'absolute',
       });
+
+      // EXPLICIT cache clear - same pattern as other triggers
+      dispatch(clearResults());
+
+      // Execute queries - interval will be picked up from Redux state
+      dispatch(executeQueries({ services }));
     },
-    [data]
+    [data, dispatch, services]
   );
   const [isCollapsed, setIsCollapsed] = useState(false);
 
--- a/src/plugins/explore/public/components/query_panel/query_panel.tsx
+++ b/src/plugins/explore/public/components/query_panel/query_panel.tsx
@@ -78,8 +78,14 @@ const QueryPanel: React.FC<QueryPanelProps> = ({ services, indexPattern }) => {
       if (timefilter) {
         timefilter.setTime(newTimeRange);
       }
+
+      // EXPLICIT cache clear - same pattern as other triggers
+      dispatch(clearResults());
+
+      // Execute queries - interval will be picked up from Redux state
+      dispatch(executeQueries({ services }));
     },
-    [timefilter]
+    [timefilter, dispatch, services]
   );
 
   const handleRefreshChange = useCallback(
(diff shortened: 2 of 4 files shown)


src/plugins/explore/public/application/legacy/discover/application/components/chart/chart.tsx
```diff
@@ -61,8 +61,14 @@ export const DiscoverChart = ({
         to: moment(ranges.to).toISOString(),
         mode: 'absolute',
       });
+
+      // EXPLICIT cache clear - same pattern as other triggers
```
ananzh, line 65: Do we still need timefilterUpdateHandler to handle query execution? Currently we timefilter update is triggered by clicking run button in datepicker in query panel. This will build searchSource using the latest timefilter. 

Why we need separate handle for chart here? Is there any bugs that you try to resolve that I misunderstand?

src/plugins/explore/public/components/query_panel/query_panel.tsx
```diff
@@ -82,8 +82,14 @@ const QueryPanel: React.FC<QueryPanelProps> = ({ services, indexPattern }) => {
       if (timefilter) {
         timefilter.setTime(newTimeRange);
       }
+
```
ananzh, line 81: Same question here. Timefilter update is handled by handleRun in src/plugins/explore/public/components/query_panel/query_panel.tsx

src/plugins/explore/public/application/legacy/discover/application/components/chart/chart.tsx
sejli: `timeFilterUpdateHandler` is run when [an element is clicked](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/query_explore/src/plugins/explore/public/application/legacy/discover/application/components/chart/histogram/histogram.tsx#L175) in the histogram. Currently, when something is clicked in the histogram, the timefilter is updated correctly but the query isn't run. Adding a handle here would make allow the search to run after the timefilter is updated.
Going from:
```
histogram clicked -> timefilter set
```
to:
```
histogram clicked -> timefilter set -> query executed
```

src/plugins/explore/public/components/query_panel/query_panel.tsx
sejli: Similar to the histogram, the `handleTimeChange` is passed down to the [`DateTimeRangePicker`](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/query_explore/src/plugins/explore/public/components/query_panel/components/footer/date_time_selector.tsx#L37), which is used when user updates the time range in the `EuiSuperDatePicker`. Previously, this would not run the query after being selected, adding the dispatches here allows the query to run after.
```
date range clicked -> time range set -> query executed
```

src/plugins/explore/public/application/legacy/discover/application/components/chart/chart.tsx
ananzh: Got it.

ananzh (dismissed):

Maosaic (dismissed):

ananzh (approved):

PR #10034, opened 2025-07-01
Update bar chart style panel

abbyhu2000:
### Description
Update bar chart style panel

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/bar/bar_vis_options.tsx
+++ b/src/plugins/explore/public/components/visualizations/bar/bar_vis_options.tsx
@@ -3,17 +3,16 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import React, { useState } from 'react';
-import { i18n } from '@osd/i18n';
-import { EuiSplitPanel, EuiButtonEmpty } from '@elastic/eui';
-import { GeneralVisOptions } from '../style_panel/general_vis_options';
+import React from 'react';
+import { EuiFlexGroup, EuiFlexItem } from '@elastic/eui';
 import { BarChartStyleControls } from './bar_vis_config';
-import { BarExclusiveVisOptions } from './bar_exclusive_vis_options';
-import { ThresholdOptions } from '../style_panel/threshold_options';
-import { GridOptionsPanel } from '../style_panel/grid_options';
-import { AxesOptions } from '../style_panel/axes_options';
 import { StyleControlsProps } from '../utils/use_visualization_types';
-import { ChartTypeSwitcher } from '../style_panel/chart_type_switcher';
+import { LegendOptionsPanel } from '../style_panel/legend/legend';
+import { ThresholdOptions } from '../style_panel/threshold/threshold';
+import { BarExclusiveVisOptions } from './bar_exclusive_vis_options';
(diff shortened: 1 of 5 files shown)


src/plugins/explore/public/components/visualizations/bar/bar_vis_options.tsx
```diff
@@ -56,159 +38,80 @@ export const BarVisStyleControls: React.FC<BarVisStyleControlsProps> = ({
...
+    <EuiFlexGroup direction="column" gutterSize="none">
+      <EuiFlexItem grow={false}>
+        <LegendOptionsPanel
+          shouldShowLegend={!notShowLegend}
+          legendOptions={{
+            show: styleOptions.addLegend,
+            position: styleOptions.legendPosition,
+          }}
+          onLegendOptionsChange={(legendOptions) => {
+            if (legendOptions.show !== undefined) {
```
kavilla, line 50: why not just check if value is undefined in update style option

```
const updateStyleOption = <K extends keyof BarChartStyleControls>(
    key: K,
    value: BarChartStyleControls[K]
  ) => {
     if (value !== undefined) {
      onStyleChange({ [key]: value });
    }
};
```

src/plugins/explore/public/components/visualizations/bar/to_expression.ts
```diff
@@ -139,7 +129,7 @@ export const createTimeBarChart = (
   // Configure bar mark
   const barMark: any = {
```
kavilla, line 130: avoid any

src/plugins/explore/public/components/visualizations/bar/bar_vis_options.tsx
```diff
@@ -56,159 +38,80 @@ export const BarVisStyleControls: React.FC<BarVisStyleControlsProps> = ({
...
+          onValueAxesChange={(valueAxes) => updateStyleOption('valueAxes', valueAxes)}
+          numericalColumns={numericalColumns}
+          categoricalColumns={categoricalColumns}
+          dateColumns={dateColumns}
+        />
+      </EuiFlexItem>
+      <EuiFlexItem grow={false}>
+        <GridOptionsPanel
+          grid={styleOptions.grid}
+          onGridChange={(grid) => updateStyleOption('grid', grid)}
```
kavilla, line 92: instead of having have these very unique props . i'm not sure who added them. but couldn't we just pass the the function to be called by the component.

like we just pass in `updateStyleOption` or really would be cool if it was just `onChange` that way we dont have to add a new prop for every configuration

src/plugins/explore/public/components/visualizations/style_panel/style_accordion.tsx
```diff
@@ -29,6 +29,7 @@ export const StyleAccordion: React.FC<StyleAccordionProps> = ({
         }
         initialIsOpen={initialIsOpen}
       >
+        <EuiSpacer size="s" />
```
kavilla, line 32: inline styling on line 26 and imo. i find it easier to use another class or something like a flex grid

abbyhu2000:
todo：clean `updateStyleOption` to just be onChange

PR #10045, opened 2025-07-01
[Explore] Support PPL filters in Explore

joshuali925:
### Description

- Support PPL filters in explore when adding from side bar or from results table
   - Filters are no longer added to `filterManager` in explore, because explore does not have UI to let user change filters. They will directly modify the PPL query in query editor
   - If user adds the same filter with the same value, it will not do anything
   - If user adds the negated version of an existing filter, it will update existing filter
   - Otherwise the filter is added as a where command after first pipe
- If user is currently focused on prompt editor (dual or single), then the same logic follows, but instead of adding `WHERE`, it adds in natural language. E.g. `show me docs` becomes `show me docs, foo is 'bar'`
- Refactor filters util functions

## Screenshot

https://github.com/user-attachments/assets/92099af4-fbcb-4532-aebb-e523afedb64d

## Changelog
- feat: Support PPL filters in Explore

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/data_table/explore_data_table.tsx
```diff
@@ -148,9 +156,17 @@ const ExploreDataTableComponent = () => {
         operation,
         indexPattern.id ?? ''
       );
-      return data.query.filterManager.addFilters(newFilters);
+      const languageConfig = services.data.query.queryString
+        .getLanguageService()
+        .getLanguage(query.language);
+      const newQuery = languageConfig?.insertFiltersToQuery?.(query, newFilters);
+      if (newQuery) {
+        dispatch(setQueryWithHistory(newQuery));
```
kavilla: is there such a thing as batch dispatches?

joshuali925: i'll refactor this, the desired UX will be set the query without running it. but good point, redux does have `batch()` which should be used if we go with existing code

src/plugins/explore/public/application/context/editor_context/editor_context.tsx
```diff
@@ -173,6 +177,7 @@ export const useEditorContext = (): EditorContextValue => {
       setBottomEditorText,
       query,
       prompt,
+      editorMode,
```
angle943: i'm conflicted about this change. I see why it could be convenient to have it here, as the consumers don't need to have a `const editorMode = useSelector(selectEditorMode)`, but `editorMode` is a redux state while the rest of the things here are the monaco editor stuff (it is true that they use the redux state to compute some things, but the output ultimately are things related to the monaco editors)

I'm afraid with this change that it can get confusing where `editorMode` is ultimately coming from. It seems more clear to me (albeit more wordy) for consumers to use `useEditorContext()` for things related to monaco editor, and `useSelector()` (aka Redux) to get redux state.


And you are aware of this, but just for context to everyone, we were "force" to have a separate mechanism to store the monaco editor state (like it's `ref` and the controlled text state) into this context instead of redux due to redux limitations.
So we are in a way forced into this not-so-ideal situation. But with this compromise in mind, I think being explicit in where we grab the data from could be more beneficial for DX in the long run.

Happy to discuss this more in detail if you disagree

src/plugins/explore/public/application/legacy/discover/application/view_components/panel/index.tsx
```diff
@@ -27,25 +27,28 @@ import {
...
   const columns = useSelector(selectColumns);
   const query = useSelector(selectQuery);
   const results = useSelector((state: any) => state.results);
-  const cacheKey = useMemo(() => {
-    const queryString = typeof query.query === 'string' ? query.query : '';
-    return defaultPrepareQuery(queryString);
-  }, [query]);
+  const cacheKey = useMemo(
+    () => defaultPrepareQuery(typeof query.query === 'string' ? query.query : ''),
+    [query]
```
angle943, line 38: random-comment: do you know why `query.query` is sometimes an array, and is that ever a practical outcome in the explore app?

src/plugins/explore/public/application/legacy/discover/application/view_components/panel/index.tsx
```diff
@@ -97,9 +100,15 @@ export function DiscoverPanel() {
         operation,
         indexPattern.id ?? ''
       );
-      return filterManager.addFilters(newFilters);
+      const languageConfig = queryString.getLanguageService().getLanguage(query.language);
+      const newText =
+        editorContext.editorMode === EditorMode.SingleQuery ||
+        editorContext.editorMode === EditorMode.DualQuery
+          ? languageConfig?.addFiltersToQuery?.(editorContext.query, newFilters)
+          : languageConfig?.addFiltersToPrompt?.(editorContext.prompt, newFilters);
```
angle943: i was hoping this component does not need to worry about `query` or `prompt`, and just call a function that magically do that, but languageConfig is outside of redux :(

angle943: not a requirement, but do you think it is worth having a hook like:

```
const useChangeQueryEditor = () => {
    const {services: {data: {queryString}}} = useOpenSearchDashboardServices();
const indexPatterns = useIndexPatterns();
   const editorMode = useSelector(selectEditorMode);
  
  const addFilter = useCallback((field: string | IndexPatternField, values: string, operation: '+' | '-') => {
      const newFilters = opensearchFilters.generateFilters(
        filterManager,
        field,
        values,
        operation,
        indexPattern.id ?? ''
      );
      const languageConfig = queryString.getLanguageService().getLanguage(query.language);
      const newText =
        editorContext.editorMode === EditorMode.SingleQuery ||
        editorContext.editorMode === EditorMode.DualQuery
          ? languageConfig?.addFiltersToQuery?.(editorContext.query, newFilters)
          : languageConfig?.addFiltersToPrompt?.(editorContext.prompt, newFilters);
  }, []);

  return { addFilter };
}
```


this way consumers can add filter without worrying about internal mechanisms, and it can be re used by other places too if use case arises.

In my above example i also called the hook `useChangeQueryEditor` so that it can be used for things like `removeFilter` or somethign in the future if we ever support that functionality


And btw, if you agree with this approach, lets put the hook within editor_context. I plan on creating serveral hooks for the editor context depending on the use case already (mostly due to some performance optimizations), something like this: 
Screenshot 2025-07-08 at 12 08 45 PM

So we can put this queryEditorManipulation hook there

src/plugins/explore/public/application/context/editor_context/editor_context.tsx
joshuali925: fair point, I was thinking `editorMode` is part of the "things related to monaco editor", but I also prefer to use redux if possible for consistency. will update

src/plugins/explore/public/application/legacy/discover/application/view_components/panel/index.tsx
joshuali925: i think they can also be DSL, but not really useful in explore
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/4a069f464bc056e0f7a925bb567100b27f06fe3f/src/plugins/data/common/opensearch_query/opensearch_query/from_lucene.test.ts#L64-L69

Although we do have a type check on if a `Query` is a PPL search query https://github.com/opensearch-project/OpenSearch-Dashboards/blob/08cfd5d3be62ee757799b85af3c9e6dc17afa966/src/plugins/query_enhancements/common/utils.ts#L145-L147

src/plugins/explore/public/application/legacy/discover/application/view_components/panel/index.tsx
joshuali925: good idea, will update. but I'd prefer to not put hooks in `context/` directory. i'll create a separate hooks directory

Maosaic (approved):

Issue #9990, opened 2025-06-26
[Meta][Banner Plugin] Feature, bug fix and enhancement for Banner Plugin


sidwang42:
## This meta issue tracks the PR/Issues for banner plugin: 
RFC #9861 

## Future Pull Requests Planed (in chronological order)
- Advanced settings for customizing the banner
- [Banner extensibility model Refactor](#10233)
- More Unit tests
- [Cypress Integration test](https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/1813)

## Feature
- [ ]  #9989
- [ ]  #10058 
- [ ]  #10264 
- [ ] #10139
  - [ ] #10270 
- [ ] #10326

## Enhancement
- [ ]  #10125 
- [ ] https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/1813
  - [ ] https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/1820 
- [ ] #10223
  - [ ] #10324 
- [ ] #10325 
- [ ] https://github.com/opensearch-project/oui/issues/1622

## Bug
- [ ] #10258
- [ ] #11503 

## Remaining Tasks

- [ ] #10380
- [ ] #10381

zhongnansu:
@sidwang42 Can you add all the remaining tasks as issues for tracking? For example, backport to older OSD, input sanitization, etc?

Issue #7111, opened 2024-06-26
[RFC] Dynamic Configuration Service


huyaboo:
## Proposal
Currently, all configs are statically determined via the config schema `defaultValues` property and the `opensearch_dashboards.yml`. This provides a centralized repository of config values but a major limitation is the inability to change configs without a server restart. For instance, if the admin needs to set the `server.maxPayloadBytes`, they would need to go into the config file, edit it, and restart Dashboards. Additionally, the existing config service does not allow for per-request config values. Say an admin needs to set different CSP rules based on the request. This is not feasible with the current model. The proposal instead is for a new core service `dynamicConfigService`, an extension of the existing `configService`, to expose a read-only dynamic config client in the `RequestHandlerContext` for plugins to consume and obtain configs from an external config store. For core services, the client will be exposed in the `dynamicConfigService.start()` method. 

## Background
Currently, the configs are handled by the `ConfigService` in [@osd-config](https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/packages/osd-config), which takes the `defaultValues` defined in a plugin/core service *config schema* and overrides them with the config values found in  `opensearch_dashboards.yml`. Take `csp` configs for example. Its config schema is the following:
```Typescript
export const config = {  
  path: 'csp',  
  schema: schema.object({  
    rules: schema.arrayOf(schema.string(), {  
      defaultValue: [  
        `script-src 'unsafe-eval' 'self'`,  
        `worker-src blob: 'self'`,  
        `style-src 'unsafe-inline' 'self'`,  
      ],  
    }),  
    strict: schema.boolean({ defaultValue: false }),  
    warnLegacyBrowsers: schema.boolean({ defaultValue: true }),  
  }),  
};
```

This means that by default, the `csp` config will be the following:
```Typescript
const DEFAULT_CSP = {
	rules: [`script-src 'unsafe-eval' 'self'`, `worker-src blob: 'self'`, `style-src 'unsafe-inline' 'self'`],
	strict: false,
	warnLegacyBrowsers: true,
};
```

If we define the following config in the `opensearch_dashboards.yml` file:
```yaml
csp.strict: true
csp.rules: ["default-src 'self'"]
```

The final config exposed to all core services and plugins will be 
```Typescript
const DEFAULT_CSP = {
	rules: [`default-src 'self'`],
	strict: true,
	warnLegacyBrowsers: true,
};
```

For plugins, the config schema can be found in `config.ts`.

[...]

PR #10058, opened 2025-07-02
[Banner Plugin] Add Api Support and Yaml config

sidwang42:
## Description

This PR enables dynamic configuration for the banner plugin by adding a server-side API endpoint and client-side integration. The banner's content, visibility, and style are now fetched from `/api/_plugins/_banner/content` on startup, allowing runtime control without editing frontend code or restarting the instance.

It replaces the previous hardcoded banner content with server-driven configuration, introduces a loading state, and ensures consistent layout behavior during banner transitions.

## Issues Resolved

* [x] Related to [[Meta Issue #9990](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9990)](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9990)
* [x] Addresses aspects of [[#9861](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9861)](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9861)

### Screenshot

image

### Testing the changes

1. Uncomment and set the following in `config/opensearch_dashboards.yml`:

   ```yaml
   banner.enabled: true
   banner.content: "This is a server-configured banner"
   ```
2. Start OpenSearch Dashboards
3. Visit the homepage; the banner should load with a spinner and then display the configured content
4. Optionally change the banner config server-side and refresh to verify dynamic update
5. Disable `banner.enabled` to ensure the banner does not render

### Changelog

- feat: Enable dynamic banner configuration via server API

### Check List

* [x] All tests pass

  * [x] `yarn test:jest`
  * [x] `yarn test:jest_integration`
* [x] New functionality includes testing
* [x] New functionality has been documented
* [ ] Update [[CHANGELOG.md](https://chatgpt.com/CHANGELOG.md)](./../CHANGELOG.md)
* [x] Commits are signed per the DCO using `--signoff`

---

Let me know if you'd like a lighter version or to move anything to a follow-up PR.


src/plugins/banner/public/plugin.ts
```diff
@@ -9,47 +9,45 @@
...
-import { CoreSetup, CoreStart, Plugin, PluginInitializerContext } from '../../../core/public';
+import { CoreSetup, CoreStart, Plugin } from '../../../core/public';
 import { BannerPluginSetup, BannerPluginStart } from './types';
 import { BannerService } from './services/banner_service';
+import { BannerApiService } from './services/banner_api_service';
 import { renderBanner, unmountBanner, setInitialBannerHeight } from './services/render_banner';
 
 export class BannerPlugin implements Plugin<BannerPluginSetup, BannerPluginStart> {
   private readonly bannerService = new BannerService();
+  private bannerApiService: BannerApiService | undefined;
```
huyaboo:
```suggestion
  private bannerApiService?: BannerApiService;
```

src/plugins/banner/public/services/banner_api_service.ts
```diff
@@ -0,0 +1,54 @@
...
+   * Fetches banner configuration from the API
+   */
+  public async fetchBannerConfig(): Promise<void> {
+    try {
+      const response = await this.http.get<BannerApiResponse>('/api/_plugins/_banner/content');
+
+      if (response.enabled) {
+        this.bannerService.updateBannerConfig({
+          text: response.content || '',
+          color: response.color || 'primary',
```
huyaboo: For these "default" configs, do we have a central place where they're defined?

src/plugins/banner/server/plugin.ts
```diff
@@ -25,16 +26,22 @@ export class BannerPlugin implements Plugin<BannerPluginSetup, BannerPluginStart
   public async setup(core: CoreSetup<BannerPluginStart>) {
     const pluginConfig: BannerPluginConfigType = await this.config$.pipe(first()).toPromise();
 
-    return {
+    const bannerSetup = {
       bannerEnabled: () => pluginConfig.enabled,
       getConfig: () => ({
-        text: pluginConfig.text,
+        text: pluginConfig.content,
```
huyaboo: Why is it `content` in the YAML config but it's `text` when accessed everywhere else? Could we not consolidate the naming to either or?

src/plugins/banner/server/routes/index.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { IRouter } from '../../../../core/server';
+import { BannerPluginSetup } from '../types';
+
+export function defineRoutes(router: IRouter, bannerSetup: BannerPluginSetup) {
+  router.get(
```
huyaboo: So we seem to be able to get the latest configs. Where do we set these configs? Is the source of truth still the server side YAML?

huyaboo (commented):
High level questions: 

1. If the configs are designed to be resolved at runtime, where can we configure them? 
2. Could we not use the `DynamicConfigService` to dynamically fetch configs at request time? 
3. Does the current approach require a server reload? If so, how is this differrent from just updating the yaml and refreshing the page?

sidwang42:
> @huyaboo 
> If the configs are designed to be resolved at runtime, where can we configure them?

Right now, changes to the config still requires a server restart and is still not runtime configure. Next PR will brings advanced settings and external source as banner content.

At the moment, the banner config is fetched from the server via an API on each page load. I'm also considering an alternative where the server polls an external source periodically (e.g., every 10 minutes) and stores the result in Advanced Settings, allowing the client to read a static value without triggering an API call every time.

Do you think one of these approaches is better? I’d really appreciate your thoughts or suggestions!

sidwang42:
>@huyaboo
> Could we not use the DynamicConfigService to dynamically fetch configs at request time?

I looked into DynamicConfigService (thanks for the RFC link: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/7111#issuecomment-2195756693), and it seems promising. However, the banner plugin is designed to support OpenSearch Dashboards versions from 1.3 to 2.19, and I believe DynamicConfigService isn’t available in many of those versions, which limits its applicability here.

>Does the current approach require a server reload? If so, how is this different from just updating the YAML and refreshing the page?

>So we seem to be able to get the latest configs. Where do we set these configs? Is the source of truth still the server side YAML?

At the moment, it’s functionally similar to updating the YAML and refreshing. However, the current design is intentionally structured to support future enhancements, like fetching from an external URL or using Advanced Settings for runtime customization without restarting the server.

src/plugins/banner/server/plugin.ts
sidwang42: Noted. Changed all `text` to `content` as banner may contains icons not just text.

src/plugins/banner/public/services/banner_api_service.ts
sidwang42: Yeah, I should avoid these defaults. Deleted all the other "default" configs and keeping only the visible one to avoid incomplete banner showing to the user.
```
isVisible: response.isVisible !== undefined ? response.isVisible : false,
```

src/plugins/banner/server/routes/index.ts
```diff
@@ -0,0 +1,46 @@
...
+    async (context, request, response) => {
+      const bannerEnabled = bannerSetup.bannerEnabled();
+
+      if (!bannerEnabled) {
+        return response.ok({
+          body: {
+            enabled: false,
+          },
+        });
+      }
```
zhongnansu: I don't think this is needed, but could you verify? If plugin is disabled, this API won't be registered , and won't be accessible

src/plugins/banner/public/services/banner_service.ts
zhongnansu, line 1: Shall we remove this service since now we are using API to retrieve the settings, and it doesn't make sense to me to persist the same set of configs through this BannerService in memory

src/plugins/banner/public/services/banner_api_service.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export class BannerApiService {
+  constructor(private readonly http: HttpSetup, private readonly bannerService: BannerService) {}
+
+  /**
+   * Fetches banner configuration from the API
+   */
+  public async fetchBannerConfig(): Promise<void> {
+    try {
+      const response = await this.http.get<BannerApiResponse>('/api/_plugins/_banner/content');
```
zhongnansu: can we simply this by moving the api call logic the ui component itSelf? API call is pretty common thing, it doesn't have too much logic for it to be created as a service.

src/plugins/banner/server/routes/index.ts
sidwang42: > @zhongnansu 
> I don't think this is needed, but could you verify? If plugin is disabled, this API won't be registered , and won't be accessible

Thanks for the comment. Tested. Ur right, this part deleted.

src/plugins/banner/public/services/banner_service.ts
sidwang42: > Shall we remove this service since now we are using API to retrieve the settings, and it doesn't make sense to me to persist the same set of configs through this BannerService in memory

Agreed.
Might benefit from this design in the future, but for the scope of this project it is redundant. Removed the services folder in the latest commit.

src/plugins/banner/public/services/banner_api_service.ts
sidwang42: > can we simply this by moving the api call logic the ui component itSelf? API call is pretty common thing, it doesn't have too much logic for it to be created as a service.

Same as above. Updated in the latest commit.

src/plugins/banner/public/plugin.ts
```diff
@@ -9,47 +9,29 @@
...
+    renderBanner(core.http);
 
-    return {
-      bannerService: this.bannerService,
-    };
+    return {};
   }
 
   public stop() {
+    // Unmount the banner component
```
huyaboo, line 31:
```suggestion

```

src/plugins/banner/public/types.ts
```diff
@@ -20,7 +18,7 @@ export interface BannerPluginSetup {}
 
 /**
  * Start contract for the banner plugin.
+ * Intentionally empty as the plugin no longer exposes any services.
```
huyaboo, line 21:
```suggestion

```

src/plugins/banner/common/constants.ts
```diff
@@ -15,11 +15,18 @@
...
+export const HIDDEN_BANNER_HEIGHT = '0px';
 
 /**
- * Height of the banner when hidden (in pixels)
+ * Default banner configuration values
  */
-export const HIDDEN_BANNER_HEIGHT = '0px';
+export const DEFAULT_BANNER_CONFIG = {
+  content: 'Banner Content',
+  color: 'primary' as const,
```
zhongnansu, line 27: why use `as const` here? What about using `enum`?

zhongnansu (approved):
left a minor comment, overall lgtm

src/plugins/banner/common/constants.ts
sidwang42: > @zhongnansu 
> why use `as const` here? What about using `enum`?

They behave similarly in this context, but `as const` is a lighter-weight alternative. Since the value isn’t reused widely across the codebase, as const fits the current design better.

zhongnansu (approved):

zhongnansu (dismissed):
Let's add unit tests for all components needed, to ensure quality @sidwang42

sidwang42:
> @zhongnansu 

Unit tests added:
here is the screenshot of the passed unit test:
image

src/plugins/banner/server/plugin.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      const initContext = coreMock.createPluginInitializerContext();
+      const plugin = new BannerPlugin(initContext);
+      const start = plugin.start(coreMock.createStart());
+
+      expect(start).toEqual({});
+    });
+  });
+
+  describe('stop', () => {
+    test('does not throw', () => {
```
zhongnansu: can you define the test case text to be more clear?

src/plugins/banner/server/routes/index.test.ts
zhongnansu, line 1: index tests are just weird name. Can we move API to its own file with readable name, `index.ts` should organize and import, not implement APIs.

src/plugins/banner/public/components/global_banner.test.tsx
```diff
@@ -0,0 +1,180 @@
...
+    ...originalReact,
+    lazy: jest.fn().mockImplementation(() => () => <div>Mocked Markdown Content</div>),
+    Suspense: ({ children, fallback }: any) => children,
+  };
+});
+
+// Mock react-markdown
+jest.mock('react-markdown', () => () => <div>Mocked Markdown Content</div>);
+
+describe('<GlobalBanner />', () => {
```
zhongnansu, line 37: do we have tests for `new home header` case?

sidwang42: Added and tests passed:
image

zhongnansu (approved):
lgtm, make sure CIs can pass before merging

config/opensearch_dashboards.yml
```diff
@@ -382,3 +382,6 @@
 
 # Set the value to true to enable the banner plugin
 # banner.enabled: true
+
+# Set the content of the banner
+# banner.content: "This is an important announcement for all users. [Learn more](https://opensearch.org) about OpenSearch."
```
xinruiba, line 387: Has the CSS issue been resolved?

src/plugins/banner/README.md
```diff
@@ -18,6 +18,14 @@ To enable or disable the banner plugin, add the following to your `opensearch_da
...
 # Enable or disable the banner plugin (default: false)
 banner.enabled: true
+
+# Configure the banner content and appearance
+banner.content: "Important announcement or notification"
+banner.color: "primary" # Options: primary, success, warning
+banner.iconType: "iInCircle" # Any valid EUI icon type
+banner.isVisible: true # Whether the banner is initially visible
+banner.useMarkdown: true # Whether to render content as markdown
+banner.size: "m" # Options: s (small), m (medium)
```
xinruiba, line 28: All these settings need to be configured under the .yml file? What's the default value?

src/plugins/banner/server/routes/get_config.ts
```diff
@@ -0,0 +1,36 @@
...
+import { BannerPluginSetup } from '../types';
+
+export function defineRoutes(router: IRouter, bannerSetup: BannerPluginSetup) {
+  router.get(
+    {
+      path: '/api/_plugins/_banner/content',
+      validate: false,
+    },
+    async (context, request, response) => {
+      const config = bannerSetup.getConfig();
```
xinruiba, line 22: Looks like this route will always return the pre-configured banner info from the .yml file, instead of making a real HTTP request to interact with other services.

config/opensearch_dashboards.yml
sidwang42: >@xinruiba
> Has the CSS issue been resolved?

Yes, it is been resolved in #10125.

src/plugins/banner/README.md
sidwang42: > All these settings need to be configured under the .yml file? What's the default value?

They don’t need to be explicitly configured. If omitted, the system will behave as if they were set to their default values.

src/plugins/banner/server/routes/get_config.ts
sidwang42: > Looks like this route will always return the pre-configured banner info from the .yml file, instead of making a real HTTP request to interact with other services.

Yeah, that’s how it currently works. The next PR will update this behavior, which will support fetching from a configurable external link.

This is tracked in #9990.

src/plugins/banner/public/render_banner.ts
```diff
@@ -11,34 +11,25 @@
...
+import { BANNER_CONTAINER_ID } from '../common';
+import { HttpStart } from '../../../core/public';
 
 /**
  * Renders the banner component into the DOM
- * @param bannerService The banner service instance
+ * @param http The HTTP client
  */
-export const renderBanner = (bannerService: BannerService): void => {
+export const renderBanner = (http: HttpStart): void => {
```
huyaboo, line 22: Do we have UT for this?

huyaboo (commented):
Left minor comments. Overall looks good

sidwang42:
About the two failing CI jobs and why they are not relevant to this PR:

* **First failure (Lighthouse audit)** image

  This failed due to a Lighthouse performance assertion:

  * **Metric**: `speed-index`
  * **Expected**: ≤ 26000
  * **Actual**: \~27196.76

  Since the banner plugin is disabled and this PR does not affect rendering or performance of the Discover app, it's unrelated to the change.
This CI failure is also occurring in recent unrelated PRs (e.g., #10256), confirming it’s not caused by changes in this PR.

* **Second failure (Functional tests on Windows - ciGroup5)**
  [Build and test / Run functional tests on Windows (ciGroup5)
  This failed after 71 minutes with repeated isolate timeouts:

  ```
  warn worker stderr Attempt:18 waiting for isolate vm-isolate to check in
  ...
  Attempt:20 waiting for isolate vm-isolate to check in
  ```

  This is a known flaky behavior in CI and not related to the changes in this PR.
Recent PRs, and former PRs all experienced similar issues.

src/plugins/banner/public/render_banner.ts
sidwang42: > Do we have UT for this?

The render_banner function is currently undergoing frequent changes. I plan to add unit tests for it in the next PR once the implementation stabilizes with the advanced settings imported.

PR #10060, opened 2025-07-03
[Explore] Execute the query when loading from saved object

FriedhelmWS:
## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts
+++ b/src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts
@@ -17,6 +17,7 @@ import {
   setQueryState,
 } from '../state_management/slices';
 import { Query } from '../../../../../data/common';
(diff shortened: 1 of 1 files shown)


src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts
```diff
@@ -68,6 +69,8 @@ export const useInitPage = () => {
           savedExplore.id,
           { type: 'explore' }
         );
+
+        dispatch(executeQueries({ services }));
```
raintygao, line 73: I was thinking this should be put in loadSavedExplore, if this is put here, will the execute time of this function be uncontrolled? It could execute at first time, but it could also execute automatically later when dependencies changed, right? Is this as expected? cc @ruanyl

ruanyl (approved):

PR #10066, opened 2025-07-04
[Explore] refactor query editor architecture

angle943:
- this is a refactor of the query editor.
- promopt mode has been tested. Still not perfect but progress

Notable changes:
- prompt editor and dual editor mode state works
- Pressing just "Enter" when theres no autosuggest fires the query correctly. This was actually a really annoying bug that effects our previous query editor too.
- introduces the `queryPanel` redux slice (actually just renamed the `systemSlice` into this) that houses "executionStatus" (previously `status` in SystemSlice), `editorMode` that is one of four permutations that the query editor can be in, and `promptModeIsAvailble`, which is a boolean of whether an agent is availble for prompt.
- Instead of having a state for whether each editor is enabled or not, or having various onChange and onRun and etc for different permutations, the centralization of the logic in redux makes it effectively have one of these handlers so that components don't need to worry about the state of the queryEditors (too much)
- the two query editor references and the text state is pulled to near the top of the plugin DOM, exposed via the `editorContext`. This was done so that the filter panel can have access to it and be able to update the query (as well as results table in the future). I actually labored a lot on how to solve this, initially we thought about putting the references to the editors in redux, but because its not serializable its not recommended, as well as it crashing redux dev tools due to how large those objects are :P . So i settled on a context. its not the most ideal, but still works
- adds unit tests and storybook for all query editor related files

Some things left to do:
- when switching dataset we must flip the editor back to single mode and change the query in the editor. also gotta check if agent is available and update `promptModeIsAvailable` accordingly.
- some uiState needs to go into the new queryPanel slice
- there is an annoying bug where if you are on the bottom editor with autocomplete open, and you flip to top, the suggestion lingers. This is due to react lifecycles
- I think performance of the editor can be improved by revising the editor hooks
- styles and classnames should be reworked
- when on dual editor state, clicking on disabled one should enable it


Maosaic:
Amazing work

Maosaic (dismissed):

src/plugins/explore/public/application/utils/state_management/actions/query_editor/on_editor_change/type_detection/type_detection.ts
```diff
@@ -58,17 +57,17 @@ export class QueryTypeDetector {
       plaintext: this._checkNaturalLanguage(normalizedQuery),
```
joshuali925: not fully sure but are static methods accessable using `this.` or should it be `QueryTypeDetector._xxx`

src/plugins/explore/public/application/utils/state_management/actions/query_editor/on_editor_run/call_agent/call_agent.ts
```diff
@@ -0,0 +1,109 @@
...
+    });
+    return;
+  }
+
+  try {
+    const params: QueryAssistParameters = {
+      question: prompt,
+      index: dataset.title,
+      // TODO: Make this not
+      language: 'PPL',
```
joshuali925, line 63: what is this todo?

src/plugins/explore/public/application/utils/state_management/slices/query/query_slice.ts
```diff
@@ -36,9 +36,19 @@ const querySlice = createSlice({
         meta: { addToHistory: true },
       }),
     },
+    setQueryStringWithHistory: {
+      reducer: (state, action: PayloadAction<string>) => {
+        // Same logic as setQueryState but with meta flag for history
+        state.query = action.payload;
```
joshuali925, line 42: is it ok to modify state directly or does redux use immer or something

src/plugins/explore/public/application/utils/state_management/utils/get_prompt_mode_is_available/get_prompt_mode_is_available.ts
```diff
@@ -0,0 +1,36 @@
...
+      return false;
+    }
+
+    // check if agent is available
+    const response: { configuredLanguages: string[] } = await services.http.get(
+      '/api/enhancements/assist/languages',
+      {
+        query: { dataSourceId: dataset.dataSource?.id },
+      }
+    );
```
joshuali925, line 28: can this use `isEnabled$` of the query assist extension? to avoid another network call

joshuali925 (dismissed):

src/plugins/explore/public/application/utils/state_management/actions/query_editor/on_editor_run/call_agent/call_agent.ts
angle943: o shoot nice catch, i meant to not hardcode PPL

src/plugins/explore/public/application/utils/state_management/slices/query/query_slice.ts
angle943: yeah redux toolkit uses immer under the hood: https://redux-toolkit.js.org/usage/immer-reducers

src/plugins/explore/public/application/utils/state_management/actions/query_editor/on_editor_change/type_detection/type_detection.ts
angle943: I think it works, but not proper. i'll update these `this`

src/plugins/explore/public/application/utils/state_management/utils/get_prompt_mode_is_available/get_prompt_mode_is_available.ts
```diff
@@ -16,19 +18,14 @@ export const getPromptModeIsAvailable = async (
...
     }
 
-    // check if agent is available
-    const response: { configuredLanguages: string[] } = await services.http.get(
-      '/api/enhancements/assist/languages',
-      {
-        query: { dataSourceId: dataset.dataSource?.id },
-      }
+    return await firstValueFrom(
+      queryAssistExtension.isEnabled$({} as QueryEditorExtensionDependencies)
```
joshuali925, line 27: not for this PR but we can probably get ride of the `dependencies` argument in the type definition later, since implementations shouldn't need it

joshuali925 (approved):

Maosaic (approved):

PR #10076, opened 2025-07-08
[Explore] only allow PPL to be filterable in explore and dashboards

joshuali925:
### Description

https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9953#discussion_r2166043524 changed PPL language `fields.filterable` to `undefined`. Rather than the language overriding `filterable` as true or false, `undefined` allows `filterable` to come from fields, which is the behavior we want for languages that support filters.

But setting it as `undefined` also enabled filtering in classic discover, where we don't have filter support for PPL.

This PR makes the PPL language filterable value to be dynamic. If the current app id is explore or dashboards, then it is `undefined` and will use the value from individual fields, otherwise PPL is not filterable.

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/public/plugin.tsx
+++ b/src/plugins/query_enhancements/public/plugin.tsx
@@ -3,10 +3,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import { i18n } from '@osd/i18n';
-import { BehaviorSubject } from 'rxjs';
+import { BehaviorSubject, Subscription } from 'rxjs';
 import moment from 'moment';
 import { CoreSetup, CoreStart, Plugin, PluginInitializerContext } from '../../../core/public';
-import { DataStorage, Filter, OSD_FIELD_TYPES } from '../../data/common';
+import { DataStorage, OSD_FIELD_TYPES } from '../../data/common';
 import {
   createEditor,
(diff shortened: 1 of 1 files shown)


angle943 (approved):

src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -42,6 +42,8 @@ export class QueryEnhancementsPlugin
   private isQuerySummaryCollapsed$ = new BehaviorSubject<boolean>(false);
   private resultSummaryEnabled$ = new BehaviorSubject<boolean>(false);
   private isSummaryAgentAvailable$ = new BehaviorSubject<boolean>(false);
+  private currentAppId$ = new BehaviorSubject<string | undefined>(undefined);
+  private appIdSubscription?: Subscription;
```
Maosaic, line 46: Why not following the same subscription pattern?

joshuali925: could you elaborate a bit more? which pattern?

Maosaic: resolved

PR #10081, opened 2025-07-09
[WIP] Add dataset select and service to explore

kavilla:
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


kavilla:
- [x]  label for select component should be `Data` temporarily
- [ ]  in the future, the temp dataset created in the saved search should be translated to using the virtual all dataset object. if i find if that not to be compat with specific data sources then we should bring it up
- [x] single button dont show manage and create
- [ ] re-add recent datasets
- [x] for details use the description component
- [x] selector text to be smaller
- [x] change timefield in details not be selector
- [ ] Manage for the dataset details iif index pattern
- [ ] datasetdetails data definition on click takes you to the appropriate view like data source or index pattern

kamingleung:
- [ ] In the three-dot menu, consider using [description list](https://oui.opensearch.org/1.19/#/display/description-list) component for displaying the meta information. Remove the borders
- [ ] Consider removing the icons and show a check on the selected item: see [single selectable](https://oui.opensearch.org/1.19/#/forms/selectable#single-selection)
- [ ] In the three-dot menu, If the timefield is read-only, show as key-value text. Avoid using disabled dropdown.

kamingleung:
- [ ] The dropdown width should be fixed, it shouldn't resize based on the text width. We should truncate the text if it's too long to fit the dropdown:
image

- [ ] Hide the "Default" badge if the dataset is not a default dataset:
image

- [ ] When user is viewing a default dataset, the badge should not be clickable or focusable, it's read-only:
image

PR #10083, opened 2025-07-09
[Explore] make source = * optional, and default to showing empty

angle943:
- make `source = *` optional. if you don't have it it will prepend it before it fires a query
- defualt to empty string when you load the page

- Some TODOs:
  - the PPLParser is not correct. for example, `search source=""` does not work
  - Keyvalue pair does not work correctly
  - There is a nasty bug (that i put a TODO on) where if you switch explore apps, the autocompletion does not work


src/plugins/data/public/query/query_string/query_string_manager.ts
```diff
@@ -336,7 +336,7 @@ export class QueryStringManager {
 
     return {
       ...newQuery,
-      query: this.getInitialDatasetQueryString(newQuery),
+      query: useInitialDatasetQueryString ? this.getInitialDatasetQueryString(newQuery) : '',
```
angle943: this is a change in `data` plugin. i'm wondering now that i have submitted this PR whether I shouldn't make this change, and just make the change in the explore plugin... I'll think more about it tomorrow

ananzh (approved):

Maosaic (approved):

PR #10104, opened 2025-07-10
[Explore] Integrate result summary component with redux states

FriedhelmWS:
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/results_summary/results_summary_panel.tsx
```diff
@@ -36,99 +28,60 @@ type AccordionState = 'closed' | 'open';
...
+    status: state.queryEditor.queryStatus.status || QueryExecutionStatus.UNINITIALIZED,
+    queryPromptState: state.queryEditor.lastExecutedPrompt,
+    dataSetState: state.query.dataset,
+    queryResults: state.results[state.query.query as string]?.hits?.hits,
+    editorMode: state.queryEditor.editorMode,
+  }));
+
+  const assistantEnabled = useMemo(() => !!core.application.capabilities?.assistant?.enabled, [
+    core.application.capabilities?.assistant?.enabled,
+  ]);
```
ruanyl: useMemo is useless here, we can simply do this:
```suggestion
  const assistantEnabled = !!core.application.capabilities?.assistant?.enabled;
```

src/plugins/explore/public/components/results_summary/results_summary_panel.tsx
```diff
@@ -36,99 +28,60 @@ type AccordionState = 'closed' | 'open';
...
+    queryState,
+    status,
+    queryPromptState,
+    dataSetState,
+    queryResults,
+    editorMode,
+  } = useSelector((state: RootState) => ({
+    queryState: state.query,
+    status: state.queryEditor.queryStatus.status || QueryExecutionStatus.UNINITIALIZED,
+    queryPromptState: state.queryEditor.lastExecutedPrompt,
```
ruanyl: Nit: maybe just call it `lastExecutedPrompt`? sounds more intuitive to me

src/plugins/explore/public/components/results_summary/results_summary_panel.tsx
```diff
@@ -138,97 +91,73 @@ export const ResultsSummaryPanel: React.FC<ResultsSummaryPanelProps> = (props) =
...
     },
-    [http, reportCountMetric, errorPrompt]
+    [http, reportCountMetric, dataSetState?.dataSource?.id]
   );
 
   useEffect(() => {
-    if (accordionState === 'open' && canGenerateSummary && !summary && !loading) {
-      // TODO properly integrate with t2ppl
+    if (
+      accordionState === 'open' &&
```
ruanyl: The `summary` here is the last result summary, does it mean the result summary will only be triggered when last result `summary` is empty? For example:
1. I run a natural language query
2. result summary is generated
3. Now I run another natural language query
4. Will it trigger new result summary for the new query?

ruanyl (commented):
Left a few comments :)

src/plugins/explore/public/components/results_summary/results_summary_panel.tsx
FriedhelmWS: The component actually re-renders (replaced by loading state) if the query is executed so the summary will not be present. The current implementation does trigger when we re-run the query: 

https://github.com/user-attachments/assets/42518cc6-74cc-4d7e-9eb2-2ff916be13ae

angle943:
I’ll review this tomorrow. Could you ensure all unit tests on explore continue passing with this change and there are no TS errors?

src/plugins/explore/public/application/utils/state_management/actions/set_dataset/set_dataset.ts
```diff
@@ -58,6 +60,14 @@ export const setDatasetActionCreator = (
     dispatch(setPromptModeIsAvailable(newPromptModeIsAvailable));
   }
 
+  const newSummaryAgentIsAvailable = await getSummaryAgentIsAvailable(
```
ruanyl: I suggest to do `Promise.allSettled([await getPromptModeIsAvailable(), await getSummaryAgentIsAvailable()])` so the two requests be fired in parallel

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
```diff
@@ -0,0 +1,164 @@
...
+    queryState: state.query,
+    lastExecutedPrompt: state.queryEditor.lastExecutedPrompt,
+    dataSetState: state.query.dataset,
+    queryResults: state.results[state.query.query as string]?.hits?.hits,
+    editorMode: state.queryEditor.editorMode,
+    queryStatus: state.queryEditor.overallQueryStatus.status,
+  }));
+
+  const assistantEnabled = core.application.capabilities?.assistant?.enabled;
+  const queryExecutedInPromptMode = useMemo(
```
ruanyl: Is this still needed? If yes, can you add a comment to describe the logic?

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
```diff
@@ -0,0 +1,164 @@
...
+  );
+
+  useEffect(() => {
+    // Close popover and reset summary when a query is re-executed
+    setIsPopoverOpen(false);
+    setSummary('');
+  }, [queryResults, setIsPopoverOpen, setSummary]);
+
+  const autoTriggered = useRef(false);
+  // Trigger auto generating for the first query
```
ruanyl: Does it open the popup automatically for the first time?

src/plugins/explore/public/application/utils/get_summary_agent_is_available/get_summary_agent_is_available.ts
```diff
@@ -0,0 +1,29 @@
...
+      query: {
+        dataSourceId,
+        agentConfigName: 'os_data2summary',
+      },
+    });
+    return res.exists;
+  } catch (error) {
+    return false;
+  }
+};
```
angle943, line 29: get we get unit test for this file?

src/plugins/explore/public/application/utils/state_management/actions/set_dataset/set_dataset.ts
angle943: +1 to @ruanyl , we don't want these two async calls to happen in sequence

src/plugins/explore/public/application/utils/state_management/actions/set_dataset/set_dataset.ts
```diff
@@ -58,6 +60,14 @@ export const setDatasetActionCreator = (
...
+    services,
+    query.dataset?.dataSource?.id!
```
angle943: what happens if dataset or dataSource is undefined?

src/plugins/explore/public/application/utils/state_management/actions/set_dataset/set_dataset.ts
```diff
@@ -58,6 +60,14 @@ export const setDatasetActionCreator = (
...
+  );
+  if (newSummaryAgentIsAvailable !== summaryAgentIsAvailable) {
+    dispatch(setSummaryAgentIsAvailable(newSummaryAgentIsAvailable));
+  }
+
```
angle943, line 76: i'm pretty sure the unit tests for this file would need to be updated for this change

src/plugins/explore/public/application/utils/state_management/selectors/query_editor/query_editor.ts
```diff
@@ -42,6 +42,10 @@ export const selectPromptToQueryIsLoading = createSelector(
   [selectState],
   (state) => state.promptToQueryIsLoading
 );
+export const selectSummaryAgentIsAvailable = createSelector(
+  [selectState],
+  (state) => state.summaryAgentIsAvailable
+);
```
angle943, line 49: lets add unit test for this, if you look at `query_editor.test.ts` you should see unit tests for all other selectors

src/plugins/explore/public/application/utils/state_management/slices/query_editor/query_editor_slice.ts
```diff
@@ -117,6 +122,7 @@ export const {
   setLastExecutedPrompt,
   setPromptModeIsAvailable,
   setPromptToQueryIsLoading,
+  setSummaryAgentIsAvailable,
```
angle943, line 140: this utility would need unit test in `query_editor_slice.test.ts`

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
```diff
@@ -266,6 +272,7 @@ const getPreloadedQueryEditorState = async (
     },
     promptModeIsAvailable,
     promptToQueryIsLoading: false,
+    summaryAgentIsAvailable: false,
```
angle943: why is this being set to false? That means the above `summaryAgentIsAvailable` is not being used. Shouldn't this just be `summaryAgentIsAvailable: summaryAgentIsAvailable,` or `summaryAgentIsAvailable,` for short?

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
```diff
@@ -246,8 +247,13 @@ const getPreloadedQueryEditorState = async (
   queryState?: QueryState
 ): Promise<QueryEditorSliceState> => {
   let promptModeIsAvailable = false;
+  let summaryAgentIsAvailable = false;
   if (queryState?.dataset) {
     promptModeIsAvailable = await getPromptModeIsAvailable(services);
+    summaryAgentIsAvailable = await getSummaryAgentIsAvailable(
+      services,
+      queryState.dataset.dataSource?.id!
```
angle943: 1. lets use promise.allSetttled for above two
2. what hapepns if dataSource? or id?

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
```diff
@@ -266,6 +272,7 @@ const getPreloadedQueryEditorState = async (
     },
     promptModeIsAvailable,
     promptToQueryIsLoading: false,
+    summaryAgentIsAvailable: false,
     editorMode,
     lastExecutedPrompt: '',
   };
```
angle943, line 274: there is an associated `redux_persistence.test.ts` file, lets update that with this change

src/plugins/explore/public/components/container/bottom_container/bottom_right_container/bottom_right_container.tsx
```diff
@@ -83,7 +82,6 @@ export const BottomRightContainer = () => {
   if (status === QueryExecutionStatus.READY) {
     return (
       <>
-        <ResultsSummaryPanel />
```
angle943, line 86: please update `bottom_right_container.test.tsx`. therte is a mock in there for this component

src/plugins/explore/public/components/query_panel/editor_stack/top_editor/edit_toolbar/edit_toolbar.tsx
```diff
@@ -64,6 +65,8 @@ export const EditToolbar = () => {
           <EuiText size="xs">{clearText}</EuiText>
         </div>
       </EuiButtonEmpty>
+      <div className="exploreEditToolbar__verticalSeparator" />
+      <ResultsSummary />
```
angle943: lets udate the `edit_toolbar.test.tsx`. Could you also ensure that the storybook is not busted with this change? Every component under the `query_panel` has an associated story that relies on `mock_provider.mocks.tsx`. That file might need to be updated. Try running `yarn storybook explore` and ensure its not busted

src/plugins/explore/public/components/query_panel/results_summary/results_summary.scss
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+@import "../shared";
+
+.exploreResultsSummary {
+  @include exploreQueryPanelButton;
+  @include exploreQueryPanelVerticalSeparator;
+}
+
+.exploreResultsSummary_popover_tooltip {
```
angle943: nitpick: why not `&_popover_tooltip {...}` and `__summaryContent {...}` inside `.exploreResultsSummary { ... }? This way if the name of class change its just one change.


And just fyi, the convention i've been seeing followed some parts of the code, as well as what i'm been following, is this: https://getbem.com/

src/plugins/explore/public/components/query_panel/results_summary/results_summary.scss
```diff
@@ -0,0 +1,21 @@
...
+  display: flex;
+  padding-inline-start: 2px;
```
angle943: nitpick (i think?): $ouiSizeXXS (i think that is 2px)

src/plugins/explore/public/components/query_panel/results_summary/results_summary.scss
```diff
@@ -0,0 +1,21 @@
...
+  transform: translate(-1px);
```
angle943: nitpick: i had to google what this translate(singleArg) does. looks like it translates horizontally. Maybe just `translateX(-1px)` to be explicit?

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
```diff
@@ -0,0 +1,164 @@
...
+    lastExecutedPrompt,
+    dataSetState,
+    queryResults,
+    editorMode,
+    queryStatus,
+  } = useSelector((state: RootState) => ({
+    queryState: state.query,
+    lastExecutedPrompt: state.queryEditor.lastExecutedPrompt,
+    dataSetState: state.query.dataset,
+    queryResults: state.results[state.query.query as string]?.hits?.hits,
```
angle943: @ananzh is this safe?

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
```diff
@@ -0,0 +1,164 @@
...
+    editorMode: state.queryEditor.editorMode,
+    queryStatus: state.queryEditor.overallQueryStatus.status,
+  }));
```
angle943: this is more nitpick, but I would prefer having multiple useSelectors using the memoized selectors so that state updates that does not belong to these will not cause a re-render

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  const assistantEnabled = core.application.capabilities?.assistant?.enabled;
+  const queryExecutedInPromptMode = useMemo(
+    // Only read the mode when the query result changed
+    () => [EditorMode.DualPrompt, EditorMode.SinglePrompt].includes(editorMode),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [queryResults]
```
angle943, line 58: why is editorMode excluded in this dependency?

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
angle943: i don't remember if we do this, but are we clearing out the `lastExecutedPrompt` in the queryEditorSlice if a non-prompt has been fired? If not should we add that logic? And that way, you can just check to see if `!lastExecutedPrompt`, instead of having this useMemo with `reqct-hooks/exhaustive-deps` lint ignore

angle943: the only value for the `lastExecutedPrompt` state is for this summary component, so lets use it and update it to fit our use case

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  return (
+    <ResultsSummaryButton
+      actionButtonVisible={actionButtonVisible}
+      feedback={feedback}
+      onFeedback={onFeedback}
+      summary={summary}
+      loading={loading}
+      onGenerateSummary={() =>
+        fetchSummary({
```
angle943: not sure if this matters, but `fetchSummary` was memoized above via `useCallback`, but its essentially creating a new function here every time, so this `ResultsSummaryButton` does not get a memoized version of this prop

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
```diff
@@ -0,0 +1,164 @@
...
+    [http, reportCountMetric, dataSetState?.dataSource?.id, summary, queryStatus]
+  );
+
+  useEffect(() => {
+    // Close popover and reset summary when a query is re-executed
+    setIsPopoverOpen(false);
+    setSummary('');
+  }, [queryResults, setIsPopoverOpen, setSummary]);
+
+  const autoTriggered = useRef(false);
```
angle943: nitpick: i think `isMounted` is better name

src/plugins/explore/public/application/utils/state_management/actions/set_dataset/set_dataset.ts
FriedhelmWS: That's the question I would like @kavilla to answer

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
FriedhelmWS: Nope, open the popover is still needed by the user

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
FriedhelmWS: Fixed

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
FriedhelmWS: Supply `undefined` to `getSummaryAgentIsAvailable ` will still make the request and check for agent on local cluster.

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
FriedhelmWS: If no result I think the results summary will just not getting triggered

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
ruanyl: Perhaps we no longer need this `autoTrigger`? because it now always requires user click

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
ruanyl: `queryExecutedInPromptMode` is only used to decide the visibility of panel action buttons `actionButtonVisible`, but the panel won't be visible if not prompt mode, so logically it's unnecessary, am I right? @FriedhelmWS

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
FriedhelmWS: Created a new useCallback to address this issue

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
FriedhelmWS: Removed this.

FriedhelmWS:
@angle943 Thanks for you comments, I've update the PR and all test would be passed now:
Screenshot 2025-07-18 at 13 23 34
Could you please review this commit https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10104/commits/dfb56076ff5e1b41587531783c410fa71716f7c0

src/plugins/explore/public/components/query_panel/results_summary/results_summary.tsx
FriedhelmWS: I think we still need this as we only want auto trigger the generation on the first query after initialization.

src/plugins/explore/public/application/utils/state_management/actions/set_dataset/set_dataset.ts
FriedhelmWS: If there is no datasource Id, the function will check for the local cluster and determine if there is a summary agent for it. I think that is the question I was asking @kavilla: if we don't have a dataSource object at this moment, does that always mean we are using a dataset from local cluster, so we intend to check an agent on the local cluster?

angle943 (approved):

PR #10106, opened 2025-07-10
[Explore] Add dynamic tab selection based on visualization compatibility

ananzh:
## Description

This PR tries multiple approaches to handle that explore automatically switches to the visualization tab when query results support visualization, otherwise stays on logs tab.

#### Approach 1 (first commit)
* useEffect to trigger detect based on result change
* Modified executeQueries() to always run visualization tab query for detection
* Added detection logic in ExploreTabsComponent when activeTabId === ''
* Changed initial activeTabId from 'logs' to '' to trigger auto-detection
* Run button and initial load now auto-detect optimal tab
* Respect saved activeTabId

#### Approach 2 (second commit and current approach)
* use component to trigger detect
* This requires await for results and a fix on overall status

This PR also allows save query to save the context in the editor not run query.

## Testing the changes

Before, dual render

https://github.com/user-attachments/assets/a1682621-1445-4ac8-ad9f-22bb9d3b8be5

After, no dual render --> auto switch
https://github.com/user-attachments/assets/0aab1b54-dd51-4e11-91eb-5e60beb3c24b

Saved query
https://github.com/user-attachments/assets/55f7b44e-0981-490d-87be-392584721808

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/tabs/tabs.tsx
```diff
@@ -43,9 +50,10 @@ export const ExploreTabsComponent = () => {
     };
   });
 
-  const activeTab = tabs.find((tab) => {
-    return tab.id === activeTabId;
-  });
+  const activeTab =
+    tabs.find((tab) => {
+      return tab.id === activeTabId;
+    }) || tabs.find((tab) => tab.id === 'logs'); // Fallback to logs if activeTabId is empty
```
angle943: should we memoize this?

src/plugins/explore/public/components/tabs/tabs.tsx
```diff
@@ -21,12 +23,17 @@ import { ExploreServices } from '../../types';
  */
 export const ExploreTabsComponent = () => {
   const dispatch = useDispatch();
-  // Get services from context
   const { services } = useOpenSearchDashboards<ExploreServices>();
   const registryTabs = services.tabRegistry.getAllTabs();
-
+  const results = useSelector((state: RootState) => state.results);
```
angle943, line 31: should we use a selector for results?

src/plugins/explore/public/components/tabs/tabs.tsx
```diff
@@ -21,12 +23,17 @@ import { ExploreServices } from '../../types';
...
   const activeTabId = useSelector(selectActiveTab);
 
+  useEffect(() => {
+    if (activeTabId === '' && results && Object.keys(results).length > 0) {
+      dispatch(detectAndSetOptimalTab({ services }));
+    }
+  }, [activeTabId, results, dispatch, services]);
```
angle943: is it possible to declaratively dispatch optimal tab within the executeQueries? I want to avoid `useEffect` unless absolutely necessary, and the logic here looks like it can be done without this sideeffect

src/plugins/explore/public/application/utils/state_management/slices/ui/ui_slice.ts
```diff
@@ -13,7 +13,7 @@ export interface UIState {
 }
 
 const initialState: UIState = {
-  activeTabId: 'logs',
+  activeTabId: '',
   showDatasetFields: true,
   prompt: '', // Initialize prompt as empty string
   showHistogram: true,
```
angle943, line 17: can we change the name of `setActiveTab` to `setActiveTabId` below? :(

src/plugins/explore/public/application/utils/state_management/slices/ui/ui_slice.ts
```diff
@@ -13,7 +13,7 @@ export interface UIState {
 }
 
 const initialState: UIState = {
-  activeTabId: 'logs',
+  activeTabId: '',
```
angle943, line 15: question #1: at the time of this creation (or in the `preloaded stuff` within redux persistence, do we have the list of registered tabs?

If so, does empty string makes sense? what does it mean when an active tab is empty?

ananzh: active tab is empty means we won't set a default tab but allow query result to determine. The current logic is to run queries for both default and visualization. It will set tab id to either to visualization if we can find a vis type to match or set to logs.

src/plugins/explore/public/application/utils/state_management/slices/ui/ui_slice.ts
ananzh: sure. can change.

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -140,28 +140,40 @@ export const executeQueries = createAsyncThunk<
...
+  const activeTabId = state.ui.activeTabId;
   const results = state.results;
 
   if (!services) {
     return;
   }
 
-  // Direct cache key computation (no computeQueryContext)
   const defaultCacheKey = defaultPrepareQuery(query);
+  const visualizationTab = services.tabRegistry?.getTab('explore_visualization_tab');
```
angle943: why is tabRegistry optional?

src/plugins/explore/public/components/tabs/tabs.tsx
joshuali925: i don't think this is that expensive

src/plugins/explore/public/components/tabs/tabs.tsx
joshuali925: +1

src/plugins/explore/public/application/utils/state_management/actions/detect_optimal_tab/detect_optimal_tab.ts
```diff
@@ -0,0 +1,72 @@
...
+};
+
+/**
+ * Determine the optimal tab based on results
+ * Use visualization tab if we can find a visualization type
+ * Otherwise, fallback to logs tab
+ */
+const determineOptimalTab = (results: any): string => {
+  if (canResultsBeVisualized(results)) {
+    return 'explore_visualization_tab';
```
joshuali925: is this always what user wants? for example, I might be in the logs tab because I want to compare numbers, and having to manually switch tab is not convenient

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -186,6 +210,8 @@ export const executeQueries = createAsyncThunk<
       )
     );
   }
+
+  // Wait for all queries to complete
   await Promise.all(promises);
```
joshuali925, line 189: should this be all settled?

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -175,8 +187,20 @@ export const executeQueries = createAsyncThunk<
     );
   }
 
-  // CONDITIONALLY execute tab query (only if queries are different and needed)
-  if (!queriesEqual && needsActiveTabQuery) {
+  // Execute visualization tab query for dynamic tab selection
+  if (needsVisualizationTabQuery) {
+    promises.push(
+      dispatch(
+        executeTabQuery({
```
joshuali925, line 168: would it be possible if the previous query is still running? is there a way to abort the previous one?

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
ananzh: fixed

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
ananzh: Yes, it is possible for a previous query to still be running when this code executes. However, there IS abort handling in place, but it's implemented at the individual query level rather than at the orchestration level. In executeQueryBase, each query creates its own AbortController. 

This is not very accurate. Compared to use_search in discover, since we have multiple query now, the current implementation has a gap where concurrent queries for the same cache key are not properly managed. I will send a follow up PR for this issue.

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
ananzh: It does not need. Let me take care of this with the below abort issue together.

src/plugins/explore/public/application/utils/state_management/actions/detect_optimal_tab/detect_optimal_tab.ts
ananzh: well this is the current decided behavior: when switch tab or load saved explore, will respect customer's choice. When user run a new query, initialize explore or switch dataset, will auto detect. I asked the same question but it seems we will follow this pattern and ask user to feedback on this behavior.

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
ananzh: fixed

src/plugins/explore/public/application/utils/state_management/actions/query_editor/run_query/run_query.ts
```diff
@@ -4,20 +4,26 @@
...
     dispatch(setQueryStringWithHistory(query));
   }
 
+  dispatch(setActiveTab(''));
   dispatch(clearResults());
-  dispatch(executeQueries({ services }));
+  dispatch(clearQueryStatusMap());
+
+  await dispatch(executeQueries({ services }));
+  dispatch(detectAndSetOptimalTab({ services }));
```
Maosaic, line 28: Do we always want to do auto switch tab whenever query run? could be annoying

angle943 (dismissed):

Maosaic (approved):

src/plugins/explore/public/application/utils/state_management/actions/query_editor/run_query/run_query.ts
ananzh: That is the current expecting behavior. We can update this if user see major performance issue.

PR #10107, opened 2025-07-10
Fix duplicate Completion Providers

LDrago27:
### Description

Both Top editor and Bottom Editor was using the use_shared_editor. This was causing multiple completion providers to be registered for the same language. This was causing same suggestion to appear multiple times in the suggestion wizard. 

This change uses a language based registry to store the completionProvider, and replace it to store only one instance of completionProvider per language.

## Testing the changes

After Change 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/query_panel/utils/use_shared_editor/use_shared_editor.ts
+++ b/src/plugins/explore/public/components/query_panel/utils/use_shared_editor/use_shared_editor.ts
@@ -33,6 +33,9 @@ type IStandaloneCodeEditor = monaco.editor.IStandaloneCodeEditor;
 
 const TRIGGER_CHARACTERS = [' '];
 
+// Registry to prevent duplicate completion provider registrations
+const completionProviderRegistry = new Map<string, monaco.IDisposable>();
+
 const languageConfiguration: LanguageConfiguration = {
   autoClosingPairs: [
     { open: '(', close: ')' },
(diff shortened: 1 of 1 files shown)


src/plugins/explore/public/components/query_panel/utils/use_shared_editor/use_shared_editor.ts
```diff
@@ -171,12 +174,25 @@ export const useSharedEditor = ({
...
       suggestionProvider
     );
-    return () => disposable.dispose();
+
+    completionProviderRegistry.set(queryLanguage, disposable);
+
+    return () => {
+      disposable.dispose();
+      completionProviderRegistry.delete(queryLanguage);
+    };
```
angle943, line 195: lets add tests for this, but even before that let me sync up with you first. i thought i fixed this issue

kavilla (commented):
do you think there is merit for version of language?

src/plugins/explore/public/components/query_panel/utils/use_shared_editor/use_shared_editor.ts
```diff
@@ -33,6 +33,9 @@ type IStandaloneCodeEditor = monaco.editor.IStandaloneCodeEditor;
 
 const TRIGGER_CHARACTERS = [' '];
 
+// Registry to prevent duplicate completion provider registrations
+const completionProviderRegistry = new Map<string, monaco.IDisposable>();
```
angle943, line 37: this scares me. Its essentially a singleton and i never am confident that this will work correctly as we want. Should we throw this into the queryEditor context? Happy to discuss

LDrago27:
Closing it in favor of https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10108

PR #10110, opened 2025-07-10
Filter out Null Values in Suggestions

LDrago27:
### Description

Filter out Null Values in Suggestions

## Changelog

- fix: Filter out Null Values in Suggestions

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/antlr/shared/utils.ts
+++ b/src/plugins/data/public/antlr/shared/utils.ts
@@ -149,18 +149,17 @@ export const formatValuesToSuggestions = <T extends { toString(): string }>(
   values: T[], // generic for any value type
   modifyInsertText?: (input: T) => string
 ) => {
-  let i = 0;
-
-  const valueSuggestions: MonacoCompatibleQuerySuggestion[] = values.map((val: T) => {
-    i++;
-    return {
-      text: val.toString(),
-      type: monaco.languages.CompletionItemKind.Value,
-      detail: SuggestionItemDetailsTags.Value,
-      sortText: i.toString().padStart(values.length.toString().length + 1, '0'), // keeps the order of sorted values
-      ...(modifyInsertText && { insertText: modifyInsertText(val) }),
-    };
-  });
+  const valueSuggestions: MonacoCompatibleQuerySuggestion[] = values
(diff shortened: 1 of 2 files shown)


src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -151,16 +151,18 @@ export const formatValuesToSuggestions = <T extends { toString(): string }>(
...
-      text: val.toString(),
-      type: monaco.languages.CompletionItemKind.Value,
-      detail: SuggestionItemDetailsTags.Value,
-      sortText: i.toString().padStart(values.length.toString().length + 1, '0'), // keeps the order of sorted values
-      ...(modifyInsertText && { insertText: modifyInsertText(val) }),
-    };
-  });
+  const valueSuggestions: MonacoCompatibleQuerySuggestion[] = values
+    .filter((val) => !!val) // Only using the notNull values
+    .map((val: T) => {
```
angle943: do `(val: T, i) => {`

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -149,18 +149,17 @@ export const formatValuesToSuggestions = <T extends { toString(): string }>(
...
-    return {
-      text: val.toString(),
-      type: monaco.languages.CompletionItemKind.Value,
-      detail: SuggestionItemDetailsTags.Value,
-      sortText: i.toString().padStart(values.length.toString().length + 1, '0'), // keeps the order of sorted values
-      ...(modifyInsertText && { insertText: modifyInsertText(val) }),
-    };
-  });
+  const valueSuggestions: MonacoCompatibleQuerySuggestion[] = values
+    .filter((val) => !!val) // Only using the notNull values
```
joshuali925: if there's a boolean suggestion, this also filters out `false`?

angle943: what about `val ?? true`? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

LDrago27: Thanks. I have updated the condition to be more specific

PR #10114, opened 2025-07-11
save to query string manager with source

angle943:
- when we save no source query to slice, we don't save it with source
- when we save it to queryStringManager, we save it with source

### DIFF
--- a/src/plugins/explore/public/application/utils/languages/types.ts
+++ b/src/plugins/explore/public/application/utils/languages/types.ts
@@ -1,0 +1,10 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Query } from '../../../../../data/common';
+
+export type QueryWithQueryAsString = Omit<Query, 'query'> & {
+  query: string;
+};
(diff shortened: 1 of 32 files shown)


src/plugins/explore/public/application/utils/languages/types.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Query } from '../../../../../data/common';
+
+export type QueryWithQueryAsString = Omit<Query, 'query'> & {
+  query: string;
+};
```
joshuali925, line 10: maybe also update https://github.com/opensearch-project/OpenSearch-Dashboards/blob/b5882e98bea143d921c789372ce6b38b0b65aeae/src/plugins/query_enhancements/common/utils.ts#L147-L147
to use this type

joshuali925 (approved):

src/plugins/explore/public/application/utils/languages/types.ts
angle943: i think this type should be defined next to Query (i think its defined in data plugin?) , but i didn't want to touch any code outside explore plugin. we can do that later

Maosaic (approved):

PR #10118, opened 2025-07-11
Add Support for PPL functions

LDrago27:
### Description

This PR aims at adding support for SPAN, MATCH, MATCH_PREFIX , MATCH_PHRASE and other relevance functions in Autocomplete. 

## Screenshot

https://github.com/user-attachments/assets/10ff943c-10f6-4a91-8475-5d353eecff41

## Changelog

- feat: Expand Support for PPL functions

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/query_panel/utils/use_shared_editor/use_shared_editor.ts
```diff
@@ -43,6 +43,7 @@ const languageConfiguration: LanguageConfiguration = {
     { open: '"', close: '"' },
     { open: "'", close: "'" },
   ],
+  wordPattern: /@?\w[\w@'.-]*[?!,;:"]*/,
```
angle943: can you leave a comment on this regex above? I have no idea what it is looking for. are these the keys that we allow?

LDrago27: Oh sure. Actually it is a weird issue with monaco editor suggestion replacement in general. 
So what happens is, if you have a long suggestions with `.` or `@`, and you have typed in a part of it when you select a suggestion, instead of replacing the previous typed in part,  it appends to it. 

For eg:
Imagine you have typed in `log.` and there is a suggestion `log.service.envname` when you select the suggestion you get `log.log.service.envname` instead of `log.service.envname`. 

The issue is what monaco editor considers as a word and replaces it with the suggestion. By default Monaco editor doesn't consider `.` and `@` to be part of the word. Therefore it doesn't consider `log.` as word and just appends the suggestion to it. The above regex basically makes it possible to have a word that contains `.` and `@` in it. So that now when a user

angle943 (approved):

sejli (approved):

PR #10125, opened 2025-07-11
[Banner Plugin] UI Implementation Refactor

sidwang42:
### Description

Adds a lightweight **header wrapper** and shared `--global-banner-height` / `$osdHeaderOffset` variables so all headers, body content, flyouts and masks automatically shift downward whenever the Global Banner is mounted. Removes all `!important` overrides and hard-coded pixel offsets.

### Issues Resolved

The concerns raised in discussion of #9989.

## Screenshot

design drawio
design 2 drawio

### Implementation Details

This change avoids hardcoded CSS overrides by dynamically offsetting headers and page content using a shared layout strategy:

#### **Banner Mounting**

* The banner is rendered as a fixed element: `#pluginGlobalBanner`
* On mount, it sets a CSS variable:

```css
--global-banner-height
```

---

#### **Header Wrapping**

* A new class `.headerGlobalNav--withBanner` is conditionally applied when the banner is enabled.
* This class modifies header positioning:

```scss
.headerGlobalNav--withBanner .expandedHeader {
  top: var(--global-banner-height);
}

.headerGlobalNav--withBanner .primaryHeader:not(.newTopNavHeader) {
  top: calc(var(--global-banner-height) + $headerWidth);
}
```

---

#### **Shared Offsets**

* Body content, flyouts, and overlay masks use shared SCSS variables:

```scss
$osdHeaderOffset: calc(#{$euiHeaderHeightCompensation} + var(--global-banner-height, 0px));
$osdHeaderOffsetExpanded: calc((#{$euiHeaderHeightCompensation} * 2) + var(--global-banner-height, 0px));
```

* These values are injected into `min-height`, `padding`, and `top` styles to ensure content shifts automatically when the banner appears.

---

#### **Fallback Safety**

* If the banner is not rendered, `--global-banner-height` defaults to `0px`, preserving the existing layout when the banner is disabled.

## Testing the changes

1. **Start Dashboards** with banner disabled – verify no layout change.
2. Enable banner plugin (`opensearch_dashboards.yml`).
3. Refresh any app page:

   * Banner appears at top.
   * Expanded Header & Primary Header reposition correctly.
   * Flyouts / overlay masks open without overlap.
4. Resize window, and switch to dark mode – confirm offsets remain correct.

## Changelog

- feat: add wrapper & CSS-variable offsets to support Global Banner without `!important` hacks

### Check List

* [ ] All tests pass

  * [ ] `yarn test:jest`
  * [ ] `yarn test:jest_integration`
* [ ] New functionality includes testing.
* [ ] New functionality has been documented.
* [ ] Update [[CHANGELOG.md](https://chatgpt.com/CHANGELOG.md)](./../CHANGELOG.md)
* [ ] Commits are signed per the DCO using `--signoff`

### DIFF
--- a/src/core/public/chrome/ui/header/header.tsx
+++ b/src/core/public/chrome/ui/header/header.tsx
@@ -213,7 +213,16 @@ export function Header({
 
   const toggleCollapsibleNavRef = createRef<HTMLButtonElement & { euiAnimate: () => void }>();
   const navId = htmlIdGenerator()();
-  const className = classnames('hide-for-sharing', 'headerGlobalNav');
+
+  // Get the banner plugin configuration
+  const bannerPluginConfig = injectedMetadata
+    ?.getPlugins()
+    ?.find((plugin: { id: string }) => plugin.id === 'banner')?.config;
+  const isBannerEnabled = bannerPluginConfig?.enabled === true;
+
+  const className = classnames('hide-for-sharing', 'headerGlobalNav', {
+    'headerGlobalNav--withBanner': isBannerEnabled,
+  });
   const { useExpandedHeader = true } = branding;
   const useApplicationHeader = headerVariant === HeaderVariant.APPLICATION;
 
@@ -657,12 +666,6 @@ export function Header({
   const renderHeader = () => {
     return useApplicationHeader ? renderApplicationHeader() : renderPageHeader();
   };
-
-  // Get the banner plugin configuration
-  const bannerPluginConfig = injectedMetadata
-    ?.getPlugins()
-    ?.find((plugin: { id: string }) => plugin.id === 'banner')?.config;
-  const isBannerEnabled = bannerPluginConfig?.enabled === true;
 
   return (
     <>
--- a/src/plugins/banner/public/index.scss
+++ b/src/plugins/banner/public/index.scss
@@ -3,6 +3,12 @@
    ================================================================= */
 
 /**
+ * Header spacing variables
+ */
+$headerWidth: $ouiSizeL * 2 + 1px;
+$expandedHeaderWidth: $headerWidth * 2;
+
+/**
  * 1. CSS Variable - Defines the space needed for the global banner
  *    Default value is 0px when no banner is present
  *    Will be dynamically set to the actual banner height when visible
@@ -19,47 +25,22 @@
   top: 0;
   left: 0;
   right: 0;
-  z-index: 1100; /* Higher than header z-index (1000) */
+  z-index: 1100;
   width: 100%;
   min-height: var(--global-banner-height);
 }
 
 /**
- * 3. Smooth transitions - Prevents jarring layout shifts when banner height changes
- */
-/* stylelint-disable @osd/stylelint/no_modifying_global_selectors */
-.euiHeader {
-  transition: top 0.3s ease;
-}
-
-/**
- * 4. Header positioning - Places headers below the banner
+ * 3. Header positioning - Places headers below the banner
  *    expandedHeader: directly below the banner
  *    primaryHeader: below the expanded header
  */
-.expandedHeader {
-  top: var(--global-banner-height) !important;
+.headerGlobalNav--withBanner .expandedHeader {
+  top: var(--global-banner-height);
 }
 
-.primaryHeader:not(.newTopNavHeader) {
-  top: calc(var(--global-banner-height) + $euiSizeXL + $euiSizeM) !important;
+.headerGlobalNav--withBanner .primaryHeader:not(.newTopNavHeader) {
+  top: calc(var(--global-banner-height) + $headerWidth);
 }
-
-/**
- * 5. Fix for body padding when header is expanded and fixed
- *    Ensures content doesn't overlap with headers
- */
-.headerIsExpanded.euiBody--headerIsFixed {
-  padding-top: calc(var(--global-banner-height) + $euiSizeXL + $euiSizeM + $euiSizeXL + $euiSizeM) !important;
-}
-
-/**
- * 6. Fix for flayout position when header is expanded and fixed
- *    Ensures content doesn't overlap with headers
- */
-.euiOverlayMask {
-  top: calc(var(--global-banner-height) + $euiSizeXL + $euiSizeM + $euiSizeXL + $euiSizeM) !important;
-}
-/* stylelint-enable @osd/stylelint/no_modifying_global_selectors */
 
 /* End of Global Banner Positioning */
--- a/src/plugins/home/public/application/components/_home.scss
+++ b/src/plugins/home/public/application/components/_home.scss
@@ -28,14 +28,16 @@
  * under the License.
  */
 
+@import "src/core/public/variables";
+
 .homWrapper {
   background-color: $euiColorEmptyShade;
   display: flex;
   flex-direction: column;
-  min-height: calc(100vh - #{$euiHeaderHeightCompensation});
+  min-height: calc(100vh - #{$osdHeaderOffset});
 
   .headerIsExpanded & {
-    min-height: calc(100vh - #{$euiHeaderHeightCompensation * 2});
+    min-height: calc(100vh - #{$osdHeaderOffsetExpanded});
   }
 }
 
--- a/src/core/public/_variables.scss
+++ b/src/core/public/_variables.scss
@@ -1,3 +1,7 @@
 @import "@elastic/eui/src/global_styling/variables/header";
 
-$osdHeaderOffset: $euiHeaderHeightCompensation;
+// Base header offset including the banner height
+$osdHeaderOffset: calc(#{$euiHeaderHeightCompensation} + var(--global-banner-height, 0px));
+
+// Double header offset for expanded state
+$osdHeaderOffsetExpanded: calc((#{$euiHeaderHeightCompensation} * 2) + var(--global-banner-height, 0px));
--- a/src/core/public/chrome/ui/header/_index.scss
+++ b/src/core/public/chrome/ui/header/_index.scss
@@ -3,7 +3,7 @@
 }
 
 .headerIsExpanded {
-  @include euiHeaderAffordForFixed($osdHeaderOffset * 2);
+  @include euiHeaderAffordForFixed($osdHeaderOffsetExpanded);
 }
 
 .headerGlobalNav .euiHeaderSectionItem:empty {
--- a/src/core/public/rendering/_base.scss
+++ b/src/core/public/rendering/_base.scss
@@ -15,7 +15,7 @@
   min-height: calc(100vh - #{$osdHeaderOffset});
 
   .headerIsExpanded & {
-    min-height: calc(100vh - #{$osdHeaderOffset * 2});
+    min-height: calc(100vh - #{$osdHeaderOffsetExpanded});
   }
 
   &.hidden-chrome {


src/plugins/home/public/application/components/_home.scss
```diff
@@ -28,14 +28,16 @@
  * under the License.
  */
 
+@import "src/core/public/variables";
```
virajsanghvi: Was this already imported in another file? It will likely be imported twice - may want to validate (probably not problematic for this particular file, but not ideal pattern to set)

src/core/public/chrome/ui/header/header.tsx
```diff
@@ -213,7 +213,16 @@ export function Header({
 
   const toggleCollapsibleNavRef = createRef<HTMLButtonElement & { euiAnimate: () => void }>();
   const navId = htmlIdGenerator()();
-  const className = classnames('hide-for-sharing', 'headerGlobalNav');
+
+  // Get the banner plugin configuration
+  const bannerPluginConfig = injectedMetadata
+    ?.getPlugins()
+    ?.find((plugin: { id: string }) => plugin.id === 'banner')?.config;
+  const isBannerEnabled = bannerPluginConfig?.enabled === true;
```
virajsanghvi, line 222: Isn't it a little weird for core to know about a particular plugin? Is idea that if someone else wanted to implement a banner plugin they'd have to give it the same id?

Looks like this is pre-existing code, so not a blocker

src/plugins/banner/public/index.scss
```diff
@@ -19,47 +25,22 @@
   top: 0;
   left: 0;
   right: 0;
-  z-index: 1100; /* Higher than header z-index (1000) */
+  z-index: 1100;
```
virajsanghvi, line 28: why remove comment?

virajsanghvi (commented):
some comments

src/plugins/banner/public/index.scss
```diff
@@ -19,47 +25,22 @@
...
-/**
- * 6. Fix for flayout position when header is expanded and fixed
- *    Ensures content doesn't overlap with headers
- */
-.euiOverlayMask {
-  top: calc(var(--global-banner-height) + $euiSizeXL + $euiSizeM + $euiSizeXL + $euiSizeM) !important;
+.headerGlobalNav--withBanner .primaryHeader:not(.newTopNavHeader) {
+  top: calc(var(--global-banner-height) + $headerWidth);
 }
-/* stylelint-enable @osd/stylelint/no_modifying_global_selectors */
```
virajsanghvi, line 63: Why is this no longer necessary?

src/plugins/banner/public/index.scss
```diff
@@ -19,47 +25,22 @@
...
-.primaryHeader:not(.newTopNavHeader) {
-  top: calc(var(--global-banner-height) + $euiSizeXL + $euiSizeM) !important;
-}
-
-/**
- * 5. Fix for body padding when header is expanded and fixed
- *    Ensures content doesn't overlap with headers
- */
-.headerIsExpanded.euiBody--headerIsFixed {
-  padding-top: calc(var(--global-banner-height) + $euiSizeXL + $euiSizeM + $euiSizeXL + $euiSizeM) !important;
```
virajsanghvi, line 53: why is this no longer necessary?

sidwang42:
> @virajsa
> Can you post a screenshot with these changes and the new page header?

The default home screen:
banner_showcase

The new page Header:
image

The current banner overlaps with the new header — I plan to address this in a follow-up PR. Supporting a non-overlapping banner with the new header would require adding conditional logic in the SCSS file, which feels out of scope for this PR’s purpose.

src/plugins/home/public/application/components/_home.scss
sidwang42: > Was this already imported in another file? It will likely be imported twice - may want to validate (probably not problematic for this particular file, but not ideal pattern to set)

It wasn’t imported elsewhere, but you're right — it probably makes more sense to import it at a higher level like src/plugins/home/public/application/index.scss. I’ll update that, thanks!

src/plugins/banner/public/index.scss
sidwang42: > @virajsanghvi 
> why remove comment?

It’s no longer necessary because body content, flyouts, and overlay masks now use shared SCSS variables like $osdHeaderOffset and $osdHeaderOffsetExpanded. These are injected into min-height, padding, and top styles, so layout shifts automatically when the banner appears.

These changes are all related to concerns raised by the already merged PR #9989.

src/plugins/banner/public/index.scss
sidwang42: >It’s no longer necessary because body content, flyouts, and overlay masks now use shared SCSS variables like $osdHeaderOffset and $osdHeaderOffsetExpanded. These are injected into min-height, padding, and top styles, so layout shifts automatically when the banner appears.

>These changes are all related to concerns raised by the already merged PR #9989.

Same as above.

src/plugins/banner/public/index.scss
sidwang42: > @virajsanghvi 
> Why is this no longer necessary?

It’s no longer necessary because body content, flyouts, and overlay masks now use shared SCSS variables like $osdHeaderOffset and $osdHeaderOffsetExpanded. These are injected into min-height, padding, and top styles, so layout shifts automatically when the banner appears.

These changes are all related to concerns raised by the already merged PR #9989.

src/core/public/chrome/ui/header/header.tsx
sidwang42: > Isn't it a little weird for core to know about a particular plugin? Is idea that if someone else wanted to implement a banner plugin they'd have to give it the same id?
> 
> Looks like this is pre-existing code, so not a blocker

You're right — it is a bit awkward for core to reference a specific plugin. In this case, it’s needed to ensure the plugin mount is only added to the DOM when enabled, and to conditionally apply the wrapper class.

Definitely open to better approaches if you have something in mind — would love to hear your thoughts!

zhongnansu (approved):
thanks for removing the css override, re-enabling linter

src/plugins/banner/public/index.scss
sidwang42: > why remove comment?

This comment was brought in from my previous PR. After reviewing similar cases in the codebase, I realized that z-index values typically aren’t commented this way — so I removed it for consistency.

src/core/public/chrome/ui/header/header.tsx
virajsanghvi: Do you have a point of view of why this doesn't use the banner overlay service? e.g. curious what answer to  https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9989/files#r2193589565 is (@kavilla)

src/plugins/banner/public/index.scss
virajsanghvi: Ah, didn't realize this was all new code

virajsanghvi:
> I plan to address this in a follow-up PR

Why are we following up vs just fixing this now?

src/core/public/chrome/ui/header/header.tsx
virajsanghvi: Read the RFC (https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9861) - given our reasoning this is standalone and not intended to be reused, not sure I fully understand why this is a plugin if its has a specific purpose that seems coupled to core.

Anyways, @zhongnansu  is this a common practice for plugins within core to be referenced by id in core code?

zhongnansu: > Anyways, @zhongnansu is this a common practice for plugins within core to be referenced by id in core code?

@virajsanghvi  There are some options to reference plugin from core, the current implementation is a lightweight solution that's acceptable to me base on the use case. Tho it might not be the best practice.

One option is like implementation in this PR. Although I didn't find similar usage, but `injectedMetaService -> getPlugin()` API does relay the plugin manifest from server side to client side. A use case like this is acceptable to me, which it just reads the feature flag and conditionally render UI.

A more common way is defining a interface in core and expose to plugin to register/set. 

Which approach do you think we should take in this case?

sidwang42:
> > I plan to address this in a follow-up PR
> 
> Why are we following up vs just fixing this now?

Thanks for the question!

I spent a good amount of time investigating possible solutions here. Unlike the default home page, the new home page and its flyout are implemented as fixed-position components. There’s currently no equivalent to `$osdHeaderOffset` that can be set to adjust their position when the banner is visible.

The primary goal of this project is to promote new OpenSearch Dashboards features (like the new home page) to existing users. Supporting banner offsets in the new home page is out of scope for this PR.

The only viable approach I found involves overriding top-level layout styles like:

```scss
.content,
.euiFlyout,
.newPageTopNavExpander {
  top: var(--global-banner-height);
}
```

However, this would override existing styles and risk regressions.

To avoid that, this PR conditionally hides the banner when the new home page is enabled. A proper solution is tracked in [the issue](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/10139), which will include a refactor of the new UI layout to introduce a --new-ui-header-offset (or equivalent) and apply it systematically across all affected components.

virajsanghvi:
> The primary goal of this project is to promote new OpenSearch Dashboards features (like the new home page) to existing users. Supporting banner offsets in the new home page is out of scope for this PR.

If the direction of the new experience is with the new header, it is odd to me that we don't support it.

src/core/public/chrome/ui/header/header.tsx
```diff
@@ -659,15 +668,9 @@ export function Header({
...
-  // Get the banner plugin configuration
-  const bannerPluginConfig = injectedMetadata
-    ?.getPlugins()
-    ?.find((plugin: { id: string }) => plugin.id === 'banner')?.config;
-  const isBannerEnabled = bannerPluginConfig?.enabled === true;
-
   return (
     <>
-      {isBannerEnabled && <div id="pluginGlobalBanner" />}
+      {isBannerEnabled && !useUpdatedHeader && <div id="pluginGlobalBanner" />}
```
virajsanghvi, line 673: Why not always just have a space to mount global banners vs making it specific to this plugin?

src/core/public/chrome/ui/header/header.tsx
virajsanghvi: I think the latter - I don't think core should really reference plugins by id (and from your comment, it seems like we don't currently do this), unless the design intent is that users can provide a plugin with this id for this behavior (although the banner plugin is a core plugin, so that seems like its not the case).

It is also odd that we're introducing a core plugin that doesn't support the direction the ux is going (new header) - is this a stop-gap solution? Do we plan to eventually support the new header?

zhongnansu (approved):
Let's make sure you are tracking all the follow-up properly in separate github issues and and linked here.

sidwang42:
> One call out, can you validate that failing CI isn't an issue?

- Build and test / Build min release artifacts on Windows x64 is failing because of:

- worker stderr [BABEL] Note: The code generator has deoptimised the styling of D:\a\OpenSearch-Dashboards\OpenSearch-Dashboards\artifacts\src\plugins\data\public\antlr\opensearch_sql\.generated\OpenSearchSQLParser.ts as it exceeds the max of 500KB
---
- Build and test / Run functional tests on Windows (ciGroup4) (pull_request) is failing after 7m is failing because of:

- build_opensearch_dashboards_platform_plugins.js

None of these two could be caused by this PR.
Thanks, for pointing out.

sidwang42:
> * We plan on picking up new header support and this pr prevents this from showing there

This will be tracked by this issue:
#10139

> * We can address proper extensibility model after this pr (as it fixes something already in the code base) - let's please just have a quick discussion with stakeholders.

This will be tracked by this issue:
#10223

ruanyl:
@sidwang42 seems this breaks plugin build: https://github.com/opensearch-project/dashboards-assistant/actions/runs/16407245910/job/46355121266?pr=589

sidwang42:
> @sidwang42 seems this breaks plugin build: https://github.com/opensearch-project/dashboards-assistant/actions/runs/16407245910/job/46355121266?pr=589

Thanks for pointing it out! 
I looked into it — the plugin build was failing because the SCSS import in index.scss used a path like @import "src/core/public/variables", which assumes the Sass loader is configured to resolve from the project root. However, for plugins, this kind of alias isn't recognized. I’ve fixed it by switching to a relative import path and submitted a PR here: #10258.

PR #10128, opened 2025-07-12
[Explore] Query editor style update

angle943:
- remove most of stylelint ignores minus two, one is for AI mode animation, the other is for overlay when you are on dual editor mode and one of them is disabled (this ones kind of stupid rule i think. i am using SASS but the fact that i am using rgba($sassVariable, 0.2) it doesn't like it)
- There is now no directly targeting of monaco editor styles using CSS
- Clean up CSS to make it more scalable and easier to update
- align text for both prompt and editor mode


sejli (dismissed):

Maosaic:
I think the style rule exception should be added to `packages/osd-stylelint-config/config/restricted_values.json`

angle943:
Oops I think I forgot about placeholder. Don't merge this

ananzh (approved):

sejli (approved):

PR #10133, opened 2025-07-14
[Explore] add switch x and y

Qxisylolo:
### Description
apply switch X and Y for all charts type(expect pie metric)

## Screenshot

https://github.com/user-attachments/assets/2451ca54-5d01-4e94-9a2b-36c7eb7499a4

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/line/to_expression.ts
+++ b/src/plugins/explore/public/components/visualizations/line/to_expression.ts
@@ -326,7 +326,7 @@ export const createFacetedMultiLineChart = (
     },
     facet: {
       field: category2Field,
-      type: 'nominal',
+      type: getSchemaFromAxisMapping(facetMapping),
       columns: 2,
       header: { title: category2Name },
     },
@@ -339,34 +339,17 @@ export const createFacetedMultiLineChart = (
           encoding: {
             x: {
               field: dateField,
-              type: 'temporal',
-              axis: applyAxisStyling(
-                {
-                  title: dateName,
-                  labelAngle: -45,
-                },
-                styles,
-                'category',
-                numericalColumns,
-                categoricalColumns,
-                dateColumns
-              ),
+              type: getSchemaFromAxisMapping(xAxisMapping),
+              axis: applyAxisStyling(xAxisMapping, styles?.grid?.xLines),
             },
             y: {
               field: metricField,
-              type: 'quantitative',
-              axis: applyAxisStyling(
-                { title: metricName },
-                styles,
-                'value',
-                numericalColumns,
-                categoricalColumns,
-                dateColumns
-              ),
+              type: getSchemaFromAxisMapping(yAxisMapping),
+              axis: applyAxisStyling(yAxisMapping, styles?.grid?.yLines),
             },
             color: {
               field: category1Field,
-              type: 'nominal',
+              type: getSchemaFromAxisMapping(colorMapping),
               legend:
                 styles?.addLegend !== false
                   ? {


src/plugins/explore/public/components/visualizations/bar/bar_vis_options.tsx
```diff
@@ -65,6 +82,13 @@ export const BarVisStyleControls: React.FC<BarVisStyleControlsProps> = ({
           categoricalColumns={categoricalColumns}
           dateColumns={dateColumns}
         />
+      </EuiFlexItem> */}
```
FriedhelmWS, line 85: Just remove it

Qxisylolo: Please ignore some of the comments for now — they are fallback notes. I’ll remove them once this PR is finalized.

src/plugins/explore/public/components/visualizations/area/area_vis_options.tsx
```diff
@@ -35,6 +36,23 @@ export const AreaVisStyleControls: React.FC<AreaVisStyleControlsProps> = ({
     onStyleChange({ [key]: value });
   };
 
+  const handleAxisStyleChange = (
```
FriedhelmWS, line 39: Can we just refactor it as the utils function so we don't have to repeat exact same code for five times.

src/plugins/explore/public/components/visualizations/bar/bar_vis_config.ts
```diff
@@ -130,6 +130,11 @@ export const createBarConfig = (): VisualizationType<'bar'> => ({
             [AxisRole.X]: { type: VisFieldType.Categorical, index: 0 },
             [AxisRole.Y]: { type: VisFieldType.Numerical, index: 0 },
           },
+          // TODO the first one should be default?
```
FriedhelmWS, line 133: Could you please remove this comment 😆

src/plugins/explore/public/components/visualizations/line/line_vis_config.ts
```diff
@@ -122,11 +122,16 @@ export const createLineConfig = (): VisualizationType<'line'> => ({
...
+          {
+            [AxisRole.X]: { type: VisFieldType.Numerical, index: 0 },
+            [AxisRole.Y]: { type: VisFieldType.Date, index: 0 },
+          },
         ],
       },
       {
         mapping: [
           {
+            // TODO :temporarily disable vertical line-bar chart
```
FriedhelmWS, line 134: Have we? 🤔

Qxisylolo: I'm not sure if we should apply the line-bar switch, as it would introduce a new axisRole: x_second, and handling this switch might become tricky. [In such way, user can also select x_2 in field selector].

It’s doable, but it requires changes in more places.

For now, the conversion between line, bar, and area only works when there's a single X and Y axis. 
Current implementation already ensures that axis settings are preserved when switching both the chart type and the X/Y axes. Since the line-bar chart is a special case of line charts, I think it's ok to temporarily disable or hide this button

src/plugins/explore/public/components/visualizations/area/area_vis_options.tsx
Qxisylolo: Will do it in following commit. thanks for pointing out

src/plugins/explore/public/components/visualizations/line/line_vis_config.ts
FriedhelmWS: 👆Make sense 👍, maybe just remove this TODO and say we don't support vertical line-bar chart 😗

FriedhelmWS:
@ruanyl btw do you have any concern on putting style settings in the axes field mapping object and persist it?

src/plugins/explore/public/components/visualizations/style_panel/axes/standard_axes_options.tsx
```diff
@@ -13,35 +13,32 @@ import {
...
+  Positions,
+  AxisRole,
+  AxisStyleStoredInMapping,
+  CompleteAxisWithStyle,
+  AxisLabels,
+} from '../../types';
 import { DebouncedTruncateField, DebouncedText } from '.././utils';
 import { StyleAccordion } from '../../style_panel/style_accordion';
 
 interface AllAxesOptionsProps {
```
FriedhelmWS, line 29: Can we rename the file to `all_axse_options.tsx` to avoid confusion? 😶

ruanyl:
> @ruanyl btw do you have any concern on putting style settings in the axes field mapping object and persist it?

@FriedhelmWS Technically no, I think this is more like a product call. 

Any suggestions? @ashwin-pc @virajsanghvi

PR #10117, opened 2025-07-11
Add dataset select and service to explore

kavilla:
### Description

https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10014

The dataset select component alongside the data view service.

**NOTE**: This is the initial foundational work. Will clean up unused and make it more unique to datasets but people can review this PR under the assumption this is similar to index patterns

The original PR: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9980 had conflicts with namespacing that typescript was getting thrown off. It had conflicts with the existing dataset and the dataset interface in the query does not nicely confirm to index pattern. The next logical namespace for me was **data view** which is a view of the data.

So **data view** acts as a direct fork of index pattern to avoid backwards compatibility issues for existing index patterns and to not break the query enhancements logic.

This also includes the change to run story book on a multiple ports. Since the storybook for this plugin is not the explore plugin I wanted the ability to run the storybook for each plugin not just one at a time.

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9791

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #10134, opened 2025-07-14
[explore] dataset select bug fixes and clean up

kavilla:
### Description

Cleans up some of the logic to extend index pattern and simplify some of the logic 

### Issues Resolved

n/a

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #10138, opened 2025-07-14
fix datasets to not call fetchDatasets on every re-render

joshuali925:
### Description

Follow up fix for #10117. Currently every re-render triggers a call to `GET /api/saved_objects/_find?fields=title&per_page=10000&type=index-pattern&workspaces=xxx`. It makes debugging other issues difficult

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/top_nav/top_nav.tsx
+++ b/src/plugins/explore/public/components/top_nav/top_nav.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { i18n } from '@osd/i18n';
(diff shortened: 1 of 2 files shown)


src/plugins/explore/public/components/top_nav/top_nav.tsx
```diff
@@ -150,30 +150,33 @@ export const TopNav = ({ setHeaderActionMenu = () => {}, savedExplore }: TopNavP
 
   const dispatch = useDispatch();
 
-  const handleDatasetSelect = (newDataset: any) => {
```
joshuali925, line 153: @kavilla why is this `any`? I tried `Dataset` but `newDataset.dataSource` will be invalid

angle943:
@kavilla could you review this?

kavilla:
handled some of the callback here: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10134 thanks for bringing up. must of have been tired when i wrote it

PR #10147, opened 2025-07-15
[Explore] fix type errors on legacy folder

angle943:
- only ones remaining are `abort_data_query_action.test.ts` and few errors related to index patterns

### DIFF
--- a/src/plugins/explore/public/components/chart/discover_chart_container.tsx
+++ b/src/plugins/explore/public/components/chart/discover_chart_container.tsx
@@ -17,6 +17,7 @@ import {
 import { RootState } from '../../application/utils/state_management/store';
 import { selectShowHistogram } from '../../application/utils/state_management/selectors';
 import { CanvasPanel } from '../../application/legacy/discover/application/components/panel/canvas_panel';
+import { Chart } from './utils';
 
 export const DiscoverChartContainer = () => {
   const { services } = useOpenSearchDashboards<ExploreServices>();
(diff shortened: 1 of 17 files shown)


src/plugins/explore/public/components/chart/discover_chart_container.tsx
```diff
@@ -66,7 +67,7 @@ export const DiscoverChartContainer = () => {
       <div className="dscCanvas__chart">
         <DiscoverChart
           bucketInterval={processedResults.bucketInterval}
-          chartData={processedResults.chartData}
+          chartData={processedResults.chartData as Chart}
```
ruanyl, line 70: Shall we update `Chart` type instead of type casting?

ruanyl (approved):
Approved, but I see lots of type casting, are we going to get rid of the "legacy" folder in the future?

src/plugins/explore/public/components/chart/discover_chart_container.tsx
angle943: Let me respond to your other comment here (not sure why but when a comment is not in a specific line i can't respond to it on github)

So the North star plan is to indeed get rid of `legacy` folder. Legacy folder is for the most part a "copy pasted" version of discover to this new plugin. However, realistically speaking, I am not sure when that will happen. What I am envisioning is that some parts of the legacy folder will need to be reworked for future use-cases, and then we should move it out of the legacy and touch it up.

However, I spent minimal time on fixing the type errors on the legacy folder, since they are copy&pasted from discover and the feature already "worked" in production (i'm using the word "worked" very loosely here :D )

Since the code was already shipped to production (aka in discover), I don't think it is worth our time to fix the typescript errors, unless a known bug exists in it. And for the `Chart` type, I don't have too much context nor did i look into it too deeply, but am hesitant about changing the type as that might have domino effect. But if you got context, feel free to update (but i don't think its high priority due to the above reasons)

But for new code we write, I def would not want to be this lax about it. only for the legacy stuff

PR #10189, opened 2025-07-16
[Explore] Log patterns container

paulstn:
### Description

Implements the patterns tab container logic. The tab will specify a specific patterns query to be sent out, and using those results as well as the results from the logs tab, displays information for the patterns table. Additionally, using the brain algorithm's pattern, it will highlight all of the dynamic elements in a sample log of the pattern grouping.

## Screenshot

image

## Changelog

- feat: container logic for the patterns tab

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -43,8 +43,16 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
     order: 15,
     supportedLanguages: [EXPLORE_DEFAULT_LANGUAGE],
 
-    prepareQuery: (query) => {
-      return typeof query.query === 'string' ? query.query : '';
+    prepareQuery: (queryString) => {
+      const patternsField = 'message'; // TODO: pull from patterns field configured in dataset
```
paulstn: This is still an active TODO, not handling this in this PR.

Maosaic:
can you add description and/or screenshot about this PR?

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -44,7 +44,15 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
     supportedLanguages: [EXPLORE_DEFAULT_LANGUAGE],
 
     prepareQuery: (query) => {
-      return typeof query.query === 'string' ? query.query : '';
+      const patternsField = 'message'; // TODO: pull from patterns field configured in dataset
+
+      // TODO: add quotes around take patternsField
+      // ` | patterns ${patternsField} method=brain | stats count() as count, take(${patternsField}, 1) as sample by patterns_field | sort - count | fields patterns_field, count, sample`
+
+      return typeof query.query === 'string' && query.query !== ''
```
angle943: empty string is an active query now, since "source = *" is optional. Do we still want this empty string check?

paulstn: I think i'll change this, I was seeing a bug that I noticed might've just been fixed recently so I'm going to rebase and modify this.

src/plugins/explore/public/components/patterns_table/utils/constants.tsx
```diff
@@ -23,14 +26,18 @@ export const patternsTableColumns: Array<EuiBasicTableColumn<PatternItem>> = [
...
-      defaultMessage: 'Pattern',
+    field: 'sample',
+    name: i18n.translate('explore.patterns.table.column.sampleLog', {
+      defaultMessage: 'Pattern Sample Log',
     }),
-    render: (pattern: string) => pattern || '—',
+    render: (sample: string) => {
+      const sanitizedSampleLog = dompurify.sanitize(sample);
+      // eslint-disable-next-line react/no-danger
+      return <span dangerouslySetInnerHTML={{ __html: sanitizedSampleLog || '—' }} />;
```
joshuali925: why set inner html?

src/plugins/explore/public/components/patterns_table/utils/utils.tsx
```diff
@@ -7,3 +7,98 @@
 export const isValidFiniteNumber = (val: number) => {
   return !isNaN(val) && isFinite(val);
 };
+
+// TODO: clean up the below. move constants, write a clearer description
+export const highlightLogUsingPattern = (log: string, pattern: string) => {
```
joshuali925, line 46: why not regex?

src/plugins/explore/public/components/patterns_table/utils/constants.tsx
paulstn: The sample log, using the helper method below called `highlightLogUsingPattern`, will include < mark> tags in the real log that we pull, for the purpose of highlighting all of the dynamic vs static elements. This is similar to the existing functionality where DQL search terms will highlight terms in the results, and so I'm utilizing the same method done there. The dangerouslySetInnerHtml is needed to have the < mark> tags be recognized as HTML.

src/plugins/explore/public/components/patterns_table/utils/utils.tsx
paulstn: The goal of this method is to highlight a sample log. So, the sample log would be something like 
`[Log] Gecko GET/something 172.198.1.1`
and the pattern string we get could be something like
`[Log] <*> GET/<*> <*IP*>`.
We could highlight all the dynamic elements in the pattern string using regex, but can't expand beyond that to also highlight the sample log. To do that, this piece of code essentially checks the static elements between dynamic elements and matches that with the sample log to find and highlight the dynamic elements.

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -10,11 +10,17 @@ import { ExploreServices } from '../types';
...
+import { RootState } from './utils/state_management/store';
+import { findDefaultPatternsField } from '../components/patterns_table/utils/utils';
 
 /**
  * Registers built-in tabs with the tab registry
  */
-export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
+export const registerBuiltInTabs = (
+  tabRegistry: TabRegistryService,
+  services?: ExploreServices
```
kavilla: couldn't services be accessed via the context?

also if it cant i would just make it not optional

src/plugins/explore/public/components/patterns_table/utils/constants.tsx
kavilla: nit: it is a little bit weird that constants file is not a file of constants. should we consider placing this somewhere else

src/plugins/explore/public/components/tabs/action_bar/results_action_bar/results_action_bar.tsx
```diff
@@ -37,6 +38,7 @@ export const DiscoverResultsActionBar = ({
 }: DiscoverResultsActionBarProps) => {
   const currentTab = useSelector(selectActiveTabId);
   const shouldShowAddToDashboardButton = currentTab !== 'explore_patterns_tab';
+  const showTabSpecificSettings = currentTab === 'explore_patterns_tab';
```
kavilla, line 41: this const is too generic. and the previous const seems so as well. it seems more like. wouldn't it be easier if we just checked in the component the specific tab. that way we wont need to find what the constant equals

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -44,18 +50,32 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
     order: 15,
     supportedLanguages: [EXPLORE_DEFAULT_LANGUAGE],
 
-    prepareQuery: (query) => {
-      return typeof query.query === 'string' ? query.query : '';
+    prepareQuery: (query, state) => {
```
kavilla: is this the only place we are using the prepareQuery? if not how come we don't just prepare the query from the patterns? why does a higher component need to handle this?

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_content.tsx
```diff
@@ -0,0 +1,121 @@
...
+      if (fieldChange) fieldChange(); // run logic from whoever calls the content here
+    });
+  };
+
+  // Show loading indicator if dataset is loading
+  if (isLoading) {
+    return <EuiLoadingSpinner />;
+  }
+
+  // Show error message if there was an error loading the dataset
```
kavilla: nit: unneeded comments

paulstn: i'll remove them

src/plugins/explore/public/application/register_tabs.ts
paulstn: this here is defining prepareQuery from the tab registry's definition, there is other code that calls upon pattern tab's prepareQuery for things such as figuring out what the cache key would be, sending out the query, etc.

src/plugins/explore/public/application/register_tabs.ts
paulstn: which context are you referring to? do you have another code snippet where services is accessed through this?

src/plugins/explore/public/components/tabs/action_bar/results_action_bar/results_action_bar.tsx
paulstn: So the reason I had this generic const was because this boolean is meant to handle if tab specific settings are shown (this would be the gear icon). For now, the patterns tab is the only one using it, so I have it set as such. I was emulating what the prev line did. 
I would assume that it could change in the future, since addToDashboardButton and tabSpecificSettings don't have to be mutually exclusive. 
So I kept the name generic, to refer to what the boolean would do if true inside of the component.

src/plugins/explore/public/application/register_tabs.ts
angle943: the services can be accessed via `useOpenSearchDashboardContext` or something, but this is a regular function, so it does not have access to hooks. And +1 on making services non-optional. Lets change it to be required

src/plugins/explore/public/application/register_tabs.ts
angle943: prepareQuery is used by all tabs that needs a custom prepare query

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -44,18 +50,32 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
...
+      // Get the selected patterns field from the Redux state
+      let patternsField = state?.tab.patterns.patterns_field;
+
+      const preparedQuery = getQueryWithSource(query);
+      if (!patternsField) {
+        patternsField = findDefaultPatternsField(state, services);
+      }
+
+      return typeof preparedQuery.query === 'string' && preparedQuery.query !== ''
```
angle943: Screenshot 2025-07-28 at 8 28 46 PM

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -44,18 +50,32 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
...
+            ` | patterns \`${patternsField}\` method=brain mode=aggregation | sort - pattern_count`
+        : preparedQuery.query;
     },
 
     component: PatternsTab,
 
     // Add lifecycle hooks
-    onActive: () => {
-      // Tab activated
+    onActive: (state?: RootState): void => {
```
angle943: can we make state required here?

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -44,18 +50,32 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
...
+      findDefaultPatternsField(state, services);
     },
-    onInactive: () => {
+
+    onInactive: (state?: RootState): void => {
```
angle943: same as above

src/plugins/explore/public/application/register_tabs.ts
angle943: but wait, doesn't `services` have the store already? can we just use that?

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -44,18 +50,32 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
...
       // Tab deactivated
+      if (!state || !services?.store) return;
```
angle943: if state and services are required, i don't think this return would be needed

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -44,18 +50,32 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
...
+      services.store.dispatch(setPatternsField(''));
```
angle943: are we sure we want to do this? If you fired a query, and user is quickly going back and forth on different tabs, shouldn't the `patternsField` just remain the same?

src/plugins/explore/public/application/register_tabs.ts
angle943: does this need `state`? Wouldn't we be able to get it from `services`?

src/plugins/explore/public/application/utils/hooks/use_tab_results.ts
```diff
@@ -15,15 +15,21 @@ import { defaultPrepareQueryString } from '../state_management/actions/query_act
...
+    results,
+  } = fullState;
 
   const cacheKey = useMemo(() => {
     const activeTab = services.tabRegistry.getTab(activeTabId);
     const prepareQuery = activeTab?.prepareQuery || defaultPrepareQueryString;
-    return prepareQuery(query);
-  }, [query, activeTabId, services]);
+
+    return prepareQuery(query, fullState);
```
angle943: IF prepareQuery does not need fullState, then this file's changes is no longer needed i think

src/plugins/explore/public/application/utils/state_management/actions/set_dataset/set_dataset.ts
```diff
@@ -29,16 +29,24 @@ export const setDatasetActionCreator = (
...
 
   dispatch(setActiveTab(''));
   dispatch(clearResults());
   dispatch(clearQueryStatusMap());
   dispatch(clearLastExecutedData());
 
+  const activeTab = services.tabRegistry.getTab(ui.activeTabId);
+  if (activeTab?.onInactive) {
+    activeTab?.onInactive(state);
+  }
```
angle943: why don't we just dispatch `setPatternsField('')`?

i'm not a fan of this approach of using activeTab.onActive because:

- is the tab really getting inactive? Say you are on logs tab and you switch dataset, does the tab become inactive?
- Say you are on the patterns tab and user switches dataset. Does the tab go back to logs? If so, wouldn't `onInactive` automatically be called?

src/plugins/explore/public/components/patterns_table/patterns_container.tsx
```diff
@@ -4,26 +4,50 @@
...
 export const PatternsContainer = () => {
-  const { results } = useTabResults();
+  const { results: patternResults } = useTabResults();
 
-  // TODO: Register custom processor for patterns tab if needed
-  //       If no need, feel free to remove this comment
-  // const tabDefinition = services.tabRegistry?.getTab?.('patterns');
-  // const processor = tabDefinition?.resultsProcessor || defaultResultsProcessor;
-  // const processedResults = processor(rawResults, indexPattern);
+  const querySelector = useSelector((state: RootState) => state.query);
```
angle943: i think there is a memoized selector of this: `useSelector(selectQuery)`

src/plugins/explore/public/components/patterns_table/patterns_container.tsx
```diff
@@ -4,26 +4,50 @@
...
+  const resultsSelector = useSelector((state: RootState) => state.results);
```
angle943: memoized version

src/plugins/explore/public/components/patterns_table/utils/constants.tsx
angle943: yeah this should be moved out of constants.

i don't think its nit. we may have checks in the future where we ignore `constants.ts` in testing coverage, but this has functional logic in it, so we don't want it ignored

src/plugins/explore/public/application/register_tabs.ts
paulstn: Changing it to be required

paulstn: `tabRegistry` is passed in as `services.tabRegistry`. it seems slightly redundant to me that we have both of these parameters. do you think i should get rid of `tabRegistry` being passed in and then define `tabRegistry` within `registerBuiltInTabs`?

src/plugins/explore/public/application/register_tabs.ts
paulstn: Would it be in `services.store`? If its just that, I'll try removing everything referencing state here.

src/plugins/explore/public/application/register_tabs.ts
paulstn: removed redundant check

src/plugins/explore/public/application/register_tabs.ts
paulstn: This code works because onInactive is only implemented for a dataset change. However, this isn't the best option since that's not what its meant to do. I'm going to change this to work for when tabs are being switched as well.

src/plugins/explore/public/application/utils/state_management/actions/set_dataset/set_dataset.ts
paulstn: If we're on the logs tab or the patterns tab, and I'm switching the dataset, I would say that the tab becomes inactive. 
Maybe we need to clear up the definition of inactive in this case, the way I see it is:

Since the entire query and what its querying is changing, we're changing everything about the page, and it could be considered inactive, since later, it'll become active again but with a new dataset. 

So I'm just seeing it as a hard reset cycle.
The reason I didn't just dispatch `setPatternsField('')` was because of extensibility. What if later on, I need to add some more logic here other than just dispatching one redux change? Calling upon my tab's onInactive seems to open to the door for that.

src/plugins/explore/public/components/patterns_table/patterns_container.tsx
paulstn: changed

src/plugins/explore/public/components/patterns_table/patterns_container.tsx
paulstn: changed

src/plugins/explore/public/application/utils/hooks/use_tab_results.ts
paulstn: removed this files changes

src/plugins/explore/public/application/register_tabs.ts
paulstn: removed state

src/plugins/explore/public/application/register_tabs.ts
paulstn: removed

src/plugins/explore/public/application/register_tabs.ts
paulstn: Removing this still results in working code. It's because on a dataset switch, onActive gets called again, which will result in a new default patterns field being selected, overriding the old one.
So onInactive is currently unimplemented essentially

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -45,17 +47,31 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
     supportedLanguages: [EXPLORE_DEFAULT_LANGUAGE],
 
     prepareQuery: (query) => {
-      return typeof query.query === 'string' ? query.query : '';
+      const state = services.store.getState();
+
+      // Get the selected patterns field from the Redux state
+      let patternsField = state.tab?.patterns?.patterns_field;
```
angle943:
```suggestion
      let patternsField = state.tab.patterns.patterns_field;
```

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -45,17 +47,31 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
...
+
+      const preparedQuery = getQueryWithSource(query);
+      if (!patternsField) {
+        patternsField = findDefaultPatternsField(services);
+      }
+
+      return preparedQuery.query !== ''
+        ? preparedQuery.query +
+            ` | patterns \`${patternsField}\` method=brain mode=aggregation | sort - pattern_count`
```
angle943: since you are already using string interperlation, lets just use that all the way for readability. IE instead of:

```
? preparedQuery.query +
            ` | patterns \`${patternsField}\` method=brain 
```

lets do:

```
`${preparedQuery.query} | patterns \`${patternsField}\` method=brain mode=aggregation | sort - pattern_count`
```

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -45,17 +47,31 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
...
+
+      // Get the selected patterns field from the Redux state
+      let patternsField = state.tab?.patterns?.patterns_field;
+
+      const preparedQuery = getQueryWithSource(query);
+      if (!patternsField) {
+        patternsField = findDefaultPatternsField(services);
+      }
+
+      return preparedQuery.query !== ''
```
angle943: I don't think `preparedQuery.query` can ever be an empty string. But even if it was, would returning the empty string from `prepareQuery` ever make sense for the patterns tab?

I personally think we shouldn't even do this check, becuse:

1. (less important) i don't think empty string is possible
2. (more important) the result of what you have when it is empty seems to me to be an inaccurate behavior. we don't ever want to pass an empty string to the patterns page, right?

src/plugins/explore/public/application/register_tabs.ts
paulstn: I intentionally included these because I believe there were situations where I've seen patterns_field not be a field inside of options? Although should I be making sure this is never the case?

src/plugins/explore/public/application/register_tabs.ts
paulstn: I'm not an expert on the behavior of `preparedQuery.query`, but I added this because there used to be an initial load of the page where the query was an empty string, and that would crash the page with my `prepareQuery`. Let me try removing this check to see if this is not needed anymore.

src/plugins/explore/public/application/register_tabs.ts
paulstn: changed anyways

src/plugins/explore/public/application/register_tabs.ts
paulstn: changed

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -159,11 +159,8 @@ export const executeQueries = createAsyncThunk<
   // Check what needs execution
   const needsDefaultQuery = !results[defaultCacheKey];
   const needsVisualizationTabQuery =
-    query.query !== '' &&
-    visualizationTabCacheKey !== defaultCacheKey &&
-    !results[visualizationTabCacheKey];
+    visualizationTabCacheKey !== defaultCacheKey && !results[visualizationTabCacheKey];
   const needsActiveTabQuery =
-    query.query !== '' &&
```
angle943, line 166: why were these removed? I think they are needed

angle943: actually i am always confused by this. I just remember i tried removing these a couple times but changed my mind for reasons I don't remember. lets get @ananzh here for input

src/plugins/explore/public/components/patterns_table/utils/constants.tsx
```diff
@@ -5,11 +5,22 @@
...
+export const SAMPLE_FIELD = 'sample_logs';
+
+// used to highlight sections in the sample log column of table
+export const DELIM_START = '<*';
+export const STD_DELIM_END = '>';
+export const UNIQ_DELIM_END = '*>';
+export const MARK_START = '<mark>';
+export const MARK_END = '</mark>';
 
 export const patternsTableColumns: Array<EuiBasicTableColumn<PatternItem>> = [
```
angle943: @paulstn casn you move this to its own file, maybe csll it `patterns-table_columns.ts` and add simple unit tests on it

src/plugins/explore/public/components/tabs/tabs.tsx
```diff
@@ -37,6 +37,12 @@ export const ExploreTabs = () => {
       dispatch(setActiveTab(selectedTab.id));
 
       const activeTab = services.tabRegistry.getTab(selectedTab.id);
+
+      if (activeTab?.onActive) {
+        activeTab?.onActive();
+      }
+      // TODO: put onInactive here? i mean the tabs are technically changing here as well
```
angle943: yes lets add it. but lets call it before the new active tab is active and before all the dispatch, ie at the top of the function:

```
const onTabsClick = useCallback(
    (selectedTab: EuiTabbedContentTab) => {
        const previousActiveTab = services.tabRegistry.getTab(activeTabId);
        previousActiveTab?.onInactive?.();
        
        dispatch(setActiveTab(selectedTab.id))
        const newActiveTab = services.tabRegistry.getTab(selectedTab.id);
        newActiveTab?.onActive?.();

```

src/plugins/explore/public/application/register_tabs.ts
angle943: I added comment in the tabs component about how to implement the onInactive. 

It should fire every time you tab out of the tab, because IMO that is what the name `onInactive` implies.

With that in mind I don't think this `services.store.dispatch(setPatternsField(''));` is no longer applicable here, since we dont' want that on every tab switch, so for dataset changes i think you should manually call this dispatch there

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
paulstn: I removed this because I found that the boolean wasn't true for situations in which it was meant to be true. query.query was `''` even when there was a query in the active tab that needed to be executed. In other words, I saw that while the active tab query was needed, `needsActiveTabQuery` was false.
I suspect this could be because query.query might not reflect the actual query, but I'm not sure.

src/plugins/explore/public/components/tabs/tabs.tsx
paulstn: oh sorry this todo was not meant to be here. i'll add this in.

src/plugins/explore/public/components/patterns_table/utils/constants.tsx
paulstn: added

src/plugins/explore/public/components/tabs/tabs.tsx
paulstn: removed onActive + todo for onInactive

src/plugins/explore/public/application/utils/state_management/actions/set_dataset/set_dataset.ts
```diff
@@ -40,6 +42,8 @@ export const setDatasetActionCreator = (
   dispatch(clearResults());
   dispatch(clearQueryStatusMap());
   dispatch(clearLastExecutedData());
+  dispatch(setPatternsField(''));
```
kavilla: i would just use https://github.com/kavilla/OpenSearch-Dashboards-1/blob/kavilla/dsselect_move/src/plugins/explore/public/application/utils/state_management/middleware/dataset_change_middleware.ts#L55

src/plugins/explore/public/services/tab_registry/tab_registry_service.ts
```diff
@@ -34,15 +34,17 @@ export interface TabDefinition {
...
-  // Optional results processor for raw results
-  resultsProcessor?: (rawResults: any, indexPattern: any, includeHistogram?: boolean) => any;
+  /**
+   * @experimental Callback for errors from query
+   * Used for handling patterns queries which fail on older versions.
+   * Will be removed when a method of finding dataset query engine version is implemented
+   *
+   * @returns A custom QueryExecutionStatus as well as an error, or undefined if standard behavior is desired
+   */
+  handleQueryError?: (error: any, cacheKey: string) => boolean;
```
kavilla, line 44: will this cause issues?

src/plugins/explore/public/services/tab_registry/tab_registry_service.ts
```diff
@@ -34,15 +34,17 @@ export interface TabDefinition {
...
 
   // UI Components
   component: (() => React.JSX.Element | null) | MemoExoticComponent<() => React.JSX.Element>;
-
-  // Optional lifecycle hooks
-  onActive?: () => void;
```
kavilla, line 44: is this ok? we are deleting these optional things. are they needed? should we just keep them there juset in case

src/plugins/explore/public/application/utils/state_management/actions/set_dataset/set_dataset.ts
paulstn: using this

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -372,26 +370,33 @@ const executeQueryBase = async (
...
-            error: error.body.error || 'Unknown Error',
-            message: {
-              details: parsedError?.error?.details || 'Unknown Error',
-              reason: parsedError?.error?.reason || 'Unknown Error',
-              type: parsedError?.error?.type,
+    // if there is no avoidDispatchingError function, dispatch Error.
+    // if there is that function, and it returns false, dispatch Error
+    if (
+      !avoidDispatchingError ||
+      (avoidDispatchingError && !avoidDispatchingError(parsedError, cacheKey))
```
kavilla, line 376: instead of current approach, which from my understanding will return back from to the client side a 400 error first if it fails adds some custom logic to prevent dispatching an error in that case in the ui and then fire the query again from the client side. which might expose too much implementation details and be confusing to anyone who is watching. so if some host has alerting associated to opensearch dashboards api calls and it's constantly returning back a 400 then it might be treated as a problem but in reality it is what we are expecting.

 we should fast follow and move this logic to Node server. if we can't check PPL features/functions before query execution, then Node server should:
* try original query first
* if it returns 400, fallback to regex query
* return back 200 if it succeeds for that secondary or return back the failure to explore

this removes need for custom error handling in client-side public folder and keeps opensearch errors internal to server. cleaner approach overall

src/plugins/explore/public/services/tab_registry/tab_registry_service.ts
paulstn: some potential issues (from our discussion but for everyone else):
- security related issues (the 'correct' user flow means sending out a call that will return with a 400)
- potential race conditions (in between the time waiting for calls, regex state being changed, strange UI results)

src/plugins/explore/public/application/utils/state_management/actions/set_dataset/set_dataset.ts
```diff
@@ -40,6 +42,8 @@ export const setDatasetActionCreator = (
...
+  dispatch(setUsingRegexPatterns(false));
```
kavilla: if we go with the server side approach then we should be able to avoid potential race condition where if a previous query failed setting this to false while the dataset was updated

src/plugins/explore/public/services/tab_registry/tab_registry_service.ts
paulstn: a potential future solution that could be implemented which wouldn't have these issues (from your suggestion):
server side handling of this flow.
1) checking if the query server can handle 'brain' pattern queries, by sending out that query
2) intercepting an error, checking to see if it matches the criteria of the server being too old
3) sending out the older patterns query, and marking in the result that it is so
This would solve the two issues above.

src/plugins/explore/public/services/tab_registry/tab_registry_service.ts
paulstn: Yes, @angle943 was mentioning that since these aren't implemented nor are they being used, to get rid of these definitions. There aren't plans to implement them from what I could understand.

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
paulstn: i like it, will do this for a fast follow.

kavilla (approved):

abbyhu2000 (approved):

PR #10214, opened 2025-07-17
[WIP] Discover: Trace

TackAdam:
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/register_tabs.ts
+++ b/src/plugins/explore/public/application/register_tabs.ts
@@ -4,59 +4,85 @@
  */
 
 import { LogsTab } from '../components/tabs/logs_tab';
+import { TracesTab } from '../components/tabs/traces_tab';
 import { TabRegistryService } from '../services/tab_registry/tab_registry_service';
 import { PatternsTab } from '../components/tabs/patterns_tab';
 import { ExploreServices } from '../types';
-import { EXPLORE_DEFAULT_LANGUAGE } from '../../common';
+import { EXPLORE_DEFAULT_LANGUAGE, ExploreFlavor } from '../../common';
 import { VisTab } from '../components/tabs/vis_tab';
 
 /**
  * Registers built-in tabs with the tab registry
  */
-export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
-  // Register Logs Tab
-  const logsTabDefinition = {
-    id: 'logs',
-    label: 'Logs',
-    flavor: [],
-    order: 10,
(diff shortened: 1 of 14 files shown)


src/plugins/explore/public/components/tabs/traces_tab.tsx
```diff
@@ -0,0 +1,39 @@
...
+      setColumns([
+        'timestamp',
+        'attributes.http.status_code',
+        'status.code',
+        'attributes.rpc.method',
+        'serviceName',
+        'durationInNanos',
+      ])
+    );
+  }, [dispatch]);
```
angle943, line 31: should we do this in the preloaded state of redux when it mounts for traces? I want to decrease the use of useEffects as much as possible in the codebase, and they are typically required when we need the react app to be in sync with some outside sync (like an api call), but this does not seem to be the case here

src/plugins/explore/public/components/formatters/use_status_code_formatter.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+      // Use the original formatField method for other fields
+      return originalFormatField(hit, fieldName);
+    };
+
+    return () => {
+      dataset.formatField = originalFormatField;
+    };
+  }, [dataset]);
+};
```
angle943, line 42: can this code live in the dataset object?

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -89,8 +115,20 @@ export const registerBuiltInTabs = (tabRegistry: TabRegistryService) => {
...
+  const path = window.location.pathname;
+  let flavor: ExploreFlavor | undefined;
+
+  if (path.includes(`/${ExploreFlavor.Logs}`)) {
+    flavor = ExploreFlavor.Logs;
+  } else if (path.includes(`/${ExploreFlavor.Traces}`)) {
+    flavor = ExploreFlavor.Traces;
+  } else if (path.includes(`/${ExploreFlavor.Metrics}`)) {
+    flavor = ExploreFlavor.Metrics;
+  }
```
angle943, line 130: lets throw an error if its not one of them

src/plugins/explore/public/components/container/bottom_container/bottom_container.tsx
Maosaic, line 1: I don't think it's ideal to do context switch at this container component level, I will propose each page(logs/traces) should build their own container.

PR #10222, opened 2025-07-17
[Explore] Merging Explore Feature Branch into main

sejli:
### Description

Rebased feature branch `query_explore` to merge into `main`

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ruanyl (approved):

ruanyl (approved):

angle943 (approved):

ruanyl (approved):

ruchidh (approved):

ruchidh:
Lighthouse performance metrics have breached the baseline. so failing ci. We need to reset the baseline and check where performance is degrading separately. For this pr, please ignore ci workflow of lighthouse, as this pr has development on explore plugin, not any major work dashboard or discover.

[
  {
    "name": "maxNumericValue",
    "expected": 28000,
    "actual": 28366.949145667117,
    "values": [
      28366.949145667117
    ],
    "operator": "<=",
    "passed": false,
    "auditId": "speed-index",
    "level": "warn",
    "url": "http://localhost:5601/app/dashboards",
    "auditTitle": "Speed Index",
    "auditDocumentationLink": "https://developer.chrome.com/docs/lighthouse/performance/speed-index/"
  },
  {
    "name": "maxNumericValue",
    "expected": 26000,
    "actual": 26869.61797359369,
    "values": [
      26869.61797359369
    ],
    "operator": "<=",
    "passed": false,
    "auditId": "speed-index",
    "level": "warn",
    "url": "http://localhost:5601/app/data-explorer/discover",
    "auditTitle": "Speed Index",
    "auditDocumentationLink": "https://developer.chrome.com/docs/lighthouse/performance/speed-index/"
  }
]

angle943 (dismissed):

LDrago27 (dismissed):

angle943:
ciGroup15Explore is broken with my change due to the fact that suggest window no longer turns on when focused by design. This is not a regression but a design decision, as pressing "spacebar" to enter AI mode will not work if suggest window is shown on focus. I don't think we should block merge due to it

ruchidh (approved):

changelogs/fragments/10001.yml
kavilla, line 1: do we need all the changelogs? since this will go into as a single commit?

cypress/support/e2e.js
```diff
@@ -5,6 +5,7 @@
 
 import '../utils/commands';
 import '../utils/commands.osd';
+import '../utils/commands.explore';
```
kavilla, line 8: were they different that we couldn't leverage the same osd commands?

cypress/utils/apps/explore/constants.js
```diff
@@ -3,8 +3,39 @@
...
+  DELETE_DATA_SOURCE: `${DS_API_PREFIX}/data-source/`,
+};
+export const DSM_API = '/internal/data-source-management/fetchDataSourceMetaData';
+
+export const BASE_QUERY_ENHANCEMENTS_API = '/api/enhancements';
+export const JOBS_API = {
+  DELETE: `${BASE_QUERY_ENHANCEMENTS_API}/jobs`,
+};
+
+export const INDEX_WITH_TIME_1 = 'data_logs_small_time_1';
```
kavilla, line 28: this seems redundant since i think we reference the same stuff from the discover tests

cypress/utils/apps/explore/shared.js
```diff
@@ -52,3 +89,146 @@ export const generateAllExploreTestConfigurations = (
...
+/**
+ * Gets Date and returns string of date in format MMM dd, yyyy @ HH:mm:ss.SSS
+ * @param {Date} date - date to format
+ * @returns {string} - in the format e.g. Jan 24, 2025 @ 16:20:08.000
+ * @example formatDate(new Date('2011-10-05T14:48:00.000Z'))
+ * // returns 'Oct 5, 2011 @ 14:48:00.000'
+ */
+export const formatDate = (date) => {
+  const months = [
+    'Jan',
```
kavilla, line 165: we def have this defined already right?

package.json
```diff
@@ -105,8 +105,11 @@
     "osd:ciGroup14Explore": "echo \"cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/*.spec.js\"",
     "osd:ciGroup15Explore": "echo \"cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/05/*.spec.js\"",
     "osd:ciGroupS3Explore": "echo \"cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/s3_tests/*.spec.js\"",
-    "generate:opensearchsqlantlr": "./node_modules/antlr4ng-cli/index.js -Dlanguage=TypeScript -o ./src/plugins/data/public/antlr/opensearch_sql/.generated -visitor -no-listener -Xexact-output-dir ./src/plugins/data/public/antlr/opensearch_sql/grammar/OpenSearchSQLLexer.g4 ./src/plugins/data/public/antlr/opensearch_sql/grammar/OpenSearchSQLParser.g4",
```
kavilla, line 108: i'm not sure if any system is relying on this script but do we think we want to maybe remove this in a major version update but have defined generate:opensearchsqlantlr and generate:grammars:sql with the earlier pointing to the later. while the earlier outputs a warning deprecated message? that way if any system is built on top of this it wont fail

cypress/support/e2e.js
angle943: yes because they have different interactions or data test subj. We can think about consolidating them later, and honestly the explore tests are written very messy, but I did not want to intermingle the two tests and possibly cause a regression in the enhancements test

packages/osd-antlr-grammar/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "generate:opensearchppl": "../../node_modules/antlr4ng-cli/index.js -Dlanguage=TypeScript -o ./src/opensearch_ppl/.generated -visitor -no-listener -Xexact-output-dir ./src/opensearch_ppl/grammar/OpenSearchPPLLexer.g4 ./src/opensearch_ppl/grammar/OpenSearchPPLParser.g4",
+    "generate:opensearchppl-simplified": "../../node_modules/antlr4ng-cli/index.js -Dlanguage=TypeScript -o ./src/opensearch_ppl_simplified/.generated -visitor -no-listener -Xexact-output-dir ./src/opensearch_ppl_simplified/grammar/OpenSearchPPLLexer.g4 ./src/opensearch_ppl_simplified/grammar/OpenSearchPPLParser.g4",
+    "generate:dql": "../../node_modules/antlr4ng-cli/index.js -Dlanguage=TypeScript -o ./src/dql/.generated -visitor -no-listener -Xexact-output-dir ./src/dql/grammar/DQLLexer.g4 ./src/dql/grammar/DQLParser.g4",
+    "generate:all": "yarn generate:opensearchsql && yarn generate:opensearchppl && yarn generate:opensearchppl-simplified && yarn generate:dql"
+  },
+  "dependencies": {
+    "antlr4ng": "^3.0.4"
```
kavilla, line 19: does that mean we can remove antlr from the root level package (if we haven't already)

packages/osd-antlr-grammar/src/index.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './opensearch_ppl';
```
kavilla, line 6: maybe alphanumeric order?

packages/osd-monaco/ppl-test.html
kavilla, line 1: what is this for? i think we should delete this. osd monaco doesn't need any context on ppl.

and i think this will get bundled in a release

cypress/utils/apps/explore/constants.js
angle943: just fyi, all `utils` for enhancements have been copied and pasted to have explore counterpart. In my initial Cypress test addition we initially only branched out of enhancement specific utils when we needed to, but then as we progressed it became very hard to manage, as further utils needed to be branched out and it was proving difficult to keep track.

We should consolidate them but i believe this should be done separately, with deep thought, instead of sweating and monkeying around the code half-hazardly when you are working on a feature

cypress/utils/apps/explore/shared.js
angle943: yes, see comment above

packages/osd-monaco/package.json
```diff
@@ -10,7 +10,9 @@
   },
   "dependencies": {
     "regenerator-runtime": "^0.13.3",
-    "monaco-editor": "^0.52.0"
+    "monaco-editor": "^0.52.0",
+    "antlr4ng": "^3.0.4",
```
kavilla, line 14: why does the osd monaco package need antlr?

packages/osd-monaco/src/ppl/constants.ts
```diff
@@ -0,0 +1,367 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const ID = 'PPL';
+
+/**
+ * PPL token type lookup table organized by Monaco editor token class
+ * Maps Monaco token types to Sets of ANTLR token types for O(1) lookup performance
+ */
+export const PPL_TOKEN_SETS = {
```
kavilla, line 12: what's the difference between this file and the xjson/lexer_rules/opensearchppl?

packages/osd-antlr-grammar/package.json
LDrago27: Nope. The osd-antlr-grammar only houses the grammar and generated files. But besides that in the root package we are using antlr library for creating AST and suggestions in Dataplugin.

packages/osd-monaco/src/xjson/lexer_rules/index.ts
```diff
@@ -45,7 +44,5 @@ export const registerLexerRules = (m: typeof monaco) => {
   m.languages.setMonarchTokensProvider(opensearchql.ID, opensearchql.lexerRules);
   m.languages.register({ id: opensearchsql.ID });
   m.languages.setMonarchTokensProvider(opensearchsql.ID, opensearchsql.lexerRules);
-  m.languages.register({ id: opensearchppl.ID });
-  m.languages.setMonarchTokensProvider(opensearchppl.ID, opensearchppl.lexerRules);
```
kavilla, line 49: should we then delete those files if they aren't registered and how come we we still keep sql in here?

src/plugins/dashboard/public/plugin.tsx
```diff
@@ -219,6 +220,9 @@ export class DashboardPlugin
     core: CoreSetup<StartDependencies, DashboardStart>,
     { share, uiActions, embeddable, home, urlForwarding, data, usageCollection }: SetupDependencies
   ): DashboardSetup {
+    // bootstrap UI Actions
```
kavilla, line 223: unneeded comment

src/plugins/data/public/antlr/opensearch_ppl/simplified_ppl_grammar/opensearch_ppl_autocomplete.ts
kavilla, line 1: @LDrago27 i thought we were gunna move these out of the data plugin?

package.json
LDrago27: @kavilla I don;t think anyone is relying on this script in particular. This script is used to generate the .generated folder from the g4 files. I think we should be safe removing the script.
In the current change we have consolidated and moved all the grammar file from the data plugin to the new package. Having both the scripts might create confusion and might lead us to have same files in multiple locations. 

But if you think we need it, we can add back the older scripts.

packages/osd-monaco/ppl-test.html
LDrago27: Yeah you are right. let me remove this.

src/plugins/data/public/query/query_string/language_service/types.ts
```diff
@@ -66,4 +67,7 @@ export interface LanguageConfig {
   supportedAppNames?: string[];
   hideDatePicker?: boolean;
   sampleQueries?: SampleQuery[];
+  addFiltersToQuery?: (query: string, filters: Filter[]) => string;
```
kavilla, line 70: query throughout the data plugin refers to an object. here we are going with string why? this conflicts with other functions in this same file reference query as an object making it inconsistent when registering a new language

src/plugins/data/public/query/query_string/language_service/types.ts
```diff
@@ -66,4 +67,7 @@ export interface LanguageConfig {
...
+  /** Add filters to the natural language prompt. */
+  addFiltersToPrompt?: (prompt: string, filters: Filter[]) => string;
```
kavilla, line 72: this seems like it if we passed a query object with a language we could do some switch case instead of defining the same function twice

changelogs/fragments/10001.yml
angle943: this will not go in as a single commit

packages/osd-monaco/package.json
LDrago27: Well , In monaco editor for lexical highlighting and error listeners it creates a PPL worker. This PPL worker uses ANTLR package to create to validate the syntax and listen for errors.

src/plugins/data/public/query/saved_query/saved_query_service.ts
```diff
@@ -38,6 +38,8 @@ import { UI_SETTINGS } from '../../../common';
 
 type SerializedSavedQueryAttributes = SavedObjectAttributes & SavedQueryAttributes;
 
+const unregisteredLangServiceApps = ['explore'];
```
kavilla, line 41: why couldn't this be a config instead of hardcoded?

src/plugins/data/public/ui/saved_query_management/saved_query_management_component.tsx
```diff
@@ -70,6 +69,7 @@ interface Props {
   onClearSavedQuery: () => void;
   closeMenuPopover: () => void;
   saveQuery: (savedQueryMeta: SavedQueryMeta, saveAsNew?: boolean) => Promise<void>;
+  saveQueryIsDisabled?: boolean;
```
kavilla, line 72: i'm confused how would saved query be disabled from the saved query management page?

packages/osd-monaco/src/ppl/constants.ts
LDrago27: @kavilla 
xjson/lexer_rules/opensearchppl was something that we used to have prior to this release. Where in order to tokenize and parse users query we were using regex along monarch tokenizer (default tokenizer provided by monaco editor). This didn't take into account the g4 grammar files and just used regex to do it. 

In the current approach we are creating a AST using ANTLR and using that to tokenize the ppl query instead of relying on regex. However it introduces a new problem ANTLR tokenizes the query and returns ANTLR equivalent tokens which can't be understood by monaco editor. (Since we are no longer using the monarch tokenizer).  Therefore this file defines how do we want to map the ANTLR generated tokens into a monaco editor tokens

packages/osd-monaco/src/xjson/lexer_rules/index.ts
LDrago27: Yeah. I might have missed these cleanup

src/plugins/data/public/ui/search_bar/index.tsx
```diff
@@ -46,3 +46,4 @@ export const SearchBar = injectI18n(withOpenSearchDashboards(WrappedSearchBar));
 export { StatefulSearchBarProps } from './create_search_bar';
 export type { SearchBarProps, SearchBarOwnProps } from './search_bar';
 export { useQueryStringManager } from './lib/use_query_string_manager';
+export { useSavedQuery } from './lib/use_saved_query';
```
kavilla, line 49: what does this do?

src/plugins/explore/public/application/components/recent_query_selector.tsx
```diff
@@ -0,0 +1,32 @@
...
+export interface RecentQuerySelectorProps {
+  size?: 'xs' | 's' | 'l';
+}
+
+/**
+ * Placeholder component for recent query selection
+ * TODO: Implement full functionality with query history
+ */
+export const RecentQuerySelector: React.FC<RecentQuerySelectorProps> = ({ size = 's' }) => {
+  const handleClick = () => {
```
kavilla, line 18: does recent query not work or something?

src/plugins/explore/public/application/hooks/editor_hooks/use_change_query_editor/index.ts
```diff
@@ -0,0 +1,6 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './use_change_query_editor';
```
kavilla, line 6: guess i'm little lost we have an index file but it only references the same file by name can we shorten this filepath but just reference the hooks with file names vs having a folder for each?

src/plugins/data/public/antlr/opensearch_ppl/simplified_ppl_grammar/opensearch_ppl_autocomplete.ts
LDrago27: Oh. So we moved out the grammar files into osd-antlr-grammar package. Unfortunately, the autocomplete also relies on client,  API calls to get the value for a column, This made it to difficult to move them out into the package. 

Therefore the core business logic still rests within the data plugin. 

Maybe we can revisit and move them out of data plugin into another dedicated plugin in a future refactor.

src/plugins/explore/public/application/hooks/editor_hooks/use_change_query_editor/use_change_query_editor.test.ts
```diff
@@ -0,0 +1,196 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { renderHook } from '@testing-library/react-hooks';
```
kavilla, line 6: i'm worried the structure of the query panel being too coupled with explore plugin. if the goal is to make this eventually global then it should likely exist outside of the explore plugin but with this current implementation it would be difficult to extract that out in the future. which would then make that itself a larger project.

are we sure we want to say this is global accessible.

src/plugins/explore/public/application/hooks/editor_hooks/use_change_query_editor/use_change_query_editor.ts
```diff
@@ -0,0 +1,59 @@
...
+    services: {
+      data: {
+        query: { filterManager, queryString },
+      },
+    },
+  } = useOpenSearchDashboards<ExploreServices>();
+  const { dataset } = useDatasetContext();
+  const setEditorText = useSetEditorText();
+  const editorMode = useSelector(selectEditorMode);
+  const query = useSelector(selectQuery);
```
kavilla, line 29: @ananzh this is kind of what i was talking about. i updated this file to use the context but it was existing there. how come we have a dataset context but then pull the query from the querystring manager. the query should have the dataset already so we kind of have two sources of truth.

src/plugins/explore/public/application/hooks/editor_hooks/use_editor_focus/index.ts
```diff
@@ -0,0 +1,6 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './use_editor_focus';
```
kavilla, line 6: this needs a use hook?

src/plugins/explore/public/application/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { Router, Route, Switch } from 'react-router-dom';
+import { Provider as ReduxProvider } from 'react-redux';
+import { Store } from 'redux';
+import { AppMountParameters } from '../../../../core/public';
+import { ExploreServices } from '../types';
+import { LogsPage } from './pages/logs';
```
kavilla, line 13: could we define pages/index and then just import all these from the same line

src/plugins/explore/public/application/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  ReactDOM.render(
+    <Router history={history}>
+      <OpenSearchDashboardsContextProvider services={services}>
+        <ReduxProvider store={store}>
+          <EditorContextProvider>
+            <DatasetProvider>
+              <services.core.i18n.Context>
+                <Switch>
+                  {/* View route for saved searches */}
+                  {/* TODO: Do we need this? We might not need to, please revisit */}
```
kavilla, line 72: is this relevant?

src/plugins/explore/public/application/pages/logs/logs_page.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+import React from 'react';
+import { EuiErrorBoundary, EuiPage, EuiPageBody } from '@elastic/eui';
+import { AppMountParameters, HeaderVariant } from 'opensearch-dashboards/public';
+import { useOpenSearchDashboards } from '../../../../../opensearch_dashboards_react/public';
+import { ExploreServices } from '../../../types';
+import { QueryPanel } from '../../../components/query_panel';
+import { useInitialQueryExecution } from '../../utils/hooks/use_initial_query_execution';
+import { useUrlStateSync } from '../../utils/hooks/use_url_state_sync';
+import { useTimefilterSubscription } from '../../utils/hooks/use_timefilter_subscription';
```
kavilla, line 16: i think with an index file we could clean up this import a lot better

src/plugins/explore/public/application/pages/logs/logs_page.tsx
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import '../explore_page.scss';
```
kavilla, line 6: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/DEVELOPER_GUIDE.md#sass-files

kavilla: that or we structure these components to inherit the same classes

src/plugins/explore/public/application/hooks/editor_hooks/use_editor_focus/index.ts
angle943: yes. there is an editor context that exposes the editor ref, editor focus state, and editor text. Because the editor text updates frequently, having one hook that grabs all data will be unperformant for the consumers. Therefore i made specific hooks with memoization for optimation, such that consumers of this hook for example do not need to be re-rendered when the text changes

src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts
```diff
@@ -0,0 +1,103 @@
...
+          dispatch(setChartType(chartType));
+          dispatch(setStyleOptions(params));
+          dispatch(setAxesMapping(axesMapping));
+        }
+        if (uiState) {
+          const { activeTab } = JSON.parse(uiState);
+          dispatch(setActiveTab(activeTab));
+        }
+
+        // Add to recently accessed
```
kavilla, line 76: unneeded

src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts
```diff
@@ -0,0 +1,103 @@
...
+          const searchSource = JSON.parse(searchSourceFields.searchSourceJSON);
+          const query = searchSource.query;
+          if (query) {
+            dispatch(setQueryState(query));
+            setEditorText(query.query);
+          }
+        }
+
+        // Update savedSearch to store just the ID (like discover)
+        // TODO: remove this once legacy state is not consumed any more
```
kavilla, line 59: is this the case?

src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts
```diff
@@ -0,0 +1,103 @@
...
+        dispatch(clearLastExecutedData());
+        dispatch(setEditorMode(EditorMode.Query));
+        dispatch(clearResults());
+        dispatch(clearQueryStatusMap());
+        dispatch(executeQueries({ services }));
+      }
+    }
+    if (error) {
+      // Navigate to management page for invalid IDs
+      // TODO: need to confirm the UI behavior for invalid ID, the current logic is copied from useSavedExplore hook
```
kavilla, line 93: did we confirm?

src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts
```diff
@@ -0,0 +1,103 @@
...
+        chrome.recentlyAccessed.add(
+          `/app/explore/${savedExplore.type ?? ExploreFlavor.Logs}#/view/${savedExplore.id}`,
+          title,
+          savedExplore.id,
+          { type: 'explore' }
+        );
+
+        dispatch(clearLastExecutedData());
+        dispatch(setEditorMode(EditorMode.Query));
+        dispatch(clearResults());
```
kavilla, line 86: i notice we kind of done a lot of of this? was there way to have a single dispatch event that does this. because i think i had to dispatch all of these and becomes a little bit not maintainable. would be easier to dispatch a single event and then who ever consumed this handled this

src/plugins/explore/public/application/utils/interfaces.ts
kavilla, line 1: i think the file name should be `types.ts` for consistency

src/plugins/explore/public/application/utils/interfaces.ts
```diff
@@ -0,0 +1,164 @@
...
+  _id: string;
+  _index?: string;
+  _type?: string;
+  _score?: number;
+}
+
+/**
+ * Interface for raw search results from OpenSearch
+ */
+export interface RawSearchResults {
```
kavilla, line 70: this doesn't look different from the data plugin why redefined the interface?

src/plugins/explore/public/application/utils/interfaces.ts
```diff
@@ -0,0 +1,164 @@
...
+  sort: number[];
+  _source: Record<string, unknown>;
+  _id: string;
+  _index?: string;
+  _type?: string;
+  _score?: number;
+}
+
+/**
+ * Interface for raw search results from OpenSearch
```
kavilla, line 68: technically doesn't have to be true it's just an OpenSearch structured response. it can be any data source

src/plugins/explore/public/application/utils/interfaces.ts
```diff
@@ -0,0 +1,164 @@
...
+  };
+}
+
+/**
+ * Base interface for processed search results
+ */
+export interface BaseProcessedSearchResults {
+  hits: ISearchResult['hits'];
+  fieldCounts: Record<string, number>;
+  dataset: Dataset;
```
kavilla, line 97: might be easier if we passed query

src/plugins/explore/public/application/utils/interfaces.ts
```diff
@@ -0,0 +1,164 @@
...
+  tabId: string;
+}
+
+/**
+ * Interface for query execution options
+ */
+export interface QueryExecutionOptions {
+  clearCache?: boolean;
+  services?: ExploreServices;
+  reason?: 'user_action' | 'tab_switch' | 'dataset_change';
```
kavilla, line 142: how is this used?

src/plugins/explore/public/application/utils/interfaces.ts
```diff
@@ -0,0 +1,164 @@
...
+  preparedQueries?: PreparedQuery[];
+}
+
+/**
+ * Interface for tab query execution options
+ */
+export interface TabQueryExecutionOptions {
+  clearCache?: boolean;
```
kavilla, line 150: why are all these optional? i dont think these should all be optional

src/plugins/explore/public/application/utils/interfaces.ts
```diff
@@ -0,0 +1,164 @@
...
+  services?: ExploreServices;
+  tabId?: string;
+  preparedQuery?: PreparedQuery;
+  cacheKey?: string;
+}
+
+/**
+ * Interface for Redux store state
+ */
+export interface ReduxStore {
```
kavilla, line 160: this seems too buried. is the application folder the only one that knows about the redux store

src/plugins/explore/public/application/utils/languages/ppl/default_prepare_ppl_query/default_prepare_ppl_query.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import { Query } from '../../../../../../../data/common';
+import { stripStatsFromQuery } from '../strip_stats_from_query';
+import { getQueryWithSource } from '../get_query_string_with_source';
+import { QueryWithQueryAsString } from '../../types';
+
+/**
+ * Default PPL query preparation for tabs (removes stats pipe for histogram compatibility)
+ */
+export const defaultPreparePplQuery = (query: Query): QueryWithQueryAsString => {
```
kavilla, line 14: tbh i wouldnt know that all defaultPreparePplQuery only strips the stats call. are we sure this is needed?

kavilla: it kind of conflicts with the initial query

src/plugins/explore/public/application/utils/languages/ppl/get_query_string_with_source/get_query_with_source.ts
```diff
@@ -0,0 +1,35 @@
...
+  const hasSource = /^\s*(search\s+)?source\s*=/.test(lowerCaseQuery);
+  const hasDescribe = /^\s*describe\s+/.test(lowerCaseQuery);
+
+  if (hasSource || hasDescribe) {
+    return { ...query, query: queryString };
+  }
+
+  const datasetTitle = query.dataset?.title || '';
+
+  let queryStringWithSource: string;
```
kavilla, line 24: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/DEVELOPER_GUIDE.md#avoid-using-var-to-declare-variables

i'm guilty of this too but i think we could make this better

src/plugins/explore/public/application/utils/state_management/README.md
```diff
@@ -0,0 +1,313 @@
+# Explore Plugin State Management
+
+## Overview
+
+The Explore plugin uses Redux Toolkit with a slice architecture for state management, featuring middleware-based persistence and logical separation of UI and infrastructure state.
```
kavilla, line 5: could this be in the docs folder?

src/plugins/explore/public/application/utils/state_management/actions/abort_controller/abort_data_query_action.ts
```diff
@@ -22,10 +22,8 @@ export function createAbortDataQueryAction(
     shouldAutoExecute: async () => true,
     execute: async (context: AbortDataQueryContext) => {
       try {
-        refs.forEach((ref) => {
-          const controller = ref.current?.abortController;
-          controller?.abort?.(context.reason);
-        });
+        // Abort all active queries using the cacheKey-based system
+        abortAllActiveQueries();
```
kavilla, line 26: does this lose the context ? if we do it this way?

src/plugins/explore/public/application/utils/state_management/actions/detect_optimal_tab/detect_optimal_tab.ts
```diff
@@ -0,0 +1,69 @@
...
+import { RootState } from '../../store';
+import { setActiveTab } from '../../slices';
+import { ExploreServices } from '../../../../../types';
+import { defaultPrepareQueryString } from '../query_actions';
+import { getVisualizationType } from '../../../../../components/visualizations/utils/use_visualization_types';
+
+/**
+ * Determine if results can be visualized
+ */
+const canResultsBeVisualized = (results: any): boolean => {
```
kavilla, line 16: what does detecting optimal tab have to be nested into state management?

src/plugins/explore/public/application/utils/state_management/actions/export_actions.ts
```diff
@@ -0,0 +1,151 @@
...
+import { Dispatch } from 'redux';
+import { saveAs } from 'file-saver';
+import { ExploreServices } from '../../../../types';
+import { AppDispatch, RootState } from '../store';
+
+/**
+ * Redux Thunk for exporting data to CSV
+ * Uses existing results from the Redux store
+ */
+export const exportToCsv = (options: { fileName?: string; services?: ExploreServices } = {}) => {
```
kavilla, line 15: what does this have to be nested into state management?

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -0,0 +1,539 @@
...
+  ProcessedSearchResults,
+} from '../../interfaces';
+import { defaultPreparePplQuery, getQueryWithSource } from '../../languages';
+
+// Module-level storage for abort controllers keyed by cacheKey
+const activeQueryAbortControllers = new Map<string, AbortController>();
+
+// Helper function to abort all active queries
+export const abortAllActiveQueries = () => {
+  activeQueryAbortControllers.forEach((controller, cacheKey) => {
```
kavilla, line 40: are we sure that if we unmount the plugin (ie navigate away) that this will abort all active queries?

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -0,0 +1,539 @@
...
+    controller.abort();
+  });
+  activeQueryAbortControllers.clear();
+};
+
+/**
+ * Default query preparation for tabs
+ */
+export const defaultPrepareQueryString = (query: Query): string => {
```
kavilla, line 49: this function seems really complex and not sure if it's helpful. right now it seems like the only purpose for this function is to strip stats from the query but then that gets obfuscated based on the name

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -0,0 +1,539 @@
...
+        `defaultPrepareQueryString encountered unhandled language: ${query.language}`
+      );
+  }
+};
+
+/**
+ * Default results processor for tabs
+ * Processes raw hits to calculate field counts and optionally includes histogram data
+ */
+export const defaultResultsProcessor: DefaultDataProcessor = (
```
kavilla, line 64: what does this have to do with state and is this not already available?

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -0,0 +1,539 @@
...
+    query.query !== '' &&
+    activeTabCacheKey !== visualizationTabCacheKey &&
+    activeTabCacheKey !== defaultCacheKey &&
+    !results[activeTabCacheKey];
+
+  const promises = [];
+
+  // Execute default query for histogram/sidebar
+  if (needsDefaultQuery) {
+    const interval = state.legacy?.interval;
```
kavilla, line 175: why is interval part of the legacy state?

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -0,0 +1,539 @@
...
+    // Create abort controller for this specific query
+    const abortController = new AbortController();
+
+    // Store controller by cacheKey for individual query abort
+    activeQueryAbortControllers.set(cacheKey, abortController);
+
+    // Reset inspector adapter
+    services.inspectorAdapters.requests.reset();
+
+    const title = i18n.translate('explore.discover.inspectorRequestDataTitle', {
```
kavilla, line 269: does this not cause i18n issues?

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -0,0 +1,539 @@
...
+      defaultMessage: 'data',
+    });
+    const description = i18n.translate('explore.discover.inspectorRequestDescription', {
+      defaultMessage: 'This request queries OpenSearch to fetch the data for the search.',
```
kavilla, line 273: i dont think this is the right wording

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -0,0 +1,539 @@
...
+
+    throw error;
+  }
+};
+
+/**
+ * Helper function to create SearchSource with common configuration
+ */
+const createSearchSourceWithQuery = async (
+  preparedQuery: any,
```
kavilla, line 404: any type

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -0,0 +1,539 @@
...
+      interval: undefined, // Tabs don't need intervals
+    },
+    thunkAPI
+  );
+});
+
+/**
+ * Helper function to transform aggregation results into chart data
+ */
+function transformAggregationToChartData(results: any, indexPattern: any): ChartData | undefined {
```
kavilla, line 500: this is new? we def have aggs data as charts why dont we leverage that? or use expressions

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -0,0 +1,539 @@
...
+  const maxTime = buckets.length > 0 ? moment(buckets[buckets.length - 1].key) : moment();
+
+  // Create chart data structure that matches Discover's Chart interface
+  return {
+    values: buckets.map((bucket: any) => ({
+      x: bucket.key,
+      y: bucket.doc_count,
+    })),
+    xAxisOrderedValues: buckets.map((bucket: any) => bucket.key),
+    xAxisFormat: { id: 'date', params: { pattern: 'YYYY-MM-DD HH:mm' } },
```
kavilla, line 527: the pattern should be pulled from advanced settings right?

src/plugins/explore/public/application/utils/state_management/actions/query_editor/load_query/load_query.ts
```diff
@@ -0,0 +1,23 @@
...
+import { useSetEditorTextWithQuery } from '../../../../../hooks';
+import { clearLastExecutedData } from '../../../slices';
+
+/**
+ * This is called when you need to load a query, it runs the loaded query
+ */
+export const loadQueryActionCreator = (
+  services: ExploreServices,
+  setEditorTextWithQuery: ReturnType<typeof useSetEditorTextWithQuery>,
+  query: string
```
kavilla, line 18: why string?

src/plugins/explore/public/application/utils/state_management/actions/query_editor/run_query/run_query.ts
```diff
@@ -0,0 +1,29 @@
...
+ * This is called when you want to run the query
+ */
+export const runQueryActionCreator = (services: ExploreServices, query?: string) => async (
+  dispatch: AppDispatch
+) => {
+  if (typeof query === 'string') {
+    dispatch(setQueryStringWithHistory(query));
+  }
+
+  dispatch(setActiveTab(''));
```
kavilla, line 23: so it seems like we do have a action to do all the disatching?

src/plugins/explore/public/application/utils/state_management/middleware/query_sync_middleware.ts
```diff
@@ -0,0 +1,51 @@
...
+ * user-initiated query executions.
+ */
+export const createQuerySyncMiddleware = (services: ExploreServices): Middleware<{}, RootState> => {
+  return (store) => (next) => (action) => {
+    const result = next(action);
+
+    if (
+      action.type === 'query/setQueryState' ||
+      action.type === 'query/setQueryWithHistory' ||
+      action.type === 'query/setQueryStringWithHistory'
```
kavilla, line 25: wouldn't it be easier just have one? always require a query object and always populate it to the history?

i dont see why we would have three different types for this. do these all do the same thing?

src/plugins/explore/public/application/utils/state_management/middleware/query_sync_middleware.ts
```diff
@@ -0,0 +1,51 @@
...
+        const queryStringQuery = services.data.query.queryString.getQuery();
+
+        if (!isEqual(queryStringQuery, queryWithSource)) {
+          services.data.query.queryString.setQuery(queryWithSource);
+        }
+
+        // Add to query history only for user-initiated query executions
+        // This prevents programmatic updates (loading saved queries, clearing, etc.) from polluting history
+        if (action.meta?.addToHistory && query.query?.trim()) {
+          const timefilter = services?.data?.query?.timefilter?.timefilter;
```
kavilla, line 41: on line 32 we access services.data.query without the optional chaining. why do we do it here?

src/plugins/explore/public/application/utils/state_management/middleware/query_sync_middleware.ts
```diff
@@ -0,0 +1,51 @@
...
+    ) {
+      const state = store.getState();
+      const query = state.query;
+      const queryWithSource = getQueryWithSource(query);
+
+      if (query.dataset && services.data?.query?.queryString) {
+        const queryStringQuery = services.data.query.queryString.getQuery();
+
+        if (!isEqual(queryStringQuery, queryWithSource)) {
+          services.data.query.queryString.setQuery(queryWithSource);
```
kavilla, line 35: should we just do this regardless? like why access the query object and do a deep compare?

src/plugins/explore/public/application/utils/state_management/selectors/index.ts
```diff
@@ -0,0 +1,90 @@
...
+const selectQueryState = (state: RootState) => state.query;
+export const selectUIState = (state: RootState) => state.ui;
+const selectResultsState = (state: RootState) => state.results;
+const selectLegacyState = (state: RootState) => state.legacy;
+export const selectTabState = (state: RootState) => state.tab;
+
+/**
+ * Query selectors
+ */
+export const selectQuery = createSelector([selectQueryState], (queryState) => queryState);
```
kavilla, line 21: this is where i get confused. why not just have one. `selectQuery` the query object should have all that we need

src/plugins/explore/public/application/utils/state_management/selectors/index.ts
```diff
@@ -0,0 +1,90 @@
...
+);
+
+export const selectDataset = createSelector([selectQueryState], (queryState) => queryState.dataset);
+
+/**
+ * UI selectors
+ */
+export const selectActiveTabId = createSelector([selectUIState], (uiState) => uiState.activeTabId);
+
+export const selectShowDatasetFields = createSelector(
```
kavilla, line 40: showFields

src/plugins/explore/public/application/utils/state_management/selectors/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  [selectTabState],
+  (tabState) => tabState.visualizations.chartType
+);
+
+export const selectAxesMapping = createSelector(
+  [selectTabState],
+  (tabState) => tabState.visualizations.axesMapping
+);
+
+export const selectActiveTab = createSelector(
```
kavilla, line 65: why not just have one selector?

src/plugins/explore/public/application/utils/state_management/selectors/index.ts
```diff
@@ -0,0 +1,90 @@
...
+export const selectColumns = createSelector(
+  [selectLegacyState],
+  (legacyState) => legacyState.columns
+);
+
+export const selectSort = createSelector([selectLegacyState], (legacyState) => legacyState.sort);
+
+export const selectSavedSearch = createSelector(
+  [selectLegacyState],
+  (legacyState) => legacyState.savedSearch
```
kavilla, line 87: we have this but not a saved explore?

src/plugins/explore/public/application/utils/state_management/slices/ui/ui_slice.ts
```diff
@@ -0,0 +1,41 @@
...
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+export interface UIState {
+  activeTabId: string;
+  showFilterPanel: boolean;
+  showHistogram: boolean;
+}
+
+const initialState: UIState = {
+  activeTabId: '',
```
kavilla, line 15: why would tab id be initially empty?

src/plugins/explore/public/application/utils/state_management/types.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+/**
+ * Application state without query state (used for URL persistence)
+ * Query state is handled separately in URL persistence
+ */
+export type AppState = Omit<RootState, 'query'>;
+
+export enum EditorMode {
+  Query = 'query',
```
kavilla, line 50: is there a difference? both are query. couldn't we treat it as language base. like NLQ vs PPL

src/plugins/explore/public/components/data_table/explore_data_table.tsx
```diff
@@ -3,151 +3,147 @@
...
+import {
+  defaultPrepareQueryString,
+  defaultResultsProcessor,
+} from '../../application/utils/state_management/actions/query_actions';
+import { useChangeQueryEditor } from '../../application/hooks';
+import { useDatasetContext } from '../../application/context';
+import { addColumn, removeColumn } from '../../application/utils/state_management/slices';
+
+const ExploreDataTableComponent = () => {
+  const { services } = useOpenSearchDashboards<ExploreServices>();
```
kavilla, line 37: i like how we use services with the context. we should do this through the app

src/plugins/explore/public/components/container/bottom_container/bottom_right_container/bottom_right_container.tsx
kavilla, line 1: i think bottom right container is a weird name like are we saying that people are embedding the bottom right container of explore into a dashboard? i think canvas or something was better.

src/plugins/explore/public/components/experience_banners/new_experience_banner/new_experience_banner.tsx
kavilla, line 1: why not banner service?

src/plugins/explore/public/components/fields_selector/discover_field.test.tsx
kavilla, line 1: should just be explore_feild?

src/plugins/explore/public/components/patterns_table/utils/utils.tsx
kavilla, line 1: i didn't know patterns was coming in on this release.

src/plugins/explore/public/application/utils/state_management/types.ts
angle943: I am distinguishing between a query language vs natural language, as the UI looks different. When we have SQL and etc in the future, that will also be considered a Query

src/plugins/explore/public/application/utils/state_management/slices/ui/ui_slice.ts
angle943: i had a long talk with anan about this, and she convinced me. An empty activeTabId means that the component (ie tab component i think) will programmatically determine the best tab based on the query. If the active tab is not empty, then that tab will be selected, even if it is not 'optimal' for the query.

This is to fulfill the use case where the user saves an stat query (should default to viz) in the logs page

angle943: @ananzh i still think a comment will be helpful for this

testing_strategy.md
kavilla, line 1: this file should be deleted

yarn.lock
```diff
@@ -1413,7 +1480,7 @@
     "@babel/helper-plugin-utils" "^7.25.9"
     "@babel/traverse" "^7.25.9"
 
-"@babel/plugin-proposal-class-properties@^7.12.1":
+"@babel/plugin-proposal-class-properties@^7.12.1", "@babel/plugin-proposal-class-properties@^7.18.0":
```
kavilla, line 1483: was this generated with the correct yarn version?

src/plugins/opensearch_dashboards_react/public/code_editor/editor_theme.ts
```diff
@@ -125,8 +125,63 @@ export function createTheme(
   };
 }
 
+export function createTheme(
+  euiTheme: typeof darkTheme | typeof lightTheme,
+  selectionBackgroundColor: string
+): monaco.editor.IStandaloneThemeData {
+  return {
+    base: euiTheme === lightTheme ? 'vs' : 'vs-dark', // Change the base theme based on theme selected
```
kavilla, line 133: is this comment needed?

src/plugins/data/public/query/query_string/language_service/types.ts
angle943: @joshuali925

src/plugins/data/public/query/query_string/language_service/types.ts
angle943: @joshuali925

src/plugins/data/public/query/saved_query/saved_query_service.ts
angle943: @ruanyl ? are you the right person to own this?

src/plugins/data/public/ui/saved_query_management/saved_query_management_component.tsx
angle943: this is a component that is used by the editor. we decided to disable the `save` when user is in prompt mode, but they can always load a saved query

src/plugins/explore/public/application/components/recent_query_selector.tsx
angle943: @ananzh ?

src/plugins/explore/public/application/hooks/editor_hooks/use_change_query_editor/index.ts
angle943: its more organizational, there are like 8 hooks in this folder, each with an unit test

src/plugins/explore/server/saved_objects/explore.ts
```diff
@@ -46,7 +46,8 @@ export const searchSavedObjectType: SavedObjectsType = {
     },
     getInAppUrl(obj) {
       return {
-        path: `/app/explore#/view/${encodeURIComponent(obj.id)}`,
+        // TODO:finalize this until flavor and view route are finalized
```
kavilla, line 49: is this still the case?

src/plugins/explore/public/application/hooks/editor_hooks/use_change_query_editor/use_change_query_editor.test.ts
angle943: this is (almost) globally accessible in the context of the explore app. If you look at the DOM tree both the redux and EditorContext are near the top

src/plugins/data/public/query/query_string/language_service/types.ts
joshuali925: Last time i checked the Query object doesn't contain prompt, prompt is not a query

src/plugins/explore/public/helpers/use_flavor_id/use_flavor_id.ts
```diff
@@ -0,0 +1,31 @@
...
+  const [flavorId, setFlavorId] = useState<ExploreFlavor | null>(null);
+  const { services } = useOpenSearchDashboards<ExploreServices>();
+
+  useEffect(() => {
+    const subscription = services.core.application.currentAppId$.subscribe((value) => {
+      const flavorFromAppId = value?.split('/')?.[1];
+      if (flavorFromAppId) {
+        setFlavorId(flavorFromAppId as ExploreFlavor);
+      } else {
+        setFlavorId(null);
```
kavilla, line 21: if flavorFromAppId could be undefined? then we could just set it directly if it exsit or not

src/plugins/explore/public/application/index.tsx
angle943: that should be fine. but i am personally not liking this whole logs page, metrics page, traces page where they are currently all the same thing. I want to see how they diverge when metrics and traces gets implemented, and reconsider whether we need to refactor it or not. but i got no issues with the index.ts file

src/plugins/explore/public/helpers/data_table_helper.tsx
```diff
@@ -82,11 +86,11 @@ export function getTimeColumn(
...
-  indexPattern: IndexPattern,
+  indexPattern: IndexPattern | Dataset,
   hideTimeField: boolean,
   isShortDots: boolean
 ): LegacyDisplayedColumn[] {
-  if (!Array.isArray(columns) || typeof indexPattern !== 'object' || !indexPattern.getFieldByName) {
+  if (!Array.isArray(columns) || !indexPattern || !indexPattern.getFieldByName) {
     return [];
   }
   // TODO: Remove overrides once we support PPL/SQL sorting
```
kavilla, line 96: im not sure if this got ported over but i think this might bite us in the future. ie what if the field is not searchable or aggregatable

src/plugins/data/public/query/query_string/language_service/types.ts
joshuali925: because this is the editor query `const editorQuery = useEditorQueryText();`, the editor query text is not a query object

src/plugins/explore/public/index.scss
```diff
@@ -2,4 +2,6 @@
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
  */
-/* stylelint-disable no-empty-source */
+
+// Import the new layout styles
```
kavilla, line 6: can be removed

src/plugins/explore/public/application/pages/logs/logs_page.tsx
angle943: "copied and pastaed from the other comment"

but i am personally not liking this whole logs page, metrics page, traces page where they are currently all the same thing. I want to see how they diverge when metrics and traces gets implemented

src/plugins/explore/public/saved_explore/transforms.ts
```diff
@@ -0,0 +1,124 @@
...
+  tabDefinition: TabDefinition,
+  tabState?: TabState,
+  dataset?: IndexPattern | Dataset
+): SavedExplore => {
+  // Serialize the state into the saved object
+  obj.type = flavorId;
+  obj.visualization = JSON.stringify({
+    // TODO: Add title to saved object
+    // Visualization has an independent title?
+    title: '',
```
kavilla, line 31: is this true?

src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts
angle943: I 100% agtrree. i made several action creators to fulfill these use case, i think we use leverage that pattern more

src/plugins/explore/public/application/utils/interfaces.ts
angle943: fyi we are planning on getting rid of the `application` folder entirely. but on top of that, i think redux should get its own folder near the root, inline with `components` and such.

src/plugins/explore/public/application/utils/languages/ppl/default_prepare_ppl_query/default_prepare_ppl_query.ts
angle943: this is very messy and its from previous naming. I also agree we should revise these. The comment makes it imply it is just for histogram, maybe the name should say that more explicitely. @ananzh ?

src/plugins/saved_objects/public/finder/saved_object_finder.tsx
```diff
@@ -167,7 +167,7 @@ class SavedObjectFinderUi extends React.Component<
     // If the current app id is explore, although explore app might not support
     // a language, it still supports all saved searches that are supported by
     // discover by redirecting to discover for backward compatibility.
-    if (currentAppId === 'explore') {
+    if (currentAppId.includes('explore/')) {
```
kavilla, line 170: i'm confused by this one. does discover get messed up?

src/plugins/explore/public/application/utils/languages/ppl/get_query_string_with_source/get_query_with_source.ts
angle943: what is wrong with this? it is using `let`.

I can technically do:

```
const queryStringWithSource = queryString.trim() === '' ? `source=${datasetTitle}` : `source=${datasetTitle} ${queryString}`
```

but i am personally not the biggest fan of ternary operators, due to readability

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
angle943: could you explain?

src/plugins/explore/public/application/utils/state_management/actions/query_editor/load_query/load_query.ts
angle943: why not? just for background this is used for loading a saved search and loading a saved query and loading recent query. The purpose is to set the editor text with the provided query string.

src/plugins/explore/public/application/utils/state_management/selectors/index.ts
angle943: i think its nice if a component only needs the query string, there can be a selector that only grabs the query string. Why?

- reduces knowledge that the component needs. if it doesn't need to know about dataset and language, it shouldn't need to. by using this `selectQuery` it is forced to know it, and we are accountable for it in unit tests
- performance: this selector will re-compute every time the queryState object changes. But what if just dataset changed? Then selectQueryString will not be recomputed

src/plugins/explore/public/application/utils/state_management/selectors/index.ts
angle943: is this comment about the name? i personally think just `fields` is too generic maybe?

src/plugins/explore/public/components/experience_banners/new_experience_banner/new_experience_banner.tsx
angle943: We could, initially this was here because it relied on the old query editor's `__enhance` functionality, but once we swapped the editor for the new one (that does not have the `__enhance` currently, i just swapped it out with this. Is this something you feel strongly to change?

src/plugins/saved_objects/public/finder/saved_object_finder.tsx
angle943: there was a bug where we introduced flavors, since the id is now `explore/logs`, `explore/traces`, etc. we can no longer to ===

angle943 (approved):

LDrago27 (approved):

src/plugins/query_enhancements/public/search/filters/filter_utils.ts
```diff
@@ -0,0 +1,108 @@
...
+    return `WHERE \`${timeFieldName}\` >= '${formatDate(
+      fromDate
+    )}' AND \`${timeFieldName}\` <= '${formatDate(toDate)}'`;
+  }
+
+  /**
+   * Convert core {@link Filter} and convert to a PPL where clause. Only
+   * supports non DSL filters.
+   */
+  public static convertFiltersToWhereClause(
```
kavilla, line 35: @LDrago27 is this something we can use some of the grammar rules with?

peterzhuamazon:
Hi. Convert to draft for the time being until https://github.com/opensearch-project/.github/issues/362 resolve tomorrow.

Thanks.

ashwin-pc:
Squashing and merging as we have decided not to push to main directly

joshuali925:
why not rebase or create merge commit? we lose all git history on squash

zhongnansu:
I still see unaddressed comments from @kavilla

ruanyl:
@joshuali925 we have pushed very hard to avoid squashing and keeping the commit history in the last a few days, you can find more info here: https://github.com/opensearch-project/.github/issues/362

src/plugins/data/public/query/saved_query/saved_query_service.ts
ruanyl: @angle943 I'm not sure, who originally added this?

joshuali925: @ruchidh https://github.com/opensearch-project/OpenSearch-Dashboards/commit/2099c78a33c58a9ed29052ff5c070d5b2bc47ea7#diff-8f9e6729b2a73ad1a432b62f4884f1b2c9131fbf0ce940093d999753ab0dea0dR41-R42?

PR #10248, opened 2025-07-21
Fix a issue that cause PPL histogram display incorrectly

Maosaic:
### Description

Previously, when converting date to timestamp, date that does not contain timezone information are being converted as local timezone. However, PPL data format does contain timezone information but require be converted to UTC timezone. Not correctly handling the timestamp conversion is causing PPL histogram display incorrectly.

Change the timestamp conversion util to treat date as UTC timezone if the data format does not contain timezone information. 

### Issues Resolved

Fix a issue that cause PPL histogram display incorrectly

## Changelog

- fix: Fix a issue that cause PPL histogram display incorrectly

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/common/data_frames/utils.test.ts
+++ b/src/plugins/data/common/data_frames/utils.test.ts
@@ -245,6 +245,55 @@ describe('convertResult', () => {
     expect(result.aggregations?.timestamp_histogram.buckets[0].doc_count).toBe(10);
   });
 
+  it('should convert date strings to UTC timestamps when no timezone info is present', () => {
+    const response: IDataFrameResponse = {
+      took: 100,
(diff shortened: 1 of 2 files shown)


TackAdam (approved):
LGTM

abbyhu2000 (approved):

abbyhu2000:
`convertResult` also is used by legacy discover, could u check and make sure PPL on legacy discover still work correctly with timezone?

Maosaic:
> `convertResult` also is used by legacy discover, could u check and make sure PPL on legacy discover still work correctly with timezone?

legacy discover also have the same issue

Issue #9898, opened 2025-06-13
[RFC] Introduce React Flow Library in OpenSearch Dashboards


ps48:
## Overview 
Introduce react flow library in OpenSearch Dashboards https://reactflow.dev/ 

## Motivation 
Currently OpenSearch Dashboards contains support for many visualization specific libraries and framework which include vega, elastic/charts and d3. There are really great and flexible frameworks for generic visualizations. But there is a need for a framework where we need Higher order components for interactive node-based visualizations that require:

- **Interactive Node Graphs**: Service maps, trace flows, and dependency visualizations
- **Workflow Visualizations**: Process flows, pipeline diagrams, and state machines  
- **Network Diagrams**: Infrastructure topology, data lineage, and relationship mapping
- **Decision Trees**: Logic flows and conditional pathways
- **Observability Workflows**: Investigative flows for troubleshooting, alert correlation, and root cause analysis

Current implementations across plugins are fragmented, leading to:
- Inconsistent user experiences
- Duplicated development effort
- Version conflicts between plugins
- Increased security maintenance overhead

## Benefits of React flow

1. **Developer Experience**
   - Declarative React API that integrates seamlessly with OSD's React ecosystem
   - TypeScript support with comprehensive type definitions
   - Extensive documentation and active community

2. **Performance & Scalability**
   - Built-in virtualization handles thousands of nodes efficiently
   - Optimized rendering with React's reconciliation
   - Memory-efficient edge rendering

3. **Customization & Flexibility**
   - Custom node and edge components
   - Pluggable layout algorithms
   - Theming support compatible with OSD's EUI theme system

4. **Accessibility**
   - WCAG compliant with keyboard navigation
   - Screen reader support
   - Focus management

5. **Maintenance & Security**
   - Active development with regular updates
   - Strong security track record
   - MIT license compatible with OpenSearch

## How is it different from above mentioned libraries

React Flow is specifically designed for interactive node-based visualizations and provides higher-order components that are not available in traditional charting libraries:

**Core Components & HOCs:**
- **ReactFlow Component**: Main container that handles the flow state, interactions, and rendering
- **Node Components**: Pre-built and customizable node types (default, input, output, group nodes)
- **Edge Components**: Various edge types (default, straight, step, smoothstep, bezier) with custom styling
- **Handle Components**: Connection points for nodes that manage edge attachment and validation
- **Background Component**: Configurable grid, dots, or lines background patterns
[...]

PR #10253, opened 2025-07-21
Trace Details page

TackAdam:
### Description
Discover: Trace Details page
Screenshot 2025-07-30 at 3 17 12 PM

The trace detail page will be navigated to using links that re-direct to the url with the following example format:
```
http://localhost:5601/lvy/w/9C3CHg/app/explore/traces/traceDetails#/?_a=(dataSourceId:e2681970-5b5b-11f0-84cd-1f27cd4c7d84,indexPattern:'otel-v1-apm-span-*',spanId:a8b6dd34808083b3,traceId:'23a88fa559da1b6c7521658fba2cb82f')
```

If a spanId is not provided such as the user entered the page with just the TraceId the span id will be selected as the root span or first available span.

### Issues Resolved

Introduce library React flow and dagrejs RFC https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9898 for constructing service map:
ServiceMap

## Screenshot
Video of all functionality:

https://github.com/user-attachments/assets/03d87849-4860-4513-894f-32d7a3461070

## Testing the changes

Included storybook can be checked with
``` yarn storybook explore ```
Storybook

Included testing can be checked with
``` yarn test:jest src/plugins/explore/public/application/pages/traces/trace_details --silent ```
Screenshot 2025-07-30 at 3 16 20 PM

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### CODE REVIEW

File: src/plugins/explore/public/application/pages/traces/trace_details/public/services/service_map.tsx [hunk 1]
```diff
@@ -0,0 +1,1209 @@
...
+              minWidth: '180px',
+              textAlign: 'left',
+              fontSize: '14px',
+              fontWeight: '500',
+              position: 'relative',
+              cursor: 'pointer',
+              opacity: isSelected ? 1 : 1,
+              transform: isSelected ? 'scale(1.02)' : 'scale(1)',
+              transition: 'all 0.2s ease-in-out',
+            }}
```
Maosaic: Not dynamic styles should be imported from `scss`

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/ppl_to_trace_hits.tsx
```diff
@@ -0,0 +1,298 @@
...
+  endTime: string;
+  durationInNanos: number;
+  'status.code': number;
+  // Dynamic attributes will be added here
+  [key: string]: any;
+  // Sort field for ordering
+  sort: number[];
+}
+
+export interface PPLResponse {
```
ananzh, line 32: Is this PPLResponse traces specific? Should we consider put these to data plugin?

TackAdam: This PPL Response is trace specific, it formats the PPL response to the required format for cleanly displaying the raw data and use in the gantt chart. I think keeping it here makes sense until we have a use-case that would require it outside of this page.

Transform

src/plugins/explore/public/application/pages/traces/trace_details/state/trace_app_state.ts
```diff
@@ -0,0 +1,82 @@
...
+  stateDefaults: TraceAppState;
+  osdUrlStateStorage: IOsdUrlStateStorage;
+}) => {
+  const initialStateFromUrl = osdUrlStateStorage.get<TraceAppState>(STATE_STORAGE_KEY);
+  const initialState: TraceAppState = {
+    ...stateDefaults,
+    ...initialStateFromUrl,
+  };
+
+  const stateContainer = createStateContainer(initialState, {
```
ananzh, line 34: This file implements a separate state management system from the main explore plugin's Redux store. Have we discussed somewhere about the state management for different pages? Currently explore uses Redux + persistence middleware for URL sync. I am curious why we create two different state management patterns within the same plugin. Should we consider integrating with existing Redux store?

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
```diff
@@ -0,0 +1,305 @@
...
+        dataSourceId: '',
+        indexPattern: 'otel-v1-apm-span-*',
+        spanId: undefined,
+      },
+      osdUrlStateStorage: osdUrlStateStorage!,
+    });
+  }, [osdUrlStateStorage]);
+
+  // Get current state values and subscribe to changes
+  const [appState, setAppState] = useState(() => stateContainer.get());
```
ananzh, line 72: This component uses local state + URL state container instead of the Redux pattern used elsewhere in explore. Seems have limited integration with existing explore services (timefilter, query string manager, etc.)

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
```diff
@@ -0,0 +1,305 @@
...
+    services: { chrome, data, osdUrlStateStorage },
+  } = useOpenSearchDashboards<DataExplorerServices>();
+
+  // Initialize URL state management
+  const { stateContainer, stopStateSync } = useMemo(() => {
+    return createTraceAppState({
+      stateDefaults: {
+        traceId: '',
+        dataSourceId: '',
+        indexPattern: 'otel-v1-apm-span-*',
```
ananzh, line 64: Do we have to hardcode for indexPattern? Also can we use dataset as in explore as indexPattern has been migrated?

src/plugins/explore/public/application/pages/traces/trace_details/public/services/service_map.tsx
```diff
@@ -0,0 +1,1059 @@
...
+  onFocusService: (serviceName: string) => void;
+  onToggleDetails: () => void;
+  openPopoverNodeId: string | null;
+  setOpenPopoverNodeId: (id: string | null) => void;
+  isSelected?: boolean;
+}) => {
+  const serviceColor = data.color;
+
+  // Calculate intensity values based on max values
+  const requestRateIntensity = data.spanCount / (data.maxValues?.maxRequestRate || 100);
```
ananzh, line 111: Curious can the data.maxValues?.maxRequestRate ever be 0 or undefined? and how do we come up 100, 0.1 and 1000000000 these numbers?

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
```diff
@@ -0,0 +1,305 @@
...
+      fetchData(spanFilters);
+    }
+  }, [traceId, dataSourceId, pplService, spanFilters, indexPattern, transformedHits.length]);
+
+  useEffect(() => {
+    if (!pplQueryData) return;
+    // Transform the PPL data to trace hits format
+    const transformed = transformPPLDataToTraceHits(pplQueryData);
+    const hits = transformed.length > 0 ? transformed : [];
+    setTransformedHits(hits);
```
ananzh, line 173: curious why we need to save both transformedHits and unfilteredHits? Also in main explore, we have results slice to save results for diff tabs but might be not useful here. 

I am thinking if we could reset the slices when navigate different pages and reuse the slices would be great. Some states are sharable. For example index pattern can use query slice. A migration function can allow user to navigate different pages without losing dataset context.

src/plugins/explore/public/application/pages/traces/trace_details/state/trace_app_state.ts
TackAdam: The trace details page uses a separate state management system because it serves a fundamentally different purpose than the main explore interface. The main Redux store manages complex query building, visualization configurations, and editor modes - none of which are relevant to trace navigation. 

The trace details page only needs to manage trace-specific navigation state (traceId, spanId, dataSourceId, indexPattern). Using a lightweight state container provides better performance and cleaner URLs while maintaining the same URL synchronization pattern.

This architectural separation follows the principle of using the right tool for the job - complex Redux for complex query workflows, simple state containers for simple navigation state.

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
TackAdam: The time-filter and query string are not used in the trace detail page. We remove that context when re-directing and just focus on the spans present in the parent trace.

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
TackAdam: The re-direction should provide the index pattern in the url, this is just a fallback to the default naming convention if it is not provided.

I haven't seen the query example version using the dataset instead of index pattern. As soon as I see a ppl example of what the call looks like that should be simple to switch out.

src/plugins/explore/public/application/pages/traces/trace_details/public/services/service_map.tsx
TackAdam: It should never be undefined or 0 as the service map is only generated if we have valid spans making up the trace. This calculation is for making the colored bars of the service map to give a representation of it's weight in the trace.

These fallbacks are just reasonable numbers for the fields to ensure that the component does not crash as filter are applied and it is re-rendered.

Screenshot 2025-07-28 at 10 45 21 AM


Screenshot 2025-07-28 at 10 44 45 AM

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
TackAdam: We maintain both transformedHits and unfilteredHits to ensure visual consistency across the Gantt chart and service map visualizations. The color mapping is generated from the complete unfiltered dataset, which prevents color assignments from shifting as users apply or remove span filters. This approach delivers a stable and intuitive user experience where services retain their visual identity throughout the filtering workflow.

I don't think the  Redux slice integration is needed here because the trace details page serves as an independent analysis view. Users typically arrive at this page from Discover:Traces when they need to perform deep-dive analysis on a specific trace, often sharing the URL with colleagues for collaborative investigation. The page is intentionally designed to be self-contained, so context loss upon navigation is not a concern.

src/plugins/explore/public/application/pages/traces/trace_details/public/gantt_chart_vega/gantt_chart_vega.scss
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.gantt-chart {
```
angle943: i fear this is too generic of a name. we usually namespace our classnames to avoid collisions. can we do something like `exploreGanttChart` or something? Thats been our convention (to add `explore` prefix to all of our classNames)

src/plugins/explore/public/application/pages/traces/trace_details/public/services/service_map.scss
```diff
@@ -0,0 +1,198 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.service-map {
```
angle943: namespace classname

src/plugins/explore/public/application/pages/traces/trace_details/public/services/service_map.scss
```diff
@@ -0,0 +1,198 @@
...
+    color: $euiColorDarkShade;
+    margin-bottom: 4px;
+
+    &--error {
+      color: $euiColorDanger;
+    }
+  }
+}
+
+.metric-bar {
```
angle943: all these should be nasmespaced

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/flyout_list_item.scss
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.flyout-list-item {
```
angle943: namespace

src/plugins/explore/public/application/pages/traces/trace_details/public/gantt_chart_vega/gantt_chart_vega.tsx
```diff
@@ -0,0 +1,154 @@
...
+export function GanttChart({
+  data,
+  colorMap,
+  height,
+  onSpanClick,
+  selectedSpanId,
+}: GanttChartProps) {
+  const containerRef = useRef<HTMLDivElement>(null);
+  const viewRef = useRef<View | null>(null);
+  const [containerWidth, setContainerWidth] = useState<number>(800);
```
angle943, line 33: you know better than me, but is this state unavoidable? vega can't automatically deduce the width of the container?

src/plugins/explore/public/application/pages/traces/trace_details/state/trace_app_state.ts
angle943: could you give an example of an URL for the detail page? And a example on a non-detail trace page? I'm going back and forth in writing a comment here, but I think examples will help

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
```diff
@@ -0,0 +1,305 @@
...
+      subscription.unsubscribe();
+    };
+  }, [stateContainer]);
+
+  const [transformedHits, setTransformedHits] = useState<any[]>([]);
+  const [spanFilters, setSpanFilters] = useState<any[]>([]);
+  const [pplQueryData, setPplQueryData] = useState<any>(null);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [isBackgroundLoading, setIsBackgroundLoading] = useState<boolean>(false);
+  const [unfilteredHits, setUnfilteredHits] = useState<any[]>([]);
```
angle943: lots of `any` here, are these unavoidable?

src/plugins/explore/public/application/pages/traces/trace_details/public/gantt_chart_vega/gantt_chart_vega.scss
TackAdam: Updated all namespace to use the correct naming convention
```.exploreGanttChart ```

File: src/plugins/explore/public/application/pages/traces/trace_details/public/services/service_map.scss (hunk 10 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/public/services/service_map.scss]: ```.exploreMetricBar ```

File: src/plugins/explore/public/application/pages/traces/trace_details/public/services/service_map.scss (hunk 9 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/public/services/service_map.scss]: ```.exploreServiceMap ```

File: src/plugins/explore/public/application/pages/traces/trace_details/public/traces/flyout_list_item.scss (hunk 11 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/public/traces/flyout_list_item.scss]: ```.exploreFlyoutListItem```

File: src/plugins/explore/public/application/pages/traces/trace_details/public/gantt_chart_vega/gantt_chart_vega.tsx (hunk 12 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/public/gantt_chart_vega/gantt_chart_vega.tsx:33]: Vega's built in spacing does not interface well with the gantt chart bars being dependent on width.
Will look into possible refactor as follow-up but haven't been able to get the built in resizing to work properly.

**Review by angle943 — DISMISSED** [review]

**Review by angle943 — DISMISSED** [review]

File: src/plugins/explore/public/application/pages/traces/trace_details/state/trace_app_state.ts (hunk 3 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/state/trace_app_state.ts:34]: Discover:Trace:
http://localhost:5601/zky/w/ah0CZq/app/explore/traces/#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))


Trace Details:
http://localhost:5601/lvy/w/9C3CHg/app/explore/traces/traceDetails#/?_a=(dataSourceId:e2681970-5b5b-11f0-84cd-1f27cd4c7d84,indexPattern:'otel-v1-apm-span-*',spanId:a8b6dd34808083b3,traceId:'23a88fa559da1b6c7521658fba2cb82f')

File: src/plugins/explore/public/application/pages/traces/trace_details/public/gantt_chart_vega/gantt_chart_spec.ts [hunk 14]
```diff
@@ -0,0 +1,283 @@
...
+        from: { data: 'spans' },
+        encode: {
+          enter: {
+            y: { scale: 'yscale', field: 'label' },
+            height: { scale: 'yscale', band: 1.0 }, // 1.0 for text to be in middle
+            fill: { field: 'color' },
+            cursor: { value: 'pointer' },
+            tooltip: {
+              signal:
+                "datum.hasError ? {'⚠️ Error': '" +
```
ps48: Does this `⚠️ Error` need localization?

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
ps48: @TackAdam let us know if you want to take this now or as fast follow when you add support for contextual links

src/plugins/explore/public/application/pages/traces/trace_details/public/gantt_chart_vega/gantt_chart_spec.ts
TackAdam: Good catch missed that one, added
```
                i18n.translate('explore.ganttChart.tooltip.error', {
                  defaultMessage: 'Error',
                }) +
```

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
TackAdam: Removed all any's

ps48 (approved):

angle943 (approved):

PR #10262, opened 2025-07-22
[Query Enhancements] improve resources API type definition

joshuali925:
### Description

This is a follow up improvement for https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9770. It adds ResourcesRequest API type and includes `errorObj.message` in the response.

## Testing the changes

mostly type changes. updated unit test

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/server/routes/resources/routes.ts
+++ b/src/plugins/query_enhancements/server/routes/resources/routes.ts
@@ -3,11 +3,31 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { schema } from '@osd/config-schema';
-import { IRouter } from 'opensearch-dashboards/server';
+import { schema, TypeOf } from '@osd/config-schema';
+import { IRouter, OpenSearchDashboardsRequest } from 'opensearch-dashboards/server';
 import { BASE_API } from '../../../common';
 import { resourceManagerService } from '../../connections/resource_manager_service';
 import { coerceStatusCode } from '..';
+
+const body = schema.object({
+  connection: schema.object({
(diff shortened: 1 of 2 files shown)


src/plugins/query_enhancements/server/routes/resources/routes.ts
```diff
@@ -46,7 +56,7 @@ export function registerResourceRoutes(router: IRouter) {
       } catch (error) {
         const errorObj = error as any;
         return response.customError({
-          body: 'Unable to get resources',
+          body: 'Unable to get resources: ' + errorObj.message,
```
joshuali925: FYI @kathyguo18 , is it ok to expose error message to the user? I think backend error message is already user facing

kathyguo18: "Unable to get resources" is the default error message. It can be overwritten by errorObj.body. 
The actual implementation of the manager should be responsible for the speific error handling. They should verify the error and add the error message to the body field. They can use backend error message in body field. 

I'm okay with updating the default error message to include errorObj.message, but we should add `errorObj.message ?? ''` to avoid "undefined in the error message.

joshuali925: Good catch, updated

Yeah I think the overwriting makes sense, either way the manager can decide what user sees. it's just that `message` is a standard field in `Error` objects but `body` is not, so it may be more convenient

sejli (approved):

Maosaic (approved):

PR #10264, opened 2025-07-22
[Banner Plugin] Integrate UI Settings Support for Configurable Global Banner

sidwang42:
### Description

This PR introduces support for managing banner settings via OpenSearch Dashboards’ advanced UI settings. It includes server-side registration of default values, server side read the settings, client-side subscription to setting changes, and unit tests for the banner render logic.

### Issues Resolved
Meta: #9990
RFC: #9861

## Screenshot

iShot_2025-07-22_15 36 16

## Testing the changes

1. Navigate to **Advanced Settings**
2. Search for `banner:` to find newly registered UI settings
3. Change any setting (e.g., `banner:content`) and observe real-time update to the global banner
4. Verify dismissal behavior, markdown rendering, and height adjustment
5. Run unit tests:

   ```bash
   yarn test:jest
   yarn test:jest_integration
   ```

## Changelog

- feat: Add global banner support via UI settings with live updates

### Check List

- [x] All tests pass

  - [x] `yarn test:jest`
- [x] Update [[CHANGELOG.md](https://chatgpt.com/CHANGELOG.md)](./../CHANGELOG.md)


src/plugins/banner/server/routes/get_config.test.ts
```diff
@@ -38,7 +38,39 @@ describe('Banner routes', () => {
...
+              }),
+            },
+          },
+        },
+        dataSource: {
+          opensearch: {
+            client: {},
+          },
+        },
+      } as any; // Cast to any to bypass TypeScript checks for the test
```
huyaboo: Could we avoid typing `any` here? OSD Core provides a `coreMock.createRequestHandlerContext();`. You can use the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals) to merge the default mock with your implementation.

src/plugins/banner/public/components/global_banner.tsx
```diff
@@ -9,47 +9,65 @@
...
+export const GlobalBanner: React.FC<GlobalBannerProps> = ({ http, uiSettings }) => {
   const [bannerConfig, setBannerConfig] = useState<BannerConfig | null>(null);
   const [isLoading, setIsLoading] = useState<boolean>(true);
   const bannerRef = useRef<HTMLDivElement>(null);
 
   // Fetch banner config from API when component mounts
+  const fetchBannerConfig = useCallback(async () => {
+    try {
+      setIsLoading(true);
+      const response = await http.get<BannerConfig>('/api/_plugins/_banner/content');
```
raintygao, line 33: Could we maintain this path as a constant in some place?

src/plugins/banner/public/components/global_banner.tsx
```diff
@@ -9,47 +9,65 @@
...
+      act(() => {
+        setBannerConfig(response);
+      });
```
raintygao: I guess this is some way for avoiding warnings when running tests? But should we include act in component file? I have concern about this.

src/plugins/banner/public/render_banner.ts
```diff
@@ -13,23 +13,24 @@ import React from 'react';
...
   if (container) {
-    ReactDOM.render(React.createElement(GlobalBanner, { http }), container);
+    ReactDOM.render(React.createElement(GlobalBanner, { http, uiSettings }), container);
 
     // Trigger resize and reflow for proper height calculation
     window.dispatchEvent(new Event('resize'));
     void document.body.offsetHeight;
   } else {
-    setTimeout(() => renderBanner(http), 50);
+    setTimeout(() => renderBanner(http, uiSettings), 50);
```
raintygao: BTW, could we leave a comment about why we use `setTimeout` here?

src/plugins/banner/server/routes/get_config.ts
```diff
@@ -19,7 +20,26 @@ export function defineRoutes(router: IRouter, bannerSetup: BannerPluginSetup) {
...
+      // Get UI settings client
+      const uiSettingsClient = context.core.uiSettings.client;
+
+      // Try to get values from UI settings, fall back to YAML config values
+      const content = await uiSettingsClient.get('banner:content');
+      const color = await uiSettingsClient.get('banner:color');
+      const iconType = await uiSettingsClient.get('banner:iconType');
+      const isVisible = await uiSettingsClient.get('banner:active');
+      const useMarkdown = await uiSettingsClient.get('banner:useMarkdown');
+      const size = await uiSettingsClient.get('banner:size');
```
raintygao: I haven't deep dive the implementation detail of uiSettings, but I suppose here using `uiSettingsClient.getAll` will have a better performance than a batch of `uiSettingsClient.get`

src/plugins/banner/public/render_banner.ts
FriedhelmWS: Would this be a good use case for `requestAnimationFrame`?

src/plugins/banner/public/components/global_banner.tsx
Maosaic: +1 
act is a testing utility only, It's designed specifically for test environments and not available in production

src/plugins/banner/public/render_banner.ts
```diff
@@ -13,23 +13,24 @@ import React from 'react';
...
+export const renderBanner = (http: HttpStart, uiSettings: IUiSettingsClient): void => {
   const container = document.getElementById(BANNER_CONTAINER_ID);
 
   if (container) {
-    ReactDOM.render(React.createElement(GlobalBanner, { http }), container);
+    ReactDOM.render(React.createElement(GlobalBanner, { http, uiSettings }), container);
 
     // Trigger resize and reflow for proper height calculation
     window.dispatchEvent(new Event('resize'));
     void document.body.offsetHeight;
```
Maosaic: Why do you need to manually trigger `resize`

src/plugins/banner/server/plugin.ts
```diff
@@ -16,26 +16,51 @@ import { BannerPluginConfigType } from './config';
...
-      bannerEnabled: () => pluginConfig.enabled,
+      bannerEnabled: () => this.pluginConfig.enabled,
       getConfig: (): BannerConfig => ({
-        content: pluginConfig.content,
-        color: pluginConfig.color,
-        iconType: pluginConfig.iconType,
-        isVisible: pluginConfig.isVisible,
-        useMarkdown: pluginConfig.useMarkdown,
-        size: pluginConfig.size,
+        content: this.pluginConfig.content,
```
huyaboo: Consider using the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals)

src/plugins/banner/public/components/global_banner.tsx
sidwang42: > Could we maintain this path as a constant in some place?

@raintygao 
Thanks for the suggestion! I did consider that, but after checking the repo, I followed the existing pattern—other places also use this route directly. Happy to refactor later if we decide to centralize it.

src/plugins/banner/public/components/global_banner.tsx
sidwang42: > +1 act is a testing utility only, It's designed specifically for test environments and not available in production

> I guess this is some way for avoiding warnings when running tests? But should we include act in component file? I have concern about this.

Yeah, I added it to suppress test warnings — definitely shouldn’t be in the component file. Thanks for the reminder, I’ll remove it. Appreciate it!

src/plugins/banner/public/components/global_banner.tsx
```diff
@@ -9,47 +9,65 @@
...
-          setIsLoading(false);
-        });
+    fetchBannerConfig();
+  }, [http, fetchBannerConfig]);
+
+  // Subscribe to UI settings changes for banner-related settings
+  useEffect(() => {
+    if (!uiSettings) return;
+
+    const subscription = uiSettings.getUpdate$().subscribe(({ key }) => {
```
zhongnansu: The single source of truth of the banner config that used to render banner should be coming from the API, here you are introducing another source. Can we stick to using API as single source?

src/plugins/banner/public/render_banner.ts
sidwang42: > Would this be a good use case for `requestAnimationFrame`?

Thanks a lot for pointing that out — I wasn’t aware of requestAnimationFrame before. Just checked the documentation and yeah, this looks like a perfect fit for our use case. Really appreciate the suggestion!

src/plugins/banner/server/routes/get_config.ts
sidwang42: > I haven't deep dive the implementation detail of uiSettings, but I suppose here using `uiSettingsClient.getAll` will have a better performance than a batch of `uiSettingsClient.get`

Noted. Will update.

src/plugins/banner/server/routes/get_config.test.ts
sidwang42: > Could we avoid typing `any` here? OSD Core provides a `coreMock.createRequestHandlerContext();`. You can use the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals) to merge the default mock with your implementation.

Sure. Thanks for pointing out.

src/plugins/banner/public/components/global_banner.tsx
sidwang42: > The single source of truth of the banner config that used to render banner should be coming from the API, here you are introducing another source. Can we stick to using API as single source?

This was added purely to trigger updates—API remains the single source of truth for the banner config.

src/plugins/banner/public/render_banner.ts
sidwang42: Deleted. Was here because the old design.

src/plugins/banner/public/components/global_banner.tsx
sidwang42: > > The single source of truth of the banner config that used to render banner should be coming from the API, here you are introducing another source. Can we stick to using API as single source?
> 
> This was added purely to trigger updates—API remains the single source of truth for the banner config.

Refactored using needs update for advanced settings, thanks for pointing out.

src/plugins/banner/server/plugin.ts
```diff
@@ -16,26 +16,46 @@ import { BannerPluginConfigType } from './config';
...
-    const pluginConfig: BannerPluginConfigType = await this.config$.pipe(first()).toPromise();
+    this.pluginConfig = await this.config$.pipe(first()).toPromise();
+
+    // Get the default UI settings
+    const bannerSettings = getBannerSettings();
+
+    // Override the default values with values from YAML config
+    if (bannerSettings['banner:content']) {
+      bannerSettings['banner:content'].value = this.pluginConfig.content;
+    }
```
zhongnansu, line 38: `bannerSettings['banner:content']?.value = this.pluginConfig.content;`
You can simplify the code using optional chaining (?.)

src/plugins/banner/server/ui_settings.ts
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { schema } from '@osd/config-schema';
+import { i18n } from '@osd/i18n';
+import { UiSettingsParams } from 'opensearch-dashboards/server';
+
+export const getBannerSettings = (): Record<string, UiSettingsParams> => {
```
huyaboo:
```suggestion
export const getDefaultBannerSettings = (): Record<string, UiSettingsParams> => {
```

nit: This seems to get the default settings, if I'm not mistaken. In that case, could we rename?

src/plugins/banner/server/routes/get_config.ts
```diff
@@ -19,7 +20,29 @@ export function defineRoutes(router: IRouter, bannerSetup: BannerPluginSetup) {
       validate: false,
     },
     async (context, request, response) => {
-      const config = bannerSetup.getConfig();
+      // Get UI settings client
+      const uiSettingsClient = context.core.uiSettings.client;
+
+      // Get all banner settings at once instead of individual calls
```
huyaboo:
```suggestion

```

huyaboo (dismissed):
Minor nits; LGTM

zhongnansu (commented):
minor comment, others lgtm

src/plugins/banner/server/plugin.ts
sidwang42: > `bannerSettings['banner:content']?.value = this.pluginConfig.content;` You can simplify the code using optional chaining (?.)

Thanks for the suggestion! 

I tried using optional chaining here, but TypeScript throws
`The left-hand side of an assignment expression may not be an optional property access.ts(2779)`
Based on this [blog](https://udn.realityripple.com/docs/Web/JavaScript/Reference/Operators/Optional_chaining) it seems that `?.` is only valid for reads, not assignments.

zhongnansu (approved):

sidwang42:
Simply rename `getBannerSettings` to `getDefaultBannerSettings`, the changes dismissed the review from earlier.

sidwang42:
Can confirm the two CIs failing are not related to this PR.
image

zhongnansu (approved):

huyaboo (approved):

PR #10265, opened 2025-07-23
[Explore] Move datepicker and run button to global header

ananzh:
### Description

This PR removes runButton from query editor footer and adds a customized button with the following functionality:

* Smart button states: Shows "Update" when query editor or datepicker changes are detected, shows "Refresh" when no changes are present
* Tab-aware behavior: Refresh respects the selected tab, while Update includes tab detection
* Input validation: Button is disabled for invalid time ranges and blocks Enter key execution when time range is invalid

### Issues Resolved
NA

## Screenshot

https://github.com/user-attachments/assets/4e3ca4e6-5863-4a8f-8377-7ba8ca8ec2e9

## Changelog
- feat: [Explore] Move datepicker and run button to global header

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/top_nav/components/smart_update_button.tsx
```diff
@@ -0,0 +1,78 @@
...
+  services: ExploreServices;
+  localDateRange?: { from: string; to: string } | null;
+}
+
+/**
+ * Smart button that shows different text based on what has changed:
+ * - "Run" when there are actual changes (query OR timefilter changed since last execution)
+ * - "Refresh" when there are no changes (same query and time, just re-run)
+ */
+export const SmartUpdateButton: React.FC<SmartUpdateButtonProps> = ({
```
kavilla: nit: smart update button doesn't fill like the correct name. since it doesn't give the insight of what it does. how do we differ this button than another button that updates

Maosaic: +1  too smart to be a button's name

src/plugins/explore/public/application/utils/state_management/actions/query_editor/run_query/run_query.ts
```diff
@@ -13,17 +13,22 @@ import { detectAndSetOptimalTab } from '../../detect_optimal_tab';
 /**
  * This is called when you want to run the query
  */
-export const runQueryActionCreator = (services: ExploreServices, query?: string) => async (
-  dispatch: AppDispatch
-) => {
+export const runQueryActionCreator = (
+  services: ExploreServices,
+  query?: string,
```
angle943: can we change this to an object prop instead of 3 arguments?

- its hard to read (what is false in `runQueryActionCreator(services, query, false)`?)
- what if query is undefined but `isUpdate` is true? Then you need to do `runQueryActionCreator(services, undefined, false)`

src/plugins/explore/public/application/utils/state_management/slices/query_editor/query_editor_slice.ts
```diff
@@ -132,6 +137,7 @@ export const {
   setPromptModeIsAvailable,
   setPromptToQueryIsLoading,
   clearLastExecutedData,
+  setQueryExecutionDisabled,
```
angle943: can we get unit test for this reducer?

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.test.tsx
```diff
@@ -88,20 +72,18 @@ describe('QueryPanelWidgets', () => {
...
-    expect(screen.getByTestId('selected-language')).toBeInTheDocument();
-    expect(screen.getByTestId('language-toggle')).toBeInTheDocument();
     expect(screen.getByTestId('query-panel-error')).toBeInTheDocument();
 
-    // Check right section components
-    expect(screen.getByTestId('date-time-range-picker')).toBeInTheDocument();
-    expect(screen.getByTestId('run-query-button')).toBeInTheDocument();
+    // Check that vertical separators are present
+    const separators = container.querySelectorAll('.exploreQueryPanelWidgets__verticalSeparator');
+    expect(separators).toHaveLength(3); // There should be 3 separators between the 4 components
```
angle943: do we really need to check if there are vertical separators -___-

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.tsx
```diff
@@ -32,14 +24,6 @@ export const QueryPanelWidgets = () => {
         {/* TODO: Actions should go here */}
         <QueryPanelError />
       </div>
-
-      {/* Right Section */}
-      <div className="exploreQueryPanelWidgets__right">
-        <SelectedLanguage />
-        <LanguageToggle />
```
angle943, line 36: why are we removing `SelectedLanguagte` an `LanguageToggle`?

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.test.tsx
```diff
@@ -134,18 +124,26 @@ describe('QueryPanelWidgets', () => {
...
-    expect(screen.queryByTestId('date-time-range-picker')).not.toBeInTheDocument();
+    // The component renders the same components regardless of dataset
+    expect(screen.getByTestId('filter-panel-toggle')).toBeInTheDocument();
+    expect(screen.getByTestId('recent-queries-button')).toBeInTheDocument();
+    expect(screen.getByTestId('save-query-button')).toBeInTheDocument();
+    expect(screen.getByTestId('query-panel-error')).toBeInTheDocument();
   });
 
-  it('hides date time picker when dataset is null', () => {
+  it('renders components when dataset is null', () => {
```
angle943: this test is no longer needed if you remove the dataset logic

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.test.tsx
```diff
@@ -122,10 +108,14 @@ describe('QueryPanelWidgets', () => {
...
-    expect(screen.queryByTestId('date-time-range-picker')).not.toBeInTheDocument();
+    // The component renders the same components regardless of dataset
+    expect(screen.getByTestId('filter-panel-toggle')).toBeInTheDocument();
+    expect(screen.getByTestId('recent-queries-button')).toBeInTheDocument();
+    expect(screen.getByTestId('save-query-button')).toBeInTheDocument();
+    expect(screen.getByTestId('query-panel-error')).toBeInTheDocument();
   });
 
-  it('hides date time picker when dataset is undefined', () => {
+  it('renders components when dataset is undefined', () => {
```
angle943: this test is no longer needed. These additions will unnecessary increase our unit test ci

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.test.tsx
```diff
@@ -110,10 +92,14 @@ describe('QueryPanelWidgets', () => {
...
-    expect(screen.getByTestId('date-time-range-picker')).toBeInTheDocument();
+    // The component renders the same components regardless of dataset
+    expect(screen.getByTestId('filter-panel-toggle')).toBeInTheDocument();
+    expect(screen.getByTestId('recent-queries-button')).toBeInTheDocument();
+    expect(screen.getByTestId('save-query-button')).toBeInTheDocument();
+    expect(screen.getByTestId('query-panel-error')).toBeInTheDocument();
   });
 
-  it('hides date time picker when dataset has no time field', () => {
+  it('renders components when dataset has no time field', () => {
```
angle943: no longer needed

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.test.tsx
```diff
@@ -88,20 +72,18 @@ describe('QueryPanelWidgets', () => {
...
-    // Check right section components
-    expect(screen.getByTestId('date-time-range-picker')).toBeInTheDocument();
-    expect(screen.getByTestId('run-query-button')).toBeInTheDocument();
+    // Check that vertical separators are present
+    const separators = container.querySelectorAll('.exploreQueryPanelWidgets__verticalSeparator');
+    expect(separators).toHaveLength(3); // There should be 3 separators between the 4 components
   });
 
-  it('shows date time picker when dataset has time field', () => {
+  it('renders components regardless of dataset time field', () => {
```
angle943: no longer needed

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.test.tsx
```diff
@@ -88,20 +72,18 @@ describe('QueryPanelWidgets', () => {
     // Check main container
     expect(container.querySelector('.exploreQueryPanelWidgets')).toBeInTheDocument();
 
-    // Check left section components
+    // Check left section components that are actually rendered
```
angle943: why do we need 'that are actually rendered'?

angle943:
I tried changing before date to like 1998 absolute date and i got this:

Screenshot 2025-07-22 at 6 02 25 PM

angle943:
Here is a bug:

- I set date time range to last fifteen years. Query is empty. i fire.
- I update time to something, like (July 2, 2018 @ 12:00:00 - july 22, 2021 @ 18:05), DON"T FIRE. But go to query editor, and type `| where AvgTicketPrice > 1` and press `CMD+ENTER` to fire query.

The result I see is that the fired query DOES NOT adhere to the date time range change. It uses the old date time range


Screenshot 2025-07-22 at 6 07 59 PM


- this example is not only applicable to running `cmd+Enter` in the query editor, but pretty much anywhere where the action fires the query. Clicking new button, loading a saved search, etc


TLDR: **you need the `from` and the `to` state in the redux**

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.test.tsx
Maosaic: Add a `data-test-subj` instead

angle943: no don't check for vertical separators. Are we going to check to see if every single `div` and span and everythign is there too? We shouldn't be using unit test to check for UI fluff. Its not helpful.

For example, hopefully @Maosaic in your show/hide field button removal, you removed one of the separators. Do you then need to update this to show 2?

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.tsx
```diff
@@ -32,14 +24,6 @@ export const QueryPanelWidgets = () => {
         {/* TODO: Actions should go here */}
         <QueryPanelError />
       </div>
-
-      {/* Right Section */}
-      <div className="exploreQueryPanelWidgets__right">
```
angle943, line 34: If you remove this div (which you shouldn't) then you should also erase the styling in the scss file

ananzh:
> I tried changing before date to like 1998 absolute date and i got this:
> 
> Screenshot 2025-07-22 at 6 02 25 PM

How did you set 1998? I tried to type 1998 and the DatePicker will ask you to input a correct date format
Screenshot 2025-07-23 at 8 11 13 AM

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.tsx
ananzh: Good catch. I think it got removed due to rebase. Let me add them back.

ananzh:
> Here is a bug:
> 
> * I set date time range to last fifteen years. Query is empty. i fire.
> * I update time to something, like (July 2, 2018 @ 12:00:00 - july 22, 2021 @ 18:05), DON"T FIRE. But go to query editor, and type `| where AvgTicketPrice > 1` and press `CMD+ENTER` to fire query.
> 
> The result I see is that the fired query DOES NOT adhere to the date time range change. It uses the old date time range
> 
> Screenshot 2025-07-22 at 6 07 59 PM
> * this example is not only applicable to running `cmd+Enter` in the query editor, but pretty much anywhere where the action fires the query. Clicking new button, loading a saved search, etc
> 
> TLDR: **you need the `from` and the `to` state in the redux**

I understand your point and I think this is debatable. This is following current discover behavior. Only when user click update button, it means they actually want to reset the time. This is current opensearch dashboard behavior for multiple plugins, like dashboard, discover, visualization.

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.tsx
ananzh: Will restore it

src/plugins/explore/public/components/top_nav/components/smart_update_button.tsx
ananzh: loll. What would be a better name? Does it smart if it can detect both Datepicker local state change through search bar and also our query panel context change?

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.test.tsx
ananzh: will clean it.

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.test.tsx
ananzh: will clean it

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.test.tsx
ananzh: will clean it

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.test.tsx
ananzh: will clean it.

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_widgets.test.tsx
ananzh: I can clean out this part.

src/plugins/explore/public/application/utils/state_management/slices/query_editor/query_editor_slice.ts
ananzh: will update the unit test

src/plugins/explore/public/application/utils/state_management/actions/query_editor/run_query/run_query.ts
ananzh: Sure. I could refactor this part.

src/plugins/explore/public/components/top_nav/components/smart_update_button.tsx
ananzh: QueryExecutionButton?

src/plugins/explore/public/application/utils/state_management/actions/query_editor/run_query/run_query.test.ts
```diff
@@ -41,134 +41,205 @@ jest.mock('../../detect_optimal_tab', () => ({
 
 describe('runQueryActionCreator', () => {
   const mockDispatch = jest.fn();
+  const mockGetState = jest.fn();
   const mockServices = { some: 'service' } as any;
   const query = 'this is some query';
 
   beforeEach(() => {
     jest.clearAllMocks();
+    // Default state with UPDATE status
```
kavilla: nit: prolly unneeded comment

src/plugins/explore/public/components/top_nav/query_execution_button/query_execution_button.tsx
```diff
@@ -0,0 +1,105 @@
...
+ * Query execution button that calculates and manages its own status in Redux.
+ * Shows different text based on button status:
+ * - "Update" when status is 'UPDATE' (there are changes)
+ * - "Refresh" when status is 'REFRESH' (no changes, just re-run)
+ * - Disabled when status is 'DISABLED' (validation errors)
+ */
+export const QueryExecutionButton: React.FC<QueryExecutionButtonProps> = ({
+  onClick,
+  editorText,
+  services,
```
kavilla: how come not using contextual services?

kavilla:
super date picker has a `showUpdateButton={true|false}` prop that does something similiar.

do you mind explaining the experience in this situation. the user didn't fire but updated the query? and refreshes the page

src/plugins/navigation/public/top_nav_menu/top_nav_menu.tsx
```diff
@@ -265,6 +265,11 @@ export function TopNavMenu(props: TopNavMenuProps): ReactElement | null {
                     </EuiFlexItem>
                     <EuiFlexItem className="globalDatePicker">
                       <div ref={datePickerRef} />
+                      {!showDatePicker && props.customSubmitButton && (
```
kavilla, line 268: why wouldn't it be similar to         showDatePicker={![TopNavMenuItemRenderType.OMITTED, false].includes(showDatePicker!)}


and it's a little confusing that the custom submit button would  be dependendent on a date picker being present or not.

if it is shouldn't it be apart of the the date picker?

kavilla:
are we not passing by ref the date picker? if we are does the data plugin need any change at all?

couldn't it just be apart of the date picker we pass in?

src/plugins/explore/public/components/top_nav/query_execution_button/query_execution_button.tsx
ananzh: will change

src/plugins/explore/public/application/utils/state_management/actions/query_editor/run_query/run_query.test.ts
ananzh: will remove

src/plugins/navigation/public/top_nav_menu/top_nav_menu.tsx
ananzh: You're right that it seems confusing! Here's what's happening in the code:
* SearchBar's logic: SearchBar only renders QueryBar if there's something to show
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/086ebe228c769a5572361fe13c15c3255d2dca93/src/plugins/data/public/ui/search_bar/search_bar.tsx#L245

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/086ebe228c769a5572361fe13c15c3255d2dca93/src/plugins/data/public/ui/search_bar/search_bar.tsx#L513

* Explore's current configuration:
```
<SearchBar
  showQueryBar={true}        // ✅ Want the container
  showQueryInput={false}     // ❌ Can’t show query input (Explore has its own query panel)
  showDatePicker={/* depends on time field */}  // ❌ False when no time field
  customSubmitButton={customSubmitButton}
/>
```
So we `return true && (false || false)`,  QueryBar doesn't render - nothing to show. So the submit button is gone.
The fix is to use the date picker's reserved UI space when it's empty. The dependency is spatial (sharing UI space), not logical.

```
┌─ TopNavMenu ──────────────────────────────┐
 │ [Menu Items] [Empty Date Area: Submit Btn]                                                                 │
└──────────────────────────────────────┘

```
The submit button isn't logically part of the date picker - it's just borrowing the UI space when the date picker isn't there.

There are other alternative approaches I considered. For example,
```
return showQueryBar && (showDatePicker || showQueryInput || customSubmitButton);
```
But I don't think it is proper. I explained more in the date picker ref.

ananzh:
> are we not passing by ref the date picker? if we are does the data plugin need any change at all?
> 
> couldn't it just be apart of the date picker we pass in?

Yeah thought about it. Here are some findings and concerns:

**SearchBar focuses on data-related concerns:**
* Query management (query input, validation, execution)
* Time filtering (date picker, time ranges)
* Data filtering (filter bar, filter management)
* Index pattern handling

**The navigation plugin ([TopNavMenu](vscode-webview://1mjg7jsknnmpem1jqo99pfdbm5i9ciu8lueeqgk5riopc3mhobj2/src/plugins/navigation/public/top_nav_menu/top_nav_menu.tsx)) is responsible for:**
* UI layout (top navigation bar structure)
* Menu items (save, share, etc.)
* Custom buttons (like your customSubmitButton)
* Screen titles and branding

The datePickerRef portal logic lives inside QueryBarTopRow, which only renders when SearchBar's shouldRenderQueryBar returns true. For our explore case:
No QueryBar renders → No portal logic runs → No submit button appears

If we modify SearchBar's core logic to account for customSubmitButton, like in datePickerRef, this violate separation of concerns (SearchBar shouldn't know about TopNavMenu's custom buttons) and might break other existing usages. Our current approach is to handle it at the TopNavMenu level where we have full control over the layout and can make the decision based on available space. This keeps the the logic isolated and doesn't affect other plugins.

ananzh:
> super date picker has a `showUpdateButton={true|false}` prop that does something similiar.
> 
> do you mind explaining the experience in this situation. the user didn't fire but updated the query? and refreshes the page

Yeah sure. Let me add some explains:

1. Query state is updated through set query methods in query slice when
* Set/change dataset
* Load saved query
* Initialize explore page
* Run query
Then two middleware will 1) sync up the query to queryStringManager 2) persist query state to url

2. DateRange is updated through setDateRange when date picker is updated then when execute query timefilterSyncMiddleware will sync with global timefilter. Global timefilter will persist the state to _g in url.


Because of this, we have one special scenario
1. User updates date picker —> setDateRange
2. User does NOT click the update/refresh button
3. User fire a query via keyboard shortcut —>  timefilterSyncMiddleware —> _g in url is updated
4. User refreshes the page —> see updated query and date range state.

For the scenario you mentioned
1. User updates 
    1. the query (types in query panel) —> will not update query state. Click update/refresh button will call run query which will call setQueryStringWithHistory 
    2. date picker —> will update local dateRange state in query editor slice
2. User does NOT click the update/refresh button --> no timefilterSyncMiddleware
3. User refreshes the page
4. Since query slice is not set and global time filter is not synced, after refresh user will see query state or date range before updates (step 1)

Maosaic (approved):

review_pr.txt
```diff
@@ -0,0 +1,9160 @@
+File: package.json
```
angle943: is this file needed?

src/plugins/explore/public/application/utils/state_management/actions/query_editor/run_query/run_query.ts
```diff
@@ -14,16 +14,25 @@ import { detectAndSetOptimalTab } from '../../detect_optimal_tab';
...
   dispatch(clearQueryStatusMap());
 
   await dispatch(executeQueries({ services }));
-  dispatch(detectAndSetOptimalTab({ services }));
+
+  // Only detect and set optimal tab for update
+  if (isUpdate) {
+    dispatch(setActiveTab(''));
+    dispatch(detectAndSetOptimalTab({ services }));
+  }
```
angle943: nice

src/plugins/explore/public/components/top_nav/query_execution_button/query_execution_button.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const query = useSelector((state: RootState) => state.query);
+
+  const determineButtonStatus = useCallback((): QueryExecutionButtonStatus => {
+    if (dateRange && isTimeRangeInvalid(dateRange)) {
+      return 'DISABLED';
+    }
+
+    const currentTimeRange = timefilter ? timefilter.getTime() : { from: 'now-15m', to: 'now' };
+    const currentQuery = getQueryWithSource(
+      queryStringManager ? queryStringManager.getQuery() : { query: '', language: 'kuery' }
```
angle943: do we want `kuery` to be the default?

src/plugins/explore/public/components/top_nav/query_execution_button/query_execution_button.tsx
```diff
@@ -0,0 +1,105 @@
...
+export const QueryExecutionButton: React.FC<QueryExecutionButtonProps> = ({
+  onClick,
+  editorText,
+}) => {
+  const { services } = useOpenSearchDashboards<ExploreServices>();
+  const dispatch = useDispatch();
+  const dateRange = useSelector(selectDateRange);
+  const timefilter = services?.data?.query?.timefilter?.timefilter;
+  const queryStringManager = services?.data?.query?.queryString;
+  const query = useSelector((state: RootState) => state.query);
```
angle943: don't we have a memoized selector for this/

src/plugins/explore/public/components/top_nav/top_nav.tsx
```diff
@@ -171,27 +182,59 @@ export const TopNav = ({ setHeaderActionMenu = () => {}, savedExplore }: TopNavP
...
+      onQuerySubmit={handleQuerySubmit}
+      onQueryChange={handleQueryChange}
+      customSubmitButton={customSubmitButton}
       groupActions={true}
       screenTitle={screenTitle}
       queryStatus={queryStatus}
-      showQueryBar={false}
+      showQueryBar={true}
+      showQueryInput={false}
+      showFilterBar={false}
```
angle943, line 232: how is the datepicker getting rendered? I see that you are passing a customSubmitZButton, but not a datepicker?

src/plugins/explore/public/components/top_nav/query_execution_button/query_execution_button.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    const hasChanges = isQueryUpdated || Boolean(isDateRangeUpdated);
+
+    return hasChanges ? 'UPDATE' : 'REFRESH';
+  }, [dateRange, editorText, timefilter, queryStringManager, query]);
+
+  useEffect(() => {
+    const status = determineButtonStatus();
+    dispatch(setQueryExecutionButtonStatus(status));
+  }, [determineButtonStatus, dispatch]);
```
angle943: if we were to have this logic in `handleQueryChange` within top_nav.tsx`, are we able to avoid this useEffect? This useEffect will run for every time the editorText will update, but I think if we put this logic to `handleQueryChange`, it won't

src/plugins/explore/public/components/top_nav/top_nav.tsx
```diff
@@ -171,27 +182,59 @@ export const TopNav = ({ setHeaderActionMenu = () => {}, savedExplore }: TopNavP
...
+      if (queryAndDateRange.dateRange) {
+        dispatch(setDateRange(queryAndDateRange.dateRange));
+      }
+    },
+    [dispatch]
+  );
+
+  const handleQuerySubmit = useCallback(() => {
+    dispatch(onEditorRunActionCreator(services, editorText));
+    dispatch(setQueryExecutionButtonStatus('REFRESH'));
```
angle943: will `onEditorRunActionCreator` always set the execution button status to REFRESH? If so, can we add `dispatch(setQueryExecutionButtonStatus('REFRESH'))` to the end of `onEditorRunActionCreator`?

review_pr.txt
ananzh: removed

src/plugins/explore/public/components/top_nav/query_execution_button/query_execution_button.tsx
ananzh: I think we need a refactor for editorText in the follow up for performance

src/plugins/explore/public/components/top_nav/top_nav.tsx
ananzh: This is to reset the button status. Could be in onEditorRunActionCreator

angle943 (dismissed):

angle943 (approved):

LDrago27 (approved):

PR #10269, opened 2025-07-23
[explore] move dataset select to query panel widgets

kavilla:
### Description

Moving the location and updating some of the state logic. Includes consolidation of the state.

### Issues Resolved

n/a

## Screenshot

Screenshot 2025-07-29 at 1 00 35 PM
Screenshot 2025-07-29 at 1 00 52 PM

## Changelog

- fix: move dataset select to query panel widgets

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data/common/data_views/utils.ts
```diff
@@ -85,11 +85,70 @@ export const getDataViewTitle = async (
...
+ * @returns The dataset type or undefined
+ */
+export const getDatasetTypeFromReference = (
+  dataSourceRef?: DataViewSavedObjectReference
+): string | undefined => {
+  if (!dataSourceRef?.name) {
+    return undefined;
+  }
+  return extractDatasetTypeFromUri(dataSourceRef.name);
+};
```
angle943, line 148: could we get some unit tests for these functions?

src/plugins/explore/public/application/utils/state_management/middleware/dataset_change_middleware.ts
```diff
@@ -0,0 +1,101 @@
...
+          services.notifications?.toasts.addError(error, {
+            title: 'Error loading dataset',
+          });
+        }
+      }
+    }
+
+    return result;
+  };
+};
```
angle943, line 101: can we get unit test for this file?

src/plugins/explore/public/application/utils/state_management/middleware/dataset_change_middleware.ts
```diff
@@ -0,0 +1,101 @@
...
+              currentDataset,
+              {
+                ...services,
+                storage: services.storage as any,
+              },
+              false
+            );
+          }
+
+          await store.dispatch(executeQueries({ services }) as any);
```
angle943, line 89: i think this is fine, but leaving this comment here anyways. Are you certain this will never executeQueries when we don't want to?

src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_query_panel_editor.ts
```diff
@@ -90,12 +91,17 @@ export interface UseQueryPanelEditorReturnType {
...
-  const [editorText, setEditorText] = useState<string>(queryString);
+  const qs = useSelector(selectQueryString);
+  const [editorText, setEditorText] = useState<string>(qs);
   const [editorIsFocused, setEditorIsFocused] = useState(false);
+  const {
+    data: {
+      dataViews,
+      query: { queryString },
+    },
+  } = services;
```
angle943, line 105: ah i think i introduced this merge conflict :D .

the `qs` vs `queryString` is very confusing. What if we didn't destructure `queryString` and instead on line 164 just did `query.queryString.getQuery().dataset`?

Personally i would just not destructure any of these though and refactor lines below to be:

```
const currentDataset = services.data.query.queryString.getQuery().dataset;

const currentDataView = await services.data.dataViews.get(...)
```

This way there is no confusion between qs vs queryString
```

File: src/plugins/explore/public/components/query_panel/query_panel_widgets/dataset_select/dataset_select.tsx [hunk 5]
```diff
@@ -0,0 +1,96 @@
...
+        services.notifications?.toasts.addError(error, {
+          title: 'Error selecting dataset',
+        });
+      }
+    },
+    [queryString, dispatch, services]
+  );
+
+  return <DatasetSelect onSelect={handleDatasetSelect} appName="explore" />;
+};
```
angle943, line 96: this change might cause the stories in `queryPanel` folder to break (because i don't think i mock DatasetSelect in services)

angle943: could we also get unit tests for this?

src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_query_panel_editor.ts
kavilla: for me, it was already an established pattern for data plugin to be destructured to `queryString`. so if you search the other files you can see we use `queryString`. it would be a little bit more confusing working between the data plugin and the explore plugin if `queryString` referred to the string in the explore plugin while it meant the manager in other places

kavilla: i would be okay making it the `userQueryString` which is a little bit more patterned

angle943: I was going to say that perhaps in explore, that is no longer the pattern, but that the `const queryString = useSelector(seletQueryString)` is the pattern, but searching for `useSelector(selectQueryString)`, its only being used in this file lol. 

I got you, but yeah maybe we can do something else other than `qs` to show that its different from the destructured queryString

angle943: I also was thinking about removing the `useSeletor(selectQueryString)` entirely, and using only the destructured queryString. But i don't think that will work if we want the default editor to have no query string

src/plugins/explore/public/application/utils/state_management/middleware/dataset_change_middleware.ts
kavilla: added!

src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_query_panel_editor.ts
kavilla: personally, i think we should just be passing around the query object. i think some other places i saw `query` was being referenced when it was the query string so it got a little bit confusing. might be easier for contributors to understand if we have a single consolidated consistent object.

kavilla: but i have updated it to be `userQueryString` similar to the language service

src/plugins/explore/public/components/query_panel/query_panel_widgets/dataset_select/dataset_select.tsx
kavilla: added. and did a junk mock. i have the story for it mocked in the data plugin but when i tried to have a service return a full component the storybook got really slow. lemme know if you think this is acceptable. if not i can create a TODO to followup on this because I think we haven't fully locked down what we want this component to look like and where it will go.

ananzh (approved):

abbyhu2000 (dismissed):

ananzh (approved):

abbyhu2000 (approved):

ananzh (approved):

PR #10270, opened 2025-07-23
[Banner Plugin] Support New UI with no overlap

### LINKED ISSUE #10139: [Banner Plugin] Support new page header
### Description

Currently, the new home page and flyout components in OpenSearch Dashboards are implemented as fixed-position elements and do not account for the `--global-banner-height` offset used by the legacy layout.

This causes layout issues (e.g., overlap with the global banner) when the banner plugin is enabled.

And therefore the banner plugin is currently disabled now.

### Expected Behavior

When the global banner is present:

* The new home page content should shift down accordingly.
* Flyouts and navigation toggle buttons should respect the global banner offset, similar to how the default home page uses `$osdHeaderOffset`.

### Proposed Solution

Introduce a layout variable (e.g., `new-ui-header-offset`) to  solve layout issues for:

* `.content`
* `.euiFlyout`
* `.newPageTopNavExpander`

This variable would incorporate the banner height and allow for consistent offset handling across both the default and new layouts.

### Tasks

* [ ] Define and inject `--new-ui-header-offset` when the banner is enabled
* [ ] Refactor affected components to use the new offset variable
* [ ] Ensure compatibility with dark mode, responsive layout
* [ ] Add tests and visual validation for banner + new home layout interaction

sidwang42:
### Description

This PR introduces support for global banner offset handling in the new UI layout of OpenSearch Dashboards. It resolves layout conflicts where the global banner overlapped with fixed-position elements such as the new home page, panel and flyout.

Specifically:

* Adjusts z-index of the banner to respect fullscreen mode and full height flyout.
* `injectedMetadata` to determine banner status and enable layout adaptation. (Temp solution with minimal change, will be handled in the next PR)
* Ensures consistent behavior between legacy and updated headers.

### Issues Resolved

fixes #10139

## Screenshot

iShot_2025-07-23_11 06 09

image

## Testing the changes

1. Enable the banner plugin.
2. Enable the new UI layout via YAML.
3. Verify that the new home page content no longer overlaps with the global banner.
4. Open and close flyouts to confirm offset respects banner height.
5. Toggle fullscreen mode to ensure banner is hidden as expected.
6. Test across light/dark themes and responsive breakpoints.

## Changelog

- fix: Apply global banner offset to new UI layout and flyouts

### Check List

- [x] All tests pass

  - [x] `yarn test:jest`
- [x] Update [[CHANGELOG.md](https://chatgpt.com/g/g-p-687e737387748191aa22ca9d4628a4e1-banner/CHANGELOG.md)](./../CHANGELOG.md)


src/plugins/banner/public/index.scss
```diff
@@ -25,7 +25,7 @@ $expandedHeaderWidth: $headerWidth * 2;
   top: 0;
   left: 0;
   right: 0;
-  z-index: 1100;
+  z-index: 999;
```
Maosaic: why does the banner need `z-index`?

src/plugins/banner/public/index.scss
```diff
@@ -43,4 +43,16 @@ $expandedHeaderWidth: $headerWidth * 2;
...
+    position: relative;
+    top: var(--global-banner-height);
+  }
+}
+
+.app-wrapper.app-wrapper-with-banner.app-wrapper-with-new {
+  position: relative;
+  top: var(--global-banner-height);
+}
+
```
Maosaic: This change will break vertical scrolling on some page

src/core/public/rendering/rendering_service.tsx
```diff
@@ -66,6 +67,8 @@ export class RenderingService {
           <AppWrapper
             chromeVisible$={chrome.getIsVisible$()}
             sidecarConfig$={overlays.sidecar.getSidecarConfig$()}
+            useUpdatedHeader={(chrome as any).useUpdatedHeader}
```
raintygao, line 70: Any reasons that we use any in this file?

raintygao: Another solution is that:
If you want to retrieve `useUpdatedHeader`, the `start` of `RenderingService` is executed later than `uiSettings start`, so you can set add `uiSettings` to the params of `start` hook, then you don't need to change chrome service

raintygao:
BTW, https://github.com/opensearch-project/dashboards-assistant uses sidecar service, not sure if this global style change would have conflicts with that, could you install this plugin and have a test?  
You need to set `assistant.chat.enabled: true` in yml file when using assistant.

src/plugins/banner/public/index.scss
sidwang42: > why does the banner need `z-index`?

You're right — z-index: 1100 ensures the banner appears above the current header, but it unfortunately blocks EUI panels and flyouts (z-index: 1000). On the other hand, setting it at or below 1000 avoids conflicts with EUI overlays, but causes overlap with header components. Shown is the below screenshot:



I'm still weighing the trade-offs. The most reasonable long-term approach might be to set the banner below 1000 and adjust the header's z-index separately — though that feels out of scope for this PR.

virajsanghvi:
Haven't looked at code yet, but design wise, should this push everything down or just the content on the right (leaving the left nav in place)? @kgcreative @lauralexis

src/core/public/chrome/chrome_service.tsx
```diff
@@ -670,4 +671,10 @@ export interface InternalChromeStart extends ChromeStart {
    * @internal
    */
   getHeaderComponent(): JSX.Element;
+
+  /**
+   * Whether to use the updated header UI
```
zhongnansu, line 676: what does "updated" mean? Are you referring to new UI? can we use "newHeader" to align with the naming convention across the codebase

src/core/public/rendering/rendering_service.tsx
```diff
@@ -66,6 +67,8 @@ export class RenderingService {
...
+            injectedMetadata={injectedMetadata}
```
zhongnansu, line 71: questions, Later when we refactor the chrome service to expose a interface of `setGlobalBanner`, then we can get rid of the usage of `injectedMetadata` right?

zhongnansu:
> Haven't looked at code yet, but design wise, should this push everything down or just the content on the right (leaving the left nav in place)? @kgcreative @lauralexis

Tagging UX designer of this project: @canascar. 

I think this is the original aligned UI/UX. If there's any change, please let us know soon, and a UX mock up in this case would help.

src/plugins/banner/public/index.scss
sidwang42: @Maosaic 
Could you clarify which pages are affected and how the scrolling breaks?

src/core/public/rendering/rendering_service.tsx
sidwang42: > Another solution is that: If you want to retrieve `useUpdatedHeader`, the `start` of `RenderingService` is executed later than `uiSettings start`, so you can set add `uiSettings` to the params of `start` hook, then you don't need to change chrome service

Great solution—thanks for the suggestion! I'm working on this approach now.

src/core/public/rendering/rendering_service.tsx
sidwang42: > questions, Later when we refactor the chrome service to expose a interface of `setGlobalBanner`, then we can get rid of the usage of `injectedMetadata` right?

yes

src/plugins/banner/public/index.scss
Maosaic: > On the other hand, setting it at or below 1000 avoids conflicts with EUI overlays

But why this is happening? You have added offset height to adjust component height under banner

src/plugins/banner/public/index.scss
Maosaic: Discover page for example, all component are design to render in a single page, vertical scroll should not happen on page level.

src/plugins/banner/public/index.scss
sidwang42: @Maosaic 
> But why this is happening? You have added offset height to adjust component height under banner

That’s true for many pages, but not all. For example, the **Dev Tools** page is wrapped in a full-page `EuiPanel`, and its help button triggers a right-side `EuiFlyout`—also designed to cover the full page.
Similarly, the **Inspect** flyout in **Discover** has the same behavior and ends up overlapping the banner.

These components don’t respect the offset and override with `!important` isn't a viable option—it breaks encapsulation and isn't permitted.

So to prevent blocking those EUI overlays, the banner’s `z-index` has to stay below 1000.
But that reintroduces overlap with header components, since they’re not layered high enough.

After hours of debugging, the only long-term fix I see is adjusting the header's stacking like:

```scss
.headerGlobalNav--withBanner .newTopNavHeader {
  z-index: 900; // lower than banner
  position: relative;
}
```

This ensures both banner visibility and proper overlay behavior.

sidwang42: Here is the screenshot of developer tool when setting z-index to 1001 (higher than the default z-index set by euiflyout and panel)
image

src/plugins/banner/public/index.scss
sidwang42: > Discover page for example, all component are design to render in a single page, vertical scroll should not happen on page level.

@Maosaic 
Working on it right now, thanks for pointing out.

src/plugins/banner/public/index.scss
Maosaic: Understand now, thank you

src/core/public/chrome/chrome_service.tsx
virajsanghvi: +1 can we use existing terminology

src/plugins/banner/public/index.scss
virajsanghvi: Can you provide screenshots of any pages that don't respect the banner height to validate they are not broken? Similarly, screenshots of any overlay scenarios that may conflict would be good to include as well.

virajsanghvi:
> > Haven't looked at code yet, but design wise, should this push everything down or just the content on the right (leaving the left nav in place)? @kgcreative @lauralexis
> 
> Tagging UX designer of this project: @canascar.
> 
> I think this is the original aligned UI/UX. If there's any change, please let us know soon, and a UX mock up in this case would help.

Ack - won't block this, but not sure I fully agree, e.g. @canascar Doesn't this flyout experience look weird due to this (it doesn't get pushed down but the left nav does): 
image

virajsanghvi (commented):
Code looks fine to me, would be good to address comments.

One q: what level of validation across all the pages have we done to validate nothing is broken? I think that's the biggest risk here

sidwang42:
@virajsanghvi 

Thanks for the review and the comments!

For validation: unfortunately, there isn’t an easy automated way to confirm banner impact across all pages. I’ve been manually clicking through all the pages to verify behavior, and I’ve also asked other teams to join the review and raise any bugs they encounter.

The way I examined the changes was by testing:

Old UI with banner enabled

Old UI with banner disabled

New UI with banner enabled

New UI with banner disabled

I also have my own document tracking all the major ui bugs found along the way — glad to share it if you’d like to review.

sidwang42:
> Can you provide screenshots of any pages that don't respect the banner height to validate they are not broken? Similarly, screenshots of any overlay scenarios that may conflict would be good to include as well.

Every page respects the banner height — the only exceptions are **EUI flyouts** or **EUI panels**. Below are the places I found where these overlays overlap with the banner plugin:

1. **Search on the home screen**

   * OUI panel displayed when performing a search
   * image

2. **Dev Tools**

   * OUI panel that takes the entire screen
   * image


3. **Flyouts on Dev Tools and Inspect in Discovery page**

   * EUI flyout overlapping with the banner
   * image

   * image
   
**There may be more cases, especially in plugin‑specific UIs, but I believe these examples cover the main components likely to cause issues.**

zhongnansu (commented):
could you address comments and resolve upstream conflicts? @sidwang42

sidwang42:
Hi All, I just updated this PR to fix the upstream conflicts.

I want to highlight this line of the change:
```css
z-index: 800 !important;
````

This is necessary to override the OUI rule:

```scss
.ouiFlyout.ouiFlyout--push {
  box-shadow: none;
  clip-path: none;
  animation-duration: 0s; // Don't animate on loading a docked nav
  border-left: $ouiBorderThick;
  z-index: $ouiZHeader - 1; // effectively 999
}
```

Without this override, the navigation icon becomes hidden behind the flyout.

The `!important` is added here because the OUI rule already assigns a high z-index (`999`), and normal overrides would not be sufficient in this case. This ensures the fix applies reliably without unintended side effects.

A more maintainable long-term fix would be to update the following section in OUI itself:
[OUI related code](https://github.com/opensearch-project/oui/blob/ed1f13f53aa106861511709567b3c3d5b76f763b/src/components/flyout/_flyout.scss#L198)

zhongnansu:
> A more maintainable long-term fix would be to update the following section in OUI itself: [OUI related code](https://github.com/opensearch-project/oui/blob/ed1f13f53aa106861511709567b3c3d5b76f763b/src/components/flyout/_flyout.scss#L198)

@sidwang42 What's your proposal for the long-term fix in OUI? Can you open an issue in OUI for tracking, and add a TODO in the current code linking that issue?

src/plugins/banner/public/index.scss
```diff
@@ -43,4 +43,21 @@ $expandedHeaderWidth: $headerWidth * 2;
...
 
+.headerGlobalNav--withBanner {
+  #globalHeaderBars {
+    position: relative;
+    top: var(--global-banner-height);
+  }
+}
+
+.headerGlobalNav--withBanner .newTopNavHeader {
+  z-index: 900; // lower than banner
```
zhongnansu: What banner are we talking about here? Can we define the banner z-index as constant, and use something similar to `$ouiZHeader - 1` to avoid hardcoded z-index number?

src/core/public/rendering/rendering_service.tsx
sidwang42: @raintygao 
Tried this approach by adding:

```ts
const useUpdatedHeader = uiSettings.get('home:useNewHomePage', false);
````

at the `start` hook params of rendering service, but it didn’t work — seems like `uiSettings` hasn’t started yet:

```
Version: 3.2.0
Build: 9007199254740991
TypeError: Cannot read properties of undefined (reading 'get')
    at RenderingService.start (...)
```
Do you have any suggestions on a better way to retrieve `useUpdatedHeader` at this point in the lifecycle?

src/core/public/chrome/chrome_service.tsx
sidwang42: Thanks for the feedback @zhongnansu @virajsanghvi 

Might have to use the updated header for now to ensure consistency, since both the chrome service, header service and rendering service systematically use the updated header.  

I do agree that using `newHeader` would better align with the existing convention. However, switching over would require updating ~70 usages, which feels out of scope for this PR. I’d be happy to open a follow‑up PR to make that broader change if the team agrees.

zhongnansu (approved):
overall lgtm, minor comments left.

sidwang42:
> BTW, https://github.com/opensearch-project/dashboards-assistant uses sidecar service, not sure if this global style change would have conflicts with that, could you install this plugin and have a test? You need to set `assistant.chat.enabled: true` in yml file when using assistant.

Thanks for pointing this out! The current refactored approach no longer touch the sidecar service or make global OUI style modifications. If dashboards-assistant does overlap with the banner, the fix would need to be handled on the dashboards-assistant side.

src/plugins/banner/public/index.scss
sidwang42: Updated, thanks for pointing out this solution.

sidwang42:
> > A more maintainable long-term fix would be to update the following section in OUI itself: [OUI related code](https://github.com/opensearch-project/oui/blob/ed1f13f53aa106861511709567b3c3d5b76f763b/src/components/flyout/_flyout.scss#L198)
> 
> @sidwang42 What's your proposal for the long-term fix in OUI? Can you open an issue in OUI for tracking, and add a TODO in the current code linking that issue?

@zhongnansu 
Sure! I’ve added the comment in the latest revision.
Here is the link to the oui issue created:
https://github.com/opensearch-project/oui/issues/1622

zhongnansu (approved):

xinruiba (approved):

PR #10273, opened 2025-07-25
Update version compatibility process between OSD and its plugins

nvinjam-amazon:
### Description

**Before the changes:**
While installing plugins with OpenSearch Dashboards, there was strict version matching between plugins and the core platform, requiring exact version matches (e.g., OpenSearch Dashboards version 3.1.0 must match plugin version 3.1.0) across major, minor, and patch versions.
**After the changes:**
Enhanced plugin version compatibility with two distinct validation modes:
- **Strict Mode (Default)**: Requires exact version matching between OpenSearch Dashboards and plugins (e.g., 3.2.0 matches only 3.2.0)
- **Ignore Mode**: Bypasses all version validation, allowing any plugin version to be installed with warnings

Added `--single-version` flag with options:
- `--single-version=strict` (default): Enforces exact version matching
- `--single-version=ignore`: Skips version validation entirely

This approach maintains backward compatibility while providing flexibility for development and testing scenarios.
### Issues
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/10294

### Issues Resolved
- Provides clear version compatibility options for different use cases.
- Maintains ecosystem stability with strict mode as default.
- Offers flexibility for development/testing with ignore mode.
- Eliminates confusion from multiple validation modes.
- Ensures consistent behavior across CLI and plugin discovery.
- Backward compatible - existing workflows continue unchanged.
## Screenshot
Screenshot 2025-07-25 at 1 29 48 PM

### Unit Testing
All the unit tests pass for:
- `opensearch_dashboards.test.js`
- `plugin_manifest_parser.test.ts`
- `settings.test.js`
- `index.test.js`
- `plugins_discovery.test.ts`

### Manual Testing
**Scenario 1: Default Strict Mode (Version Mismatch)**
```
(25-08-14 20:55:48) <0> [~/GitHub/OpenSearch-Dashboards]  
dev-dsk-nvinjam-2b-86031268 % ./scripts/use_node scripts/opensearch_dashboards_plugin.js install file://$(pwd)/test-plugin.zip

Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme
Attempting to transfer from file:///home/nvinjam/GitHub/OpenSearch-Dashboards/test-plugin.zip
Transferring 717 bytes....................
Transfer complete
Retrieving metadata from plugin archive
Extracting plugin archive
Extraction complete
Plugin installation was unsuccessful due to error "Plugin testPlugin [2.0.0] is incompatible with OpenSearch Dashboards [3.2.0]. Strict mode requires exact version match."
```
**Scenario 2: Ignore Mode (Version Mismatch)**
```
(25-08-14 20:56:21) <70> [~/GitHub/OpenSearch-Dashboards]  
dev-dsk-nvinjam-2b-86031268 % ./scripts/use_node scripts/opensearch_dashboards_plugin.js install --single-version=ignore file://$(pwd)/test-plugin.zip

Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme
Attempting to transfer from file:///home/nvinjam/GitHub/OpenSearch-Dashboards/test-plugin.zip
Transferring 717 bytes....................
Transfer complete
Retrieving metadata from plugin archive
Extracting plugin archive
Extraction complete
WARNING: Plugin testPlugin [2.0.0] major version differs from OpenSearch Dashboards [3.2.0]. Plugin may not function correctly.
Plugin installation complete
```
**Scenario 3: Invalid Mode**
```
(25-08-14 20:56:47) <0> [~/GitHub/OpenSearch-Dashboards]  
dev-dsk-nvinjam-2b-86031268 % ./scripts/use_node scripts/opensearch_dashboards_plugin.js install --single-version=loose file://$(pwd)/test-plugin.zip

Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme

/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/commander/index.js:396
      val = fn(val, self[name] === undefined ? defaultValue : self[name]);
            ^
Error: Invalid single-version mode: loose. Use 'strict' or 'ignore'.
    at /local/home/nvinjam/GitHub/OpenSearch-Dashboards/src/cli_plugin/install/index.js:70:17
    at Command.<anonymous> (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/commander/index.js:396:13)
    at Command.emit (node:events:517:28)
    at Command.call (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/commander/index.js:751:14)
    at Command.<anonymous> (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/src/cli/command.js:111:22)
    at apply (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/lodash/lodash.js:489:27)
    at Command.wrapper [as parseOptions] (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/lodash/lodash.js:5379:16)
    at Command.listener (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/commander/index.js:255:23)
    at Command.emit (node:events:517:28)
    at Command.parseArgs (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/commander/index.js:672:12)
    at Command.parse (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/commander/index.js:459:21)
    at Object.<anonymous> (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/src/cli_plugin/cli.js:76:9)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Module._compile (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/pirates/lib/index.js:136:24)
    at Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
    at Object.newLoader [as .js] (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/pirates/lib/index.js:141:7)
    at Module.load (node:internal/modules/cjs/loader:1203:32)
    at Function.Module._load (node:internal/modules/cjs/loader:1019:12)
    at Module.require (node:internal/modules/cjs/loader:1231:19)
    at Module.Hook._require.Module.require (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/require-in-the-middle/index.js:80:39)
    at Module.Hook._require.Module.require (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/require-in-the-middle/index.js:80:39)
    at Module.Hook._require.Module.require (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/require-in-the-middle/index.js:80:39)
    at Module.Hook._require.Module.require (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/require-in-the-middle/index.js:80:39)
    at Module.Hook._require.Module.require (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/node_modules/require-in-the-middle/index.js:80:39)
    at require (node:internal/modules/helpers:177:18)
    at Object.<anonymous> (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/src/cli_plugin/dev.js:32:1)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
    at Module.load (node:internal/modules/cjs/loader:1203:32)
    at Function.Module._load (node:internal/modules/cjs/loader:1019:12)
    at Module.require (node:internal/modules/cjs/loader:1231:19)
    at require (node:internal/modules/helpers:177:18)
    at Object.<anonymous> (/local/home/nvinjam/GitHub/OpenSearch-Dashboards/scripts/opensearch_dashboards_plugin.js:31:1)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
    at Module.load (node:internal/modules/cjs/loader:1203:32)
    at Function.Module._load (node:internal/modules/cjs/loader:1019:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:128:12)
    at node:internal/main/run_main_module:28:49
```
**Scenario 4: CLI Help**
```
(25-08-14 20:55:28) <64> [~/GitHub/OpenSearch-Dashboards]  
dev-dsk-nvinjam-2b-86031268 % ./scripts/use_node scripts/opensearch_dashboards_plugin.js install --help


Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme
Usage: bin/opensearch-dashboards-plugin install [options] <plugin/url>

install a plugin

Arguments:

  plugin/url                undefined

Options:
  -q, --quiet               disable all process messaging except errors
  -s, --silent              disable all process messaging
  -c, --config <path>       path to the config file (default: "/local/home/nvinjam/GitHub/OpenSearch-Dashboards/config/opensearch_dashboards.yml")
  -t, --timeout <duration>  length of time before failing; 0 for never fail
  --single-version <mode>   set version validation method: "strict", "ignore" (default: "strict")
  -h, --help                output usage information
```
## Changelog
- feat: Enhanced plugin version compatibility with strict/ignore modes.
- feat: Add `--single-version` flag with 'strict' and 'ignore' options.
### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


src/core/server/plugins/discovery/plugin_manifest_parser.ts
```diff
@@ -285,21 +285,14 @@ function isVersionCompatible(
     return true;
   }
 
-  const coercedActualOpenSearchDashboardsVersion = coerce(actualOpenSearchDashboardsVersion);
-  if (coercedActualOpenSearchDashboardsVersion == null) {
-    return false;
-  }
+  const coercedActual = coerce(actualOpenSearchDashboardsVersion);
```
virajsanghvi: Not clear to me where isVersionCompatible is used - does this function not need to take strict mode into account?

nvinjam-amazon: Hi @virajsanghvi, Thanks for reviewing the changes.

```isVersionCompatible``` function doesn't need strict mode because it's used during plugin discovery/startup for already installed plugins, while the ```--strict-version``` flag only applies during plugin installation via CLI command. They operate at different stages with different requirements.

**Comment by virajsanghvi** [comment]
Can you take a look at failing CI to see what's going on?

**Comment by nvinjam-amazon** [comment]
> Can you take a look at failing CI to see what's going on?

Yes, that's my bad, Somehow, I haven't included the ```settings.test.js``` file, will update the commit with file shortly and also fixing the formatting error.

File: src/cli_plugin/install/opensearch_dashboards.test.js [hunk 2]
```diff
@@ -95,38 +96,142 @@ describe('opensearchDashboards cli', function () {
...
+          };
 
-          expect(() => assertVersion(settings)).toThrowErrorMatchingInlineSnapshot(
-            `"Plugin foo [2.0.0] is incompatible with OpenSearch Dashboards [1.0.0]"`
+          expect(() => assertVersion(testSettings)).toThrowErrorMatchingInlineSnapshot(
+            `"Plugin foo [3.0.0] is incompatible with OpenSearch Dashboards [2.1.0]. Strict mode requires exact version match."`
           );
         });
+
+        // Tests for --single-version modes
```
huyaboo, line 172: Do we have a test in the case of a dash after the version? E.g. `2.17.2-some-suffix-3.3.1`

src/cli_plugin/install/index.js
```diff
@@ -62,6 +62,17 @@ export function installCommand(program) {
       'length of time before failing; 0 for never fail',
       parseMilliseconds
     )
+    .option(
+      '--single-version <mode>',
+      'set version validation method: "strict", "ignore"',
+      (value) => {
+        if (!['strict', 'ignore'].includes(value)) {
+          throw new Error(`Invalid single-version mode: ${value}. Use 'strict' or 'ignore'.`);
```
huyaboo, line 70: So we will only have exact version checks or none at all? Why did we remove the option to only check major version?

src/cli_plugin/install/opensearch_dashboards.js
```diff
@@ -55,9 +56,40 @@ export function assertVersion(settings) {
...
+      if (!semver.eq(actualVersion, expectedVersion)) {
+        throw new Error(
+          `Plugin ${settings.plugins[0].id} [${actual}] is incompatible with OpenSearch Dashboards [${expected}]. Strict mode requires exact version match.`
+        );
+      }
+      break;
+
+    case 'ignore':
+      if (actualVersion.major !== expectedVersion.major) {
+        console.warn(
```
huyaboo:
```suggestion
        logger.warn(
```

Probably better to add the logger as an arg here.

huyaboo (commented):
Left comments. Thanks for improving the platform build

src/cli_plugin/install/opensearch_dashboards.test.js
nvinjam-amazon: Yes, we already have couple of tests for this exact scenario.

src/cli_plugin/install/opensearch_dashboards.js
nvinjam-amazon: Let me update it, thanks for suggestion

src/cli_plugin/install/index.js
nvinjam-amazon: This was discussed during the meeting, Agreed to move with only 'strict' and 'ignore' options.

nvinjam-amazon: Please refer to this https://quip-amazon.com/eXODAlkvfO3R/OSD-Architecture-Review-Board-Agenda#temp:C:CNe96f462fab8004cab83cb1068e 

Thanks!!

AMoo-Miki:
We can certainly do better than just `strict` and `ignore`. A `loose` setting that allows a major version match would make this a lot safer for end users.

An even better solution would be to define a range in the plugin's `package.json` and match OSD's version against that; this can be achieved with just 2-3 changes. We could use a custom prop like `supportedOpenSearchDashboardsVersions` or a well-established standard of:
```
"peerDependencies": {
    "@opensearch-project/opensearch-dashboards": "^2.0.0 || ^3.0.0"
  },
```
... then simply expose it in `parse`: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/c20ba59215e8f0bb1d3ac19f85a6388dad864591/src/cli_plugin/install/settings.js#L65-L82

and then change the validation in https://github.com/opensearch-project/OpenSearch-Dashboards/blob/43fe5d50bb9bb1d16dc98640eff2468c567c0bc2/src/cli_plugin/install/opensearch_dashboards.js#L56-L62

... to use `semver.satisfies`.

src/cli_plugin/install/index.js
AMoo-Miki: Could you help me understand the reasoning behind limiting it to just 'strict' and 'ignore'? I am concerned that this binary approach oversimplifies the real-world usage.

PS, I posted a flexible solution below that can be implemented with minimal effort and maximum benefit.

src/cli_plugin/install/opensearch_dashboards.js
```diff
@@ -55,9 +56,40 @@ export function assertVersion(settings) {
...
+      if (!semver.eq(actualVersion, expectedVersion)) {
+        throw new Error(
+          `Plugin ${settings.plugins[0].id} [${actual}] is incompatible with OpenSearch Dashboards [${expected}]. Strict mode requires exact version match.`
+        );
+      }
+      break;
+
+    case 'ignore':
+      if (actualVersion.major !== expectedVersion.major) {
+        logger.log(
```
huyaboo, line 86:
```suggestion
        logger.warn(
```

huyaboo:
> We can certainly do better than just `strict` and `ignore`. A `loose` setting that allows a major version match would make this a lot safer for end users.
> 
> An even better solution would be to define a range in the plugin's `package.json` and match OSD's version against that; this can be achieved with just 2-3 changes. We could use a custom prop like `supportedOpenSearchDashboardsVersions` or a well-established standard of:
> 
> ```
> "peerDependencies": {
>     "@opensearch-project/opensearch-dashboards": "^2.0.0 || ^3.0.0"
>   },
> ```
> 
> ... then simply expose it in `parse`:
> 
> https://github.com/opensearch-project/OpenSearch-Dashboards/blob/c20ba59215e8f0bb1d3ac19f85a6388dad864591/src/cli_plugin/install/settings.js#L65-L82
> 
> and then change the validation in
> 
> https://github.com/opensearch-project/OpenSearch-Dashboards/blob/43fe5d50bb9bb1d16dc98640eff2468c567c0bc2/src/cli_plugin/install/opensearch_dashboards.js#L56-L62
> 
> ... to use `semver.satisfies`.

If we were to implement this, what would happen to existing plugins that haven't yet adopted this convention? Until each plugin adopts this, `strict` and `ignore` are the only options we can expose right? Not saying I disagree, but this can be done incrementally

huyaboo (dismissed):
LGTM

src/core/server/plugins/discovery/plugin_manifest_parser.ts
virajsanghvi: nit: can we not drop Version from this variable names?

virajsanghvi: If someone installed a plugin with ignore, which enables plugin versions to be able to do mv bumps on their own schedule, would this then cause an issue if its comparing version? Or does this only apply to discovering plugins as your comment suggests (and the assumption there is version should match)?

nvinjam-amazon: Yes, will update the variable names with ```coercedActualOpenSearchDashboardsVersion``` and ```coercedExpectedOpenSearchDashboardsVersion```

nvinjam-amazon:
This coercion only formats version strings for parsing, it doesn't validate compatibility. The ```--single-version ignore``` flag affects installation-time checks, while this code runs at plugin discovery/runtime. Plugins installed with ignore mode will work fine since coercion just standardizes the version format without rejecting based on version differences.

src/core/server/plugins/discovery/plugin_manifest_parser.ts
nvinjam-amazon: This coercion only formats version strings for parsing, it doesn't validate compatibility. The `--single-version ignore` flag affects installation-time checks, while this code runs at plugin discovery/runtime. Plugins installed with ignore mode will work fine since coercion just standardizes the version format without rejecting based on version differences.

huyaboo (approved):

src/cli_plugin/install/index.js
zhyuanqi: Yeah. Will we have a loose option of just make sure the major version is the same in the future?

nvinjam-amazon: Yes, `loose` option could be the long term solution and will be implemented as a default behavior in the future.

src/core/server/plugins/discovery/plugin_manifest_parser.ts
virajsanghvi: What was it the semver comparison for? It appears it throws an error when it fails: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10273/files#diff-3a59d77fcb2057189a459799ffe416a5885a09ff4d82848c9457289b26540a7eR208

I guess I'm just confused as to what this is specifically preventing.

nvinjam-amazon: This semver comparison prevents plugins from loading at runtime when they're incompatible with the current OpenSearch Dashboards version. It's a safety check that runs during plugin discovery - even if a plugin was installed with `--single-version ignore`, this prevents it from actually loading and potentially crashing the system if there's a version mismatch. It protects against scenarios like upgrading OpenSearch Dashboards but having old incompatible plugins still installed.

virajsanghvi: Doesn't this conflict with where we want to go? I thought goal of enabling ignoring versions is so plugins can version independently of OSD, but can communicate compatibility separately.

The default remains strict matching, so this doesn't necessarily have to be blocking to this PR, but are we tracking cleaning this up somewhere when we want to enable the above?

src/core/server/plugins/discovery/plugin_manifest_parser.test.ts
```diff
@@ -139,7 +138,7 @@ test('return error when plugin version is missing', async () => {
   });
 });
 
-test('return error when plugin expected OpenSearch Dashboards version is lower than actual version', async () => {
+test('return error when plugin expected OpenSearch Dashboards version has different major version', async () => {
```
ruanyl, line 141: Why change these unit tests? They should remain the same as the default behavior hasn't been changed, am I right?

src/core/server/plugins/discovery/plugin_manifest_parser.ts
```diff
@@ -286,19 +286,15 @@ function isVersionCompatible(
   }
 
   const coercedActualOpenSearchDashboardsVersion = coerce(actualOpenSearchDashboardsVersion);
-  if (coercedActualOpenSearchDashboardsVersion == null) {
```
ruanyl, line 289: Just wondering, is `coerce` a heavy function so it returns earlier before?

src/cli_plugin/install/index.js
kavilla: +1 @AMoo-Miki. The reason with not implementing loose from from the start (but eventually) was because loose began to become overly complicated. these two options accepts the risk that it could break but i think the experience for loose seemed to be not decided. 
> PS, I posted a flexible solution below that can be implemented with minimal effort and maximum benefit.

and where you talking about about the plugin matrix?

kavilla (approved):

virajsanghvi:
Have we looked into bundle-analyzer failure? Is it just failing generally atm? Don't think this should be impacted by this PR, but want to validate

nvinjam-amazon:
> Have we looked into bundle-analyzer failure? Is it just failing generally atm? Don't think this should be impacted by this PR, but want to validate

I'm not sure, why it is failing. it failed for 3 times and have to check with others regarding this.
cc: @kavilla @huyaboo

src/core/server/plugins/discovery/plugin_manifest_parser.ts
virajsanghvi: @nvinjam-amazon  Can you respond to this? Want to validate we're tracking this somewhere as our current story seems inconsistent

nvinjam-amazon: Hi @virajsanghvi, Will collaborate with @kavilla regarding this and create an action item for that. Might be better to create another issue for that. Will let you know the updates on that. Thanks!

nvinjam-amazon: If you want me to add anything to this, please let me know.

nvinjam-amazon:
> Have we looked into bundle-analyzer failure? Is it just failing generally atm? Don't think this should be impacted by this PR, but want to validate

Checked with @kavilla  regarding this, not an issue for this PR.

kavilla:
@virajsanghvi yeah, I called it out on the original PR for that workflow but essentially that will check the bundle size for each plugin. It has a baseline file and each PR would check the size of the bundle and diff the size within the PR and the baseline size. If the diff is bigger than ~5%. Then the assumption is the changes are too large. 

The original intent being that not to block large PRs but for explicity overriding this check if the changes were required to be that large. An example would be like a contributor builds a whole service to do what another service already provides but isn't well documented or exposed to the public.

The one issue though is that once maintainers approve the PR the baseline file is not updated but the actual bundle is getting bigger.

So then even though this PR could actually have a very small delta in size compared to the current mainline. It is still being compared to the last time the baseline was updated which might have been 10 commits. So in reality the check is comparing the bundle size of 10 PRs. So it can be ignored.

cc : @nvinjam-amazon

PR #10275, opened 2025-07-28
[Explore] Fix axes selector i18n message and throw error in embeddable

FriedhelmWS:
### Description

1. Add proper i18n translation for axes selector combo box field group label text
2. Throws an error when loading a saved visualization embeddable in the dashboard but cannot find a matching rule (in this case, no visualization can be created, so it's better to throw an early error to prevent unchecked errors).

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/visualizations/style_panel/axes/axes_selector.tsx
```diff
@@ -210,6 +210,25 @@ export const AxesSelectPanel: React.FC<AxesSelectPanelProps> = ({
...
+      case VisFieldType.Categorical:
+        return i18n.translate('explore.stylePanel.fieldType.categorical', {
+          defaultMessage: 'Categorical fields',
+        });
+      case VisFieldType.Date:
+        return i18n.translate('explore.stylePanel.fieldType.date', {
+          defaultMessage: 'Date fields',
+        });
+      default:
+        return `${type} fields`;
```
wanglam: Nit: Do we need a i18n for this?

src/plugins/explore/public/embeddable/explore_embeddable.tsx
```diff
@@ -357,7 +357,12 @@ export class ExploreEmbeddable
         };
       } else {
         const axesMapping = convertStringsToMappings(visualization.axesMapping, allColumns);
-        const matchedRule = findRuleByIndex(visualization.axesMapping, allColumns); // FIXME when no rule matched
+        const matchedRule = findRuleByIndex(visualization.axesMapping, allColumns);
+        if (!matchedRule) {
+          throw new Error(
```
ruanyl, line 362: Q: what will user see on the screen when throw?

FriedhelmWS: Screenshot 2025-07-29 at 09 56 07
The user will see the empty screen for the visualization with a console error. Ideally maybe we should also display the error on the UI but personally suggested this is not an urgent thing so we can leave it for now.

ruanyl (approved):

ruanyl (approved):

PR #10295, opened 2025-07-30
[Explore] Show results summary for default query

FriedhelmWS:
## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/results_summary/results_summary.tsx
```diff
@@ -40,7 +40,9 @@ export const ResultsSummary: React.FC<ResultsSummaryProps> = ({
   );
   const dataSetState = useSelector(selectDataset);
   const queryResults = useSelector(
-    (state: RootState) => state.results[state.query.query as string]?.hits?.hits
+    (state: RootState) =>
+      state.results[state.query.query as string]?.hits?.hits ||
+      Object.values(state.results)[0]?.hits?.hits // default query results
```
ruanyl: Q: Why the cache key is `0`? shouldn't it be the default query?

FriedhelmWS: I think it is pretty much equivalent. In the initial state, there will be only one result, which is the first item of that `state.results` list. Since there is no query string stored in the global state in the initial state, we cannot use `state.query.query` to get the result by key.

ruanyl: I'm not sure how the current "cache" works, does it reset on every new search? I think it's better to use the current default query of "Logs" tab which is more robust. The current code pretty much rely on the implementation details of the cache.

FriedhelmWS: Done, now if it cannot retrieve the query result using the query string from the state, it will use the default query of that data source.

kavilla:
do you mind checking out this one: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10269

and see if this has any issues with the summary. i am migrating the explore stuff to treat the query and dataset as the same thing since they are:

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/common/query/types.ts#L39

i have added a middleware file but not sure if that captured or broke your implementation

FriedhelmWS:
> do you mind checking out this one: #10269
> 
> and see if this has any issues with the summary. i am migrating the explore stuff to treat the query and dataset as the same thing since they are:
> 
> https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/common/query/types.ts#L39
> 
> i have added a middleware file but not sure if that captured or broke your implementation

Hi @kavilla, I think it should be ok as the summary component re-renders when the dataset changes so it always generate the summary base on the newest state. Let me wait until your PR get merged to see.

wanglam (approved):

PR #10296, opened 2025-07-30
[Explore] Discover visualization ux enhancement

ruanyl:
### Description
1. ensure minimum space between x-axis labels
2. auto focus on axis field selector when switching chart type
3. default to show table if cannot create auto chart

## Changelog

- fix: discover visualization ux enhancement and fix

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/embeddable/explore_embeddable.test.tsx
+++ b/src/plugins/explore/public/embeddable/explore_embeddable.test.tsx
@@ -185,6 +185,10 @@ describe('ExploreEmbeddable', () => {
     expect(embeddable.type).toBe(EXPLORE_EMBEDDABLE_TYPE);
   });
 
+  test('should have return inspector adaptors', () => {
+    expect(embeddable.getInspectorAdapters()).not.toBeUndefined();
+  });
+
   test('initializes search props correctly', () => {
     // @ts-ignore - accessing private property for testing
     const searchProps = embeddable.searchProps;


src/plugins/explore/public/components/visualizations/visualization_builder.ts
```diff
@@ -277,9 +275,15 @@ export class VisualizationBuilder {
           this.setStyles({ styles: visConfig?.ui.style.defaults, type: autoVis.chartType });
         }
       } else {
-        this.setCurrentChartType(undefined);
+        const visConfig = visualizationRegistry.getVisualizationConfig('table');
```
kavilla, line 278: should we have like a constant defined higher as like a `defaultVis`?

ruanyl: Thanks, I think it makes sense if the `defaultVis` is used in multiple places, since now it's only used once here, I will keep it this way

src/plugins/explore/public/embeddable/explore_embeddable.tsx
```diff
@@ -449,4 +449,8 @@ export class ExploreEmbeddable
     this.node = node;
     this.renderComponent(node, this.searchProps);
   }
+
+  public getInspectorAdapters() {
+    return this.inspectorAdaptors;
+  }
```
SuZhou-Joe, line 455: Do we still need this change? do not find code consuming the method.

src/plugins/explore/public/components/visualizations/style_panel/axes/axes_selector.tsx
```diff
@@ -82,6 +83,17 @@ export const AxesSelectPanel: React.FC<AxesSelectPanelProps> = ({
     }
   };
 
+  useEffect(() => {
+    // Make sure to initially focus on first axis field selector if current field selection is empty
+    if (isEmpty(currentMapping)) {
+      setTimeout(() => {
```
SuZhou-Joe, line 89: Nit: better to have some comment to elaborate the timeout.

src/plugins/explore/public/embeddable/explore_embeddable.tsx
ruanyl: Yes, this is need by embeddable to display "inspect" button on visualization

src/plugins/explore/public/components/visualizations/style_panel/axes/axes_selector.tsx
ruanyl: Would the comment on line 87 good enough?

SuZhou-Joe: I guess the timeout is waiting for input component mount completely. And the 500ms timeout may better to have some comment.

SuZhou-Joe (approved):

PR #10306, opened 2025-07-30
[Explore]bug fix: add one default range for metric + fix threshold in line-bar chart

Qxisylolo:
### Description
This pr addresses some bugs:
1. add one default range for metric
2. fix threshold in line-bar chart
3. add debouncing for angle options in scatter's exclusive panel
4. disable "Trucate after" option when disable the pie chart's label option

## Screenshot

https://github.com/user-attachments/assets/c753579c-c773-42cd-b6ec-f6b0a0489c30

截屏2025-07-31 07 41 10

## Changelog
- fix: add one default range for metric + fix threshold in line-bar chart

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/line/to_expression.ts
+++ b/src/plugins/explore/public/components/visualizations/line/to_expression.ts
@@ -176,6 +176,10 @@ export const createLineBarChart = (
     },
   };
 
+  const barWithThresholdLayer = {
+    layer: [barLayer],
+  };
+
   const lineLayer = {
     mark: buildMarkConfig(styles, 'line'),
     encoding: {
@@ -218,13 +222,13 @@ export const createLineBarChart = (
     },
   };
 
-  layers.push(barLayer, lineLayer);
-
   // Add threshold layer if enabled
   const thresholdLayer = createThresholdLayer(styles.thresholdLines, styles.tooltipOptions?.mode);
   if (thresholdLayer) {
-    layers.push(thresholdLayer);
+    barWithThresholdLayer.layer.push(thresholdLayer);
   }
+
+  layers.push(barWithThresholdLayer, lineLayer);
 
   // Add time marker layer if enabled
   const timeMarkerLayer = createTimeMarkerLayer(styles);
--- a/src/plugins/explore/public/components/visualizations/metric/metric_vis_config.ts
+++ b/src/plugins/explore/public/components/visualizations/metric/metric_vis_config.ts
@@ -23,6 +23,7 @@ export const defaultMetricChartStyles: MetricChartStyleControls = {
   fontSize: 60,
   useColor: false,
   colorSchema: ColorSchemas.BLUES,
+  customRanges: [{ min: 0, max: 100 }],
 };
 
 export const createMetricConfig = (): VisualizationType<'metric'> => ({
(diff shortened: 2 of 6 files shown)


src/plugins/explore/public/components/visualizations/scatter/scatter_exclusive_vis_options.tsx
```diff
@@ -27,6 +28,15 @@ export const ScatterExclusiveVisOptions = ({ styles, onChange }: ScatterVisOptio
     });
   };
 
+  const [pointAngle, handlePointAngle] = useDebouncedNumericValue(
+    styles.angle,
+    (val) => onChange({ ...styles, angle: val }),
+    {
+      min: 0,
```
kavilla, line 35: i see we define min max to the style accordian as well so if we had to change the min and max (even though unlikely) we will have to update it in two places.

wouldn't it be easier to have a higher level constants file or types or some factory, essentially a single place where a developer can see defined vis configs. and then within the component options.  we are just accessing the configs? 

that way we define things once and as a developer i can just look in one place to configure these values

src/plugins/explore/public/components/visualizations/metric/metric_vis_config.ts
```diff
@@ -23,6 +23,7 @@ export const defaultMetricChartStyles: MetricChartStyleControls = {
   fontSize: 60,
   useColor: false,
   colorSchema: ColorSchemas.BLUES,
+  customRanges: [{ min: 0, max: 100 }],
```
ruanyl, line 27: Is `100` percent?

src/plugins/explore/public/components/visualizations/scatter/scatter_exclusive_vis_options.tsx
ruanyl: @kavilla the `min` and `max` here are the parameter of `useDebouncedNumericValue`, not default vis styles

src/plugins/explore/public/components/visualizations/metric/metric_vis_config.ts
Qxisylolo: it's just a number. I refer to the legacy vis design, where the default range is [0, 10000]. I'm thinking it might be  more helpful to include a max value as well, instead of only specifying [{ min: 0 }]. 

截屏2025-07-31 08 15 59

FriedhelmWS: Could we add a comment here tho?

ruanyl (approved):

src/plugins/explore/public/components/visualizations/line/to_expression.ts
```diff
@@ -218,14 +222,14 @@ export const createLineBarChart = (
...
-  layers.push(barLayer, lineLayer);
-
   // Add threshold layer if enabled
   const thresholdLayer = createThresholdLayer(styles.thresholdLines, styles.tooltipOptions?.mode);
   if (thresholdLayer) {
-    layers.push(thresholdLayer);
+    barWithThresholdLayer.layer.push(thresholdLayer);
   }
 
+  layers.push(barWithThresholdLayer, lineLayer);
```
wanglam, line 231: Does this mean the thresholdLayer should be in front of barLayer and behind lineLayer? The original issue is that the thresholdLayer is in front of barLayer and lineLayer, which is incorrect.

Qxisylolo: The issue occurs because there are two y axes, and Vega-Lite doesn't know which one the thresholdLayer belongs to. the layer order is not the cause. 
barWithThresholdLayer is to group thresholdLayer and barLayer into the same layer and explicitly bind them to the same y axis.

PR #10311, opened 2025-07-31
[Explore] Fix Legend Visibility Logic for Visualizations

lunalzm:
### Description

This PR updates the legend visibility logic for line, area, bar, heatmap, and scatter chart visualizations. Previously, the LegendOptionsPanel appeared prematurely when more than two fields were available, even if only two axes were mapped or before a COLOR field was selected. The changes ensure that:

The legend options are hidden when only X and Y axes are mapped, regardless of the number of available fields.
The legend options appear only after a COLOR or FACET field is selected, as required for visualizations like grouped, faceted, or colored charts.

The updates modify the *vis_options.tsx files for each visualization type to check axisColumnMappings for COLOR or FACET fields, replacing previous logic based on available column counts. This improves the user experience by preventing irrelevant legend options from appearing until necessary.

### Before
image

### After
image
image

## Changelog
- fix: Disable "Show legend" when color unset

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/visualizations/scatter/scatter_vis_options.tsx
```diff
@@ -77,7 +80,7 @@ export const ScatterVisStyleControls: React.FC<ScatterVisStyleControlsProps> = (
           {shouldShowLegend && (
             <EuiFlexItem grow={false}>
               <LegendOptionsPanel
-                shouldShowLegend={true}
+                shouldShowLegend={shouldShowLegend}
```
wanglam, line 83: Nit: I think this value should always be `true` here, right? We don't need to modify this line. I'm OK to keep the current implementation the same with other places.

raintygao (approved):

src/plugins/explore/public/components/visualizations/scatter/scatter_vis_options.tsx
ruanyl: We may not need this prop any more, let's clean it up in followup PR

ruanyl (approved):

src/plugins/explore/public/components/visualizations/scatter/scatter_vis_options.tsx
lunalzm: I've fixed this issue in this PR. https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10358

PR #10314, opened 2025-07-31
[TEST] Fix shared links for PPL in explore

abbyhu2000:
### Description

Fix shared links for PPL in explore

## Changelog
- test: Fix shared links for PPL in explore

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/03/shared_links.spec.js
```diff
@@ -80,10 +79,11 @@ export const runSharedLinksTests = () => {
       indexPattern: INDEX_PATTERN_WITH_TIME_1,
       index: INDEX_WITH_TIME_1,
     }).forEach((config) => {
-      describe.skip(`${config.testName}`, () => {
+      describe(`${config.testName}`, () => {
         const queryString = getQueryString(config);
 
-        it(`should handle shared document links correctly for ${config.testName}`, () => {
+        // We do not have share single document or surrounding documents links for PPL in explore
```
kavilla, line 85: since these tests are source controlle should we just delete this test? or do we have an issue open of all the skip tests that we need to get back to and we can link it here as a todo?

abbyhu2000: yea we should just delete all the skip tests in explore. I will have another PR to delete all the skipped/unneeded tests

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/03/shared_links.spec.js
```diff
@@ -187,18 +179,18 @@ export const runSharedLinksTests = () => {
           openShareMenuWithRetry();
           cy.getElementByTestId('exportAsSavedObject').find('input').should('not.be.disabled');
```
kavilla, line 180: is the element actually called export as saved object? is it the same as discover? if so its confusing because we have the export saved object feature where as this i think it creates a saved object . or perhaps this could have a better test name describing what it does?

because shared link as snapshot could technically still have the saved object id if we created a shared link from a saved explore. but i think the difference between the two would be it will just load the saved object whereas a snapshot could be the saved object + current state. does this test end up passing even if state is included? same for snapshot test

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/03/shared_links.spec.js
```diff
@@ -141,17 +141,9 @@ export const runSharedLinksTests = () => {
...
           cy.explore.setQueryEditor(queryString, { parseSpecialCharSequences: false });
 
-          // Set filter for DQL/Lucene
-          if (config.hasDocLinks) {
-            cy.submitFilterFromDropDown(testData.filter[0], 'is', testData.filter[1], true);
-          }
-
           // Add fields from side panel
-          testData.fields.forEach((field, i) => {
+          testData.fields.forEach((field) => {
```
kavilla, line 145: tbh i think we should be sorting

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/03/shared_links.spec.js
abbyhu2000: yes it is the same as discover. the test id is called 'exportAsSnapshot' and 'exportAsSavedObject'
Screenshot 2025-07-31 at 4 21 34 PM
i agree they could be better named

abbyhu2000: yea this test only tests scenarios:
1. snapshot link captures current state
2. saved object link shared the saved explore 

it doesn't test: 
after we save as saved explore and we make changes and we share:
1. snapshot link should include the changes
2. saved object link should not include the unsaved changes

I could add a todo to add tests on the latter two @kavilla

LDrago27 (approved):

kavilla (approved):

Issue #10223, opened 2025-07-17
[Banner Plugin] Banner extensibility model Refactor


sidwang42:
## Description

The current banner plugin accesses its configuration via injectedMetadata.getPlugins() by filtering on a hardcoded plugin ID ('banner'). This introduces tight coupling between core and plugin implementation details, which breaks encapsulation and limits extensibility.

Key architectural concerns include:

Core code should not depend on plugin identifiers for behavior.

The design prevents alternative banner implementations without duplicating the same plugin ID, which is fragile.

This issue proposes refactoring both the plugin integration model to support runtime flexibility and UI consistency.

### Proposed Solution

[WIP]

### Tasks

[WIP]

PR #10324, opened 2025-08-03
[Banner Plugin] Refactor Banner Extensibility Model via Chrome Service

sidwang42:
## Summary
This PR refactors the banner plugin to decouple it from `injectedMetadata.getPlugins()` and remove the hardcoded dependency on the 'banner' plugin ID.
Instead, the banner is now managed via the Chrome Service, providing a cleaner and more extensible ownership model with minimal change to both core and plugin sides.

## Background
Previously, the banner plugin accessed its configuration by filtering `injectedMetadata.getPlugins()` for the 'banner' plugin ID.
### This approach introduced several issues:

- Tight coupling between core and plugin implementation.
- Fragile extensibility (alternative banner implementations required reusing the same plugin ID).
- Plugin‑specific mounting point logic embedded in OSD core.
## Refactor Approach
We evaluated three possible approaches:
- Overlay Service — treat banner as an overlay.
- Header Service — extend header service since the banner is visually part of the header.
- Chrome Service (Chosen) — extend the fundamental chrome service.
### Decision: Extend the Chrome Service.
Provides a central, consistent ownership model.
Avoids introducing a plugin‑specific mount point in core.
Minimizes code change across both plugin and core.
## Changelog
-feat: Refactor Banner Extensibility Model via Chrome Service
## Changes in this PR
Added setGlobalBanner API to Chrome Service for banner rendering.
Removed hardcoded injectedMetadata.getPlugins() dependency.
Updated the banner plugin to register via Chrome Service instead of metadata filtering.
## Related Issues
Fixes: [https://github.com/opensearch-project/OpenSearch-Dashboards/issues/10223](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/10223)
Listed in: [https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9990](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9990)


src/plugins/banner/public/plugin.test.ts
```diff
@@ -31,23 +34,27 @@ describe('BannerPlugin', () => {
...
 
   describe('stop', () => {
-    test('calls unmountBanner', () => {
+    test('banner is automatically unmounted when plugin stops', () => {
       const plugin = new BannerPlugin();
       plugin.stop();
 
-      expect(unmountBannerMock).toHaveBeenCalledTimes(1);
+      // No explicit assertion needed as the banner is automatically unmounted
+      // when the plugin is stopped, as noted in the implementation comment
```
zhongnansu: remove?

src/plugins/banner/public/plugin.ts
```diff
@@ -21,13 +22,16 @@ export class BannerPlugin implements Plugin<BannerPluginSetup, BannerPluginStart
...
+      component: React.createElement(GlobalBanner, { http: core.http }),
+    });
 
     return {};
   }
 
   public stop() {
-    unmountBanner();
+    // The banner will be automatically removed when the plugin is stopped
+    // because the component will be unmounted
```
zhongnansu: do we need the comment?

src/plugins/banner/public/plugin.test.ts
virajsanghvi: Also curious, what is purpose of this test if its not asserting anything? That nothing is thrown?

src/plugins/banner/public/plugin.ts
virajsanghvi: What causes the component to be unmounted on stop?

sidwang42: > do we need the comment?



> What causes the component to be unmounted on stop?


Agreed — I’ll remove these. They were left over from the previous design where we explicitly had an unmountBanner function. Since the Chrome service now manages the interface, a manual unmount isn’t needed. If unmounting is required in the future, it should be handled through the Chrome service. That makes the current stop logic here unnecessary.

ruanyl:
I feel it's better fit Overlay Service consider we have existing banner service there already, I'd extend the existing banner service. Any particular reason of adding this to chrome service?

src/core/public/chrome/ui/header/header.tsx
```diff
@@ -669,7 +663,7 @@ export function Header({
 
   return (
     <>
-      {isBannerEnabled && <div id="pluginGlobalBanner" />}
+      {globalBanner && <div className="globalBanner">{globalBanner.component}</div>}
```
ruanyl: I'd not add the mount point to header, they are logically distinct elements with separate concerns and responsibilities. The rendering_service would be a more appropriate location for the mount point.

sidwang42:
> I feel it's better fit Overlay Service consider we have existing banner service there already, I'd extend the existing banner service. Any particular reason of adding this to chrome service?

Thanks for the feedback, @ruanyl.

Extending the current banner service wasn’t considered because the existing banner service is not designed for a persistent global header banner — it’s scoped to temporary overlays. The global banner directly impacts header layout, including adjustments to the osdHeaderOffset, which makes it functionally tied to the header.

While the rendering service could technically host the mount point, placing it in chrome/header keeps the logic consistent with its visual and functional behavior as part of the header rather than a transient overlay. This ensures layout adjustments remain centralized and avoids coupling with rendering service concerns.

ruanyl:
> The global banner directly impacts header layout, including adjustments to the osdHeaderOffset, which makes it functionally tied to the header.

Hi @sidwang42, thanks, could you elaborate why this makes it functionally tied to the header? My understanding is it affects the page content offset(app-wrapper), doesn't seem it's functionally tied to header.

sidwang42:
> > The global banner directly impacts header layout, including adjustments to the osdHeaderOffset, which makes it functionally tied to the header.
> 
> Hi @sidwang42, thanks, could you elaborate why this makes it functionally tied to the header? My understanding is it affects the page content offset(app-wrapper), doesn't seem it's functionally tied to header.

Hi @ruanyl, great question — happy to clarify.

The global banner is visually and structurally fixed to the top of the page, spanning full width and sitting above the app content — much like the header itself. This fixed position means it's always visible and pushes the page content down, just like a header does.

From a code perspective, it integrates into the layout through the osdHeaderOffset variable, which is used specifically for adjusting the header and layout spacing. The banner adds to this offset, not just affecting app-wrapper spacing arbitrarily, but working with the header’s layout logic.

In contrast, overlay service components (like flyouts and modals) don’t participate in layout flow. They float over the app without modifying spacing or offsets — their z-index simply stacks them above existing UI. That’s fundamentally different from what the banner does.

So while the banner doesn’t live inside the` <Header>` component, its behavior aligns closely with it — both in visual placement and in how it affects the overall page structure. That’s why I believe it's more functionally tied to the header than an overlay.

Happy to discuss further if needed!

zhongnansu (approved):

src/core/public/chrome/ui/header/header.tsx
zhongnansu: @sidwang42 Can we make this  header banner JSX a react component, and import here for better separation and extensibility?

ruanyl:
@sidwang42 Can you dismiss the global banner? Are there screenshots to share?

sidwang42:
> @sidwang42 Can you dismiss the global banner? Are there screenshots to share?

Yes, the banner can be dismissed. Here are the screenshots:
image

image

src/core/public/chrome/ui/header/header.tsx
sidwang42: > @sidwang42 Can we make this header banner JSX a react component, and import here for better separation and extensibility?

@zhongnansu 
Yeah, that makes a lot of sense. After checking the repo, I noticed similar patterns used for breadcrumbs, controls extension, and the help menu—all implemented as separate React components. I’ve updated this PR to follow that convention. Thanks for pointing it out!

zhongnansu (approved):

PR #10337, opened 2025-08-05
[Explore] Refactor generated query to have max height and scrollable

angle943:
### Description

- before, the generated query had a line-height set to 1, and it would show ellipses if overflowing. Now we are going to make it scrollable if needed, and never ellipses:

Screenshot 2025-08-04 at 6 38 20 PM

Screenshot 2025-08-04 at 6 38 39 PM

## Changelog
- refactor: change generated query UI

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/query_panel/query_panel_generated_query/query_panel_generated_query.tsx
+++ b/src/plugins/explore/public/components/query_panel/query_panel_generated_query/query_panel_generated_query.tsx
@@ -6,14 +6,14 @@
 import React from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { i18n } from '@osd/i18n';
-import { EuiBadge, EuiIcon, EuiText } from '@elastic/eui';
+import { EuiButtonEmpty, EuiIcon, EuiText } from '@elastic/eui';
 import { selectLastExecutedTranslatedQuery } from '../../../application/utils/state_management/selectors';
 import { useEditorFocus, useSetEditorTextWithQuery } from '../../../application/hooks';
 import { clearLastExecutedData } from '../../../application/utils/state_management/slices';
 import './query_panel_generated_query.scss';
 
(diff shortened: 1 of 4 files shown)


LDrago27 (approved):

src/plugins/explore/public/components/query_panel/query_panel_generated_query/query_panel_generated_query.tsx
```diff
@@ -6,14 +6,14 @@
...
-import { EuiBadge, EuiIcon, EuiText } from '@elastic/eui';
+import { EuiButtonEmpty, EuiIcon, EuiText } from '@elastic/eui';
 import { selectLastExecutedTranslatedQuery } from '../../../application/utils/state_management/selectors';
 import { useEditorFocus, useSetEditorTextWithQuery } from '../../../application/hooks';
 import { clearLastExecutedData } from '../../../application/utils/state_management/slices';
 import './query_panel_generated_query.scss';
 
 const editQueryText = i18n.translate('explore.queryPanel.queryPanelGeneratedQuery.editQuery', {
-  defaultMessage: 'Edit query',
+  defaultMessage: 'Replace query',
```
ruanyl, line 16: Nit: should the key `explore.queryPanel.queryPanelGeneratedQuery.editQuery` be updated accordingly?

angle943: I thought about it, but decided against it. The `dataTestId` also has the name `EditQuery`, and if I rename this, I think that should be renamed as well. But then all of our tests will also need updating. And next week product may decide to rename this label as "Move this Query to the editor", or so on.

I think no matter what we decide to label this to the end users, programattically `Edit query` makes the most sense in the codebase. So I think its worth keeping it as `Edit Query` even if this label changes several times in the future

PR #10340, opened 2025-08-05
[Explore] Fix EuiSelect Issue in Firefox Browser

lunalzm:
### Description

This pull request fixes an issue where the EuiSelect component options could not be selected in the Firefox browser. Changing the `value` prop to `defaultValue` resolves the issue.

### Before
euiSelect_before

### After

https://github.com/user-attachments/assets/48bb0510-6b49-49c4-8bd8-0ec0a6572b7d

## Testing the changes

1. Open the application in Firefox.
2. Navigate to Discover
3. Run a PPL query and select "Visualization" tab
4. Observe "Legend" options and select one of them
4. Test the same functionality in Chrome to ensure no regression.

## Changelog
- fix: Fix EuiSelect Issue in Firefox Browser

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ruanyl (approved):

FriedhelmWS (approved):

raintygao (dismissed):

SuZhou-Joe:
Nit: if it is a common issue shall we raise an issue in https://github.com/opensearch-project/oui ?

src/plugins/explore/public/components/chart/timechart_header/timechart_header.tsx
```diff
@@ -181,7 +181,7 @@ export function TimechartHeader({
                       label: display,
                     };
                   })}
-                value={interval}
+                defaultValue={interval}
```
SuZhou-Joe: defaultValue usually means the component won't be a controlled component, will this change introduce any regression? Like will there be other place that may mutate the state?

src/plugins/explore/public/components/chart/timechart_header/timechart_header.tsx
SuZhou-Joe, line 1: The state management looks weird to me actually, we are trying to sync props.stateInterval and interval state, why don't we just use props.stateInterval?

src/plugins/explore/public/components/visualizations/style_panel/axes/standard_axes_options.tsx
```diff
@@ -210,7 +210,7 @@ export const AllAxesOptions: React.FC<AllAxesOptionsProps> = ({
                       >
                         <EuiSelect
                           compressed
-                          value={
+                          defaultValue={
```
wanglam: What about set selected in option item instead pass `defaultValue`? The code would like
```
options={[
                        {
                          value: 'horizontal',
                          text: 'Horizontal',
                          selected: axis.labels.rotate === 0,
                        },
                        {
                          value: 'vertical',
                          text: 'Vertical',
                          selected: axis.labels.rotate === -90,
                        },
                        { value: 'angled', text: 'Angled', selected: axis.labels.rotate === -45 },
                      ]}
```
I'm not sure if `defaultValue` will be updated if state change from outside and component doesn't mount/unmount.

lunalzm: Ok, I will add "key" to `<EuiSelect>`

src/plugins/explore/public/components/chart/timechart_header/timechart_header.tsx
lunalzm: Ok, I will add "key" to `<EuiSelect>`

wanglam (approved):

ruanyl (dismissed):

src/plugins/explore/public/components/visualizations/style_panel/axes/standard_axes_options.tsx
```diff
@@ -209,8 +209,15 @@ export const AllAxesOptions: React.FC<AllAxesOptionsProps> = ({
                         })}
                       >
                         <EuiSelect
+                          key={
+                            axis.labels.rotate === 0
+                              ? 'horizontal'
+                              : axis.labels.rotate === -90
+                              ? 'vertical'
+                              : 'angled'
```
ruanyl: Can we extract this logic? I can see the same code for `defaultValue`

lunalzm: Sure!

ruanyl (approved):

AMoo-Miki (changes requested):
Using `defaultValue` instead of `value` fundamentally changes the component's behavior. `defaultValue` only sets the initial value when the component first mounts - it cannot be updated programmatically afterward. If the user changes the value and then the code attempts to change it later, the programmatic change will be ignored and not reflected in the UI.
* `value` creates a controlled component where state in React drives the value, allowing programmatic updates.
* `defaultValue` creates an uncontrolled component where DOM manages the state after initial render.

Using `defaultValue` in place of `value` when we need programmatic control is architecturally incorrect and will break expected functionality.


Additionally, if this is an OUI problem (which I am pretty sure it is not), the problem should be fixed there.

lunalzm:
@AMoo-Miki 

#### Analysis:
The root cause of the issue is that in Firefox, selecting an option from the `<EuiSelect>` triggers a `mouseup` event that bubbles up to the parent `<EuiResizableContainer>`. That container listens for all `mouseup` events and calls its internal `reset()` logic, which forces a re-render and re-initializes state from the old `stateInterval` value. 
```
/oui/src/components/resizable_container/resizable_container.tsx

const onMouseUp = useCallback(() => {
    actions.reset();
  }, [actions]);
 ```
The reset itself doesn’t directly affect the `EuiSelect` or its state, but the re-render triggered by the state change (e.g., isDragging changing) can lead to re-mount or re-render with outdated props, exacerbating the issue in Firefox due to its stricter `<select>` rendering behavior.
As a result, the newly selected option is immediately replaced by the original value. Chrome does not bubble the same event in this way, so it works fine there.

#### Fix:
Prevent the `mouseup` event from bubbling to the resizable container by adding an explicit handler to the `<EuiSelect>`:
```
<EuiSelect
  ...
  onMouseUp={(e) => e.stopPropagation()}
/>
```
This stops the event before it reaches the container, so Firefox no longer triggers the unintended `reset()` and the selected option persists correctly.

wanglam (dismissed):

lunalzm:
> Nit: if it is a common issue shall we raise an issue in https://github.com/opensearch-project/oui ?

@SuZhou-Joe Now I have found the root cause and can solve this bug by preventing the event from bubbling to the resizable container. However, the core problem is that the resizable container should not be reset when not dragging. Should we file an issue with OUI?

wanglam:
> @AMoo-Miki
> 
> The root cause of the issue is that in Firefox, selecting an option from the `<EuiSelect>` triggers a `mouseup` event that bubbles up to the parent `<EuiResizableContainer>`. That container listens for all `mouseup` events and calls its internal `reset()` logic, which forces a re-render and re-initializes state from the old `stateInterval` value.
> 
> ```
> /oui/src/components/resizable_container/resizable_container.tsx
> 
> const onMouseUp = useCallback(() => {
>     actions.reset();
>   }, [actions]);
> ```
> 
> The reset itself doesn’t directly affect the `EuiSelect` or its state, but the re-render triggered by the state change (e.g., isDragging changing) can lead to re-mount or re-render with outdated props, exacerbating the issue in Firefox due to its stricter `<select>` rendering behavior. As a result, the newly selected option is immediately replaced by the original value. Chrome does not bubble the same event in this way, so it works fine there.
> 
> #### Fix:
> Prevent the `mouseup` event from bubbling to the resizable container by adding an explicit handler to the `<EuiSelect>`:
> 
> ```
> <EuiSelect
>   ...
>   onMouseUp={(e) => e.stopPropagation()}
> />
> ```
> 
> This stops the event before it reaches the container, so Firefox no longer triggers the unintended `reset()` and the selected option persists correctly.

From my perspective, the `onMouseUp` event handler should follow the same pattern as `onMouseMove` in the `OuiResizableContainer` component, which should only bind when `isDragging` is true. Maybe we can raise a PR to fix this behavior in the OUI project. For a quick fix in OSD, I'm OK with adding `stopPropagation()` to the onMouseUp handler in the `EuiSelect` component.

ruanyl (dismissed):

ruanyl:
@AMoo-Miki can you take another look please? need your approval to unblock

lunalzm:
Hi @AMoo-Miki, can you help review this PR? If there is no problem, please approve it.

PR #10351, opened 2025-08-06
[Keyboard-shortcuts] Add keyboard shortcuts foundation and architecture

natel926:
### Description
This PR implements Phase 1 of the keyboard shortcuts system, providing the foundational
architecture for plugins to register and manage keyboard shortcuts in OpenSearch Dashboards.
## Changelog
feat: Add keyboard shortcuts foundation and core service architecture
### Check List
- [ ] All tests pass
  - [ ] `yarn test:jest`


config/opensearch_dashboards.yml
```diff
@@ -384,4 +384,11 @@
...
-# banner.content: "This is an important announcement for all users. [Learn more](https://opensearch.org) about OpenSearch."
\ No newline at end of file
+# banner.content: "This is an important announcement for all users. [Learn more](https://opensearch.org) about OpenSearch."
+
+opensearch.ignoreVersionMismatch: true
+data_source.enabled: true
+workspace.enabled: true
+data:
+  savedQueriesNewUI:
+    enabled: true
```
angle943: lets remove these changes

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { KeyboardShortcutsSetup, KeyboardShortcutsStart, ShortcutDefinition } from './types';
```
angle943: nitpick: lets call it (and file names and everything) `KeyboardShortcut`, not `KeyboardShortcuts`

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+
+/** @internal */
```
angle943: lets use

```
/**
 * @internal
*/
```

Its in the JSDoc format: https://jsdoc.app/about-getting-started

Also, lets add `@experimental`. This way we can change the API in the near future without worrying that it is a breaking change

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+export class KeyboardShortcutsService {
+  // Phase 1.6: Basic Registration System - Simple shortcut storage
```
angle943: lets remove all these phase comments

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+  private shortcuts = new Map<string, ShortcutDefinition>();
+  private isListening = false;
```
angle943: do we need this? can we just start listening and stop listening in the start() and stop()?

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  // Phase 1.2: Basic Service Class Structure
+  public setup(): KeyboardShortcutsSetup {
+    return {
+      register: (shortcuts) => this.register(shortcuts),
```
angle943: nitpick:

`register: this.register,`

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
angle943: nvm i think its safe to have to avoid possible double event listeners

src/core/public/keyboard_shortcuts/types.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export interface ShortcutDefinition {
+  id: string;
+  pluginId: string;
+  name: string;
+  category: string;
+  keys: string;
+  execute: ShortcutHandler;
+}
```
angle943: lets add comments using JS docs to these, see: https://github.com/angle943/OpenSearch-Dashboards/blob/ef3a9c07a07ffb7f7bfa662d29c88ed487b7e187/src/core/types/app_category.ts

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
angle943: nvm part 2: since its passing a reference when we start event listener, it will never be registered twice. There will always be one event listener no matter how many times you call `start`

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+        shortcut.execute(event);
+      } catch (error) {
+        // eslint-disable-next-line no-console
+        console.error(`Error executing keyboard shortcut ${shortcut.id}:`, error);
+      }
+    }
+  };
+
+  // Phase 1.4: Basic Event Listener Setup - Global event handling
+  private startEventListening(): void {
```
angle943: lets call this `startEventListener()`

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+    }${event.metaKey ? 'cmd+' : ''}${event.key.toLowerCase()}`;
+
+    const shortcut = this.shortcuts.get(eventKeyString);
+
+    if (shortcut) {
+      event.preventDefault();
+      event.stopPropagation();
+
+      try {
+        shortcut.execute(event);
```
angle943: do plugin devs who are registering keyboard events need this `event` object?

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+    this.shortcuts.clear();
+  }
+
+  // Phase 1.6: Basic Registration System
+  private register(shortcuts: ShortcutDefinition[]): string[] {
+    const registeredIds: string[] = [];
+
+    shortcuts.forEach((shortcut) => {
+      // Construct full ID: shortcut.id + "." + shortcut.pluginId
+      const fullId = `${shortcut.id}.${shortcut.pluginId}`;
```
angle943: lets create a function for this, like:

```
private getNamespacedIdForKeyboardShortcut = (shortcut: Pick<ShortcutDefinition, 'id' | 'pluginId'>) => `${shortcut.id}.${shortcut.pluginId}`;
```

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+    this.stopEventListening();
+    this.shortcuts.clear();
+  }
+
+  // Phase 1.6: Basic Registration System
+  private register(shortcuts: ShortcutDefinition[]): string[] {
+    const registeredIds: string[] = [];
+
+    shortcuts.forEach((shortcut) => {
+      // Construct full ID: shortcut.id + "." + shortcut.pluginId
```
angle943: unnecessary comment

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+      const fullId = `${shortcut.id}.${shortcut.pluginId}`;
+
+      // Basic storage - no validation in Phase 1
+      this.shortcuts.set(shortcut.keys.toLowerCase(), shortcut);
```
angle943: make a function instead of `shortcut.keys.toLowerCase()`, and for phase 1 you can just have that lower case it

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+      registeredIds.push(fullId);
+    });
+
+    return registeredIds;
```
angle943: why does this need to return this registered ids?

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+    };
+  }
+
+  public stop() {
+    this.stopEventListening();
+    this.shortcuts.clear();
+  }
+
+  // Phase 1.6: Basic Registration System
+  private register(shortcuts: ShortcutDefinition[]): string[] {
```
angle943: what do you think about the argument being a single one, ie:

`private register(shortcut: ShortcutDefinition)`

if plugin dev need to register more than one, they can just call it multiple times

angle943: the issue i see with this is that if plugin dev needs to register one, they always need to wrap it in an array. we can technically make this method handle both array and single instance, but thats just overkill i think

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { KeyboardShortcutsSetup, KeyboardShortcutsStart, ShortcutDefinition } from './types';
+
+/** @internal */
+export class KeyboardShortcutsService {
+  // Phase 1.6: Basic Registration System - Simple shortcut storage
+  private shortcuts = new Map<string, ShortcutDefinition>();
```
angle943: i thought we agreed on `ShortcutDefinition[]` for the value?

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+      // Basic storage - no validation in Phase 1
+      this.shortcuts.set(shortcut.keys.toLowerCase(), shortcut);
+      registeredIds.push(fullId);
+    });
+
+    return registeredIds;
+  }
+
+  private unregister(fullId: string): void {
+    // Find and remove shortcut by full ID
```
angle943: unnecessary comment

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+    for (const [key, shortcut] of this.shortcuts.entries()) {
+      if (`${shortcut.id}.${shortcut.pluginId}` === fullId) {
```
angle943: if you make the `getNamespacedIdForKeyboardShortcut` i mentioned in the other comment, you can just call it here. That way if you ever need to change this in the future, you can do it in one place

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  // Phase 1.4: Basic Event Listener Setup - Global event handling
+  private startEventListening(): void {
+    if (this.isListening) return;
+
+    document.addEventListener('keydown', this.handleKeyboardEvent, true);
+    this.isListening = true;
+  }
+
+  private stopEventListening(): void {
```
angle943: stopEventListener

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+        console.error(`Error executing keyboard shortcut ${shortcut.id}:`, error);
+      }
+    }
+  };
+
+  // Phase 1.4: Basic Event Listener Setup - Global event handling
+  private startEventListening(): void {
+    if (this.isListening) return;
+
+    document.addEventListener('keydown', this.handleKeyboardEvent, true);
```
angle943: can you figure out what this third argument does, and see if it is needed or not?

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+        return;
+      }
+    }
+  }
+
+  // Phase 1.5: Input Context Filtering - Text input detection
+  private isTextInputActive(target: EventTarget | null): boolean {
+    if (!target || !(target as HTMLElement).tagName) return false;
+
+    const element = target as HTMLElement;
```
angle943: instead of doing type coercion, lets do this: https://stackoverflow.com/a/50326668

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+    const tagName = element.tagName.toLowerCase();
+
+    // Check for input elements
+    if (tagName === 'input') {
+      const inputType = (element as HTMLInputElement).type.toLowerCase();
+      const textInputTypes = ['text', 'search', 'url', 'tel', 'email', 'password'];
+      return textInputTypes.includes(inputType);
+    }
```
angle943: why can't we return true for all tagName === 'input'`?

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+      }
+    }
+  }
+
+  // Phase 1.5: Input Context Filtering - Text input detection
+  private isTextInputActive(target: EventTarget | null): boolean {
+    if (!target || !(target as HTMLElement).tagName) return false;
+
+    const element = target as HTMLElement;
+    const tagName = element.tagName.toLowerCase();
```
angle943: lets not toLowerCase it. Since browsers capitalize all by convention, lets follow that convention in the code

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
natel926: The third argument is about useCapture.

useCapture: false(default):

Example 

1. User presses `Ctrl+S`
2. Input field handles event first
3. Input calls `event.stopPropagation()`
4. Shortcut service never receives the event
5. Save shortcut doesn't work

With useCapture: true:

1. User presses `Ctrl+S`
2. Shortcut service handles event first
3. Service calls `event.preventDefault()` and `event.stopPropagation()`
4. Save shortcut works, input doesn't get the event

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+      const inputType = (element as HTMLInputElement).type.toLowerCase();
+      const textInputTypes = ['text', 'search', 'url', 'tel', 'email', 'password'];
+      return textInputTypes.includes(inputType);
+    }
+
+    // Check for other text input elements
+    if (tagName === 'textarea') return true;
+
+    // Check for contenteditable elements
+    if (element.contentEditable === 'true') return true;
```
angle943: what is this for?

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+    for (const [key, shortcut] of this.shortcuts.entries()) {
+      if (`${shortcut.id}.${shortcut.pluginId}` === fullId) {
+        this.shortcuts.delete(key);
+        return;
+      }
+    }
+  }
+
+  // Phase 1.5: Input Context Filtering - Text input detection
+  private isTextInputActive(target: EventTarget | null): boolean {
```
angle943: to future proof this, how about we rename this to `shouldIgnoreKeyboardEventForTarget()`?

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  // Phase 1.7: Global Shortcut Execution - Basic shortcut execution
+  private handleKeyboardEvent = (event: KeyboardEvent): void => {
+    // Phase 1.5: Skip if typing in text input
+    if (this.isTextInputActive(event.target)) {
+      return;
+    }
+
+    // Basic key matching - no normalization in Phase 1
+    const eventKeyString = `${event.ctrlKey ? 'ctrl+' : ''}${event.altKey ? 'alt+' : ''}${
```
angle943: lets extract this to a method.

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+      event.shiftKey ? 'shift+' : ''
+    }${event.metaKey ? 'cmd+' : ''}${event.key.toLowerCase()}`;
+
+    const shortcut = this.shortcuts.get(eventKeyString);
+
+    if (shortcut) {
+      event.preventDefault();
```
angle943:
```
Whether or not to use e.preventDefault() and e.stopPropagation() with document.addEventListener() depends entirely on the desired behavior for the specific event listener.
e.preventDefault():
Purpose: Prevents the browser's default action for a particular event.
When to use with document.addEventListener(): Use it when you are listening for an event on the document and want to override the browser's default handling of that event.
Example: Preventing the default context menu from appearing on a right-click anywhere on the page:
JavaScript

    document.addEventListener('contextmenu', function(event) {
        event.preventDefault();
        // Custom logic for right-click
    });
e.stopPropagation():
Purpose: Prevents the event from bubbling up or capturing down the DOM tree.
When to use with document.addEventListener(): Use it when you are listening for an event on the document and want to prevent that event from reaching other event listeners on parent or child elements. This is less common for document listeners compared to element-specific listeners, as the document is often the highest level.
Example: If you had a global keyboard shortcut on the document, and you wanted to prevent it from triggering if the user was typing in an input field:
JavaScript

    document.addEventListener('keydown', function(event) {
        if (event.target.tagName === 'INPUT' || event.target.tagName === 'TEXTAREA') {
            event.stopPropagation(); // Stop propagation if focus is on an input
        }
        // Global keyboard shortcut logic
    });
In summary:
Use e.preventDefault() when you want to stop the default browser action for an event on the document.
Use e.stopPropagation() when you want to control the flow of the event through the DOM, preventing it from reaching other listeners.
You should always consider the specific interaction and the potential side effects before employing either method, especially stopPropagation(), as it can sometimes lead to unexpected behavior if not carefully managed.
```

- e.preventDefault() is necessary i think as we want to override browser's default behavior. But lets add a comment for that
- e.stopPropagation() i don't think is necessary. lets remove it until we realize it becomes a problem

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
```diff
@@ -0,0 +1,132 @@
...
+    this.isListening = true;
+  }
+
+  private stopEventListening(): void {
+    if (!this.isListening) return;
+
+    document.removeEventListener('keydown', this.handleKeyboardEvent, true);
+    this.isListening = false;
+  }
+}
```
angle943: letes add a test file for this

src/plugins/dashboard/public/plugin.tsx
```diff
@@ -608,6 +608,34 @@ export class DashboardPlugin
...
+              return;
+            }
+
+            // eslint-disable-next-line no-console
+            console.log('Pressed shift+d - Navigating to Dashboard!');
+            core.application.navigateToApp('dashboards');
+          },
+        },
+      ]);
+    }
```
angle943: lets not add this as part of this PR

src/plugins/discover/public/plugin.ts
```diff
@@ -418,6 +418,34 @@ export class DiscoverPlugin
...
+              return;
+            }
+
+            // eslint-disable-next-line no-console
+            console.log('Pressed shift+d - Navigating to Discover!');
+            core.application.navigateToApp('discover');
+          },
+        },
+      ]);
+    }
```
angle943: lets not add this as part of this PR

src/plugins/visualize/public/plugin.ts
```diff
@@ -297,6 +297,34 @@ export class VisualizePlugin
...
+              return;
+            }
+
+            // eslint-disable-next-line no-console
+            console.log('Pressed shift+g - Navigating to visualization!');
+            core.application.navigateToApp('visualize');
+          },
+        },
+      ]);
+    }
```
angle943: lets not add this as part of this PR

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
natel926: correct me if I'm wrong. I think register: this.register creates a "detached" function that loses its connection to the service instance, causing runtime errors when the method tries to access this.shortcuts.

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
natel926: Yes it should be ShortcutDefinition[]

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
natel926: I agree, the array wrapping is a minor inconvenience, but the overloaded approach would add method complexity and runtime type checking.

src/core/public/keyboard_shortcuts/keyboard_shortcuts_service.ts
angle943: did you run into errors?

config/opensearch_dashboards.yml
```diff
@@ -384,4 +384,4 @@
 # banner.enabled: true
 
 # Set the content of the banner
-# banner.content: "This is an important announcement for all users. [Learn more](https://opensearch.org) about OpenSearch."
```
angle943: can we remove this change too. i think you added a whitespace. This file should not be part of this PR

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,129 @@
...
+    this.stopEventListener();
+    this.shortcuts.clear();
+  }
+
+  private normalizeKeyboardShortcutString = (str: string): string => str.toLowerCase();
+
+  private getNamespacedIdForKeyboardShortcut = (
+    shortcut: Pick<ShortcutDefinition, 'id' | 'pluginId'>
+  ) =>
+    `${this.normalizeKeyboardShortcutString(shortcut.id)}.${this.normalizeKeyboardShortcutString(
```
angle943: i don't think this function should call `normalizeKeyboardShortcutString`. that function is going to have logic later like putting `+` and everything, that is irrelevant for `getNamespacedIdForKeyboardShortcut

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  public stop() {
+    this.stopEventListener();
+    this.shortcuts.clear();
+  }
+
+  private normalizeKeyboardShortcutString = (str: string): string => str.toLowerCase();
+
+  private getNamespacedIdForKeyboardShortcut = (
```
angle943: `getNamespacedId` (we are in the keyboardShortcutService so don't need to say keyboardShortcut again)

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,129 @@
...
+      unregister: (id) => this.unregister(id),
+    };
+  }
+
+  public stop() {
+    this.stopEventListener();
+    this.shortcuts.clear();
+  }
+
+  private normalizeKeyboardShortcutString = (str: string): string => str.toLowerCase();
```
angle943: how about `getNormalizedKey`

I think `Key` is sufficient because thats also what vscode calls it: https://code.visualstudio.com/docs/configure/keybindings#_keyboard-rules

src/core/public/keyboard_shortcut/keyboard_shortcut_service.test.js
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { keyboardShortcutServiceMock } from './keyboard_shortcut_service.mock';
```
angle943: lets test the actual service, not the mock

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  private getNamespacedIdForKeyboardShortcut = (
+    shortcut: Pick<ShortcutDefinition, 'id' | 'pluginId'>
+  ) =>
+    `${this.normalizeKeyboardShortcutString(shortcut.id)}.${this.normalizeKeyboardShortcutString(
+      shortcut.pluginId
+    )}`;
+
+  private getEventKeyString = (event: KeyboardEvent): string => {
```
angle943, line 48: how about `getNormalizedKeyFromKeyboardEvent`.

But also, i see why we need two functions to normalize... One we need to normalize for developeres who are registering a keyboard shortcut, and one we need to normalize for the keyboard event. Is there a good way to reduce code duplication between the two?

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,129 @@
...
+    return `${event.ctrlKey ? 'ctrl+' : ''}${event.altKey ? 'alt+' : ''}${
+      event.shiftKey ? 'shift+' : ''
+    }${event.metaKey ? 'cmd+' : ''}${this.normalizeKeyboardShortcutString(event.key)}`;
```
angle943: for readability can we build a string instead?

let key = '';
if (event.ctrlKey) {
    key += 'shift+';
 }


etc

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,129 @@
...
+  };
+
+  private register(shortcuts: ShortcutDefinition[]): void {
```
angle943: lets have this register a single one? i think i left a comment about this before

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,129 @@
...
+    shortcuts.forEach((shortcut) => {
+      const key = this.normalizeKeyboardShortcutString(shortcut.keys);
+      const existingShortcuts = this.shortcuts.get(key) || [];
+      existingShortcuts.push(shortcut);
+      this.shortcuts.set(key, existingShortcuts);
```
angle943: i'm probably being OCD, but instead of using `.push()` (which mutates the existing array), can we do:

```
this.shortcuts.set(key, [...existingShortcuts, shortcut]);
```
My fear is that what if a key event happens while this logic is currently in the state of happening.

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,129 @@
...
+    });
+  }
+
+  private unregister(fullId: string): void {
```
angle943: maybe acll it `namespacedId`? lets have consistency

angle943: also, are plugin developers calling this?

If so, how would they know they need to pass in `pluginId.id`? Would it be better to put Pick<KeyboardShortcut, 'id' | 'pluginId'> so that devs don't need to worry about implementation details?

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,129 @@
...
+    for (const [key, shortcuts] of this.shortcuts.entries()) {
```
angle943: this is inefficient when we have a lot of registered keys. lets implement the second data structure we talked about

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,129 @@
...
+      const filteredShortcuts = shortcuts.filter(
+        (shortcut) => this.getNamespacedIdForKeyboardShortcut(shortcut) !== fullId
+      );
+
+      if (filteredShortcuts.length === 0) {
+        this.shortcuts.delete(key);
```
angle943: this is deleting for a key that the passed argument is unrelated to. i don't think this logic should be here. it should be handled when you are deleting the last entry in your specific key.

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,129 @@
...
+      } else if (filteredShortcuts.length !== shortcuts.length) {
+        this.shortcuts.set(key, filteredShortcuts);
+      }
+    }
+  }
+
+  private isHTMLElement(target: EventTarget | null): target is HTMLElement {
```
angle943, line 111: what is this typescript magic? o.O (i was not aware of this, TIL)

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,129 @@
...
+  private handleKeyboardEvent = (event: KeyboardEvent): void => {
+    if (this.shouldIgnoreKeyboardEventForTarget(event.target)) {
+      return;
+    }
+
+    const eventKeyString = this.getEventKeyString(event);
+    const shortcuts = this.shortcuts.get(eventKeyString);
+
+    if (shortcuts && shortcuts.length > 0) {
+      event.preventDefault();
```
angle943, line 137: add a comment that we want to prevent browser specific keybindings if they conflict

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  private handleKeyboardEvent = (event: KeyboardEvent): void => {
+    if (this.shouldIgnoreKeyboardEventForTarget(event.target)) {
+      return;
+    }
+
+    const eventKeyString = this.getEventKeyString(event);
+    const shortcuts = this.shortcuts.get(eventKeyString);
+
+    if (shortcuts && shortcuts.length > 0) {
```
angle943: nitpick: `if (shortcuts?.length > 0)`

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,129 @@
...
+      event.preventDefault();
+
+      shortcuts.forEach((shortcut) => {
```
angle943: we should execute the last one, not each one

angle943:
Note to maintainers: I have marked this PR's changelog as SKIP because for some reason, the changelog entry kept failing with `feat: ` prefix. We will have more PRs concerning this feat and in those PRs, we will check to ensure the changelog is properly added

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,159 @@
...
+      key += 'alt+';
+    }
+
+    if (event.shiftKey) {
+      key += 'shift+';
+    }
+
+    if (event.metaKey) {
+      key += 'cmd+';
+    }
```
angle943, line 65: later we will refactor this to unify mac's command and window's ctrl and so on, rifhrt?

src/core/public/keyboard_shortcut/index.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
```
kavilla, line 4: wrong license header

here is the right one

```
/*
 */
 ```

File: src/core/public/keyboard_shortcut/keyboard_shortcut_service.mock.ts [hunk 49]
```diff
@@ -0,0 +1,37 @@
+/*
```
kavilla, line 1: same as above

```
/*
 */
 ```

File: src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts [hunk 50]
```diff
@@ -0,0 +1,159 @@
+/*
```
kavilla, line 1: license header

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,159 @@
...
+  private getNormalizedKey = (str: string): string => str.toLowerCase();
+
+  private getNamespacedId = (shortcut: Pick<ShortcutDefinition, 'id' | 'pluginId'>) =>
+    `${shortcut.id.toLowerCase()}.${shortcut.pluginId.toLowerCase()}`;
+
+  private getEventKeyString = (event: KeyboardEvent): string => {
+    let key = '';
+
+    if (event.ctrlKey) {
+      key += 'ctrl+';
```
kavilla, line 52: i think these should be enums or something in a constants file

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,159 @@
...
+  private isHTMLElement(target: EventTarget | null): target is HTMLElement {
+    return target !== null && 'tagName' in target;
+  }
+
+  private shouldIgnoreKeyboardEventForTarget(target: EventTarget | null): boolean {
+    if (!this.isHTMLElement(target)) return false;
+
+    const tagName = target.tagName;
+
+    if (tagName === 'INPUT' || tagName === 'TEXTAREA') {
```
kavilla, line 120: nit: 

`return (tagName === 'INPUT' || tagName === 'TEXTAREA')`

and are we sure this will properly capture all that should be ignore

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,159 @@
...
+        this.shortcutsMapByKey.set(key, filteredShortcuts);
+      }
+    }
+  }
+
+  private isHTMLElement(target: EventTarget | null): target is HTMLElement {
+    return target !== null && 'tagName' in target;
+  }
+
+  private shouldIgnoreKeyboardEventForTarget(target: EventTarget | null): boolean {
```
kavilla, line 115: should. eventtarget ever be null?

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    if (shortcuts?.length) {
+      // Prevent browser-specific keybindings if they conflict with our shortcuts
+      event.preventDefault();
+
+      const shortcut = shortcuts[shortcuts.length - 1];
+      try {
+        shortcut.execute();
+      } catch (error) {
+        // eslint-disable-next-line no-console
```
kavilla, line 143: avoid disabling linter infavor of some logging service

src/core/public/keyboard_shortcut/types.ts
```diff
@@ -0,0 +1,62 @@
+/*
```
kavilla, line 1: licesnse header

src/core/public/keyboard_shortcut/types.ts
```diff
@@ -0,0 +1,62 @@
...
+   * Format: 'ctrl+s', 'shift+d', 'alt+enter', etc.
+   * Case-insensitive, normalized internally
+   */
+  keys: string;
+
+  /**
+   * Function to execute when the shortcut is triggered
+   * Can be async for operations that require API calls
+   */
+  execute: () => void | Promise<void>;
```
kavilla, line 61: should we just make this void and have another funciton like `execute$` for the promise. it's difficult if an api could or could not be async or not

src/plugins/discover/public/plugin.ts
```diff
@@ -417,7 +417,6 @@ export class DiscoverPlugin
 
   start(core: CoreStart, plugins: DiscoverStartPlugins) {
     setUiActions(plugins.uiActions);
-
```
kavilla, line 420: uneeded

src/plugins/visualize/public/plugin.ts
```diff
@@ -297,6 +297,7 @@ export class VisualizePlugin
     setApplication(core.application);
     setIndexPatterns(plugins.data.indexPatterns);
     setQueryService(plugins.data.query);
+
```
kavilla, line 300: uneeded

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
angle943: this will be refactored later to handle it better. @natel926 lets leave a comment saying it is temporary and will be refactored later

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
angle943: this will be refactored later to handle it better. @natel926 lets leave a comment saying it is temporary and will be refactored later

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
angle943: For some reason typescript thinks yes. This is moreso to handle type error, but perhaps there are? or perhaps `EventTarget` is a generic type that is used for other cases irrelevant to us that does not have eventtarget

src/core/public/keyboard_shortcut/types.ts
angle943: synced offline and we decided to not handle promise functions

natel926: Agreed. I will update it on next pr

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
natel926: Yes. will hand it on next pr.

ps48:

PR #10352, opened 2025-08-06
Add Correlations Plugin - Saved Object Type Registration

sumukhswamy:
### Description

This PR introduces a new plugin called "correlations" that registers a saved object type for storing correlation. 
This plugin provides a foundation for managing correlations between traces and logs

Object added for Correlations

```
{
  "id": "trace-to-logs-correlation-5678",
  "type": "correlations",
  "migrationVersion": "1.0.0",
  "updatedAt": "2025-07-16T10:15:30.000Z",
  "attributes": {
    "type": "APM-Correlation",
    "entities": [
      {
        "tracesDataset":{
          "id":"refernces[0].id"
        }
      },
      {
        "logsDataset":{
          "id":"refernces[1].id",
          "meta": {
            "logServiceNameField": "service.name",
            "logSpanIdField": "span_id",
            "logTraceIdField": "trace_id",
            "timestamp": "@timestamp"
          }
        }
      },
      {
        "<Plugin-name>DataConnection":{
          "id": "refernces[2].id"
        }
      }
    ]
  },
  "references": [
    {
      "name": "entities[0].index",
      "type": "index-pattern",
      "id": "6ff30050-72ec-11f0-b9fd-951b6e375a38"
    },
    {
      "name": "entities[1].index",
      "type": "index-pattern",
      "id": "72048c50-731a-11f0-8fdf-87b448b7eb0b"
    },
    {
      "name": "<plugin>DataConnection",
      "type": "data-connection",
      "id": "1234567890"
    }
  ]
}

```

## Changelog
-feat: adds a correlation plugin which creates a saved object for the related traces and logs

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/correlations/common/types.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export interface CorrelationSavedObjectAttributes extends SavedObjectAttributes {
+  title: string;
+  description: string;
+  correlationType: string;
+  version: string;
+  entities: any[];
+}
+
+export interface CorrelationSavedObject {
```
ps48: Can this just extend savedObject itself? So we don't have to declare individual fields inherited from the SO like id, type, etc?

src/plugins/correlations/common/types.ts
```diff
@@ -0,0 +1,29 @@
...
+ * GitHub history for details.
+ */
+
+import { SavedObjectAttributes, SavedObjectReference } from 'opensearch-dashboards/server';
+
+export interface CorrelationSavedObjectAttributes extends SavedObjectAttributes {
+  title: string;
+  description: string;
+  correlationType: string;
+  version: string;
```
ps48, line 12: Is this following semver style versioning? Can you please add an example object in the comments. Also later when we create the helper functions like create/update, will we validate the field contents apart from the type itself?

src/plugins/correlations/common/types.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { SavedObjectAttributes, SavedObjectReference } from 'opensearch-dashboards/server';
+
+export interface CorrelationSavedObjectAttributes extends SavedObjectAttributes {
+  title: string;
+  description: string;
```
ps48: Do we need these for P0?

src/plugins/correlations/server/saved_objects/correlations.ts
```diff
@@ -0,0 +1,55 @@
...
+      version: {
+        type: 'keyword',
+      },
+      entities: {
+        type: 'object',
+        enabled: false,
+      },
+    },
+  },
+  migrations: {},
```
ps48, line 35: Do we need the migration version here? May be you can check how OSD Dashboards does it?

src/plugins/correlations/common/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  correlationType: string;
```
ps48, line 10: we need some basic convention here. Something like `correlation-pluginname/featurename`. Again this can be added in the comments as a full fledged example.

src/plugins/correlations/server/plugin.ts
```diff
@@ -0,0 +1,44 @@
...
+    this.logger.debug('correlations: Setup');
+
+    core.savedObjects.registerType(correlationsSavedObjectType);
+
+    return {};
+  }
+
+  public start(core: CoreStart) {
+    this.logger.debug('correlations: Started');
+    return {};
```
ps48, line 34: Any helper functions we want to expose here for read/update?

src/plugins/correlations/server/saved_objects/correlations.ts
```diff
@@ -0,0 +1,55 @@
...
+export const correlationsSavedObjectType: SavedObjectsType = {
+  name: 'correlations',
+  hidden: false,
+  namespaceType: 'single',
+  management: {
+    icon: 'link',
+    defaultSearchField: 'title',
+    importableAndExportable: true,
+    getTitle(obj) {
+      return obj.attributes.correlationType || `Correlation ${obj.id}`;
```
ps48: May be we can just keep correlation-type for now. Id usually doesn't make sense in the title. Also, if we plan to support `title` then lets use it.

src/plugins/correlations/server/plugin.ts
sumukhswamy: nothing as of now, gonna expose a savedobject client which can be accessed outside

ruanyl:
Should this plugin be in core? Sounds better in [dashboards-observability](https://github.com/opensearch-project/dashboards-observability)

sumukhswamy:
> Should this plugin be in core? Sounds better in [dashboards-observability](https://github.com/opensearch-project/dashboards-observability)

the correlation we are implementing is closely associated with the Discover traces experience which would make it better to be part of Dashboards

src/plugins/correlations/server/saved_objects/correlations.ts
```diff
@@ -0,0 +1,41 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { SavedObjectsType } from '../../../../core/server';
+
+export const correlationsSavedObjectType: SavedObjectsType = {
```
ps48, line 9: do we want to call this out as experimental and schema and APIs can change in near future

sumukhswamy: sure

ruanyl:
> > Should this plugin be in core? Sounds better in [dashboards-observability](https://github.com/opensearch-project/dashboards-observability)
> 
> the correlation we are implementing is closely associated with the Discover traces experience which would make it better to be part of Dashboards

But then should it be in discover? I may not have the full picture, could you elaborate why we want a new plugin for this? Thanks :)

src/plugins/correlations/common/types.ts
```diff
@@ -0,0 +1,23 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
```
kavilla, line 3: Could we update these to be the right license header

src/plugins/correlations/common/types.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface CorrelationSavedObjectAttributes extends SavedObjectAttributes {
+  correlationType: string;
+
+  version: string;
+  entities: any[];
+}
+
+export interface CorrelationSavedObject extends SavedObject<CorrelationSavedObjectAttributes> {
+  type: 'correlations';
```
kavilla, line 17: Normally type is singular. But is this because you can have a single object with multiple correlations?

src/plugins/correlations/common/types.ts
```diff
@@ -0,0 +1,23 @@
...
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { SavedObject, SavedObjectAttributes } from '../../../core/public';
+
+export interface CorrelationSavedObjectAttributes extends SavedObjectAttributes {
+  correlationType: string;
```
kavilla, line 10: If this is attributes of the correlation it seems little bit redundant to name the field correlationType. Could it not just be type?

src/plugins/correlations/common/types.ts
kavilla: +1 would be good to get more insight how this value will be used and how it will be updated . I'm assuming it will be updated on our side. And does it conflict with the version on the saved objects already. That is encoded to account for the number of revisions

src/plugins/correlations/opensearch_dashboards.json
```diff
@@ -0,0 +1,8 @@
+{
+  "id": "correlations",
+  "version": "opensearchDashboards",
+  "server": true,
+  "ui": true,
```
kavilla, line 5: How come ?  I think the pr description says server side plugin so then we dont need the UI to be true and we should delete the public file

src/plugins/correlations/common/types.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  version: string;
+  entities: any[];
```
kavilla, line 13: Type any is really general I think we should nail down the object prior to registering it and migrating it

kavilla (changes requested):
Could we consider this going into query enhancements plugin? Did you use the plugin generator script for this? A dedicated plugin to registering one type might just had more bloat. 

Unless we intend to do more with this plugin, I think it should go into query enhancements so we can justify not doing capabilities API checks . If needs to be a plugin it should be disabled by default and not register a new type

Also I thought we were planning on making a generic object that could be used by other parts of the app

src/plugins/correlations/common/types.ts
sumukhswamy: yes as this could be any number of correlated objects

src/plugins/correlations/common/types.ts
sumukhswamy: the reason for this is that any plugin could use it for relating the services it needs, this would help in narrowing down which plugin is currently using this

src/plugins/correlations/opensearch_dashboards.json
sumukhswamy: will update the PR description, to have access to the savedobject client, I have added the public folder

src/plugins/correlations/common/types.ts
sumukhswamy: this is the experimental schema which has a generic object being used by the dependent plugin

PR #10355, opened 2025-08-06
[Traces] Update for datasets, redirection, and accordion timeline

TackAdam:
### Description
[Traces] Update for datasets, redirection, and accordion timeline
1. Switch from using index pattern to using datasets. This required the dataset{index pattern} to be created for the page to work.
2. Add the timeline tab to the drop down accordion which shows the gantt chart and current selected span
3. Add redirection to the span column
4. Minor UI adjustments to Trace Detail page for hiding columns in span / tree view

## Screenshot

New tab only present on the trace flavor, and redirection only applied to spanId on the trace page.

https://github.com/user-attachments/assets/730b6215-da5d-45dd-a7de-34905309a93b

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_panel.tsx
```diff
@@ -183,51 +127,79 @@ export function SpanDetailPanel(props: {
...
+    const calculateGanttHeight = (spanCount: number): number => {
+      const rowHeight = 35;
+      const baseHeight = 100;
 
-    if (spanCount === 0) {
-      return 150;
-    }
+      if (spanCount === 0) {
+        return 150;
+      }
```
angle943, line 141: how did this get by CI?

src/plugins/explore/public/components/data_table/tabe_cell/table_cell.tsx
```diff
@@ -6,33 +6,129 @@
...
   columnId: string;
   isTimeField?: boolean;
   onFilter?: DocViewFilterFn;
   fieldMapping?: any;
   sanitizedCellValue: string;
+  rowData?: any;
 }
 
+// Custom URL functionality for trace navigation
+const isOnTracesPage = (): boolean => {
```
angle943: 1.  can we rename `tabe_cell` folder to `table_cell`?
2. can we add unit tests for these small functions? (there is no export)
3. (idea) i am worried about adding trace-specific functionalities to this generalized component. Just for organizational sake, can we take these out of this file?

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_panel.tsx
angle943: nvm its because you added useMemo

src/plugins/explore/public/components/data_table/tabe_cell/table_cell.tsx
TackAdam: 1. Folder renamed 
2. Added Test
```
 PASS  src/plugins/explore/public/components/data_table/table_cell/trace_utils/trace_utils.test.tsx (5.285 s)
  trace_utils
    isOnTracesPage
      ✓ should return true when pathname includes /explore/traces (1 ms)
      ✓ should return true when hash includes /explore/traces
      ✓ should return true when both pathname and hash include /explore/traces (1 ms)
      ✓ should return false when neither pathname nor hash include /explore/traces
      ✓ should return false when pathname is empty
      ✓ should handle partial matches correctly
    isSpanIdColumn
      ✓ should return true for spanId
      ✓ should return true for span_id
      ✓ should return true for spanID (1 ms)
      ✓ should return false for other column names
      ✓ should be case sensitive
      ✓ should handle null and undefined
    extractTraceIdFromRowData
      ✓ should return empty string for null or undefined rowData
      ✓ should extract traceId from direct field
      ✓ should extract trace_id from direct field
      ✓ should extract traceID from direct field (3 ms)
      ✓ should extract traceId from _source.traceId
      ✓ should extract trace_id from _source.trace_id
      ✓ should extract traceID from _source.traceID
      ✓ should prioritize direct fields over _source fields
      ✓ should return empty string if no trace ID fields are found
      ✓ should return empty string if trace ID field exists but is not a string
      ✓ should return empty string if trace ID field is empty string
      ✓ should handle nested _source field that is null (1 ms)
      ✓ should handle missing _source field gracefully
      ✓ should handle complex nested objects
    buildTraceDetailsUrl
      ✓ should build URL with span ID and trace ID
      ✓ should build URL with only span ID when trace ID is empty
      ✓ should use default values when dataset is null
      ✓ should use default values when dataset properties are missing
      ✓ should handle base path correctly
      ✓ should handle no base path (1 ms)
      ✓ should handle special characters in dataset values
      ✓ should handle empty span ID
    handleSpanIdNavigation
      ✓ should open new window with correct URL (1 ms)
      ✓ should strip HTML tags from sanitized cell value
      ✓ should trim whitespace from span ID
      ✓ should handle complex HTML tags
      ✓ should work when no trace ID is found in row data
      ✓ should handle null row data
      ✓ should handle null dataset (1 ms)
      ✓ should handle empty sanitized cell value
    SpanIdLink
      ✓ should render span ID link with correct text (14 ms)
      ✓ should strip HTML tags from sanitized cell value in display (4 ms)
      ✓ should call handleSpanIdNavigation when clicked (4 ms)
      ✓ should display popout icon (2 ms)
      ✓ should handle whitespace in sanitized cell value (2 ms)
      ✓ should work with null row data (3 ms)
      ✓ should work with null dataset (1 ms)
      ✓ should have correct tooltip text (2 ms)

Tests:       50 passed, 50 total
Snapshots:   0 total
Time:        6.27 s
✨  Done in 12.70s.
```

3.  Moved trace specific functionality to 
``` src/plugins/explore/public/components/data_table/table_cell/trace_utils ```

File: src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_panel.tsx [hunk 3]
```diff
@@ -57,35 +25,27 @@ export interface TraceFilter {
...
-  const {
-    chrome,
-    spanFilters,
-    setSpanFiltersWithStorage,
-    payloadData,
-    onSpanSelect,
-    dataSourceMDSId,
-    colorMap,
-  } = props;
+  const { chrome, spanFilters, payloadData, onSpanSelect, dataSourceMDSId, colorMap } = props;
```
kavilla: Can some of the services like chrome be accessed by the context?

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_panel.tsx
```diff
@@ -57,35 +25,27 @@ export interface TraceFilter {
...
   const containerRef = useRef<HTMLDivElement | null>(null);
-  const [availableWidth, setAvailableWidth] = useState<number>(window.innerWidth);
-  const isLocked = useObservable(chrome.getIsNavDrawerLocked$());
+  const [availableWidth, setAvailableWidth] = useState<number>(800); // Default width
```
kavilla: Why 800 and should it be defined somewhere in a constant?

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_panel.tsx
```diff
@@ -96,39 +56,23 @@ export function SpanDetailPanel(props: {
...
+    if (props.isEmbedded) {
+      // In embedded mode, set width once based on container size and don't listen to resize events
+      if (containerRef.current) {
+        const containerWidth = containerRef.current.getBoundingClientRect().width;
+        setAvailableWidth(containerWidth > 0 ? containerWidth : 800);
+      }
+      return; // No cleanup needed, no listeners added
+    }
+
+    // Add resize listeners
```
kavilla: Nit: probably unneeded comment

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_panel.tsx
```diff
@@ -183,51 +127,79 @@ export function SpanDetailPanel(props: {
...
   }, [payloadData]);
 
-  // Calculate dynamic height based on number of spans
-  const calculateGanttHeight = (spanCount: number): number => {
-    const rowHeight = 35;
-    const baseHeight = 100;
+  const ganttChart = useMemo(() => {
+    // Calculate dynamic height based on number of spans
+    const calculateGanttHeight = (spanCount: number): number => {
+      const rowHeight = 35;
```
kavilla, line 136: How come we used this number?

src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_helpers.tsx
```diff
@@ -21,48 +25,47 @@ export interface PPLQueryRequest {
         language: string;
         format: string;
         dataset: {
-          dataSource: {
-            id: string;
```
kavilla, line 25: Dataset should still have a data source if u needed it

angle943 (dismissed):

src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_helpers.tsx
```diff
@@ -21,48 +25,47 @@ export interface PPLQueryRequest {
...
-            title: string;
-            type: string;
-          };
           id: string;
           title: string;
           type: string;
-          isRemoteDataset: boolean;
```
kavilla, line 32: I guess I'm not too sure why we are redefining the search request interface. And it doesn't seem too directly tied with ppl because SQL can look like this. Do you mind providing some insight here

src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx
```diff
@@ -12,8 +12,12 @@ export interface PPLQueryParamsWithFilters extends PPLQueryParams {
 export interface PPLSpanQueryParams {
   traceId: string;
   spanId: string;
-  dataSourceId: string;
-  indexPattern: string;
+  dataset: {
```
kavilla: Imo you should just import the type from the data plugin

src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx
```diff
@@ -24,12 +28,12 @@ export class TracePPLService extends PPLService {
   }
 
   private buildPPLQueryWithFilters(
-    indexPattern: string,
+    datasetTitle: string,
```
kavilla: Imo I would just pass in the full dataset object and access the title. That way if you need other stuff in the future you can access it without updating the params

src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx
```diff
@@ -42,18 +46,25 @@ export class TracePPLService extends PPLService {
...
-      return await this.executeQuery(dataSourceId, indexPattern, pplQuery);
+      const datasetWithoutTime = {
+        id: dataset.id,
+        title: dataset.title,
+        type: dataset.type,
+        // Omit timeFieldName to prevent automatic time filtering
+      };
+
+      // Execute using the base class method with the modified dataset
+      return await this.executeQuery(datasetWithoutTime, pplQuery);
```
kavilla, line 63: How come we need this?

src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx
```diff
@@ -63,20 +74,18 @@ export class TracePPLService extends PPLService {
...
   async fetchSpanDetails(params: PPLSpanQueryParams): Promise<any> {
-    const { traceId, spanId, dataSourceId, indexPattern, limit = 100 } = params;
+    const { traceId, spanId, dataset, limit = 100 } = params;
 
-    if (!traceId || !spanId || !dataSourceId || !indexPattern) {
-      throw new Error(
-        'Missing required parameters: traceId, spanId, dataSourceId, and indexPattern'
-      );
+    if (!traceId || !spanId || !dataset) {
+      throw new Error('Missing required parameters: traceId, spanId, and dataset');
```
kavilla, line 76: Should we say missing one of the required parameters ?

src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx
```diff
@@ -63,20 +74,18 @@ export class TracePPLService extends PPLService {
...
-      const pplQuery = `source = ${indexPattern} | where traceId = "${traceId}" | where spanId = "${spanId}" | head ${limit}`;
+      // Construct the PPL query to filter by trace ID and span ID using the dataset title
+      const pplQuery = `source = ${dataset.title} | where traceId = "${traceId}" | where spanId = "${spanId}" | head ${limit}`;
 
       // Execute using the base class method
-      return await this.executeQuery(dataSourceId, indexPattern, pplQuery);
+      return await this.executeQuery(dataset, pplQuery);
     } catch (error) {
       // eslint-disable-next-line no-console
       console.error('PPL Span Query Error:', error);
```
kavilla, line 87: We should try to avoid using console log and disabling it. We can use the logging service

src/plugins/explore/public/application/pages/traces/trace_details/state/trace_app_state.ts
```diff
@@ -11,8 +11,12 @@ import {
 
 export interface TraceAppState {
   traceId: string;
-  dataSourceId: string;
-  indexPattern: string;
+  dataset: {
```
kavilla: Could import the dataset type but sorry little bit confused on the interface name. I see it's traces appstate. But it has a dataset. Shouldn't the dataset already be apart of the query state. We are not duplicating the dataset in the app state and query state right

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
```diff
@@ -275,24 +282,47 @@ export const TraceDetails: React.FC<TraceDetailsProps> = ({ setMenuMountPoint })
   };
 
   // Set up ResizeObserver to detect when the main panel size changes
+  // Only enable this in non-embedded mode to avoid crashes in embedded contexts
   useEffect(() => {
-    if (!mainPanelRef.current) return;
+    if (!mainPanelRef.current || isEmbedded) return;
 
+    let resizeTimeout: NodeJS.Timeout;
```
kavilla, line 289: I guess if it's how other visualizations do it then it's fine. But I would be surprised if didn't have utils for this. Or does the way we present the chart very unique. Because I'm not sure if I would be able to modify this file if requirements are updated.

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
```diff
@@ -407,14 +437,9 @@ export const TraceDetails: React.FC<TraceDetailsProps> = ({ setMenuMountPoint })
...
                               spanFilters={spanFilters}
-                              setSpanFiltersWithStorage={setSpanFiltersWithStorage}
                               payloadData={JSON.stringify(transformedHits)}
                               isGanttChartLoading={isBackgroundLoading}
-                              dataSourceMDSId={dataSourceId}
-                              dataSourceMDSLabel={undefined}
-                              traceId={traceId}
-                              pplService={pplService}
-                              indexPattern={indexPattern}
+                              dataSourceMDSId={dataset.id}
```
kavilla: Dataset should have a data source .id

src/plugins/explore/public/components/data_table/tabe_cell/table_source_cell.scss
kavilla, line 1: Nice catch on the typo !

src/plugins/explore/public/components/data_table/table_cell/table_source_cell.scss
```diff
@@ -0,0 +1,5 @@
+.exploreTableSourceCell {
+  .exploreDescriptionListFieldTitle {
+    background-color: tintOrShade($euiColorPrimary, 90%, 70%);
```
kavilla, line 3: Could we verify darkmode

src/plugins/explore/public/components/data_table/table_cell/trace_utils/trace_utils.tsx
```diff
@@ -0,0 +1,110 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from 'react';
+import { EuiToolTip, EuiLink, EuiIcon } from '@elastic/eui';
+import { i18n } from '@osd/i18n';
+
+export const isOnTracesPage = (): boolean => {
+  return (
+    window.location.pathname.includes('/explore/traces') ||
```
kavilla, line 14: Did the application service functions not work for this?

src/plugins/explore/public/components/data_table/table_cell/trace_utils/trace_utils.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }
+  }
+
+  return '';
+};
+
+export const buildTraceDetailsUrl = (
+  spanIdValue: string,
+  traceIdValue: string,
+  dataset: any
```
kavilla, line 42: Avoid any

src/plugins/explore/public/components/data_table/table_cell/trace_utils/trace_utils.tsx
```diff
@@ -0,0 +1,110 @@
...
+): string => {
+  const origin = window.location.origin;
+  const pathname = window.location.pathname;
+
+  // Get the base path before /app
+  const basePathMatch = pathname.match(/^(.*?)\/app/);
```
kavilla, line 48: I would def use the application service helper functions this is supported with a simple util function you can navigate while also provided a path

src/plugins/explore/public/components/data_table/table_cell/trace_utils/trace_utils.tsx
```diff
@@ -0,0 +1,110 @@
...
+  return (
+    <EuiToolTip
+      content={i18n.translate('explore.spanIdLink.redirectTooltip', {
+        defaultMessage: 'Redirect to trace details',
+      })}
+    >
+      <EuiLink
+        onClick={handleSpanIdClick}
+        data-test-subj="spanIdLink"
+        style={{ cursor: 'pointer', display: 'flex', alignItems: 'center', gap: '4px' }}
```
kavilla: Avoid inline styling. Little bit difficult to find this if it required updating in the future

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
```diff
@@ -0,0 +1,267 @@
...
+import {
+  transformPPLDataToTraceHits,
+  TraceHit,
+} from '../../../application/pages/traces/trace_details/public/traces/ppl_to_trace_hits';
+import { generateColorMap } from '../../../application/pages/traces/trace_details/public/traces/generate_color_map';
+
+const extractTraceIdFromHit = (hit: any): string | null => {
+  // Try different possible field paths for trace ID
+  const possiblePaths = [
+    'traceId',
```
kavilla: There might be a cleaner solution in the future, but for now I think I saw above that the list of possible traces is different than here. Is that intentional? If not then could we define those as a constants in a constant file and reuse them

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
```diff
@@ -0,0 +1,267 @@
...
+  return null;
+};
+
+const getNestedValue = (obj: any, path: string): any => {
+  return path.split('.').reduce((current, key) => {
+    return current && current[key] !== undefined ? current[key] : null;
+  }, obj);
+};
+
+const extractIndexPattern = (indexPattern: any, dataset: any): string => {
```
kavilla: In reality the context should be updated if not then it might break. But non the less you shouldn't have to extract it. You can just access the dataset from the querystring manager.

Like services.data.query.queryString.getQuery(). dataset.

The URL state should be listening to changes on the query string manager.

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
```diff
@@ -0,0 +1,267 @@
...
+        return titleMatch[1];
+      }
+    }
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.warn('Failed to extract index pattern from URL:', error);
+  }
+
+  // Fallback to default pattern
+  return 'otel-v1-apm-span-*';
```
kavilla: I would just throw an error tbh

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
```diff
@@ -0,0 +1,267 @@
...
+};
+
+const isOnTracesFlavor = (): boolean => {
+  const currentPath = window.location.pathname;
```
kavilla, line 52: I think this is duplicate and we should prolly just use the e app service helper functions

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
```diff
@@ -0,0 +1,267 @@
...
+// Trace Details view component for the doc viewer accordion - Timeline only
+export function TraceDetailsView({ hit, indexPattern }: DocViewRenderProps) {
+  const [transformedHits, setTransformedHits] = useState<TraceHit[]>([]);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  const {
+    services: { data, chrome },
+  } = useOpenSearchDashboards<DataExplorerServices>();
+
+  // Get dataset context to access the current index pattern
```
kavilla: Nit: I think the next line conveys this information so we prolly don't these comments

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
```diff
@@ -0,0 +1,267 @@
...
+  const { dataset } = useDatasetContext();
+
+  // Check if we're on traces flavor
+  const isOnTraces = isOnTracesFlavor();
+
+  // Extract trace information from the hit
+  const traceInfo = useMemo(() => {
+    if (!isOnTraces) {
```
kavilla, line 72: Sorry confused, do we not unmount? How will this be triggered if not on traces page. Unless we are always listening which might be a memory concern

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
```diff
@@ -0,0 +1,267 @@
...
+        id: dataset?.id || 'default-dataset-id',
+        title: indexPatternTitle,
+        type: dataset?.type || 'INDEX_PATTERN',
+        timeFieldName: dataset?.timeFieldName,
+      },
+    };
+  }, [hit, indexPattern, dataset, isOnTraces]);
+
+  // Create PPL service instance
+  const pplService = useMemo(() => (data ? new TracePPLService(data) : undefined), [data]);
```
kavilla, line 99: What is the ppl service?

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
```diff
@@ -0,0 +1,267 @@
...
+    }
+  }, [pplService, traceInfo]);
+
+  // Only show trace details view when on traces flavor
+  if (!isOnTraces) {
+    return null;
+  }
+
+  // Show empty state if no trace data is available
+  if (!traceInfo) {
```
kavilla, line 151: Do we not have a helper function company so that we can import and keep consistency between the pages? It will become a burden if tasked to modify an zero data state if I have to find each instance

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
```diff
@@ -0,0 +1,267 @@
...
+        />
+      </div>
+    );
+  }
+
+  return (
+    <div className="exploreTraceDetailsView">
+      {isLoading ? (
+        <div className="exploreTraceDetailsView__loadingContainer">
+          <EuiText>Loading trace timeline...</EuiText>
```
kavilla: I18n? And is it worth specifying. Could it not be just loading. The reason is because I think we had at once a vis type of timeline. Which could be confusing if we now reuse it to mean gantt

src/plugins/explore/public/plugin.ts
```diff
@@ -113,6 +114,20 @@ export class ExplorePlugin
 
     this.docViewsRegistry = new DocViewsRegistry();
     setDocViewsRegistry(this.docViewsRegistry);
+    this.docViewsRegistry.addDocView({
+      title: i18n.translate('explore.docViews.timeline.title', {
```
kavilla: I think this title should be updated to specific traces

src/plugins/explore/public/plugin.ts
```diff
@@ -113,6 +114,20 @@ export class ExplorePlugin
...
+        defaultMessage: 'Timeline',
+      }),
+      order: 5,
+      component: TraceDetailsView,
+      shouldShow: (hit) => {
+        // Only show the Timeline tab when on the traces flavor
+        const currentPath = window.location.pathname;
```
kavilla, line 125: Should use application service

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
```diff
@@ -0,0 +1,267 @@
...
+        </div>
+      ) : transformedHits.length > 0 ? (
+        <SpanDetailPanel
+          chrome={chrome}
+          spanFilters={[]}
+          payloadData={JSON.stringify(transformedHits)}
+          isGanttChartLoading={false}
+          dataSourceMDSId={dataset?.id || ''}
```
kavilla: I would just pass the full dataset so update the prop. The dataset has a data source property. So this id passed here will likely be the wrong id

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
```diff
@@ -0,0 +1,267 @@
...
+          colorMap={colorMap}
+          onSpanSelect={undefined}
+          selectedSpanId={traceInfo.spanId || undefined}
```
kavilla, line 202: If traceinfo is undefined it will be an error so I think it will be if spanid is undefined right? So then does passing undefined here become redundant

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
```diff
@@ -0,0 +1,267 @@
...
+    <div className="exploreTraceDetailsView">
+      {isLoading ? (
+        <div className="exploreTraceDetailsView__loadingContainer">
+          <EuiText>Loading trace timeline...</EuiText>
+        </div>
+      ) : transformedHits.length > 0 ? (
+        <SpanDetailPanel
+          chrome={chrome}
+          spanFilters={[]}
+          payloadData={JSON.stringify(transformedHits)}
```
kavilla, line 198: Are we sure about stringifying data responses? Is this a lot of data

src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_helpers.tsx
TackAdam: That code was needed for the ppl request to work when using the index. With the index pattern/dataset implementation this is how the query looks. I believe the dataset.id is replacing the datasource.id
```
{
    "query": "source = otel-v1-apm-span* | where traceId = \"17f89aba8a521c0a41f05f4d14830ca4\" | head 100",
    "language": "PPL",
    "dataset": {
        "id": "6ff30050-72ec-11f0-b9fd-951b6e375a38",
        "title": "otel-v1-apm-span*",
        "type": "INDEX_PATTERN"
    },
    "format": "jdbc"
}
```

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_panel.tsx
TackAdam: Height of the bars in the gantt chart for proper layout
Screenshot 2025-08-06 at 1 27 50 PM

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
TackAdam: Yes, tested with traces that where much larger than normal expectations with no errors.

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
TackAdam: Linting error without this, If the spanId is undefined we have fallback logic to find the root span and select it from the trace.

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
TackAdam: Refactor as datasource is not needed in span detail panel with the other changes.
Deleted it.

src/plugins/explore/public/plugin.ts
TackAdam:
```
      title: i18n.translate('explore.docViews.trace.timeline.title', {
        defaultMessage: 'Timeline',
      }),
```

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
TackAdam: Added and switched too 
          <EuiText>
            {i18n.translate('explore.docViews.traceDetails.loading', {
              defaultMessage: 'Loading trace gantt chart...',
            })}
          </EuiText>

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
TackAdam: Fetches the Parent Trace info to make our timeline visualizations.

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
TackAdam: Added data to the decency of the useMemo to prevent memory concern, and the early return will effectively unmount when not on the trace page.

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
TackAdam: Removed.

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
TackAdam: Removed this code block

src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx
TackAdam: Removed and swapped to 
```
    const queryDataset = data?.query?.queryString?.getQuery()?.dataset;
    const currentDataset = queryDataset || dataset;
    ```

File: src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx (hunk 22 above)
Comment (TackAdam) [reply src/plugins/explore/public/components/doc_viewer/trace_details_view/trace_details_view.tsx]: added
```
src/plugins/explore/public/utils/trace_field_constants.ts
```
```
/*
 */

export const TRACE_ID_FIELD_PATHS = [
  'traceId',
  'trace_id',
  'traceID',
  '_source.traceId',
  '_source.trace_id',
  '_source.traceID',
  'fields.traceId',
  'fields.trace_id',
  'fields.traceID',
] as const;

export const SPAN_ID_FIELD_PATHS = [
  'spanId',
  'span_id',
  'spanID',
  '_source.spanId',
  '_source.span_id',
  '_source.spanID',
  'fields.spanId',
  'fields.span_id',
  'fields.spanID',
] as const;

````

File: src/plugins/explore/public/components/data_table/table_cell/trace_utils/trace_utils.tsx (hunk 21 above)
Comment (TackAdam) [reply src/plugins/explore/public/components/data_table/table_cell/trace_utils/trace_utils.tsx]: Moved too 
```
.exploreSpanIdLink {
  cursor: pointer;
  display: flex;
  align-items: center;
  gap: 4px;
}
```

File: src/plugins/explore/public/components/data_table/table_cell/trace_utils/trace_utils.tsx (hunk 18 above)
Comment (TackAdam) [reply src/plugins/explore/public/components/data_table/table_cell/trace_utils/trace_utils.tsx:14]: No the history was not working with the explore flavors. It always returned empty string and could not re-direct based on that. Can re-visit in follow up with assistance.

File: src/plugins/explore/public/components/data_table/table_cell/table_source_cell.scss (hunk 17 above)
Comment (TackAdam) [reply src/plugins/explore/public/components/data_table/table_cell/table_source_cell.scss:3]: When renaming the file i got the linting error that we can not modify global eui setting so this was required. 
Checked dark mode and appeared correct.

File: src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx (hunk 16 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx]: I am using the only example I have at the moment which is making an index pattern and navigating to the Discover Trace page. 
The url present when i select that index is
```
http://localhost:5601/w/-k1kHM/app/explore/traces#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-5m,to:now))&_q=(dataset:(id:'6ff30050-72ec-11f0-b9fd-951b6e375a38',timeFieldName:endTime,title:'otel-v1-apm-span*',type:INDEX_PATTERN),language:PPL,query:'')&_a=(legacy:(columns:!(spanId,status.code,attributes.http.status_code,serviceName,name,durationInNanos),interval:auto,isDirty:!f,sort:!()),tab:(logs:(),patterns:(usingRegexPatterns:!f)),ui:(activeTabId:logs,showHistogram:!t))
```

From this I did not see a data source
```
dataset:(id:'6ff30050-72ec-11f0-b9fd-951b6e375a38',timeFieldName:endTime,title:'otel-v1-apm-span*',type:INDEX_PATTERN)
```

File: src/plugins/explore/public/application/pages/traces/trace_details/state/trace_app_state.ts (hunk 14 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/state/trace_app_state.ts]: The trace app state is for the url of the trace page and allow us to update it as different spans are selected and allow a shareable link.

File: src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx (hunk 13 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx:87]: Can take this up as a refactor, haven't looked into the logging service.

File: src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx (hunk 12 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx:76]: I don't think that is necessary. This page will be navigated to through the Discover Trace links and a user should not be in this situation, it is just a safety net.

File: src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx (hunk 11 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx:63]: We need to run the query without a specific time field as the span we enter the page with will not have the context of its Parents trace timeframe. We need to remove the time field so that we get all spans related to the Trace to form are visualization's ect..

File: src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx (hunk 10 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx]: Updated to
```
  private buildPPLQueryWithFilters(
    dataset: Dataset,
    traceId: string,
    limit: number,
    filters: Array<{ field: string; value: any }> = []
  ): string {
    let query = `source = ${dataset.title} | where traceId = "${traceId}"`;
```

File: src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx (hunk 9 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_trace.tsx]: Added import { Dataset } from '../../../../../../../data/common';
```
  dataset: Dataset;
```

File: src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_helpers.tsx (hunk 8 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/server/ppl_request_helpers.tsx:32]: Needed the page to make specific trace queries that included the trace id and dynamically re-call with filters specific to traces. Can look into a refactor later if required.

File: src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_panel.tsx (hunk 5 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_panel.tsx]: Removed.

File: src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_panel.tsx (hunk 4 above)
Comment (TackAdam) [reply src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_panel.tsx]: Added to const file, needed for re-rendering / fallback to prevent crashes
```
// Default fallback width for span detail panel when container width cannot be determined
  DEFAULT_AVAILABLE_WIDTH: 800,
  ```

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_panel.tsx
TackAdam: I'm not sure, can take this as a follow-up to check.

angle943: chrome and things like that should not be passed as props, but use `useOpenSearchDashboardContext()` whenever you need them

angle943 (approved):

kavilla (approved):

PR #10358, opened 2025-08-06
[Explore] Infer tooltip date format dynamically

lunalzm:
### Description

This PR updates date format in the chart tooltip. This change automatically determines the appropriate time unit (e.g., seconds, minutes, days) based on the actual timestamp values in the dataset.

### Before
image

### After
image

## Testing the changes

1. Create area/line/bar charts with temporal fields using data with different time granularities (e.g., seconds, minutes, days, months).
2. Hover over data points and verify that the tooltip date format adjusts to match the inferred precision from the dataset.
3. Confirm that fallback formatting (%b %d, %Y %H:%M:%S) is applied when inference fails.

## Changelog
- fix: infer tooltip date format dynamically

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/line/to_expression.ts
+++ b/src/plugins/explore/public/components/visualizations/line/to_expression.ts
@@ -12,6 +12,7 @@ import {
   ValueAxisPosition,
 } from './line_chart_utils';
 import { createThresholdLayer, getStrokeDash } from '../style_panel/threshold/utils';
+import { inferTimeUnitFromTimestamps, timeUnitToFormat } from '../utils/utils';
 
 /**
  * Rule 1: Create a simple line chart with one metric and one date
@@ -36,6 +37,12 @@ export const createSimpleLineChart = (
   const metricName = styles.valueAxes?.[0]?.title?.text || yAxisColumn?.name;
   const dateName = styles.categoryAxes?.[0]?.title?.text || xAxisColumn?.name;
   const layers: any[] = [];
+
+  const getTooltipFormat = (axis: typeof xAxisColumn, fallback: string): string => {
+    return dateField
+      ? timeUnitToFormat[inferTimeUnitFromTimestamps(transformedData, dateField)] ?? fallback
+      : fallback;
+  };
 
   const mainLayer = {
     mark: buildMarkConfig(styles, 'line'),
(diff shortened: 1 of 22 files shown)


src/plugins/explore/public/components/visualizations/line/to_expression.ts
```diff
@@ -390,6 +429,12 @@ export const createFacetedMultiLineChart = (
   // Create a mark config for the faceted spec
   const facetMarkConfig = buildMarkConfig(styles, 'line');
 
+  const getTooltipFormat = (axis: typeof xAxisMapping, fallback: string): string => {
```
wanglam: Can we separate a `getTooltipFormat` to the `utils.ts`? Something like below:
```
const getTooltipFormat = (  data: Array<Record<string, any>>,
  field: string, fallback = '%b %d, %Y %H:%M:%S'): string => {
```

lunalzm: Sure! I'll exact this logic in `utils.ts`.

src/plugins/explore/public/components/visualizations/utils/utils.ts
```diff
@@ -225,3 +225,72 @@ export const getSchemaByAxis = (
...
+ * @param field The field name for the temporal axis
+ * @param fallback The fallback format string (default: '%b %d, %Y %H:%M:%S')
+ * @returns The format string for the tooltip
+ */
+export const getTooltipFormat = (
+  data: Array<Record<string, any>>,
+  field: string | undefined, // Updated to accept string | undefined
+  fallback = '%b %d, %Y %H:%M:%S'
+): string => {
+  const timeUnit = inferTimeUnitFromTimestamps(data, field || ''); // Use empty string if field is undefined
```
wanglam: What about changing the field type to `string | undefined` in `inferTimeUnitFromTimestamps`? This function already handles the case where `field` is `undefined` in the function body.

lunalzm: Ok, I'll do that!

lunalzm: I've changed this logic in this commit.
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10358/commits/8b0b34b8b46df93486e23ef859b12d66e70bccbb

wanglam (approved):

raintygao (approved):

PR #10362, opened 2025-08-07
[Explore] Introduce facet filter

Maosaic:
### Description

What are facets?

To help users quickly narrow down their search data, facets will provide access filtering commonly understood fields. Currently, we are introducing 3 default facet field(`serviceName`/`span.attributes.http@status_code`/`status.code`).

This PR introduces the UI components of facets filter.

## Screenshot

https://github.com/user-attachments/assets/73bc6a9a-8e5f-4340-90bf-d21b96245f8b

## Changelog

- feat: Introduce facet filter

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/10362.yml
+++ b/changelogs/fragments/10362.yml
@@ -1,0 +1,2 @@
+feat:
+- Introduce facet filter ([#10362](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10362))
--- a/src/plugins/explore/public/components/fields_selector/discover_sidebar.scss
+++ b/src/plugins/explore/public/components/fields_selector/discover_sidebar.scss
@@ -56,9 +56,35 @@
   }
 }
 
+.exploreSideBar__facetField {
+  width: 100%;
+
+  /* extra indentation for faceted field */
+  /* stylelint-disable-next-line @osd/stylelint/no_modifying_global_selectors */
+  .euiButtonEmpty__content {
+    font-size: $euiFontSizeXS;
+    font-weight: $euiFontWeightSemiBold;
+    justify-content: flex-start;
+    padding-left: $euiSizeM + 2px + $euiSizeM + $euiSizeS;
+  }
+}
+
 .exploreSideBar__item {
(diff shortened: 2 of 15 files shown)


TackAdam:
1. Should we be hiding this facet field when on the logs flavor? We only want this displayed while on traces
Screenshot 2025-08-07 at 11 39 39 AM
2. For the status.code section is there a way we can alias the values to be as follows

```
0 = Unset
1 = Ok
2 = Error
```
https://github.com/open-telemetry/opentelemetry-specification/blob/v1.46.0/specification/trace/api.md#set-status
Screenshot 2025-08-07 at 11 41 01 AM
3. I don't see the facet for "attributes.http.status_code" showing up by default
Screenshot 2025-08-07 at 11 45 51 AM
4. I think if we are able to alias the name it would make sense for status.code to be "OTel.status" and for attributes.http.status_code to just be "http.status" and serviceName should just be "Services (x)"
5. Can we adjust the order so the two error fields show up first and then the "Services (x)"

TackAdam:
Offline discussion. LGTM just pending switch from 'span.attributes.http@status_code' to 'attributes.http.status_code'

TackAdam (approved):

Maosaic:
> 1. Should we be hiding this facet field when on the logs flavor? We only want this displayed while on traces
> 
> 2. For the status.code section is there a way we can alias the values to be as follows
> ```
> 0 = Unset
> 1 = Ok
> 2 = Error
> ```
> 
> 3. I don't see the facet for "attributes.http.status_code" showing up by default 
>
> 4. I think if we are able to alias the name it would make sense for status.code to be "OTel.status" and for attributes.http.status_code to just be "http.status" and serviceName should just be "Services (x)" 
>
>5. Can we adjust the order so the two error fields show up first and then the "Services (x)"

1#&4# will need more discussion, 2# can be achieved by index pattern field mapping, updated the mapping for 3#, 5# the default order is sort by value hit count, also updated the UI to display count.

angle943 (approved):

ananzh (approved):

PR #10363, opened 2025-08-07
Fix Error Highlighting after switching languages

LDrago27:
### Description

Currently if we switch from PPL to AI mode and then back to PPL, the syntax highlighting does n't work. This fixes the issue.

## Screenshot

https://github.com/user-attachments/assets/6a2c032e-6eab-4ed9-a6b7-b875e620c15c

## Changelog

- fix: Error Highlighting after switching languages

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/packages/osd-monaco/src/ppl/language.ts
+++ b/packages/osd-monaco/src/ppl/language.ts
@@ -7,7 +7,7 @@ import { monaco } from '../monaco';
 import { ID, PPL_TOKEN_SETS } from './constants';
 import { registerWorker } from '../worker_store';
 import { PPLWorkerProxyService } from './worker_proxy_service';
-import { getPPLLanguageAnalyzer } from './ppl_language_analyzer';
+import { getPPLLanguageAnalyzer, PPLValidationResult } from './ppl_language_analyzer';
 // @ts-ignore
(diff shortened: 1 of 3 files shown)


packages/osd-monaco/src/ppl/language.ts
```diff
@@ -119,139 +119,114 @@ const setupPPLTokenization = () => {
...
-      disposables.forEach((d) => d.dispose());
-      monaco.editor.setModelMarkers(model, OWNER, []);
-    });
-    modelHandlers.clear();
-
-    // Clean up global disposables
-    allDisposables.forEach((d) => d.dispose());
+    disposables.forEach((d) => d.dispose());
     pplWorkerProxyService.stop();
   };
```
angle943, line 231: do we need to make changes to the other languages as well?

LDrago27: For other languages we don't have the Error listeners hooked up and they work based on the monarch tokenizer. So technically we don't need to do anything right now for other languages.

But definitely if we want to introduce new language that are hooked up to custom parser and wanted syntax highlighting. We can extract some of the logic into a shared entity , but the overall framework remains the same.

abbyhu2000 (approved):

PR #10366, opened 2025-08-07
[Explore] reset legacy state when changing dataset

angle943:
### Description

- i got a bug ticket where if you are on one dataset, you select some columns, switch dataset, and download CSV, the old selected columns appear
- This was because we were not resetting the legacy state in redux. This resets it

## Changelog
- fix: Reset legacy state when switching dataset

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/utils/state_management/middleware/dataset_change_middleware.ts
+++ b/src/plugins/explore/public/application/utils/state_management/middleware/dataset_change_middleware.ts
@@ -4,6 +4,7 @@
  */
 
 import { Middleware } from '@reduxjs/toolkit';
+import { AnyAction } from 'redux';
 import { isEqual } from 'lodash';
 import { Dataset, DEFAULT_DATA } from '../../../../../../data/common';
 import { RootState } from '../store';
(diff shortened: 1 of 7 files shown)


src/plugins/explore/public/application/utils/state_management/middleware/dataset_change_middleware.ts
```diff
@@ -57,6 +59,7 @@ export const createDatasetChangeMiddleware = (
       store.dispatch(clearLastExecutedData());
       store.dispatch(setPatternsField(''));
       store.dispatch(setUsingRegexPatterns(false));
+      store.dispatch((resetLegacyStateActionCreator(services) as unknown) as AnyAction);
```
angle943, line 62: this is really not ideal to just tack on a list of dispatches here, but i want to spend time when we free up time to create a clean action creator for places like this where we do this, but not now. lets let the app bake a bit and come up with a clean refactor later

Maosaic (approved):

PR #10385, opened 2025-08-08
[Explore] Default to AI mode when available instead of PPL

ananzh:
### Description

Default to AI mode when available instead of PPL
Default to PPL mode when no AI available

### Issues Resolved

NA

## Screenshot

https://github.com/user-attachments/assets/6f89af41-092b-4c49-a2ad-9f4a0257c803

Switching dataset

https://github.com/user-attachments/assets/faf6503a-2b55-490a-a4e5-cec0f307660b

Test
image

## Changelog

- feat: [Explore] Default to AI mode when available instead of PPL

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/ai_editor.spec.js
+++ b/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/ai_editor.spec.js
@@ -89,18 +89,18 @@ const runAiEditorTests = () => {
         cy.explore.setDataset(config.dataset, DATASOURCE_NAME, config.datasetType);
         setDatePickerDatesIfRelevant(config.language.name);
 
-        cy.getElementByTestId('queryPanelFooterLanguageToggle').contains('PPL');
+        cy.getElementByTestId('queryPanelFooterLanguageToggle').contains('AI');
 
         // Test via keyboard clicks
-        cy.explore.setQueryEditor(' ');
-        cy.getElementByTestId('queryPanelFooterLanguageToggle').contains('AI');
         cy.explore.setQueryEditor('{esc}');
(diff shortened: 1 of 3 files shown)


src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
```diff
@@ -270,7 +271,7 @@ const getPreloadedQueryEditorState = async (
     },
     promptModeIsAvailable,
     promptToQueryIsLoading: false,
-    editorMode: DEFAULT_EDITOR_MODE,
+    editorMode: promptModeIsAvailable ? EditorMode.Prompt : DEFAULT_EDITOR_MODE,
```
kavilla, line 274: doesn't the redux state get wiped out on each dataset change. unfortunately i think it's the current problem with the implementation pattern.

there's not singular place where this logic is occurring. we do have middleware. like is it expected to default AI for every dataset change?

ananzh: https://github.com/user-attachments/assets/9ef6c2d7-9b78-4be6-a151-253811f98300

This is the current behavior. The middleware doesn't reset the editor mode on dataset changes. So if change dataset and if user set to PPL, then it won't default to AI mode.

ananzh: I could submit a follow up PR to update the middleware. I think we need to handle two places. Any idea for refactor to make it to a singular place?

LDrago27 (approved):

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
angle943: according to the comment in slack thread, we only want AI mode as default on initialization, not on dataset change

PR #10388, opened 2025-08-09
[Explore] Preserve active tab when saving and loading explores in dashboards

ananzh:
## Issues 
When users save an explore visualization from the "Visualization" or "Patterns" tab and then load it in a dashboard, the explore always displays the "Logs" tab instead of preserving the originally active tab.

#### The Root Cause

The issue was in the save process where the system was saving `tabDefinition.id` instead of the current `activeTabId` from the UI state. This meant that regardless of which tab was active when saving, the saved explore would always default to the tab definition's ID rather than the user's actual selection.

#### The fix
In this PR, we modified the save process to capture and use the current `activeTabId` from the UI state

## Screenshot

https://github.com/user-attachments/assets/fdf23abb-d0dd-42f1-bf0b-daf67dde2951

## Changelog

- fix: [Explore] Preserve active tab when saving and loading explores in dashboards

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/top_nav/top_nav.tsx
+++ b/src/plugins/explore/public/components/top_nav/top_nav.tsx
@@ -106,6 +106,7 @@ export const TopNav = ({ setHeaderActionMenu = () => {}, savedExplore }: TopNavP
         tabState,
         flavorId,
         tabDefinition,
+        activeTabId: uiState.activeTabId,
       },
       clearEditors,
       savedExplore
@@ -120,6 +121,7 @@ export const TopNav = ({ setHeaderActionMenu = () => {}, savedExplore }: TopNavP
     flavorId,
     tabDefinition,
     clearEditors,
+    uiState.activeTabId,
   ]);
 
   useEffect(() => {
--- a/src/plugins/explore/public/components/top_nav/top_nav_links/get_top_nav_links.ts
+++ b/src/plugins/explore/public/components/top_nav/top_nav_links/get_top_nav_links.ts
@@ -25,6 +25,7 @@ export const getTopNavLinks = (
     tabState: TabState;
     flavorId: string | null;
     tabDefinition: TabDefinition | undefined;
+    activeTabId: string;
   },
   clearEditors: ReturnType<typeof useClearEditors>,
   savedExplore?: SavedExplore
--- a/src/plugins/explore/public/saved_explore/transforms.ts
+++ b/src/plugins/explore/public/saved_explore/transforms.ts
@@ -31,7 +31,8 @@ export const saveStateToSavedObject = (
   flavorId: string,
   tabDefinition: TabDefinition,
   visState?: VisState,
-  dataset?: IndexPattern | Dataset
+  dataset?: IndexPattern | Dataset,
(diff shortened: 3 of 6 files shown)


src/plugins/explore/public/components/top_nav/top_nav.tsx
```diff
@@ -106,6 +106,7 @@ export const TopNav = ({ setHeaderActionMenu = () => {}, savedExplore }: TopNavP
         tabState,
         flavorId,
         tabDefinition,
+        activeTabId: uiState.activeTabId,
```
kavilla, line 109: what's the difference between flavorId , tabState, and tabDefinition? it seems like tabDefinition is just set by uiState.activeTabId as well

kavilla: oh was it because tabdefinition was always defined as logs or something? if so can we just get rid of tabDefinition and pass down tabId

src/plugins/explore/public/components/top_nav/top_nav_links/get_top_nav_links.ts
```diff
@@ -25,6 +25,7 @@ export const getTopNavLinks = (
     tabState: TabState;
     flavorId: string | null;
     tabDefinition: TabDefinition | undefined;
+    activeTabId: string;
```
kavilla, line 28: tab definition is a union type with `undefined` and flavorId is unioned typed with `null`.

i'm not sure if flavorId is now redudant information but if those are null and undefined then activeTabId will def be. so i think we should just make those types more defined. if we dont have a tabDefinition or flavorid why even render something. just go to a default experience

src/plugins/explore/public/saved_explore/transforms.ts
```diff
@@ -43,8 +44,9 @@ export const saveStateToSavedObject = (
     params: visState?.styleOptions ?? {},
     axesMapping: visState?.axesMapping,
   });
+
   obj.uiState = JSON.stringify({
-    activeTab: tabDefinition.id,
+    activeTab: activeTabId || tabDefinition.id,
```
kavilla, line 49: nit: and unrelated to this pr but given that this all gets serialized into data that is stored. as a developer i would understand that the `tab` in uiState would be the `activeTab`. since i dont think we have any other mention of tab in the state. we could even probably have been even more greedy and just said `t` was for t.

but the less picky we get with what we go with the more cost there is incurred to the user. since it is a saved object and will get stored somewhere

kavilla: by now that im looking at the saved object why didn't we didn't we just mimic the visualizatoin saved object and the search object?

src/plugins/explore/public/components/top_nav/top_nav_links/get_top_nav_links.ts
Maosaic: FYI, I'm working on a change that actually utilize flavorId, so each tab will only be shown in the flavor they have registered

abbyhu2000 (approved):

PR #10391, opened 2025-08-11
[Explore] Remove Explore from Patterns tab

ananzh:
### Description

Remove Explore from Patterns tab

Screenshot 2025-08-11 at 9 07 03 AM

### Issues Resolved

NA

## Changelog

- fix: [Explore] Remove Explore from Patterns tab

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/10391.yml
+++ b/changelogs/fragments/10391.yml
@@ -1,0 +1,2 @@
+fix:
+- [Explore] Remove Explore from Patterns tab ([#10391](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10391))


paulstn (approved):
Removing it for now is a good idea. Making the change to have Export work for Patterns tab requires some significant effort. In the current export, the table is pulled straight from the results, depending on the columns selected by the user.
But this framework doesn't work easily with the patterns table, since we do some post processing of the results. We would need to add in separate export to csv functionality specifically for patterns and create some way for tabs to change their export functionality in an extensible way.

src/plugins/explore/public/components/tabs/action_bar/results_action_bar/results_action_bar.tsx
```diff
@@ -38,6 +38,7 @@ export const DiscoverResultsActionBar = ({
 }: DiscoverResultsActionBarProps) => {
   const currentTab = useSelector(selectActiveTabId);
   const shouldShowAddToDashboardButton = currentTab !== 'explore_patterns_tab';
+  const shouldShowExportButton = currentTab !== 'explore_patterns_tab';
```
paulstn, line 41:
```suggestion
  const shouldShowExportButton = currentTab !== 'explore_patterns_tab'; // TODO: patterns tab should soon support exporting, this flag should be removed after this happens
```

LDrago27 (approved):

PR #10386, opened 2025-08-08
Add Correlations Plugin - Saved Object Type Registration

sumukhswamy:
### Description

This PR introduces a new plugin called "correlations" that registers a saved object type for storing correlation. 
This plugin provides a foundation for managing correlations between traces and logs

Object added for Correlations

```
{
  "id": "trace-to-logs-correlation-5678",
  "type": "correlations",
  "migrationVersion": "1.0.0",
  "updatedAt": "2025-07-16T10:15:30.000Z",
  "attributes": {
    "type": "APM-Correlation",
    "entities": [
      {
        "tracesDataset":{
          "id":"refernces[0].id"
        }
      },
      {
        "logsDataset":{
          "id":"refernces[1].id",
          "meta": {
            "logServiceNameField": "service.name",
            "logSpanIdField": "span_id",
            "logTraceIdField": "trace_id",
            "timestamp": "@timestamp"
          }
        }
      },
      {
        "<Plugin-name>DataConnection":{
          "id": "refernces[2].id"
        }
      }
    ]
  },
  "references": [
    {
      "name": "entities[0].index",
      "type": "index-pattern",
      "id": "6ff30050-72ec-11f0-b9fd-951b6e375a38"
    },
    {
      "name": "entities[1].index",
      "type": "index-pattern",
      "id": "72048c50-731a-11f0-8fdf-87b448b7eb0b"
    },
    {
      "name": "<plugin>DataConnection",
      "type": "data-connection",
      "id": "1234567890"
    }
  ]
}

```

## Changelog

-feat: adds a correlation plugin which creates a saved object for the related traces and logs

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #10393, opened 2025-08-11
Trace Details: Log correlation (tabs+redirect)

TackAdam:
### Description
Utilize correlation object introduced in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10386 to determine if a trace dataset has an associated log dataset. If it does, render the logs relevant to the trace and allow re-direction with context to Discover:Logs.

LogsTab

Details outlined here for creating the object and testing: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10386#issuecomment-3176321047

## Screenshot

https://github.com/user-attachments/assets/25767599-8076-4de1-858c-4de765efb568

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/application/pages/traces/trace_details/public/logs/correlation_service.ts
```diff
@@ -0,0 +1,218 @@
...
+}
+
+export interface IndexPatternAttributes {
+  title: string;
+  timeFieldName?: string;
+  type?: string;
+}
+
+export class CorrelationService {
+  constructor(private savedObjectsClient: SavedObjectsClientContract) {}
```
angle943: what is this witchcraft? o.O I didn't know this was a thing. Is this some pattern that C# has that typescript brought over, or is this pretty standard across OO languages like Java as well?

src/plugins/explore/public/application/pages/traces/trace_details/public/logs/correlation_service.ts
```diff
@@ -0,0 +1,218 @@
...
+
+    try {
+      const logDatasets = await this.checkCorrelationsForLogs(dataset);
+
+      if (logDatasets.length > 0) {
+        // Fetch logs for the trace using the first log dataset
+        const logsResponse = await fetchTraceLogsByTraceId(data, {
+          traceId,
+          dataset: logDatasets[0], // Use first log dataset
+          limit: 1000,
```
angle943: why 1000? We have this UI setting called sampleSize, which defaults to 500. Should we use that here too? `services.uiSettings.get(SAMPLE_SIZE_SETTING)`

src/plugins/explore/public/application/pages/traces/trace_details/public/logs/logs_data_table.tsx
```diff
@@ -0,0 +1,194 @@
...
+      setPageSize(page.size);
+    }
+    if (sort) {
+      setSortField(sort.field);
+      setSortDirection(sort.direction);
+    }
+  };
+
+  if (isLoading) {
+    return <div>Loading logs...</div>;
```
angle943: is this temp? Should we use i18n?

src/plugins/explore/public/application/pages/traces/trace_details/public/logs/logs_data_table.tsx
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  const tableProps = {
+    items: sortedLogs,
+    columns,
+    loading: isLoading,
+    tableLayout: 'auto' as const,
+    'data-test-subj': 'logs-data-table',
```
angle943: should we do something like `exploreTracesLogsDataTable` or something? I don't know what prefix convention we have been following for Traces, but we should prefix this data-test-subj

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
```diff
@@ -445,6 +487,16 @@ export const TraceDetails: React.FC<TraceDetailsProps> = ({
                               activeView={activeTab}
                             />
                           )}
+
+                          {activeTab === 'logs' && (
```
angle943: is there an enum we can use for this?

Maosaic:
I'm a bit concern with the current folder structure, it seems `traces detail` is a pretty large component which is more similar like a plugin, do we want to put it deep nested inside explore? Also not quite understand the split logic between `trace` `log` and `server`, I suggest to create folders for main components on the page.

src/plugins/explore/public/application/pages/traces/trace_details/public/logs/correlation_service.ts
TackAdam: The syntax constructor(private savedObjectsClient: SavedObjectsClientContract) is a TypeScript feature called parameter properties (constructor shorthand).

What it does:

    Declares a private property savedObjectsClient on the class
    Automatically assigns the constructor parameter to that property
    Sets the property’s type to SavedObjectsClientContract

Equivalent to writing:
```
export class CorrelationService {
  private savedObjectsClient: SavedObjectsClientContract;
  constructor(savedObjectsClient: SavedObjectsClientContract) {
    this.savedObjectsClient = savedObjectsClient;
  }
}
```
Language comparison:

    TypeScript: has this feature (parameter properties)
    JavaScript: no equivalent (requires manual property assignment)
    C#: similar ideas exist (e.g., records/primary-ctor-like patterns), but this is not the source
    Java: no equivalent (manual declaration + assignment)

src/plugins/explore/public/application/pages/traces/trace_details/public/logs/correlation_service.ts
TackAdam: Good idea, 500 default should be more than enough, updated.
```
    try {
      const logDatasets = await this.checkCorrelationsForLogs(dataset);

      if (logDatasets.length > 0) {
        const sampleSize = this.uiSettings.get(SAMPLE_SIZE_SETTING);
        
        // Fetch logs for the trace using the first log dataset
        const logsResponse = await fetchTraceLogsByTraceId(data, {
          traceId,
          dataset: logDatasets[0], // Use first log dataset
          limit: sampleSize,
        });
```

src/plugins/explore/public/application/pages/traces/trace_details/public/logs/logs_data_table.tsx
TackAdam: Might be replaced, but adding i18n for time being, good catch. Thanks

src/plugins/explore/public/application/pages/traces/trace_details/public/logs/logs_data_table.tsx
TackAdam: Updated too 
```
 'data-test-subj': 'trace-logs-data-table',
```
Will re-visit adjusting all data-test-subj after discussion on if we are moving the components regarding @Maosaic comment.

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
TackAdam: Sure added constant folder and file for

```
export enum TraceDetailTab {
  TIMELINE = 'timeline',
  SPAN_LIST = 'span_list',
  TREE_VIEW = 'tree_view',
  SERVICE_MAP = 'service_map',
  LOGS = 'logs',
}

export type TraceDetailTabType = TraceDetailTab;
```

TackAdam:
@Maosaic 
Thanks for calling this out, I agree trace details has grown into a feature-sized area.

I’m open to moving it as a follow-up once we align on where it should live. For this PR I kept it under explore to keep scope contained, but I’m happy to open an issue/RFC to decide the final home and do the move in a clean follow-up.

traces: all span processing + rendering (timeline/tree/Gantt/service map integrations, client state).
logs: correlation + logs table. This uses a different PPL request and needs contextual inputs, so I kept it separate from span rendering
server: thin wrappers for the PPL requests that fetch raw data for the page; keeps request/response shapes isolated from UI.

I’ve tried to group by responsibility and keep components that change together in the same slice. I’m open to adjusting if come up with a more intuitive structure.

angle943 (approved):

Maosaic (approved):

PR #10394, opened 2025-08-11
[Explore] Patterns field fixes

paulstn:
### Description

Fixes:
- re-enables patterns tab
- an issue with datasets of type INDEXES would not have options appear in the pattern field options because of an issue with the types of fields
- an issue where changing the patterns field back to a field already computed did not update the table
- the patterns field selector being a gear instead of showing up as "Patterns Field: {pattern field}"

## Screenshot

Old:
image

New (ignoring styling):
image

## Changelog

- fix: Functional issues with patterns field selector

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/register_tabs.ts
+++ b/src/plugins/explore/public/application/register_tabs.ts
@@ -11,6 +11,7 @@ import {
   EXPLORE_DEFAULT_LANGUAGE,
   EXPLORE_LOGS_TAB_ID,
   EXPLORE_VISUALIZATION_TAB_ID,
+  EXPLORE_PATTERNS_TAB_ID,
 } from '../../common';
 import { VisTab } from '../components/tabs/vis_tab';
 import { getQueryWithSource } from './utils/languages';
@@ -54,27 +55,27 @@ export const registerBuiltInTabs = (
     flavor: [ExploreFlavor.Logs, ExploreFlavor.Metrics],
     order: 15,
     supportedLanguages: [EXPLORE_DEFAULT_LANGUAGE],
-  
+
     prepareQuery: (query) => {
       const state = services.store.getState();
-  
+
       // Get the selected patterns field from the Redux state
       let patternsField = state.tab.patterns.patternsField;
-  
+
       const preparedQuery = getQueryWithSource(query);
       if (!patternsField) {
         patternsField = findDefaultPatternsField(services);
       }
-  
+
       if (state.tab.patterns.usingRegexPatterns)
         return regexPatternQuery(preparedQuery.query, patternsField);
-  
+
       return brainPatternQuery(preparedQuery.query, patternsField);
     },
-  
+
     handleQueryError: (error, cacheKey) => {
       const state = services.store.getState();
-  
+
       /**
        * The below conditional is checking for the error returned when attempting to use a BRAIN
        * query on an older version of the querying engine. If this error appears, an attempt is made
@@ -104,7 +105,7 @@ export const registerBuiltInTabs = (
             cacheKey: regexPatternQuery(preparedQuery.query, patternsField),
           })
         );
-  
+
         // set the old cacheKey to uninitialized to finalize loading, our new tab query has new cacheKey
         services.store.dispatch(
           setIndividualQueryStatus({
@@ -117,13 +118,13 @@ export const registerBuiltInTabs = (
             },
           })
         );
-  
+
         return true;
       }
-  
+
       return false;
     },
-  
+
     component: PatternsTab,
   });
 


src/plugins/explore/public/application/utils/hooks/use_tab_results.ts
```diff
@@ -19,11 +18,9 @@ export const useTabResults = () => {
...
   const results = useSelector((state: RootState) => state.results);
 
-  const cacheKey = useMemo(() => {
-    const activeTab = services.tabRegistry.getTab(activeTabId);
-    const prepareQuery = activeTab?.prepareQuery || defaultPrepareQueryString;
-    return prepareQuery(query);
-  }, [query, activeTabId, services]);
+  const activeTab = services.tabRegistry.getTab(activeTabId);
+  const prepareQuery = activeTab?.prepareQuery || defaultPrepareQueryString;
+  const cacheKey = prepareQuery(query);
```
paulstn: Getting rid of the useMemo here
Everything in the dependency array was not enough to update the cacheKey. The (real/undiscovered) cacheKey could change due to something outside of query, activeTab, and service. In my case it was due to redux state. The user sets the patternField to something else. The intended cacheKey will change. I don’t anticipate this being a performance hit because prepareQuery is only meant to create a string.

src/plugins/explore/public/components/tabs/action_bar/results_action_bar/results_action_bar.scss
```diff
@@ -4,7 +4,7 @@
   padding: $euiSizeXS 0;
 
   &__explore-download-csv-flex-item {
-    margin-left: auto;
+    margin-left: 10px;
```
paulstn: I've changed the React here to be the one aligning items, instead of requiring that "Export" be the thing aligning all the elements to the right.

src/plugins/explore/public/components/tabs/action_bar/results_action_bar/results_action_bar.tsx
```diff
@@ -77,30 +77,35 @@ export const DiscoverResultsActionBar = ({
...
-              {shouldShowAddToDashboardButton && (
-                <EuiFlexItem grow={false}>
-                  <SaveAndAddButtonWithModal dataset={dataset} />
-                </EuiFlexItem>
-              )}
-            </>
-          ) : null}
+          <EuiFlexItem grow={false}>
+            {dataset && rows?.length ? (
+              <EuiFlexGroup
```
paulstn, line 82: Created another FlexGroup here as to group together all the elements on the right side of the action bar. So that the above FlexGroup can use "spaceBetween" instead.

src/plugins/explore/public/application/utils/hooks/use_tab_results.ts
angle943: i'm okay with removing the `useMemo` but i'm moreso concerned with why it doesn't work with the `useMemo`. Shouldn't `query` change when you change the `patternField`, because you are changing the query by doing that?

I am testing this locally and notice that the `query.query` does not change when switching onto the patterns tab. I think that is a bug. The query slice is supposed to stay in sync with the query object from the data plugin (minus the omission of "source = *") and I think that is no longer the case.

Solvingfor that should make the `useEffect` work again

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
```diff
@@ -3,26 +3,39 @@
...
+import { EuiButtonEmpty, EuiPopover } from '@elastic/eui';
 import React, { useState } from 'react';
+import { FormattedMessage } from '@osd/i18n/react';
+import { useSelector } from 'react-redux';
 import { PatternsSettingsPopoverContent } from './patterns_settings_popover_content';
+import { selectPatternsField } from '../../../../application/utils/state_management/selectors';
 
 export const PatternsSettingsPopoverButton = () => {
   const [isPopoverOpen, setIsPopoverOpen] = useState(false);
+  const patternsField = useSelector(selectPatternsField) ?? '';
```
angle943: lets not do this `?? ''` because that is inaccurate (i don't think we ever want this button to show `Patterns field: `.

I am thinking about how to best handle this... My initial thought was to have a 

```
if (!patternsField) {
   return null;
}
```

before the `return (<EuiPopover ....`, but what if for some reason we get to this tab with the patternsField being undefined? Shouldn't they have a way to manually select a patterns field?

I think the best solution is:

1. Remove the `?? '';` in line 15 to avoid inaccuracies
2. Put a ternary condition in the child of `EuiButtonEmpty`:

```
{patternsField ? <FormattedMessage
            id="explore.discover.patterns.settings.fieldSelector"
            defaultMessage="Patterns field: {patternsField}"
            values={{
              patternsField,
            }}
          /> : 
         // handling the unlikely scenario where patternsField is undefined
         <FormattedMessage
            id="explore.discover.patterns.settings.fieldSelectorNoSelection"
            defaultMessage="Select patterns field"
          />
}
```

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
```diff
@@ -3,26 +3,39 @@
...
+            id="explore.discover.patterns.settings.fieldSelector"
+            defaultMessage="Patterns field: {patternsField}"
+            values={{
+              patternsField,
+            }}
+          />
+        </EuiButtonEmpty>
       }
       isOpen={isPopoverOpen}
       closePopover={closePopover}
```
angle943, line 48: should we make the popover appear to the leftDown, since this appears so much on the right side of screen? See Popover prop `anchorPosition`: https://oui.opensearch.org/1.21/#/layout/popover

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
kavilla: what about if not found just the first field of type keyword you find on the dataset

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
```diff
@@ -3,26 +3,39 @@
...
+  const patternsField = useSelector(selectPatternsField) ?? '';
 
   const openPopover = () => setIsPopoverOpen(true);
   const closePopover = () => setIsPopoverOpen(false);
 
   return (
     <EuiPopover
       button={
-        <EuiButtonIcon
+        <EuiButtonEmpty
```
kavilla, line 23: ah i see we are moving away from just the gear and showing the pattern

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_content.tsx
```diff
@@ -7,7 +7,8 @@ import { EuiCallOut, EuiLoadingSpinner, EuiSelect, EuiText } from '@elastic/eui'
 import React from 'react';
 import { FormattedMessage } from '@osd/i18n/react';
 import { useDispatch, useSelector } from 'react-redux';
-import { DataView } from 'src/plugins/data/common';
+import { DataView } from 'src/plugins/data/common/data_views';
+import { IndexPatternField } from 'src/plugins/data/common';
```
kavilla, line 11: you should be able to use dataviewfield and update the import on line 10

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_content.tsx
```diff
@@ -29,12 +30,12 @@ const gatherOptions = (dataset?: DataView) => {
...
   // Filter out fields that are not suitable for patterns
   // For example, exclude scripted fields or meta fields
-  const filteredFields = fields.filter((field) => {
+  const filteredFields = fields.filter((field: IndexPatternField) => {
     return (
       !field.scripted &&
       !dataset.metaFields.includes(field.name) &&
       !field.subType &&
-      ['text', 'keyword'].includes(field?.esTypes?.[0] ?? '')
+      field.type === 'string'
```
kavilla, line 38: but i dont think if field is unindexed then it shouldn't be selectable. so we should consider checking if text or keyword

src/plugins/explore/public/components/tabs/action_bar/results_action_bar/results_action_bar.scss
kavilla: there should be sass variables that we can use

src/plugins/explore/public/components/tabs/action_bar/results_action_bar/results_action_bar.tsx
```diff
@@ -77,30 +77,35 @@ export const DiscoverResultsActionBar = ({
...
+                  <EuiFlexItem grow={false}>
+                    <PatternsSettingsPopoverButton />
+                  </EuiFlexItem>
+                )}
+                {shouldShowExportButton && (
+                  <EuiFlexItem
+                    grow={false}
+                    className="explore-results-action-bar__explore-download-csv-flex-item"
+                  >
+                    <DiscoverDownloadCsv indexPattern={dataset as any} rows={rows} hits={hits} />
```
kavilla: i don thtink it needs any for dataset. it should be the same

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
kavilla: what about if not found just the first field of type keyword you find on the dataset.

that or or we can make an experimental ui setting for explore plugin that has this value default to `message` if it isn't able to find a better pattern field

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
paulstn: Yes, i put some images in the description

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
paulstn: ah yeah that seems like a good thing to specify, ill do that

paulstn: actually the export button doesn't have it specified, when we reintroduce the export button if might feel weird to have one pop up left and one down. unless we primarily have it go down, which happens with `leftUp`. 

leftDown:
image


leftUp:
image

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_content.tsx
paulstn: sometimes for datasets of type INDEXES it doesn't have the esTypes specified. 
does checking this type (which i believe is the discover type, right?) not work as well?

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_content.tsx
paulstn: What would I need dataviewfield for? The only reason I'm using DataView here is to specify a parameter type

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
angle943: i think there will always be found one, so its not a realistic scenario (correct me if i'm wrong @paulstn )

But its a matter of typing, since patternsField in the type is typed as optional, since in other tabs it will be undefined.

I don't think it should be the responsbility of this component of handling the unlikely case where it is optional, but have a better handling of the unlikely scenario.

Also if we are defaulting to something (such as "first field of type keyword on dataset"), it should be done outside of this component, within redux.

But again, i think this will always be defined in patterns tab due to how its defined, so i'd rather not add extra logic that is not needed

src/plugins/explore/public/components/tabs/action_bar/results_action_bar/results_action_bar.scss
paulstn: Changed to euiSizeS

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
angle943: can you update the export button too? The export button used to be on the left side of the screen so thats prob why

src/plugins/explore/public/components/tabs/action_bar/results_action_bar/results_action_bar.tsx
paulstn: This is just code I moved, I haven't modified it. On the removed side its on line 87

src/plugins/explore/public/application/utils/hooks/use_tab_results.ts
paulstn: Hmm I don't think that I'm changing `query` when `patternsField` changes, from what i know about the design. `query` just contains information about the user specified query and dataset, to the best of my knowledge:
image

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
paulstn: Yes so @angle943 is correct in that ideally, the patterns field would be discovered before this component is being rendered. I'm just checking to make sure that this is actually the case, and there are no potential weird race conditions where it could be undefined.

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
paulstn: do we want the export button to also be `leftUp`?

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
paulstn: After looking through it, I believe that since the activeTab is [set by the same code](https://github.com/paulstn/OpenSearch-Dashboards/blob/e2ca14a10f4edd3bd80e227a5a345b9ebdc45aba/src/plugins/explore/public/components/tabs/tabs.tsx#L37-L40) that calls upon the patterns tab `prepareQuery`, and the redux state activeTab is what [then renders the patterns field selector component](https://github.com/paulstn/OpenSearch-Dashboards/blob/e2ca14a10f4edd3bd80e227a5a345b9ebdc45aba/src/plugins/explore/public/components/tabs/action_bar/results_action_bar/results_action_bar.tsx#L42-L43) here, we can guarantee that there will be a patterns field set before the patterns field selector is rendered (nothing is async). 
So in the unlikely case that this code changes or otherwise, I'm thinking that to be safe I'm just going to go with @angle943 's suggestion where I change the language to just be a more coherent call to action where it says "Select patterns field"

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
angle943: up to you, just try to make both experience consistent and looking good

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
paulstn: from the other discussion thread on the same line, we're going to have the selector read out a 'call to action' in the unlikely case that the field is not found.

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
paulstn: Going to include this when the patterns tab reintroduced the export functionality, its unneeded right now.

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_content.tsx
kavilla: oh that is interesting. can you share examples of which index and field this happens in so i can check if its intended. 

i believe fields should get an esTypes mapping. if not we might have incorrect type mapping or it converted too early. the mappings are defined here: github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/common/osd_field_types/osd_field_types_factory.ts#L44

so [keyword](https://docs.opensearch.org/latest/field-types/supported-field-types/keyword/)  has different functionalities compared to [wildcard](https://docs.opensearch.org/latest/field-types/supported-field-types/wildcard/) and it might not do what you are intending. But we will need to make sure that any type of OSD field type of string can work. in general i think what you have here is ok. I think maybe we should check if it's aggregatable then you should be good

or are you saying the field type might map to something other than text keyword? looking at the original code i see it makes an assumption the first esType is always text or keyword. which im not sure if that is guranteed

src/plugins/explore/public/components/tabs/action_bar/results_action_bar/results_action_bar.tsx
kavilla: gotcha. ignore the following if you don't have any other changes.

but if you need to push more changes do you mind cleaning this part up 🥺

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_content.tsx
kavilla: so `DataViewField` extends `IndexPatternField` like how `DataView` extends `IndexPattern`.

Explore should use `DataView`. You are able to use the IndexPatternField interchangeably because it extends but for simplicity if you are already importing DataView, then you should be able to just do  
```
import { DataView. DataViewField } from 'src/plugins/data/common/data_views';
```

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.test.tsx
```diff
@@ -17,31 +29,57 @@ jest.mock('./patterns_settings_popover_content', () => ({
   )),
 }));
 
+// Test harness component to provide IntlProvider
```
kavilla: nit: can probably remove this comment

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.test.tsx
```diff
@@ -17,31 +29,57 @@ jest.mock('./patterns_settings_popover_content', () => ({
...
+const TestHarness = () => {
```
kavilla: nit: i think our current patter for constants like this is `get${name of component}`

so it would be like `getPatternsSettingsPopoverButton()`

kavilla (approved):

src/plugins/explore/public/application/utils/hooks/use_tab_results.ts
Maosaic: The current way to get `TabDefinition` is through `services.tabRegistry.getTab(activeTabId)`, which return a snapshot, if your `TabDefinition` change, the snapshot you get in the componnet will not get updated.

Your component might have trigger a re-render while `patternField` changed, but I don't think the problem is properlly fixed, because it depends on a implicit side effect.

angle943: what if we re-add the `useMemo` and add `patternField` as a dependency? Might need eslint-ignore (with a comment explaining why) but that might make it more explicit

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_content.tsx
paulstn: It's not really working because I'm getting fields through dataset.fields.getAll(), which gives me a field list of type `IndexPatternField[]`. Is there some other way of getting the fields which will be of type `DataViewField`? Even though these types seem to pretty much be identical, they dont seem to have any relation for me to try and cast it over.

src/plugins/explore/public/components/tabs/action_bar/results_action_bar/results_action_bar.tsx
paulstn: cleaning it up! 😁

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_content.tsx
paulstn: `DataViewField` is unavailable in `DataView`, so will make this change after the above is implemented.

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.test.tsx
paulstn: just for the name of the constant here? i was kind of copying how download_csv.test.tsx worked (incl the name here), since that component is essentially the template for this one.

paulstn: I think there are issues with keeping it like a function, the name being in camelCase is causing linting issues. So I'm keeping the name as `PatternsSettingsPopoverButtonHarness`, to make sure that the `render(<PatternsSettingsPopoverButtonHarness />)` works properly in all the tests.

src/plugins/explore/public/application/utils/hooks/use_tab_results.ts
paulstn: ^doing this, but also adding a `TODO` as to not make it a long term solution. The potential issues with this include:
- other prepareQuery methods by tabs using state and failing to keep their cacheKey updated when the state changes
- keeping a tab specific piece of logic within logic which is meant to be agnostic to the tab implementations

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_content.tsx
paulstn: I didn't know this at the time, but the ESType has no bearing on whether the PPL `patterns` command works or not. It seems to work on fields of type `keyword`, `wildcard`, `text`, etc.
Even knowing if its not aggregatable doesn't really matter, because the PPL `patterns` command would work on it regardless. 
All of this really does just get covered by checking if the field's osd type is `"string"`

src/plugins/explore/public/components/tabs/action_bar/patterns_settings/patterns_settings_popover_button.tsx
paulstn: I've removed this change because it clashes with the supposed direction of the caret. If there is a UI ask for it, I'll reinclude it, but for now, we're not going to have the export button and the patterns field selector appear at the same time, as export is not being included with patterns.

ananzh (approved):

angle943 (approved):

PR #10396, opened 2025-08-12
[Explore] Refactor Language Switch in Query Editor

LDrago27:
### Description

- Refactor Language Switch in Query Editor
- Show suggestions when we have no text in Query Editor

## Screenshot

https://github.com/user-attachments/assets/277c293c-323a-4c79-887e-d3294a5b63e6

## Changelog

- fix: Refactor Language Switch in Query Editor 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/application/hooks/editor_hooks/use_language_switch/use_language_switch.ts
```diff
@@ -0,0 +1,35 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { useCallback } from 'react';
+import { useDispatch } from 'react-redux';
+import { EditorMode } from '../../../utils/state_management/types';
+import { setEditorMode } from '../../../utils/state_management/slices';
+import { useEditorRef } from '../use_editor_ref';
+
+export interface UseLanguageSwitchReturn {
```
angle943: nitpick: instead of `useLanguageSwitch`, `useSwitchLanguage`? `useLanguageSwitch` sounds like yoda speak

src/plugins/explore/public/application/hooks/editor_hooks/use_language_switch/use_language_switch.ts
```diff
@@ -0,0 +1,35 @@
...
+      if (range) {
+        setTimeout(() => editorRef.current?.setSelection(range), 300);
+      }
+    },
+    [dispatch, editorRef]
+  );
+
+  return {
+    switchEditorMode,
+  };
```
angle943: why return an object? can we just have it return `switchEditorMode`? or better yet:

```
return useCallback(...)
```

src/plugins/explore/public/components/query_panel/query_panel_widgets/language_toggle/language_toggle.tsx
```diff
@@ -27,25 +27,21 @@ export const LanguageToggle = () => {
   const promptModeIsAvailable = useSelector(selectPromptModeIsAvailable);
   const isPromptMode = useSelector(selectIsPromptEditorMode);
   const language = useSelector(selectQueryLanguage);
-  const dispatch = useDispatch();
   const editorRef = useEditorRef();
```
angle943: editorRef is no longer needed

src/plugins/explore/public/application/hooks/editor_hooks/use_language_switch/use_language_switch.ts
```diff
@@ -0,0 +1,35 @@
...
+  const dispatch = useDispatch();
+  const editorRef = useEditorRef();
+
+  const switchEditorMode = useCallback(
+    (mode: EditorMode) => {
+      dispatch(setEditorMode(mode));
+      // select all
+      const range = editorRef.current?.getModel()?.getFullModelRange();
+      if (range) {
+        setTimeout(() => editorRef.current?.setSelection(range), 300);
```
kavilla: why a set timeout? should it just set? or are we trying to debounce here?

angle943: we are trying to debounce. The `dispatch(setEditorMode(mode))` should happen first, but dispatches happen later. this settimeout is sort of mimicking the `dispatch` behavior.

Just fyi i initially had this set to 100, but it wasn't working on neo, so i requested suchit to set it to 300.

but we should add a comment here suchit to describe why we are using settimeout. We can also use `debounce` to be more explicit on what we are trying to do, but i think a comment would suffice?

src/plugins/explore/public/components/query_panel/query_panel_widgets/language_toggle/language_toggle.tsx
LDrago27: Thanks Removed it

src/plugins/explore/public/application/hooks/editor_hooks/use_language_switch/use_language_switch.ts
LDrago27: Updated it.

src/plugins/explore/public/application/hooks/editor_hooks/use_language_switch/use_language_switch.ts
LDrago27: Sure.

src/plugins/explore/public/application/hooks/editor_hooks/use_language_switch/use_language_switch.ts
LDrago27: The idea was to have a small delay after the mode switches and then select all the text in the editor till now.

angle943 (dismissed):

src/plugins/explore/public/application/hooks/editor_hooks/use_switch_language/use_switch_language.ts
```diff
@@ -0,0 +1,29 @@
...
+    (mode: EditorMode) => {
+      dispatch(setEditorMode(mode));
+      // select all
+      const range = editorRef.current?.getModel()?.getFullModelRange();
+      if (range) {
+        setTimeout(() => editorRef.current?.setSelection(range), 300);
+      }
+    },
+    [dispatch, editorRef]
+  );
```
Maosaic, line 26: Can we store the timeout ID and clear it before setting a new one?

```
const timeoutRef = useRef<NodeJS.Timeout | null>(null);
...
const switchEditorMode = useCallback(
  (mode: EditorMode) => {
    // Clear existing timeout
    if (timeoutRef.current) {
      clearTimeout(timeoutRef.current);
    }

    dispatch(setEditorMode(mode));
    
    const range = editorRef.current?.getModel()?.getFullModelRange();
    if (range) {
      timeoutRef.current = setTimeout(() => {
        editorRef.current?.setSelection(range);
        timeoutRef.current = null; // Clean up reference
      }, 300);
    }
  },
  [dispatch, editorRef]
);

// Cleanup on unmount
useEffect(() => {
  return () => {
    if (timeoutRef.current) {
      clearTimeout(timeoutRef.current);
    }
  };
}, []);
```

angle943: why? we never want to cancel this right. the `dispatch(setEditorMode()` doesn't cancel, so why should this one?

angle943: either we cancel everything or we should cancel this one. The only diff is the 300 delay, and might lead to some funky behaviors when switching quickly back and forth between modes, but I think thats a niche case that we don't need to over engineer for

PR #10397, opened 2025-08-12
[Keyboard-shortcuts] Add Key Processing & Normalization

natel926:
### Description

This PR implements Phase 2 of the keyboard shortcuts system, providing the core key processing engine that transforms raw keyboard input into standardized, cross-platform key representations.

## Changelog
- feat: Add key Processing & normalization

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/public/keyboard_shortcut/key_parser.ts
+++ b/src/core/public/keyboard_shortcut/key_parser.ts
@@ -104,8 +104,8 @@ type ModifierKey = typeof MODIFIER_ORDER[number];
 type PlatformMappingKeys = 'ctrl' | 'control' | 'meta' | 'win' | 'super' | 'command' | 'cmd';
 
 interface PlatformModifierMappings {
-  readonly mac: Record<PlatformMappingKeys, string>;
-  readonly other: Record<PlatformMappingKeys, string>;
+  readonly mac: Record<PlatformMappingKeys, ModifierKey>;
+  readonly other: Record<PlatformMappingKeys, ModifierKey>;
 }
 
 type DisplayMappingKeys =
@@ -296,32 +296,31 @@ export class KeyStringParser {
       return plusKeyResult;
     }
 
-    const parts = keyString
-      .toLowerCase()
-      .split('+')
-      .map((part) => {
-        if (part === ' ') {
-          return 'space';
-        }
-        return part.trim();
-      })
-      .filter((part) => part.length > 0);
-
+    const rawParts = keyString.toLowerCase().split('+');
     const modifiers: ModifierKey[] = [];
     const keys: string[] = [];
+    const modifierSet = new Set<ModifierKey>();
 
-    for (const part of parts) {
+    for (const rawPart of rawParts) {
+      let part = rawPart.trim();
+
+      if (part === ' ') {
+        part = 'space';
+      }
+
+      if (part.length === 0) continue;
+
       if (this.isModifier(part)) {
         const normalizedModifier = this.normalizeModifier(part);
-        if (normalizedModifier && !modifiers.includes(normalizedModifier as ModifierKey)) {
-          const platformMappedModifier = this.applyPlatformMapping(
-            normalizedModifier
-          ) as ModifierKey;
-          if (!modifiers.includes(platformMappedModifier)) {
+        if (normalizedModifier) {
+          const platformMappedModifier = this.applyPlatformMapping(normalizedModifier);
+
+          if (!modifierSet.has(platformMappedModifier)) {
+            modifierSet.add(platformMappedModifier);
             modifiers.push(platformMappedModifier);
           }
         }
-      } else if (part.length > 0) {
+      } else {
         const normalizedKey = this.normalizeKey(part);
         if (normalizedKey.length > 0) {
           keys.push(normalizedKey);
@@ -329,7 +328,11 @@ export class KeyStringParser {
       }
     }
 
-    this.sortModifiers(modifiers);
+    modifiers.sort((a, b) => {
+      const aIndex = KeyStringParser.MODIFIER_INDEX_MAP.get(a) ?? -1;
+      const bIndex = KeyStringParser.MODIFIER_INDEX_MAP.get(b) ?? -1;
+      return aIndex - bIndex;
+    });
 
     if (modifiers.length > 0 && keys.length === 0) {
       return modifiers.join('+');
@@ -339,7 +342,6 @@ export class KeyStringParser {
       return this.buildKeyString(modifiers, keys[0]);
     }
 
-    // For multiple keys, include modifiers in the result so validation can catch invalid combinations
     if (modifiers.length > 0) {
       return `${modifiers.join('+')}+${keys.join('+')}`;
     }
@@ -401,28 +403,20 @@ export class KeyStringParser {
     return KeyStringParser.VALID_MODIFIERS.has(normalized as ModifierKey) ? normalized : null;
   }
 
-  private applyPlatformMapping(modifier: string): string {
+  private applyPlatformMapping(modifier: string): ModifierKey {
     const mappings =
       this.platform === 'mac'
         ? KeyStringParser.PLATFORM_MODIFIER_MAPPINGS.mac
         : KeyStringParser.PLATFORM_MODIFIER_MAPPINGS.other;
 
-    const normalized = modifier.toLowerCase();
-    return mappings[normalized as keyof typeof mappings] || normalized;
+    const normalized = modifier.toLowerCase() as PlatformMappingKeys;
+    return mappings[normalized] || (normalized as ModifierKey);
   }
 
   private buildKeyString(modifiers: string[], key: string): string {
     if (modifiers.length === 0) return key;
     if (!key) return modifiers.join('+');
     return `${modifiers.join('+')}+${key}`;
-  }
-
-  private sortModifiers(modifiers: ModifierKey[]): void {
-    modifiers.sort((a, b) => {
-      const aIndex = KeyStringParser.MODIFIER_INDEX_MAP.get(a) ?? -1;
-      const bIndex = KeyStringParser.MODIFIER_INDEX_MAP.get(b) ?? -1;
-      return aIndex - bIndex;
-    });
   }
 
   private normalizeKey(key: string): string {
@@ -460,12 +454,12 @@ export class KeyStringParser {
    *
    * @example
    *
-   * parser.isValidKeyString("ctrl+s"); // succeeds
-   * parser.isValidKeyString("ctrl+"); // throws "Malformed key string: invalid '+' character placement"
-   * parser.isValidKeyString("ctrl a"); // throws "Chord sequences are not supported"
+   * parser.validateKeyString("ctrl+s"); // succeeds
+   * parser.validateKeyString("ctrl+"); // throws "Malformed key string: invalid '+' character placement"
+   * parser.validateKeyString("ctrl a"); // throws "Chord sequences are not supported"
    *
    */
-  public isValidKeyString(keyString: string): void {
+  public validateKeyString(keyString: string): void {
     if (keyString === null || keyString === undefined) {
       throw new Error(`Key string cannot be null or undefined`);
     }
@@ -526,17 +520,20 @@ export class KeyStringParser {
       );
     }
 
-    if (keyCount === 2 && modifierCount > 0) {
-      throw new Error(
-        `Malformed key string: two-key sequences cannot have modifiers: "${keyString}"`
-      );
-    }
-
     if (keyCount > 1 && modifierCount > 0) {
+      if (keyCount === 2) {
+        throw new Error(
+          `Malformed key string: two-key sequences cannot have modifiers: "${keyString}"`
+        );
+      }
       throw new Error(
         `Malformed key string: multiple non-modifier keys with modifiers: "${keyString}"`
       );
     }
+  }
+
+  public isValidKeyString(keyString: string): void {
+    return this.validateKeyString(keyString);
   }
 
   public getDisplayString(keyString: string): string {
--- a/src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
+++ b/src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
@@ -52,7 +52,7 @@ export class KeyboardShortcutService {
 
   private register(shortcut: ShortcutDefinition): void {
     try {
-      this.keyParser.isValidKeyString(shortcut.keys);
+      this.keyParser.validateKeyString(shortcut.keys);
     } catch (error) {
       throw new Error(
         `Invalid keyboard shortcut key string: "${shortcut.keys}" for shortcut "${shortcut.id}" in plugin "${shortcut.pluginId}": ${error.message}`
--- a/src/core/public/keyboard_shortcut/key_parser.test.ts
+++ b/src/core/public/keyboard_shortcut/key_parser.test.ts
@@ -134,7 +134,6 @@ describe('KeyStringParser', () => {
     it('should handle space key variations', () => {
       expect(parser.normalizeKeyString('Space')).toBe('space');
       expect(parser.normalizeKeyString('Ctrl+Space')).toBe('ctrl+space');
-      expect(parser.normalizeKeyString('Ctrl+ ')).toBe('ctrl+space');
     });
 
     it('should remove duplicate modifiers', () => {
@@ -607,16 +606,32 @@ describe('KeyStringParser', () => {
       );
     });
 
-    it('should allow valid space key combinations', () => {
-      expect(() => parser.isValidKeyString('ctrl+ ')).not.toThrow();
-      expect(() => parser.isValidKeyString('shift+ ')).not.toThrow();
+    it('should reject implicit space key combinations', () => {
+      expect(() => parser.isValidKeyString('ctrl+ ')).toThrow(
+        'only modifier keys found, no action key'
+      );
+      expect(() => parser.isValidKeyString('shift+ ')).toThrow(
+        'only modifier keys found, no action key'
+      );
+      expect(() => parser.isValidKeyString('alt+ ')).toThrow(
+        'only modifier keys found, no action key'
+      );
+    });
+
+    it('should allow explicit space key combinations', () => {
+      expect(() => parser.isValidKeyString('ctrl+space')).not.toThrow();
+      expect(() => parser.isValidKeyString('shift+space')).not.toThrow();
       expect(() => parser.isValidKeyString(' ')).not.toThrow();
-      expect(() => parser.isValidKeyString('alt+ ')).not.toThrow();
+      expect(() => parser.isValidKeyString('alt+space')).not.toThrow();
     });
 
     it('should handle edge cases with spaces correctly', () => {
-      expect(() => parser.isValidKeyString('ctrl+ ')).not.toThrow();
-      expect(() => parser.isValidKeyString('shift+ ')).not.toThrow();
+      expect(() => parser.isValidKeyString('ctrl+ ')).toThrow(
+        'only modifier keys found, no action key'
+      );
+      expect(() => parser.isValidKeyString('shift+ ')).toThrow(
+        'only modifier keys found, no action key'
+      );
 
       expect(() => parser.isValidKeyString('ctrl a')).toThrow('Chord sequences are not supported');
       expect(() => parser.isValidKeyString('shift+s f1')).toThrow(


src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,459 @@
...
+  '{': 'braceleft',
+  '}': 'braceright',
+  '|': 'pipe',
+  ':': 'colon',
+  '"': 'doublequote',
+  '<': 'less',
+  '>': 'greater',
+  '?': 'question',
+  '~': 'tilde',
+};
```
angle943, line 98: can we update these to be the web keys? https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -115,9 +89,15 @@ export class KeyboardShortcutService {
   private shouldIgnoreKeyboardEventForTarget(target: EventTarget | null): boolean {
     if (!this.isHTMLElement(target)) return false;
 
-    const tagName = target.tagName;
+    const element = target;
+    const tagName = element.tagName;
+
+    if (tagName === 'INPUT' || tagName === 'TEXTAREA' || tagName === 'SELECT') {
```
angle943: should we do `.includes()` here as well, just like what you do for the role?

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -18,6 +13,7 @@ import { KeyboardShortcutSetup, KeyboardShortcutStart, ShortcutDefinition } from
 export class KeyboardShortcutService {
   private shortcutsMapByKey = new Map<string, ShortcutDefinition[]>();
   private namespacedIdToKeyLookup = new Map<string, string>();
+  private keyParser = new KeyStringParser();
```
angle943, line 16: do you need this keyparser? Does keyparser contain any state of it's own? Can it just be used statically?

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -40,37 +36,15 @@ export class KeyboardShortcutService {
...
-
-    if (event.shiftKey) {
-      key += 'shift+';
-    }
-
-    if (event.metaKey) {
-      key += 'cmd+';
+  private register(shortcut: ShortcutDefinition): void {
+    if (!this.keyParser.isValidKeyString(shortcut.keys)) {
+      return;
```
angle943, line 63: lets throw an error here to let devs know they screwed up

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,459 @@
...
+
+    if (trimmed.length === 1) {
+      return trimmed;
+    }
+
+    return trimmed;
+  }
+
+  public isValidKeyString(keyString: string): boolean {
+    if (!keyString || typeof keyString !== 'string') {
```
angle943: lets trust typescript, we don't need `typeof keyString !== 'string'`

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,459 @@
...
+      return false;
+    }
+
+    try {
+      const normalized = this.normalizeKeyString(keyString);
+      const parts = normalized.split('+');
```
angle943: what if they have a ` ` (a space) to try to register a chord (multi-key sequences)? Should we also throw an error for that scenario since we are not going to allow devs to register multi-key sequences?

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,459 @@
...
+
+      if (parts.length === 0) {
+        return false;
+      }
+
+      const lastPart = parts[parts.length - 1];
+      if (this.isModifier(lastPart)) {
+        return false;
+      }
```
angle943: why do we need this check for last part? shouldn't the `for` loop below handle this already?

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,459 @@
...
+      right: '→',
+      plus: '+',
+    },
+  } as const;
+
+  private static readonly MAX_CACHE_SIZE = 1000;
+  private platform: 'mac' | 'windows' | 'linux';
+  private keyStringCache = new Map<string, string>();
+
+  constructor(platform?: 'mac' | 'windows' | 'linux') {
```
angle943: we synced offline, decided we don't need this platform

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,459 @@
...
+   * parser.normalizeKeyString("Ctrl+S") // → "cmd+s"
+   * parser.normalizeKeyString("shift+ctrl+f1") // → "cmd+shift+f1"
+   *
+   * // On Windows/Linux:
+   * parser.normalizeKeyString("Cmd+S") // → "ctrl+s"
+   * parser.normalizeKeyString("ALT+F4") // → "alt+f4"
+   *
+   */
+  public normalizeKeyString(keyString: string): string {
+    if (this.keyStringCache.has(keyString)) {
```
angle943, line 262: nice, i like

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,459 @@
...
+
+    return trimmed;
+  }
+
+  public isValidKeyString(keyString: string): boolean {
+    if (!keyString || typeof keyString !== 'string') {
+      return false;
+    }
+
+    try {
```
angle943: can any of this try block throw an error? do we need this try catch?

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,459 @@
...
+      const lastPart = parts[parts.length - 1];
+      if (this.isModifier(lastPart)) {
+        return false;
+      }
+
+      for (let i = 0; i < parts.length - 1; i++) {
+        if (!this.isModifier(parts[i])) {
+          return false;
+        }
+      }
```
angle943, line 508: we can also do:
 
`
 return parts.every(!this.isModifier);
`

angle943: but wait i'm getting confused, why does every part have to be a modifier for it to be a valid key string?

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,459 @@
...
+    if (this.keyStringCache.has(keyString)) {
+      const value = this.keyStringCache.get(keyString)!;
+      this.keyStringCache.delete(keyString);
+      this.keyStringCache.set(keyString, value);
+      return value;
+    }
+
+    const result = this.computeNormalizedKeyString(keyString);
+
+    if (this.keyStringCache.size >= KeyStringParser.MAX_CACHE_SIZE) {
```
angle943: i don't think this is needed:

- we most likely wont ever get to 1000 of registered keys
- even if we did, its just a map of strings, so not a heavy burden on memory

Lets remove both MAX_CACHE_SIZE and this logic of code, because every code we write, we are responsible for unit testing & maintaining it, so I would like to ship minimum amount of code wherever possible

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,459 @@
...
+
+  private static readonly MAX_CACHE_SIZE = 1000;
+  private platform: 'mac' | 'windows' | 'linux';
+  private keyStringCache = new Map<string, string>();
+
+  constructor(platform?: 'mac' | 'windows' | 'linux') {
+    this.platform = platform ?? this.detectPlatform();
+  }
+
+  private detectPlatform(): 'mac' | 'windows' | 'linux' {
```
angle943: don't think this is needed

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,459 @@
...
+    ['win', 'cmd'],
+    ['super', 'cmd'],
+    ['option', 'alt'],
+    ['command', 'cmd'],
+    ['opt', 'alt'],
+  ]);
+
+  private static readonly VALID_MODIFIERS = new Set(MODIFIER_ORDER);
+
+  private static readonly PLATFORM_MODIFIER_MAPPINGS = {
```
angle943: lets type both the object value of `mac` and `opther` so the keys must be the same across both objects

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,459 @@
...
+      win: 'ctrl',
+      super: 'ctrl',
+      control: 'ctrl',
+      option: 'alt',
+      opt: 'alt',
+    },
+  } as const;
+
+  private static readonly DISPLAY_MAPPINGS = {
+    mac: {
```
angle943, line 189: lets type

src/core/public/keyboard_shortcut/key_parser.ts
natel926: Yes. I was thinking should I add the MAX_CACHE_SIZE as well and totally agree that we won't get to that much registered keys. However, I think we should add clearCache in stop() right?

```
public stop() {
  this.stopEventListener();
  this.shortcutsMapByKey.clear();
  this.namespacedIdToKeyLookup.clear();
  this.keyParser.clearCache();  // <- 
}
```

src/core/public/keyboard_shortcut/key_parser.ts
natel926: Yes. we should throw an error for that scenario. I added the check in validateInput method.

src/core/public/keyboard_shortcut/key_parser.ts
natel926: The isValidKeyString method is a validation function that must always return a  true or false and never throw errors. However, inside this method, we call normalizeKeyString, which is designed to throw errors when it encounters invalid input. To maintain the validation method's contract of always returning a boolean, I use a try-catch block to catch any errors thrown by normalizeKeyString and convert them to false.

src/core/public/keyboard_shortcut/key_parser.ts
angle943: yeah

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,558 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const SPECIAL_KEY_MAPPINGS: Record<string, string> = {
```
angle943, line 6: Lets leave a comment describing what this is used for. Something like:

```
These are the allowed keys this service supports outside of alphanumeric values. The key of this object represents the lowercased value of the `key` that we accept, and the value represents the lowercased value of KeyboardEvent.key
```

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,558 @@
...
+  return: 'enter',
+  backspace: 'backspace',
+  delete: 'delete',
+
+  // Navigation keys
+  home: 'home',
+  end: 'end',
+  pageup: 'pageup',
+  pagedown: 'pagedown',
+  insert: 'insert',
```
angle943, line 26: what do you think about not supporting these navigational keys? Were they in the BRD? Trying to reduce complexity

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,558 @@
...
+  f3: 'f3',
+  f4: 'f4',
+  f5: 'f5',
+  f6: 'f6',
+  f7: 'f7',
+  f8: 'f8',
+  f9: 'f9',
+  f10: 'f10',
+  f11: 'f11',
+  f12: 'f12',
```
angle943, line 44: what do you think about not supporting these function keys? Were they in the BRD? Trying to reduce complexity

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,558 @@
...
+  numpad6: 'numpad6',
+  numpad7: 'numpad7',
+  numpad8: 'numpad8',
+  numpad9: 'numpad9',
+  numpaddecimal: 'numpaddecimal',
+  numpadenter: 'numpadenter',
+  numpadplus: 'numpadplus',
+  numpadminus: 'numpadminus',
+  numpadmultiply: 'numpadmultiply',
+  numpaddivide: 'numpaddivide',
```
angle943, line 97: what do you think about not supporting these numeric kepad keys? Were they in the BRD? Trying to reduce complexity

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,558 @@
...
+
+    const withoutPlusKey = trimmed.slice(0, -2);
+
+    if (withoutPlusKey.length === 0) {
+      return 'plus';
+    }
+
+    const modifierPart = this.computeNormalizedKeyString(withoutPlusKey);
+    return modifierPart ? `${modifierPart}+plus` : 'plus';
+  }
```
angle943, line 291: what if they wrote `++cmd`?

angle943: Also, what if they wrote `alt+g++`? Should that be allowed?

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,558 @@
...
+    }
+
+    const result = this.computeNormalizedKeyString(keyString);
+
+    this.keyStringCache.set(keyString, result);
+
+    return result;
+  }
+
+  private handlePlusKeySpecialCase(keyString: string): string | null {
```
angle943, line 276: I think just `handlePlusKey` is good enough

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,558 @@
...
+
+    if (keys.length === 1) {
+      return this.buildKeyString(modifiers, keys[0]);
+    }
+
+    if (modifiers.length > 0) {
+      return `${modifiers.join('+')}+${keys.join('+')}`;
+    }
+
+    return keys.join('+');
```
angle943, line 349: 1. should we allow `chords`? (ie `g d`)
2. Should we allow multi-keys in one keypress? (ie `cmd+g+a`)

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,558 @@
...
+  readonly other: Record<DisplayMappingKeys, string>;
+}
+
+export class KeyStringParser {
+  private static readonly MODIFIER_INDEX_MAP = new Map([
+    ['ctrl', 0],
+    ['alt', 1],
+    ['shift', 2],
+    ['cmd', 3],
+  ]);
```
angle943, line 139: can we put a comment on top of `MODIFIER_INDEX_MAP` to denote what it is used for?

angle943: Also, some of these constant definitions are outside of the class (like `SPECIAL_KEY_MAPPINGS`) while others are inside the class as private properties. What is the difference?

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,558 @@
...
+    'meta',
+    'win',
+    'super',
+    'control',
+    'option',
+    'command',
+    'opt',
+  ]);
+
+  private static readonly MODIFIER_ALIASES = new Map([
```
angle943, line 155: comment on top of this property plz

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,558 @@
...
+    ['control', 'ctrl'],
+    ['meta', 'cmd'],
+    ['win', 'cmd'],
+    ['super', 'cmd'],
+    ['option', 'alt'],
+    ['command', 'cmd'],
+    ['opt', 'alt'],
+  ]);
+
+  private static readonly VALID_MODIFIERS = new Set(MODIFIER_ORDER);
```
angle943, line 165: a little confusing that you list all modifiers above in MODIFIER_KEYS yet this one states there are only certain VALID_MODIFIERS

src/core/public/keyboard_shortcut/key_parser.ts
natel926: "++cmd" would throw a validation error because the parser only allows "++" at the end of a string.

natel926: "alt+g++" would also be invalid while the "++" at the end is correctly parsed as the plus key (resulting in "alt+g+plus"), it gets rejected because two-key sequences cannot have modifiers.

src/core/public/keyboard_shortcut/key_parser.ts
natel926: Chords (g d) - Currently not allowed, for current implementation, we reject space-separated sequences with "Chord sequences are not supported" error, though it does support some two-key sequences like "g+d" if "g" is in the allowed prefixes.

natel926: Multi-keys (cmd+g+a) - Currently not allowed, the validation rejects multiple non-modifier keys with modifiers, throwing "two-key sequences cannot have modifiers" error.

src/core/public/keyboard_shortcut/key_parser.ts
natel926: I think the complexity is already handled, the parser logic doesn't change whether we support them.

src/core/public/keyboard_shortcut/key_parser.ts
natel926: I will leave a comment describing what this is used for

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -48,48 +44,44 @@ export class KeyboardShortcutService {
...
+      );
     }
 
-    key += this.getNormalizedKey(event.key);
-
-    return key;
-  };
-
-  private register(shortcut: ShortcutDefinition): void {
     if (!this.config.enabled) {
```
angle943, line 62: i think this should be at the top of `register` method

src/core/public/keyboard_shortcut/key_parser.ts
angle943: For example, my laptop computer doesn't have these keys (as well as numpad). But adding these will allow plugin devs to use these, which will make the keyboard shortcut not helpful to users. We cannot easily control what keys the plugin devs decides to implement, but we must ensure that they register keys that would work for all users, and the way we can control that is through our API

VSCode supporting these is one thing, since the configurations are specific to each user, but the keyboard shortcut we are supporting needs to be working not for a specific user, but rather for all users that the app is going to be used for

src/core/public/keyboard_shortcut/key_parser.ts
angle943: I see that logic below now, but I'm getting worried about code maintainability and readabilty, but I think i have a solution (will post in a different comment).

But just addressing your comment, `++cmd` is not allowed, but would `g+cmd` be allowed? I think so based on the code, but this is inconsistent since we are giving special treatment to +

But again, no action item needed here, I think my other comment will address this (look for standalone comment)

angle943 (changes requested):
I have checked out the code and was trying to go through each method, especially in `key_parser`. First off, this was not an easy thing to do so good job on this work!

My gut reaction was that `key_parser` class is very hard to follow and began worrying about maintainability of it, but I understood while reading the code why the code was written this way. But I then began to think that this is too complicated and there has to be an easier way to handle it, and I re-looked at VSCode docs and I think they solved it.

The two complexities I think that comes with this work is in these two areas:

- Handling of the `+` sign
- Handling of modifier keys

The `+` sign is problematic because it is used both for combinations as well as the keypress, and therefore you had to handle special logic to accomodate for it (such as only allowing `++` at the end)

Handling of modifier keys is problematic because of it's conjunction with the `+` sign as well as the order it comes in.

But I think VSCode solved both of these problems here: https://code.visualstudio.com/docs/configure/keybindings#_accepted-keys

- They explicitely mention which keys are allowed, and notice that `+` is not mentioned anywhere. This is because `+` is not itself a keyboard key, but rather it is `shift+=`. We should also have a predefined set of keys, which I think should be a subset of what they have (to make our keyboard shortcuts more accessible for all people, including people without numpads):

- `a-z`, `0-9` (I think we should avoid function keys for accessibility, and it also looks like they don't accept `A-Z`?)
- `, -, =, [, ], \, ;, ', ,, ., / (same as what they have
- `left, up, right, down` (remove the pageup, pagedown, end, home)
- `tab, enter, escape, space, backspace, delete`
- And thats it. Maybe we can even enforce the casing so they don't do something like `Space`? I am not sure if VSCode does that. EDIT: I think we should enforce casing. That way we don't need to worry about normalizing, but furthermore, we will have a table like https://code.visualstudio.com/docs/configure/keybindings#_accepted-keys in our README or HOW-TO somewhere. It'll also make it easier to `cmd+f` on IDE to see if someone is using the key combination that you have registered. Along that line, I think we should restrict registrating of keyboard shortcut to just mac-keyboard shortcut. This way we can have consistent registration system across the codebase, and makes `cmd+f` easier.

Now the problem with this approach is that when user presses (Shift+=) (aka the `+` key), on our `window.addEventListener`, the `e.key` comes back as `+`. This is problematic, since everything i said above is void. But thankfully there is a solve for this, and that is to not use `e.key` but rather `e.code`.

Accoding to MDN: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

```
The KeyboardEvent.code property represents a physical key on the keyboard (as opposed to the character generated by pressing the key). In other words, this property returns a value that isn't altered by keyboard layout or the state of the modifier keys.
```

The second point is how we handle modifier keys. Based on VSCode's documentation, they explicitely say that modifiers are like `Ctrl+`, `Shift+` and so on, instead of saying `Ctrl`, `Shift`, etc.  I think we should also incorporate this (with a twist that i will add later).

But this way, the `plus` sign will be part of the modifier keys, so we don't need to handle the `+` separately.

Now the "twist" that I'm thinking about has to do with the ordering of the modifier keys. And this "twist" is up to you whether you want to implement it or not. Because we need to normalize the keys, we gotta make these two keys the same:

`Cmd+Shift+p` and `Shift+Cmd+p`

The twist proposal is that we hardcode specific keypresses as modifiers. For example, for mac, we will only allow the following modifiers:

```
Ctrl+
Shift+
Alt+
Cmd+
Ctrl+Shift+
Alt+Shift+
Cmd+Shift+
```

So the combination modifiers are considered one modifier, and it has to be registered in that order (with the `Shift+` coming second). We gain two things by doing this:

- We don't have to worry about normalizing the order of the modifier keys
- By enforcing a strict combination of modifier keys, we avoid (IMO weird) key combinations, such as `Cmd+Ctrl+Alt+p` or so on. I think `Shift` plus another modifier is pretty standard, but multi-non-shift-modifiers is less common I think


Conclusion: Making it "harder" for devs to register a keyboard shortcut is not bad as long as we have good documentation (like VSCode does), and can be even good by enforcing consistency across the codebase (such as specific modifier keys and the order, casing of the registering keys, etc).

PR #10413, opened 2025-08-14
Optimize Value Suggestions in Autocomplete

LDrago27:
### Description

Currently for larger datasets the process of fetching column Values is quite slow , making the autocomplete experience quite laggy. This PR introduces the following in a bid to reduce the wait time. 

- We introduce 2 new fields to IndexPattern **autoCompleteValues** and **topAggValues**

Whenever a value suggestion is triggered the following flow is executed
- If we already have results cached in autoCompleteValues we show those values
- If we don't have results cached, we try to showcase the  **Top Five Values** computed by sidebar to user and fire an async query to fetch column values from the API and update the cache in background
- In case we don't have **Top Five Values**, we synchronously fire the query to fetch column values and update the cache

Comparison:
Before:
https://github.com/user-attachments/assets/e69f6e5d-57c3-43c3-93bd-a5b9c8f29229

We were triggering the API call to fetch values each time, and on average in the above scenario was taking around 6s to show values on screen. (Anecdotal evidence based on the testing on above dataset)

After the change:
https://github.com/user-attachments/assets/1a94a59c-888c-4a02-aa5e-f7c6c2f85d2c

Now if we have the value present in the current query results we show the top values from the query result, while updating the cache using the API call. For all subsequent calls we are using the cached values from API call. 
So we are now showing suggestions to user instantaneously, also the API call durations have been slashed to around 1.5s. (Anecdotal evidence based on the testing on above dataset)

## Changelog

- fix: Optimize Value Suggestions in Autocomplete 
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/common/index_patterns/fields/index_pattern_field.ts
+++ b/src/plugins/data/common/index_patterns/fields/index_pattern_field.ts
@@ -148,6 +148,10 @@ export class IndexPatternField implements IFieldType {
     return this.aggregatable && !notVisualizableFieldTypes.includes(this.spec.type);
   }
 
+  public isSuggestionAvailable() {
+    return ([OSD_FIELD_TYPES.STRING, OSD_FIELD_TYPES.BOOLEAN] as string[]).includes(this.spec.type);
+  }
+
   public toJSON() {
     return {
       count: this.count,
--- a/src/plugins/data/common/index_patterns/types.ts
+++ b/src/plugins/data/common/index_patterns/types.ts
@@ -185,8 +185,15 @@ export interface FieldSpec {
   readFromDocValues?: boolean;
   subType?: IFieldSubType;
   indexed?: boolean;
-  autoCompleteValues?: string[];
-  topQueryValues?: string[];
+
+  /**
+   * @experimental These fields are experimental and are subject to change
+   * TODO: Refactor and move them into DataViewField
+   */
+  suggestions?: {
+    autoCompleteValues?: string[];
+    topAggValues?: string[];
+  };
 }
 
 export type IndexPatternFieldMap = Record<string, FieldSpec>;
--- a/src/plugins/data/common/index_patterns/types.ts
+++ b/src/plugins/data/common/index_patterns/types.ts
@@ -185,6 +185,7 @@ export interface FieldSpec {
   readFromDocValues?: boolean;
   subType?: IFieldSubType;
   indexed?: boolean;
+  autoCompleteValues?: string[];
 }
 
 export type IndexPatternFieldMap = Record<string, FieldSpec>;
--- a/src/plugins/data/common/index_patterns/types.ts
+++ b/src/plugins/data/common/index_patterns/types.ts
@@ -186,6 +186,7 @@ export interface FieldSpec {
   subType?: IFieldSubType;
   indexed?: boolean;
   autoCompleteValues?: string[];
+  topQueryValues?: string[];
 }
 
 export type IndexPatternFieldMap = Record<string, FieldSpec>;
--- a/src/plugins/data/common/index_patterns/types.ts
+++ b/src/plugins/data/common/index_patterns/types.ts
@@ -191,8 +191,8 @@ export interface FieldSpec {
    * TODO: Refactor and move them into DataViewField
    */
   suggestions?: {
-    autoCompleteValues?: string[];
-    topAggValues?: string[];
+    values?: string[]; // fetched from the dataset
+    topValues?: string[]; // computed from the query result
   };
 }
 
--- a/src/plugins/data/common/search/search_source/types.ts
+++ b/src/plugins/data/common/search/search_source/types.ts
@@ -104,6 +104,7 @@ export interface SearchSourceFields {
   timeout?: string;
   terminate_after?: number;
   df?: IDataFrame;
+  skipTimeFilter?: boolean;
 }
 
 export interface SearchSourceOptions {
--- a/src/plugins/data/public/antlr/shared/utils.test.ts
+++ b/src/plugins/data/public/antlr/shared/utils.test.ts
@@ -20,6 +20,13 @@ import { HttpSetup } from 'opensearch-dashboards/public';
 import { IUiSettingsClient } from 'opensearch-dashboards/public';
 import { UI_SETTINGS } from '../../../common';
 
+// Mock the getDataViews service
+jest.mock('../../services', () => ({
+  getDataViews: jest.fn(() => ({
+    saveToCache: jest.fn(),
+  })),
+}));
+
 describe('fetchData', () => {
   it('should fetch data using the dataSourceRequestHandler', async () => {
     const mockTables = ['table1', 'table2'];
--- a/src/plugins/data/public/antlr/shared/utils.test.ts
+++ b/src/plugins/data/public/antlr/shared/utils.test.ts
@@ -24,6 +24,16 @@ import { UI_SETTINGS } from '../../../common';
 jest.mock('../../services', () => ({
   getDataViews: jest.fn(() => ({
     saveToCache: jest.fn(),
+    get: jest.fn().mockResolvedValue({
+      id: 'test-index-pattern',
+      title: 'test-index',
+      timeFieldName: 'timestamp',
+    }),
+    convertToDataset: jest.fn().mockResolvedValue({
+      id: 'test-dataset',
+      title: 'test-index',
+      type: 'INDEX_PATTERN',
+    }),
   })),
 }));
 
@@ -358,17 +368,29 @@ describe('fetchColumnValues', () => {
   let mockServices: IDataPluginServices;
   let mockHttp: HttpSetup;
   let mockUiSettings: IUiSettingsClient;
+  let mockSearchSource: any;
 
-  const createMockIndexPattern = (fieldType: string) =>
-    (({
+  const createMockIndexPattern = (fieldType: string, hasField: boolean = true) => {
+    const mockField = hasField
+      ? (({
+          type: fieldType,
+          isSuggestionAvailable: jest.fn().mockReturnValue(fieldType === 'number' ? false : true),
+          spec: {
+            suggestions: {
+              autoCompleteValues: undefined,
+              topAggValues: undefined,
+            },
+          },
+        } as unknown) as IndexPatternField)
+      : undefined;
+
+    return ({
       id: 'test-index',
       fields: {
-        find: jest.fn().mockReturnValue({
-          type: fieldType,
-          spec: { autoCompleteValues: undefined, topQueryValues: undefined },
-        } as IndexPatternField),
+        getByName: jest.fn().mockReturnValue(mockField),
       },
-    } as unknown) as IndexPattern);
+    } as unknown) as IndexPattern;
+  };
 
   beforeEach(() => {
     mockHttp = ({
@@ -386,6 +408,20 @@ describe('fetchColumnValues', () => {
       }),
     } as unknown) as IUiSettingsClient;
 
+    // Mock SearchSource
+    mockSearchSource = {
+      setFields: jest.fn().mockReturnThis(),
+      fetch: jest.fn().mockResolvedValue({
+        hits: {
+          hits: [
+            { _source: { 'test-column': 'value1' } },
+            { _source: { 'test-column': 'value2' } },
+            { _source: { 'test-column': 'value3' } },
+          ],
+        },
+      }),
+    };
+
     mockServices = ({
       http: mockHttp,
       uiSettings: mockUiSettings,
@@ -395,6 +431,11 @@ describe('fetchColumnValues', () => {
             getQuery: jest.fn().mockReturnValue({ dataset: { dataSource: { id: 'test-id' } } }),
           } as unknown) as QueryStringContract,
         } as unknown) as IQueryStart,
+        search: {
+          searchSource: {
+            create: jest.fn().mockResolvedValue(mockSearchSource),
+          },
+        },
         indexPatterns: {
           saveToCache: jest.fn(),
         } as unknown,
@@ -453,20 +494,59 @@ describe('fetchColumnValues', () => {
     expect(result).toEqual([]);
   });
 
-  it('should fetch and return column values for string fields', async () => {
-    const mockResponse = {
-      body: {
-        fields: [
-          {
-            values: ['value1', 'value2', 'value3'],
-          },
-        ],
-      },
-    };
-
-    (mockHttp.fetch as jest.Mock).mockResolvedValue(mockResponse);
-
+  it('should fetch and return column values using PPL and SearchSource', async () => {
     const mockIndexPattern = createMockIndexPattern('string');
+
+    const result = await fetchColumnValues(
+      'test-table',
+      'test-column',
+      mockServices,
+      mockIndexPattern,
+      'INDEX_PATTERN'
+    );
+
+    // Verify SearchSource was created and configured correctly
+    expect(mockServices.data.search.searchSource.create).toHaveBeenCalled();
+    expect(mockSearchSource.setFields).toHaveBeenCalledWith({
+      index: expect.objectContaining({
+        id: 'test-index-pattern',
+        title: 'test-index',
+        timeFieldName: 'timestamp',
+      }),
+      query: {
+        query: 'source = `test-table` | top 10 `test-column`',
+        language: 'PPL',
+        dataset: expect.objectContaining({
+          id: 'test-dataset',
+          title: 'test-index',
+          type: 'INDEX_PATTERN',
+        }),
+      },
+    });
+
+    // Verify fetch was called
+    expect(mockSearchSource.fetch).toHaveBeenCalled();
+
+    // Verify values were extracted from hits
+    expect(result).toEqual(['value1', 'value2', 'value3']);
+  });
+
+  it('should handle SearchSource fetch errors', async () => {
+    mockSearchSource.fetch.mockRejectedValue(new Error('Search API Error'));
+
+    const result = await fetchColumnValues(
+      'test-table',
+      'string-column',
+      mockServices,
+      createMockIndexPattern('string'),
+      'INDEX_PATTERN'
+    );
+
+    expect(result).toEqual([]);
+  });
+
+  it('should return empty array when field is undefined', async () => {
+    const mockIndexPattern = createMockIndexPattern('string', false);
 
     const result = await fetchColumnValues(
       'test-table',
@@ -476,42 +556,20 @@ describe('fetchColumnValues', () => {
       'INDEX_PATTERN'
     );
 
-    expect(mockHttp.fetch).toHaveBeenCalledWith({
-      method: 'POST',
-      path: '/api/enhancements/search/sql',
-      body: JSON.stringify({
-        query: {
-          query:
-            'SELECT `string-column` FROM `test-table` GROUP BY `string-column` ORDER BY COUNT(`string-column`) DESC LIMIT 10',
-          language: 'SQL',
-          format: 'jdbc',
-          dataset: { dataSource: { id: 'test-id' } },
-        },
-      }),
-    });
-
-    expect(result).toEqual(['value1', 'value2', 'value3']);
+    expect(result).toEqual([]);
   });
 
-  it('should handle API errors', async () => {
-    (mockHttp.fetch as jest.Mock).mockRejectedValue(new Error('SQL API Error'));
+  it('should return empty array when field is not suggestion available', async () => {
+    const mockField = ({
+      type: 'string',
+      isSuggestionAvailable: jest.fn().mockReturnValue(false),
+      spec: { suggestions: {} },
+    } as unknown) as IndexPatternField;
 
-    await expect(
-      fetchColumnValues(
-        'test-table',
-        'string-column',
-        mockServices,
-        createMockIndexPattern('string'),
-        'INDEX_PATTERN'
-      )
-    ).rejects.toThrow('SQL API Error');
-  });
-
-  it('should return empty array when field is undefined', async () => {
     const mockIndexPattern = ({
       id: 'test-index',
       fields: {
-        find: jest.fn().mockReturnValue(undefined),
+        getByName: jest.fn().mockReturnValue(mockField),
       },
     } as unknown) as IndexPattern;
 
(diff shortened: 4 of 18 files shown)


src/plugins/data/common/index_patterns/types.ts
```diff
@@ -185,6 +185,8 @@ export interface FieldSpec {
   readFromDocValues?: boolean;
   subType?: IFieldSubType;
   indexed?: boolean;
+  autoCompleteValues?: string[];
```
kavilla: Could we use the data views field spec instead ? That way we know we aren't messing with non experimental stuff

kavilla: I don't think "auto complete" would be the sole consumer of this. That is one use case but it's not the only use case.

Could there be a better name here that conveys that these are values that were present . Does "values" not work ? (Although I can understand the reason not to just name the values)

Also shouldnt it use a record? Like we wouldn't want to show duplicate values right?

Finally I think we should limit the size of the values. If there is too high of a cardinality then it might be too much values to suggest. Perhaps this could be name like "recentValues" or "latestEntries"  because if there is like a 100 different unique values we can probably assume we are getting the values it whatever the user preferred. 

I can see a good argument that the user would only would want to see or need to see up to 10 unique records values. For example if a user's service is experiencing outages.  And most of their logs are 400s 404s 500 etc then probably they won't care to see as much 200s in their auto complete. They likely just want to see the error fast

src/plugins/data/public/antlr/shared/utils.test.ts
```diff
@@ -20,6 +20,13 @@ import { HttpSetup } from 'opensearch-dashboards/public';
 import { IUiSettingsClient } from 'opensearch-dashboards/public';
 import { UI_SETTINGS } from '../../../common';
 
+// Mock the getDataViews service
+jest.mock('../../services', () => ({
+  getDataViews: jest.fn(() => ({
```
kavilla, line 25: Do we have a getdataviewsmockcontract if not I will make one

src/plugins/data/common/index_patterns/types.ts
```diff
@@ -185,6 +185,8 @@ export interface FieldSpec {
...
+  topQueryValues?: string[];
```
kavilla: What if instead of two new fields we make one new field that is a object that has these two values set on them. The reason why is because I feel like these values should be in sync with each other and updated together

kavilla: And are we sure it's of type string?   Like each osd field as an osd field type it maps too. Wouldn't it make sense that this maps to that?

Also could explain how will this be sorted?  Like should this be some sorted list or something or maybe more than just a string array?  Is the first index the most top value or is it the last index ? There wouldn't be anyway to know which field we should show at the very top or not

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +107,94 @@ export const fetchColumnValues = async (
...
+      !services.uiSettings.get(UI_SETTINGS.QUERY_ENHANCEMENTS_SUGGEST_VALUES) ||
+      !fieldInOsd ||
+      !allowedType.includes(fieldInOsd.type) ||
+      !indexPattern
+    ) {
+      return;
+    }
+
+    const limit = services.uiSettings.get(UI_SETTINGS.QUERY_ENHANCEMENTS_SUGGEST_VALUES_LIMIT);
+    const dataset = services.data.query.queryString.getQuery().dataset;
```
kavilla: I think you should just use the index pattern passed in . The best option tho is if you have passed in the query object. At which point u access query.dataset to get the type of dataset and the full index pattern object from the service. Like directly passed in not from accessing a Singleton. To protect ourselves from race conditions

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +107,94 @@ export const fetchColumnValues = async (
...
+
+    const values = (
+      await fetchFromAPI(
```
kavilla: Oh we don't just use the SQL search strategy via search.search ? We should prolly do that vs making raw http calls if we are

src/plugins/explore/public/components/fields_selector/lib/get_details.ts
```diff
@@ -68,5 +69,12 @@ export function getDetails(
       bucket.display = indexPattern.getFormatterForField(field).convert(bucket.value);
     }
   }
+
+  // Save updated index pattern with topQueryValues to cache
+  if (indexPattern && indexPattern.id) {
```
kavilla: If you could just update to dataView. Or dataset since that is where we want to be

src/plugins/explore/public/components/fields_selector/lib/get_details.ts
```diff
@@ -68,5 +69,12 @@ export function getDetails(
...
+    const services = getServices();
```
kavilla: Is this from the context or could. We just use the context ?

src/plugins/explore/public/components/fields_selector/lib/get_details.ts
```diff
@@ -33,6 +33,7 @@ import { i18n } from '@osd/i18n';
 import { getFieldValueCounts } from './field_calculator';
 import { IndexPattern, IndexPatternField } from '../../../../../data/public';
 import { OpenSearchSearchHit } from '../../../types/doc_views_types';
+import { getServices } from '../../../application/legacy/discover/opensearch_dashboards_services';
 
 export function getDetails(
   field: IndexPatternField,
```
kavilla: Update to use the right type for me? Then we can modify that file instead of touching. Index pattern field

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +107,94 @@ export const fetchColumnValues = async (
...
+  fieldInOsd: IndexPatternField | undefined
+): Promise<void> => {
+  try {
+    // Check if conditions allow API call
+    if (!datasetType || !Object.values(DEFAULT_DATA.SET_TYPES).includes(datasetType)) {
+      return;
+    }
+    if (
+      !services.uiSettings.get(UI_SETTINGS.QUERY_ENHANCEMENTS_SUGGEST_VALUES) ||
+      !fieldInOsd ||
```
kavilla, line 179: Why allow for undefined? If it just returns

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +107,94 @@ export const fetchColumnValues = async (
...
+      !allowedType.includes(fieldInOsd.type) ||
+      !indexPattern
+    ) {
+      return;
+    }
+
+    const limit = services.uiSettings.get(UI_SETTINGS.QUERY_ENHANCEMENTS_SUGGEST_VALUES_LIMIT);
```
kavilla, line 185: Ah we should just use this limit to define the size of the array on the field

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +107,94 @@ export const fetchColumnValues = async (
...
+        })
+      )
+    ).body.fields[0].values;
+
+    if (values) {
+      // Update the field with fresh API values
+      fieldInOsd.spec.autoCompleteValues = values;
+
+      // Save the updated IndexPattern to cache
+      getDataViews().saveToCache(indexPattern.id!, indexPattern as any);
```
kavilla, line 214: I don't think u need the any

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +107,94 @@ export const fetchColumnValues = async (
...
+    }
+  } catch (error) {
+    // Propagate the error instead of silently failing
+    throw error;
```
kavilla: Does this just catch and throw it? Prolly we don't need it then

src/plugins/explore/public/components/fields_selector/lib/field_calculator.ts
```diff
@@ -100,6 +100,11 @@ const getFieldValueCounts = (params: FieldValueCountsParams): FieldValueCounts =
       };
     }
 
+    // Store top values in field's topQueryValues for caching
+    if (counts.length > 0) {
+      field.spec.topQueryValues = counts.map((bucket) => String(bucket.value));
```
kavilla: I think we can then pick a better name because are bucketed counts aggregated ?  For ppl we do a query and agg call. So  is this actually the topAggValues? I think query might be confusing


What is the expected values that will be stored here? Could this always be derived ?!

kavilla (commented):
Could we mark these fields with the experimental annotation ?  I think we should really consider adding an explicit Boolean or specifying in the limits UI setting . That setting to -1 will disable field value caching. Just in case we need to protect ourselves and give the users a backdoor If we found out we might have missed something or if we cause blot.


Also I think we should have more explicit names or types. Could u also include an example values (, could be online comments )

Thank you for this. Cool stuff

LDrago27:
> Could we mark these fields with the experimental annotation ? I think we should really consider adding an explicit Boolean or specifying in the limits UI setting . That setting to -1 will disable field value caching. Just in case we need to protect ourselves and give the users a backdoor If we found out we might have missed something or if we cause blot.
> 
> Also I think we should have more explicit names or types. Could u also include an example values (, could be online comments )
> 
> Thank you for this. Cool stuff

Hi @kavilla , Thanks for comments. Based on our offline sync. These are some of the things I have updated.
- Use Search Source for making the call 
- Updated the structure of newly added fields
- Updated the API to fetch column Values
- Moved out the allowed datatypes for suggestion to IndexPatternFields that now exposes a method isSuggestionAvailable to  provide this information

src/plugins/data/common/search/search_source/types.ts
```diff
@@ -104,6 +104,7 @@ export interface SearchSourceFields {
   timeout?: string;
   terminate_after?: number;
   df?: IDataFrame;
+  skipTimeFilter?: boolean;
```
kavilla, line 107: whats this for?

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -57,7 +58,7 @@ const fetchFromAPI = async (http: HttpSetup, body: string) => {
   try {
     return await http.fetch({
       method: 'POST',
-      path: '/api/enhancements/search/sql',
+      path: `/api/enhancements/search/_sql`,
```
kavilla, line 61: has tthis been broken? anyways are we not able to use the search.search?

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +105,102 @@ export const fetchColumnValues = async (
...
+    return fieldInOsd.spec.suggestions.autoCompleteValues;
+  }
+
+  // Return topQueryValues if available and fire async API call to update the cache for subsequent calls
+  if (
+    fieldInOsd?.spec.suggestions?.topAggValues &&
+    fieldInOsd?.spec.suggestions?.topAggValues.length > 0
+  ) {
+    // Fire async API call to update cache in non-blocking manner
+    updateFieldValuesAsync(table, column, services, indexPattern, datasetType, fieldInOsd);
```
kavilla: should we not await here?

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +105,102 @@ export const fetchColumnValues = async (
...
+  return fieldInOsd?.spec.suggestions?.autoCompleteValues ?? [];
+};
+
+// Non-blocking async function to update field values in background
+const updateFieldValuesAsync = async (
+  table: string,
+  column: string,
+  services: IDataPluginServices,
+  indexPattern: IndexPattern,
+  datasetType: string | undefined,
```
kavilla, line 168: we should log an improvement to just pass in query dataset and drop the type and index pattern

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +105,102 @@ export const fetchColumnValues = async (
...
+  await updateFieldValuesAsync(table, column, services, indexPattern, datasetType, fieldInOsd);
+
+  // Return the results of synchronous calls
+  return fieldInOsd?.spec.suggestions?.autoCompleteValues ?? [];
+};
+
+// Non-blocking async function to update field values in background
+const updateFieldValuesAsync = async (
+  table: string,
+  column: string,
```
kavilla, line 165: is this the same thing as the field title? which is field in osd

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +105,102 @@ export const fetchColumnValues = async (
...
+  // Fire a synchronous query to fetch values
+  await updateFieldValuesAsync(table, column, services, indexPattern, datasetType, fieldInOsd);
+
+  // Return the results of synchronous calls
+  return fieldInOsd?.spec.suggestions?.autoCompleteValues ?? [];
+};
+
+// Non-blocking async function to update field values in background
+const updateFieldValuesAsync = async (
+  table: string,
```
kavilla, line 164: nit: i think it might be cleaner to pass by object since this is getting long

src/plugins/data/public/antlr/shared/utils.ts
kavilla: i see this got resolved . could we add some context then in line if possible

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +105,102 @@ export const fetchColumnValues = async (
...
+    }
+
+    const limit = services.uiSettings.get(UI_SETTINGS.QUERY_ENHANCEMENTS_SUGGEST_VALUES_LIMIT);
+
+    const dataView = await getDataViews().get(indexPattern.id!);
+    const dataset = await getDataViews().convertToDataset(dataView);
+
+    const searchSource = await services.data.search.searchSource.create();
+    searchSource.setFields({
+      index: dataView,
```
kavilla, line 192: if i remember correctly is index a string type?

src/plugins/data/common/search/search_source/types.ts
LDrago27: @kavilla 
Currently the PPL Interceptor always appends the timeFilter in the query. 

SQL language doesn't have a timePicker, but uses this util to fetch column Values. Since the PPL interceptor always adds in a timefilter so there are times SQL timepicker is set to default of 15 mins which results in us not receiving the appropriate results. In a bid to maintain backwards compatibility with SQL language, which don't have a timePicker, I am adding this as a way to bypass adding the timeFilter.

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +105,102 @@ export const fetchColumnValues = async (
...
+    ) {
+      return;
+    }
+
+    const limit = services.uiSettings.get(UI_SETTINGS.QUERY_ENHANCEMENTS_SUGGEST_VALUES_LIMIT);
+
+    const dataView = await getDataViews().get(indexPattern.id!);
+    const dataset = await getDataViews().convertToDataset(dataView);
+
+    const searchSource = await services.data.search.searchSource.create();
```
kavilla, line 190: i am pro using the search source because you can actually get some pretty nifty handling of state

however, i think my original suggestion was to to do `services.data.search()` was that not working?

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +105,102 @@ export const fetchColumnValues = async (
...
+    searchSource.setFields({
+      index: dataView,
+      query: {
```
kavilla, line 193: this alone might be worth exposing higher up in this file or function like `createQuerySomethingSpecific`

that way we can see how we making the query

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +105,102 @@ export const fetchColumnValues = async (
...
+        query: `source = ${escapeIdentifier(table)} | top ${limit} ${escapeIdentifier(column)}`,
+        language: 'PPL',
+        dataset,
+      },
+      skipTimeFilter: true,
```
kavilla: i think we passed this as osoemthing we can set but tbh i think we do what the time filter no? it would give more correlated values

src/plugins/data/public/antlr/shared/utils.ts
kavilla: any more context on this one

src/plugins/data/public/antlr/shared/utils.ts
```diff
@@ -104,49 +105,102 @@ export const fetchColumnValues = async (
...
+    });
+
+    const response = await searchSource.fetch();
+
+    // Extract field values from OpenSearch hits response
```
kavilla: i would remove `OpenSearch` technically it's just made to look like a OpenSearch response but it is open to be other things as well

src/plugins/data/common/index_patterns/types.ts
```diff
@@ -185,6 +185,15 @@ export interface FieldSpec {
   readFromDocValues?: boolean;
   subType?: IFieldSubType;
   indexed?: boolean;
+
+  /**
+   * @experimental These fields are experimental and are subject to change
+   * TODO: Refactor and move them into DataViewField
```
kavilla, line 191: Thank you

src/plugins/data/common/index_patterns/types.ts
```diff
@@ -185,6 +185,15 @@ export interface FieldSpec {
...
+   */
+  suggestions?: {
+    autoCompleteValues?: string[];
```
kavilla: nit: one more pitch if this could be just `values`. autocomplete is one user but it doesn't have to be the only user of this new field

src/plugins/data/common/index_patterns/types.ts
```diff
@@ -185,6 +185,15 @@ export interface FieldSpec {
...
+    topAggValues?: string[];
```
kavilla: if we are skipping the timeFilter and if we are building for observability which usually involve time. should we just always be using the time filter then.

then we could just call it `topValues` which would give us a clean interface like

```
suggestions: {
   values: []
   topValues: []
}

which would make this reusuable

File: src/plugins/data/public/antlr/shared/utils.ts (hunk 16 above)
Comment (LDrago27) [reply src/plugins/data/public/antlr/shared/utils.ts]: Nope we are showing them results obtained from query and updating it in background. We don't want this to be a blocking operation.

File: src/plugins/data/public/antlr/shared/utils.ts (hunk 18 above)
Comment (LDrago27) [reply src/plugins/data/public/antlr/shared/utils.ts:165]: Yup.

File: src/plugins/data/public/antlr/shared/utils.ts (hunk 9 above)
Comment (LDrago27) [reply src/plugins/data/public/antlr/shared/utils.ts:179]: Actually the main reason to allow for undefined is for typescript.

File: src/plugins/data/public/antlr/shared/utils.ts (hunk 15 above)
Comment (LDrago27) [reply src/plugins/data/public/antlr/shared/utils.ts:61]: Oh we have updated this to use the search.search in the current version. I am no longer using this.

File: src/plugins/data/common/index_patterns/types.ts (hunk 27 above)
Comment (LDrago27) [reply src/plugins/data/common/index_patterns/types.ts]: Sounds good.

File: src/plugins/data/public/antlr/shared/utils.ts (hunk 24 above)
Comment (LDrago27) [reply src/plugins/data/public/antlr/shared/utils.ts]: Updated.

File: src/plugins/data/public/antlr/shared/utils.ts (hunk 12 above)
Comment (LDrago27) [reply src/plugins/data/public/antlr/shared/utils.ts]: Agreed. We werenot doing anything. Currently removed it to allow for it fail silently without blocking the cx.

File: src/plugins/data/public/antlr/shared/utils.ts (hunk 19 above)
Comment (LDrago27) [reply src/plugins/data/public/antlr/shared/utils.ts:164]: Sure. Will refactor it later.

**Comment by LDrago27** [comment]
@kavilla I liked your idea of having the suggestion being filtered by time. 
So instead of removing the skipTimeFilter param that I introduced I felt it is better to have controls. 

For all queries we will be setting the param as False except in SQL where we set is as True. Since we have the limitation of SQL not having timePicker which actually sets the time range in the search source.

File: src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts [hunk 28]
```diff
@@ -91,13 +96,69 @@ export const defaultResultsProcessor: DefaultDataProcessor = (
...
+        grouped: false,
+      });
+
+      // Extract top values from the result buckets
+      if (result.buckets && result.buckets.length > 0) {
+        const topValues = result.buckets.map((bucket) => String(bucket.value));
+        fieldUpdates.push({ field, topValues });
+      }
+    } catch (error) {
+      // Silently continue on field processing errors
```
sejli, line 137: Do we want to log this somewhere and continue?

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -91,13 +96,69 @@ export const defaultResultsProcessor: DefaultDataProcessor = (
...
+    }
+  });
+
+  // Batch update all fields in the IndexPattern
+  if (fieldUpdates.length > 0) {
+    fieldUpdates.forEach(({ field, topValues }) => {
+      // Update the IndexPattern field
+      const indexPatternField = dataset.fields.getByName(field.name);
+      if (indexPatternField) {
+        const indexPatternFieldWithSuggestions = indexPatternField as any;
```
sejli: There's a couple `any`'s in this function, feel like we could type them correctly.

LDrago27: @sejli Thanks. Removed the any.

src/plugins/data/common/index_patterns/fields/index_pattern_field.ts
```diff
@@ -148,6 +148,10 @@ export class IndexPatternField implements IFieldType {
     return this.aggregatable && !notVisualizableFieldTypes.includes(this.spec.type);
   }
 
+  public isSuggestionAvailable() {
```
sejli, line 151: Nit: May want to consider renaming this function. Initially, it looked like it was to check if there's a suggestion available, when its actual purpose is to check if it's possible to generate suggestions on this field. For example, like `isSuggestable` or something of the sort.

LDrago27: Thanks. Will clean up in a followup PR

abbyhu2000 (approved):

PR #10416, opened 2025-08-18
Feature/partial results

### LINKED ISSUE #10374: [BUG] When the max bucket size is hit in a query returning partial or no results, the response should indicate that clearly
**Describe the bug**

When the max bucket size is hit & when partial or no results are returned, the response needs to inform that the results are partial.
When the bucket size limit was being hit, observed inconsistencies.
On some customers a toast popup (on OpenSearch Dashboards) clearly showed that something had gone wrong, though it was still unclear what.
In other customers, the dashboard simply appeared fine, however the data had not been updated. This can lead customers to think that the data is complete & they may miss vitally important information for their business.
This is occurring even with a simple table visualisation.

When https://github.com/opensearch-project/OpenSearch/issues/18982 is fixed, use the details in the response to display a message clearly indicating that the results are not complete.

For improved usability, allow a mechanism to resend the request with an increased bucket size. This could be a follow-up ticket if not addressed in this ticket.

**To Reproduce**
Create a simple table viz that has many terms aggs in it. Set the bucket size to some very small number and then use a terms agg that is greater than this. 

**Expected behavior**
The feedback in the returned results should be more clearly escalated that there are partial or no results and the data that is shown should still be processed and updated but appropriate feedback needs to be clearly presented to the user.

**OpenSearch Version**
2.19.0

**Dashboards Version**
2.19.0

**Plugins**

NA

**Screenshots**

None

OS: Alma Linux
Version 9.4

**Additional context**

NA

ogprakash:
### Description

When the max bucket size is hit and in the aggregation response partial or no results are returned, the response should clearly inform users that the results are incomplete.

This PR adds a partial results warning in the table visualization when results are truncated due to the bucket limit.

OpenSearch already exposes sum_other_doc_count in the aggregation response, which allows us to detect partial results (> 0 means some buckets were dropped due to the bucket size limit).

### Issues Resolved
closes #10374 

## Screenshot

Screenshot 2025-08-18 at 7 34 37 PM

## Testing the changes

1. Create a table visualization.
2. Set bucket size very small (e.g. 1–2). A "Partial results" warning should appear.
3. Increase bucket size to a higher value. The warning should disappear and show complete results.

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/search/expressions/opensearchaggs.ts
+++ b/src/plugins/data/public/search/expressions/opensearchaggs.ts
@@ -201,6 +201,31 @@ const handleCourierRequest = async ({
 
   (searchSource as any).finalResponse = resp;
 
+  function extractSumOtherDocCount(aggregations) {
+    let results = 0;
+
+    function traverse(obj, path = '') {
+      if (typeof obj !== 'object' || obj === null) {
+        return;
+      }
+
+      // Check if current object has sum_other_doc_count
+      if (obj.hasOwnProperty('sum_other_doc_count')) {
+        results += obj.sum_other_doc_count;
+      }
+
+      // Recursively traverse all properties
+      Object.keys(obj).forEach((key) => {
+        const newPath = path ? `${path}.${key}` : key;
+        traverse(obj[key], newPath);
+      });
+    }
+
+    traverse(aggregations);
+    return results;
+  }
+  const sumOtherDocCounts = extractSumOtherDocCount(resp.aggregations);
+
   const parsedTimeRange = timeRange ? calculateBounds(timeRange) : null;
   const tabifyParams = {
     metricsAtAllLevels,
@@ -215,6 +240,9 @@ const handleCourierRequest = async ({
     (searchSource as any).finalResponse,
     tabifyParams
   );
+
+  const tr: any = (searchSource as any).tabifiedResponse;
+  tr.meta = { ...(tr.meta || {}), sumOtherDocCount: sumOtherDocCounts };
 
   inspectorAdapters.data.setTabularLoader(
     () =>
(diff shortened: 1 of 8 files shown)


src/plugins/data/public/search/expressions/opensearchaggs.ts
```diff
@@ -201,6 +201,31 @@ const handleCourierRequest = async ({
...
 
+  function extractSumOtherDocCount(aggregations) {
+    let results = 0;
+
+    function traverse(obj, path = '') {
+      if (typeof obj !== 'object' || obj === null) {
+        return;
+      }
+
+      // Check if current object has sum_other_doc_count
```
ruanyl: Could you have a detailed comment to describe what's `sum_other_doc_count` and why we need to count it?

src/plugins/data/public/search/expressions/opensearchaggs.ts
```diff
@@ -201,6 +201,31 @@ const handleCourierRequest = async ({
 
   (searchSource as any).finalResponse = resp;
 
+  function extractSumOtherDocCount(aggregations) {
+    let results = 0;
+
+    function traverse(obj, path = '') {
```
ruanyl: It seems `path` is passed but not actually been used?

src/plugins/data/public/search/expressions/opensearchaggs.ts
```diff
@@ -296,7 +324,7 @@ export const opensearchaggs = (): OpenSearchaggsExpressionFunctionDefinition =>
       abortSignal: (abortSignal as unknown) as AbortSignal,
     });
 
-    const table: OpenSearchDashboardsDatatable = {
+    const table: OpenSearchDashboardsDatatable & { meta?: any } = {
```
ruanyl: Perhaps we can update interface `OpenSearchDashboardsDatatable` to add `meta?` property?

src/plugins/vis_type_table/public/components/table_vis_app.tsx
```diff
@@ -49,6 +54,12 @@ export const TableVisApp = ({
             direction={direction === 'column' ? 'row' : 'column'}
             alignItems={direction === 'column' ? 'flexStart' : 'stretch'}
           >
+            {shouldShowWarning && (
+              <EuiCallOut title="Warning: Partial results" color="warning" iconType="alert">
+                Some data was omitted due to bucket size limits.
```
ruanyl: would be nice to have `i18n` for these texts

PR #10419, opened 2025-08-19
[Explore] Refactor visualization registry

ruanyl:
### Description
Refactor visualization registry to remove the usage of `matchIndex`

## Changelog

- refactor: update visualization registry to remove the usage of `matchIndex`

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/chart_type_selector.tsx
+++ b/src/plugins/explore/public/components/visualizations/chart_type_selector.tsx
@@ -50,7 +50,7 @@ export const ChartTypeSelector = <T extends ChartType>({
   }, [chartType]);
 
   // Map columns to add value and text fields for select component
-  const { numericalColumns = [], categoricalColumns = [], dateColumns = [] } = visualizationData;
+  const { numericalColumns, categoricalColumns, dateColumns } = visualizationData;
 
   // Get icon type based on chart type
   const getChartIconType = (type: string): string => {
--- a/src/plugins/explore/public/components/visualizations/chart_type_selector.tsx
+++ b/src/plugins/explore/public/components/visualizations/chart_type_selector.tsx
@@ -14,12 +14,12 @@ import {
   EuiPanel,
 } from '@elastic/eui';
 import React, { useEffect, useMemo, useRef, useState } from 'react';
-import { ALL_VISUALIZATION_RULES } from './rule_repository';
 import { VisualizationRule } from './types';
 import { ChartType } from './utils/use_visualization_types';
 import { CHART_METADATA } from './constants';
 import { isChartType } from './utils/is_chart_type';
 import { VisData } from './visualization_builder.types';
+import { visualizationRegistry } from './visualization_registry';
 
 interface ChartTypeSelectorProps<T extends ChartType> {
   visualizationData: VisData;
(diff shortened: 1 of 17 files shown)


src/plugins/explore/public/components/visualizations/chart_type_selector.tsx
```diff
@@ -120,7 +107,7 @@ export const ChartTypeSelector = <T extends ChartType>({
     );
 
     return processed;
-  }, [baseChartTypeMapping, numericalColumns, categoricalColumns.length, dateColumns.length]);
+  }, [baseChartTypeMapping, numericalColumns, categoricalColumns, dateColumns]);
```
wanglam, line 110: The columns will be set to a new array in every render if it's undefined. Will it become a performance issue?

src/plugins/explore/public/components/visualizations/rule_repository.ts
```diff
@@ -210,9 +228,15 @@ const oneMetricTwoCateOneDateRule: VisualizationRule = {
   id: 'one-metric-two-category-one-date',
   name: '1 Metric & 2 Category & 1 Date',
   description: 'Multiple time series visualizations',
-  matches: (numerical, categorical, date) =>
-    numerical.length === 1 && categorical.length === 2 && date.length === 1,
-  matchIndex: [1, 2, 1],
+  matches: (numerical, categorical, date) => {
```
wanglam: What about separate a function something like `generateMatches` and pass the old `matchIndex` as the argument to share these match logic?

wanglam (approved):

src/plugins/explore/public/components/visualizations/rule_repository.ts
ruanyl: Good point, I will update

src/plugins/explore/public/components/visualizations/chart_type_selector.tsx
ruanyl: I don't quite get it, can you clarify please?

wanglam: I'm referring [line 53](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10419/files/971067a751f4365a022bd6be64bb533f87b9c3f0#diff-f6ed4d38764f54613519ecdcf723401a889164d781c2d6938092fd00f8ac1f34R53) in the same file. Every render will generate a new empty array if these columns doesn't exists in the `visualizationData`. It will trigger a new function call to generate `chartTypeMappedOptions`. It's not a big issue.

ruanyl: I removed the default value, the columns should always have a default value which shouldn't be undefined.

wanglam (approved):

src/plugins/explore/public/components/visualizations/rule_repository.ts
```diff
@@ -273,14 +296,22 @@ const oneMetricTwoCateHighCardRule: VisualizationRule = {
...
+      categorical.length === 2 &&
+      (numerical[0].uniqueValuesCount >= 7 ||
+        categorical[0].uniqueValuesCount >= 7 ||
+        categorical[1].uniqueValuesCount >= 7)
+    ) {
+      return 'EXACT_MATCH';
+    }
+    return 'COMPATIBLE_MATCH';
+  },
   chartTypes: [
```
Hailong-am, line 315: could this replaced by compare too?

ruanyl: Good point, I think yes, I will update in next PR.

Issue #10398, opened 2025-08-12
[RFC] Keyboard Shortcuts System for OpenSearch Dashboards


natel926:
### Introduction
This RFC proposes introducing the keyboard shortcuts system provides a centralized, context-aware framework for managing keyboard shortcuts across OpenSearch Dashboards. It replaces fragmented, plugin-specific implementations with a unified service that handles conflict resolution, cross-platform compatibility, and user discoverability.

### Problem Statement

OpenSearch Dashboards has scattered keyboard shortcut implementations across plugins:

#### Fragmented Implementation

* Each plugin implements independent keyboard handling
* Shared components lack coordinated shortcut behavior
* No standardized registration API

Note: For now, text editor shortcuts (Monaco, etc.) are out of scope and remain editor-managed.

#### Poor User Experience

* No help system to discover available shortcuts
* Inconsistent behavior across plugins
* Hidden functionality unknown to users

#### Technical Debt

* Code duplication for common keyboard patterns
* No deterministic conflict resolution in federated plugin environment

### Architecture Overview

The keyboard shortcuts system is implemented as a core service in OpenSearch Dashboards, providing centralized management of keyboard shortcuts across all applications and plugins.

* Single Global Event Listener - One keydown listener for the entire application
* Context-Aware Help System - Shows relevant shortcuts based on current app context
* Multi-Key Sequence Support - Handles complex shortcuts like "g -> d" for navigation
* Plugin Integration - Easy registration API for plugins and components

### System Architecture Diagram

<img width="471" height="461" alt="Image" src="https://github.com/user-attachments/assets/1241dfc6-b8ba-4329-a1d5-b903a56c3d4f" />

### Event Flow Diagram

<img width="381" height="556" alt="Image" src="https://github.com/user-attachments/assets/a4301f60-ec17-48f7-81c6-4af0884cdae9" />

### Registration Flow Diagram

<img width="471" height="951" alt="Image" src="https://github.com/user-attachments/assets/4103ed69-2925-492b-8a46-0294020bafa9" />

### Solution: Console Warning for Conflict Detection

The solution adds conflict detection during shortcut registration without changing the core last-registered-wins behavior. When a duplicate key is detected, the system logs a detailed warning to help developers identify and resolve conflicts.

### Current Keyboard Shortcut Constraints

  Supported Keys (ALLOWED_KEYS)

  The keyboard shortcut service currently supports a limited set of keys for security and consistency:

[...]

PR #10424, opened 2025-08-20
[Keyboard-shortcuts] Update new Key Processing & Normalization

natel926:
### Description

This PR implements Phase 2 of the keyboard shortcuts system, providing the core key processing engine that transforms raw keyboard input into standardized, cross-platform key representations.

## Changelog
- feat: Add key Processing & normalization

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,555 @@
...
+/**
+ * Allowed prefix keys for two-key sequences.
+ *
+ * usage patterns:
+ * - g+d: Go to Discover
+ * - g+v: Go to Visualizations
+ * - g+b: Go to Dashboard
+ *
+ */
+export const SEQUENCE_PREFIX = new Set(['g']);
```
ananzh: So we only supporting g as a sequence prefix?

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,555 @@
...
+      throw new Error(`Key string cannot be whitespace-only: "${keyString}"`);
+    }
+
+    if (keyString.includes(' ') && keyString !== 'space') {
+      throw new Error(`Extra spaces and chord sequences are not allowed: "${keyString}"`);
+    }
+
+    const normalized = keyString.toLowerCase();
+
+    if (normalized.startsWith('+') || normalized.endsWith('+')) {
```
ananzh: how about ctrl++ (zoom in)?

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -48,48 +44,44 @@ export class KeyboardShortcutService {
...
     const existingShortcuts = this.shortcutsMapByKey.get(key) || [];
-    this.shortcutsMapByKey.set(key, [...existingShortcuts, shortcut]);
 
+    if (existingShortcuts.length > 0) {
+      const conflictingShortcuts = existingShortcuts
+        .map((s) => `${s.id} (${s.pluginId})`)
+        .join(', ');
+      // eslint-disable-next-line no-console
+      console.warn(
+        `keyboard shortcut conflict detected for key "${shortcut.keys}". ` +
```
ananzh, line 75: nice warning. I am curious if we consider a way to check for conflicts before registration or provide some option to prevent registration on conflict

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -143,7 +142,7 @@ export class KeyboardShortcutService {
       return;
     }
 
-    const eventKeyString = this.getEventKeyString(event);
+    const eventKeyString = this.keyParser.getEventKeyString(event);
     const shortcuts = this.shortcutsMapByKey.get(eventKeyString);
 
     if (shortcuts?.length) {
```
ananzh, line 151: I think we need some document in the service or here (const shortcut = shortcuts[shortcuts.length - 1];) that "Last registered shortcut wins"

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -129,9 +121,16 @@ export class KeyboardShortcutService {
   private shouldIgnoreKeyboardEventForTarget(target: EventTarget | null): boolean {
     if (!this.isHTMLElement(target)) return false;
 
-    const tagName = target.tagName;
+    const element = target;
+    const tagName = element.tagName;
+
+    const inputTags = ['INPUT', 'TEXTAREA', 'SELECT'];
```
ananzh, line 125: Do we also need to check for contenteditable elements? I think we don't have contenteditable components. Will we put some concerns for future expansion? or we are focusing on current osd?

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -48,48 +44,44 @@ export class KeyboardShortcutService {
...
+      );
     }
 
-    key += this.getNormalizedKey(event.key);
-
-    return key;
-  };
-
-  private register(shortcut: ShortcutDefinition): void {
     if (!this.config.enabled) {
```
ananzh, line 53: should we move this part before try catch? do we still need validation if there is no config enabled?

src/core/public/keyboard_shortcut/key_parser.ts
natel926: Yes, for now we only support g, but we may support more later.

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
angle943: I think its fine because they can just remove it.

I like the idea of preventing registration on conflict, but the main issue is that we cannot figure out what conflicts exists when there are external plugins involved. For example, externalA and externalB both registered same shortcut key without being aware. This will cause the app to crash when the two plugins come together, which is ultimattely a bad experience to the end users.

Thats why we agreed on "last registration win" approach, and always allow conflicts

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,555 @@
...
+  }
+
+  /**
+   * Validates the input key string for basic format and structural issues.
+   * Throws descriptive errors for invalid inputs.
+   */
+  private validateKeyString(keyString: string): void {
+    if (keyString === null || keyString === undefined) {
+      throw new Error(`Key string cannot be null or undefined`);
+    }
```
angle943: lets trust typescript and assume that this can't happen. i think we can remove this check. or better yet, see the comment below

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,555 @@
...
+
+    if (keyString === '') {
+      throw new Error(`Key string cannot be empty: "${keyString}"`);
+    }
```
angle943: 1. we can just throw "Key string cannot be empty" and not have the string interpolation. Its always going to be empty so its not helpful

2. Instead of having the above `keyString === null` || `keyString === undefined` above, why don't we just remove that condition and here we just do `if (!keyString)`? That will handle both `null`, `undefined`, and `empty string` scenario

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,555 @@
...
+ */
+export const SEQUENCE_PREFIX = new Set(['g']);
+
+/**
+ * Valid modifier combinations in their canonical order.
+ *
+ * Each combination is listed in canonical order to ensure consistent normalization
+ * regardless of the input order (e.g., "shift+cmd+a" becomes "cmd+shift+a").
+ */
+
```
angle943: i fear this file is getting too long, can we move: `ALLOWED_KEYS`, `SEQUENCE_PREFIX`, `VALID_MODIFIER_COMBINATIONS`, `VALID_MODIFIERS` to `./constants.ts`

and lets move `DisplayMappingKeys`, `PlatformDisplayMappings` to `./types.ts`

src/core/public/keyboard_shortcut/key_parser.ts
angle943: nitpick: usually `Map` and `Set` are good data structures for runtime creation of the data structures, but for a predefined one like this its usually better to use an array or object (because of typescript)

```
export const SEQUENCE_PREFIX = ['g'] as const
```

(the `as const` will make it readonly)

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,555 @@
...
+      tab: '⇥',
+      escape: 'Esc',
+      space: 'Space',
+      up: '↑',
+      down: '↓',
+      left: '←',
+      right: '→',
+      plus: '+',
+    },
+  } as const;
```
angle943: I think `CODE_TO_KEY_MAPPING` and `DISPLAY_MAPPINGS` can be move to `./constants.ts` to keep this file concise

angle943: btw, `constants.ts` doesn't need an associated unit test file

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,555 @@
...
+
+  private keyStringCache = new Map<string, string>();
```
angle943: (no action item) this is a good example of `Map` usage over an js object because of it's dynamic nature

src/core/public/keyboard_shortcut/key_parser.ts
angle943: that would be `ctrl+=` (there is no physical `+` key on the keyboard)

angle943: But how about `normalized.contains('++')`? Should we check for that?

Ex: `Ctrl++m`, `Ctrl+++z`

src/core/public/keyboard_shortcut/key_parser.ts
angle943: actually scratch what i said. `Set.has()` has avg time complexity of O(1). not that it matters much here, but still, good to know

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
natel926: Yes I added the check contenteditable elements before, but removed it after searching the code base since we don't currently have any contenteditable components. You're right, I should put it back for future expansion.

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
natel926: yes we should, since there's no need for validation if the config is not enabled.

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,555 @@
...
+    if (normalized.startsWith('+') || normalized.endsWith('+')) {
+      throw new Error(`Malformed key string: invalid '+' character placement: "${keyString}"`);
+    }
+
+    if (/\+{2,}/.test(normalized)) {
+      throw new Error(`Malformed key string: invalid consecutive '+' characters: "${keyString}"`);
+    }
+  }
+
+  private computeNormalizedKeyString(keyString: string): string {
```
angle943: What about this?

1. make `ALLOWED_KEYS` as a readonly array instead of Set.

2. Make VALID_MODIFIER_COMBINATIONS be:

```
export const VALID_MODIFIER_COMBINATIONS = [
'shift+', 'alt+', 'cmd+',, 'alt+shift+', 'cmd+shift+', 'cmd+alt+', 'cmd+alt+shift+'] as const
```
(i removed the empty string one)

3. make a `utils.ts` file in this folder, and add the following (this is in `utils` because it is not a `const`, and there are logical units of code in here, meaning it needs an associated unit test):

```
import {ALLOWED_KEYS, VALID_MODIFIER_COMZBINATIONS} from './constants';

// This can be removed once Regexp.escape() is available
export const escapeKey = (key: string) => {
  return ['.', '[', ']', '.', '/', '\\'].includes(key) ? `\\${key}` : key;
};

// This can be removed once Regexp.escape() is available
export const escapeModifier = (key: string) => {
  return key.replaceAll('+', '\\+');
};

export const VALID_KEY_STRING_REGEX = new RegExp(
  `^(?:${VALID_MODIFIER_COMBINATIONS.map(escapeModifier).join('|')})?(?:${ALLOWED_KEYS.map(
    escapeKey
  ).join('|')})$`
);

// and unit test the crap out of this. you can iterate over both VALID_MODIFIER_COMBINATIONS and ALLOWED_KEYS to do the test
```

Then you don't need `validateKeyString()` and in `computeNormalizedKeyString`, you can just do:

```
private computeNormalizedKeyString = (keyString: string) => {
const lowercasedKeyString = keyString.toLowerCase();
 if (!VALID_KEY_STRING_REGEX.test(lowercasedKeyString)) throw new Error(`invalid key: ${keyString}`);

return lowercasedKeyString;
}
```

The downside of this is that it doesn't provide much help to dev if they have an error, but if we have clear documentation then its up to devs to figure out what they did wrong. The upside is a lot less complexity (and can test the regex using loops to ensure it works)


ALSO, we can remove `computeNormarmalizedKeyString` and just add that logic to `normalizeKeyString`, something like this:

```
public normalizeKeyString(keyString: string): string {
    if (this.keyStringCache.has(keyString)) {
      return this.keyStringCache.get(keyString)!;
    }
   const lowercasedKeyString = keyString.toLowerCase();
    if (!VALID_KEY_STRING_REGEX.test(lowercasedKeyString)) {
      throw new Error(`Invalid keyString: ${keyString}`);
    }
    this.keyStringCache.set(keyString, lowercasedKeyString);

    return lowercasedKeyString;
  }

// double check that this is correct. i hurriedly copied & pasted
```

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,555 @@
...
+    this.keyStringCache.set(keyString, result);
+
+    return result;
+  }
+
+  /**
+   * Validates the input key string for basic format and structural issues.
+   * Throws descriptive errors for invalid inputs.
+   */
+  private validateKeyString(keyString: string): void {
```
angle943: don't think we need this if we use regex

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,555 @@
...
+   *   // User presses Ctrl+S on Windows → "cmd+s" (normalized to Mac-style)
+   *
+   *   // Compare with config shortcut (both are Mac-style)
+   *   if (keyString === parser.normalizeKeyString("Ctrl+S")) {
+   *     // Execute save action - both strings are "cmd+s"
+   *   }
+   * });
+   *
+   */
+  public getEventKeyString(event: KeyboardEvent): string {
```
angle943, line 73: this we need, but lets use a string instead of array, and just add to the string in correct order with the `+`

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,555 @@
...
+  private getKeyFromCode(code: string): string | null {
+    const mappedKey = KeyStringParser.CODE_TO_KEY_MAPPING[code];
+    return mappedKey && ALLOWED_KEYS.has(mappedKey) ? mappedKey : null;
+  }
+
+  /**
+   * Finds the canonical modifier combination that matches the given modifiers.
+   * Returns the modifiers in their predefined order.
+   */
+  private getCanonicalModifierOrder(modifiers: string[]): string[] {
```
angle943: i don't think we need this with regex approach

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,555 @@
...
+      }
+    }
+    throw new Error(`Invalid modifier combination: [${modifiers.join(', ')}]`);
+  }
+
+  /**
+   * Builds a normalized key string from modifiers and keys.
+   * Handles the 3 supported shortcut types with proper formatting.
+   */
+  private buildKeyString(modifiers: string[], keys: string | string[]): string {
```
angle943: i don't think we need this with regex approach

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
angle943: yeah good to protect future cases since we can't control that

src/core/public/keyboard_shortcut/key_parser.ts
angle943: actually maybe we don't even need the cache with this approach

src/core/public/keyboard_shortcut/key_parser.ts
```diff
@@ -0,0 +1,134 @@
...
+   * parser.normalizeKeyString("cmd++") // → throws Error
+   *
+   */
+  public normalizeKeyString(keyString: string): string {
+    const lowercasedKeyString = keyString.toLowerCase();
+    if (!VALID_KEY_STRING_REGEX.test(lowercasedKeyString)) {
+      throw new Error(
+        `Invalid key combination: "${keyString}". ` +
+          `Use supported keys (a-z, 0-9, punctuation, arrows, special keys) with valid modifiers (cmd, alt, shift). ` +
+          `Examples: "cmd+s", "alt+shift+a", "g+d"`
```
angle943, line 40: g+d should not be allowed. And what if they wrote `shift+alt+a` and is getting this error? I would just be vague:

`Invalid key combination: ${keyString}. Please refer to our documentation to see the allowed formats`

angle943: You could do this in a follow up PR

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -50,46 +46,40 @@ export class KeyboardShortcutService {
...
 
-  private register(shortcut: ShortcutDefinition): void {
-    if (!this.config.enabled) {
-      return;
+    if (existingShortcuts.length > 0) {
+      const conflictingShortcuts = existingShortcuts
+        .map((s) => `${s.id} (${s.pluginId})`)
+        .join(', ');
+      // eslint-disable-next-line no-console
+      console.warn(
```
angle943, line 74: you could do this in a followup PR, but lets only show this when we are in prod env

angle943 (approved):

ps48:

PR #10425, opened 2025-08-20
[Explore] Infer date format of axis

lunalzm:
### Description

This PR introduces dynamic inference of date formats for axis labels and tooltips in time-based bar charts, line charts and  heatmaps. The changes enhance the date format functions by integrating a new utility, buildTemporalAxisFormatting, which infers appropriate date formats based on the temporal range and intervals in the dataset. This ensures that axis labels and tooltips display dates in a format that matches the data's granularity (e.g., seconds, minutes, hours, days, months, or years), improving readability and user experience. 

Additionally, this PR also fixes a small bug. There is a small problem in the graph generated by vega-lite. If the first scale mark on the horizontal axis is very close to the intersection of the x-axis and y-axis, the scale value label will not be displayed directly below the scale mark, but will be displayed to the right of the bottom. This will cause the adjacent scale mark on the right to be squeezed out.

##### Test case 1:
`source = opensearch_dashboards_sample_data_ecommerce | stats count() , avg(products.base_price) by span(order_date , 26m)`

image

##### Test case 2:
`source = opensearch_dashboards_sample_data_ecommerce | stats count() by span(order_date , 18s)`

image

##### Test case 1:
`source = opensearch_dashboards_sample_data_ecommerce | stats count() , avg(products.base_price) by span(order_date , 26m)`

image

##### Test case 2:
`source = opensearch_dashboards_sample_data_ecommerce | stats count() by span(order_date , 18s)`

image

#### Before
image

#### After
image

## Changelog
- feat: Infer axis date format

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/types.ts
+++ b/src/plugins/explore/public/components/visualizations/types.ts
@@ -224,6 +224,26 @@ export enum AggregationType {
   NONE = 'none',
 }
 
+export interface BucketOptions {
+  aggregationType?: AggregationType;
+  // exclusive for time-series histogram
+  bucketTimeUnit?: TimeUnit;
+
+  // exclusive for numerical histogram
+  bucketSize?: number;
+  bucketCount?: number;
+}
+
+export enum TimeUnit {
+  AUTO = 'auto',
+  YEAR = 'year',
+  MONTH = 'yearmonth',
+  DATE = 'yearmonthdate',
+  HOUR = 'yearmonthdatehours',
+  MINUTE = 'yearmonthdatehoursminutes',
+  SECOND = 'yearmonthdatehoursminutesseconds',
+}
+
 export interface AxisConfig {
   grid: boolean;
   gridColor?: string;
--- a/src/plugins/explore/public/components/visualizations/utils/utils.ts
+++ b/src/plugins/explore/public/components/visualizations/utils/utils.ts
@@ -27,16 +27,9 @@ export const applyAxisStyling = (
     // Grid settings
     grid: gridEnabled,
     labelSeparation: 8,
-    orient: axisStyle?.position, // Apply position
-    title: axisStyle?.title.text || axis?.name, // Apply title settings
+    orient: axisStyle?.position,
+    title: axisStyle?.title.text || axis?.name,
   };
-
-  // Apply position
-
-  fullAxisConfig.orient = axisStyle?.position;
-
-  // Apply title settings
-  fullAxisConfig.title = axisStyle?.title.text;
 
   // Apply axis visibility
   if (!axisStyle?.show) {
@@ -48,14 +41,22 @@ export const applyAxisStyling = (
   }
 
   // Apply label settings
-  const showLabels = axisStyle?.labels?.show ?? true;
-  fullAxisConfig.labels = showLabels;
+  if (axisStyle?.labels) {
+    fullAxisConfig.labels = !!axisStyle.labels.show;
+    if (fullAxisConfig.labels) {
+      fullAxisConfig.labelAngle = 0;
+      fullAxisConfig.labelLimit = 100;
 
-  if (showLabels) {
-    fullAxisConfig.labelAngle = axisStyle?.labels?.rotate ?? fullAxisConfig.labelAngle;
-    fullAxisConfig.labelLimit = axisStyle?.labels?.truncate ?? fullAxisConfig.labelLimit;
-    fullAxisConfig.labelOverlap = 'greedy';
-    fullAxisConfig.labelFlush = false;
+      if (axisStyle.labels.rotate !== undefined) {
+        fullAxisConfig.labelAngle = axisStyle.labels.rotate;
+      }
+      if (axisStyle.labels.truncate !== undefined && axisStyle.labels.truncate > 0) {
+        fullAxisConfig.labelLimit = axisStyle.labels.truncate;
+      }
+
+      fullAxisConfig.labelOverlap = 'greedy';
+      fullAxisConfig.labelFlush = false;
+    }
   }
 
   if (axis?.schema === VisFieldType.Date) {
--- a/src/plugins/explore/public/components/visualizations/utils/utils.test.ts
+++ b/src/plugins/explore/public/components/visualizations/utils/utils.test.ts
@@ -93,7 +93,7 @@ describe('applyAxisStyling', () => {
     });
     expect(config.labels).toBe(true);
     expect(config.labelAngle).toBe(0);
-    expect(config.labelLimit).toBe(0);
+    expect(config.labelLimit).toBe(100);
   });
 
   it('preserves existing labelAngle and labelLimit when rotate and truncate are provided', () => {


src/plugins/explore/public/components/visualizations/utils/utils.ts
```diff
@@ -269,5 +269,186 @@ export const getTooltipFormat = (
...
+    .sort((a, b) => a - b);
+
+  if (ts.length < 2) return {};
+
+  const rangeMs = ts[ts.length - 1] - ts[0];
+  const out: { tickCount?: number; format?: string; labelExpr?: string } = {};
+
+  out.tickCount = pickTickCount(rangeMs);
+
+  if (data.length > 1000) {
```
ruanyl: What does `1000` mean here?

src/plugins/explore/public/components/visualizations/utils/utils.ts
```diff
@@ -256,6 +214,48 @@ export const inferTimeUnitFromTimestamps = (
...
+
+  const range = (timestamps[timestamps.length - 1] - timestamps[0]) / 1000;
+
+  let minInterval = Infinity;
+  for (let i = 1; i < timestamps.length; i++) {
+    const interval = (timestamps[i] - timestamps[i - 1]) / 1000;
+    if (interval > 0 && interval < minInterval) {
+      minInterval = interval;
+    }
+  }
```
ruanyl: `minInterval` not used?

src/plugins/explore/public/components/visualizations/utils/utils.ts
```diff
@@ -269,5 +269,186 @@ export const getTooltipFormat = (
...
+  minute: 1000 * 60,
+  hour: 1000 * 60 * 60,
+  day: 1000 * 60 * 60 * 24,
+  week: 1000 * 60 * 60 * 24 * 7,
+  month: 1000 * 60 * 60 * 24 * 30,
+  year: 1000 * 60 * 60 * 24 * 365,
+} as const;
+
+const TICK_INTERVALS: Array<{ unit: TimeUnitName; step: number; durMs: number }> = [
+  { unit: 'second', step: 1, durMs: 1 * DURATION.second },
```
ruanyl: How does `step` work? Didn't see how it's used

src/plugins/explore/public/components/visualizations/utils/utils.ts
```diff
@@ -269,5 +269,186 @@ export const getTooltipFormat = (
...
+
+  { unit: 'week', step: 1, durMs: 1 * DURATION.week },
+
+  { unit: 'month', step: 1, durMs: 1 * DURATION.month },
+  { unit: 'month', step: 3, durMs: 3 * DURATION.month },
+
+  { unit: 'year', step: 1, durMs: 1 * DURATION.year },
+];
+
+export function pickTickCount(rangeMs: number, targetMin = 10, targetMax = 20): number {
```
ruanyl: Can simply use tick count values like: "millisecond", "second", "minute", "hour", "day", "week", "month", and "year"?

src/plugins/explore/public/components/visualizations/utils/utils.ts
```diff
@@ -269,5 +269,186 @@ export const getTooltipFormat = (
...
+
+  return out;
+}
+
+export const applyAxisStyling = (
+  axis?: VisColumn,
+  axisStyle?: StandardAxes,
+  disableGrid?: boolean,
+  data?: Array<Record<string, any>>
+): any => {
```
yyfamazon: NIT: Maybe it is better to define a specific interface, such as AxisConfig, for the return type of the applyAxisStyling function, rather than using any, to ensure that consumers know exactly what data structure to expect

src/plugins/explore/public/components/visualizations/utils/utils.ts
```diff
@@ -269,5 +269,186 @@ export const getTooltipFormat = (
...
+  if (axisStyle?.labels) {
+    if (!axisStyle?.labels.show) {
+      fullAxisConfig.labels = false;
+    } else {
+      fullAxisConfig.labels = true;
+      // Apply label rotation/alignment
+      if (axisStyle?.labels.rotate !== undefined) {
+        fullAxisConfig.labelAngle = axisStyle?.labels.rotate;
+      }
+
```
yyfamazon: NIT: You can refactor the label settings logic by leveraging Optional Chaining (?.) and Nullish Coalescing (??) to avoid unnecessary nested if/else statements to make the code concise

src/plugins/explore/public/components/visualizations/utils/utils.ts
lunalzm: I agree that defining a specific interface like `AxisConfig` for the return type of `applyAxisStyling` would improve type safety and clarity for consumers of the function. Using `any` is the logic that wrote before and I can improve this part in this PR.

src/plugins/explore/public/components/visualizations/utils/utils.ts
lunalzm: You're right!

src/plugins/explore/public/components/visualizations/utils/utils.ts
lunalzm: You're correct. This appears to be an leftover from a previous iteration of the code.

src/plugins/explore/public/components/visualizations/utils/utils.ts
lunalzm: I wanted to write the calculation logic as `const count = Math.max(2, Math.floor(rangeMs / cand.durMs * cand.step) + 1)` but now `durMs` is already multiplied by step, so I can remove `step`.

src/plugins/explore/public/components/visualizations/utils/utils.ts
lunalzm: Do you mean to use the time unit string supported by Vega-Lite (such as "minute", "day", "month") as the value of tickCount without writing a lot of logic to calculate the specific number?

src/plugins/explore/public/components/visualizations/utils/utils.ts
lunalzm: Large data volumes (>1000): For performance reasons, use a simple format. 
Small data volumes (≤1000): For presentation purposes, use a complex labelExpr to dynamically adjust the format for a smarter approach.

ruanyl (approved):

yyfamazon:
looks great!

src/plugins/explore/public/components/visualizations/utils/utils.ts
```diff
@@ -46,24 +42,18 @@ export const applyAxisStyling = (
...
-      // Apply label truncation
-      if (axisStyle?.labels.truncate !== undefined && axisStyle?.labels.truncate > 0) {
-        fullAxisConfig.labelLimit = axisStyle?.labels.truncate;
-      }
+  const showLabels = axisStyle?.labels?.show ?? true;
+  fullAxisConfig.labels = showLabels;
+
+  if (showLabels) {
+    fullAxisConfig.labelAngle = axisStyle?.labels?.rotate ?? fullAxisConfig.labelAngle;
+    fullAxisConfig.labelLimit = axisStyle?.labels?.truncate ?? fullAxisConfig.labelLimit;
```
wanglam: It seems we don't have a default `fullAxisConfig.labelAngle` and `fullAxisConfig.labelLimit` here. Using these values as fallback will still result in `undefined`. Is this intentional?

src/plugins/explore/public/components/visualizations/utils/utils.ts
```diff
@@ -46,24 +42,18 @@ export const applyAxisStyling = (
...
-      // Apply label rotation/alignment
-      if (axisStyle?.labels.rotate !== undefined) {
-        fullAxisConfig.labelAngle = axisStyle?.labels.rotate;
-      }
-
-      // Apply label truncation
-      if (axisStyle?.labels.truncate !== undefined && axisStyle?.labels.truncate > 0) {
-        fullAxisConfig.labelLimit = axisStyle?.labels.truncate;
-      }
+  const showLabels = axisStyle?.labels?.show ?? true;
```
wanglam: Seems `showLabels` will fallback with `true` here. Seems it's a different behavior with previous implementation. Is this intentional?

src/plugins/explore/public/components/visualizations/utils/utils.ts
lunalzm: You're right! I will define default values for `labelAngle` and `labelLimit` in the initial `fullAxisConfig` object.

src/plugins/explore/public/components/visualizations/utils/utils.ts
lunalzm: Ok, I'll move the `showLabels` logic inside the `if (axisStyle?.labels)` check.

wanglam (approved):

src/plugins/explore/public/components/visualizations/types.ts
```diff
@@ -223,24 +223,24 @@ export enum AggregationType {
   NONE = 'none',
 }
 
-export interface BucketOptions {
```
ruanyl: Should not be removed?

lunalzm: Sorry about that, I’ll add it back.

wanglam (approved):

ruanyl (approved):

Issue #9688, opened 2025-04-16
`.vscode` configs should be renamed to `*.example.json`


huyaboo:
I have some personal `launch.json`, `settings.json`, and `tasks.json` configured for this project. It's kind of annoying to have to pull changes and stash my `launch.json` configs. 

The `launch.json` should be renamed to something like `launch.example.json` so users can copy them over to their own `launch.json` and can be ignored by git.

huyaboo:
Closing issue as multi root workspaces in VS Code is a workaround, albeit not an elegant one

PR #10426, opened 2025-08-20
Add common VSCode launch configurations to `launch.json`

huyaboo:
Signed-off-by: Huy Nguyen <73027756+huyaboo@users.noreply.github.com>

### Description

This change adds some common OSD run configurations for VSCode users. This allows for quick one-click launches with debugging capabilities:
1. Start the Dashboards server only (with default yml configurations)
2. Start the OpenSearch snapshot cluster only
3. Start both Dashboards and OpenSearch snapshot cluster together (for a one-click end-to-end debugging environment) 

### Issues Resolved

Related to #9688

## Testing the changes

Ran the configurations in my VSCode IDE instance

## Changelog
- chore: add common Dashboards run configurations

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,7 +5,7 @@
      * Attaches a debugger to an already running OSD instance
      */
     {
-      "name": "Attach Debugger to running Dashboards Server",
+      "name": "Attach debugger to running Dashboards",
       "type": "pwa-node",
       "request": "attach",
       "port": 9229
@@ -14,7 +14,7 @@
      * Starts the OSD instance only
      */
     {
-      "name": "Start Dashboards Server",
+      "name": "Start Dashboards",
       "type": "node",
       "request": "launch",
       "runtimeExecutable": "yarn",
@@ -29,10 +29,10 @@
     },
     /**
      * Starts the OpenSearch snapshot server
-     * - Everytime the server is stopped, data will NOT be persisted. Thus, if you need to persist data, you can add a configuration to spin up your OpenSearch cluster of choice
+     * - Everytime this launch configuration runs, it overwrites your previous data. Thus, if you need to persist data, you can add a configuration in your VSCode workspace to run '${workspaceFolder}/.opensearch/{version}/bin/opensearch'
      */
     {
-      "name": "Start OpenSearch Cluster Snapshot",
+      "name": "Start OpenSearch cluster snapshot",
       "type": "node",
       "request": "launch",
       "runtimeExecutable": "yarn",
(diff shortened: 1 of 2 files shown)


.vscode/launch.json
```diff
@@ -1,11 +1,62 @@
...
+      "name": "Attach Debugger to running Dashboards Server",
+      "type": "pwa-node",
+      "request": "attach",
+      "port": 9229
+    },
+    /**
+     * Starts the OSD instance only
+     */
+    {
+      "name": "Start Dashboards Server",
```
kavilla: this should just be like `Start Dashboards`

the config above will attach directly to the Node server (which includes the code in the server folders).

But this will start OSD completely so it conflicts with the first config title

.vscode/launch.json
```diff
@@ -1,11 +1,62 @@
...
+      "runtimeArgs": [],
+      "env": {
+        "NODE_ENV": "development"
+      },
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "openOnSessionStart"
+    },
+    /**
+     * Starts the OpenSearch snapshot server
+     * - Everytime the server is stopped, data will NOT be persisted. Thus, if you need to persist data, you can add a configuration to spin up your OpenSearch cluster of choice
```
kavilla: it's actually flipped.

Everytime the server is started it will overwrite your previous snapshot data.

the files get placed in `.opensearch/{version}`

so if you run `yarn opensearch snapshot`, then stop it and then the next day you can actually re-access that same instance with the data available with by doing `./.opensearch/{version}/bin/opensearch`

.vscode/launch.json
```diff
@@ -1,11 +1,62 @@
...
+  "compounds": [
+    /**
+     * This is the run configuration to startup OSD end-to-end
+     * 1. Starts up OpenSearch Snapshot server
+     * 2. Starts up Dashboards server
+     */
+    {
+      "name": "Start Dashboards with OpenSearch Cluster Snapshot",
+      "configurations": ["Start OpenSearch Cluster Snapshot", "Start Dashboards Server"],
+      // If either Dashboards or OpenSearch server is stopped, both servers will be stopped (set this to false to stop an individual server without stopping all servers)
```
joshuali925, line 58: comments are not allowed in json, should it be launch.jsonc? or does vscode handle json with comments and we are ok with them being red here?

.vscode/launch.json
huyaboo: Noted. I guess from my perspective, `Start Dashboards` sounds kinda confusing since I'd expect everything to just "work", which is why I added the "Server" keyword in the launch name

.vscode/launch.json
huyaboo: TIL. Updated description

.vscode/launch.json
huyaboo: VSCode provides syntax highlighting and Intellisense support for `launch.json`, `settings.json`, and `tasks.json`. This includes support for comments in JSON. Not sure how they fare in other IDEs

angle943: it shows errors on my IDE (webstorm) but i don't think it matters:

- No reason non-vscode users need to look at this file
- looking at CI i don't think it fails any type/lint checks

PR #10433, opened 2025-08-22
[Keyboard-shortcuts] Add two-key sequence support

natel926:
### Description

This PR introduce the implement support for two-key sequences (e.g., "g d") in the keyboard shortcut system with temporal state management and validation.

Manages temporal state for two-key sequences with configurable timeout (default 1000ms)

## Changelog

- feat: Add a two-key sequence support

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/core/public/keyboard_shortcut/sequence_handler.ts
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Normalizes and validates sequence key strings
+   *
+   * @param keyString - Sequence string to normalize and validate
+   * @returns Normalized sequence string
+   * @throws Error if sequence format is invalid
+   */
+  public normalizeKeyString(keyString: string): string {
+    const normalized = keyString.toLowerCase().trim();
+    const parts = normalized.split(/\s+/).filter((part) => part.length > 0);
```
angle943:
```suggestion
    const parts = normalized.split(' ').filter((part) => !!part.length);
```

Regex should only be used if standard string does not suffice, the problem with regex its hard to read

src/core/public/keyboard_shortcut/utils.ts
```diff
@@ -20,3 +20,13 @@ export const VALID_KEY_STRING_REGEX = new RegExp(
...
+
+export const SINGLE_LETTER_REGEX = /^[a-z]$/;
+
+/**
+ * Detects if a key string represents a sequence (two space-separated keys)
+ * Uses simple heuristic for routing to appropriate parser
+ */
+export const isSequenceKeys = (keys: string): boolean => {
+  return /^\s*\w+\s+\w+\s*$/.test(keys);
+};
```
angle943: can we have unit tests for this funciton?

src/core/public/keyboard_shortcut/utils.ts
```diff
@@ -20,3 +20,13 @@ export const VALID_KEY_STRING_REGEX = new RegExp(
...
 );
+
+export const SINGLE_LETTER_REGEX = /^[a-z]$/;
+
+/**
+ * Detects if a key string represents a sequence (two space-separated keys)
+ * Uses simple heuristic for routing to appropriate parser
+ */
+export const isSequenceKeys = (keys: string): boolean => {
+  return /^\s*\w+\s+\w+\s*$/.test(keys);
```
angle943: What do you think about just checking to see if it contains a `space`? See my comment in keyboard_shortcut_service.ts

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -54,7 +59,9 @@ export class KeyboardShortcutService {
       return;
     }
 
-    const key = this.keyParser.normalizeKeyString(shortcut.keys);
+    const key = isSequenceKeys(shortcut.keys)
```
angle943: what if they typed `g d a`. I think with this logic, it would go to `this.keyParser.normalizeKeyString()`. But i think multi sequence keys should always be handled by `sequence_handler`.

src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts
```diff
@@ -54,7 +59,9 @@ export class KeyboardShortcutService {
...
+      ? this.sequenceHandler.normalizeKeyString(shortcut.keys)
+      : this.keyParser.normalizeKeyString(shortcut.keys);
 
     const namespacedId = this.getNamespacedId(shortcut);
```
angle943, line 61: lets put 

```
const namespacedId = this.getNamespacedId(shortcut);
    if (this.namespacedIdToKeyLookup.has(namespacedId)) {
      throw new Error(
        `Shortcut "${shortcut.id}" from plugin "${shortcut.pluginId}" is already registered`
      );
    }
```
before we get the `key`. Logically speaking we want to ensure the id is unique before we do anything

src/core/public/keyboard_shortcut/utils.ts
angle943: and if we just do `space`, then we don't even need this funciton

src/core/public/keyboard_shortcut/sequence_handler.ts
```diff
@@ -0,0 +1,98 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { SEQUENCE_PREFIX } from './constants';
+import { SINGLE_LETTER_REGEX } from './utils';
+
+export class SequenceHandler {
+  private firstKey: string | null = null;
+  private sequenceTimer: number | null = null;
```
angle943: nitpick: can we call this `sequenceTimerId`? I had to look up MDN docs to see what this did, so others might be confused as well

src/core/public/keyboard_shortcut/sequence_handler.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  private readonly timeout: number;
+
+  constructor(timeout: number = 1000) {
```
angle943: letes extract this timeout to a constant at the top of this files (or you can put it in ./constants) called `EXPORT CONST SEQUENCE_TIMEOUT_MS = 1000`

src/core/public/keyboard_shortcut/sequence_handler.ts
```diff
@@ -0,0 +1,98 @@
...
+    this.clearSequenceTimer();
+
+    const sequenceKey = `${this.firstKey} ${key}`;
+    this.resetSequence();
+
+    // // Only start new sequence if key is a valid sequence prefix
+    if (SEQUENCE_PREFIX.has(key)) {
+      this.firstKey = key;
+      this.startSequenceTimer();
+    }
```
angle943, line 70: hmmm  if this `if` condition is true, say they typed `g g`, then we would return `g g` from this function.

This is fine in terms of how we handle it, but its not technically correct in how this class handles that? Because it counts the second `g` as the first key.

What if we changed this as 
```
if (SEQUENCE_PREFIX.has(key)) {
      this.firstKey = key;
      this.startSequenceTimer();
      return null;
    }
```

and when we call this method, if have a ```
const sequenceKey = this.sequenceHandler.processSecondKey(eventKeyString);
if (!!sequenceKey) {
  this.executeShortcutForKey(event, sequenceKey);
}
```

i think this will make it less confusing, as i was confused why we still return `${this.firstKey} ${key}`; even if its a sequence key

File: src/core/public/keyboard_shortcut/utils.ts (hunk 3 above)
Comment (natel926) [reply src/core/public/keyboard_shortcut/utils.ts]: what do you think about this then we do not need the isSequenceKeys? 

```
const key = shortcut.keys.includes(' ')
      ? this.sequenceMatcher.normalizeKeyString(shortcut.keys)
      : this.keyParser.normalizeKeyString(shortcut.keys);
```

Comment (angle943) [reply src/core/public/keyboard_shortcut/utils.ts]: that works. just make sure you handle trailing and leading whitespaces correctly.

you could even do `shortcut.keys.trim().includes(' ')` if you don't want to deal with that

I just want to avoid the scenario where they are trying to input single key, but get a sequential key error

Comment (natel926) [reply src/core/public/keyboard_shortcut/utils.ts]: Yes and inside the normalizeKeyString.  we have the trim already. 
```
public normalizeKeyString(keyString: string): string {
    const normalized = keyString.toLowerCase().trim();
```

File: src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts (hunk 4 above)
Comment (natel926) [reply src/core/public/keyboard_shortcut/keyboard_shortcut_service.ts]: `g d a `would go to this.sequenceHandler.normalizeKeyString(shortcut.keys)

File: src/core/public/keyboard_shortcut/sequence_handler.ts (hunk 8 above)
Comment (natel926) [reply src/core/public/keyboard_shortcut/sequence_handler.ts:70]: I'm thinking actually we should not start new sequence even if `SEQUENCE_PREFIX.has(key)`. for example, if we registered "g g" and "g d".  pressing g -> g will execute "g g" and need to press g again then -> d to execute the "g d" shortcut. I found Github keyboard shortcut also using this approach. 
```
 public processSecondKey(key: string): string {
    this.clearSequenceTimer();

    const sequenceKey = `${this.firstKey} ${key}`;
    this.resetSequence();

    return sequenceKey;
  }
```

natel926: synced with Justin offline. Keep it unchanged.

angle943 (approved):

ps48:

PR #10444, opened 2025-08-22
[Explore] disable charts in traces

Maosaic:
### Description

Temporarily aisabling charts on Traces page, because currently these charts are built base on query result which is restricted by the query response size(`ppl` currently has limit on response with first 10k result), we want to move aggregation query base implementation, which will imrove the accuracy against large size data but will add cost by making extra query.

## Changelog

- chore: Disable charts in traces

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/container/bottom_container/bottom_right_container/bottom_right_container.tsx
+++ b/src/plugins/explore/public/components/container/bottom_container/bottom_right_container/bottom_right_container.tsx
@@ -18,11 +18,14 @@ import { executeQueries } from '../../../../application/utils/state_management/a
 import { DiscoverChartContainer } from '../../../../components/chart/discover_chart_container';
 import { useDatasetContext } from '../../../../application/context';
 import { ResizableVisControlAndTabs } from './resizable_vis_control_and_tabs';
+import { useFlavorId } from '../../../../helpers/use_flavor_id';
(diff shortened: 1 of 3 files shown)


src/plugins/explore/public/components/container/bottom_container/bottom_right_container/bottom_right_container.tsx
```diff
@@ -78,7 +81,7 @@ export const BottomRightContainer = () => {
   if (status === QueryExecutionStatus.READY || status === QueryExecutionStatus.ERROR) {
     return (
       <>
-        <DiscoverChartContainer />
+        {flavorId !== ExploreFlavor.Traces && <DiscoverChartContainer />}
```
kavilla, line 84: framework wise. perhaps some re-factorization is in order. originally i suggested an embedded approach are we sure we don't want this

angle943 (approved):

src/plugins/explore/public/components/container/bottom_container/bottom_right_container/bottom_right_container.tsx
Maosaic: Updated in the PR Description

LDrago27 (approved):

PR #10446, opened 2025-08-22
[Query Enhancement] Fix Enter command behaviour in query editor

LDrago27:
### Description

Currently in Query Enhancement if you start with DQL/Lucene and switch into PPL/SQL. After  switching the editor pressing enter removes the query and reruns an empty query giving errors. Also the enter key can't be used to accept suggestions in the new editor. 

This PR adds an explicit action for Enter key. 
- Now we can accept suggestion using enter when the suggestion wizard is visible 
- Outside of suggestion window, pressing enter moves to next line. 
- Users can still make use of cmd+enter to run the query.

## Changelog

- [Query Enhancement] Fix Enter command behaviour in query editor
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/query_editor/editors/default_editor/index.tsx
+++ b/src/plugins/data/public/ui/query_editor/editors/default_editor/index.tsx
@@ -50,7 +50,21 @@ export const DefaultInput: React.FC<DefaultInputProps> = ({
       }, 100);
     });
 
-    // Return the original editor instance
+    // Overiding the Enter command
+    editor.addCommand(monaco.KeyCode.Enter, () => {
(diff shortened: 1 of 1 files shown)


kavilla:
is this the reason why on PPL we get a blank screen after hitting the keyboard shortcut? Should we backport this?

LDrago27:
> is this the reason why on PPL we get a blank screen after hitting the keyboard shortcut? Should we backport this?

Yup but only in Query Enhancements. Let me check with ashwin if we want to backport to older versions. Definitely an annoying bug.

PR #10448, opened 2025-08-25
[Cypress] Discover:Traces + Trace Details Sanity

TackAdam:
### Description
[Cypress] Discover:Traces + Trace Details Sanity testing.

Sanity test creates workspace and verifies Discover:Traces has correct display for no index patterns.
Adds the Otel sample data using UI.
Creates an index-pattern.
Verifies the Discover:Traces page renders corretly and has the spanId re-direction.
Clicks the spanId re-dirction and intercepts the call.
Manually goes to the Trace Details page and verifies it renders correctly.

testPassing

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/discover_traces.spec.js
```diff
@@ -0,0 +1,123 @@
...
+      });
+      cy.osd.waitForLoader(true);
+      cy.getElementByTestId('discoverNoIndexPatterns').should('be.visible');
+    });
+
+    it('should create trace index pattern and navigate to trace details', function () {
+      // Navigate to sample data page
+      cy.osd.navigateToWorkSpaceSpecificPage({
+        workspaceName: workspaceName,
+        page: 'import_sample_data',
```
kavilla, line 37: there's an api we can invoke to just install the sample data

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/discover_traces.spec.js
```diff
@@ -0,0 +1,123 @@
...
+            timeout: 30000,
+          }).should('be.visible');
+        }
+      });
+
+      // Create index pattern for traces
+      cy.log('Creating trace index pattern');
+      cy.createWorkspaceIndexPatterns({
+        workspaceName: workspaceName,
+        indexPattern: INDEX_PATTERN_NAME.replace('*', ''),
```
kavilla, line 58: i understand we do this in other locations but i actually have no clue why we do it other parts of the app.

do you know why we create an index pattern where we replace the `*` with a empty character?

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/discover_traces.spec.js
```diff
@@ -0,0 +1,123 @@
...
+        ' @ ' +
+        now.toLocaleTimeString('en-US', {
+          hour12: false,
+          hour: '2-digit',
+          minute: '2-digit',
+          second: '2-digit',
+        }) +
+        '.000';
+
+      cy.explore.setTopNavDate('Aug 1, 2025 @ 00:00:00.000', currentDate, true);
```
kavilla: couldn't we just use relative time "now"

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/discover_traces.spec.js
ananzh: +1

TackAdam: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/cypress/utils/apps/workspace/commands.js#L8

// Don't use * in the indexPattern it adds it by default at the end of name

Just following the convention in place, could replace the 
const INDEX_PATTERN_NAME = 'otel-v1-apm-span-*';
but think that might hurt read-ability unless we change every use.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/discover_traces.spec.js
TackAdam: // Click on dataset selector to close Syntax options if blocking time-picker
      cy.getElementByTestId('datasetSelectButton').should('be.visible').click();

      // Set time range to capture OTEL sample data - last 2 months
      cy.explore.setRelativeTopNavDate('2', 'Months ago');
      
      Switched to this to simplify and prevent flakiness.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/discover_traces.spec.js
TackAdam: I didn't see an example used in our existing test and am hesitant to introduce the command to be used in neo as this use of sample data is most likely temporary

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/discover_traces.spec.js
kavilla: i'm confused by what that comment is saying the source code. on the creation side the pattern doesn't get created automatically with a * it will just append the * to the pattern when the query is made. even with that`otel-v1-apm-span-*` is still the same as `otel-v1-apm-span-**`.

does the test fail if we simply just do 
```
cy.createWorkspaceIndexPatterns({
        workspaceName: workspaceName,
        indexPattern: INDEX_PATTERN_NAME,
 ...
 ```

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/discover_traces.spec.js
kavilla: sorry not too sure what you mean here. it would be calling essentially what this ui actions are interacting with. just skipping the added time and resources for this page to be loaded into the dom and just hitting the api directly

kavilla: we should have just ported over all the existing commands instead of creating new ones. 

you can copy these over https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/cypress/utils/commands.js#L614-L646

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/discover_traces.spec.js
TackAdam: No it still works, i can switch to that if you want and do a separate PR to fix all the other ones if we want to remove the confusion.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/discover_traces.spec.js
TackAdam: Tested and not working out of the box. Will look into this as follow-up if it sample data is maintained.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/discover_traces.spec.js
angle943: Background ex: `INDEX_PATTERN_NAME = 'data_logs_*'`

the problem is that when we want to create an index pattern of this type, that `createWorkspaceIndexPattern` needs `data_logs_` as the index pattern.

Sure we can use `data_logs_**`, but that would mean that whenever we need the index pattern in our tests, we would have to update it to not look for `INDEX_PATTERN_NAME`, but rather `INDEX_PARTTERN_NAME + '*'`

so the problem is whether we want to have the issue here or there. Prob the cleanest fix is to update `createWorkspaceIndexPatterns` to strip the `*` at the end

angle943 (approved):

Maosaic (approved):

PR #10449, opened 2025-08-26
[Explore] Metric viz with sparkline

ruanyl:
### Description
Update metric visualization to support sparkline

Screenshot 2025-08-26 at 11 28 04

## Changelog

- feat: update metric viz to support sparkline

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/rule_repository.ts
+++ b/src/plugins/explore/public/components/visualizations/rule_repository.ts
@@ -79,6 +79,7 @@ const oneMetricOneDateRule: VisualizationRule = {
     { ...CHART_METADATA.line, priority: 100 },
     { ...CHART_METADATA.area, priority: 80 },
     { ...CHART_METADATA.bar, priority: 60 },
+    { ...CHART_METADATA.metric, priority: 40 },
   ],
   toSpec: (
     transformedData,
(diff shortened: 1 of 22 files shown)


src/plugins/explore/public/components/visualizations/rule_repository.ts
```diff
@@ -533,14 +531,26 @@ const oneMetricRule: VisualizationRule = {
...
   ) => {
-    return createSingleMetric(
-      transformedData,
-      numericalColumns,
-      categoricalColumns,
-      dateColumns,
-      styleOptions,
-      axisColumnMappings
-    );
+    switch (chartType) {
```
wanglam, line 534: Nit: What is this switch for? Seems we execute the same `createSingleMetric` in both default and 'metric' case.

src/plugins/explore/public/components/visualizations/style_panel/value/value_calculation_selector.tsx
```diff
@@ -0,0 +1,150 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { EuiSuperSelect, EuiText } from '@elastic/eui';
+import React from 'react';
+
+const VALUE_CALCULATION_OPTIONS = [
+  {
+    value: 'last',
+    inputDisplay: 'Last',
```
wanglam, line 12: Nit: Do we need i18n for these options?

src/plugins/explore/public/components/visualizations/rule_repository.ts
ruanyl: You're right, it's not needed now, we will add gauge chart which will need this

src/plugins/explore/public/components/visualizations/style_panel/value/value_calculation_selector.tsx
ruanyl: Right, I will update in next PR

raintygao (approved):

PR #10452, opened 2025-08-26
[Explore] Traces - improve gantt chart

Maosaic:
### Description

1. Fix left padding, and give more space to display service name
2. Truncate service name text in the middle
3. Add Opacity and hightlight selected span
4. Make span in expanded row selectable
5. Move the error/fault indication from the right of the bar chart to the right of the service name  

## Screenshot

Screenshot 2025-08-25 at 9 31 46 PM

## Changelog

- feat: improve gantt chart

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/pages/traces/trace_details/public/gantt_chart_vega/gantt_chart_spec.ts
+++ b/src/plugins/explore/public/application/pages/traces/trace_details/public/gantt_chart_vega/gantt_chart_spec.ts
@@ -5,7 +5,7 @@
 
 import { Spec } from 'vega';
 import { i18n } from '@osd/i18n';
-import { GANTT_CHART_CONSTANTS, TOTAL_PADDING, calculateLeftPadding } from './gantt_constants';
+import { GANTT_CHART_CONSTANTS, calculateTextPadding } from './gantt_constants';
 
 export function createGanttSpec(
   height: number,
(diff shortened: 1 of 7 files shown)


src/plugins/explore/public/application/pages/traces/trace_details/public/gantt_chart_vega/gantt_chart_spec.ts
```diff
@@ -212,10 +207,19 @@ export function createGanttSpec(
             fontWeight: { value: 'bold' },
             baseline: { value: 'middle' },
             fill: { value: isDarkMode ? '#ffffff' : '#000000' },
-            text: { field: 'serviceName' },
+            text: {
+              signal:
```
LDrago27, line 211: Why not extract it into a function where we can pass the `datum.serviceName` or `datum.name` as params.

src/plugins/explore/public/application/pages/traces/trace_details/public/gantt_chart_vega/gantt_constants.ts
```diff
@@ -28,17 +26,23 @@ export const GANTT_CHART_CONSTANTS = {
...
   GANTT_CHART_CONSTANTS.BUFFER_PADDING;
 
 /**
  * Calculate left padding based on container width
  * Uses responsive sizing to ensure labels have adequate space
  */
-export const calculateLeftPadding = (containerWidth: number): number => {
-  return Math.max(60, Math.min(90, containerWidth * 0.08));
+export const calculateTextPadding = (containerWidth: number): number => {
+  return Math.max(120, Math.min(180, containerWidth * 0.12));
```
LDrago27, line 47: nit: We can maybe define the numbers 120,180 and 0.12 as constants which we can reference here. It is not clear right now, what these numbers mean.

src/plugins/explore/public/application/pages/traces/trace_details/public/gantt_chart_vega/gantt_chart_spec.ts
Maosaic: `datum.serviceName` and `datum.name` are not variables we can access here, they are part of the string get pass to vega

PR #10459, opened 2025-08-27
ai chat initial development

angle943:
- ai chat dev


src/plugins/chat/.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
angle943, line 1: We should change this plugin's name to something else. I started out thinking that I was just going to build the Chat UI components but now we agreed that it will contain everything

ruanyl:
@angle943 Curious what's this PR about? Are we adding a chat bot?

angle943:
@ruanyl yeah

PR #10465, opened 2025-08-29
[Explore] add histogram

Qxisylolo:
### Description
This pr adds histogram

1. numerical  histogram

https://github.com/user-attachments/assets/55a5583b-3297-4330-88f9-0506265c8175

2. time-series histogram

https://github.com/user-attachments/assets/a529526a-175f-447a-8602-3c2720116bb9

截屏2025-09-03 10 21 29

截屏2025-09-03 10 24 29

4. single value bar

https://github.com/user-attachments/assets/e9a31f86-698b-4726-9aa3-9e1d2a34b35c

5. categorical histogram

截屏2025-09-03 10 08 57

截屏2025-09-03 10 09 03

## Changelog
- feat: add histogram

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/bar/bar_vis_config.ts
+++ b/src/plugins/explore/public/components/visualizations/bar/bar_vis_config.ts
@@ -45,7 +45,7 @@ export interface BarChartStyleControls {
   titleOptions: TitleOptions;
 
   // histogram bucket config
-  bucket: BucketOptions;
+  bucket?: BucketOptions;
 }
 
 export const defaultBarChartStyles: BarChartStyleControls = {
@@ -123,7 +123,7 @@ export const defaultBarChartStyles: BarChartStyleControls = {
     titleName: '',
   },
   bucket: {
-    aggregationType: AggregationType.COUNT,
+    aggregationType: AggregationType.SUM,
     bucketTimeUnit: TimeUnit.AUTO,
   },
 };
--- a/src/plugins/explore/public/components/visualizations/bar/bar_vis_config.ts
+++ b/src/plugins/explore/public/components/visualizations/bar/bar_vis_config.ts
@@ -15,6 +15,9 @@ import {
   AxisRole,
   StandardAxes,
   TitleOptions,
+  AggregationType,
+  BucketOptions,
+  TimeUnit,
 } from '../types';
 import { BarVisStyleControls, BarVisStyleControlsProps } from './bar_vis_options';
 
@@ -40,6 +43,9 @@ export interface BarChartStyleControls {
   switchAxes: boolean;
 
   titleOptions: TitleOptions;
+
+  // histogram bucket config
+  bucket: BucketOptions;
 }
 
 export const defaultBarChartStyles: BarChartStyleControls = {
@@ -116,6 +122,10 @@ export const defaultBarChartStyles: BarChartStyleControls = {
     show: false,
     titleName: '',
   },
+  bucket: {
+    aggregationType: AggregationType.COUNT,
+    bucketTimeUnit: TimeUnit.AUTO,
+  },
 };
 
 export const createBarConfig = (): VisualizationType<'bar'> => ({
@@ -175,6 +185,13 @@ export const createBarConfig = (): VisualizationType<'bar'> => ({
         [AxisRole.X]: { type: VisFieldType.Numerical, index: 0 },
         [AxisRole.Y]: { type: VisFieldType.Categorical, index: 0 },
         [AxisRole.COLOR]: { type: VisFieldType.Categorical, index: 1 },
+      },
+      {
+        [AxisRole.X]: { type: VisFieldType.Numerical, index: 0 },
+        [AxisRole.Y]: { type: VisFieldType.Numerical, index: 1 },
+      },
+      {
+        [AxisRole.X]: { type: VisFieldType.Numerical, index: 0 },
       },
     ],
   },
--- a/src/plugins/explore/public/components/visualizations/bar/bucket_options.tsx
+++ b/src/plugins/explore/public/components/visualizations/bar/bucket_options.tsx
@@ -5,20 +5,20 @@
 
 import { i18n } from '@osd/i18n';
 import { EuiFormRow, EuiSelect, EuiFieldNumber } from '@elastic/eui';
-import React from 'react';
+import React, { useMemo } from 'react';
 import { BarChartStyleControls } from './bar_vis_config';
 import { AggregationType, TimeUnit, BucketOptions } from '../types';
-import { getAggregationnType, getTimeUnits } from '../utils/collections';
+import { getAggregationType, getTimeUnits } from '../utils/collections';
 import { useDebouncedValue } from '../utils/use_debounced_value';
 import { StyleAccordion } from '../style_panel/style_accordion';
 
-interface BucketVisOptionsProps {
+interface BucketOptionsPanelProps {
   styles: BarChartStyleControls['bucket'];
   bucketType: 'time' | 'num' | 'single';
   onChange: (styles: BarChartStyleControls['bucket']) => void;
 }
 
-export const BucketVisOptions = ({ styles, onChange, bucketType }: BucketVisOptionsProps) => {
+export const BucketOptionsPanel = ({ styles, onChange, bucketType }: BucketOptionsPanelProps) => {
   const updateBucketOption = <K extends keyof BucketOptions>(key: K, value: BucketOptions[K]) => {
     onChange({
       ...styles,
@@ -36,8 +36,8 @@ export const BucketVisOptions = ({ styles, onChange, bucketType }: BucketVisOpti
     (val) => updateBucketOption('bucketCount', val)
   );
 
-  const labelType = getAggregationnType();
-  const timeUnits = getTimeUnits();
+  const labelType = useMemo(() => getAggregationType(), []);
+  const timeUnits = useMemo(() => getTimeUnits(), []);
 
   return (
     <StyleAccordion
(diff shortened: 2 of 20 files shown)


src/plugins/explore/public/components/visualizations/bar/bucket_options.tsx
```diff
@@ -0,0 +1,118 @@
...
+            }
+            options={labelType}
+          />
+        </EuiFormRow>
+      )}
+
+      {bucketType === 'time' && (
+        <EuiFormRow
+          label={i18n.translate('explore.stylePanel.bar.bucket.timeUnit', {
+            defaultMessage: 'Time interval selection',
```
ruanyl: What about simply call it:
```suggestion
            defaultMessage: 'Interval',
```

src/plugins/explore/public/components/visualizations/bar/bucket_options.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { useDebouncedValue } from '../utils/use_debounced_value';
+import { StyleAccordion } from '../style_panel/style_accordion';
+
+interface BucketVisOptionsProps {
+  styles: BarChartStyleControls['bucket'];
+  bucketType: 'time' | 'num' | 'single';
+  onChange: (styles: BarChartStyleControls['bucket']) => void;
+}
+
+export const BucketVisOptions = ({ styles, onChange, bucketType }: BucketVisOptionsProps) => {
```
ruanyl:
```suggestion
export const BucketOptions = ({ styles, onChange, bucketType }: BucketVisOptionsProps) => {
```

src/plugins/explore/public/components/visualizations/bar/bucket_options.tsx
```diff
@@ -0,0 +1,118 @@
...
+    styles?.bucketSize,
+    (val) => updateBucketOption('bucketSize', val)
+  );
+
+  const [localBucketCount, setLocalBucketCount] = useDebouncedValue<number | undefined>(
+    styles?.bucketCount,
+    (val) => updateBucketOption('bucketCount', val)
+  );
+
+  const labelType = getAggregationnType();
```
ruanyl:
```suggestion
  const labelType = getAggregationType();
```

src/plugins/explore/public/components/visualizations/bar/bucket_options.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const timeUnits = getTimeUnits();
```
ruanyl: Shall we use `useMemo` for `getAggregationType` and `getTimeUnits`?

src/plugins/explore/public/components/visualizations/bar/bucket_options.tsx
Qxisylolo: I will use BucketOptionsPanel as BucketOptions is already used as interface. thanks~

src/plugins/explore/public/components/visualizations/bar/bucket_options.tsx
Qxisylolo: updated, thanks!

src/plugins/explore/public/components/visualizations/bar/bucket_options.tsx
Qxisylolo: yes, use useMemo will prevent unnecessary execution on every render. thanks for the advice~

src/plugins/explore/public/components/visualizations/types.ts
```diff
@@ -210,12 +210,32 @@ export interface RangeValue {
   max?: number;
 }
 
-export enum LabelAggregationType {
+export enum AggregationType {
   SUM = 'sum',
   MEAN = 'mean',
   MAX = 'max',
-  MIN = 'min',
+  COUNT = 'count',
```
Hailong-am, line 218: remove min by intention?

Hailong-am (commented):
quick question, does the aggregation done by vega or the ppl itself?

Qxisylolo:
> quick question, does the aggregation done by vega or the ppl itself?

yes, it is done by vega-lite

src/plugins/explore/public/components/visualizations/types.ts
Qxisylolo: Yes, I removed min because there’s a chance the data could be undefined, and since we don’t currently have preprocessing to handle invalid data, the result of min could be misleading.

ruanyl: Can vega be configured to ignore invalid values?

Qxisylolo: Vega will ignore null for example https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAdxoATemhBQ5mJDVlwATooxwaZLFp2y9B46bVI9aUMSSCGcCGgDaoAA80AEYlAE80ACYAXwVAqPC0AGZY+NRgyjFE1FkGQUFogF1YkGQjAGstJiQTJThZHTUDMlcQINRQADMaOEE1LSClTDCcOC0ARwYkWzpnGlIQEoiOkCQyMiM+TDH0BANTbt7+9AihkZ2QSenMWZuF6IegA. 

However, there may be cases where the field contains a string among the numbers, which will cause an error.

Qxisylolo: Removing min was just a instant idea. if null makes min misleading, the same could happen with other aggregations.  I will bring min back and Should we add a transform to filter out invalid fields

ruanyl: How could it work for `max`? I assume it will have the same problem?

Hailong-am: > Yes, I removed min because there’s a chance the data could be undefined, and since we don’t currently have preprocessing to handle invalid data, the result of min could be misleading.

I am a little confused here, does `min` behaves same as `max`?

Hailong-am:
> > quick question, does the aggregation done by vega or the ppl itself?
> 
> yes, it is done by vega-lite

 will that have performance issue for large dataset? I know that PPL could return at most 10k data by default. can we have a test for that?

src/plugins/explore/public/components/visualizations/types.ts
Qxisylolo: If min fails, max would fail as well. I just realized that an invalid field, which can cause errors. At the time, I removed min only because the large number of nulls could make its value misleading.

ruanyl: > Should we add a transform to filter out invalid fields

I think we should not filter out invalid values automatically for the user, because that may cause inaccurate result, for example the result of `count` maybe inaccurate if non-numerical values are filter out.

I suggest to keep the original data unchanged, if the invalid value causes problem when creating visualization, it should be better fixed by the query, or in the future, we can provide options to let user select specifically if invalid value should be filter out or not.

Qxisylolo: ok I will bring min back,thanks!

ruanyl (dismissed):

Hailong-am (approved):

src/plugins/explore/public/components/visualizations/bar/bar_vis_config.ts
```diff
@@ -40,6 +43,9 @@ export interface BarChartStyleControls {
   switchAxes: boolean;
 
   titleOptions: TitleOptions;
+
+  // histogram bucket config
+  bucket: BucketOptions;
```
ruanyl: Q: Did handle old saved object which doesn't have `bucket` in the config gracefully?

Qxisylolo: Thanks for pointing that out, we will initialize old save object with default bucket options

src/plugins/explore/public/components/visualizations/bar/bucket_options.tsx
```diff
@@ -0,0 +1,130 @@
...
+    >
+      {bucketType !== 'single' && (
+        <EuiFormRow
+          label={i18n.translate('explore.stylePanel.bar.bucket.type', {
+            defaultMessage: 'Type',
+          })}
+        >
+          <EuiSelect
+            compressed
+            value={styles?.aggregationType}
```
ruanyl: Instead of calling `onChange` in `useEffectOnce`, I think better to do this:

```suggestion
            value={styles?.aggregationType ?? defaultBarChartStyles.bucket.aggregationType}
```

src/plugins/explore/public/components/visualizations/bar/bucket_options.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+      {bucketType === 'time' && (
+        <EuiFormRow
+          label={i18n.translate('explore.stylePanel.bar.bucket.timeUnit', {
+            defaultMessage: 'Interval',
+          })}
+        >
+          <EuiSelect
+            compressed
+            value={styles?.bucketTimeUnit}
```
ruanyl: The same here, we can use a default value from `defaultBarChartStyles` if `bucketTimeUnit` not set

Qxisylolo: updated, thanks!

ruanyl (approved):

Hailong-am (approved):

PR #10470, opened 2025-09-01
[Explore] Table visualization enhancement

lunalzm:
### Description

This PR enhances the Table visualization by adding column alignment options, optional column filters, and table footers with statistical calculations.

- Column Alignment: Users can now configure cell content alignment (Auto, Left, Center, Right) for each column.
- Column Filters: Added optional Excel-like filters for each column, with a toggle to enable/disable them via settings.
- Table Footers: Introduced footers with support for statistical calculations (Last, Average, Min, Max, Total) on selected fields.
- Cell customization options:
   - Type: Colored text, colored background

### Column Filters

https://github.com/user-attachments/assets/7b0bc5b1-7ebf-43bb-b9be-16ab7bd5caf2

### Table Footers

https://github.com/user-attachments/assets/7117f31a-d936-40bd-8327-ded003da33a6

### Cell Customization

https://github.com/user-attachments/assets/66cf6558-6925-42f6-bd48-d5d28f1c4b87

## Changelog
- feat: Table visualization enhancement

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/table/table_vis.tsx
+++ b/src/plugins/explore/public/components/visualizations/table/table_vis.tsx
@@ -120,12 +120,20 @@ export const TableVis = React.memo(({ rows, columns, styleOptions }: TableVisPro
       const cellValue = Object.prototype.hasOwnProperty.call(filteredRows, rowIndex)
         ? filteredRows[rowIndex][columnId]
         : null;
-      const color =
-        columnCellType !== 'auto'
-          ? getThresholdByValue(cellValue, styleOptions?.thresholds, {
-              defaultColor: styleOptions?.baseColor,
-            })
-          : undefined;
+
+      let color: string | undefined;
+      if (
+        columnCellType !== 'auto' &&
+        styleOptions?.thresholds &&
+        styleOptions.thresholds.length > 0
+      ) {
+        const threshold = getThresholdByValue(cellValue, styleOptions.thresholds);
+        if (threshold) {
+          color = threshold.color;
+        } else {
+          color = styleOptions.baseColor;
+        }
+      }
 
       return (
         <CellValue
--- a/src/plugins/explore/public/components/visualizations/utils/utils.ts
+++ b/src/plugins/explore/public/components/visualizations/utils/utils.ts
@@ -12,6 +12,7 @@ import {
   VisColumn,
   AxisColumnMappings,
   AxisSupportedStyles,
+  Threshold,
 } from '../types';
 
 export const applyAxisStyling = (
@@ -274,25 +275,12 @@ export const getTooltipFormat = (
  * Determines the color for a value based on a set of thresholds.
  * @param value - The value to evaluate (e.g., a number, string, or any type that can be converted to a number).
  * @param thresholds - Array of threshold objects with `value` (number) and `color` (string) properties.
- * @param options - Optional configuration object.
- * @param options.isNumeric - Function to determine if the value is valid for threshold comparison (defaults to checking if the value is a number).
- * @param options.defaultColor - Fallback color if no threshold matches (defaults to '#000000').
- * @returns The color string for the matching threshold, the default color, or undefined if no color applies.
+ * @returns The matched threshold
  */
 export function getThresholdByValue<T>(
   value: any,
-  thresholds?: Array<{ value: number; color: string }>,
-  options: {
-    isNumeric?: (value: any) => boolean;
-    defaultColor?: string;
-  } = {}
-): string | undefined {
-  const { isNumeric = (v: any) => !isNaN(Number(v)), defaultColor = '#000000' } = options;
-
-  if (!thresholds || !isNumeric(value)) {
-    return undefined;
-  }
-
+  thresholds: Threshold[] = []
+): Threshold | undefined {
   const numValue = Number(value);
   if (isNaN(numValue)) {
     return undefined;
@@ -304,9 +292,9 @@ export function getThresholdByValue<T>(
   // Find the first threshold where the value is greater than or equal to the threshold value
   for (const threshold of sortedThresholds) {
     if (numValue >= threshold.value) {
-      return threshold.color;
+      return threshold;
     }
   }
 
-  return defaultColor;
+  return undefined;
 }


ruanyl:
@lunalzm can you please add description and maybe a screen record?

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -3,26 +3,68 @@
...
+    columns.forEach((col) => (uniques[col.column] = new Set()));
+    rows.forEach((row) => {
+      columns.forEach((col) => {
+        if (row.hasOwnProperty(col.column)) {
+          uniques[col.column].add(row[col.column]);
+        }
+      });
+    });
+    return Object.fromEntries(
+      Object.entries(uniques).map(([key, set]) => [key, Array.from(set).sort()])
```
wanglam: Nit: Do we need to remove empty columns?

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -3,26 +3,68 @@
...
+  calculation: CalcType;
+}
+
 export const TableVis = React.memo(({ rows, columns, styleOptions }: TableVisProps) => {
   const pageSize = styleOptions?.pageSize ? styleOptions.pageSize : 10;
   const [visibleColumns, setVisibleColumns] = useState(() => columns.map(({ column }) => column));
-
   const [pagination, setPagination] = useState({ pageIndex: 0, pageSize });
+  const [filters, setFilters] = useState<Record<string, FilterConfig>>({});
+  const [popoverOpen, setPopoverOpen] = useState<string | null>(null);
```
wanglam: Prefer renaming to something like `popoverOpenColumnId`.

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -38,21 +80,171 @@ export const TableVis = React.memo(({ rows, columns, styleOptions }: TableVisPro
...
+    const hasValues = Array.isArray(config.values) && config.values.length > 0;
+    const hasSearch = typeof config.search === 'string' && config.search.trim() !== '';
+    const strVal = (v: any) => (v == null ? '' : String(v));
+    const sVal = strVal(value);
+    const sSearch = (config.search || '').trim();
+    const toNum = (v: any) => (v == null || v === '' ? NaN : Number(v));
+
+    if (op === 'contains') {
+      if (hasSearch) return sVal.toLowerCase().includes(sSearch.toLowerCase());
+      if (hasValues) return config.values.includes(value);
```
wanglam: Shall we need to handle `hasSearch` and `hasValues`?

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -38,21 +80,171 @@ export const TableVis = React.memo(({ rows, columns, styleOptions }: TableVisPro
...
+      return true;
+    }
+
+    if (op === '=' || op === '!=') {
+      if (hasValues) {
+        const hit = config.values.includes(value);
+        return op === '=' ? hit : !hit;
+      }
+      if (!hasSearch) return false;
+      const eq = value === (isNaN(Number(sSearch)) ? sSearch : Number(sSearch));
```
wanglam: Shall we replace second `sSearch` with `NaN` ?

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -38,21 +80,171 @@ export const TableVis = React.memo(({ rows, columns, styleOptions }: TableVisPro
...
+  }, [rows, filters]);
+
+  const availableFieldSet = useMemo(() => new Set(columns.map((c) => c.column)), [columns]);
+  const normalizedFooterCalcs = useMemo<Calc[]>(() => {
+    const raw = styleOptions?.footerCalculations ?? [];
+
+    const processed: Calc[] = (raw as any[]).map(
+      (c): Calc =>
+        'fields' in c
+          ? { fields: (c.fields ?? []).map(String), calculation: c.calculation as CalcType }
```
wanglam: Shall we need to filter null or undefined fields?

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -38,21 +80,171 @@ export const TableVis = React.memo(({ rows, columns, styleOptions }: TableVisPro
...
+
+    normalizedFooterCalcs.forEach(({ fields, calculation }) => {
+      fields.forEach((field) => {
+        const values = getNumericSeries(field);
+        if (values.length === 0) {
+          footer[field] = '-';
+          return;
+        }
+
+        switch (calculation) {
```
wanglam: Can we reuse calculateValue in the `src/plugins/explore/public/components/visualizations/utils/utils.ts` file?

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -38,21 +80,171 @@ export const TableVis = React.memo(({ rows, columns, styleOptions }: TableVisPro
...
+      });
+    });
+    return footer;
+  }, [filteredRows, styleOptions?.showFooter, normalizedFooterCalcs]);
+
   const renderCellValue = useMemo(() => {
     return ({ rowIndex, columnId, setCellProps }: EuiDataGridCellValueElementProps) => {
-      return rows.hasOwnProperty(rowIndex) ? rows[rowIndex][columnId] : null;
+      const alignment = styleOptions?.globalAlignment || 'auto';
+      setCellProps?.({ style: { textAlign: alignment === 'auto' ? 'left' : alignment } });
```
wanglam: Why we need this auto option if always been set to left here?

src/plugins/explore/public/components/visualizations/table/table_vis_footer.tsx
```diff
@@ -0,0 +1,318 @@
...
+
+    return numericalColumns
+      .filter((col) => !selectedFields.includes(col.column))
+      .map((col) => ({
+        id: col.column,
+        label: col.name || col.column,
+      }));
+  };
+
+  const canAddCalculation = (() => {
```
wanglam: Prefer wrap with `useMemo`

src/plugins/explore/public/components/visualizations/table/table_vis_options.tsx
```diff
@@ -35,28 +44,96 @@ export const TableVisStyleControls: React.FC<TableVisStyleControlsProps> = ({
...
+    },
+    [updateStyleOption]
+  );
+
   const [localPageSize, handlePageSizeChange] = useDebouncedNumericValue(
     styleOptions.pageSize,
     onPageSizeChange
   );
 
+  const alignmentOptions = [
```
wanglam: Can we move constants outside of render function body?

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
lunalzm: Good point! Thanks for catching that. I’ll update the code to filter out empty columns so the return looks like:
```
return Object.fromEntries(
  Object.entries(uniques)
    .map(([key, set]) => [key, Array.from(set).sort()])
    .filter(([, arr]) => arr.length > 0)
);
```

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
lunalzm: OK!

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
lunalzm: I’ll update the logic so that both search and values are taken into account. The new implementation will be:
```
if (op === 'contains') {
  const matchSearch = !hasSearch || sVal.toLowerCase().includes(sSearch.toLowerCase());
  const matchValues = !hasValues || config.values.includes(value);
  return matchSearch && matchValues;
}
```

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
lunalzm: Thanks for the suggestion! In our case, the = and != operators only apply to numerical columns, so this branch is intended to handle numeric equality checks. Falling back to the raw string (sSearch) isn’t actually needed here, since we don’t support string equality for non-numerical fields. I’ll simplify this logic to only work with numbers to avoid confusion.
```
const eq = value === Number(sSearch);
```

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
lunalzm: Ok!

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
lunalzm: Sure!

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
lunalzm: I will change auto to "right-align the value column and left-align the other columns".

src/plugins/explore/public/components/visualizations/table/table_vis_footer.tsx
lunalzm: Good point 👍. I’ll wrap this calculation in a `useMemo` so it only recomputes when `numericalColumns` or `localCalculations` change. Thanks for catching this!

src/plugins/explore/public/components/visualizations/table/table_vis_options.tsx
lunalzm: Sure!

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
wanglam: Oh, that's my fault. I mean we can reuse the `calculateValue` function in [this file](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/explore/public/components/visualizations/utils/calculation.ts). It's an existing function, what do you think about it?

src/plugins/explore/public/components/visualizations/table/table_vis_options.tsx
```diff
@@ -5,13 +5,61 @@
...
+  { value: 'right', text: 'Right' },
+];
+
+const cellTypeOptions = [
+  { value: 'auto', text: 'Default (Black Text)' },
+  { value: 'colored_text', text: 'Colored Text' },
+  { value: 'colored_background', text: 'Colored Background' },
+];
+
+const CustomThresholdPanel: React.FC<{
```
ruanyl: It doesn't seem necessary to create a `CustomThresholdPanel` component

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -3,56 +3,325 @@
...
+    if (!styleOptions?.showColumnFilter) {
+      setFilters({});
+    }
+  }, [styleOptions?.showColumnFilter]);
+
+  const matchesFilter = (value: any, config: FilterConfig) => {
+    const op = config.operator || FilterOperator.Contains;
+    const hasValues = Array.isArray(config.values) && config.values.length > 0;
+    const hasSearch = typeof config.search === 'string' && config.search.trim() !== '';
+    const sVal = value == null ? '' : String(value);
```
ruanyl:
```suggestion
    const sVal = value === null ? '' : String(value);
```

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -3,56 +3,325 @@
...
+    const sSearch = (config.search || '').trim();
+    const toNum = (v: any) => (v == null || v === '' ? NaN : Number(v));
```
ruanyl: Is it intentional to use `==`?
```suggestion
    const toNum = (v: any) => (v === null || v === '' ? NaN : Number(v));
```

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -3,56 +3,325 @@
...
+    }));
+  }, []);
+
+  useEffect(() => {
+    if (!styleOptions?.showColumnFilter) {
+      setFilters({});
+    }
+  }, [styleOptions?.showColumnFilter]);
+
+  const matchesFilter = (value: any, config: FilterConfig) => {
```
ruanyl: Let's move this function to `table/utils.ts`

src/plugins/explore/public/components/visualizations/table/table_vis_filter.tsx
```diff
@@ -0,0 +1,341 @@
...
+}
+
+interface TableColumnHeaderProps {
+  col: VisColumn;
+  showColumnFilter?: boolean;
+  popoverOpen: boolean;
+  setPopoverOpen: (open: boolean) => void;
+  filters: Record<string, FilterConfig>;
+  setFilters: Dispatch<SetStateAction<Record<string, FilterConfig>>>;
+  uniques: any[];
```
wanglam: Nit: I'm wondering if we can use `unknown[]` instead of `any[]` for `uniques` and `values`. It's safer.

src/plugins/explore/public/components/visualizations/table/table_vis_config.ts
```diff
@@ -6,13 +6,32 @@
...
+  footerCalculations?: Array<{
+    fields: string[];
+    calculation: CalculationMethod;
+  }>;
+  cellType?: 'auto' | 'colored_text' | 'colored_background';
+  thresholds?: Threshold[];
+  baseColor?: string;
 }
 
 const defaultTableChartStyles: TableChartStyleControls = {
```
ruanyl:
```suggestion
const defaultTableChartStyles: Required<TableChartStyleControls> = {
```

src/plugins/explore/public/components/visualizations/table/table_vis_filter.tsx
```diff
@@ -0,0 +1,341 @@
...
+    return <span>{col.name}</span>;
+  }
+
+  const defaultOperator =
+    col.schema === 'numerical' ? FilterOperator.Equal : FilterOperator.Contains;
+  const currentFilter = filters[col.column] || { values: [], operator: defaultOperator };
+  const isFilterActive = currentFilter.values.length > 0 || !!currentFilter.search;
+
+  return (
+    <div className="table-column-header">
```
ruanyl: Can we rename these css classes with prefix like `visTableColumnHeader` so avoid conflict?

src/plugins/explore/public/components/visualizations/table/table_vis_footer.tsx
```diff
@@ -0,0 +1,326 @@
...
+};
+
+const CALCULATIONS: CalculationMethod[] = ['total', 'last', 'mean', 'min', 'max'];
+
+export const TableFooterStyleControls: React.FC<TableFooterStyleControlsProps> = ({
+  styleOptions,
+  onStyleChange,
+  numericalColumns = [],
+}) => {
+  const toNew = (c: any) =>
```
ruanyl: can we avoid `any` here?

ruanyl: Prefer more readable name than `toNew`

src/plugins/explore/public/components/visualizations/table/table_vis_footer.tsx
ruanyl, line 1: Prefer renaming `table_vis_footer` to `table_vis_footer_options` and `TableFooterStyleControls` to `TableFooterOptions`

src/plugins/explore/public/components/visualizations/table/table_vis_footer.tsx
```diff
@@ -0,0 +1,326 @@
...
+      ? {
+          fields: (c.fields || []).filter((f: any) => typeof f === 'string'),
+          calculation: c.calculation as CalculationMethod,
+        }
+      : {
+          fields: c.field ? [String(c.field)] : [],
+          calculation: c.calculation as CalculationMethod,
+        };
+
+  const [localCalculations, setLocalCalculations] = useState<
```
ruanyl: can we avoid this local state? 

I think you can do:
```ts
const localCalculations = useMemo(() => (styleOptions.footerCalculations || []).map(toNew), [styleOptions.footerCalculations])
``` 

Then you don't need `syncToParent`

src/plugins/explore/public/components/visualizations/table/table_vis_filter.tsx
lunalzm: OK.

src/plugins/explore/public/components/visualizations/table/table_vis_options.tsx
lunalzm: Yes, your suggestion is reasonable.

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
lunalzm: Sure!

src/plugins/explore/public/components/visualizations/table/table_vis_footer.tsx
```diff
@@ -0,0 +1,326 @@
...
+          >
+            <>
+              {localCalculations.map((calc, index) => (
+                <div key={index}>
+                  <EuiFormRow>
+                    <EuiFlexGroup gutterSize="s" alignItems="center">
+                      <EuiFlexItem>
+                        <EuiSelect
+                          compressed
+                          options={getCalculationOptionsForIndex(index)}
```
ruanyl: I think you can update `getCalculationOptionsForIndex` to accept `calc` directly for simplicity
```suggestion
                          options={getCalculationOptions(calc)}
```

src/plugins/explore/public/components/visualizations/table/table_vis_filter.tsx
lunalzm: OK!

src/plugins/explore/public/components/visualizations/table/table_vis_footer.tsx
lunalzm: Ok, I'll change it to `normalizeCalculation`.

src/plugins/explore/public/components/visualizations/table/table_vis_footer.tsx
lunalzm: Sure!

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -3,56 +3,264 @@
...
+
+  useEffect(() => {
+    setPagination({ pageIndex: 0, pageSize });
+  }, [filteredRows.length, pageSize]);
+
+  const availableFieldSet = useMemo(() => new Set(columns.map((c) => c.column)), [columns]);
+  const normalizedFooterCalcs = useMemo<Calc[]>(() => {
+    const raw = styleOptions?.footerCalculations ?? [];
+
+    const processed: Calc[] = (raw as any[]).map(
```
ruanyl: Q: Why `as any[]` here? 

`raw` type should be this right?
```ts
Array<{
    fields: string[];
    calculation: CalculationMethod;
  }>
```

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -3,56 +3,264 @@
...
+        if (result == null) {
+          footer[field] = '-';
+        } else {
+          const label = calculation.charAt(0).toUpperCase() + calculation.slice(1);
+          footer[field] = `${label}: ${result}`;
+        }
+      });
+    });
+    return footer;
+  }, [filteredRows, styleOptions?.showFooter, normalizedFooterCalcs]);
```
ruanyl: Suggest to move `footerValues` to `renderFooterCellValue` to avoid `useMemo`

src/plugins/explore/public/components/visualizations/utils/utils.ts
```diff
@@ -271,3 +273,84 @@ export const getTooltipFormat = (
   const timeUnit = inferTimeUnitFromTimestamps(data, field);
   return timeUnit ? timeUnitToFormat[timeUnit] ?? fallback : fallback;
 };
+
+// Function to calculate luminance of a hex color
+export const getLuminance = (hexColor: string): number => {
```
ruanyl: can you move util functions to `table/utils.ts` if they are table specific?

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -3,56 +3,264 @@
...
+      const thresholds = [...styleOptions.thresholds].sort((a, b) => b.value - a.value);
+      for (const threshold of thresholds) {
+        if (numValue >= threshold.value) {
+          return threshold.color;
+        }
+      }
+      return styleOptions.baseColor || '#000000';
+    },
+    [columnTypes, styleOptions?.baseColor, styleOptions?.thresholds]
+  );
```
ruanyl: I suggest to extract this to a util function in `utils/utils.ts`
```
function getThresholdByValue(value: any, thresholds, defaultColor?) {
// ...
}
```

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -3,56 +3,264 @@
...
+          footer[field] = `${label}: ${result}`;
+        }
+      });
+    });
+    return footer;
+  }, [filteredRows, styleOptions?.showFooter, normalizedFooterCalcs]);
+
+  const getCellColor = useCallback(
+    (value: any, columnId: string): string | undefined => {
+      if (columnTypes[columnId] !== 'numerical' || !styleOptions?.thresholds) return undefined;
```
ruanyl: I think we don't need to check `columnTypes[columnId] !== 'numerical'` here, because it does convert the value to number later `const numValue = Number(value);`. 

This will make the implementation more resilient, for example, for status code column, string value `"400"`, `"500"` can also be colored by thresholds.

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
lunalzm: Ok, I'll resolve this problem.

ruanyl: And why do we need map and filter here?

lunalzm: We use the `map` to clean each calculation by removing any fields that are no longer available, and then the `filter`  ensures we only keep calculations that still have at least one valid field.

src/plugins/explore/public/components/visualizations/utils/utils.ts
lunalzm: Sure!

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
lunalzm: Good idea 👍

src/plugins/explore/public/components/visualizations/table/table_cell_type_options.tsx
```diff
@@ -0,0 +1,149 @@
...
+export interface TableCellTypeOptionsProps {
+  styleOptions: TableChartStyleControls;
+  onStyleChange: (newStyle: Partial<TableChartStyleControls>) => void;
+  numericalColumns?: VisColumn[];
+  axisColumnMappings?: StyleControlsProps<TableChartStyleControls>['axisColumnMappings'];
+  updateVisualization?: StyleControlsProps<TableChartStyleControls>['updateVisualization'];
+}
+
+const cellTypeOptions = [
+  { value: 'auto', text: 'Default (Black Text)' },
```
ruanyl: Let's just call it `default`, because text color may different depends on the current theme.
```suggestion
  { value: 'auto', text: 'Default' },
```

src/plugins/explore/public/components/visualizations/table/table_vis_config.ts
```diff
@@ -6,13 +6,37 @@
 import React from 'react';
 import { VisualizationType } from '../utils/use_visualization_types';
 import { TableVisStyleControls } from './table_vis_options';
+import { Threshold } from '../types';
+import { CalculationMethod } from '../utils/calculation';
+
+export interface CellTypeConfig {
+  field: string;
+  type: 'auto' | 'colored_text' | 'colored_background';
```
ruanyl: Let's create a TS type, I see it's repeating
```
type ColorMode = 'auto' | 'colored_text' | 'colored_background'
```

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
lunalzm: OK

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
lunalzm: Got it!

src/plugins/explore/public/components/visualizations/table/table_vis_config.ts
lunalzm: Got it!

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -17,42 +24,205 @@ interface TableVisProps {
...
+      const columnCellType = cellTypes.find((ct) => ct.field === columnId)?.type || 'auto';
+      const alignment = styleOptions?.globalAlignment || 'auto';
+      const textAlign =
+        alignment === 'auto'
+          ? columnTypes[columnId] === 'numerical'
+            ? 'right'
+            : 'left'
+          : alignment;
+      const cellValue = Object.prototype.hasOwnProperty.call(filteredRows, rowIndex)
+        ? (filteredRows as any)[rowIndex][columnId]
```
ruanyl: `as any` seems unnecessary
```suggestion
        ? filteredRows[rowIndex][columnId]
```

src/plugins/explore/public/components/visualizations/table/table_vis_filter.tsx
```diff
@@ -0,0 +1,341 @@
...
+
+export const ColumnFilterContent: React.FC<ColumnFilterContentProps> = ({
+  col,
+  currentFilter,
+  onApply,
+  onClear,
+  onCancel,
+  uniques,
+}) => {
+  const numericOps = [
```
ruanyl: Move this constant out of the component

src/plugins/explore/public/components/visualizations/table/table_vis_footer_options.tsx
```diff
@@ -0,0 +1,332 @@
...
+
+export type TableFooterStyleControlsProps = StyleControlsProps<TableChartStyleControls> & {
+  numericalColumns?: VisColumn[];
+};
+
+const CALCULATIONS: CalculationMethod[] = ['total', 'last', 'mean', 'min', 'max'];
+
+interface FooterCalculationInput {
+  field?: unknown;
+  fields?: unknown[];
```
ruanyl: why it has `field` and `fields`?

src/plugins/explore/public/components/visualizations/table/table_vis_footer_options.tsx
```diff
@@ -0,0 +1,332 @@
...
+  calculation: CalculationMethod;
+}
+
+export const TableFooterOptions: React.FC<TableFooterStyleControlsProps> = ({
+  styleOptions,
+  onStyleChange,
+  numericalColumns = [],
+}) => {
+  const normalizeCalculation = (c: FooterCalculationInput) => {
```
ruanyl: Prefer move `normalizeCalculation` it out of the component or just move it into the `useMemo` of `localCalculations`

src/plugins/explore/public/components/visualizations/table/table_vis_footer_options.tsx
```diff
@@ -0,0 +1,332 @@
...
+      onStyleChange({ [key]: value });
+    },
+    [onStyleChange]
+  );
+
+  const onShowFooterChange = useCallback(
+    (checked: boolean) => {
+      updateStyleOption('showFooter', checked);
+      if (!checked) {
+        updateStyleOption('footerCalculations', []);
```
ruanyl: Do we need to reset `footerCalculations` to empty? I feel it's better experience to keep it so user can toggle between the two state while memorizing the footer configurations.

src/plugins/explore/public/components/visualizations/table/table_vis_footer_options.tsx
```diff
@@ -0,0 +1,332 @@
...
+      updateStyleOption('footerCalculations', next);
+    },
+    [updateStyleOption]
+  );
+
+  const onCalculationChange = useCallback(
+    (index: number, value: CalculationMethod) => {
+      const next = [...localCalculations];
+      next[index] = { ...next[index], calculation: value };
+      syncToParent(next);
```
ruanyl: Prefer to get rid of `syncToParent` function and call `updateStyleOption` directly

src/plugins/explore/public/components/visualizations/table/table_vis_footer_options.tsx
```diff
@@ -0,0 +1,332 @@
...
+    syncToParent(next);
+  }, [localCalculations, numericalColumns, syncToParent]);
+
+  const getCalculationOptions = useCallback(
+    (calc: { fields: string[]; calculation: CalculationMethod }) => {
+      const currentValue = calc.calculation;
+      const usedByOthers = new Set<CalculationMethod>(
+        localCalculations.filter((c) => c !== calc).map((c) => c.calculation)
+      );
+      const baseOptions: Array<{ value: CalculationMethod; text: string }> = [
```
ruanyl: Should the options be derived from `CALCULATIONS` variable?

src/plugins/explore/public/components/visualizations/table/table_vis_footer_options.tsx
```diff
@@ -0,0 +1,332 @@
...
+        { value: 'max', text: 'Max' },
+      ];
+      return baseOptions.filter(
+        (opt) => opt.value === currentValue || !usedByOthers.has(opt.value)
+      );
+    },
+    [localCalculations]
+  );
+
+  const getFieldOptionsForIndex = (index: number) => {
```
ruanyl: `index` never used?

src/plugins/explore/public/components/visualizations/table/table_vis_filter.tsx
lunalzm: Sure!

src/plugins/explore/public/components/visualizations/table/table_vis_footer_options.tsx
lunalzm: A calculation can have multiple fields, and each field can be added or deleted independently.

src/plugins/explore/public/components/visualizations/table/table_vis_footer_options.tsx
lunalzm: I will choose the former method to modify.

src/plugins/explore/public/components/visualizations/table/table_vis_footer_options.tsx
lunalzm: Sounds good!

src/plugins/explore/public/components/visualizations/table/table_vis_footer_options.tsx
lunalzm: Ok, I'll use `updateStyleOption('footerCalculations', next);` directly.

src/plugins/explore/public/components/visualizations/table/table_vis_footer_options.tsx
lunalzm: Then I change it to 
```
const baseOptions = CALCULATIONS.map((calc) => ({
    value: calc,
    text: calc.charAt(0).toUpperCase() + calc.slice(1),
})); 
```
Is it okay?

src/plugins/explore/public/components/visualizations/table/table_vis_footer_options.tsx
lunalzm: Yes, so I'll change it to `const getFieldOptions = () => {}`.

src/plugins/explore/public/components/visualizations/table/table_vis_footer_options.tsx
lunalzm: The `field` is redundant. I will change it like:
```
interface FooterCalculation {
    fields: string[];
    calculation: CalculationMethod;
}
```

src/plugins/explore/public/components/visualizations/table/table_vis_footer_options.tsx
```diff
@@ -0,0 +1,306 @@
...
+    const available = numericalColumns.map((c) => c.column).filter((c) => !selected.has(c));
+    if (available.length === 0) return;
+
+    const usedCalculations = new Set<CalculationMethod>(
+      localCalculations.map((c) => c.calculation)
+    );
+
+    const availableCalculation = CALCULATIONS.find((calc) => !usedCalculations.has(calc)) as
+      | CalculationMethod
+      | undefined;
```
ruanyl:
```suggestion
    const availableCalculation = CALCULATIONS.find((calc) => !usedCalculations.has(calc));
```

PR #10485, opened 2025-09-04
[Query enhancements] support custom component to create data structure

joshuali925:
### Description

In #8255, we added multi-select table for selecting multiple children as the next data structure. But it's not flexible enough, and there's no support for customizing that table component.

This PR allows each data structure to provide a custom UI component for how to select its `children` as the next data structure.

The interface is 
```tsx
export interface DataStructure {
  ...
  /** Whether data structure supports multi-selecting children. The selected
   * children will be combined using {@link DatasetTypeConfig['combineDataStructures']} */
  multiSelect?: boolean;
  /** Custom component to select children and create next data structure or
   * dataset. multiSelect will be ignored if using custom component */
  DataStructureCreator?: React.ComponentType<DataStructureCreatorProps>;
```
```tsx
export interface DataStructureCreatorProps<FetchOptions = unknown> {
  /** list of selected data structures */
  path: DataStructure[];
  /** update list of selected data structures */
  setPath: (newPath: DataStructure[]) => void;
  /** level of current data structure */
  index: number;
  /** select the next data structure or dataset */
  selectDataStructure: (item: DataStructure | undefined, newPath: DataStructure[]) => Promise<void>;
  /** fetch data structure using fetch in type config */
  fetchDataStructure: (
    path: DataStructure[],
    dataType: string,
    options?: FetchOptions
  ) => Promise<DataStructure>;
}
```

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/common/datasets/types.ts
+++ b/src/plugins/data/common/datasets/types.ts
@@ -32,6 +32,23 @@ export interface DataSourceMeta {
   supportsTimeFilter?: boolean;
 }
 
+export interface DataStructureCreatorProps<FetchOptions = unknown> {
+  /** list of selected data structures */
+  path: DataStructure[];
+  /** update list of selected data structures */
+  setPath: (newPath: DataStructure[]) => void;
+  /** level of current data structure */
+  index: number;
+  /** select the next data structure or dataset */
+  selectDataStructure: (item: DataStructure | undefined, newPath: DataStructure[]) => Promise<void>;
+  /** fetch data structure using fetch in type config */
+  fetchDataStructure: (
+    path: DataStructure[],
+    dataType: string,
+    options?: FetchOptions
+  ) => Promise<DataStructure>;
+}
+
 /**
  * Represents the hierarchical structure of data within a data source.
  *
@@ -125,7 +142,12 @@ export interface DataStructure {
   children?: DataStructure[];
   hasNext?: boolean;
   paginationToken?: string;
+  /** Whether data structure supports multi-selecting children. The selected
+   * children will be combined using {@link DatasetTypeConfig['combineDataStructures']} */
   multiSelect?: boolean;
+  /** Custom component to select children and create next data structure or
+   * dataset. multiSelect will be ignored if using custom component */
+  DataStructureCreator?: React.ComponentType<DataStructureCreatorProps>;
   columnHeader?: string;
   /** Optional metadata for the data structure */
   meta?: DataStructureMeta;
--- a/src/plugins/data/public/ui/dataset_selector/dataset_explorer.tsx
+++ b/src/plugins/data/public/ui/dataset_selector/dataset_explorer.tsx
@@ -203,15 +203,23 @@ export const DatasetExplorer = ({
                 >
                   <h3>{current.columnHeader}</h3>
                 </EuiTitle>
-                {current.multiSelect ? (
+                {current.DataStructureCreator ? (
+                  <current.DataStructureCreator
+                    path={path}
+                    setPath={setPath}
+                    index={index}
+                    selectDataStructure={selectDataStructure}
+                    // @ts-ignore custom component can have their own fetch options
+                    fetchDataStructure={fetchNextDataStructure}
+                  />
+                ) : current.multiSelect ? (
                   <DatasetTable
                     services={services}
                     path={path}
                     setPath={setPath}
                     index={index}
-                    explorerDataset={explorerDataset}
                     selectDataStructure={selectDataStructure}
-                    fetchNextDataStructure={fetchNextDataStructure}
+                    fetchDataStructure={fetchNextDataStructure}
                   />
                 ) : (
                   <EuiSelectable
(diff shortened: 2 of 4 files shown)


src/plugins/data/public/ui/dataset_selector/dataset_table.tsx
```diff
@@ -8,22 +8,12 @@ import { i18n } from '@osd/i18n';
 import { FormattedMessage } from '@osd/i18n/react';
 import React, { useRef, useState } from 'react';
 import { IDataPluginServices } from '../..';
-import { DataStructure } from '../../../common';
+import { DataStructureCreatorProps, DataStructure } from '../../../common';
 import { DatasetTypeConfig, DataStructureFetchOptions } from '../../query';
 import { getQueryService } from '../../services';
 
-interface DatasetTableProps {
+interface DatasetTableProps extends DataStructureCreatorProps<DataStructureFetchOptions> {
```
kavilla, line 15: nit:
since `DatasetTableProps` would it make more sense just make the base class like `DatasetCreationComponentProps` 

few reasons:
* we aren't really creating data structures we are trying to create a dataset
* `Component` because creator implies that the component itself will be the one responsible for creating.

src/plugins/data/common/datasets/types.ts
```diff
@@ -32,6 +32,23 @@ export interface DataSourceMeta {
...
+  /** list of selected data structures */
+  path: DataStructure[];
+  /** update list of selected data structures */
+  setPath: (newPath: DataStructure[]) => void;
+  /** level of current data structure */
+  index: number;
+  /** select the next data structure or dataset */
+  selectDataStructure: (item: DataStructure | undefined, newPath: DataStructure[]) => Promise<void>;
+  /** fetch data structure using fetch in type config */
+  fetchDataStructure: (
```
kavilla, line 45: how come this over `fetchNextDataStructure`?  the name for me indicated that if i pass a path then i would get the `next` data structure.

however if `fetchDataStructure` then i would be expect to get my `current` data structure

src/plugins/data/public/ui/dataset_selector/dataset_explorer.tsx
```diff
@@ -203,15 +203,23 @@ export const DatasetExplorer = ({
                 >
                   <h3>{current.columnHeader}</h3>
                 </EuiTitle>
-                {current.multiSelect ? (
+                {current.DataStructureCreator ? (
```
kavilla, line 206: should we do something similar to index pattern and indices?

instead of having conditional. we just pass the dataset creation component in. in which case it would be  whatever is the default.

otherwise we will have this contract that 
if `DatasetCreationComponent` is defined then regardless of `multiSelect` it will show whatever you pass us

src/plugins/data/common/datasets/types.ts
```diff
@@ -125,7 +142,12 @@ export interface DataStructure {
   children?: DataStructure[];
   hasNext?: boolean;
   paginationToken?: string;
+  /** Whether data structure supports multi-selecting children. The selected
+   * children will be combined using {@link DatasetTypeConfig['combineDataStructures']} */
   multiSelect?: boolean;
+  /** Custom component to select children and create next data structure or
+   * dataset. multiSelect will be ignored if using custom component */
+  DataStructureCreator?: React.ComponentType<DataStructureCreatorProps>;
```
kavilla, line 150: nit:

would it make more sense to be like
`customExplorerComponent` or `customComponent` or something. contract wise to put the comment into the name.`

src/plugins/data/common/datasets/types.ts
joshuali925: if i pass in the current `path: [Indexes, Cluster]`, it will fetch the `children` of the last data structure: `Cluster` is returned, with all indices as children.

i found `fetchNextDataStructure` confusing since the next data structure should be `Index`, not `Cluster`, but this function still returns `Cluster` data structure. maybe fetch children?

src/plugins/data/public/ui/dataset_selector/dataset_table.tsx
joshuali925: we can change name, i couldn't think of a good name. but

> we aren't really creating data structures we are trying to create a dataset

my understanding is that dataset is something user can search, but maybe this custom component would be used to create/select a data structure, which will have another custom component that would be used to create/select the dataset?

src/plugins/data/public/ui/dataset_selector/dataset_explorer.tsx
joshuali925: this is to avoid breaking change for other types. If we now ask dataset to pass in the component, then all dataset type config need to change?

src/plugins/data/common/datasets/types.ts
joshuali925: wouldn't they be too generic? Developers probably won't understand what the component does by looking at the name

sejli (approved):

PR #10486, opened 2025-09-04
[Discover Bug] Remove extra empty space above editor and fix default horizontal scroll in logs table

ruchidh:
### Description
This pr fixes 2 ui bugs in new discover
1. Remove extra line above query editor.
2. Fix default horizontal scroll in discover table.

## Screenshot

Screenshot 2025-09-04 at 4 21 29 PM

https://github.com/user-attachments/assets/dab12fc4-1800-4965-82e4-d3cfcfa42495

## Changelog
- fix: Remove extra empty space above editor and fix default horizontal scroll in logs table

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/pages/explore_page.scss
+++ b/src/plugins/explore/public/application/pages/explore_page.scss
@@ -62,7 +62,6 @@ $osdHeaderOffset: $euiHeaderHeightCompensation;
 }
 
 .dscCanvas {
-
   // Ensure proper spacing and layout
   .dscCanvas__queryPanel {
     margin-bottom: $euiSizeS;


config/opensearch_dashboards.yml
```diff
@@ -387,4 +387,4 @@
 # banner.content: "This is an important announcement for all users. [Learn more](https://opensearch.org) about OpenSearch."
 
 # Set the value to true to enable the keyboard shortcuts
-# opensearchDashboards.keyboardShortcuts.enabled: true
\ No newline at end of file
+# opensearchDashboards.keyboardShortcuts.enabled: true
```
angle943: nitpick: can we not check in the `opensearch_dashboards.yml`? its unnecessary change and adds frustration when checking out your PR locally (when we have changes in this file)

src/plugins/data/public/ui/query_editor/query_editor_top_row.tsx
```diff
@@ -416,27 +417,31 @@ export default function QueryEditorTopRow(props: QueryEditorTopRowProps) {
...
   return (
-    <EuiFlexGroup
-      className={classes}
-      responsive={!!props.showDatePicker}
-      gutterSize="xs"
-      direction="column"
-      justifyContent="flexEnd"
-    >
+    <>
+      {/* Portals for dataset selector and date picker */}
```
angle943: can we remove this comment? Its pretty clear what goes here, and if we add more stuff here later, i'm pretty sure its going to get stale real fast

src/plugins/data/public/ui/search_bar/search_bar.scss
```diff
@@ -0,0 +1,4 @@
+.globalQueryEditor,
+.globalQueryBar {
+    display: contents;
+}
```
angle943: this is very dangerous, its going to make the styling applied to these elements void. That means that every styling we have for these two elements are essentially unapplied, which leads to unmaintainable code.

But i just looked through the codebase and see this is a frustrating problem with no simple fix. I think we should scope this styling to be specific to `explore`. So define & scope this within a explore page styling, and a comment would be good explaining why its so. And I think in the context of explore, we only care about `.globalQueryEditor`.

src/plugins/explore/public/components/data_table/explore_data_table.scss
```diff
@@ -1,4 +1,9 @@
 .explore-table-container {
   height: 100%;
   mask-image: none;
+
+  /* stylelint-disable-next-line @osd/stylelint/no_modifying_global_selectors */
+  .euiFlexGroup {
+    width: 100%;
+  }
```
angle943, line 7: this is going to target all `.euiFlexGroup` within the explore-table-container. Lets not ignore these stylelint group unless absolute necessary.

you can just add a class to `EuiFlexGroup` and target it like that

angle943 (changes requested):

src/plugins/explore/public/components/data_table/explore_data_table.scss
ruchidh: got it

src/plugins/data/public/ui/search_bar/search_bar.scss
ruchidh: good point, will add this

config/opensearch_dashboards.yml
ruchidh: there are no changes though, i tried update it with main too but seems it is still showing up though no difefrence

angle943: you can just not check it in when you commit

ruchidh: sure, will take care of this from next time

angle943 (approved):

LDrago27 (approved):

PR #10487, opened 2025-09-05
[Explore] Add facetedFields to DataView/IndexPattern

Maosaic:
### Description

1. Introduce a new dataset atrribute `facetedFields` to allow dataset to control faceted fields filters
2. Only show faceted fields filter on Traces page

## Screenshot

Screenshot 2025-09-04 at 7 34 12 PM

Screenshot 2025-09-04 at 7 33 45 PM

## Changelog

- feat: Allow faceted fields filter control by dataset 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/dashboard/public/application/components/dashboard_top_nav/__snapshots__/dashboard_top_nav.test.tsx.snap
+++ b/src/plugins/dashboard/public/application/components/dashboard_top_nav/__snapshots__/dashboard_top_nav.test.tsx.snap
@@ -885,6 +885,7 @@ exports[`Dashboard top nav render in embed mode 1`] = `
               "getInjectedVar": [MockFunction],
             },
             "keyboardShortcut": Object {
+              "getAllShortcuts": [MockFunction],
               "register": [MockFunction],
               "unregister": [MockFunction],
               "useKeyboardShortcut": [MockFunction],
@@ -2043,6 +2044,7 @@ exports[`Dashboard top nav render in embed mode, and force hide filter bar 1`] =
               "getInjectedVar": [MockFunction],
             },
             "keyboardShortcut": Object {
+              "getAllShortcuts": [MockFunction],
               "register": [MockFunction],
               "unregister": [MockFunction],
               "useKeyboardShortcut": [MockFunction],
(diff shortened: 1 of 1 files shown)


src/plugins/data/common/data_views/data_views/data_views.ts
```diff
@@ -394,6 +395,7 @@ export class DataViewsService {
     const parsedTypeMeta = typeMeta ? JSON.parse(typeMeta) : undefined;
     const parsedFieldFormatMap = fieldFormatMap ? JSON.parse(fieldFormatMap) : {};
     const parsedFields: DataViewFieldSpec[] = fields ? JSON.parse(fields) : [];
+    const parsedFacetedFields = facetedFields ? JSON.parse(facetedFields) : undefined;
```
kavilla, line 398: the original plan was to have fieldspec actually contain the facet information. the reason being is that there is now two sources of conflicting fields.

for example,

user creates a dataset with this change and in theory it would look like
```
{
    title: index-*
    fields: [
       timestamp,
       message,
       statusCode
    ],
    facetedFields: [ "statusCode" ]
 }
 ```
 
 then they update their mapping or their index templates to change from `statusCode` to `status` which is one way users have been able increase the searchable results before hitting a limit. so now they refresh their fields in their dataset to be
 
 ```
 {
    title: index-*
    fields: [
       timestamp,
       message,
       status
    ],
    facetedFields: [ "statusCode" ]
 }
 ```
 
now the dataset saved object has a faceted field that actually doesn't exist. it also gives room for the future for being able to be able to display the alias for a field that has a facet.
 
 @LDrago27 has actually recently done something i think with the field that stores topValues https://github.com/opensearch-project/OpenSearch-Dashboards/commit/0716100515442be28e2614e49e2fccd45eea1687#diff-e4a8adf64459836ae64feb53e43aea814e5821c4482220e80f9127ef3628511f i wonder if we could just leverage that and add one more property like `facetedValues` within the suggestions and then a top level property like `isFaceted` or something. then have a function in the dataset to be similar to `getTimeField()` to be like `getFacetedFields()`.
 
 comes with the added benefit of displaying the facetedValues for the auto suggestion first

PR #10489, opened 2025-09-05
[Explore] Patterns Flyout 'search with pattern' redirect

paulstn:
### Description

Implements 'search with pattern' functionality in the Patterns' tab flyout.
This button, when clicked, will create a query that will select every document with that pattern for the row in the pattern table, and open the logs tab with this query.

https://github.com/user-attachments/assets/0deab5a6-45b2-4e95-afeb-7c0453267f7d

For example, within the opensearch sample flights index, the most common pattern for `Origin` is `<*> <*> <*> International Airport`. The query will be 
```
source = opensearch_dashboards_sample_data_flights | patterns `Origin` method=brain | where patterns_field = '<*> <*> <*> Airport'
```

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/patterns_table/patterns_table_flyout/patterns_flyout_update_search.tsx
+++ b/src/plugins/explore/public/components/patterns_table/patterns_table_flyout/patterns_flyout_update_search.tsx
@@ -36,9 +36,26 @@ export const PatternsFlyoutUpdateSearch = ({ patternString }: PatternsFlyoutUpda
   const setEditorText = useSetEditorText();
   const dispatch = useDispatch();
 
-  const query = useSelector(selectQuery);
-  const patternsField = useSelector(selectPatternsField);
+  const originalQuery = useSelector(selectQuery);
+  const selectedPatternsField = useSelector(selectPatternsField);
   const usingRegexPatterns = useSelector(selectUsingRegexPatterns);
+
+  // craft query that will select for all documents with specific pattern
+  const createSearchPatternQuery = (patternsField: string) => {
+    const preparedQuery = getQueryWithSource(originalQuery);
+    return usingRegexPatterns
+      ? regexUpdateSearchPatternQuery(preparedQuery.query, patternsField, patternString)
+      : brainUpdateSearchPatternQuery(preparedQuery.query, patternsField, patternString);
+  };
+
+  // move to 'logs' flow with 'update search' query
+  const redirectToLogsWithQuery = (query: string) => {
+    dispatch(setQueryStringWithHistory(query)); // sets query to be ran
+    setEditorText(query); // sets the text in the editor
+    dispatch(setActiveTab(EXPLORE_LOGS_TAB_ID)); // moves tab over to logs
+    dispatch(executeQueries({ services }));
+    closePatternsTableFlyout();
+  };
 
   return (
     <EuiButton
@@ -47,20 +64,10 @@ export const PatternsFlyoutUpdateSearch = ({ patternString }: PatternsFlyoutUpda
       })}
       iconType={'continuityBelow'}
       onClick={() => {
-        if (!patternsField) throw new Error('no patterns field');
+        if (!selectedPatternsField) throw new Error('no patterns field');
 
-        // craft query that will select for all documents with specific pattern
-        const preparedQuery = getQueryWithSource(query);
-        const newQuery = usingRegexPatterns
-          ? regexUpdateSearchPatternQuery(preparedQuery.query, patternsField, patternString)
-          : brainUpdateSearchPatternQuery(preparedQuery.query, patternsField, patternString);
-
-        // move to 'logs' flow with 'update search' query
-        dispatch(setQueryStringWithHistory(newQuery));
-        setEditorText(newQuery);
-        dispatch(setActiveTab(EXPLORE_LOGS_TAB_ID));
-        dispatch(executeQueries({ services }));
-        closePatternsTableFlyout();
+        const newQuery = createSearchPatternQuery(selectedPatternsField);
+        redirectToLogsWithQuery(newQuery);
       }}
     >
       {i18n.translate('explore.patterns.flyout.updateSearch', {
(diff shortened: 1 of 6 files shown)


paulstn:
The currently failing test is from an unrelated component, where the snapshot is incorrect. This test is failing for other PRs currently as well. Can ignore that test for now.

src/plugins/explore/public/components/patterns_table/patterns_table_flyout/patterns_flyout_update_search.tsx
```diff
@@ -0,0 +1,75 @@
...
+  };
+
+  return (
+    <EuiButton
+      aria-label={i18n.translate('explore.patterns.flyout.updateSearchWithPattern', {
+        defaultMessage: 'Update search with pattern',
+      })}
+      iconType={'continuityBelow'}
+      onClick={() => {
+        if (!selectedPatternsField) throw new Error('no patterns field');
```
ruchidh: Can we not move this call back to a method ?
Reason
1. Moving the callback to a method (e.g., a named function or class method) keeps the render logic clean and readable.
2. It avoids creating a new function instance on every render, which can help with performance and prevent unnecessary re-renders

src/plugins/explore/public/components/patterns_table/patterns_table_flyout/patterns_flyout_event_table.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  if (!dataset || !patternsField || !timeFieldName)
+    throw new Error('Dataset, patterns field, or time field is not appearing for event table');
+
+  const [fetchedItems, setFetchedItems] = useState<EventTableItem[]>([]);
+  const [pageIndex, setPageIndex] = useState(0);
+  const [tableLoading, setTableLoading] = useState(false);
+
+  const eventResults = async (page: number) => {
+    /**
```
ruchidh, line 55: What do you think about handling error for this async call ?

like ,


const eventResults = async (page: number) => {
  setError(null);
  try {
    // ...existing logic
  } catch (err) {
    setError('Failed to fetch event results'); // we can set any error too to display client error message in case fails
    setFetchedItems([]);
  }
  setTableLoading(false);
};

src/plugins/explore/public/components/patterns_table/patterns_table_flyout/patterns_flyout_event_table.tsx
```diff
@@ -0,0 +1,146 @@
...
+        event: row._source[patternsField],
+      };
+    });
+
+    setFetchedItems(items);
+    setTableLoading(false);
+  };
+
+  useEffect(() => {
+    setTableLoading(true);
```
ruchidh, line 111: The effect only runs once, but if any prop or selector changes, the data will not refresh.
is this useeffect required so ?

like patternstring, dataset, query , etc.

src/plugins/explore/public/components/patterns_table/patterns_table_flyout/patterns_flyout_event_table.tsx
```diff
@@ -0,0 +1,146 @@
...
+    eventResults(0);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  return (
+    <EuiBasicTable
+      items={fetchedItems}
+      columns={[
+        {
```
ruchidh: The columns array is recreated on every render. Use useMemo for performance.
can we move this to a method which is optimized ?

src/plugins/explore/public/components/patterns_table/patterns_table_flyout/patterns_flyout_event_table.tsx
```diff
@@ -0,0 +1,146 @@
...
+  };
+
+  useEffect(() => {
+    setTableLoading(true);
+    eventResults(0);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  return (
+    <EuiBasicTable
```
ruchidh, line 142: If possible, ensure the table is accessible (ARIA roles, etc.), depending on what EuiBasicTable provides.

angle943 (dismissed):

src/plugins/explore/public/components/patterns_table/patterns_table_flyout/patterns_flyout_update_search.tsx
paulstn: Ah okay, yeah I'll move this to a method.

src/plugins/explore/public/components/patterns_table/patterns_table_flyout/patterns_flyout_event_table.tsx
paulstn: This table is only within the flyout, which doesn't change anything in redux for the selectors, and the whole component stops rendering if the `patternString` or `totalItemCount` changes. 

The main purpose of this useEffect is just to run this initial fetch when the table gets rendered, is there another good way of doing this? I think from the perspective of keeping this logic as simple as possible, this is the best way to indicate when I'm calling `eventResults`

src/plugins/explore/public/components/patterns_table/patterns_table_flyout/patterns_flyout_event_table.tsx
paulstn: Sounds good, I'll include this!

src/plugins/explore/public/components/patterns_table/patterns_table_flyout/patterns_flyout_event_table.tsx
paulstn: Will make this change.

For the future, is this only an issue for when I have variability in the column names? i.e. for this example, because I'm using `timeFieldName` and `patternsField`?

src/plugins/explore/public/components/patterns_table/patterns_table_flyout/patterns_flyout_event_table.tsx
ruchidh: then this logic seems fit. if not required any changes to table

angle943 (approved):

PR #10491, opened 2025-09-05
[Explore] Introduce dataset SignalType

Maosaic:
## Changelog

-feat: Introduce dataset SignalType

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/common/index_patterns/index_patterns/index_pattern.ts
+++ b/src/plugins/data/common/index_patterns/index_patterns/index_pattern.ts
@@ -43,6 +43,7 @@ import {
   IndexPatternFieldMap,
   IndexPatternSpec,
   SavedObjectReference,
+  SignalType,
   SourceFilter,
   TypeMeta,
 } from '../types';
@@ -68,6 +69,7 @@ interface SavedObjectBody {
   fieldFormatMap?: string;
   typeMeta?: string;
   type?: string;
+  signalType?: string;
 }
 
 type FormatFieldFn = (
@@ -100,6 +102,7 @@ export class IndexPattern implements IIndexPattern {
   public sourceFilters?: SourceFilter[];
   public dataSourceRef?: SavedObjectReference;
   public fieldsLoading?: boolean;
+  public signalType?: SignalType;
   private originalSavedObjectBody: SavedObjectBody = {};
   private shortDotsEnable: boolean = false;
   private fieldFormats: FieldFormatsStartCommon;
@@ -136,6 +139,7 @@ export class IndexPattern implements IIndexPattern {
     this.description = spec.description;
     this.timeFieldName = spec.timeFieldName;
     this.sourceFilters = spec.sourceFilters;
+    this.signalType = spec.signalType;
 
     this.fields.replaceAll(Object.values(spec.fields || {}));
     this.type = spec.type;
@@ -242,6 +246,7 @@ export class IndexPattern implements IIndexPattern {
       typeMeta: this.typeMeta,
       type: this.type,
       dataSourceRef: this.dataSourceRef,
+      signalType: this.signalType,
     };
   }
 
@@ -382,6 +387,7 @@ export class IndexPattern implements IIndexPattern {
       fieldFormatMap,
       type: this.type,
       typeMeta: this.typeMeta ? JSON.stringify(this.typeMeta) : undefined,
+      signalType: this.signalType,
     };
   }
 
--- a/src/plugins/data/common/index_patterns/types.ts
+++ b/src/plugins/data/common/index_patterns/types.ts
@@ -37,6 +37,12 @@ import { IFieldType } from './fields';
 
 export type FieldFormatMap = Record<string, SerializedFieldFormat>;
 
+export enum SignalType {
+  LOGS = 'logs',
+  METRICS = 'metrics',
+  Traces = 'traces',
+}
+
 export interface IIndexPattern {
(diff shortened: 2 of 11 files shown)


src/plugins/explore/public/components/query_panel/query_panel_widgets/dataset_select/dataset_select.test.tsx
```diff
@@ -53,16 +56,28 @@ jest.doMock('../../../../../../opensearch_dashboards_react/public', () => ({
...
+            onFilter?: (dataset: any) => boolean;
+          }) => {
+            capturedOnFilter = onFilter;
+            return (
+              <div data-test-subj="dataset-select">
+                <button
+                  data-test-subj="dataset-select-button"
+                  onClick={() => onSelect({ id: 'test-dataset', type: 'index_pattern' })}
+                >
+                  Select Dataset
```
ruchidh, line 73: do we expect II8n here for select dataset

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
```diff
@@ -166,7 +167,36 @@ const fetchFirstAvailableDataset = async (
...
+            if (dataView?.signalType !== SignalType.Traces) {
+              return dataset;
+            }
+          }
+        } catch (error) {
+          // Continue to next dataset if this one fails
+          continue;
+        }
+      }
+      return undefined; // No compatible dataset found
```
ruchidh, line 196: if a function does not explicitly return a value, or if it uses a return; statement without a specified value, the function will implicitly return undefined

just in case we want to avoid hardcoded names

src/plugins/explore/public/components/query_panel/query_panel_widgets/dataset_select/dataset_select.test.tsx
Maosaic: this is test file

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
Maosaic: This is to maintain exisiting function behaviour

src/plugins/data/common/index_patterns/types.ts
```diff
@@ -37,6 +37,12 @@ import { IFieldType } from './fields';
 
 export type FieldFormatMap = Record<string, SerializedFieldFormat>;
 
+export enum SignalType {
+  LOGS = 'logs',
+  METRICS = 'metrics',
+  Traces = 'traces',
```
angle943, line 43: why only Traces not capitalized?

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
```diff
@@ -166,7 +167,36 @@ const fetchFirstAvailableDataset = async (
         typeConfig.toDataset([pattern])
       ) ?? [];
 
-    return fetchedDatasets.length > 0 ? fetchedDatasets[0] : undefined;
+    // Filter by SignalType compatibility
+    if (fetchedDatasets.length > 0) {
+      for (const dataset of fetchedDatasets) {
+        try {
+          const dataView = await services.data?.dataViews?.get(
```
angle943, line 174: should we batch this?

```
const dataViewPromises = fetchedDatasets.map(dataset =>
    services.data?.dataViews?.get(dataset.id, ...)
  );
  const dataViews = await Promise.allSettled(dataViewPromises);
```

the current implementation awaits them one by one

Maosaic: We're trying to find the first match, so we can avoid unnecessary fetch

src/plugins/data/common/index_patterns/types.ts
angle943: @TackAdam will take care of this in his followup PR

src/plugins/data/common/index_patterns/index_patterns/index_pattern.ts
```diff
@@ -100,6 +102,7 @@ export class IndexPattern implements IIndexPattern {
   public sourceFilters?: SourceFilter[];
   public dataSourceRef?: SavedObjectReference;
   public fieldsLoading?: boolean;
+  public signalType?: SignalType;
```
kavilla, line 105: this should be just a string.

`signalType` doesn't isn't limited to observability but based on this enum it is. it's fine that we have predefined enums but this field shouldn't be reserved for observability datasets only

Maosaic: We can expand the enum anytime we need to, even we don't and the response coming back with a different value that is not on the enum, it shouldn't cause runtime error.

kavilla: scalability will become a concern. observability does seem to be a large use case of OSD but it isn't the only use case. the framework is flexible enough for set this `signalType` to be `Events` or `Errors`

but it will make development a little bit more difficult if before my plugin can register a new dataset type of a new signal type. the core repo will need to update the enum to accept this new signal type or throw `ignores` in their code while they write.

normally im all for the enums but in this case this enum makes datasets less flexible. what about a union type 
```
public signalType?: SignalType | string;
```

PR #10499, opened 2025-09-08
Add sanitization for axis label and name 

abbyhu2000:
### Description

This adds sanitization to fix the security issue when axis labels are rendered as HTML without proper sanitization, allowing malicious HTML content to be injected and stored within visualizations.

## Changelog
- fix: Add sanitization for axis label and name 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/10499.yml
+++ b/changelogs/fragments/10499.yml
@@ -1,0 +1,2 @@
+fix:
+- Add sanitization for axis label and name ([#10499](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10499))


ruanyl (approved):

packages/osd-ui-shared-deps/flot_charts/jquery_flot_axislabels.js
```diff
@@ -153,37 +153,46 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
-            this.elem.css('left', box.left + box.width - this.labelWidth +
-                          'px');
-        }
+    HtmlAxisLabel.prototype.draw = function (box) {
+      const sanitizedAxisName = dompurify
+        .sanitize(this.axisName, {
+          ALLOWED_TAGS: [],
+          ALLOWED_ATTR: [],
+        })
+        .replace(/[^a-zA-Z0-9_-]/g, '');
```
angle943, line 162: not sure how this works, but doesn't this replace everything to empty string?

abbyhu2000:
```
dompurify.sanitize(this.axisName, {
  ALLOWED_TAGS: [],
  ALLOWED_ATTR: []
})
```
uses DOMPurify library to strip all HTML tags and attributes from axisName.

`.replace(/[^a-zA-Z0-9_-]/g, '')` restricts the output to only alphanumeric characters, underscores, and hyphens
so any character not matching a-z, A-Z, 0-9, _, or - gets removed. 
So the combination above is to create an absolutely safe string that can be used without any risk of XSS

sejli (approved):

PR #10504, opened 2025-09-09
Trace:Details Cypress testing

TackAdam:
### Description
Trace:Details Cypress testing
Remove sanity test that are covered.
Combine Discover:Trace and Trace Detail testing as they are dependent and use same workspace to save CI time.

## Screenshot
Screenshot 2025-09-10 at 11 51 19 AM

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


angle943 (dismissed):

ruchidh (approved):

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
```diff
@@ -0,0 +1,633 @@
...
+  TRACE_INDEX_PATTERN,
+  TRACE_TIME_FIELD,
+  TRACE_INDEX,
+} from '../../../../../../utils/apps/explore/constants';
+import { getRandomizedWorkspaceName } from '../../../../../../utils/apps/explore/shared';
+import { prepareTestSuite } from '../../../../../../utils/helpers';
+
+const workspaceName = getRandomizedWorkspaceName();
+
+const traceDetailsTestSuite = () => {
```
kavilla: how come we add these tests into the traces discover test file?

all the setup seems similar but will add more resource consumption

TackAdam: This is the only way to get the environment set-up that i am aware of. My understanding is that we want each test suite isolated so they need to make their workspace / add data and then run the test.

Traces discover is navigated to through the Discover:Trace page so there is small over-lap but this is required to have the url for the Trace:Details page to run all of its test.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
```diff
@@ -0,0 +1,633 @@
...
+      });
+
+      // Navigate to traces page and capture trace URL
+      cy.osd.navigateToWorkSpaceSpecificPage({
+        workspaceName: workspaceName,
+        page: 'explore/traces',
+        isEnhancement: true,
+      });
+
+      // Click on dataset selector to close Syntax options if blocking time-picker
```
kavilla: you dont need to do this way. there is logic so we don't have users having to click the dataset button every time. so we can mimic that behavior like this

```
cy.window().then((win) => {
  win.localStorage.setItem('hasSeenInfoBox_PPL', true);
});
```

this will be more consistent because if for some reason it doesn't find the datasetSelectButton then this test will fail on stuff that is unrelated to this test

TackAdam: Good call out, was following the existing pattern for this issue. Will add a task for a follow-up to refactor that usage through-out are cypress testing.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
```diff
@@ -0,0 +1,633 @@
...
+
+      // Wait for span links and capture trace URL
+      cy.get('[data-test-subj="spanIdLink"]', { timeout: 30000 }).should('exist');
+
+      // Intercept window.open to capture URL
+      cy.window().then((win) => {
+        cy.stub(win, 'open').as('windowOpen');
+      });
+
+      cy.get('[data-test-subj="spanIdLink"]').first().click();
```
kavilla: i think we should improve the test subject in the actual code. to get us more leverage.

we can interpolate the span id into the data test subj so like `spanIdLink-${spanId}` or even if you want you could just do like `spanIdLink-${index}` so like `spanIdLink-0` `spanIdLink-1`...

this way we can then re-use the custom cyrpess commands and also find the exact element you are looking for

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
```diff
@@ -0,0 +1,633 @@
...
+    beforeEach(() => {
+      cy.visit(traceUrl);
+      cy.osd.waitForLoader(true);
+    });
+
+    describe('Page Load and Structure', () => {
+      it('should load trace details page with correct structure', () => {
+        // Verify page loads with trace ID in title
+        cy.url().should('include', 'traceDetails');
+        cy.get('h1').should('contain', 'Trace:');
```
kavilla: we should put a test subj on this if we don't already have one. cy.get('h1') will get a `h1` it doesn't necessarily restrict to only having 1 `h1` element on the DOM

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
```diff
@@ -0,0 +1,633 @@
...
+
+        // Verify Timeline tab is active by default
+        cy.get('button[role="tab"][aria-selected="true"]').should('contain', 'Timeline');
+
+        // Verify main panels are visible
+        cy.get('[data-test-subj="span-gantt-chart-panel"]').should('be.visible');
```
kavilla: we should use the existing custom  cypress commands `cy.getElementByTestId`

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
```diff
@@ -0,0 +1,633 @@
...
+
+        // Check for gantt chart container with fallback
+        cy.get('body').then(($body) => {
+          if ($body.find('.exploreGanttChart__container').length > 0) {
+            cy.get('.exploreGanttChart__container').should('be.visible');
+          } else {
+            // Fallback: check for SVG in the gantt panel
```
kavilla: this is a little bit confusing. is this saying that this test might sometimes do one thing or another thing?

shouldn't we know what is going to happen in the test?

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
```diff
@@ -0,0 +1,633 @@
...
+                    if ($tooltip.text().trim().length > 0) {
+                      // Tooltip has content, test passes
+                      cy.wrap($tooltip).should('contain.text', 'ms').or('not.be.empty');
+                    } else {
+                      // Fallback: just verify tooltip element exists
+                      cy.wrap($tooltip).should('exist');
+                    }
+                  });
+                } else {
+                  // No tooltip found, skip this test gracefully
```
kavilla: wouldn't we want to make sure the tooltip does exist?

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
```diff
@@ -0,0 +1,633 @@
...
+              // Trigger mouseover event
+              cy.wrap($path).trigger('mouseover', { force: true });
+
+              // Check if tooltip exists and try to make it visible
+              cy.get('body').then(($body) => {
+                if ($body.find('.vg-tooltip').length > 0) {
+                  // Tooltip exists, check if it becomes visible or has content
+                  cy.get('.vg-tooltip').should('exist');
+
+                  // Try different approaches to verify tooltip functionality
```
kavilla: is there no test subject?

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
```diff
@@ -0,0 +1,633 @@
...
+
+        // Verify right panel has span data
+        cy.contains('Service identifier').should('be.visible');
+        cy.contains('Span ID').should('be.visible');
+      });
+    });
+  });
+};
+
+prepareTestSuite('Trace Details Interactive Tests', traceDetailsTestSuite);
```
kavilla: if this is is what is outputted to the log. it is easier to find if the file name matches the spec name

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
```diff
@@ -0,0 +1,633 @@
...
+      });
+
+      it('should load trace data within reasonable time', () => {
+        // Verify page loads quickly
+        cy.get('[data-test-subj="span-gantt-chart-panel"]', { timeout: 15000 }).should(
+          'be.visible'
+        );
+
+        // Verify gantt chart and span data is loaded
+        cy.get('body').then(($body) => {
```
kavilla: if i remember correctly. the problem with accessing body is that once you access the body it will access it successfully right away. however if the body was accessed before the elements were ever loaded then it will never refetch the body.

we should avoid access the body like this. it will make the test flakey

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
kavilla: for me its a little bit law of diminishing returns. roughly, every spec file setup adds an additional 10 mins to the CI. also each setup hits the system hard and starves out resources. which causes tests to fail because the tests are actually not ran in isolation even though we want it to be that way https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/cypress.config.ts#L61.

imo, i think saving an extra +10 mins on the CI and less flaky results because the resources consumed is preferable. so we could just have the traces test file setup up the workspace datasource and dataset. and then we have two describes for `Discover:Trace` and `Trace:Details`

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
kavilla: thanks yeah. i started doing that whenever i touch a file. that does this

kavilla (dismissed):

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
TackAdam: As discussed combined testing files to traces.spec.js

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
TackAdam: Added
cy.window().then((win) => {
  win.localStorage.setItem('hasSeenInfoBox_PPL', true);
});
And removed previous click item implementation as discussed.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
TackAdam: Replaced with
      cy.getElementByTestId('headerApplicationTitle').contains(
        'Trace: 68b0ad76fc05c5a5f5e3738d42b8a735'
      );

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
TackAdam: Added where possible
cy.getElementByTestId('span-gantt-chart-panel').should('be.visible');

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
TackAdam: Refactored to be reliable

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
TackAdam: Refactored to be reliable

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/04/traces_details.spec.js
TackAdam: Adjusted to 
prepareTestSuite('Traces', traceTestSuite);

angle943 (dismissed):

Maosaic (approved):

angle943 (approved):

PR #10505, opened 2025-09-09
Capitalize PPL Keywords

LDrago27:
### Description

Capitalize PPL Keywords in suggestions. 

## Screenshot

https://github.com/user-attachments/assets/45df39c6-ac20-496b-a2e8-65476b850a86

## Changelog

- fix: Capitalize PPL Keywords in suggestions

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/10505.yml
+++ b/changelogs/fragments/10505.yml
@@ -1,0 +1,2 @@
+fix:
+- Capitalize PPL Keywords in suggestions ([#10505](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10505))
--- a/cypress/utils/apps/explore/autocomplete.js
+++ b/cypress/utils/apps/explore/autocomplete.js
@@ -126,14 +126,14 @@ const getLanguageSpecificConfig = (language, config) => {
     case QueryLanguages.PPL.name:
       return {
         initialCommands: [
-          { value: 'source', input: 's' },
+          { value: 'SOURCE', input: 's' },
           { value: '=' },
           { value: getDatasetName('data_logs_small_time_1', config.datasetType) },
           { value: '|' },
-          { value: 'where', input: 'w' },
+          { value: 'WHERE', input: 'w' },
         ],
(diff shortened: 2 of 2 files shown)


angle943 (approved):

src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
```diff
@@ -289,24 +289,23 @@ export const getSimplifiedPPLSuggestions = async ({
...
             return {
-              text: sk.value.toLowerCase(),
-              insertText: getInsertText(sk.value.toLowerCase(), 'keyword', isInQuotes),
+              text: sk.value,
+              insertText: getInsertText(sk.value, 'keyword', isInQuotes),
               type: monaco.languages.CompletionItemKind.Keyword,
               detail: SuggestionItemDetailsTags.Keyword,
               // sortText is the only option to sort suggestions, compares strings
-              sortText:
-                PPL_SUGGESTION_IMPORTANCE.get(sk.id)?.importance ?? '98' + sk.value.toLowerCase(), // '98' used to devalue every other suggestion
```
paulstn, line 309: The `.toLowerCase()` here is being used in sortText, which compares the values of the strings. The value of uppercase text will be lower than lowercase, right? 

Also, when we don't put a function on text to either make them all lowercase or all uppercase, they might not be all normalized, and the order could be different. (which isn't an issue with the current code since you changed it to just show the value as is in the grammar, so text in all caps will appear before lowercase which is expected) 

Additionally, there might be some characters which have their order changed, due to being in between uppercase and lowercase letters, such as square brackets, underscore, backslash, etc. 

Ultimately there isn't anything here you should change IMO, but its worth keeping an eye on.

LDrago27: Thanks @paulstn . For providing the context on the change. I will check if this changes the relative ordering.

PR #10513, opened 2025-09-10
[Explore] Split aggregations query from logs query

sejli:
## Changelog

- fix: [Explore] Split Aggregations Query from Logs Query

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
+++ b/src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
@@ -64,6 +64,13 @@ export const defaultPrepareQueryString = (query: Query): string => {
 };
 
 /**
+ * Prepare cache key for data table queries (no aggregations)
+ */
+export const prepareDataTableCacheKey = (query: Query): string => {
+  return `datatable:${defaultPrepareQueryString(query)}`;
+};
+
+/**
  * Default results processor for tabs
  * Processes raw hits to calculate field counts and optionally includes histogram data
  * Also updates topQueryValues for string fields to improve autocomplete performance
@@ -182,7 +189,7 @@ export const histogramResultsProcessor: HistogramDataProcessor = (
 };
 
 /**
- * Enhanced executeQueries orchestrator (simplified - no cache logic)
+ * Enhanced executeQueries orchestrator - executes queries independently without blocking
  */
 export const executeQueries = createAsyncThunk<
   void,
@@ -199,6 +206,44 @@ export const executeQueries = createAsyncThunk<
   }
 
   const defaultCacheKey = defaultPrepareQueryString(query);
+  // Use separate cache keys for data table and histogram
+  const dataTableCacheKey = prepareDataTableCacheKey(query);
+  const histogramCacheKey = defaultCacheKey;
+
+  // Check what needs execution for core queries
+  const needsDataTableQuery = !results[dataTableCacheKey];
+  const needsHistogramQuery = !results[histogramCacheKey];
+
+  const promises = [];
+
+  // Execute query without aggregations
+  if (needsDataTableQuery) {
+    promises.push(
+      dispatch(
+        executeDataTableQuery({
+          services,
+          cacheKey: dataTableCacheKey,
+          queryString: defaultPrepareQueryString(query),
+        })
+      )
+    );
+  }
+
+  if (needsHistogramQuery) {
+    const interval = state.legacy?.interval;
+    promises.push(
+      dispatch(
+        executeHistogramQuery({
+          services,
+          cacheKey: histogramCacheKey,
+          queryString: defaultPrepareQueryString(query),
+          interval,
+        })
+      )
+    );
+  }
+
+  // Handle tab queries as before (keeping existing tab logic)
   const visualizationTab = services.tabRegistry.getTab('explore_visualization_tab');
   let visualizationTabPrepareQuery = defaultPrepareQueryString;
   if (visualizationTab?.prepareQuery) {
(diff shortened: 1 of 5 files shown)


ruanyl:
Great fix!

ruanyl:
@sejli btw, the unit test failures seems relate to this PR

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -199,6 +206,44 @@ export const executeQueries = createAsyncThunk<
   }
 
   const defaultCacheKey = defaultPrepareQueryString(query);
+  // Use separate cache keys for data table and histogram
+  const dataTableCacheKey = prepareDataTableCacheKey(query);
+  const histogramCacheKey = defaultCacheKey;
```
angle943: nitpick: curious why histogram became the default cache key? I feel like histogram should have special handling instead instead of the data table

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -583,6 +623,28 @@ export const executeTabQuery = createAsyncThunk<
...
+>('query/executeDataTableQuery', async (params, thunkAPI) => {
+  return executeQueryBase(
+    {
+      ...params,
+      includeHistogram: false, // Data table doesn't need histogram
+      interval: undefined, // Data table doesn't need intervals
+    },
+    thunkAPI
+  );
+});
```
angle943, line 675: can we add unit test for this funciton?

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -63,6 +74,150 @@ export const defaultPrepareQueryString = (query: Query): string => {
...
+      return {
+        key: sourceValues[1],
+        value: sourceValues[0],
+      };
+    });
+  }
+
+  const tempResult: ISearchResult = { ...rawResults, aggregations: {} };
+
+  Object.entries(responseAggs).forEach(([id, value]) => {
```
paulstn: Instead of creating a `responseAggs` in line 165 and then deconstructing it here to create the `tempResults` with timestamped buckets, I think we should just create `tempResults` with the timestamped buckets from the get-go. 

Having looked at the code for how the `ppl_search_strategy` deconstructs the result into k/v pairs and then the dataframe later does the below bucketing, I understand why it was put in this way, but since this new function is very directly creating the histogram, let's just make it as simple as possible.

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -397,12 +595,20 @@ const executeQueryBase = async (
       .ok({ json: rawResults });
 
     // Store RAW results in cache
-    const rawResultsWithMeta: ISearchResult = {
+    let rawResultsWithMeta: ISearchResult = {
```
paulstn, line 456: Instead of having `lets` throughout this function, would it be simple enough to keep all of the `consts`? I believe there was some reason for preferring `consts`, does anyone else who knows want to chime in for why?

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -63,6 +74,150 @@ export const defaultPrepareQueryString = (query: Query): string => {
...
+    histogramConfigs,
+    aggs,
+    effectiveInterval,
+    fromDate,
+    toDate,
+    timeFieldName: dataView.timeFieldName,
+  };
+};
+
+export const buildQuery = (
```
paulstn: Could you move this function outside into a utils file?

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -63,6 +74,150 @@ export const defaultPrepareQueryString = (query: Query): string => {
   }
 };
 
+/**
+ * Creates histogram configuration with computed interval and time range
+ */
+const createHistogramConfigWithInterval = (
```
paulstn: Can you also move this function outside into a utils file?

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -347,13 +535,23 @@ const executeQueryBase = async (
...
+      histogramConfig = createHistogramConfigWithInterval(dataView, interval, services, getState);
+    }
+
     const preparedQueryObject = {
       ...query,
       dataset,
-      query: cacheKey,
+      query:
+        isHistogramQuery && histogramConfig
+          ? buildQuery(queryString, histogramConfig, services)
```
paulstn: I think the `createHistogramConfigWithInterval` and `buildQuery` flow might contain a little redundancy here. From a general overview, it seems to me like we only need the timeFieldName from dataView, the interval, and the current query to create the histogram query. Looking at the `createHistogramConfigWithInterval` and `buildQuery` functions, I think they contain a lot of vestigial components from the old DSL way of creating the histogram. 

Could you take another look at these functions and see if you can remove a lot of the unneeded complexity?

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
sejli: We use `const` to prevent reassigning something to another variable. I changed this to `let` because `rawResultsWithMeta` is possibly updated if the incoming result is part of a histogram query.

paulstn: There was some other reason to prefer using `const`, to do with performance, and to get around the behavior of needing to reassign a variable, I believe that usage of iterative methods (such as `map()`) substituted. I know that @kavilla commented on some of my other prs for this, he would definitely know whether or not this is important, but for now, I am fine with just letting this be.

paulstn (commented):
All changes should be good to handle in a future PR.

PR #10527, opened 2025-09-16
[Explore] Support data link for table chart

lunalzm:
### Description

This PR adds support for data links in the table chart visualization. Users can now configure multiple data links for specific fields in the table, allowing cells in those fields to be clickable links to external URLs. The feature includes a multi-select Fields combo box in the Data Link modal, enabling users to associate a single link with multiple fields and allowing fields to have multiple associated links. When a cell has multiple links, a popover with a context menu displays the available links. The implementation ensures type safety and uses human-readable field names in the UI while storing internal column IDs for linking.

## Screenshot

https://github.com/user-attachments/assets/855f2184-c03e-4075-a4d5-6ddc41457edc

### Add a Data Link:
1. Open the style panel and click "Add Link" in the Data Links section.
2. In the modal, enter a title, a URL (e.g., https://example.com/${__value.text}), and select multiple fields from the Fields combo box.
3. Save the link and confirm it appears in the link list.

### Test Single Link:
1. In the table, click a cell in a column with a single link.
2. Verify that it navigates to the resolved URL in the correct tab (_blank or _self based on the "Open in new tab" setting).

### Test Multiple Links:

1. Add another link for the same field(s) and save.
2. Click a cell in a column with multiple links.
3. Confirm that a popover appears with a context menu listing all link titles.
4. Select a link from the menu and verify navigation to the correct URL.

### Edit and Delete Links:

1. Edit an existing link and verify that the modal pre-populates with the correct title, URL, and selected fields.
2. Delete a link and confirm it is removed from the link list and no longer applies to table cells.

## Changelog

- feat: Add data link support to table chart

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/visualizations/table/cell_value.tsx
```diff
@@ -37,5 +72,84 @@ export const CellValue = (props: Props) => {
...
 
-  return value;
+  const resolveUrl = (url: string) => {
+    return url.replace('${__value.text}', encodeURIComponent(value || ''));
+  };
+
+  const getAnchorPosition = () => {
+    if (!buttonRef.current || !columnWidth) return 'downCenter';
+    const textWidth = buttonRef.current.scrollWidth;
+    if (textWidth > columnWidth || columnWidth > 200) {
```
raintygao: Why do we compare columnWidth with 200, could you elaborate more?

src/plugins/explore/public/components/visualizations/table/data_link_options.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+const DataLinkModal: React.FC<DataLinkModalProps> = ({
+  link,
+  onSave,
+  onCancel,
+  availableFields,
+}) => {
+  const [title, setTitle] = useState(link?.title || '');
+  const [url, setUrl] = useState(link?.url || '');
+  const [openInNewTab, setOpenInNewTab] = useState(link?.openInNewTab || false);
```
raintygao: Although the result would be same, but I think here using `??` is better than `||`? If the value of `openInNewTab` is `false`, actually we don't need `||` for fallback?

src/plugins/explore/public/components/visualizations/table/data_link_options.tsx
```diff
@@ -0,0 +1,281 @@
...
+        return fieldOption ? { label: fieldOption.label, value: fieldOption.value } : null;
+      })
+      .filter((option): option is { label: string; value: string } => option !== null) || []
+  );
+
+  const isSaveDisabled = !title.trim() || !url.trim() || selectedFields.length === 0;
+
+  const handleSave = () => {
+    onSave({
+      id: link?.id || Math.random().toString(36).substr(2, 9),
```
raintygao: What is `Math.random().toString(36).substr(2, 9)` for?  Is this for generating unique ID?

src/plugins/explore/public/components/visualizations/table/data_link_options.tsx
ruanyl, line 1: FYI, don't forget to add unit tests for this component :)

src/plugins/explore/public/components/visualizations/table/data_link_options.tsx
lunalzm: Yes.

src/plugins/explore/public/components/visualizations/table/cell_value.tsx
lunalzm: The `columnWidth > 200` check is to handle cells with very long text.
By default the popover tries to align to the text’s center, which looks off when the text overflows the cell.
When the column is wide enough, we switch to align based on the cell width instead, so the popover position stays consistent with the cell rather than the long text.

src/plugins/explore/public/components/visualizations/table/data_link_options.tsx
raintygao: This approach seems not always unique, OSD has already used `uuid`, what about leveraging this?

lunalzm: Good point!

src/plugins/explore/public/components/visualizations/table/cell_value.tsx
```diff
@@ -3,20 +3,55 @@
...
+    colorMode,
+    dataLinks,
+    isPopoverOpen,
+    setPopoverOpen,
+    columnId,
+  } = props;
+  const buttonRef = useRef<HTMLButtonElement>(null);
+  const [columnWidth, setColumnWidth] = useState<number | null>(null);
+
+  useEffect(() => {
```
ruanyl: Can you add comment to this `useEffect` to describe the intention?

src/plugins/explore/public/components/visualizations/table/cell_value.tsx
```diff
@@ -37,5 +72,84 @@ export const CellValue = (props: Props) => {
...
+    href: resolveUrl(link.url),
+    target: link.openInNewTab ? '_blank' : '_self',
+    rel: link.openInNewTab ? 'noopener noreferrer' : undefined,
+  }));
+
+  if (setPopoverOpen && isPopoverOpen !== undefined) {
+    return (
+      <EuiPopover
+        button={
+          <button
```
ruanyl: Should this be `a` tag?

src/plugins/explore/public/components/visualizations/table/cell_value.tsx
```diff
@@ -37,5 +72,84 @@ export const CellValue = (props: Props) => {
...
+        anchorPosition={getAnchorPosition()}
+        repositionOnScroll
+      >
+        <EuiContextMenu
+          initialPanelId={0}
+          panels={[
+            {
+              id: 0,
+              title: i18n.translate('explore.stylePanel.table.cellValue.selectLink', {
+                defaultMessage: 'Select a link',
```
ruanyl: What about not showing the title? And instead add a OUI `link` icon to the context menu items? https://oui.opensearch.org/1.21/#/display/icons

src/plugins/explore/public/components/visualizations/table/data_link_options.tsx
lunalzm: Ok, I've already add it.

ruanyl (approved):

src/plugins/explore/public/components/visualizations/table/cell_value.tsx
```diff
@@ -37,5 +55,71 @@ export const CellValue = (props: Props) => {
     });
   }, [setCellProps, value, textAlign, color, colorMode]);
 
-  return value;
+  const resolveUrl = (url: string) => {
+    return url.replace('${__value.text}', encodeURIComponent(value || ''));
```
wanglam, line 59: Nit: Do we need any place to store `${__value.text}` as a constant variable?

src/plugins/explore/public/components/visualizations/table/data_link_options.tsx
```diff
@@ -0,0 +1,333 @@
...
+        >
+          <EuiFlexGroup wrap gutterSize="s" alignItems="center">
+            {selectedFields.map((field) => (
+              <EuiFlexItem grow={false} key={field.value}>
+                <EuiBadge
+                  color="hollow"
+                  iconType="cross"
+                  iconSide="right"
+                  onClick={() => handleRemoveField(field.value)}
+                  onClickAriaLabel={`Remove ${field}`}
```
wanglam: Seems `field` is an object here. Shall we use `${field.label}` here?

src/plugins/explore/public/components/visualizations/table/data_link_options.tsx
```diff
@@ -0,0 +1,333 @@
...
+  styleOptions,
+  onStyleChange,
+  numericalColumns = [],
+  categoricalColumns = [],
+  dateColumns = [],
+}) => {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [editingLink, setEditingLink] = useState<DataLink | undefined>(undefined);
+
+  const availableFields = [
```
wanglam: Nit: seems this will be passed to `DataLinkModal` as a dependency of `useMemo`, can we wrap it with useMemo?

src/plugins/explore/public/components/visualizations/table/cell_value.tsx
lunalzm: For now we only have `${__value.text}`, so let’s just keep it this way.

src/plugins/explore/public/components/visualizations/table/data_link_options.tsx
lunalzm: Sure!

PR #10535, opened 2025-09-17
[Explore] Patterns Calcite workarounds

paulstn:
### Description

Due to some issues in the `patterns` command coming in from PPL, related to when Calcite gets enabled, this PR aims to work around as many of those issues as possible without creating something that will be too jarring when reverting back after these bugs stop existing. 

Here are the bugs that show up in the UI when calcite gets enabled: 
- Highlighting / future features on dynamic element selection is broken
- sample_logs not returned with results, stopping the page from rendering (as expected, there was no error state implemented here)
- Search with patterns redirects with a query as expected, but this leads to issues where the logs matching a pattern aren't returned, because the pattern unexpectedly changes without an aggregation being performed on it, for fields with high cardinality (unsolvable one-way door, better to wait for backend fix)
- Events table does not show any results

These are the workarounds where these issues (barring search with patterns) can be circumvented:
- changing the query being sent out (for brain) to explicitly do an aggregation and specify the stats/fields needed as to not rely on the mode=aggregation  (which changes unreliably under calcite)
- modifying the highlighting code to now be more general as to work with the new type of delimiter that calcite uses for the patterns_field
- using a query with a workaround for the sample log table, particularly in specifying a fields clause for the fields needed, which seems to return a consistent patterns string. arbitrary enough change to keep in even when the backend bug is fixed

Videos with these workarounds implemented:

https://github.com/user-attachments/assets/b5f3cfed-21de-4177-afaa-5f898d28b218

^Displays working patterns page with calcite enabled and calcite disabled. Includes a rendering page, working highlighting, and accurate events table. Shows how the search with query will put in the correct query, but currently that correct query hits the bug in calcite.

https://github.com/user-attachments/assets/a271653c-4f43-4e25-9a06-e0c33a8660d9

^Displays working regex pattern workflow.

## Changelog

- breaking: Patterns tab adaptation to calcite issues.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/patterns_table/patterns_container.tsx
```diff
@@ -53,7 +54,8 @@ const PatternsContainerContent = () => {
     const hasAllRequiredFields = requiredFields.every((field) => field in hit._source);
 
     if (!hasAllRequiredFields) {
-      return <></>;
+      // doesn't match normal fields or calcite fields
+      return <EuiCallOut title="Expected schema not found" color="danger" iconType="alert" />;
```
angle943: can we use i18n for the title?

paulstn: done

Maosaic (approved):

sejli (approved):

PR #10540, opened 2025-09-18
[Explore] Patterns tab Error Handling updates

paulstn:
### Description

image
Invalid query doesn't error out the page anymore (the 'errors' popping up from an incorrect query were not thrown as before), and now there is a UI page that shows the pattern query attempted.

## Changelog

- feat: Custom patterns error page
- fix: Prevent general query error from hiding tabs

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/patterns_table/utils/utils.ts
+++ b/src/plugins/explore/public/components/patterns_table/utils/utils.ts
@@ -215,8 +215,12 @@ export const findDefaultPatternsField = (services: ExploreServices): string => {
     return field.type === 'string';
   });
 
+  if (!logResults?.hits?.hits?.[0]) {
+    throw new Error('Cannot access hits from logs tab');
+  }
+
   // Get the first hit if available
-  const firstHit = logResults?.hits?.hits?.[0];
+  const firstHit = logResults.hits.hits[0];
 
   if (firstHit && firstHit._source && filteredFields) {
     // Find the field with the longest value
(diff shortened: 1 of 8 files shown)


src/plugins/explore/public/components/patterns_table/utils/utils.ts
```diff
@@ -215,8 +215,12 @@ export const findDefaultPatternsField = (services: ExploreServices): string => {
     return field.type === 'string';
   });
 
+  if (!logResults?.hits?.hits?.[0]) {
+    throw new Error('Cannot access hits from logs tab');
+  }
```
angle943, line 220: i'm weary of throwing errors like this as our fallback is pretty bad (and things we typically we want to do when you catch an error, such as log it, is not possible in our system), is there a way not to throw error?

angle943: edit: if this throws error, are you going to show the `patterns error guard` like your screenshot? because then its not bad, as they still have access to each tabs

paulstn: Oh so in this function, I throw a lot of errors for unexpected behavior (which was "causing" the original problem), but a step above where this is called, I run a try catch for this function so this error isn't going to be thrown and handled like any other error. 
I'm fine with changing this to however it needs to be, though. Should I remove all the `throw new Errors` I have in my code?

angle943: we synced offline, its fine as long as we do not get our users to a state where they no longer have access to tabs

LDrago27 (approved):

PR #10557, opened 2025-09-23
[Explore] show query panel even if there are no datasets

joshuali925:
### Description

Currently Discover shows a blank page when there is associated datasource (opensearch cluster) but no index-pattern. This is confusing to the user, because they should be able to still search the cluster using `Indexes` without creating an index-pattern, but they must create an index pattern first in order to see the dataset picker that will be used to pick `Indexes`.

This PR allows selecting/creating a dataset when there's no current dataset selected.

## Changelog
- feat: show query panel in Explore even if there are no datasets

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/pages/logs/logs_page.tsx
+++ b/src/plugins/explore/public/application/pages/logs/logs_page.tsx
@@ -18,7 +18,6 @@ import { useUrlStateSync } from '../../utils/hooks/use_url_state_sync';
 import { useTimefilterSubscription } from '../../utils/hooks/use_timefilter_subscription';
 import { useHeaderVariants } from '../../utils/hooks/use_header_variants';
 import { NewExperienceBanner } from '../../../components/experience_banners/new_experience_banner';
-import { useDatasetContext } from '../../context';
 import { BottomContainer } from '../../../components/container/bottom_container';
 import { TopNav } from '../../../components/top_nav/top_nav';
 import { useInitPage } from '../../../application/utils/hooks/use_page_initialization';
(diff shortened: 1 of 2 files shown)


src/plugins/explore/public/application/pages/logs/logs_page.tsx
```diff
@@ -94,7 +92,7 @@ export const LogsPage: React.FC<Partial<Pick<AppMountParameters, 'setHeaderActio
             <NewExperienceBanner />
 
             <div className="dscCanvas__queryPanel">
-              {dataset && !isLoading ? <QueryPanel /> : null}
+              <QueryPanel />
```
angle943, line 95: hmm why didn't we do this before

angle943: there's no error that you see with rendering the query panel when there is no dataset? (or page is loading?)

angle943: should we also make the changes for traces and metrics page?

joshuali925: > there's no error that you see with rendering the query panel when there is no dataset? 

there's no error, user sees two blank panels (side bar and results tab), and the query panel that includes dataset picker won't show up at all

> should we also make the changes for traces and metrics page?

maybe but i left it out since they should be working more out of the box, without requiring user to manually pick the index from `Indexes`

angle943: I think Traces was having that issue as well. Up to you to make the change or not but not a blocker

PR #10257, opened 2025-07-22
Enhance t2ppl to add time parsing logics

zhichao-aws:
### Description

Enhance t2ppl to add time parsing logics.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #10558, opened 2025-09-23
[Explore] update prompt. pass time information

angle943:
### Description

- adds `timefield` and `currentTime` to prompt Request as those were implemented here: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10257

## Changelog
- feat: add timefield and currentTime to prompt

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/utils/state_management/actions/query_editor/on_editor_run/call_agent/call_agent.ts
+++ b/src/plugins/explore/public/application/utils/state_management/actions/query_editor/on_editor_run/call_agent/call_agent.ts
@@ -64,7 +64,7 @@ export const callAgentActionCreator = createAsyncThunk<
       // TODO: when we introduce more query languages, this should be no longer be hardcoded to PPL
       language: 'PPL',
       dataSourceId: dataset.dataSource?.id,
-      currentTime: moment().toISOString(),
+      currentTime: moment().format('YYYY-MM-DD HH:mm:ss'),
       timeField: dataset.timeFieldName,
     };
 
@@ -76,7 +76,11 @@ export const callAgentActionCreator = createAsyncThunk<
     );
 
     if (response.timeRange) {
-      services.data.query.timefilter.timefilter.setTime(response.timeRange);
+      const convertedTimeRange = {
+        from: moment(response.timeRange.from, 'YYYY-MM-DD HH:mm:ss').toISOString(),
+        to: moment(response.timeRange.to, 'YYYY-MM-DD HH:mm:ss').toISOString(),
+      };
+      services.data.query.timefilter.timefilter.setTime(convertedTimeRange);
     }
 
     dispatch(runQueryActionCreator(services, response.query));
--- a/src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
+++ b/src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import moment from 'moment';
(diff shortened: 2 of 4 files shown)


src/plugins/explore/public/application/utils/state_management/actions/query_editor/on_editor_run/call_agent/call_agent.ts
```diff
@@ -63,6 +64,8 @@ export const callAgentActionCreator = createAsyncThunk<
       // TODO: when we introduce more query languages, this should be no longer be hardcoded to PPL
       language: 'PPL',
       dataSourceId: dataset.dataSource?.id,
+      currentTime: moment().toISOString(),
+      timeField: dataset.timeFieldName,
```
joshuali925, line 68: should we update the function used for classic discover as well
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a365de35b0c150377f36d51e65f6702e67cb54d9/src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx#L72-L108

Note on line 106 it already updates time range picker if present from agent response. This was written before agent supports time range, is the agent response contract the same or did it change?
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a365de35b0c150377f36d51e65f6702e67cb54d9/src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx#L106

angle943: updated. contract is same but based on: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10257, it doesn't send the timerange back in ISO format, so i updated it to convert it to ISO format

src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
```diff
@@ -74,6 +75,8 @@ export const QueryAssistBar: React.FC<QueryAssistInputProps> = (props) => {
       index: selectedIndex,
       language: props.dependencies.language,
       dataSourceId: selectedDataset?.dataSource?.id,
+      currentTime: moment().format('YYYY-MM-DD HH:mm:ss'),
```
joshuali925, line 78: why do we need currentTime from frontend instead of node server?

angle943: I've asked a similar question along with why timezone is omitted, here was the response:

```
Hi Justin. It's by designed to omit the timezone issue.
Let's say user is not in standard UTC timezone, and the user could ask question like "what's the error rate from 1pm to 2pm". And the time range depends on the user context (i.e. UTC or local zone selected at time picker)
What we do is omitting the time zone information, and the current_time has the same context with user's mind. Which means for LLM, 1pm to 2pm is just 1pm to 2pm in user's context. And LLM will also return a time string like this 2025-05-02 03:05:42 . If user are using UTC+N in time picker, then it's 2025-05-02 03:05:42 UTC+N.
```

so based on this, it looks like it uses the currentTime to get the user's relative time

TackAdam (approved):

src/plugins/query_enhancements/public/query_assist/components/query_assist_bar.tsx
joshuali925: hmm this doesn't appear to be very flexible, since OSD supports setting the timezone in advanced settings to something other than browser

angle943: @joshuali925 might be worth bringing it up with @zhichao-aws

Maosaic (approved):

PR #10562, opened 2025-09-24
[Explore] Add toggle to show full time range on time-based axes

lunalzm:
### Description

This PR adds a new toggle option "Show full time range" in the chart axis settings. When enabled, the visualization will render the full selected time range on the time axis, even if no data points exist at the start or end of the range. 

### Line chart
image
image

### Area chart
image
image

### Bar chart
image
image

## Testing the changes
1. Create a time-based chart.
2. In the Axes settings panel, enable the Show full time range toggle.
3. Verify that the chart axis expands to cover the entire selected time range.
4. Disable the toggle and confirm the chart axis adjusts to the data range only.

## Changelog

- feat: Add toggle to show full time range on time-based axes

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/visualizations/area/area_vis_config.ts
```diff
@@ -42,6 +42,7 @@ export interface AreaChartStyleOptions {
   titleOptions?: TitleOptions;
 
   thresholdOptions?: ThresholdOptions;
+  showFullTimeRange: boolean;
```
ruanyl: This should be optional

src/plugins/explore/public/components/visualizations/utils/utils.ts
```diff
@@ -62,7 +62,12 @@ export const applyAxisStyling = (
   }
 
   if (axis?.schema === VisFieldType.Date) {
-    fullAxisConfig.format = { seconds: '%I:%M:%S', milliseconds: '%I:%M:%S.%L' };
+    fullAxisConfig.format = {
+      hours: '%H:%M',
+      minutes: '%H:%M',
+      seconds: '%H:%M:%S',
+      milliseconds: '%H:%M:%S.%L',
```
ruanyl, line 77: Can you leave a comment to describe why `hours, minutes, seconds, milliseconds` are necessary?

src/plugins/explore/public/components/visualizations/utils/utils.ts
```diff
@@ -328,3 +333,67 @@ export const mergeStyles = (dest: ChartStyles, source: StyleOptions | undefined)
 
   return mergeWith(copiedDest, source, customMerge);
 };
+
+export function applyTimeRangeToEncoding(
+  mainLayerEncoding?: any,
+  axisColumnMappings?: AxisColumnMappings,
+  timeRange?: { from: string; to: string },
```
ruanyl, line 348: Perhaps this is unlike to happed, but what if the range of the data is out of the time range?

src/plugins/explore/public/components/visualizations/utils/utils.ts
lunalzm: Thanks for pointing that out.
Previously there was a suggestion to switch from 12-hour format to 24-hour format for better clarity.
That’s why hours, minutes, seconds, and milliseconds are now defined with the 24-hour format (`%H:%M` etc.) instead of the previous 12-hour (`%I:%M:%S`).

src/plugins/explore/public/components/visualizations/utils/utils.ts
```diff
@@ -328,3 +333,67 @@ export const mergeStyles = (dest: ChartStyles, source: StyleOptions | undefined)
...
+
+  if (mainLayerEncoding && mainLayerEncoding[targetRole]) {
+    mainLayerEncoding[targetRole].scale = {
+      ...(mainLayerEncoding[targetRole].scale || {}),
+      ...scaleConfig,
+    };
+    return null;
+  }
+
+  return {
```
wanglam: Nit: seems the results not been used.

src/plugins/explore/public/components/visualizations/utils/utils.ts
lunalzm: Original visualization:

image



I tried this locally. If the data range extends beyond the `scale.domain`, the chart can go blank.

image

To avoid the blank screen, we can add `"clamp": true` to the scale. This keeps points visible by clamping them to the domain edges even when the data range is larger.
```
"x": {
  "type": "temporal",
  "scale": {
    "domain": [ /* ... */ ],
    "clamp": true
  }
}
```
image

ruanyl (approved):

PR #10574, opened 2025-09-26
Date String Normalization

lunalzm:
### Description

Currently, time fields in the result set may appear as plain strings (e.g., "2025-09-24 14:50:00") without timezone information.
As a result, visualizations always display them in UTC, and the times do not adjust according to the system timezone settings (e.g., UTC+8).

### Before
image
image

### After
image

## Changelog

- feat: Date String Normalization

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/visualizations/utils/normalize_result_rows.ts
```diff
@@ -7,6 +7,14 @@ import { OpenSearchSearchHit } from '../../../types/doc_views_types';
 import { FIELD_TYPE_MAP } from '../constants';
 import { VisColumn, VisFieldType } from '../types';
 
+const normalizeDateString = (value: string): string => {
```
SuZhou-Joe: Should we add a UT to cover this function? And could you please add some comment to say that PPL will always use UTC as timezone so we need to append TZ string?

lunalzm: Good point! I'll add a small unit test to cover this and add a comment clarifying that.

ruanyl (approved):

ruanyl (dismissed):

SuZhou-Joe (approved):

angle943:
I think CiGroup16Explore has a failure due to the changes to this code. could we ensure all tests are passing?

TackAdam:
Testing locally think for the traces correlation object this change will require update https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10768 (This would need to be fixed in trace details)

Think it has custom fix for the time issue that would need to be adjusted.
Without change (Shows up)
Screenshot 2025-10-22 at 11 08 04 AM


After change (Missing)
Screenshot 2025-10-22 at 11 07 17 AM

lunalzm:
Hi @TackAdam , thanks for pointing that out! Could you clarify a bit more about how this change affects the trace correlation object or the trace details view?
Specifically, does the issue relate to how the time fix in #10768 modifies the trace timestamps, or how those timestamps are being used when correlating traces?
I just want to understand what part of the correlation logic or rendering is impacted so I can check it properly.

earthling-shruti:
Pulled the PR and tested with the latest cypress data. Related logs show up in the Discover traces and trace details page on redirect. We should be good on that front. 👍 Thanks, @TackAdam for adding an integ test to verify the behavior as well!

ruanyl:
@TackAdam would you be able to review and give it an approval?

PR #10582, opened 2025-09-26
Fix focus  add/remove column 

ruchidh:
### Description

 Changed the React key from field.name + index to just field.name since:
  - Field names are unique and stable identifiers
  - This ensures React correctly tracks each field component across re-renders
  - Focus management will now work properly when fields are added/removed
  

## Screenshot

https://github.com/user-attachments/assets/fdd47fed-6a1b-4c20-b3b9-f43843abc3ed

## Changelog
- fix:  Remove index from field key to fix column focus issue

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/fields_selector/field_list.tsx
+++ b/src/plugins/explore/public/components/fields_selector/field_list.tsx
@@ -50,11 +50,11 @@ export const FieldList = ({
         {title}
       </EuiButtonEmpty>
       {expanded &&
-        fields.map((field: DataViewField, index) => {
+        fields.map((field: DataViewField) => {
           return (
             <EuiPanel
               data-attr-field={field.name}
-              key={field.name + index}
+              key={field.name}
               paddingSize="none"
               hasBorder={false}
               hasShadow={false}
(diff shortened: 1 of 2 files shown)


src/plugins/explore/public/components/fields_selector/field_list.tsx
```diff
@@ -50,11 +50,11 @@ export const FieldList = ({
         {title}
       </EuiButtonEmpty>
       {expanded &&
-        fields.map((field: DataViewField, index) => {
+        fields.map((field: DataViewField) => {
           return (
             <EuiPanel
               data-attr-field={field.name}
-              key={field.name + index}
+              key={field.name}
```
angle943, line 57: why was previous implementation causing the focus bug to happen? Shouldn't `field.name + index` be unique as well?

ruchidh: You're right to question this.  I actually looked what viraj suggested in the task and found out he is right about index.
so,
  field.name + index would theoretically be unique, the problem lies in how React 
  handles list reordering and component state with keys.
 
  React sees fieldA went from key "fieldA0" to key "fieldA2" and treats it as a completely new component, breaking focus.
 
  so, the fix removes the index from the key to ensure React maintains component identity across re-renders, preserving focus state.
  
  and Yes, this change is better for performance as well as this avoids unnecessary DOM destruction/creation cycles.

angle943 (approved):

Issue #10490, opened 2025-09-05
[RFC] Enable **Workspaces**, New Discover with PPL, and PPL Visualization Framework by default in OpenSearch Dashboards


anirudha:
## Summary
- Make the **Workspaces** feature enabled by default.  
- Default the **New Discover app** with **Piped Processing Language (PPL)** as the primary query language for log analytics.  
- Introduce a **new visualization framework powered by PPL** and ship it enabled by default.  

Together, these defaults modernize the Dashboards experience: context-aware data separation (Workspaces), consistent analytics query language (PPL), and a next-gen visualization engine.

---

## Motivation
- **Workspaces**: Now stable with CRUD, ACL, per-workspace settings, and use-case presets. Default-on reduces friction and aligns with the new home page.  
- **New Discover**: PPL provides SQL-like syntax optimized for logs/observability. Users shouldn’t have to toggle between query languages — PPL should be the default.  
- **Visualizations**: Today, visualizations are tied to saved object plumbing. A PPL-native framework simplifies chart definitions, enables richer log-centric UIs, and aligns with future Observability use-cases.

---

## Goals
1. Flip `workspace.enabled` default to **true**.  
2. Default the home page to Workspace view.  
3. Make **PPL the default query language** in New Discover.  
4. Enable a **new PPL-powered visualization framework** by default, with a clear migration path.  
5. Provide migration helpers for Security multi-tenancy and legacy visualizations.

---

## User Experience / Defaults

```yaml
# Workspaces
workspace.enabled: true
uiSettings:
  overrides:
    "home:useNewHomePage": true

# Log Analytics / Discover
discover.defaultApp: "new"        # new Discover app
discover.defaultLanguage: "ppl"   # default query language = PPL

# Visualizations
visualizations.framework: "ppl"   # enable new framework by default

joshuali925:
related: [RFC] New Discover experience for Observability Workspaces https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9826

seraphjiang:
Do we have migration path for customer whose instance haven't enable workspace, if not, what is instruction to them.
e.g. 
they used to use default settings of osd (workspace disable by default).
after we make workspace default, customer need to specific disable workspace when they upgrade their osd instance.

ashwin-pc:
[...]

PR #10588, opened 2025-09-27
Make explore experience default

abbyhu2000:
### Description

auto turn on the following config when explore is turned on
```
data.savedQueriesNewUI.enabled: true
opensearchDashboards.branding:
  useExpandedHeader: false
uiSettings:
  overrides:
    "theme:version": "v9"
    "home:useNewHomePage": true
    "query:enhancements:enabled": true
```

## Changelog

- feat: explore experience will auto turn on few other flags

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


angle943 (approved):

src/plugins/explore/server/index.ts
```diff
@@ -18,4 +18,28 @@ export const config: PluginConfigDescriptor<ConfigSchema> = {
     enabled: true,
   },
   schema: configSchema,
+  deprecations: ({ rename, unused }) => [
```
LDrago27, line 21: @abbyhu2000 Are deprecations the right way to do it. it seems from prior usage as a way to just annotate settings that are unused. 

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/58f770e095e859007d66543936df6ce90799df55/src/plugins/newsfeed/server/index.ts#L42

abbyhu2000: yea it was a bit of the customized usage of deprecations. But after investigation, i realized to auto-enable these settings when explore flag is on is to either use deprecations or to modify the core system code during plugin discovery phase(src/core/server/plugins/discovery/plugins_discovery.ts). I feel like changing the core plugin discovery code might be against the architecture, thus i utilize the deprecations as a hack. @ashwin-pc

ashwin-pc: Lets just call out this hack and mark it as not a pattern we want to follow in the comments. Given our start to build 4.0, this is just a stop gap solution with minimal changes

ashwin-pc: I dont think this needs to block this PR

abbyhu2000: Will merge this and add the comment in the following PR.

ps48:
linked with https://github.com/opensearch-project/OpenSearch-Dashboards/issues/10490

PR #10606, opened 2025-09-30
[Explore vis] add zoom in interaction to explore vis

ruanyl:
### Description
https://github.com/user-attachments/assets/f5dde893-6c94-4fed-bc4f-ff96b74e0076

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/utils/time_range_brush.ts
+++ b/src/plugins/explore/public/components/visualizations/utils/time_range_brush.ts
@@ -1,0 +1,31 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { getColors } from '../theme/default_colors';
+
+export const createTimeRangeBrush = (options: { timeAxis: 'x' | 'y' }) => {
+  const colors = getColors();
+  return {
+    name: 'timeRangeBrush',
+    select: {
+      type: 'interval',
+      encodings: [options.timeAxis],
+      mark: { fill: colors.text, strokeOpacity: 0 },
+    },
+  };
+};
+
+export const createTimeRangeUpdater = (options: { dateField: string; timeAxis: 'x' | 'y' }) => {
+  return {
+    name: 'applyTimeFilter',
+    value: null,
+    on: [
+      {
+        events: 'pointerup',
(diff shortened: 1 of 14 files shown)


src/plugins/explore/public/embeddable/explore_embeddable.tsx
```diff
@@ -8,6 +8,7 @@ import { merge, Subscription } from 'rxjs';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { i18n } from '@osd/i18n';
+import moment from 'moment';
```
wanglam: Nit: Seems moment not been used in added code for this file.

src/plugins/explore/public/components/visualizations/visualization_container.tsx
```diff
@@ -53,6 +66,23 @@ export const VisualizationContainer = () => {
     };
   }, [visualizationBuilder]);
 
+  const onSelectTimeRange = useCallback(
+    (timeRange?: TimeRange) => {
+      if (timeRange) {
+        dispatch(
+          setDateRange({
+            from: moment(timeRange.from).toISOString(),
+            to: moment(timeRange.to).toISOString(),
```
Hailong-am, line 75: nit: do we need to check whether timeRange from and to is valid?

src/plugins/explore/public/components/visualizations/utils/time_range_brush.ts
```diff
@@ -0,0 +1,31 @@
...
+    name: 'applyTimeFilter',
+    value: null,
+    on: [
+      {
+        events: 'pointerup',
+        update: `timeRangeBrush ? opensearchDashboardsSelectTimeRange(timeRangeBrush) : null`,
+      },
+    ],
+  };
+};
```
Hailong-am, line 33: nit: maybe we can extract those magic strings into constants

src/plugins/explore/public/components/visualizations/visualization_container.tsx
ruanyl: time range picker should be able to gracefully handle invalid value, it set the time range to default `last 15 minutes`.

src/plugins/explore/public/components/visualizations/utils/time_range_brush.ts
ruanyl: updated

PR #10607, opened 2025-09-30
Add explore experience modal 

abbyhu2000:
### Description

* Display an explore experience modal when user land on homepage for the first time; skip display the modal for all subsequent times.
* Create a saved object config for recording if user has been showed this modal or not with a fall back logic of utilizing session storage for read only users.
* Add a yml config flag `#home.disableExperienceModal: false` to disable display of this.

## Screenshot

https://github.com/user-attachments/assets/daed61d1-62d2-48da-8374-42becb8a5920

## Changelog

- feat: Add explore experience modal 

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/home/public/application/components/explore_experience.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+interface ConfigNoticeModalProps {
+  onClose: () => void;
+}
+
+export const ExperienceSelectionModal = ({ onClose }: ConfigNoticeModalProps) => {
+  return (
+    <EuiModal onClose={onClose} maxWidth={800}>
+      <EuiModalHeader>
+        <EuiModalHeaderTitle>OpenSearch Dashboards Configuration Notice</EuiModalHeaderTitle>
```
ruanyl, line 30: Shall we wrap these texts with i18n?

src/plugins/home/public/application/components/home_app.js
```diff
@@ -90,6 +93,59 @@ export function HomeApp({ directories, solutions }) {
...
+        setIsCheckingExperience(false);
+      }
+    };
+
+    checkExperienceSelection();
+  }, [savedObjectsClient]);
+
+  const dismissExperienceSelection = async () => {
+    try {
+      await savedObjectsClient.create(
```
ruanyl: What if the user doesn't have saved objects write permission? Will the user always see the modal every time navigate to homepage?

angle943: +1. should we use localStorage?

abbyhu2000: i think using local storage will make this a per browser setting; but i think @ashwin-pc wants this to be a per instance setting. So they only see this when accessing the instance for the first time. What do u think? @ashwin-pc

ashwin-pc: If they don't have write access, they should not see this. They are by default not an admin

ashwin-pc: This is an admin and first time only screen

abbyhu2000: add a fallback logic to utilize session storage to set the dismissal state for read only users

src/plugins/home/public/application/components/explore_experience.tsx
abbyhu2000: will update UI components with i18n in the next PR after @kgcreative 's review

src/plugins/home/public/application/components/home_app.js
ashwin-pc: lets use local storage. Session storage is cleared after logout

ananzh:
Screenshot 2025-10-01 at 10 21 24 AM

tests are passed. only need to resolve one conflict. let's merge it without wait for another 2hr and block 3.3 release.

LDrago27 (approved):

src/plugins/home/public/application/components/home_app.js
abbyhu2000: will change to localstorage in fast follow up pr

ps48:
linked with https://github.com/opensearch-project/OpenSearch-Dashboards/issues/10490

PR #10639, opened 2025-10-03
chat(feat): implement plugin configuration controls for chat and context_provider plugins

ananzh:
### Description

We want to disable contextProvider and chat completely from browser and server side when no explore or when not config/ Need to use contextProvider as optionalPlugins not requiredBundles. Otherwise contextProvider will be loaded into browser. In this PR we change to use dependency injection for contextProvider. The challenge is that React hooks can't be passed through dependency injection in the traditional way because they need to be called at the component level and follow the rules of hooks. Therefore, we need to update the contextProvider plugin to 1) expose its hooks through the plugin's start interface, so they can be accessed via dependency injection 2) add a server side. Otherwise no way to read configuration from `opensearch_dashboards.yml` so plugin always loaded in browser regardless of settings. Also without server we could not use `optionalPlugins` pattern (requires server component). No change on how we use the hooks, only change how we import them. 

* Fixed chat.enabled to properly disable plugin when set to false or when chat.agUiUrl is missing 
* Added contextProvider.enabled with server-side validation and dependency injection pattern

#### Both server side and browser: 
No chat when explore is enabled but chat.enabled is false or no chat.agUiUrl. 
No contextProvider when explore is enabled but contextProvider.enable is false
Screenshot 2025-10-02 at 9 34 13 PM
Screenshot 2025-10-02 at 9 33 49 PM

Setup chat when explore is enabled and both chat.enabled and chat.agUiUrl are configed. 
Setup contextProvider when explore is enabled and contextProvider.enable is true
Screenshot 2025-10-02 at 9 33 39 PM

Screenshot 2025-10-02 at 9 30 34 PM

No chat or contextProvider when no explore
Screenshot 2025-10-02 at 9 34 56 PM

### Issues Resolved

NA

## Changelog
- feat: chat(feat): implement plugin configuration controls for chat and context_provider plugins

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/index.tsx
+++ b/src/plugins/explore/public/application/index.tsx
@@ -11,7 +11,10 @@ import { Store } from 'redux';
 import { AppMountParameters } from '../../../../core/public';
 import { ExploreServices } from '../types';
 import { LogsPage } from './pages/logs';
-import { OpenSearchDashboardsContextProvider } from '../../../opensearch_dashboards_react/public';
+import {
+  OpenSearchDashboardsContextProvider,
+  useOpenSearchDashboards,
+} from '../../../opensearch_dashboards_react/public';
 import { DatasetProvider } from './context';
 import { ExploreFlavor } from '../../common';
 import { TracesPage } from './pages/traces';
@@ -28,13 +31,14 @@ interface ExploreRouteProps {
 type ExploreComponentProps = ExploreRouteProps &
   Partial<Pick<AppMountParameters, 'setHeaderActionMenu'>>;
 
+const NOOP_PAGE_CONTEXT_HOOK = (options?: any): string => '';
+
 // Component that handles page context for all Explore flavors
 const ExplorePageContextProvider: React.FC<{
   children: React.ReactNode;
-  services: ExploreServices;
-}> = ({ children, services }) => {
-  const usePageContext = services.contextProvider?.hooks?.usePageContext || (() => {});
-
+}> = ({ children }) => {
+  const { services } = useOpenSearchDashboards<ExploreServices>();
+  const usePageContext = services.contextProvider?.hooks?.usePageContext || NOOP_PAGE_CONTEXT_HOOK;
   usePageContext({
     description: 'Explore application page context',
     convert: (urlState: any) => ({
@@ -92,7 +96,7 @@ export const renderApp = (
           <EditorContextProvider>
             <DatasetProvider>
               <services.core.i18n.Context>
-                <ExplorePageContextProvider services={services}>
+                <ExplorePageContextProvider>
                   <Switch>
                     {/* View route for saved searches */}
                     {/* TODO: Do we need this? We might not need to, please revisit */}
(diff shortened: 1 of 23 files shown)


src/plugins/explore/public/application/index.tsx
```diff
@@ -32,10 +31,13 @@ type ExploreComponentProps = ExploreRouteProps &
 // Component that handles page context for all Explore flavors
 const ExplorePageContextProvider: React.FC<{
   children: React.ReactNode;
-}> = ({ children }) => {
+  services: ExploreServices;
+}> = ({ children, services }) => {
+  const usePageContext = services.contextProvider?.hooks?.usePageContext || (() => {});
+
```
Maosaic: we sould only pass `usePageContext` into `ExplorePageContextProvider` to minimal rerender.

Also, we are not creating any context here, is this intended? Why we are calling it `ContextProvider`?

src/plugins/explore/public/components/data_table/explore_data_table.tsx
```diff
@@ -145,6 +145,7 @@ const ExploreDataTableComponent = () => {
             scrollToTop={scrollToTop}
             onAddColumn={onAddColumn}
             onRemoveColumn={onRemoveColumn}
+            services={services}
```
Maosaic: Why drilling `services` prop? can we use  `useOpenSearchDashboards<ExploreServices>()` in child component when we need services`?

Maosaic:
Should we split the changes that consuming context_provider plugin in Explore into a different PR?

src/plugins/explore/public/application/index.tsx
ananzh: I could pass usePageContext into that. But I don't think it would causes any re-render as it is only used once in the component and not passed to child components as props. Also the usePageContext hook itself handles the function internally.

ananzh: The ExplorePageContextProvider component is not creating a React Context Instead, it just uses the hook which can automatically capture page context from URL state. Basically it is to convert the URL state to the format needed. I could re-name it as ExplorePageContextRegister.

src/plugins/explore/public/components/data_table/explore_data_table.tsx
ananzh: +1 will update

Maosaic (dismissed):

abbyhu2000 (approved):

ananzh:
* [Build and Verify on Linux (ciGroup4) (pull_request)](https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/18270047796/job/52014509187?pr=10639) failures is a know issue and we have a fix https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10649

* The other two failed cypress tests [Run cypress tests / Run cypress tests (osd:ciGroup13) (pull_request)](https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/18270047803/job/52069085004?pr=10639) and  [Run cypress tests / Run cypress tests (osd:ciGroup15Explore) (pull_request)](https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/18270047803/job/52069085129?pr=10639) are also known issue https://github.com/opensearch-project/OpenSearch-Dashboards/issues/10648

ananzh:
ciGroup13 for query enhancements can pass locally 
 
Screenshot 2025-10-06 at 10 16 37 AM

PR #10660, opened 2025-10-07
Allow more use cases for serverless collections

amsiglan:
### Description

This PR has 2 sets of changes:

Exposes the permission control client and relevant utility that can be used together with the Workspace plugin server client to determine the permission mode id for the current user for a given workspace id. This will be leveraged in other plugins to gate user actions based on permission mode. For e.g. in security analytics we can block detector creation if user has read only permissions

On create workspace page, it limits the use case type to essentials if all the data sources connected to the application are OpenSearch Serverless Collections. This PR refactors the logic to allow plugins to specify the use case types they want to be supported for collections. By default the list include essentials use case type.

## Changelog
- feat: Allow more use cases for serverless collections

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/10660.yml
+++ b/changelogs/fragments/10660.yml
@@ -1,0 +1,2 @@
+feat:
+- Allow more use cases for serverless collections; Expose utilities for deriving workspace permissions ([#10660](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10660))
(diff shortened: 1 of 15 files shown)


SuZhou-Joe (commented):
Most of the change look good to me, but about the permission, we may want to migrate it to saved objects core and that may break the integration with other plugins. Can we do not expose it from workspace?

amsiglan:
> Most of the change look good to me, but about the permission, we may want to migrate it to saved objects core and that may break the integration with other plugins. Can we do not expose it from workspace?

~~Right now it should be okay exposing these from the workspace plugin since the permission check is based on workspace object.~~ Reverted the change to expose the permission related utilities

SuZhou-Joe (dismissed):

SuZhou-Joe (commented):
@amsiglan Could you please fix the CI?

amsiglan:
> @amsiglan Could you please fix the CI?

Pushed fix for the tests

ZilongX (approved):

amsiglan:
The failing CIs are not related to the changes in this PR

PR #10676, opened 2025-10-08
Trace:Details Replace Timeline gantt chart with updated hierarchy timeline table

yanatha99:
### Description

This addresses the TODO's from https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10642 except the "Service legends" change and improves the code organization for traces.

Refactor `SpanDetailTable` into separate `SpanListTable` and `SpanHierarchyTable` under new `span_detail_tables` directory
- `SpanListTable` - normal list view with filtering and sorting
- `SpanHierarchyTable` - foldout hierarchy view with timeline waterfall bars
- `Span`- shared component to render cells for the span tables
- `HierarchySpan` - encapsulates the drop-down hierarchy and spacing along with the span name, operation name, and error icon indicator
- Also moved `timeline_waterfall_bar` under new directory

Replace Timeline gantt chart with updated `SpanHierarchyTable`
- Replaced gantt chart in `SpanDetailPanel` "timeline" view with `SpanHierarchyTable`
- Removed "Tree view" from `SpanDetailPanel` and span tabs

## Screenshot

new-timeline-view

## Testing the changes

1. From Traces page go to any trace that contains at least one error
2. The default view loaded for trace details should be the "Timeline" tab, which should be the `SpanHierarchyTable` with Timeline waterfall bars shown in the example gif
3. Only spans that have errors should have an alert icon in the "Span" column
4. Adjust the "Span" column width by click dragging. When width is small should see that any truncated text has "...". No matter what the column width is, the alert icon should be present if the span has an error
5. Hover over any text in the "Span" column, should see tooltip with full text
6. If desired can also follow 4-5 from testing changes in https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10642 to test the Timeline column. Behavior here should be unchanged.

## Changelog
- feat: Replace Timeline gantt chart with updated hierarchy timeline table

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


yanatha99:
The force push before the linting fix commit was because forgot to sign off on that commit and had to reset to previous commit in order to push commit with sign off. Was unable to push when just amending commit with sign off as it was treated as commit with same hash.

yanatha99:
Was unable to test integ test changes locally, since my device to slow and timeout while setting up the test resources. Pushed to PR to test here and also for collaborators I'm working with to pull and test/modify locally.

yanatha99:
I cleaned up the commit history while also rebasing from main. Removed the integ test change commit and stashed it to instead unstash it on top of changes from mainline

yanatha99:
Did a force push since needed to rebase latest changes from main.

TackAdam:
Fast follow-up items
1. The fly-out version needs the spacing of columns adjusted to fit by default. (The width of the bar visualizations is to long for fly-out)
FlyoutSpacing
2. The Trace detail page should utilize the full space for the "Timeline panel" for cases where the span list exceeds the container. We are currently wasting a bunch of white space underneath the panel and introducing scroll-bar when unnecessary.

Screenshot 2025-10-10 at 10 31 48 AM

FullPageTimeLine

yanatha99:
Already have changes for both (1) and (2) fast follow items staged as part of change for moving the "Service legend" button. Waiting for this PR to get merged first before commit and push.

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_tables/span_hierarchy_table.tsx
```diff
@@ -0,0 +1,232 @@
...
+      setTotal(hierarchy.length);
+
+      // Auto-expand all spans by default to show the complete tree structure
+      const allSpanIds = gatherAllSpanIds(hierarchy);
+      setExpandedRows(allSpanIds);
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('Error parsing payloadData in SpanDetailTableHierarchy:', error);
+    } finally {
+      setIsSpansTableDataLoading(false);
```
Maosaic, line 89: It seems `isSpansTableDataLoading` never be true?

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_tables/span_list_table.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+      setItems(pageSpans);
+      setTotal(spans.length);
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('Error parsing payloadData in SpanDetailTable:', error);
+    } finally {
+      setIsSpansTableDataLoading(false);
+    }
+  }, [props.payloadData, props.DSL, props.filters, tableParams]);
```
Maosaic, line 136: Is props.DSL a intented dependency?

TackAdam:
Nit not blocking:
The span bar's are not centered with text / in their window

Screenshot 2025-10-10 at 11 34 49 AM

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_tables/span_list_table.tsx
```diff
@@ -0,0 +1,203 @@
...
+      const end = start + tableParams.size;
+      const pageSpans = spans.slice(start, end);
+
+      setItems(pageSpans);
+      setTotal(spans.length);
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('Error parsing payloadData in SpanDetailTable:', error);
+    } finally {
+      setIsSpansTableDataLoading(false);
```
Maosaic, line 134: It seems isSpansTableDataLoading never be true?

Maosaic:
I see a few file have `react-hooks/exhaustive-deps` disabled, can you add comment why you need to disabled it?

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_tables/span_hierarchy_table.tsx
yanatha99: I noticed that as well, but the SpanHierarchyTable and SpanListTable are my refactors and splitting of the [SpanDetailsTable](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_table.tsx).
This matches pre-existing behavior, and do not want to further increase the scope of this change to look into it.

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_tables/span_list_table.tsx
yanatha99: Referring back to earlier reply. I don't want to further increase the scope of my change to include fixes to the pre-existing logic from [SpanDetailsTable](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_table.tsx)

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_tables/span_list_table.tsx
yanatha99: Referring back to earlier reply. I don't want to further increase the scope of my change to include fixes to the pre-existing logic from [SpanDetailsTable](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_table.tsx)

yanatha99:
Referring back to earlier reply. I don't want to further increase the scope of my change to include fixes to the pre-existing logic from [SpanDetailsTable](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_table.tsx)

I only added the `react-hooks/exhaustive-deps` disabled comment in one place in my PR, since that useEffect was once again a direct port of a hook from the former SpanDetailsTable. My guess is that original file must've been made before the stricter linting checks, or there was some exception given to bypass.
Any other comments to disable that check were already there before my change.

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_tables/timeline_waterfall_bar/index.ts
TackAdam, line 1: Personally prefer the export from the component itself to be used as referencing it seems simpler. Don't see a benefit of having this inside the same folder as the original export. Would consider removing this in follow-up

yanatha99:
> Nit not blocking: The span bar's are not centered with text / in their window
> 
> Screenshot 2025-10-10 at 11 34 49 AM

This is intentional desgin from https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10642 in order to fit the bar within the ruler scale provided by the column header. 
If the bar took up the whole width, then the column header scale would be inaccurate, or if the scale was stretched the beginning and end would be cutoff by the column width.

PR #10690, opened 2025-10-08
Explore: update dataset configurator

Maosaic:
### Description

Introduce a V2 dataset configurator that allow saving dataset when dataset creation.

Enable V2 dataset configurator in Explore

## Screenshot

https://github.com/user-attachments/assets/708e5d79-068d-4c1c-a3a5-d84b29209418

## Changelog

- feat: Allow saving dataset from dataset configurator

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/dataset_select/dataset_select.tsx
+++ b/src/plugins/data/public/ui/dataset_select/dataset_select.tsx
@@ -44,13 +44,13 @@ export interface DetailedDataset extends Dataset {
 export interface DatasetSelectProps {
   onSelect: (dataset: Dataset) => void;
   supportedTypes?: string[];
-  singalType: string | null;
+  signalType: string | null;
 }
 
 /**
  * @experimental This component is experimental and may change in future versions
  */
-const DatasetSelect: React.FC<DatasetSelectProps> = ({ onSelect, supportedTypes, singalType }) => {
+const DatasetSelect: React.FC<DatasetSelectProps> = ({ onSelect, supportedTypes, signalType }) => {
   const { services } = useOpenSearchDashboards<IDataPluginServices>();
   const isMounted = useRef(true);
   const [isOpen, setIsOpen] = useState(false);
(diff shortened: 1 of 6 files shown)


src/plugins/data/public/ui/dataset_select/dataset_select.tsx
```diff
@@ -43,20 +43,14 @@ export interface DetailedDataset extends Dataset {
 
 export interface DatasetSelectProps {
   onSelect: (dataset: Dataset) => void;
-  appName: string;
   supportedTypes?: string[];
-  onFilter?: (dataset: Dataset) => boolean;
+  singalType: string | null;
```
angle943: signalType

src/plugins/data/public/ui/dataset_selector/configurator/configurator_v2.test.tsx
```diff
@@ -0,0 +1,188 @@
...
+  },
+};
+
+const mockOnConfirm = jest.fn();
+const mockOnCancel = jest.fn();
+const mockOnPrevious = jest.fn();
+
+beforeEach(() => {
+  jest.clearAllMocks();
+  // @ts-expect-error TS2345 TODO(ts-error): fixme
```
angle943: for tests can we just use `any`? i think we are tracking `@ts-expect-error` to be fixed in the future, but we don't really need to fix it for tests

src/plugins/data/public/ui/dataset_select/dataset_select.tsx
```diff
@@ -43,20 +43,14 @@ export interface DetailedDataset extends Dataset {
 
 export interface DatasetSelectProps {
   onSelect: (dataset: Dataset) => void;
-  appName: string;
```
joshuali925, line 46: why are we removing appName here? appName is used to populate context
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/42e95956deef035003d65c689daee3b1f3bf7ae4/src/plugins/data/public/ui/dataset_select/create_dataset_select.tsx#L22-L33

which will then be used for filtering out unsupported datasets
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/42e95956deef035003d65c689daee3b1f3bf7ae4/src/plugins/data/public/ui/dataset_selector/advanced_selector.tsx#L71-L71

PR #10692, opened 2025-10-09
[Explore] Restrict patterns tab to only default datasets

sejli:
### Description

Patterns currently only works for index pattern and index datasets. This change restricts the UI to reflect that.

## Screenshot

image

image

## Changelog

- fix: Restrict patterns tab to default datasets

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/10692.yml
+++ b/changelogs/fragments/10692.yml
@@ -1,0 +1,2 @@
+fix:
+- Restrict patterns tab to default datasets ([#10692](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10692))


src/plugins/explore/public/components/tabs/tabs.tsx
```diff
@@ -65,7 +67,18 @@ export const ExploreTabs = () => {
 
   // Display tabs that registered under current flavor
   const tabs: EuiTabbedContentTab[] = registryTabs
-    .filter((registryTab) => registryTab.flavor.includes(flavorId))
+    .filter((registryTab) => {
+      const registeredFlavor = registryTab.flavor.includes(flavorId);
```
paulstn, line 71: In the future, might be good to expose some functionality like this to register_tabs to let tabs specify the behavior in which they remain hidden, but this works for now.

PR #10695, opened 2025-10-09
global semantic search: PLM

s-zx:
### Description

To get the semantic information of user input and existing search results, we use pre-trained models (PLM) to generate corresponding embeddings(vectors). By comparing the distance between each embedding, we can find comparably relevant results.

Specifically, we use transformer.js to run pre-trained model in our browser easily. They provide  pipeline  API, which simplifies complex tasks by abstracting away the intricacies of tokenization (converting raw text into numerical IDs the model understands), preprocessing (adding special tokens, padding, and truncation), model inference (executing the ONNX model via ONNX Runtime), and post-processing (extracting the relevant output, such as pooling the embeddings). Further more, it provides methods like  cos_sim  to compute the cosine similarity between two embeddings.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/public/chrome/utils.test.ts
+++ b/src/core/public/chrome/utils.test.ts
@@ -196,83 +196,83 @@ describe('searchNavigationLinks', () => {
 
   const allAvailableCaseId = ['test-group'];
 
-  it('should return matching visible and enabled links', () => {
+  it('should return matching visible and enabled links', async () => {
     const query = 'child';
-    const result = searchNavigationLinks(allAvailableCaseId, navGroupMap, query);
+    const result = await searchNavigationLinks(allAvailableCaseId, navGroupMap, query);
 
     expect(result).toHaveLength(1);
-    // expect(result[0]).toEqual(
-    //   expect.objectContaining({
-    //     id: 'child',
-    //     title: 'Child Link',
-    //     navGroup: mockedNavGroup,
-    //   })
-    // );
+    expect(result[0]).toEqual(
+      expect.objectContaining({
+        id: 'child',
+        title: 'Child Link',
+        navGroup: mockedNavGroup,
+      })
+    );
   });
 
-  it('should return child links when searching by parent title', () => {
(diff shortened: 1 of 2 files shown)


src/core/public/chrome/ui/header/header_search_bar.tsx
```diff
@@ -196,11 +197,15 @@ export const HeaderSearchBar = ({ globalSearchCommands, panel, onSearchResultCli
     [globalSearchCommands, onSearchResultClick]
   );
 
+  const debouncedSearch = useMemo(() => {
+    return debounce(search, 500); // 300ms delay, adjust as needed
```
wanglam, line 201: Nit: Will 500ms has better UI performance than 300ms?

src/core/public/chrome/chrome_service.tsx
```diff
@@ -124,6 +124,7 @@ interface ConstructorParams {
 
 export interface SetupDeps {
   uiSettings: IUiSettingsClient;
+  http: HttpStart;
```
wanglam: Can we use `HttpSetup` here?

src/core/public/chrome/utils.test.ts
wanglam, line 1: We'd better to fix the failed unit tests in this file.

src/plugins/workspace/server/routes/index.ts
```diff
@@ -343,6 +347,104 @@ export function registerRoutes({
...
+          const loadingTimeMs = endTime - startTime;
+
+          console.log('Model loaded and ready for inference.');
+          console.log(`Model loading took: ${loadingTimeMs.toFixed(2)} ms`);
+        } else {
+          console.log('Model already loaded. Reusing existing pipeline.');
+        }
+
+        // Generate embeddings for links
+        const linkEmbeddings = await Promise.all(
```
wanglam, line 389: How long will it take to generate embeddings for links? Maybe we can store these info in the memory.

PR #10696, opened 2025-10-09
global semantic search: neural sparse search

s-zx:
### Description

Neural Sparse Search offers a lightweight yet effective approach for semantic search by representing text as sparse vectors where most elements are zero. This method bridges the gap between traditional keyword matching and dense neural embeddings.

Neural Sparse Search works in two phases:

1. Document Processing: Documents are tokenized and converted into sparse vector representations where only meaningful tokens have non-zero values.
2. Query Processing: User queries undergo the same tokenization process, creating sparse vectors that can be efficiently compared with document vectors.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/core/public/chrome/ui/header/header_search_bar.tsx
```diff
@@ -196,11 +197,15 @@ export const HeaderSearchBar = ({ globalSearchCommands, panel, onSearchResultCli
     [globalSearchCommands, onSearchResultClick]
   );
 
+  const debouncedSearch = useMemo(() => {
+    return debounce(search, 500); // 300ms delay, adjust as needed
```
FriedhelmWS, line 201: Nit: comment need changes

src/plugins/workspace/server/sparse_search.ts
```diff
@@ -0,0 +1,97 @@
...
+    score: number;
+  }> {
+    const nlpBertTokenizer = new BertWordPieceTokenizer({ vocabContent: Object.keys(this.vocab) });
+    const tokenizedResult = nlpBertTokenizer.tokenizeSentence(query);
+    const tokensArray = tokenizedResult.tokens;
+    console.log('Tokenization: ', tokensArray);
+
+    const queryVec = this.buildQueryVector(tokensArray);
+    console.log('Non-zero query dimensions count: ', Object.keys(queryVec).length);
+    console.log('Non-zero query vector: ', queryVec);
```
yuye-aws, line 81: neural sparse search supports natural language query. You can search with existing model_id

src/core/public/chrome/utils.ts
```diff
@@ -255,27 +255,49 @@ export function searchNavigationLinks(
...
+        return {
+          ...link,
+          parentNavLinkTitle,
+          navGroup,
+        };
+      });
   });
+
+  try {
+    console.log('All links:', allSearchAbleLinks);
```
FriedhelmWS, line 277: Nit: `AllSearchableLinks`.

src/plugins/workspace/server/sparse_search.ts
yuye-aws: Maybe you can provide some context on why you're doing tokenization here? At first glance, you're doing tokenization to obtain the sparse query vector with IDF value. This is the doc-only search mode: https://docs.opensearch.org/latest/query-dsl/specialized/neural-sparse/

src/plugins/discover/public/plugin.ts
```diff
@@ -266,6 +266,7 @@ export class DiscoverPlugin
     core.application.register({
       id: PLUGIN_ID,
       title: 'Discover',
+      description: 'Analyze your data in OpenSearch and visualize key metrics.',
```
FriedhelmWS, line 269: Just for curiosity, would the performance of semantic search be more accurate if we can have a more explanatory and detailed description for each application?

src/plugins/workspace/server/routes/index.ts
```diff
@@ -113,6 +116,13 @@ const updateWorkspaceAttributesSchema = schema.object({
   ...workspaceOptionalAttributesSchema,
 });
 
+let jsonData: {
```
FriedhelmWS, line 119: Is it safe to make it as a global variable in nodejs? Would it be nice to create a class, make `jsonData` a private field of that class and provide dedicated method to manipulate the `jsonData`?

src/plugins/workspace/server/routes/index.ts
```diff
@@ -343,6 +353,51 @@ export function registerRoutes({
...
+        }),
+      },
+    },
+    async (context, req, res) => {
+      try {
+        const { query } = req.body;
+
+        if (!jsonData) {
+          const filePath = path.join(__dirname, 'doc_vectors.json');
+          const data = await readFile(filePath, 'utf8');
```
FriedhelmWS, line 378: Would be nice to have error handling for file reading here I guess.

src/plugins/workspace/server/sparse_search.ts
s-zx: Yes, you are right. We want to use doc-only neural sparse search to achieve semantic search in frontend without using backend service. We generate doc vetor in advance and store them in frontend side. Then we tokenize the query to obtain the sparse query vector with IDF value. After that, relevance is calculated using a dot product between query and document vectors.

src/core/public/chrome/utils.ts
s-zx: Thanks!

src/plugins/discover/public/plugin.ts
s-zx: Yes, it will imrpove the relevance. We can enrich them in the future.

src/plugins/workspace/server/routes/index.ts
s-zx: Yes, this is a good concern. We should make it private.

src/plugins/workspace/server/sparse_search.ts
yuye-aws: Doc-only queries supports natural language query

s-zx: I see, this way we don't need tokenization anymore.

PR #10698, opened 2025-10-09
Trace Details: Remove service.name column

jessrlew:
### Description

Remove duplicate service name column from traces table. `attributes.service.name` is not needed since we already have 'resource.attributes.service.name'. 

## Screenshot
Screenshot 2025-10-09 at 9 23 46 AM

## Changelog
- fix: remove service.name column from traces table

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/fields_selector/lib/group_fields.ts
+++ b/src/plugins/explore/public/components/fields_selector/lib/group_fields.ts
@@ -34,7 +34,6 @@ import { FieldFilterState, isFieldFiltered } from './field_filter';
 // TODO: Use data set defined faceted field
 const FACET_FIELDS = [
   'resource.attributes.service.name',
-  'serviceName',
   'attributes.http.status_code',
   'status.code',
 ] as const;
--- a/src/plugins/explore/server/trace_ui_settings.ts
+++ b/src/plugins/explore/server/trace_ui_settings.ts
@@ -19,7 +19,6 @@ export const traceUiSettings: Record<string, UiSettingsParams> = {
       'status.code',
       'attributes.http.status_code',
       'resource.attributes.service.name',
-      'serviceName',
       'kind',
       'name',
       'durationNano',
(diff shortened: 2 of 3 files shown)


src/plugins/explore/public/components/fields_selector/lib/group_fields.ts
```diff
@@ -33,7 +33,6 @@ import { FieldFilterState, isFieldFiltered } from './field_filter';
 
 // TODO: Use data set defined faceted field
 const FACET_FIELDS = [
-  'attributes.service.name',
   'resource.attributes.service.name',
   'serviceName',
```
TackAdam: Think we should also remove serviceName, as the dataprepper schema will show duplicate
After

src/plugins/explore/server/trace_ui_settings.ts
```diff
@@ -18,7 +18,6 @@ export const traceUiSettings: Record<string, UiSettingsParams> = {
       'spanId',
       'status.code',
       'attributes.http.status_code',
-      'attributes.service.name',
       'resource.attributes.service.name',
       'serviceName',
```
TackAdam: Same, think we should remove 'serviceName', if both schema's have 'resource.attributes.service.name',

src/plugins/explore/public/helpers/data_table_helper.tsx
```diff
@@ -84,10 +84,8 @@ function getColumnDisplayName(column: string): string {
...
     case 'resource.attributes.service.name':
       return 'Service';
-    case 'attributes.service.name':
-      return 'Service.Name';
     case 'attributes.http.status_code':
-      return 'Status code';
+      return 'Status Code';
     case 'status.code':
       return 'Status';
     case 'spanId':
```
angle943, line 91: should we be translating all these?

jessrlew: Good point! Are the column headers usually translated?

angle943: yeah, every text that we show to end users should be wrapped in i18n of some sort. but no biggie, it can be done in a followup

angle943: oops i spoke too soon. didn't think thoroughly. table headers we don't usually translate because they are dynamic. the ones for traces are static but i think its fine not to translate. translating would be weird if we had other non-translated column headers

PR #10703, opened 2025-10-09
Redesigned logs tab component with accordion for dataset and expandable rows

kakshak07:
### Description

Redesigned logs tab component with accordion and expandable rows. Each row has message with code block for clear display.

### Issues Resolved

Made the display for logs component better

## Screenshot

image

image

image

## Testing the changes

Added unit testing for the new component and tested it properly. Also ran the test locally.

## Changelog

- feat: Added dataset with accordion
- feat: Made logs row expandable
- feat: Display 10 logs
- feat: Added testing for the componenets

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`
- [x]  New functionality includes testing.

### DIFF
--- a/src/plugins/explore/public/application/pages/traces/trace_details/public/logs/dataset_logs_table.tsx
+++ b/src/plugins/explore/public/application/pages/traces/trace_details/public/logs/dataset_logs_table.tsx
@@ -106,6 +106,10 @@ export const DatasetLogsTable: React.FC<DatasetLogsTableProps> = ({
         }),
         truncateText: true,
         sortable: false,
+        width: '300px',
+        css: {
+          minWidth: '200px',
+        },
       },
       {
         field: 'spanId',


TackAdam:
What is the purpose of the row being expandable if only the message is displayed. I see it is copyable but seems unnecessary or like more information should be present?
Screenshot 2025-10-10 at 1 33 08 PM

TackAdam:
Update
Last change added truncation so the expand drop down now makes sense

TackAdam:
Not blocking, think we should have discussion on only showing 10 logs (current behavior). It encourages user's to use the full experience (Discover: Logs) but can be mis-leading without a call-out on the page indicating it is possibly a subset.

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
```diff
@@ -69,6 +69,8 @@ export interface TraceDetailsProps {
   isEmbedded?: boolean;
   isFlyout?: boolean;
 }
+// Displaying only 10 logs in the tab
+export const LOGS_DATA = 10;
```
TackAdam, line 73: Lets follow-up discussion on this / adding pagination back

kamingleung:
> Not blocking, think we should have discussion on only showing 10 logs (current behavior). It encourages user's to use the full experience (Discover: Logs) but can be mis-leading without a call-out on the page indicating it is possibly a subset.

+1, we should provide additional context: I would suggest we rename the accordion header to this:
> From dataset [dataset_name]
> 10 recent results
image
@TackAdam @kakshak07

PR #10706, opened 2025-10-10
[Explore] fix explore storybook by bringing over mocks from discover to explore

angle943:
### Description

- `yarn storybook explore` was not working for some time, as the build would fail with lots of errors
- issue was that we were importing `indexPatternMock` from the discover plugin like this:

```
  import { indexPatternMock } from '../../../../discover/public';
```

- This was breaking storybook because this import would bring in the entire discover plugin with its plugin initialization code and dependencies, which were failing to resolve properly.

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/__mock__/index.test.mock.ts
+++ b/src/plugins/explore/public/__mock__/index.test.mock.ts
@@ -1,0 +1,31 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { DataSource } from 'src/plugins/data/public';
+
+interface DataSourceConfig {
+  name: string;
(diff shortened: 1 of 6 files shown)


src/plugins/explore/public/__mock__/index.test.mock.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  async testConnection(): Promise<boolean> {
+    return true;
+  }
+
+  async runQuery(queryParams: any) {
+    return { data: {} };
+  }
+}
```
angle943, line 31: this file and index-pattern_mock.ts are just copied and pasted from the discover plugin

PR #10714, opened 2025-10-13
Introduce schema mapping

Maosaic:
### Description

Introducing schema mapping, allowing user to map data field to certain attribute that UI requires. Adding OTel schema mappings as the first tenant of this feature. 

Schema mappings will be stored with dataset following the format below
```
{
  "schemaMappings": {
    "<schemaName>": {
      "<attributeName>": "<fieldName>"            
      }
    }
}
```

UI define schema attribute in `src/plugins/data/public/ui/dataset_selector/configurator/schema_config.ts`

--- 
Refactor the `ConfiguratorV2` component, service methods now called once at component initialization instead of multiple times throughout render, extract component base on single, clear responsibility.

  Changes

  Extracted Components:
  - SaveAsDatasetOption - Handles "Save as dataset" checkbox with disabled state warning
  - DatasetMetadataFields - Manages dataset name and description inputs with async warning

  Custom Hook:
  - useDatasetFields - Encapsulates field fetching logic, preventing potential infinite loops from useEffect dependencies

## Screenshot

https://github.com/user-attachments/assets/4e306060-cbc0-49be-8b30-ecdd35f56855

## Changelog

-feat: Introducing schema mapping

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #10721, opened 2025-10-14
Fix indexPattern link on Asset table

Maosaic:
### Description

Fix a problem causing indexPattern link on Asset table failed to load when dataset management plugin is enabled.

Saved object server side compute URL of `indexPattern` type as `/indexPatterns/patterns/*` which does not match with dataset management plugin route.
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/server/saved_objects/index_patterns.ts#L45-L56

This PR update dataset management plugin to register a redirection from `/indexPatterns` route to `/datasets`

This PR also added unit test to cover some components introduced by https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10714
 
## Changelog

- fix: Fix indexPattern link on Asset table

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/dataset_management/public/plugin.test.ts
+++ b/src/plugins/dataset_management/public/plugin.test.ts
@@ -25,7 +25,7 @@ describe('DiscoverPlugin', () => {
         management: managementPluginMock.createSetupContract(),
       })
     ).not.toThrow();
-    expect(setupMock.application.register).toBeCalledTimes(1);
+    expect(setupMock.application.register).toBeCalledTimes(2);
     waitFor(() => {
       expect(setupMock.chrome.navGroup.addNavLinksToGroup).toBeCalledTimes(1);
     });


TackAdam:
Working for Trace/Log indexs


https://github.com/user-attachments/assets/7052cd3f-f173-4460-87ac-e72601d03e30

src/plugins/dataset_management/public/plugin.ts
```diff
@@ -97,6 +97,32 @@ export class DatasetManagementPlugin
       return pathInApp && `/patterns${pathInApp}`;
     });
 
+    // Forward indexPatterns management page to datasets management page
+    urlForwarding.forwardApp('indexPatterns', DM_APP_ID, (path) => {
```
paulstn, line 101: Would we be removing the indexPatterns management page code at some point, if its being deprecated? Is it being deprecated?

paulstn (approved):

src/plugins/dataset_management/public/plugin.ts
Maosaic: Eventaully yes, for now they are coexist

Issue #10614, opened 2025-10-01
[RFC] Field Statistics


paulstn:
# Field Statistics in Explore

## Background

The Field Statistics feature is a new addition to OpenSearchDashboards' Explore plugin that provides users with comprehensive statistical information about fields in their indices. This feature addresses the need for users to quickly understand the composition and distribution of their data without writing complex queries.

The feature will be implemented as a new tab in the Explore interface, similar to the existing Logs, Patterns, and Visualization tabs. It will display a table showing field names, document counts, distinct value counts, and additional type-specific statistics in an expandable row format.

### Component Structure

```
src/plugins/explore/public/components/
├── tabs/
│   └── field_stats_tab.tsx              # Main tab component
└── field_stats/
    ├── field_stats_container.tsx        # Data fetching and state management
    ├── field_stats_table.tsx            # Table component with column definitions
    ├── field_stats_row_details.tsx      # Expandable row details
    ├── field_stats_queries.ts           # PPL query generation functions
    ├── field_stats_types.ts             # TypeScript interfaces and types
    └── detail_sections/
        ├── top_values_section.tsx       # Top values detail component
        ├── numeric_summary_section.tsx  # Numeric statistics component
        └── date_range_section.tsx       # Date range detail component
```

### Tab Registration

The Field Stats tab will be registered in `register_tabs.ts`:

```typescript
tabRegistry.registerTab({
  id: EXPLORE_FIELD_STATS_TAB_ID,
  label: 'Field Stats',
  flavor: [ExploreFlavor.Logs, ExploreFlavor.Metrics],
  order: 25, // After Visualization tab
  supportedLanguages: [EXPLORE_DEFAULT_LANGUAGE],
  component: FieldStatsTab,
});
```

### Field Statistics Query Strategy

When the tab is opened, we don't need a query to find all of the fields. Instead, we'll use the existing Redux selector `selectDataset` to get the current index pattern and then use `getIndexPatternFieldList` to get all fields.

#### Getting Fields Without Query

```typescript
const dataset = useSelector(selectDataset);
const fields = useMemo(() => {
  return getIndexPatternFieldList(dataset, {});
}, [dataset]);
```

#### Basic Field Statistics Query

For each field, we'll fetch the following information:

```ppl
source = <index>
| stats count(`<field>`) as count,
        dc(`<field>`) as dc,
        count() as total_count
| eval percentage_total = (count * 100.0) / total_count
```

[...]

PR #10723, opened 2025-10-14
[Explore] Field Statistics Tab

paulstn:
### Description

This PR introduces a new **Field Statistics** tab to the Explore plugin in OpenSearch Dashboards. The feature provides users with comprehensive statistical analysis of fields in their indices, enabling quick data profiling and exploration without writing complex queries.

The Field Stats tab displays a sortable, expandable table showing field-level statistics including document counts, distinct value counts, and type-specific detailed analytics such as top values, numeric summaries, date ranges, and example values.

Related RFC: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/10614

### Field Statistics Tab

- New tab available in the Explore interface for both Logs and Metrics flavors
- Positioned after the Visualization tab
- Supports PPL as the default query language

#### 1. **Field Statistics Table**

A comprehensive, sortable table displaying all non-meta fields with:

- Field name and type (with icon)
- Document count and percentage
- Distinct value count (approximate)
- Expandable rows for detailed field-specific statistics

image

#### 2. **Expandable Row Details**

Dynamic detail sections that adapt to field type:

**String/Keyword Fields:**

- Top 10 most frequent values with counts and percentages

image

**Numeric Fields:**

- Top 10 values
- Summary statistics (min, median, average, max)

image

**Date Fields:**

- Date range (earliest and latest timestamps)

image

**Boolean Fields:**

- Top values distribution

**Complex Fields (geo_point, geo_shape, binary, object):**

- Example values from the first 10 documents

image

#### 3. **Loading States**

- Centered loading spinner with "Searching in progress..." message during initial data fetch
- Per-row loading indicators when expanding fields for detailed statistics

image

### Component Structure

```
src/plugins/explore/public/components/
├── tabs/
│   └── field_stats_tab.tsx              # Main tab component
└── field_stats/
    ├── field_stats_container.tsx        # Data fetching and state management
    ├── field_stats_table.tsx            # Table component with sorting
    ├── field_stats_table_columns.tsx    # Column definitions
    ├── field_stats_row_details.tsx      # Expandable row details renderer
    ├── field_stats_queries.ts           # PPL query generation and execution
    ├── field_stats_detail_sections.ts   # Registry of detail sections
    ├── detail_sections/
    │   ├── top_values_detail.tsx        # Top values component and config
    │   ├── numeric_summary_detail.tsx   # Numeric statistics component and config
    │   ├── date_range_detail.tsx        # Date range component and config
    │   └── examples_detail.tsx          # Examples component and config
    └── utils/
        ├── field_stats_types.ts         # TypeScript interfaces
        ├── field_stats_utils.ts         # Utility functions
        ├── field_stats.stubs.ts         # Test stubs
        └── constants.ts                 # Constants and configurations
```

#### 1. **Plugin Architecture for Detail Sections**

The feature uses a registry-based architecture that makes it easy to add new detail sections:

```typescript
// detail_sections/histogram_detail.tsx
export const histogramDetailConfig: DetailSectionConfig<HistogramData> = {
  id: 'histogram',
  title: 'Distribution',
  applicableToTypes: ['number'],
  fetchData: async (fieldName, dataset, services) => {
    // Fetch histogram data
  },
  component: HistogramSection,
};

// utils.ts
export const DETAIL_SECTIONS: DetailSectionConfig[] = [
  topValuesDetailConfig,
  numericSummaryDetailConfig,
  dateRangeDetailConfig,
  examplesDetailConfig,
  // new configs
];
```

Each detail section is self-contained with:

- Its own PPL query function
- React component for rendering
- Configuration specifying applicable field types
- Data fetching logic

#### 2. **Lazy Loading with Parallel Queries**

- Basic field statistics are fetched in parallel for all fields on tab load
- Detailed statistics are only fetched when a row is expanded (lazy loading)
- All applicable detail sections for a field are fetched in parallel when expanded

#### 3. **Field Filtering**

Automatically filter out:

- Meta fields (e.g., `_id`, `_index`, `_source`)
- Multi-fields (e.g., `.keyword` subfields)
- Scripted fields

### State Management

The `FieldStatsContainer` manages multiple state pieces:

- `fieldStats`: Basic statistics for all fields (Record<string, FieldStatsItem>)
- `isLoading`: Global loading state for initial fetch
- `expandedRows`: Set of currently expanded field names
- `fieldDetails`: Detailed statistics for expanded fields (FieldDetailsMap)
- `detailsLoading`: Set of fields currently loading details

### Accessing the Feature

1. Navigate to the Explore plugin in OpenSearch Dashboards
2. Select a data source and index
3. Click on the **Field Stats** tab
4. View field statistics in the sortable table
5. Click the expand button to see detailed statistics for any field

### Use Cases

**Data Profiling:**

- Quickly understand field distributions without writing queries
- Identify fields with high cardinality
- Find fields with missing data (low document counts)

**Data Quality Analysis:**

- Check for unexpected values in string fields
- Verify numeric ranges are within expected bounds
- Identify date ranges and time spans

**Query Optimization:**

- Find fields with few distinct values (good candidates for term aggregations)
- Identify high-cardinality fields that may slow queries
- Understand data distribution before creating visualizations

## Follow ups

- Pagination for the table, make it as simple as just a simple pagination, still need to wait for all top level queries
- Action Bar, have it show the field count
- Bug: Top values contains null for documents that this field isn’t a part of, making the percentage go crazy (>100). E.g. phpmemory in sample_logs
- Hide for other dataset types

## Changelog

- feat: Explore Field Statistics

### DIFF
--- a/src/plugins/explore/public/components/field_stats/field_stats_container.test.tsx
+++ b/src/plugins/explore/public/components/field_stats/field_stats_container.test.tsx
@@ -5,6 +5,22 @@
 
 import React from 'react';
 import { render, waitFor } from '@testing-library/react';
+
+// Mock the data plugin before any imports that use it
+jest.mock('../../../../data/public', () => ({
+  indexPatterns: {
+    isDefault: jest.fn(() => true),
+  },
+}));
+
+jest.mock('../../application/context/dataset_context/dataset_context');
+jest.mock('../../../../opensearch_dashboards_react/public');
+jest.mock('./utils/field_stats_utils');
+jest.mock('./field_stats_queries');
+jest.mock('./field_stats_table', () => ({
+  FieldStatsTable: jest.fn(() => <div data-test-subj="field-stats-table">Mocked Table</div>),
+}));
+
 import { FieldStatsContainer } from './field_stats_container';
 import { useDatasetContext } from '../../application/context/dataset_context/dataset_context';
 import { useOpenSearchDashboards } from '../../../../opensearch_dashboards_react/public';
@@ -24,14 +40,6 @@ import {
   createMockDataset,
   createMockFieldStatsItem,
 } from './utils/field_stats.stubs';
-
-jest.mock('../../application/context/dataset_context/dataset_context');
-jest.mock('../../../../opensearch_dashboards_react/public');
-jest.mock('./utils/field_stats_utils');
-jest.mock('./field_stats_queries');
-jest.mock('./field_stats_table', () => ({
-  FieldStatsTable: jest.fn(() => <div data-test-subj="field-stats-table">Mocked Table</div>),
-}));
 
 const mockUseDatasetContext = useDatasetContext as jest.MockedFunction<typeof useDatasetContext>;
 const mockUseOpenSearchDashboards = useOpenSearchDashboards as jest.MockedFunction<
--- a/src/plugins/explore/public/components/field_stats/field_stats_queries.test.ts
+++ b/src/plugins/explore/public/components/field_stats/field_stats_queries.test.ts
@@ -41,21 +41,28 @@ describe('field_stats_queries', () => {
 
   describe('executeFieldStatsQuery', () => {
     const mockServices = createMockServices();
-    const mockSearchSource = {
-      setFields: jest.fn(),
-      fetch: jest.fn(),
-    };
 
     beforeEach(() => {
       jest.clearAllMocks();
-      (mockServices.data.search.searchSource.create as jest.Mock).mockResolvedValue(
-        mockSearchSource
-      );
+      (mockServices.data.search.searchSource.create as jest.Mock).mockResolvedValue({
+        setFields: jest.fn(),
+        setField: jest.fn(),
+        setParent: jest.fn(),
+        fetch: jest.fn().mockResolvedValue({}),
+      });
     });
 
     it('executes query successfully', async () => {
       const mockResults = { hits: { total: 100 } };
-      mockSearchSource.fetch.mockResolvedValue(mockResults);
+      const mockSearchSource = {
+        setFields: jest.fn(),
+        setField: jest.fn(),
+        setParent: jest.fn(),
+        fetch: jest.fn().mockResolvedValue(mockResults),
+      };
+      (mockServices.data.search.searchSource.create as jest.Mock).mockResolvedValue(
+        mockSearchSource
+      );
 
       const result = await executeFieldStatsQuery(
         mockServices,
@@ -69,7 +76,15 @@ describe('field_stats_queries', () => {
     });
 
     it('sets correct searchSource fields', async () => {
-      mockSearchSource.fetch.mockResolvedValue({});
+      const mockSearchSource = {
+        setFields: jest.fn(),
+        setField: jest.fn(),
+        setParent: jest.fn(),
+        fetch: jest.fn().mockResolvedValue({}),
+      };
+      (mockServices.data.search.searchSource.create as jest.Mock).mockResolvedValue(
+        mockSearchSource
+      );
 
       await executeFieldStatsQuery(mockServices, 'source = test | stats count()', 'test-id');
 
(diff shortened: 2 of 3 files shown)


TackAdam:
Investigating vrs 3.2 back-end issues found to address in follow-up
1. Empty fields in the table due to missing dependency's needed for count/percentage. Might want to show ? indicator saying why these fields are empty but that they are still sorted correctly.
3 2EmptyFields
2. Body field got the following error
3 2BodyError
3. And tested with very large dataset got the following circuit breaker exception
```
CircuitBreakingException[[script] Too many dynamic script compilations within, max: [75/5m]; please use indexed, or scripts with parameters instead; this limit can be changed by the [script.context.filter.max_compilations_rate] setting];\n\nFor more details, please send request for Json format to see the raw response from OpenSearch engine.",
```

src/plugins/explore/public/components/field_stats/field_stats_container.tsx
```diff
@@ -0,0 +1,157 @@
...
+            return undefined;
+          }
+        })();
+
+        // Step 2: fetch field stats per row
+        const results = await Promise.allSettled(
+          fields.map(async (field) => {
+            try {
+              const query = getFieldStatsQuery(dataset.title, field.name);
+              const result = await executeFieldStatsQuery(
```
sejli, line 77: Any way for user to limit the max number or cancel these? Feel like there's going to be a lot of queries that are executed. I'm not familiar with the queries that are run, but is it possible to pass all the fields in a single query and have the engine return them all?

src/plugins/explore/public/components/field_stats/field_stats_container.tsx
```diff
@@ -0,0 +1,157 @@
...
+                  errorMessage,
+                },
+              };
+            }
+          })
+        );
+
+        const allFieldStats: Record<string, FieldStatsItem> = {};
+        results.forEach((result) => {
+          if (result.status === 'fulfilled' && result.value) {
```
sejli, line 107: Maybe we can leverage the existing [query status](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/explore/public/application/utils/state_management/slices/query_editor/query_editor_slice.ts#L57)?

src/plugins/explore/public/components/field_stats/field_stats_detail_sections.ts
```diff
@@ -0,0 +1,36 @@
...
+import { DetailSectionConfig } from './utils/field_stats_types';
+import { topValuesDetailConfig } from './detail_sections/top_values_detail';
+import { numericSummaryDetailConfig } from './detail_sections/numeric_summary_detail';
+import { dateRangeDetailConfig } from './detail_sections/date_range_detail';
+import { examplesDetailConfig } from './detail_sections/examples_detail';
+
+/**
+ * Registry of all detail sections that can be displayed in expanded field rows.
+ *
+ * ## How to Add a New Detail Section
```
sejli, line 15: Could add some more context for what a detail section is for

sejli (approved):
LGTM. Only main concern I have would be a performance hit if there are a bunch of queries that get lined up from multiple fields.

Others are just nits. I think the file structure can also be improved upon. There's a bunch of components for field stats that can be organized into directories. 

Cypress tests should be a fast follow up.

src/plugins/explore/public/components/field_stats/field_stats_container.tsx
paulstn: I do not believe it would be possible to pass all the fields in a single query and have the engine return them all. As for cancellation, that isn't supported yet within explore, but when that happens, it would be a good enhancement. I think letting queries fail should also be enough since they should time out.

src/plugins/explore/public/components/field_stats/field_stats_container.tsx
paulstn: There would be a lot of field statistics queries, and setting individual query status requires a cacheKey, which would be too much state to put into redux. Cacheing field stats results within `.kibana` or anywhere else could be a requirement we look at later, due to its complexity.

TackAdam (approved):

paulstn:
@TackAdam For performance enhancements, I'm potentially going to use `stats count(eval(isnotnull(<field>)))` to chain together aggregations for multiple fields within the same query, and batch queries that way. Currently vetting this command, since it seems to be a >3.3 opensearch command.

sejli (approved):

PR #10724, opened 2025-10-14
Explore:Traces Empty state

TackAdam:
### Description
Explore:Traces Empty state
Add error handling for missing fields by informing user.
Disabled the span-id re-direction if it would lead to a broken page.

## Screenshot
SpanId disabled:
Hover

Fly-out with info:
Flyout

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


Maosaic (approved):

src/plugins/explore/public/application/pages/traces/trace_details/trace_view.tsx
```diff
@@ -405,169 +437,170 @@ export const TraceDetails: React.FC<TraceDetailsProps> = ({
...
+                  handleErrorFilterClick={handleErrorFilterClick}
+                  logDatasets={logDatasets}
+                  logsData={logsData}
+                  isLogsLoading={isLogsLoading}
+                />
+              </EuiPanel>
+            </div>
+
+            {/* Filter badges section */}
+            {spanFilters.length > 0 && (
```
paulstn, line 466: Nit: doesn't have to be handled now, but personally, I think a good way of cutting down large UI components and making sure that they are a lot more readable is to separate out React UI components by what they would be semantically, just so that I could look at something and go "oh those are the headers, that is that thing, etc". Even with functions that are put into props, putting those in functions defined up above or in util files makes this a lot more readable.

This might not make sense in all cases so take this with a grain of salt.

ps48 (approved):

src/plugins/explore/public/utils/trace_field_validation.ts
```diff
@@ -0,0 +1,214 @@
...
+): {
+  isValid: boolean;
+  missingFields: string[];
+  presentFields: string[];
+} => {
+  const missingFields: string[] = [];
+  const presentFields: string[] = [];
+
+  const spanId = extractFieldFromRowData(rowData, SPAN_ID_FIELD_PATHS);
+  if (spanId) {
```
paulstn, line 119: I've noticed that there is a repetitive pattern here with extract -> check -> push. We could try and make this a bit more maintainable and readable. 

(Completely optional suggestion):
We have an array of each of these fields we are checking for - this array could contain the name, the path we expect to find it at (to handle nested values), anything else needed.
Then, we have this function work to perform validation after on that array. 

A potential downside of doing my suggestion is that each field we check for is abstract enough that all of the various options inside of the array that we put down could grow to be larger than just doing this manually (which could be why you've done it this way). I haven't fully checked but take a look and see if this could help.

PR #10725, opened 2025-10-15
Trace:Details Bug fixes and visual improvements

yanatha99:
### Description

Multiple bug fixes and visual improvements for Trace Details page
- Fix vertical alignment of cell content in `SpanHierarchyTable` to all be centered
- Fix nested field filtering with shared `applySpanFilters` util
- Replace hard-coded `SpanHierarchyTable` default height of `500px` with `auto` when there are fewer entries and capped at `70vh` when there are more entries
- Fix duplicate ppl queries on page load or filter change
- Fix double re-render on filter change

## Screenshot

Trace Details page before fixes
trace-details-before-fixes

Trace Details page after fixes
trace-details-after-fixes

## Testing the changes

1. Go to details page of any trace
3. In the "Timeline" tab validate that all the contents of each row in table are vertically centered and align with each other (none of them should be lower or higher than other content in the same row)
4. Go to the "Meta data" tab of the right drawer and apply any filter from there
5. Validate that the expected spans still show in table
6. Validate that the page does not seem to "flicker" and re-render multiple times
7. Open chrome (or whatever web browser) dev tools and look at network calls
8. Validate that on page refresh only one ppl query is made
9. Validate that when adding or removing filters only one ppl query is made

## Changelog
- fix: trace details bug fixes and visual improvements

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_tables/span_hierarchy_table.test.tsx
+++ b/src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_tables/span_hierarchy_table.test.tsx
@@ -168,15 +168,6 @@ describe('SpanHierarchyTable', () => {
     consoleSpy.mockRestore();
   });
 
-  it('shows all hierarchy columns by default', () => {
-    const { getByTestId } = render(<SpanHierarchyTable {...defaultProps} />);
-
-    const visibleColumns = JSON.parse(getByTestId('visible-columns').textContent || '[]');
-    expect(visibleColumns).toContain('span');
-    expect(visibleColumns).toContain('timeline');
-    expect(visibleColumns).toContain('durationInNanos');
-  });
-
   it('passes availableWidth to RenderCustomDataGrid', () => {
     const propsWithWidth = {
       ...defaultProps,
@@ -190,18 +181,6 @@ describe('SpanHierarchyTable', () => {
         availableWidth: 800,
       })
     );
-  });
-
-  it('includes expected hierarchy columns', () => {
-    render(<SpanHierarchyTable {...defaultProps} />);
-
-    const mockCall = mockRenderCustomDataGrid.mock.calls[0]?.[0];
-    expect(mockCall).toBeDefined();
-    const columnIds = mockCall!.columns.map((col: any) => col.id);
-
-    expect(columnIds).toContain('span');
-    expect(columnIds).toContain('timeline');
-    expect(columnIds).toContain('durationInNanos');
   });
 
   it('renders HierarchySpanCell for span column', () => {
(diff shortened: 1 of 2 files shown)


TackAdam:
Overall it looks good, this is a great improvement reducing duplicate calls.
One blocking concern though:
The removed hidden column functionality on the Timeline. The user is able to switch tabs to get them back or refreshing but we don't want to cause confusion for first time experience if this is accidentally clicked.

https://github.com/user-attachments/assets/0ffabbb7-9c1d-40b0-841a-39f270a48ebc

The old behavior allowed the user to add it back
OldBehavior

yanatha99:
> Overall it looks good, this is a great improvement reducing duplicate calls. One blocking concern though: The removed hidden column functionality on the Timeline. The user is able to switch tabs to get them back or refreshing but we don't want to cause confusion for first time experience if this is accidentally clicked.
> 
>  HideColumnIssue.mov 
> The old behavior allowed the user to add it back OldBehavior

Thanks for the callout! Removed column actions from the `SpanHierarchyTable` all together now with `
fix: disable column actions for SpanHierarchyTable`

yanatha99:
`fix: Stop column widths resizing when selecting spans` is from [TackAdam](https://github.com/TackAdam) added it to this PR as well

TackAdam (approved):
Thanks for implementing requested changes

yanatha99:
Rebased changes from main and also fixed unit test to account for this change `
fix: disable column actions for SpanHierarchyTable`. No other diffs between this change set and previously approved change set.

TackAdam (dismissed):

ruchidh (dismissed):

yanatha99:
Rebased from main and fixed merge conflicts with https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10728. 
No other diffs between this change set and previously approved change set.

Did find a bug when testing locally, not sure which commit from the rebase introduced it. Dismissing a filter still correctly sets the spanFilters state, but the table items don't update accordingly. 
Can see from the gif in PR description that this was working after my changes, but one of the rebased commits breaks it.
Would rather just get this PR in now, so it doesn't keep churning and I have to keep fixing bugs introduced by other changes.

TackAdam:
> Rebased from main and fixed merge conflicts with #10728. No other diffs between this change set and previously approved change set.
> 
> Did find a bug when testing locally, not sure which commit from the rebase introduced it. Dismissing a filter still correctly sets the spanFilters state, but the table items don't update accordingly. Can see from the gif in PR description that this was working after my changes, but one of the rebased commits breaks it. Would rather just get this PR in now, so it doesn't keep churning and I have to keep fixing bugs introduced by other changes.

This can be fixed with adding the following to trace_view line 544
```
key={`span-panel-${visualizationKey}-${spanFilters.length}-${transformedHits.length}`}
```
This will cause the proper re-render when filter changes.
This can be a fast follow today/tomorrow.

TackAdam (approved):

PR #10727, opened 2025-10-15
Query cancel for new discover

ruchidh:
### Description
 Summary

  This PR implements a comprehensive cancel button feature for query execution across OpenSearch Dashboards, allowing users to cancel running
   queries in both the standard QueryEditor and custom QueryExecutionButton scenarios. 
  The implementation provides a unified cancel button
  experience that works consistently across different data source types (OpenSearch and S3).

  What Changed

  Core Features

  - Cancel Button Visibility: Added showCancelButton, onQueryCancel, and isQueryRunning props to enable cancel functionality
  - Timeout-based UI: Cancel button appears after 50ms delay to avoid flickering for quick queries
  - Cross-component Integration: Works seamlessly with both QueryEditor and custom submit buttons
  - Data Source Compatibility: Functions correctly for both OpenSearch and S3 data source types

## Screenshot

https://github.com/user-attachments/assets/d6c59ac2-5eb6-47c9-b4e7-473d2819113a

Screenshot 2025-10-15 at 3 52 36 PM

## Changelog
- feat: Add cancel query feature to discover

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/search_bar/search_bar.test.tsx
+++ b/src/plugins/data/public/ui/search_bar/search_bar.test.tsx
@@ -127,9 +127,8 @@ function wrapSearchBarInContext(testProps: any) {
       ...startMock.uiSettings,
       get: jest.fn((key) => {
         if (key === 'query:enhancements:enabled') return true;
-        if (key === 'timepicker:quickRanges') return [
-          { from: 'now-15m', to: 'now', display: 'Last 15 minutes' }
-        ];
+        if (key === 'timepicker:quickRanges')
+          return [{ from: 'now-15m', to: 'now', display: 'Last 15 minutes' }];
         return startMock.uiSettings.get(key);
       }),
     },


sejli:
There are some tests that are failing due to the changes, some related to snapshots. Could you fix the tests and run the linter? Typically, we have a precommit validation that runs that should run the linter on changed files.

config/opensearch_dashboards.yml
```diff
@@ -415,4 +415,4 @@
 
 # @experimental Set the value to true to enable context provider
 # contextProvider:
-#   enabled: true
\ No newline at end of file
+#   enabled: true
```
TackAdam, line 418: Was this included by mistake?

src/plugins/data/public/ui/query_editor/_query_editor.scss
```diff
@@ -86,6 +86,13 @@
   max-width: 32px;
 }
 
+// Global cancel button styling - applies everywhere
+.osdQueryEditor__cancelButton {
+  border: 1px solid #BD271E !important;
```
TackAdam: Believe the hardcoded colors will be lint blocker, need oui color usage.

config/opensearch_dashboards.yml
ruchidh: not sure why it's showing up, though i didn't make any changes, tried to reset but still shows up

TackAdam:
Can you rename PR tittle?

src/plugins/data/public/ui/query_editor/_query_editor.scss
```diff
@@ -86,6 +86,13 @@
...
+  border-radius: 4px !important;
```
TackAdam: Is the border-radius required? I don't see a difference without it and ideally we want to avoid hard coding padding. If it is required we should use EuiSizeXs.

config/opensearch_dashboards.yml
angle943: you can do:

```
git restore --staged config/opensearch_dashboatrds.yml
```

before commiting.

And better yet, something we learned from @ashwin-pc , if you have temporary`opensearch_dashboards.yml` changes you need to make, create a `opensearch_dashboards.dev.yml` within this folder and have your personal changes in there. Then you never have to interact with the `opensearch_dashboards.yml` file. This is what mine looks like:

Screenshot 2025-10-21 at 11 47 17 AM

angle943: and i believe its showing up because you prob made changes to this file for your local configs, and your IDE is auto-formatting it.

TBH its more proper to add the extra whitespace like your change does, but ideally in the future its best to leave this file out of the diff to make it easier for reviewers.

You can use vim to undo this change if you really wanted and commit it, but not a blocker

ruchidh: thanks justin that would help

src/plugins/data/public/ui/query_editor/_query_editor.scss
```diff
@@ -86,6 +86,13 @@
...
+  background-color: transparent !important;
+}
```
angle943, line 94: +1 to what adam said, but even better i don't think any of these styles are required. Could you try removing these styles, removing the classnames from the two cancel buttons, and adding the prop `display="base"`?

src/plugins/explore/public/components/top_nav/query_execution_button/query_execution_button.tsx
```diff
@@ -31,13 +35,67 @@ export interface QueryExecutionButtonProps {
...
+    return () => {
+      if (cancelButtonTimerRef.current) {
+        clearTimeout(cancelButtonTimerRef.current);
+      }
+      if (minimumDisplayTimerRef.current) {
+        clearTimeout(minimumDisplayTimerRef.current);
+      }
+    };
+  }, [reduxShouldShowCancelButton, shouldShowCancelButton]);
+
```
angle943, line 54: it looks like this useEffect and the other one in search bar is very similar. is it possible to encapsulate these logic in a hook and have the two components call that hook?

src/plugins/data/public/ui/query_editor/_query_editor.scss
ruchidh: This would require to set border, without it the default icon is not taking the required css as per mock design .

Adam, can you try again, may be after refresh. 
Also, have updated the hardcoded color

ruchidh: Image

TackAdam: Got it, ya my cache kept boarder when i removed. Definitely want the border included.

src/plugins/explore/public/components/top_nav/query_execution_button/query_execution_button.tsx
ruchidh: Sure justin, i already have encapsulated that in next pr of dashboard cancel, but if you say i would implement here too.

ruchidh:
> when i first load the page (ie refresh), i don't see the cancel button. Could we update the logic to have the cancel there too?

in case of initial load, I deliberately put it on timefilter update, not on load
Because the initial load is pretty faster and the cancel button ui was going away pretty fast , kind of looking flicker
So I have put logic only when the api delays, then it should appear, because for fast load, there is hardly any action user can take because the button goes away too quickly.

I would implement for initial load too, if this looks blocker to you. Let me know

PR #10729, opened 2025-10-15
Add correlations tab to support linking Trace and Logs Datasets 

ps48:
### Description

Add support for correlations tab for Datasets feature. 
* Users can create/edit a correlation from traces datasets, under the correlated datasets tab
* Users can view a associated correlations from the logs dataset tab
* Users can edit schema mappings during the correlations creation or later when editing a correlation
* Limits -> 1 trace can only be part of 1 correlation and each correlation can have max 5 logs datasets

### Issues Resolved

#10386 
 
## Screenshot / Testing the changes

https://github.com/user-attachments/assets/96c5b728-c02e-4c3e-8dd4-d4893969a030

## Changelog

 - feat: Add correlations tab to support linking Trace and Logs Datasets 
 
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/dataset_management/public/components/edit_dataset/tabs/correlated_datasets_tab/configure_correlation_modal/field_mapping_editor.tsx
+++ b/src/plugins/dataset_management/public/components/edit_dataset/tabs/correlated_datasets_tab/configure_correlation_modal/field_mapping_editor.tsx
@@ -306,9 +306,31 @@ export const FieldMappingEditor: React.FC<FieldMappingEditorProps> = ({
   );
 
   // Handler to start editing a dataset
-  const handleStartEdit = useCallback((datasetId: string) => {
-    setEditingDatasetId(datasetId);
-  }, []);
+  const handleStartEdit = useCallback(
+    (datasetId: string) => {
+      // If there's already a dataset being edited, revert its changes first
+      if (editingDatasetId !== null && editingDatasetId !== datasetId) {
+        // Revert previous dataset to original mappings
+        setDatasetFieldMappings((prev) =>
+          prev.map((ds) =>
+            ds.datasetId === editingDatasetId
+              ? { ...ds, mappings: { ...originalMappings[editingDatasetId] } }
+              : ds
+          )
+        );
+
+        // Clear invalid fields for previous dataset
+        setInvalidFields((prev) => {
+          const { [editingDatasetId]: _, ...rest } = prev;
+          return rest;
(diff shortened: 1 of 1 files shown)


TackAdam:
Testing this change against the Traces workflow. Required update to the log fetching as the correlation is a different object structure in src/plugins/explore/public/application/pages/traces/trace_details/public/logs/correlation_service.ts

Works great, nice change!

https://github.com/user-attachments/assets/cb276f37-4e45-4c15-804f-4b53f93d8d6c

TackAdam:
NIT: The Traces "Correlated datasets" does not update with a (1) to indicate one exist after creation. 
The logs does show this so it might mis-lead user's if the behavior is different here
UIFeedback

Logs:
Screenshot 2025-10-15 at 3 07 38 PM

ps48:
> NIT: The Traces "Correlated datasets" does not update with a (1) to indicate one exist after creation. The logs does show this so it might mis-lead user's if the behavior is different here UIFeedback
> 
> Logs: Screenshot 2025-10-15 at 3 07 38 PM

This came from Ming to not have a counter for Correlated datasets and 1 trace can only be mapped to 1 correlation object

src/plugins/dataset_management/public/components/edit_dataset/tabs/correlated_datasets_tab/configure_correlation_modal/field_mapping_editor.tsx
```diff
@@ -0,0 +1,734 @@
...
+                color="success"
+                iconType="check"
+                size="s"
+              />
+              <EuiSpacer size="m" />
+            </>
+          )}
+
+        {/* Show error callout only when there are missing mappings */}
+        {hasErrors && missingMappings.length > 0 && (
```
earthling-shruti, line 713: Should we show a generic unknown error in other cases where hasErrors is true?

ps48: If `hasErrors = true`, then at least one row found a matching entry in missing mappings with non-empty missingFields, which means `missingMappings.length > 0` must be `true`.  `hasErrors` is kept as a separate variable for react state update race conditions. 

TL;DR generic unknown error cannot be caught with hasErrors it just checks for missing mappings.

PR #10743, opened 2025-10-17
Pass 5 rows as context if no row is expanded

goyamegh:
### Description
1. When passing context to agent, if a few rows are passed as context, it helps in improving the accuracy of generating PPL queries because the attributes and other fields are available for reference. I'm only passing 5 rows for now, and if there are no expanded rows present.
2. Improve the system prompt by adding details about PPL queries.

### Issues Resolved

NA

## Screenshot

Screenshot 2025-10-17 at 1 25 22 AM

## Testing the changes

Tested against Langgraph agent.

## Changelog

- feat: pass top 5 rows as default context and improve system prompt for ppl accuracy

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/packages/osd-agents/src/agents/langgraph/react_agent.ts
+++ b/packages/osd-agents/src/agents/langgraph/react_agent.ts
@@ -40,6 +40,9 @@ export interface ReactAgentState {
   threadId?: string; // Store thread identifier
   runId?: string; // Store run identifier
   modelId?: string; // Store model identifier from forwardedProps for dynamic model selection
+  // Duplicate detection
+  duplicateCallCount?: number; // Counter for consecutive duplicate tool calls
+  recentToolSignatures?: string[]; // Signatures of recent tool calls for duplicate detection
 }
 
 /**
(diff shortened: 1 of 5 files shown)


ananzh:
I don;t think we should pass default rows. It might confuse future dynamic expanded rows. For example, if I expand several rows after explore loaded, how could I separate default rows and user expanded rows? If we need fields to improve explore context, we could modify page context to pass fields and fields mapping.

goyamegh:
> I don;t think we should pass default rows. It might confuse future dynamic expanded rows. For example, if I expand several rows after explore loaded, how could I separate default rows and user expanded rows? If we need fields to improve explore context, we could modify page context to pass fields and fields mapping.

@ananzh thanks for the review. If you notice, these are only passed when no row is expanded. Further, I am also exploring how to modify the system prompt to use field mappings directly by the LLM. I would suggest to merge this for now, as this helps in improving the accuracy of the agent dramatically. 

Another area I'm exploring is how to manage this context effectively. Few ideas are showing the context usage in the evals, logging them for easier debugging, etc. I don't think it'll be a single shot answer, because we need to be mindful of the context window here. Let me know what you think.

PR #10739, opened 2025-10-16
fix: Add Copy Icon for TraceId and relocate for Span overview

mattishii26:
### Description

- Traces detail page > Span details. Move copy icon to the right of the value
- Trace detail page > Add copy icon to Trace ID badge

### Issues Resolved

- Traces detail page > Span details. Move copy icon to the right of the value
- Trace detail page > Add copy icon to Trace ID badge

## Screenshot

image

image

image

## Testing the changes

Manual testing and unit tests

## Changelog
- fix: Add copy icon for Trace ID
- fix: Relocate copy icon for Span overview to be on left side

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #10738, opened 2025-10-16
made tab persistent

kakshak07:
### Description

Made tab persistent across different span selection and default fallback to Overview in case of tab not found

### Issues Resolved

Made tab persistent across different span selection and default fallback to Overview in case of tab not found, even if you switch between span and timeline persistent of tabs is saved

## Screenshot

https://github.com/user-attachments/assets/13e8ae04-d0ed-4528-b9dd-585b2d152902

## Testing the changes

Added testing for the same

## Changelog

- feat: Made tab persistent across different span selection

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [x] Update [CHANGELOG.md](./../CHANGELOG.md)
- [x] Commits are signed per the DCO using --signoff

PR #10731, opened 2025-10-15
fix: Trace view to have narrower resizer and tabs panel to be transparent

mattishii26:
### Description
- fix: Trace view padding to match Log/Trace view
- fix: Tabs panel to not have a background and border

### Issues Resolved

- Trace detail page, reduce panel spacing
- Trace detail > Remove panel border around tabs

## Screenshot

Screenshot 2025-10-15 at 3 04 49 PM

image

## Changelog
- fix: Trace view padding to match Log/Trace view
- fix: Tabs panel to not have a background and border

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #10742, opened 2025-10-17
Trace:Details Enhance span status filter functionality

yanatha99:
### Description

Replace singular "Error filter" with new Span status filter that can filter for : Error, OK, or Unset statuses.
- Created `SpanStatusFilter` new component with popover containing selectable status filters. If there are multiple status filters selected ANDs them together, which follows established "Meta data" filtering behavior for trace details 
- Updated `applySpanFilters` util to handle all span statuses, not just Error. Uses the same logic as the span detail flyout for determining Ok and Error. Just uses the `span.status.code` directly for determining Unset, so there can be overlap between Unset and Error or Unset and Ok, but that is expected.
- Removed "Error filter" component and related logic from `TraceDetailTabs`
- Added `SpanStatusFilter` to `SpanHierarchyTable` secondary toolbar

### Issues Resolved

- one liner fix from @TackAdam for https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10725#issuecomment-3412436551
- removed extra line at top of `SpanDetailPanel`

## Screenshot

Old behavior with "Error filter"
span-error-filter

New behavior with added functionality from Span status filter
span-status-filter

## Testing the changes

1. Go to details page of trace with at least one error span
2. Click the "Filter by status" button to open popover with filter options
3. Select the "Unset" status filter
4. For each span still shown in table check its "Raw span" in the "Span details" panel and validate that `"status": { "code": "UNSET" }` is present
5. Unselect the filter and select the "Error" status filter in the popover
6. Validate that the only spans shown have the error alert in "Span" column
7. Unselect the filter and select the "Ok" status filter in the popover
8. Validate that the only spans shown do not have error alert in "Span" column
9. Try mixing and matching the status filters and verify the expected spans show e.g. if "Error" and "Ok" filters selected there should be no spans, as there is no overlap between those two statuses
10. Try mixing the status filters with the "Meta data" filters from "Span details" panel and verify the expected spans show
11. Try removing the status filters by unselecting option in the "Filter by status" popover, dismissing the filter badge, and clearing all filters.
12. While changing the status filters make sure the count next to "Filter by status" accurately represents the number of selected status filters

## Changelog

- feat: Add span status filters to trace details

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
[...]

PR #10745, opened 2025-10-17
Trace:Details Add span status filters. Combine bug fixes.

yanatha99:
### Description

This PR combines the following PRs to resolve merge conflicts and proactively handle any potential interaction bugs between the changes by testing them all together
- feat improvement https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10742 @yanatha99 
- bug fix https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10738 @kakshak07 
- visual fix https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10739 @mattishii26 
- visual fix https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10731 @mattishii26
Can view those PR's for more detailed description of each change

Additional commits not from those PRs
- `fix: change span overview service identifier to operation` @yanatha99 - Span Overview tab service identifier now correctly uses the span operation instead of service name
- `chore: Improve SpanStatusFilter unit testing` @yanatha99 

## Screenshot

Can view the recordings/screenshots from the listed PRs to see each of the changes separately.

How Trace details looks with all these changes together.
trace-details-after-changes

## Testing the changes

Follow the testing steps in the linked PRs.

To test `fix: change span overview service identifier to operation`
1. Go to Discover Traces page 
2. Note the "Service identifier" of any span in the table
3. Click the "Time" link of that span to open the trace detail flyout
4. In the flyout look at the Span details Overview tab and verify the "Service identifier" there matches (2)

## Changelog
- feat: Add span status filters to trace details. Combine bug fixes.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


yanatha99:
After discussion with @TackAdam going to fast follow this to change span status filter to only allow single selection at a time.

TackAdam:
Callout as fast fallow:
Switch the filter by spans to be single select instead of and clause.
Screenshot 2025-10-17 at 10 49 37 AM

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/ppl_resolve_helpers.tsx
```diff
@@ -282,6 +282,13 @@ export function isSpanError(span: any): boolean {
   return false;
 }
 
+export function isSpanOk(span: any): boolean {
```
Maosaic, line 285: "OK" is vague, I recommend call it `isSpanNotErrored`

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_tables/span_hierarchy_table.tsx
```diff
@@ -175,6 +176,10 @@ export const SpanHierarchyTable: React.FC<SpanTableProps> = (props) => {
   ];
 
   const secondaryToolbar = [
+    SpanStatusFilter({
+      spanFilters: props.filters,
+      setSpanFiltersWithStorage: props.setSpanFiltersWithStorage || (() => {}),
+    }),
```
Maosaic, line 182: `SpanStatusFilter` is a function component calling it directly might break it

src/plugins/explore/public/application/pages/traces/trace_details/public/top_nav_buttons.tsx
```diff
@@ -79,10 +81,24 @@ export const TraceTopNavMenu: React.FC<TraceTopNavMenuProps> = ({
...
+                        })}
+                      </EuiText>
+                    </EuiFlexItem>
+                    <EuiFlexItem grow={false}>
+                      <EuiIcon
+                        type="copy"
+                        size="s"
+                        style={{ marginLeft: '4px' }}
+                        aria-label="Copy Trace ID"
+                      />
```
Maosaic, line 99: Any function need to be added on the copy icon?

src/plugins/explore/public/application/pages/traces/trace_details/public/top_nav_buttons.tsx
```diff
@@ -79,10 +81,24 @@ export const TraceTopNavMenu: React.FC<TraceTopNavMenuProps> = ({
...
+                          defaultMessage: 'Trace ID: {traceId}',
+                          values: { traceId },
+                        })}
+                      </EuiText>
+                    </EuiFlexItem>
+                    <EuiFlexItem grow={false}>
+                      <EuiIcon
+                        type="copy"
+                        size="s"
+                        style={{ marginLeft: '4px' }}
```
angle943, line 97: can we avoid inline styles?

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_tables/span_hierarchy_table.tsx
angle943: +1. we can just replace it with <SpanStatusFilter .... />

But why are we passing empty function as an alternative to `setSpanFiltersWithStorage`?

Is `props.setSpanFiltersWithStorage` being optional a realistic scenario?

If it is always existing, we should update the type of this component to make it non-nullable.

If it is not existing, we should update the props of `<SpanStatusFilter />` to make that prop optional, and handle it accordingly there

yanatha99: We had previous bug where nesting custom functional components on TraceDetails was identified as a React anti-pattern (which honestly surprised me as well) and caused re-render loops. This is why you'll see this pattern in a few places, calling the function directly rather than as a component. 

As it is currently, the SpanStatusFilter works as intended, but I can fast follow with additional changes and more thoroughly test which approach is better for rendering this component specifically.

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/ppl_resolve_helpers.tsx
yanatha99: Don't even need this function tbh, It's a relic of when I was tinkering around with slightly more complex logic to determine if a span is "OK" that mirrored the `isSpanError` function behavior. Now since it's just a negation of that `isSpanError` function I'll remove it in follow up.

Also the naming convention was to distinguish that this is how we determine if a span status is "OK".

src/plugins/explore/public/application/pages/traces/trace_details/public/top_nav_buttons.tsx
yanatha99: No, the function is within the nav button that the icon has been added to. The button already handles copying, this was just to add a visual indicator within that button to make it clear to users that they could copy the traceId.

src/plugins/explore/public/application/pages/traces/trace_details/public/top_nav_buttons.tsx
yanatha99: Yes, not going to block the PR on this, but can do a follow up change. I'll let the other devs on this project know as well to not inline the styles.

src/plugins/explore/public/application/pages/traces/trace_details/public/traces/span_detail_tables/span_hierarchy_table.tsx
yanatha99: As to why the prop is set up the way it is. This is because SpanHierarchyTable uses SpanTableProps which the SpanListTable also uses. The SpanListTable does not have the new SpanStatusFilter, so setSpanFiltersWithStorage is optional in the SpanTableProps. Although we know we always call SpanHierarchyTable with setSpanFiltersWithStorage passed, typescript does not know that and flags the setSpanFiltersWithStorage as possibly being undefined when it's passed to the SpanStatusFilter. The SpanStatusfilter requires the setSpanFilterWithStorage to actually change the filters applied to spans.
As part of fast follow can remove the empty function and just add a non-null assertion

There a few other props that fall into a similar category as this where they're only used by the SpanHierarchyTable but no the SpanListTable. This is due to the SpanTableProps being a relic of when both tables were very similar and defined in a singular file that I broke up to be more readable and testable after changing the SpanHierarchyTable to represent waterfall timeline view with hierarchical organization of the spans. Do want to separate out the SpanTableProps into some base props and hierarchy props that inherit from the base, but it's not completely straight forward due to how tangled the two tables were originally.

Another change that can be made is to go ahead and just make setSpanFiltersWithStorage required and pass it to both tables and add the SpanStatusFilter to SpanListTable as well. Reason this wasn't done for this PR, was UX had only wanted the filter on the SpanHierarchyTable at the time.

Maosaic: I'm not sure why nesting custom functional components on `TraceDetails` consider as React anti-pattern, but calling functional components as a plain function is defintly anti-pattern.

`SpanStatusFilter` uses React hooks (useState/useEffect/useMemo/useCallback), calling it as a plain function violates the Rules of Hooks and will cause runtime errors and lifecycle bugs.

angle943 (approved):

yanatha99:
Fast follow changes
1. Change `SpanStatusFilter` from multi-select to single-select
2. Render `SpanStatusFilter` as FC
3. Remove empty function passed to `setSpanFiltersWithStorage` and add non-null assertion
4. Remove `isSpanOk` function, just directly use `!isSpanError`
5. Add new lines for styling on copy icon

PR #10749, opened 2025-10-17
feat: Read schemaMappings for logs from dataset object

earthling-shruti:
#### Improved trace to log correlation for multiple log datasets
Logs related to traces and spans appear in a few places:
- Trace detail page in the "related logs" tab.
- Trace detail page in the "span detail" flyout.
- Discover traces page in the "trace detail" flyout.
The related logs tab in the trace detail page was previously updated to show logs from multiple logs datasets. This PR updates the other two experiences (same component) to also show logs from multiple trace datasets.
#### Use schema mappings from datasets
Previously, schema mappings were stored in the APM-Correlation object for trace to logs correlation. Now, schema mappings are stored as part of the logs dataset. Update the same components to read schema mappings from the logs datasets instead of the correlation object.
#### Other minor fixes
- The Discover links to logs should always appear, even when there are no related logs available. This will allow the user to explore more logs in Discover.
- Log count in the related logs tab was incorrectly showing log count for a single log dataset. Update this number to reflect log counts from all related logs datasets.
- Before, the saved objects plugin was used to fetch the logs dataset. Now, the correlation service relies on the data plugin to fetch the logs dataset. The saved object API is now the fallback mechanism in case the data plugin isn't available. Note: We should remove the saved object dependency since the [data plugin is required by Explore](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/92ae684ef56ac64fe97c3d37e393819a236f28bf/src/plugins/explore/opensearch_dashboards.json#L9). Will do this as a follow-up.
- Fix the type error on trace detail link from the trace detail flyout in Discover traces page.

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10729, https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10714

## Screenshot
Screenshot 2025-10-17 at 12 39 47 PM
Screenshot 2025-10-17 at 10 27 26 AM
Screenshot 2025-10-17 at 9 55 21 AM
Screenshot 2025-10-17 at 9 54 59 AM

## Changelog
- feat: use schemamapping from logs dataset, improve multiple log correlation support

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/pages/traces/trace_details/public/logs/span_logs_tab.test.tsx
+++ b/src/plugins/explore/public/application/pages/traces/trace_details/public/logs/span_logs_tab.test.tsx
@@ -92,6 +92,21 @@ describe('SpanLogsTab', () => {
 
       // Dataset name is displayed
       expect(screen.getByText('logs-*')).toBeInTheDocument();
+
+      // Description is rendered
+      expect(screen.getByText('View logs related to this specific span')).toBeInTheDocument();
+
+      // Dataset accordion container is present
+      expect(screen.getByTestId('dataset-accordion-logs-dataset-id')).toBeInTheDocument();
+
+      // "View in Discover Logs" button is rendered
+      expect(screen.getByText('View in Discover Logs')).toBeInTheDocument();
+
+      // Component is not showing the loading state
+      expect(screen.queryByTestId('loading-spinner')).not.toBeInTheDocument();
+
+      // Component is not showing the "no datasets" message
+      expect(screen.queryByText('No logs found for this dataset')).not.toBeInTheDocument();
     });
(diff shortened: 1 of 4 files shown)


TackAdam:
The redirection when there are no logs button click does not work, it probably requires new functionality for the button, as the previous one depends on the log being present to form the url with span id ect...
RedirectionNotWorking

earthling-shruti:
Thanks, @TackAdam - I ended up force pushing and closing this PR by mistake. Removed the link for datasets with no logs in a new PR - https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10762

PR #10750, opened 2025-10-17
Bumping OpenSearch version to 3.3 in Cypress workflows

sejli:
### Description

Bumps OpenSearch version to 3.3 in Cypress workflows

## Changelog

- infra: Bump OpenSearch version to 3.3 in Cypress workflows

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


paulstn:
I'm pretty sure we need to update the test `rule_matching_vis.spec.js` in `osd:ciGroup16Explore`. Even when running that test locally with 3.3 OpenSearch, it fails with the same issue:
image

I believe this is because of an underlying change in the PPL query's response, after calcite had been enabled on that release.

paulstn:
Closed in favor of: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10771 which includes the needed test fixes.

PR #10754, opened 2025-10-19
[Explore Vis] Fix table visualiztion issues

lunalzm:
### Description

This PR addresses several critical issues with table visualizations in the Explore plugin:

  1. Field Order Issue: Table columns were being displayed in type-based order (numerical → categorical → date) instead of preserving the original field order from the query response.
  2. Column Visibility Issue: When running new queries, fields would be hidden by default.
  3. Fix pagination not working
  4. Changed the table cell default text alignment to `left` to mitigate the table cell text move issue

The fixes ensure that table visualizations now correctly respect the field order specified in queries and properly manage column visibility across query executions.

#### Before
Screenshot 2025-10-20 at 01 04 31

#### After
Screenshot 2025-10-20 at 01 06 42

#### Before
Screenshot 2025-10-20 at 01 08 27

#### After
image

## Testing the changes

  1. Field Order Test:
  `SOURCE = opensearch_dashboards_sample_data_logs | FIELDS @timestamp, extension, bytes, agent`
    - Verify fields appear in order: @timestamp → extension → bytes → agent
    - Previously would show: bytes → @timestamp → extension → agent
  2. Column Visibility Test:
  `SOURCE = opensearch_dashboards_sample_data_logs | FIELDS bytes, memory, referer, url, agent`
    - Verify all fields are hidden

## Changelog
- fix: Fix table vis fields order not the same as fields order in the query
- fix: table vis pagination not working due to unnecessary rerendering
- fix: change the default table column alignment to left

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/table/table_vis.tsx
+++ b/src/plugins/explore/public/components/visualizations/table/table_vis.tsx
@@ -176,7 +176,7 @@ export const TableVis = React.memo(
 
     const renderFooterCellValue = useCallback(
       ({ columnId, setCellProps }: EuiDataGridCellValueElementProps) => {
-        const alignment = styleOptions?.globalAlignment || 'auto';
+        const alignment = styleOptions?.globalAlignment || 'left';
         const textAlign =
           alignment === 'auto'
             ? columnTypes[columnId] === 'numerical'


src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -100,6 +101,30 @@ export const TableVis = React.memo(
       }
     }, [styleOptions?.showColumnFilter]);
 
+    useEffect(() => {
```
ruanyl: Could you leave a comment on this `useEffect` to describe the logic?

lunalzm: Sure!

src/plugins/explore/public/components/visualizations/visualization_render.tsx
```diff
@@ -53,11 +53,13 @@ export const VisualizationRender = ({
...
 
   const columns = useMemo(() => {
-    return [
+    const allColumns = [
       ...(visualizationData?.numericalColumns ?? []),
       ...(visualizationData?.categoricalColumns ?? []),
       ...(visualizationData?.dateColumns ?? []),
     ];
+
+    return allColumns.sort((a, b) => a.id - b.id);
```
ruanyl: Can you also check `explore_embeddable.tsx` to see if the same logic should apply there? or perhaps move this logic to `table_vis` component?

src/plugins/explore/public/components/visualizations/visualization_render.tsx
```diff
@@ -113,11 +113,7 @@ export const VisualizationRender = ({
 
   if (visConfig?.type === 'table') {
     return (
-      <TableVis
-        styleOptions={visConfig.styles as TableChartStyle}
-        rows={visualizationData.transformedData ?? []}
-        columns={columns}
-      />
+      <TableVis styleOptions={visConfig.styles as TableChartStyle} rows={rows} columns={columns} />
```
Hailong-am: should rows be `visualizationData.transformedData ?? []`?

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -25,8 +25,13 @@ interface TableVisProps {
 
 export const TableVis = React.memo(
   ({ rows, columns, styleOptions, pageSizeOptions, showStyleSelector }: TableVisProps) => {
+    const sortedColumns = useMemo(() => columns.sort((a, b) => a.id - b.id), [columns]);
```
wanglam: Sort will update the original columns, I prefer to construct a new array for sorting.

src/plugins/explore/public/components/visualizations/visualization_render.tsx
ruanyl, line 116: I changed it to `rows` which is defined as this, should be the same
```
  const rows = useMemo(() => {
    return visualizationData?.transformedData ?? [];
  }, [visualizationData?.transformedData]);
```

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
ruanyl: good point, I will update

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
Hailong-am: is this just a replacement by new columns?

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -100,6 +105,33 @@ export const TableVis = React.memo(
       }
     }, [styleOptions?.showColumnFilter]);
 
+    // Handle column visibility when columns change
+    // If there are no common fields between old and new columns, show all new columns
+    // If there are common fields, preserve existing visible columns and add any new ones
+    useEffect(() => {
+      const currentFieldNames = sortedColumns.map(({ name }) => name);
+      const prevFieldNames = prevFieldNamesRef.current;
```
wanglam: I prefer to delete this ref, can we calculate the new columns based `prevVisibleColumns` in `setVisibleColumns` callback and `sortedColumns`?

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
ruanyl: @lunalzm How do you reproduce the issue?

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
Hailong-am: should this be `left` as well?

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
ruanyl: @Hailong-am @lunalzm I removed this `useEffect`, and pushed a commit that fix the column visibility issue from the root

Hailong-am: I just read the comment for this method and didn't read the logic fully, so i think it just a replacement by new columns. There have different property in `VisColumn` interface, i see we are using `id`, `name` and `column`, not sure what's the difference. 


`const sortedColumns = useMemo(() => [...columns].sort((a, b) => a.id - b.id), [columns]);`

`const newColumnIds = sortedColumns.map(({ column }) => column);`

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
ruanyl: that's right

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
lunalzm: This issue sometimes happens when I run a new query or refresh the page. Some columns get hidden by default.

https://github.com/user-attachments/assets/900d3cb6-de75-46ef-9e94-7161b22fa5cc

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
ruanyl: The overall implementation has changed, this `useEffect` no longer exists now

src/plugins/explore/public/components/visualizations/table/table_vis.tsx
```diff
@@ -25,8 +25,12 @@ interface TableVisProps {
 
 export const TableVis = React.memo(
   ({ rows, columns, styleOptions, pageSizeOptions, showStyleSelector }: TableVisProps) => {
+    const sortedColumns = useMemo(() => [...columns].sort((a, b) => a.id - b.id), [columns]);
```
Hailong-am, line 28: nit: we can make the columns is ordered in props in parent component, that could reduce the change.

ruanyl: This component is used at different places, having the sorting here is to avoid having the same logic at multiple places

ruanyl:
Checked the failed test is unrelated, merging now

PR #10759, opened 2025-10-20
add navigate to page tool

jiapingzeng:
### Description

Add page navigation AG-UI tool.

Chat window and session is not currently preserved on navigation. Will address in future PR.

## Screenshot
Tool can be detected by agent, then tried navigating to visualizations page which worked:
image

Temporarily disabled navigate to show that tool use renders correctly:
image

## Testing the changes

Verified that backend agent can see the tool and that I can navigate to visualizations page from chat

## Changelog

- feat: [Context Provider][Tools] Add navigate to page tool

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/.github/workflows/cypress_workflow.yml
+++ b/.github/workflows/cypress_workflow.yml
@@ -52,7 +52,7 @@ env:
   COMMENT_TAG: "[MANUAL CYPRESS TEST RUN RESULTS]"
   COMMENT_SUCCESS_MSG: ":white_check_mark: Cypress test run succeeded!"
   COMMENT_FAILURE_MSG: ":x: Cypress test run failed!"
-  LATEST_VERSION: "2.17.0"
+  LATEST_VERSION: '3.3.0'
 
 jobs:
   cypress-tests:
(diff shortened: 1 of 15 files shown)


src/plugins/chat/public/actions/navigate_action.tsx
```diff
@@ -0,0 +1,155 @@
...
+import React from 'react';
+import { EuiPanel, EuiText, EuiIcon, EuiFlexGroup, EuiFlexItem } from '@elastic/eui';
+import { useAssistantAction } from '../../../context_provider/public';
+import { useOpenSearchDashboards } from '../../../opensearch_dashboards_react/public';
+import { CoreStart } from '../../../../core/public';
+
+interface NavigateArgs {
+  appId: string;
+  path?: string;
+  description?: string;
```
ananzh, line 15: navigation has two other optional properties, replace and state. state is used to pass context/data and it might not be the correct way for context sharing for chatbot. It is often used to share some app states which I don't think this tool should manage. for `replace`, it can control whether navigation creates a new history entry (push) or replaces the current one (replace) which is useful if we need to navigate back and force. let's add `replace?: boolean` to follow the navigation pattern.

src/plugins/chat/public/actions/navigate_action.tsx
```diff
@@ -0,0 +1,155 @@
...
+      required: ['appId'],
+    },
+    handler: async (args) => {
+      const { appId, path, description } = args;
+
+      try {
+        if (!appId || typeof appId !== 'string') {
+          throw new Error('appId is required and must be a string');
+        }
+
```
ananzh, line 55: should we also validate appId in services.application.applications$.getValue()? it might not be available or accessible. for example, explore is not accessible in analytic workspaces.

ananzh:
Nit: if we want to distinguish between appId not available or no permission, we could use 

```
  const currentApps = services.application.applications$.getValue();
  const targetApp = currentApps.get(args.appId);

   if (!targetApp) {
    ...
   }

  if (targetApp.status === AppStatus.inaccessible) {
     ...
    }
```

PR #10771, opened 2025-10-21
Bumping OpenSearch version to 3.3 in Cypress workflows

paulstn:
### Description

Bumps OpenSearch version to 3.3 in Cypress workflows. 

Additionally, fix the `rule_matching_vis.spec.js` cypress test. This test started failing in 3.3 OpenSearch due to the underlying PPL query starting to return different results. 
```
source=${datasetName} | stats count() as count by span(timestamp, 1d) as timestamp, category
```
would return the error:
```
field [timestamp] not found; input fields are: [agent, timestamp0, count]
```
due to the dataset already containing a field named `timestamp`, which causes PPL to force the second alias to show as `timestamp0`.

To fix this issue, the alias was instead changed to prevent this naming overlap:
```
source=${datasetName} | stats count() as count by span(timestamp, 1d) as timestamp_span, category
```

## Changelog

- infra: Bump OpenSearch version to 3.3 in Cypress workflows
- fix: Fixed the rule_matching_vis cypress test to use a working query

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/10771.yml
+++ b/changelogs/fragments/10771.yml
@@ -1,0 +1,5 @@
+infra:
+- Bump OpenSearch version to 3.3 in Cypress workflows ([#10771](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10771))
+
+fix:
+- Fixed the rule_matching_vis cypress test to use a working query ([#10771](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10771))


.github/workflows/cypress_workflow.yml
```diff
@@ -283,9 +283,12 @@ jobs:
           rm -f opensearch-*.tar.gz
         shell: bash
 
-      - name: Remove security plugin
+      - name: Remove security plugin and plugins that depend on security
         if: matrix.config == 'query_enhanced' || matrix.config == 'explore'
         run: |
+          /bin/bash -c "yes | ./opensearch-${{ env.LATEST_VERSION }}/bin/opensearch-plugin remove opensearch-skills"
```
TackAdam, line 289: Can we clarify why these additional removes are needed?

paulstn: I believe security does not get removed without the removal of these plugins as well. @sejli could you confirm if this is why you removed it originally? I'll look into whether I can remove this any other way soon.

PR #10782, opened 2025-10-22
Encode and decode dataset id as URL

Maosaic:
### Description

Fix dataset navigation for IDs with special characters

  Solution

  - Encode dataset IDs using encodeURIComponent() when constructing navigation URLs
  - Decode dataset IDs using decodeURIComponent() when reading from route parameters

  Changes Made

  Encoding (when navigating):
  - use_dataset_selector.tsx:57,78 - Dataset creation/duplicate navigation
  - dataset_table_columns.tsx:46 - Table row click navigation
  - tabs/utils.ts:146 - Field edit navigation (getPath function)
  - scripted_fields_table/components/header/header.tsx:46 - Create scripted field navigation

  Decoding (when reading URL params):
  - edit_dataset_container.tsx:20 - Reading dataset ID from route
  - create_edit_field_container.tsx:22 - Reading dataset ID from route

## Changelog

- fix: Encode and decode dataset id as URL

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/10782.yml
+++ b/changelogs/fragments/10782.yml
@@ -1,0 +1,2 @@
+fix:
+- Encode and decode dataset id as URL ([#10782](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10782))


TackAdam (approved):

d-buckner:
Changes look safe to me, but I believe we may be better served solving for this a layer or two down so we solve for this special character issue in urls across the whole app. For example, having the core router be concerned with encoding/decoding rather than ui components. Curious of your thoughts here.

Maosaic:
> Changes look safe to me, but I believe we may be better served solving for this a layer or two down so we solve for this special character issue in urls across the whole app. For example, having the core router be concerned with encoding/decoding rather than ui components. Curious of your thoughts here.

That would be great if we can find a way to do so, but I'm wondering if it's possible to have url decode build in when reading params

angle943 (approved):

Issue #10741, opened 2025-10-17
[RFC] Global Search Enhancements for New Home


wanglam:
## Summary

This RFC proposes three key enhancements to the global search functionality in OpenSearch Dashboards when the new home is enabled:

1. **Expanded Header with Centralized Global Search**: Repurpose the expanded header to display only a centered global search bar, removing all other header elements when both new home and `useExpandedHeader` are enabled
2. **Global Assets Search with @ Prefix**: Introduce `@` prefix to search for saved objects (dashboards, visualizations) directly from the global search bar
3. **Global Search Submit Commands**: Add extensible command system triggered by Enter key to enable actions like opening a chatbot with the search query

**Note**: These enhancements are specifically designed for and only work when the new home is enabled.

### Current State and Problems

**Problem 1: Underutilized Expanded Header with New Home**
- The expanded header space could be better utilized for global features when new home is active
- There is a need to provide a prominent location for global search functionality in the new home experience

**Problem 2: Limited Action Options from Search**
- Users can only click on search results to navigate
- No way to take alternative actions with the search query (e.g., start a chatbot conversation)
- Limits the extensibility of the search experience

**Problem 3: Missing Assets Search**
- Users cannot search for their saved dashboards and visualizations from the global search
- Must navigate to management pages to find assets
- Reduces discoverability and efficiency

### Proposed Solution

This RFC proposes three technical enhancements:

1. **UI Change**: When new home and `useExpandedHeader` are both enabled, repurpose the expanded header to display only a centered global search bar, removing all other header elements (branding logo, navigation controls, etc.)

2. **New Search Capability**: Introduce `@` prefix to enable asset-only search, allowing users to search for saved objects (dashboards, visualizations) directly from the global search bar with workspace-aware filtering

3. **Extensibility Feature**: Add a submit command system that executes when users press Enter, enabling plugins to register custom actions (primary use case: opening AI chatbot with search query)

### Current User Flow (New Home Enabled)

**Finding a Dashboard**:
```
User wants to find a dashboard
    ↓
Navigate to Assets
    ↓
Use table search/filters
    ↓
Click on dashboard
    ↓
Navigate to dashboard
```

### New User Flow (New Home Enabled)

**Finding a Dashboard**:
```
User wants to find a dashboard
    ↓
Click on global search (top center or left nav)
    ↓
Type dashboard name
    ↓
See results instantly (grouped by Pages/Assets)
    ↓
[...]

PR #10789, opened 2025-10-23
Add assets search and enhance search commands

wanglam:
### Description

This PR enhances the global search functionality by adding two new capabilities:

1. **Assets Search Command**: Enables searching for dashboards and visualizations directly from the global search bar. Users can now quickly find and navigate to saved objects without leaving their current context.

2. **Enhanced Command Actions**: Extended the `GlobalSearchCommand` interface to support action-based commands that execute when users press Enter in the search bar, providing a more intuitive way to perform search-related actions.

Key changes include:
- Added `searchAssets` function to search for dashboards and visualizations via the saved objects API
- Extended `GlobalSearchCommand` interface with `action` property to support Enter-key triggered actions
- Enhanced the search bar UI to display command input placeholders and handle Enter key events for command execution

### Issues

#10741 

## Screenshot

image

## Changelog

- feat: Add assets search command to find dashboards and visualizations from global search
- feat: Enhance global submit commands for Enter-key triggered actions in global search

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/dashboard/public/application/components/dashboard_top_nav/__snapshots__/dashboard_top_nav.test.tsx.snap
+++ b/src/plugins/dashboard/public/application/components/dashboard_top_nav/__snapshots__/dashboard_top_nav.test.tsx.snap
@@ -148,6 +148,7 @@ exports[`Dashboard top nav render in embed mode 1`] = `
               "globalSearch": Object {
                 "getAllSearchCommands": [MockFunction],
                 "getAllSearchCommands$": [MockFunction],
+                "registerSearchCommand": [MockFunction],
                 "unregisterSearchCommand": [MockFunction],
               },
               "logos": Object {
@@ -1319,6 +1320,7 @@ exports[`Dashboard top nav render in embed mode, and force hide filter bar 1`] =
               "globalSearch": Object {
                 "getAllSearchCommands": [MockFunction],
                 "getAllSearchCommands$": [MockFunction],
+                "registerSearchCommand": [MockFunction],
                 "unregisterSearchCommand": [MockFunction],
               },
               "logos": Object {
@@ -2490,6 +2492,7 @@ exports[`Dashboard top nav render in embed mode, components can be forced show b
               "globalSearch": Object {
                 "getAllSearchCommands": [MockFunction],
                 "getAllSearchCommands$": [MockFunction],
+                "registerSearchCommand": [MockFunction],
                 "unregisterSearchCommand": [MockFunction],
               },
               "logos": Object {
@@ -3661,6 +3664,7 @@ exports[`Dashboard top nav render in full screen mode with appended URL param bu
               "globalSearch": Object {
                 "getAllSearchCommands": [MockFunction],
                 "getAllSearchCommands$": [MockFunction],
+                "registerSearchCommand": [MockFunction],
                 "unregisterSearchCommand": [MockFunction],
               },
               "logos": Object {
@@ -4832,6 +4836,7 @@ exports[`Dashboard top nav render in full screen mode, no componenets should be 
               "globalSearch": Object {
                 "getAllSearchCommands": [MockFunction],
                 "getAllSearchCommands$": [MockFunction],
+                "registerSearchCommand": [MockFunction],
                 "unregisterSearchCommand": [MockFunction],
               },
               "logos": Object {
@@ -6003,6 +6008,7 @@ exports[`Dashboard top nav render with all components 1`] = `
               "globalSearch": Object {
                 "getAllSearchCommands": [MockFunction],
                 "getAllSearchCommands$": [MockFunction],
+                "registerSearchCommand": [MockFunction],
                 "unregisterSearchCommand": [MockFunction],
               },
               "logos": Object {


src/core/public/chrome/global_search/global_search_service.ts
```diff
@@ -46,16 +47,28 @@ export interface GlobalSearchCommand {
...
+ */
+export interface GlobalSearchSubmitCommand {
+  id: string;
+  name: string;
+  run: (payload: { content: string }) => void;
 }
 
 export interface GlobalSearchServiceSetupContract {
   registerSearchCommand(searchCommand: GlobalSearchCommand): void;
+  registerSearchSubmitCommand(searchResultCommand: GlobalSearchSubmitCommand): void;
```
ruanyl: what's the difference between `registerSearchCommand` and `registerSearchSubmitCommand`? would be nice to add comment to each of them

src/core/public/chrome/global_search/global_search_service.ts
```diff
@@ -96,16 +110,39 @@ export class GlobalSearchService {
...
+    if (newCommands.length === this.searchSubmitCommands$.getValue().length) {
+      return;
+    }
+    this.searchSubmitCommands$.next(newCommands);
+  }
+
   public setup(): GlobalSearchServiceSetupContract {
     return {
       registerSearchCommand: this.registerSearchCommand.bind(this),
+      registerSearchSubmitCommand: this.registerSearchSubmitCommand,
```
ruanyl: I don't seem to see where is this function called `registerSearchSubmitCommand`, did I miss anything?

ruanyl:
Can we specify the type of assets when searching? Right now it seems to search all types of saved object, can we specify type in the query? for example: `@vis:test`, this will search visualization with keyword `test`

src/core/public/chrome/global_search/global_search_service.ts
wanglam: For now the only search submit command is in the dashboards-assistant repo. Will raise a new PR for adding an example for this new methods.

src/core/public/chrome/global_search/global_search_service.ts
```diff
@@ -46,16 +47,127 @@ export interface GlobalSearchCommand {
...
    * @param callback callback function when search is done
    */
-  run(value: string, callback?: () => void): Promise<ReactNode[]>;
+  run(value: string, callback?: () => void, abortSignal?: AbortSignal): Promise<ReactNode[]>;
 }
 
+/**
+ * @experimental
+ */
+export interface GlobalSearchSubmitCommand {
```
ruanyl: Agreed to use an unified Command interface, instead of introducing a new `GlobalSearchSubmitCommand`, it should extend the existing interface `GlobalSearchCommand`

src/core/public/chrome/global_search/global_search_service.ts
```diff
@@ -46,16 +47,127 @@ export interface GlobalSearchCommand {
    * @param value search query
    * @param callback callback function when search is done
    */
-  run(value: string, callback?: () => void): Promise<ReactNode[]>;
+  run(value: string, callback?: () => void, abortSignal?: AbortSignal): Promise<ReactNode[]>;
```
ruanyl: Can we update the function interface for better extensibility? in case there will be more context we will need to pass in the future
```suggestion
  run(value: string, options: {callback?: () => void, abortSignal?: AbortSignal}): Promise<ReactNode[]>;
```

ruanyl (approved):

src/core/public/chrome/global_search/global_search_service.ts
wanglam: Good suggestion. I'm prefer to keep `value: string` and `callback?: () => void` as the old implementation. Since I don't want to change the navigation search implementation. The new `abortSignal` can be wrapped with options. What do you think about it?

wanglam:
> Can we specify the type of assets when searching? Right now it seems to search all types of saved object, can we specify type in the query? for example: `@vis:test`, this will search visualization with keyword `test`

I think it can be a feature support in the future. Store a mapping for assets types and register more alias. We can reuse the assets search logic. We need to figure out the alias conflicts due to `@` already is an alias. For current implementation, it will try to search `vis:test`.

src/core/public/chrome/ui/header/header_search_bar.tsx
```diff
@@ -121,51 +130,74 @@ export const HeaderSearchBar = ({ globalSearchCommands, panel, onSearchResultCli
...
+  );
 
   const onSearch = useCallback(
     async (value: string) => {
-      const filteredCommands = globalSearchCommands.filter((command) => {
+      const abortController = new AbortController();
+      ongoingAbortControllersRef.current.push({ controller: abortController, query: value });
+      if (enterKeyDownRef.current) {
+        globalSearchCommands
+          .filter((item) => !!item.action)
```
ruanyl, line 157: Nit: this `filter` seem unnecessary, because it checks `command.action` before calling it in `forEach`

src/core/public/chrome/ui/header/header_search_bar.tsx
```diff
@@ -121,51 +130,74 @@ export const HeaderSearchBar = ({ globalSearchCommands, panel, onSearchResultCli
...
+        enterKeyDownRef.current = false;
+        setIsPopoverOpen(false);
+        setSearchValue('');
+        searchBarInputRef.current?.blur();
+        return;
+      }
+      const commandsWithoutActions = globalSearchCommands.filter(
+        (command) => command.type !== 'ACTIONS'
+      );
+      const filteredCommands = commandsWithoutActions.filter((command) => {
```
ruanyl, line 172: Could you leave a comment here to explain why having  `filteredCommands`? It seem it is sorted as `[alias/search command, action commands]`

src/core/public/chrome/ui/header/header_search_bar.tsx
```diff
@@ -181,14 +213,23 @@ export const HeaderSearchBar = ({ globalSearchCommands, panel, onSearchResultCli
               [type]: (acc[type] || []).concat(items),
             };
           }, {} as Record<SearchCommandKeyTypes, ReactNode[]>);
-
         const sections = Object.entries(searchResults).map(([key, items]) => {
           const sectionHeader = SearchCommandTypes[key as SearchCommandKeyTypes].description;
-          return resultSection(items, sectionHeader);
+          return resultSection(items, key !== 'ACTIONS' ? sectionHeader : undefined);
```
ruanyl, line 218: I would suggest not have this special logic `key !== 'ACTIONS'`, instead we could have a extra `displayName` property to `SearchCommandTypes` which can be set to empty

src/core/public/chrome/ui/header/header_search_bar.tsx
```diff
@@ -202,16 +243,32 @@ export const HeaderSearchBar = ({ globalSearchCommands, panel, onSearchResultCli
       incremental
       onSearch={onSearch}
       fullWidth
-      placeholder={i18n.translate('core.globalSearch.input.placeholder', {
-        defaultMessage: 'Search the menu',
-      })}
+      placeholder={
+        globalSearchCommands.find((item) => item.inputPlaceholder)?.inputPlaceholder ??
```
ruanyl, line 247: This makes it to only display the first command that has `inputPlaceholder`, I would suggest to make the placeholder to be computed from all different commands.

src/core/public/chrome/global_search/global_search_service.ts
```diff
@@ -76,7 +213,11 @@ export interface GlobalSearchServiceStartContract {
  * @experimental
  */
 export class GlobalSearchService {
-  private searchCommands = [] as GlobalSearchCommand[];
+  private searchCommands$ = new BehaviorSubject<GlobalSearchCommand[]>([]);
```
Hailong-am, line 216: nit: why did we change it observable?

wanglam: Need to register some search commands in the start phase. I've expose a `registerSearchCommands` method in the start contract.

src/core/public/chrome/ui/header/header_search_bar.tsx
wanglam: Yes. In the current implementation, only alias command will be execute if user start with alias. Add filter here to make sure all action commands can be executed and rendered in the end. Let me added in the next PR.

wanglam:
Checked failed CI not related to this PR. Should be safe to merge.

PR #10816, opened 2025-10-28
Fix creating index-patterns with ccs causes search threadPool to be exhausted

### LINKED ISSUE #8596: Creating index-patterns causes search threadPool to be exhausted
when I set the ccs in the cluster settings(the address is pointed to itself to reproduce)
```
PUT _cluster/settings
{
  "persistent": {
      "cluster": {
         "remote": {
            "es8_es11": {
               "mode": "proxy",
               "proxy_address": "127.0.0.1:9300"
            }
         }
      }
  }
}
```
and create a new index-patterns in the dashboard,  I find that search is immediately exhausted in the target cluster:
image
And search slow log is as follows:
image
```
POST *:*/_search
{
  "size": 0,
  "timeout": "30000ms",
  "query": {
    "match_all": {
      "boost": 1
    }
  },
  "track_total_hits": 2147483647,
  "aggregations": {
    "indices": {
      "terms": {
        "field": "_index",
        "size": 200,
        "min_doc_count": 1,
        "shard_min_doc_count": 0,
        "show_term_doc_count_error": false,
        "order": [
          {
            "_count": "desc"
          },
          {
            "_key": "asc"
          }
        ]
      }
    }
  }
}
```
As the `total_shards`=35720, which equals the total number shards of the cluster, we know that all the indices are aggregated by the query.

image
we know that when query is from the "Index_Patterns".

It seems that we request the query to get the indices list of target cluster, when the target cluster is very large, this query is very dangerous.

If we should add the api in the opensearch code to get indices list of the ccs target cluster?

**Additional context**

kkewwei:
### Description

The Dashboard currently uses the *:*/_search endpoint to retrieve indices from CCS clusters, a compatibility measure for Elasticsearch versions prior to 7.10. For later versions, the `RestResolveIndexAction` interface is supported and can fully replace the *:*/_search endpoint.

This means the *:*/_search endpoint can be removed from the Dashboard code without impacting functionality in this scenario.

### Issues Resolved

closes #8596

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/dataset_management/public/components/create_dataset_wizard/lib/get_indices.test.ts
+++ b/src/plugins/dataset_management/public/components/create_dataset_wizard/lib/get_indices.test.ts
@@ -105,11 +105,10 @@ describe('getIndices', () => {
     });
 
     expect(http.get).toHaveBeenCalled();
-    expect(result.length).toBe(4);
+    expect(result.length).toBe(3);
     expect(result[0].name).toBe('f-alias');
     expect(result[1].name).toBe('foo');
-    expect(result[2].name).toBe('opensearch_dashboards_sample_data_ecommerce');
-    expect(result[3].name).toBe('remoteCluster1:bar-01');
+    expect(result[2].name).toBe('remoteCluster1:bar-01');
   });
 
   it('should ignore ccs query-all', async () => {


kkewwei:
It seems that I should delete the method `getIndicesViaSearch` as well.

Hailong-am:
There has another place have same logic, we might need remove it as well. 

`src/plugins/dataset_management/public/components/create_dataset_wizard/lib/get_indices.ts`

Also, could you share screenshot when you delete this, CCS is still working as expected?

kkewwei:
@Hailong-am Thank you for your review. 

I have deleted the same logic. 

Your code changes are welcome! Since I’m unfamiliar with Dashboards, I’ve seen your modifications but not yet fully understood their purpose. To proceed with updates in fix_8596, I’ll temporarily move your changes to fix_8596_backup.

kkewwei:
> Also, could you share screenshot when you delete this, CCS is still working as expected?

We fix it in opendistro710, it's ok with ccs.
image

PR #10825, opened 2025-10-29
[Explore Vis] Sanitize field values containing special characters

lunalzm:
### Description

This PR fixes a bug where visualizations fail to render when data values contain special characters such as single quotes ('), dots (.), square brackets ([ ]).

### Before
image

### After
image

## Testing the changes
1. Create a multiLineChart (with one metric, one date, and one categorical column).

- Sample query: `source = opensearch_dashboards_sample_data_ecommerce | STATS count() by SPAN(order_date, 1h), category`

2. Confirm visualization renders without errors even with special characters in data values.

## Changelog
- fix: Sanitize field values containing special characters

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/utils/create_hover_state.ts
+++ b/src/plugins/explore/public/components/visualizations/utils/create_hover_state.ts
@@ -120,10 +120,9 @@ function createHiddenBarLayer(axisConfig: AxisConfig, options: Options & { barOp
   if (options.showTooltip) {
     if (axisConfig.color && !Array.isArray(axisConfig.y)) {
       const uniqueColorFieldValues = new Set(
-        (options.data ?? []).flatMap((d) => {
+        (options.data ?? []).map((d) => {
           const value = d[axisConfig.color?.name ?? ''];
-          if (value == null) return [];
-          return Array.isArray(value) ? value.filter((v) => v != null) : [value];
+          return Array.isArray(value) ? value.toString() : String(value);
         })
       );
       tooltip = createTooltip([


src/plugins/explore/public/components/visualizations/utils/create_hover_state.ts
```diff
@@ -127,12 +120,15 @@ function createHiddenBarLayer(axisConfig: AxisConfig, options: Options & { barOp
   if (options.showTooltip) {
     if (axisConfig.color && !Array.isArray(axisConfig.y)) {
       const uniqueColorFieldValues = new Set(
-        (options.data ?? []).map((d) => d[axisConfig.color?.name ?? ''])
+        (options.data ?? []).flatMap((d) => {
+          const value = d[axisConfig.color?.name ?? ''];
+          return Array.isArray(value) ? value : [value];
```
wanglam: Nit: do we need check if value if undefined here?

src/plugins/explore/public/components/visualizations/utils/utils.ts
```diff
@@ -402,3 +402,12 @@ export function applyTimeRangeToEncoding(
     };
   }
 }
+
+export const sanitizeFieldName = (fieldName: string | any): string => {
```
Hailong-am, line 412: do we have a full list characters that need to sanitize? or does vega have such built in ability

lunalzm: Currently, this issue has only been found to occur with these four special characters. Vega-Lite doesn't have a built-in sanitization utility for field names. Field names are used directly in expressions and JSON paths, so certain characters need escaping to prevent parsing errors.

Hailong-am: sounds like we need to escape all reserved keywords of JSON path? How about `@`?

lunalzm: Currently, `@` does not cause this problem.
image

Hailong-am (dismissed):

src/plugins/explore/public/components/visualizations/utils/create_hover_state.ts
```diff
@@ -127,12 +120,16 @@ function createHiddenBarLayer(axisConfig: AxisConfig, options: Options & { barOp
   if (options.showTooltip) {
     if (axisConfig.color && !Array.isArray(axisConfig.y)) {
       const uniqueColorFieldValues = new Set(
-        (options.data ?? []).map((d) => d[axisConfig.color?.name ?? ''])
+        (options.data ?? []).flatMap((d) => {
+          const value = d[axisConfig.color?.name ?? ''];
+          if (value == null) return [];
+          return Array.isArray(value) ? value.filter((v) => v != null) : [value];
```
ruanyl: Can you add a comment to explain the logic here? If array is flattened, will tooltip show correct values?

d-buckner:
This sanitization logic seems very brittle. Do we know how we get this malformed field names in the first place?

If we know the source I'd prefer to address there rather than do string replacements.

lunalzm:
> This sanitization logic seems very brittle. Do we know how we get this malformed field names in the first place?
> 
> If we know the source I'd prefer to address there rather than do string replacements.

Hi @d-buckner , the malformed field names actually originate from a `pivot` operation in `createCrosshairLayers` function. Here's what happens:
When a color field exists, the code performs a `pivot `operation, converting the color field's value into new field names. Then, it extracts unique values ​​for the color field from `options.data` and these values ​​come directly from user data. But we cannot predict the format or special characters of the value stored by the user. Therefore, we can only use replacement to handle this.

ruanyl:
This is fixed here: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10836/files#diff-8d3139833771793e8aef5c30a56811c908d04f0a004817478c2ac25b8d5974b4R31
Vega support this internally with `[fieldName]`, no need to do manual escape.

PR #10831, opened 2025-10-30
Serve web worker assets with cache-busting

d-buckner:
### Description

Serve code editor worker files with proper cache-busting via buildHash in the URL path, following the same pattern used for application bundles. This allows us to verify worker assets are same origin.

Worker files are now served at /{buildHash}/editor/workers/* where buildHash is derived from env.packageInfo.buildNum:
- Development: Number.MAX_SAFE_INTEGER (9007199254740991)
- Distribution: Actual build number from package.json

This ensures browsers and CDNs fetch fresh worker files when OSD is upgraded, preventing stale cache issues that could cause version mismatches between application code and worker threads.

Changes:
- Add worker_config.ts with buildHash-based URL generation helpers
- Register static worker route at /{buildHash}/editor/workers/*
- Add test coverage for worker route registration

## Changelog
- feat: add route for serving worker bundles

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


packages/osd-monaco/src/worker_config.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Worker filenames (relative paths).
+ * Monaco package only defines its worker files, not server routing.
+ */
+export const WORKER_FILES = {
+  ppl: 'ppl.editor.worker.js',
+  json: 'json.editor.worker.js',
+  xjson: 'xjson.editor.worker.js',
+} as const;
+
+export type WorkerId = keyof typeof WORKER_FILES;
```
angle943, line 16: i see you are not a fan of enums? :D

packages/osd-monaco/src/worker_config.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @param buildHash - The build hash for cache-busting (e.g., buildNum.toString())
+ * @returns Object with all worker URLs
+ */
+export function getWorkerUrls(buildHash: string) {
+  return {
+    ppl: `/${buildHash}/editor/workers/${WORKER_FILES.ppl}`,
+    json: `/${buildHash}/editor/workers/${WORKER_FILES.json}`,
+    xjson: `/${buildHash}/editor/workers/${WORKER_FILES.xjson}`,
+  } as const;
+}
```
angle943, line 42: nitpick: we've been trying to up our unit test coverage. its kind of unnecessary to add for these honestly, but to maintain some sort of unit test comprehensiveness in an app as big as ours, i think we should add unit tests including functions like these.

I don't think its a blocker, but as a followup u think you can sneak in some simple unit tests for these functions?

angle943:
@d-buckner do you know why there is no CHANGELOG file in this PR?

packages/osd-monaco/src/worker_config.ts
d-buckner: Yup, happy to add unit tests here in my next pr

d-buckner:
> @d-buckner do you know why there is no CHANGELOG file in this PR?

Looks like there was a node version since I was gone and I hadn't picked that up. Just updated and verified everything is generating properly now.

packages/osd-monaco/src/worker_config.ts
d-buckner: I don't mind em! I just needed these to be indexed for keyed lookup.

PR #10837, opened 2025-10-31
Fix recent queries not working for DQL and Lucene

### LINKED ISSUE #10810: [BUG] Discover "Recent queries" button doesn't work with DQL or Lucene
**Describe the bug**

Discover "Recent queries" button doesn't work with DQL or Lucene (only works with SQL and PPL).

**To Reproduce**

1. Enable the new Discover experience with `explore.enabled: true` and `workspace.enabled: true`
2. Set the query language to Lucene
3. Run a few queries to build up some history
4. Click the "Recent queries" button
5. Nothing happens

**Expected behavior**

It should load a table with the recent queries in it like it does when the query language is set to PPL or SQL

**OpenSearch Version**
3.3.1

**Dashboards Version**
3.3.0

**Plugins**

N/A

**Screenshots**

N/A

 - OS: Windows 10
 - Browser and version: Chrome 141.0.7390.123

**Additional context**

N/A

ahopp:
## Description
Fixes the Recent Queries button not working for DQL and Lucene query languages in the Discover plugin.

## Issues Resolved
Fixes #10810

## Root Cause
The Recent Queries feature was broken for DQL (kuery) and Lucene queries due to two issues in the query editor (`src/plugins/data/public/ui/query_editor/query_editor.tsx`):

**Visibility condition bug**: The `RecentQueriesTable` component visibility was gated by `isRecentQueryVisible && useQueryEditor`. The `useQueryEditor` flag is `false` for DQL and Lucene queries, preventing the dropdown from ever showing.

**Missing history save**: The `onSubmit` function never called `queryString.addToQueryHistory()`, so queries were not being saved to history in the first place.

##### `src/plugins/data/public/ui/query_editor/query_editor.tsx`

**Line 487**: Removed `&& useQueryEditor` from `RecentQueriesTable` visibility condition
- Before: `isVisible={isRecentQueryVisible && useQueryEditor}`
- After: `isVisible={isRecentQueryVisible}`

**Lines 157-160**: Added `queryString.addToQueryHistory()` call in `onSubmit` function

>      // Add query to queryString history for Recent Queries feature
>      if (currentQuery.query?.trim()) {
>        queryString.addToQueryHistory(currentQuery, dateRange);
>      }

## Testing
Tested locally with DQL and Lucene queries in the Discover plugin:
  - Executed multiple DQL and Lucene queries (e.g., order_id:570663, day_of_week:Thursday)
  - Clicked the "Recent queries" button
  - Verified that the dropdown appears and shows all executed queries
  - Confirmed queries can be re-run by clicking them from the list

DQL:
Image

Lucene:
Image

### Check List
  - All tests pass
  - New functionality includes testing
  - Commits are signed per the DCO using --signoff


ahopp:
Updated PR to align to template and added changeset file for PR

ahopp:
@ananzh, @ashwin-pc, or @abbyhu2000? Thoughts on the code changes?

ahopp:
Or maybe @TackAdam or @d-buckner? 👀

src/plugins/data/public/ui/query_editor/query_editor.tsx
```diff
@@ -154,6 +154,11 @@ export const QueryEditorUI: React.FC<Props> = (props) => {
         persistedLogRef.current.add(currentQuery.query);
       }
 
+      // Add query to queryString history for Recent Queries feature
+      if (currentQuery.query?.trim()) {
+        queryString.addToQueryHistory(currentQuery, dateRange);
```
ruchidh, line 206: have we tested the change for ppl too ? This is a known issue, but it seems to be happening to ppl also, can you confirm that this changes fixes across all language type?

ruchidh (dismissed):

src/plugins/data/public/ui/query_editor/query_editor.tsx
ahopp: I believe I did, but it's been a bit so let me spin up a local environment with the fix and see if I can get a screenshot of the fix working for PPL.

LDrago27:
@ahopp Thanks for the change. 

Currently for the PPL, SQL languages the recent queries exists as a button on the footer of the query editor. For Lucene/DQL we are using a compressed version of the query editor which lacks the footer. Your change introduces the table but there isn;t a way to show/hide the recent queries table. 

Current Behaviour for PPL/SQL
Screenshot 2025-11-19 at 12 46 16 PM

@ahopp if you don;t mind, can you address this functionality of show/hide the recent query table for Lucene/DQL. This would make the PR feature complete.

Feel free to reach out if you need any help.

LDrago27 (dismissed):
Please refer to this comment https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10837#issuecomment-3554574927

ahopp:
Can you clarify? When testing a local environment,  with the changes, for PPL (including invalid queries) I get this experience; 

image

If I click into the "Recent queries", I get this;

image

With DQL it has/had the previous experience of the "Recent queries" button is on the footer of the query editor, but only when you click into the query bar itself;

image

image

image

Are you requiring we implement that same the UX implemented by the PPL/SQL dev before we can fix this bug?

ahopp:
Also, @LDrago27, are you asking to implement the same CX or do we need to modify for the different treatment for DQL and Lucene? My assumption is you want to modify SingleLineInput to always show the footer (matching PPL/SQL behavior)? If so the changes needed would be; 

- Remove the editorIsFocused condition in SingleLineInput (src/plugins/data/public/ui/query_editor/editors/shared.tsx:177)
- Update CSS to handle the always-visible footer for single-line editors
- Test that it doesn't break the compact single-line layout

Is that correct?

LDrago27:
> Also, @LDrago27, are you asking to implement the same CX or do we need to modify for the different treatment for DQL and Lucene? My assumption is you want to modify SingleLineInput to always show the footer (matching PPL/SQL behavior)? If so the changes needed would be;
> 
> * Remove the editorIsFocused condition in SingleLineInput (src/plugins/data/public/ui/query_editor/editors/shared.tsx:177)
> * Update CSS to handle the always-visible footer for single-line editors
> * Test that it doesn't break the compact single-line layout
> 
> Is that correct?

@ahopp yeah. I was thinking the same, maybe we can add in the footer element into the singleLineQuery Editor at the bottom. That can house this information. Very Similar to what we have in Multiline Editor.

I don't like the idea of adding it on the query editor where only on clicking the option is being shown. Recent Query table eats up a lot of space on discover and users not having an obvious option to toggle it on/off will be a bad experience.

ahopp:
Actually, I think I'll also have to adjust the container styling as well. I think the current CX (non-persistent) allows for overlapping of the lower elements for DQL/Lucene while the PPL/SQL CX has a container with padding. I'll take a look, but let me know if that makes sense...and what you are asking for before we can fix this bug.

ahopp:
I'm just going to update based on those assumptions. This is what it will look like with the updated CX for Lucene and DQL; 

Not expanded for "Recent queries"
image

Expanded for "Recent queries"
image

Important to note: This is implemented outside of the form control wrapper just like PPL/SQL was. This means this isn't part of a UI library that can be easily reused.

ahopp:
I've made the updates. 

Somewhat related, when I look at the code, I see drift between the PPL/SQL work and the established CX. I think there needs to be some mechanism to propagate these changes into OUI or the core UX as part of the code review.

LDrago27 (approved):
Looks Good. @ahopp

ruchidh (approved):

PR #10844, opened 2025-11-03
[Bug][Datasets] Fix missing data source information in dataset selector for namespaced index patterns

ananzh:
## Description

When clicking the dataset selector button to choose an index pattern that was created with a data source, the data source information is not displayed. This causes:

* UI Issue: The dropdown shows only the index pattern name (e.g., `data_logs_small_time_1`) instead of the full namespaced name (e.g., `data-logs-2::data_logs_small_time_1`)

* Query Failure: When the dataset is selected, PPL/SQL queries fail with `IndexNotFoundException` because they are generated without the data source reference

Screenshot 2025-10-30 at 10 32 31 PM

Screenshot 2025-11-01 at 9 50 41 PM

#### Root Cause

The issue occurs specifically for index patterns created using the namespaced ID format, where the data source ID is embedded in the index pattern ID itself (e.g., `8561bce0-b7a4-11f0-a904-fb55c930393e::data_logs_small_time_1`).

In OpenSearch Dashboards, there are two ways to associate a data source with an index pattern:

* Traditional method: Store the data source reference in the `references` array of the saved object
* Namespaced ID method: Embed the data source ID directly in the index pattern ID using `::` delimiter

The `fetchIndexPatterns()` function in `index_pattern_type.ts` only checked the `references` array for data source information, but for namespaced IDs, the `references` array is empty. This caused the data source information to be lost when populating the dataset selector dropdown.

### Solution

Enhanced the `fetchIndexPatterns()` function to extract data source IDs from both sources:

* First, check the traditional `references` array (backward compatibility)
* If not found, parse the index pattern ID for the namespaced format (`dataSourceId::indexPatternName`)

## Changelog
- fix: [Bug] Fix missing data source information in dataset selector for namespaced index patterns

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/saved_search.spec.js
+++ b/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/01/saved_search.spec.js
@@ -55,7 +55,7 @@ const runSavedSearchTests = () => {
           cy.visit(`/w/${workspaceId}/app/data-explorer/discover`);
         });
 
-        cy.setDataset(config.dataset, DATASOURCE_NAME, config.datasetType, true);
+        cy.setDataset(config.dataset, DATASOURCE_NAME, config.datasetType);
         cy.osd.grabIdsFromDiscoverPageUrl();
 
         // using a POST request to create a saved search to load


LDrago27 (approved):

LDrago27:
Ignore the failure of Cigroup 16. It is unrelated to this change is being addressed by https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10873.

src/plugins/data/public/query/query_string/dataset_service/lib/index_pattern_type.ts
```diff
@@ -105,12 +105,23 @@ const fetchIndexPatterns = async (client: SavedObjectsClientContract): Promise<D
...
-        .map((savedObject) => savedObject.references.find((ref) => ref.type === 'data-source')?.id)
+        .map((savedObject) => {
+          // First try to get from references
+          const refDataSourceId = savedObject.references.find((ref) => ref.type === 'data-source')
+            ?.id;
+          if (refDataSourceId) {
+            return refDataSourceId;
+          }
+          // If not in references, check if the ID contains :: (namespaced format)
+          if (savedObject.id.includes('::')) {
```
SuZhou-Joe, line 120: Not related to this PR but weird why don't we keep the reference way for data source for dataset? Detecting special chars to parse data source id is introducing a dependency on how data source based index pattern id gets generated.

PR #10849, opened 2025-11-04
Enhance query action service to allow flyout registration

ps48:
### Description

Enhance query action service to allow flyout registration

### Issues Resolved

- Allow external plugins to register flyouts action query action button in explore UI 
- Add support for `queryInEditor` for un-executed in explore, to be exported for query action consumers
- Export `getQueryWithSource` for plugins to consume the `queryInEditor` with/without source details in query

## Screenshot

https://github.com/user-attachments/assets/5c9fe0e3-fa4c-45ca-a2a6-4fc0139b5ff3

## Changelog

- feat: Enhance query action service to allow flyout registration

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/docs/plugins/explore/query-panel-actions.md
+++ b/docs/plugins/explore/query-panel-actions.md
@@ -46,7 +46,7 @@ export interface FlyoutActionConfig {
   getLabel(deps: QueryPanelActionDependencies): string;
   getIcon?(deps: QueryPanelActionDependencies): IconType;
   component: React.ComponentType<FlyoutComponentProps>;
-  onOpen?(deps: QueryPanelActionDependencies): void;
+  onFlyoutOpen?(deps: QueryPanelActionDependencies): void;
 }
 
 type QueryPanelActionConfig = ButtonActionConfig | FlyoutActionConfig;
@@ -75,8 +75,10 @@ export interface QueryPanelActionDependencies {
 }
 ```
 
+**Important:** Both `query.query` and `queryInEditor` are **already transformed** with the `source = <dataset>` clause added by the explore plugin. External plugins receive ready-to-execute queries and do NOT need to perform any transformation.
+
 **Note:** The `query` object includes:
-- `query.query`: The query string
+- `query.query`: The query string (pre-transformed with source clause)
 - `query.language`: The query language (PPL, SQL, etc.)
 - `query.dataset`: The selected dataset/index pattern (includes title, dataSource, signalType, etc.)
 
@@ -132,6 +134,7 @@ const CreateMonitorFlyout: React.FC<FlyoutComponentProps> = ({
         </EuiTitle>
       </EuiFlyoutHeader>
       <EuiFlyoutBody>
+        {/* queryInEditor already includes source clause, ready to execute */}
         <p>Query: {dependencies.queryInEditor}</p>
         <p>Dataset: {dependencies.query.dataset?.title}</p>
         <p>Language: {dependencies.query.language}</p>
@@ -154,7 +157,7 @@ export class ExamplePlugin {
       getLabel: () => 'Create monitor (flyout)',
       getIcon: () => 'bell',
       component: CreateMonitorFlyout, // Pass your flyout component
-      onOpen: (deps) => {
+      onFlyoutOpen: (deps) => {
         // Optional: Log or perform actions when flyout opens
         console.log('Opening create monitor flyout');
       }
--- a/src/plugins/explore/public/types.ts
+++ b/src/plugins/explore/public/types.ts
@@ -21,7 +21,6 @@ import {
   IndexPatternsContract,
   FilterManager,
   TimefilterContract,
-  Query,
 } from 'src/plugins/data/public';
 import { EmbeddableSetup, EmbeddableStart } from 'src/plugins/embeddable/public';
 import { DashboardSetup, DashboardStart } from 'src/plugins/dashboard/public';
@@ -55,7 +54,6 @@ import {
   QueryPanelActionsRegistryService,
   QueryPanelActionsRegistryServiceSetup,
 } from './services/query_panel_actions_registry';
-import { QueryWithQueryAsString } from './application/utils/languages';
 import { ChatPluginStart } from '../../chat/public';
 
 // ============================================================================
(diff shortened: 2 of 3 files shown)


src/plugins/explore/public/types.ts
```diff
@@ -75,6 +77,7 @@ export interface ExplorePluginStart {
   urlGenerator?: UrlGeneratorContract<'EXPLORE_APP_URL_GENERATOR'>;
   savedSearchLoader: SavedExploreLoader;
   savedExploreLoader: SavedExploreLoader;
+  getQueryWithSource: (query: Query) => QueryWithQueryAsString;
```
angle943: i'm very hesistant with this change, and don't know what the right path forward is.

Currently the `optional source` is a explore-specific thing that arose out of product requirement. Other pages (including old discover) don't have this concept. 

Because currently it is a explore-specific feature, i don't think other plugins should be inundated with this concept, and so i think explore should prepare the query before passing it onto other plugins.

But i am not sure if we want to keep this 'optional source' as a explore-specific thing moving forward. Ideally I think the backend PPL service should handle this themselves, and we should be able to pass the source information as a separate parameter.

but since that is not the case now, i think we should keep this as a explore-specific rthing. Lets not expose this function, and lets handle it before passing the query state to other plugins

src/plugins/explore/public/application/utils/state_management/slices/query_editor/query_editor_slice.ts
```diff
@@ -46,6 +47,7 @@ const initialState: QueryEditorSliceState = {
   dateRange: undefined,
   isQueryEditorDirty: false,
   hasUserInitiatedQuery: false,
+  currentEditorQuery: '',
```
angle943: we intentionally took out the editor query state out of redux due to performance issues. if you originally did this to get the current editor text, i suggest `useEditorText()`: https://github.com/angle943/OpenSearch-Dashboards/blob/8eff1b6ec32de66ce1c29cb0b79c3f7bccb4e200/src/plugins/explore/public/application/hooks/editor_hooks/use_editor_text/use_editor_text.ts

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_actions/query_panel_actions.tsx
```diff
@@ -3,78 +3,120 @@
...
+    },
+    [dependencies]
   );
 
+  // Get active flyout config
+  const activeFlyout = useMemo(() => {
+    if (!activeFlyoutId) return null;
+    const action = registry.getAction(activeFlyoutId);
+    return action?.actionType === 'flyout' ? (action as FlyoutActionConfig) : null;
+  }, [activeFlyoutId, registry]);
```
angle943: https://oui.opensearch.org/1.21/#/layout/flyout

It looks like for flyouts, we need to keep an `isOpen` state.

Looking at your implementation here, it looks like the external plugins will maintain that state, whereas this component maintains the 'activeFlyout' state.

But if, for whatever reason, the external plugin set their local `isOpen` state to false, this component has no way of knowing that, and there does not seem to be logic to close the flyout.

I agree that this component needs to maintain `activeFlyoutId` state, and I also think its fine for the external plugins to maintain the `isOpen` state.

But perhaps whats missing here, and would lead to less bugs down the road, is to keep both a `activeFlyout` computed variable (like you did here), but also introduce a new `activeFlyoutIsOpen` computed variable based on the state of the external plugin's `isOpen` state.

Looking at the contract we give to the external plugins, it looks like we are only exposing the `onOpen` callback, but not necessary the `isOpen` state. So this component does not have a way of knowing the open state of the external plugin's state. I think we should include the `isOpen` state and introduce the `activeFlyoutIsOpen` computed variable here, and use that to conditionally render the flyout

src/plugins/explore/public/types.ts
ps48: I agree, removed this export `getQueryWithSource` as this is an explore internal function. The current change now is that the `queryInEditor` is already transformed. [8a0dafb](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10849/commits/8a0dafb99d2d6a96040211d4e15f5fea3d07e260)

src/plugins/explore/public/application/utils/state_management/slices/query_editor/query_editor_slice.ts
ps48: Reverted this change, using now `useEditorText` to get the current query  [c984e86](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10849/commits/c984e864a83e320b3766b2dad83a52b315858e66)

src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_actions/query_panel_actions.tsx
ps48: Based on the current implementation the `activeFlyoutState` is always in the discover code. This is the source of truth for all flyouts registered. So any plugin registering their plugin will always use `closeFlyout()` callback function. Updated the naming of the variables and state for better readability  https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10849/commits/8a0dafb99d2d6a96040211d4e15f5fea3d07e260

docs/plugins/explore/query-panel-actions.md
```diff
@@ -16,77 +16,152 @@ Within the [Explore Plugin's setup method](https://github.com/opensearch-project
...
+    <EuiFlyout onClose={closeFlyout} size="m">
+      <EuiFlyoutHeader hasBorder>
+        <EuiTitle size="m">
+          <h2>Create Monitor</h2>
+        </EuiTitle>
+      </EuiFlyoutHeader>
+      <EuiFlyoutBody>
+        {/* queryInEditor already includes source clause, ready to execute */}
+        <p>Query: {dependencies.queryInEditor}</p>
+        <p>Dataset: {dependencies.query.dataset?.title}</p>
```
ruchidh, line 139: null check with query not required , is it handled already at prop level ?
dependencies.query?.dataset?

ruchidh (approved):

docs/plugins/explore/query-panel-actions.md
ps48: This just some sample plugin code in md file. So plugin implementation needs to take care of null checks.

PR #10858, opened 2025-11-04
Fix Add to Dashboard Modal showing only Last 20 Dashboards

LDrago27:
### Description

The Add to Dashboard Modal only shows 20 Dashboards in the drop down. This is because of the following 
```
        const res = await savedObjectsClient.find({
          type: 'dashboard',
        });
``` 
Without the pagination, it only shows 20 results due to the default value of perpage optional parameter. 

This PR updates the Selection to use a EuiComboBox. Where by default we show the last 20 Dashboards in dropdown but if you type in Search Term it now shows all the Dashboards matching the search term. 

## Testing the changes

https://github.com/user-attachments/assets/e77a3f1c-70a3-4af7-a344-57a75d27ac6a

## Changelog

- fix: Add to Dashboard Modal showing only Last 20 Dashboards

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/add_to_dashboard_modal.tsx
+++ b/src/plugins/explore/public/components/visualizations/add_to_dashboard_modal.tsx
@@ -20,11 +20,12 @@ import {
   EuiRadio,
 } from '@elastic/eui';
 import { FormattedMessage } from '@osd/i18n/react';
-import React, { useEffect, useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import { SavedObjectsClientContract } from 'src/core/public';
 import { DebouncedFieldText } from './style_panel/utils';
-import { OnSaveProps, DashboardInterface } from './add_to_dashboard_button';
+import { OnSaveProps } from './add_to_dashboard_button';
 import { useSavedExplore } from '../../application/utils/hooks/use_saved_explore';
+import { useExistingDashboard } from '../../application/utils/hooks/use_existing_dashboard';
 
 interface AddToDashboardModalProps {
   savedObjectsClient: SavedObjectsClientContract;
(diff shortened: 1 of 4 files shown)


TackAdam (dismissed):

src/plugins/explore/public/components/visualizations/add_to_dashboard_modal.tsx
```diff
@@ -58,14 +61,50 @@ export const AddToDashboardModal: React.FC<AddToDashboardModalProps> = ({
     ((selectedOption === 'existing' && selectDashboard) ||
       (selectedOption === 'new' && newDashboardName));
 
-  useEffect(() => {
-    const loadAllDashboards = async () => {
```
d-buckner, line 62: This all looks like it will work great, but this modal is looking like a catch all for all sorts of logic which is causing some readability/maintenance issues.

I feel that if we implement this new functionality in a new hook (ex. `useDashboardSearch`), we can prevent exacerbating the readability issue here. That hook can be responsible for all of this business logic and provide us with really readable and easy to use functionality that others can benefit from too.

Ex.
```
const dashboardSearch = useDashboardSearch();

...
{dashboardSearch.isLoading && <...}

LDrago27: @d-buckner Actually That makes sense here. Abstracting away the fetching of dashboards from the view component will definitely simplify the component. 

I will update the PR to refactor the logic to fetch dashboards into a separate hook.

ruanyl:
@LDrago27 Thanks for fixing the issue, could you fix the failed cypress tests as well introduced by this PR? https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/19087876531/job/54604061315

PR #10860, opened 2025-11-05
Initialize workers from endpoint rather than blob

d-buckner:
### Description
Changes worker initialization from blob-based inline source to URL-based loading with cache-busting support. Workers are now served as static assets from the server with build hash-based URLs.

Key changes:
- Remove worker_store system for inline worker registration
- Remove raw-loader dependency and usage
- Add setBuildHash() to configure worker URL generation
- Centralize worker creation in monaco_environment
- Update PPL and xjson language modules to use new system

## Changelog
- feat: load worker bundles from url

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -25516,7 +25516,7 @@ string-length@^4.0.1:
     char-regex "^1.0.2"
     strip-ansi "^6.0.0"
 
-"string-width-cjs@npm:string-width@^4.2.0":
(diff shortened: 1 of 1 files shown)


packages/osd-monaco/package.json
```diff
@@ -22,7 +22,6 @@
     "css-loader": "^5.2.7",
     "del": "^6.1.1",
     "file-loader": "^6.2.0",
-    "raw-loader": "^4.0.2",
```
angle943, line 25: i gain pleasure when i see red lines in package.json dependencies

d-buckner: I'm glad I'm in good company

PR #10886, opened 2025-11-08
[Draft] Fix index pattern selection

TackAdam:
### Description
Current issue is there is no pagination so we only show the first 100 results. This change allows async searching as the user types. But we should combine the sections instead of trying to maintain index / index wildcard. 
Example below

Currently we have selection between index and index wildcard
Screenshot 2025-11-07 at 4 11 19 PM

We should aim to use the better UI instead of drop down options ie move towards:
https://playground.opensearch.org/app/indexPatterns/create
Screenshot 2025-11-07 at 4 10 55 PM

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


Maosaic:
Thanks Adam for making this change, I think the cypress test need to be updated since the "select scope by" option is removed.

PR #10895, opened 2025-11-10
[chat]Persist chatbot state in local storage

wanglam:
### Description

This change introduces persistent state management for the chatbot window using browser localStorage. The implementation extends the ChatService to track comprehensive window state including open/closed status, layout mode (sidecar or fullscreen), and padding size, automatically persisting changes to localStorage and restoring the chatbot to its previous configuration on application start. 

## Changelog

- feat: Persist chatbot state in local storage

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/10888.yml
+++ b/changelogs/fragments/10888.yml
@@ -1,0 +1,2 @@
+fix:
+- Explore embeddable not render due to explore services not set ([#10888](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10888))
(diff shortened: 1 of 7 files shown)


d-buckner:
How are we handling invalidation if a user logs out and signs in to another user? Or are we thinking about these preferences as client preferences instead of user preferences?

wanglam:
> How are we handling invalidation if a user logs out and signs in to another user? Or are we thinking about these preferences as client preferences instead of user preferences?

Yes. For now it's a client preferences. It won't be cleared after user logout. This is an existing feature in the [dashboards assistant repo](https://github.com/opensearch-project/dashboards-assistant/blob/main/public/utils/persisted_chatbot_state.ts). I just sync back to the OSD chat plugin to keep the two chatbot have the same behavior.

Issue #10907, opened 2025-11-12
[BUG] workspace collaborator ID not trimmed cause unexpected behavior


gaobinlong:
**Describe the bug**

For now, we don't trim the collaborator ID when adding collaborators for workspace, the problem is that the read ID doesn't contain any leading whitespace or trailing whitespace, so the acl permission control doesn't work.

**To Reproduce**
1. Go to 'Collaborators' inside a workspace
2. Click on 'Add collaborators'
3. Add a user with leading whitespace.

**Expected behavior**
The leading whitespace or trailing whitespace could be removed automatically.

**OpenSearch Version**
3.3.0

**Dashboards Version**
Please list the version of OpenSearch Dashboards being used.
3.3.0
**Plugins**

Please list all plugins currently enabled.

**Screenshots**

<img width="824" height="315" alt="Image" src="https://github.com/user-attachments/assets/d3402c1f-1d5c-43d7-bdae-32d10c2713fc" />

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

PR #10908, opened 2025-11-12
[Workspace] Trim collaboratorId when adding collaborators for workspace

gaobinlong:
### Description

Trim collaborator ID when adding collaborators for workspace, if the leading whitespace or trailing whitespace are not trimmed, the collaborator may not work.

### Issues Resolved
Resolves https://github.com/opensearch-project/OpenSearch-Dashboards/issues/10907

## Screenshot

image

## Testing the changes
1. In the modal of adding collaborators, input some ID with leading whitespace or trailing whitespace, these spaces will be automatically remove.
2. Check the collaborator ID in the .kibana index, any whitespace in prefix or suffix are removed.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/06/rule_matching_vis_table.spec.js
+++ b/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/06/rule_matching_vis_table.spec.js
@@ -39,7 +39,7 @@ export const runCreateVisTests = () => {
 
     it('should create a table visualization using a query with multiple fields', () => {
       const query = `source=${datasetName} | stats count() by category, response_time | head 10`;
-      cy.explore.createVisualizationWithQuery(query, 'Table', datasetName, {
+      cy.explore.createVisualizationWithQuery(query, 'table', datasetName, {
         shouldManualSelectChartType: true,
       });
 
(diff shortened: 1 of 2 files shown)


SuZhou-Joe:
Wondering if we should trim the id in server side.

src/plugins/workspace/public/components/add_collaborators_modal/workspace_collaborator_input.tsx
```diff
@@ -53,7 +53,7 @@ export const WorkspaceCollaboratorInput = ({
 }: WorkspaceCollaboratorInputProps) => {
   const handleCollaboratorIdChange = useCallback(
     (e) => {
-      onCollaboratorIdChange(e.target.value, index);
+      onCollaboratorIdChange(e.target.value?.trim(), index);
```
Hailong-am: is there a case a collaborator id contains space inside like `first last`? with this, user may not able to input space in the middle

gaobinlong: At a second thought, trim the input is not necessary, only trim the new added collaborators in onAddCollaborators() in enough. And even though in most cases userId or groupId doesn't contain whitespace in the middle, but we should provide some flexibility for users. Reverted this change yet.

gaobinlong:
> Wondering if we should trim the id in server side.

At first time, I want to implement this in server side, but found that we cannot differentiate new added collaborators and existing collaborators because the update workspace API only supports full update, not partial update, this will cause an issue that any existing collaboratorIds containing leading whitespace or trailing whitespace will be trimmed, this maybe a breaking change I think.

Hailong-am:
ci group 16 failure not related to this change.

PR #10909, opened 2025-11-12
[Explore Vis]Simplify threshold logic

Qxisylolo:
### Description
this pr simplifies threshold logic

## Changelog
- feat: Simplify threshold logic

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced threshold handling and color consistency across gauge and metric visualizations.
  * Improved support for undefined values and edge cases in threshold processing.

* **Tests**
  * Added comprehensive unit tests for threshold generation, filtering, range handling, and deduplication.


src/plugins/explore/public/components/visualizations/metric/to_expression.ts
```diff
@@ -63,24 +60,23 @@ export const createSingleMetric = (
...
-
   function targetFillColor(
     useThresholdColor: boolean,
     threshold?: Threshold[],
     baseColor?: string
   ) {
     const newThreshold = threshold ?? [];
 
-    const newBaseColor = baseColor ?? getColors().statusGreen;
+    if (calculatedValue === undefined || calculatedValue < minBase) {
```
ruanyl: It seems we don't consider `maxBase`, I'm just thinking shall we also not consider `minBase` to make it consistent? so just make the text color solely based on thresholds.

src/plugins/explore/public/components/visualizations/bar_gauge/bar_gauge_utils.ts
```diff
@@ -157,3 +142,68 @@ export const generateParams = (
...
+  }
+
+  if (lastBelowIndex !== -1) {
+    result.splice(0, lastBelowIndex);
+    result[0] = { ...result[0], value: minBase };
+  } else {
+    result.unshift({ value: minBase, color: defaultColor });
+  }
+
+  const valueResults: Threshold[] = [];
```
ruanyl: I would create another function to calculate the `valueResults` separately

ruanyl (dismissed):

src/plugins/explore/public/components/visualizations/metric/to_expression.ts
Qxisylolo: sure, updated

src/plugins/explore/public/components/visualizations/bar_gauge/bar_gauge_utils.ts
```diff
@@ -157,3 +142,75 @@ export const generateParams = (
...
+export const generateThresholds = (
+  minBase: number,
+  maxBase: number,
+  thresholds: Threshold[],
+  baseColor: string | undefined
+) => {
+  const defaultColor = baseColor ?? getColors().statusGreen;
+
+  // sort thresholds by value and dedupe threshold by value
+  thresholds = thresholds
```
wanglam, line 155: Nit: can be refactor with code below:
```tsx
thresholds = Array.from(
  thresholds.reduce((map, threshold) => {
    map.set(threshold.value, threshold);
    return map;
  }, new Map<number, Threshold>()).values()
)
.sort((t1, t2) => t1.value - t2.value);
```
It seems more clear to me.

src/plugins/explore/public/components/visualizations/bar_gauge/bar_gauge_utils.ts
```diff
@@ -157,3 +142,75 @@ export const generateParams = (
...
+      while (
+        thresholdIndex < thresholds.length - 1 &&
+        thresholds[thresholdIndex + 1].value <= stop
+      ) {
+        thresholdIndex++;
+      }
+
+      // Add valid threshold for this stop
+      if (thresholds[thresholdIndex].value <= stop) {
+        valueThresholds.push({ value: stop, color: thresholds[thresholdIndex].color });
```
wanglam, line 211: Do we need to return here if thresholdIndex equal `thresholds.length - 1`? 
Imagine the input parameters like below:
minBase = 0; maxBase = 100; valueStops = [10,20,30,40,50]; thresholds = [{ value: 15, color: "green" }];
The result will be:
```
[{"value":20,"color":"green"},{"value":30,"color":"green"},{"value":40,"color":"green"},{"value":50,"color":"green"}]
```
I'm not sure if it's meet our requirements.

Qxisylolo: Yes, this is intended, if all the value stops are beyond the last threshold, then we will use that threshold to create value thresholds, so that, the gradient mode in different bar will be consist.

Issue #10811, opened 2025-10-27
[BUG]Dashboard filters persist across index patterns when switching from Saved Objects to Discover view (OpenSearch 2.19 / 3.1)


ruchidh:
### Description
We have observed an issue in OpenSearch Dashboards where saved query filters persist across different index patterns when navigating between views. This behavior appears to be a regression introduced in OpenSearch versions 2.19 and 3.1, as it was not present in version 2.17.

### Steps to Reproduce
1. Create a saved search with query filter in discover, save as `search-1`
2. Create another search without query filter using different index pattern in discover, save as `search-2`
3. Access `search-1` discover dashboard through Saved objects
4. Navigate back to dashboard, then open `search-2` through Saved objects
5. Observe that filter from `search-1` still persists in discover view

### Current Behavior
- Query filters persist across different index patterns when navigating between saved searches
- Filter from first search remains active even when switching to a different saved search

### Expected Behavior
- Query filters should not persist when switching between different saved searches
- Each saved search should maintain its own independent filter state

### Analysis
- Successfully reproduced the issue following the above steps
- Behavior confirmed in OpenSearch 2.19 and 3.1
- Issue not present in version 2.17

### Environment
- OpenSearch Version: 2.19/3.1

### Additional Context
Goal: Understand whether this behavior is expected for OS 2.19

### Labels
- type: bug
- component: opensearch-dashboards
- version: 2.19
- severity: medium

ruchidh:
<img width="1024" height="404" alt="Image" src="https://github.com/user-attachments/assets/288efcc8-f374-481a-9142-14b636ded8db" />

Maosaic:
It will need further deep dive, but it might related to how we load saved search and this change we made https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9529

andrross:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/andrross) [2](https://github.com/cwperks)

Issue #10911, opened 2025-11-12
[BUG] "New" discover does not clear out query bar filters


paulstn:
**Describe the bug**

In Discover, when the "New" Discover nav bar header button is clicked, the filters are not cleared out.

(Before clicking "New")
<img width="1033" height="262" alt="Image" src="https://github.com/user-attachments/assets/da3b459c-1fc8-4fbb-a4f3-30fbf6a32700" />

(After clicking "New")
<img width="989" height="208" alt="Image" src="https://github.com/user-attachments/assets/a0c14d47-0c29-49a1-8f8a-f6323bd02300" />

**To Reproduce**
1. Go to Discover with DQL enabled
2. Add a filter through the filter bar
3. Click on the "New" button (plus shaped nav header link)

There is a workaround: When opening the filter options, there is an option to "Remove all" filters:

<img width="404" height="373" alt="Image" src="https://github.com/user-attachments/assets/d68eabff-6eee-47d1-ad30-e2edea37859a" />

**Expected behavior**
The filter bar gets cleared out. 

**Dashboards Version**
2.19 -> main (3.3)

paulstn:
Fixed with https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10913

PR #10913, opened 2025-11-13
[Discover] Skip loading filter if navigating to a saved search without params

Maosaic:
### Description

When user click on saved search on Assets page, it will navigate the user to Discover page and load the saved search. However, if user had visited Discover before in same session, Discover will automatcally append the filter state from the previous visit. 

This PR allow skip loading previous filter, when user visit Discover page with URL contains saved search without params.

### Issues Resolved

fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/10811

## Screenshot

Before:

https://github.com/user-attachments/assets/0c23219f-69b0-40ab-9ef6-633be71607dc

After:

https://github.com/user-attachments/assets/18c58b13-c813-4696-b426-ac8d4f810821

## Changelog

- fix: Skip loading filter if navigating to a saved search without params

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/state_sync/connect_to_query_state.ts
+++ b/src/plugins/data/public/query/state_sync/connect_to_query_state.ts
@@ -77,7 +77,7 @@ export const connectStorageToQueryState = (
       syncKeys.push('appFilters');
     }
 
-    const initialStateFromURL: QueryState = osdUrlStateStorage.get('_q') ?? {
+    const initialState: QueryState = osdUrlStateStorage.get('_q') ?? {
       query: queryString.getDefaultQuery(),
       // If caller specifies to skip filters from memory, use empty array
       filters: syncConfig.skipAppFiltersFromMemory ? [] : filterManager.getAppFilters(),
@@ -85,7 +85,7 @@ export const connectStorageToQueryState = (
 
     if (!osdUrlStateStorage.get('_q')) {
       // set up initial '_q' flag in the URL to sync query and filter changes
-      osdUrlStateStorage.set('_q', initialStateFromURL, {
+      osdUrlStateStorage.set('_q', initialState, {
         replace: true,
       });
       // clear existing query and apply default query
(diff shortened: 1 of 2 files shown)


src/plugins/discover/public/application/view_components/canvas/top_nav.tsx
```diff
@@ -77,9 +77,16 @@ export const TopNav = ({ opts, showSaveQuery, isEnhancementsEnabled }: TopNavPro
     : [];
 
   const syncConfig = useMemo(() => {
+    // Check if navigating to a saved search without params - skip filter persistence
+    const currentHash = window.location.hash.split('?')[0];
+    const isSavedSearchUrl = currentHash.includes('#/view/');
+    const hasUrlParams = window.location.hash.includes('?');
+    const shouldSkipFilters = isSavedSearchUrl && !hasUrlParams;
```
ruanyl: Didn't get why it should check `hasUrlParams` here, could you elaborate please?

src/plugins/data/public/query/state_sync/connect_to_query_state.ts
```diff
@@ -74,7 +79,8 @@ export const connectStorageToQueryState = (
 
     const initialStateFromURL: QueryState = osdUrlStateStorage.get('_q') ?? {
```
ruanyl: maybe unrelated to this PR: the name initialStateFromURL is a bit misleading, because it's not just url state, it also reads in-memory state

src/plugins/data/public/query/state_sync/connect_to_query_state.ts
ruanyl, line 1: Maybe unrelated to this PR, it seems in `connectStorageToQueryState`, we're doing two-way state syncing `memory state <-> url sate`, do you know the intention?

src/plugins/discover/public/application/view_components/canvas/top_nav.tsx
Maosaic: Agree, I think we don't need this check here, I was also confused by the naming of `initialStateFromURL`, it seems it will prioritize URL params in `connectStorageToQueryState`, so we don't need params check here.

paulstn:
If possible, would you be able to fix the "open" behavior within this PR?: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/10911

src/plugins/data/public/query/state_sync/connect_to_query_state.ts
Maosaic: Just by reading the code, I think `connectStorageToQueryState` does 2 things:
1. When the app(like Discover) initialize, read url state and update memory state
2. After initialize, update url state if memory state get updated

src/plugins/data/public/query/state_sync/connect_to_query_state.test.ts
```diff
@@ -215,6 +215,108 @@ describe('connect_storage_to_query_state', () => {
     const updatedStorage = osdUrlStateStorage.get('_q');
     expect(previousStorage).not.toStrictEqual(updatedStorage);
   });
+
+  test('when skipAppFiltersFromMemory is true, state initializes with empty filters even if filterManager has app filters', () => {
```
abbyhu2000, line 219: Should we also add a test case for`skipAppFiltersFromMemory` with some global filters? I assume `skipAppFiltersFromMemory` only skip app filters but not global filters right? @Maosaic

Maosaic: yes, `connectStorageToQueryState` only enteract with `_q`

PR #10915, opened 2025-11-14
Avoids URIError in `@osd/std` url methods

paulstn:
### Description

Urls, when being parsed through any functionality within OSD that uses the @osd/std package’s methods for `modifyUrl`, `isRelativeUrl`, or `getUrlOrigin`, eventually go on to use the `parse` method from the `node:url` package. 

When the URL is malformed, the `parse` method will throw a `URIError`, which is not handled by the current code. This means that malformed URLs can crash the page.
An exmaple service that uses modifyUrl that causes the application to crash:
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/http/base_path_service.ts#L87

I've added tests that would've previously passed in user input that can cause crashes, and changed the `url` file's methods to catch errors and return values that make sense to default to. 

The example malicious URL that I've written into tests has a malformed `pass` in the `auth` section of the url, since that is where the `node:url`'s `parse` method would explicitly throw a URIError. 
```
http://user:%E0%A4@example.com
```

## Changelog

- fix: @osd-std package url parse method does not throw page crashing errors

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


packages/osd-std/src/url.ts
```diff
@@ -122,28 +126,36 @@ export function modifyUrl(
...
-  if (!obj.protocol && !obj.hostname) {
-    return null;
+  try {
+    const obj = parseUrl(url);
+    if (!obj.protocol && !obj.hostname) {
+      return null;
+    }
+    return `${obj.protocol}//${obj.hostname}${obj.port ? `:${obj.port}` : ''}`;
+  } catch (error) {
+    return url; // Safe fallback to original url
```
ZilongX: as commented  in L149 - "Returns the origin (protocol + host + port) from given `url` if `url` is a valid absolute url, or null otherwise" it should return null when error is caught since it's likely a malformed url

ZilongX:
generally looking good with some comments, thanks @paulstn for the quick fix. Meanwhile there are a bunch of checking test failing for this PR (6+) which doesn't seem to be related the changes introduced in, is it common all PRs ? Then we should fix all those failed test cases

packages/osd-std/src/url.ts
ruanyl: Agree with @ZilongX, should better to return `null` if the url failed to parse

Flyingliuhub: agree, we should better handle here

paulstn: This does sound better - I'll make the change.

paulstn:
> generally looking good with some comments, thanks @paulstn for the quick fix. Meanwhile there are a bunch of checking test failing for this PR (6+) which doesn't seem to be related the changes introduced in, is it common all PRs ? Then we should fix all those failed test cases

They seemed to be unrelated, a test rerun has them all passing now.

Edit: They seem to be related to something else that is causing every PR to fail.

ruanyl (approved):

angle943 (approved):

PR #10933, opened 2025-11-18
Clear out filter bar on "New" Discover action

paulstn:
### Description

 When users click the "New" button in Discover to start a fresh search, app-level filters
  from the previous view persist instead of being cleared. This prevents users from getting
  a clean slate, forcing them to manually remove filters. The query text clears correctly,
  but filters do not.

  Expected behavior: Clicking "New" should provide a completely fresh Discover view with no
  filters or query, identical to opening Discover for the first time.

  Actual behavior: Query clears as expected, but filters remain from the previous view.

  Root Cause

  A previous change (PR #9529) modified the filter initialization logic to preserve existing
   filters when loading saved searches, which was necessary to correctly handle snapshot
  URLs. However, this change inadvertently affected the "New" button flow as well. When
  creating a new search (no saved search ID), the code was still merging in existing filters
   from the filter manager instead of starting fresh.

  Solution

  This fix adds a conditional check to distinguish between two scenarios:
  - Loading an existing saved search: Preserves and merges existing filters (maintains the
  snapshot URL fix)
  - Creating a new search: Starts with an empty filter array (restores the original "New"
  button behavior)

  The query already handled this distinction correctly, so this fix aligns the filter
  behavior with the query behavior.

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/10911

## Screenshot

Showing the filters clearing out upon clicking "New" Discover action:

https://github.com/user-attachments/assets/01b1ce98-ff52-4c24-a28a-5fd74102156b

Showing the original fix to correctly load saved search from snapshot url works as expected:

https://github.com/user-attachments/assets/39c59b61-f8f8-4c1b-9a68-2ae7f3f68648

## Testing the changes

  - Added a new test case to verify filters are cleared when creating a new search
  - All existing tests pass, confirming backward compatibility
  - Manual testing confirms "New" button now properly clears both query and filters

## Changelog

- fix: Clears out filter bar on "New" Discover action

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/test/common/config.js
+++ b/test/common/config.js
@@ -44,7 +44,17 @@ export default function () {
     opensearchTestCluster: {
       license: 'oss',
       from: 'snapshot',
-      serverArgs: ['search.concurrent_segment_search.mode=none'],
+      serverArgs: [
+        'search.concurrent_segment_search.mode=none',
+        // Disable disk-based shard allocation to prevent index creation blocks in CI
+        'cluster.routing.allocation.disk.threshold_enabled=false',
+        // Set very low disk watermarks for testing
+        'cluster.routing.allocation.disk.watermark.low=1gb',
+        'cluster.routing.allocation.disk.watermark.high=500mb',
+        'cluster.routing.allocation.disk.watermark.flood_stage=100mb',
+        // Disable read-only index block when disk space is low
+        'cluster.blocks.read_only_allow_delete=false',
+      ],
     },
 
     osdTestServer: {


Maosaic:
I wonder if snapshot URL without saved search will break with this change

paulstn:
> I wonder if snapshot URL without saved search will break with this change

It does break, we need to handle the fix some other way.

PR #10934, opened 2025-11-18
[chat]Add getThreadId$ in chat service

wanglam:
### Description

Expose observable thread id in chat service, so other plugins can observe this value to do some actions after thread updates.

## Changelog

- feat: [chat]Add getThreadId$ in chat service

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/test/common/config.js
+++ b/test/common/config.js
@@ -44,7 +44,17 @@ export default function () {
     opensearchTestCluster: {
       license: 'oss',
       from: 'snapshot',
-      serverArgs: ['search.concurrent_segment_search.mode=none'],
+      serverArgs: [
+        'search.concurrent_segment_search.mode=none',
+        // Disable disk-based shard allocation to prevent index creation blocks in CI
+        'cluster.routing.allocation.disk.threshold_enabled=false',
+        // Set very low disk watermarks for testing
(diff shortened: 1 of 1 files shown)


src/plugins/chat/public/services/chat_service.ts
```diff
@@ -55,20 +55,30 @@ export class ChatService {
   // ChatWindow ref for delegating sendMessage calls to proper timeline management
   private chatWindowRef: React.RefObject<ChatWindowInstance> | null = null;
 
+  private get threadId() {
+    return this.threadId$.getValue();
+  }
+
   constructor() {
```
Hailong-am, line 62: what's the use case for plugins to get current chat thread id? this sounds like a internal stuff inside the chat.

wanglam: This is a private method only for chat service internal use. We already expose one getThreadId before. It will be used to retrieve a conversation id for suggested actions service. Currently, we add a new observable thread id to let plugin detect the thread changes. One use case would be investigation plugin, since the chatbot will clear the dynamic context when start new thread. Investigation plugin can rebuilt these context by observe the thread id updates.

PR #10943, opened 2025-11-20
fix(QueryStringManager): fix Could not subscribe to application.currentAppId$

joshuali925:
### Description

Partially revert #9466. Current implementation doesn't work because QueryStringManager is constructed during setup, and setApplication happens during start. In playground or any instance we see `Could not subscribe to application.currentAppId$` in console warning.

## Changelog

- fix: fix Could not subscribe to application.currentAppId$

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


joshuali925:
this revert is not a good fix, because `getCurrentAppId` called during setup still cannot get appId. It only fixes when calls are after setup. Closing for now

PR #10959, opened 2025-11-21
Integrate dynamic config to disable global settings updates for non-admin users

Qxisylolo:
### Description
This pr integrates dynamic config to disable global settings updates for non-admin users

Issue https://github.com/opensearch-project/OpenSearch-Dashboards/issues/10960

## Testing the changes

Add the `uiSettings.globalScopeEditable.enabled` setting in the YAML configuration. (If this value cannot be found in the config store, the dynamic config service will fall back to reading it from opensearch_dashboards.yml).

```
uiSettings.globalScopeEditable.enabled
```

1. when uiSettings.globalScopeEditable.enabled is false and user is non admin

Global scope ui settings are not able to edit

截屏2025-11-21 21 39 56

2. When a user tries to edit global-scope UI settings in the OSD frontend, all settings in the application settings become non-editable.

截屏2025-11-21 21 41 09

## Changelog
- feat: Integrate dynamic config to disable global settings updates for non-admin users

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

## Summary by CodeRabbit

* **New Features**
  * Dynamic control to restrict who can edit global UI settings based on runtime config and user role.
  * Advanced Settings now respects dashboard-admin scope and dynamic global-scope editability.

* **Configuration**
  * Added `globalScopeEditable` option (runtime-controllable) to govern global UI settings editability.

* **Tests**
  * Added comprehensive unit tests covering permission branches, update/bulkUpdate paths, and error cases for the permission wrapper.


src/plugins/advanced_settings/public/management_app/advanced_settings.tsx
```diff
@@ -204,6 +208,7 @@ export class AdvancedSettingsComponent extends Component<
           isPermissionControlled:
             all[setting[0]].scope === UiSettingScope.DASHBOARD_ADMIN && !isDashboardAdmin,
           userSettingsEnabled,
+          isGlobalScopeEditable,
```
FriedhelmWS: Is that possible to reuse the isPermissionControlled attribute here? I think for the field component it doesn't have to know it is disabled by permission in which way.

src/plugins/advanced_settings/public/management_app/components/field/field.tsx
```diff
@@ -365,7 +367,11 @@ export class Field extends PureComponent<FieldProps> {
               minLines={6}
               maxLines={30}
               isReadOnly={
-                isOverridden || isPermissionControlled || preferBrowserSetting || !enableSaving
+                isOverridden ||
+                isPermissionControlled ||
+                preferBrowserSetting ||
+                !enableSaving ||
+                !isGlobalScopeEditable
```
FriedhelmWS: Can we refactoring these into a constant variable so that we can reuse them.

src/core/server/ui_settings/saved_objects/dynamic_config_controlled_ui_settings_wrapper.ts
```diff
@@ -0,0 +1,120 @@
...
+    wrapperOptions: SavedObjectsClientWrapperOptions
+  ): Promise<boolean> {
+    // If saved object permission is disabled, everyone should be treated as admin here
+
+    const dynamicConfigServiceStart = await this.dynamicConfig.getStartService();
+    const store = dynamicConfigServiceStart.getAsyncLocalStore();
+    const client = dynamicConfigServiceStart.getClient();
+
+    try {
+      const dynamicConfig = await client.getConfig(
```
FriedhelmWS, line 70: Is that possible to use the capabilities that we already registered? 🤔 Otherwise I think it would be better to cache the dynamic config result to prevent multiple asynchronized function calling every time we want to update the global settings saved object.

src/core/server/ui_settings/saved_objects/dynamic_config_controlled_ui_settings_wrapper.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  public wrapperFactory: SavedObjectsClientWrapperFactory = (wrapperOptions) => {
+    const updateUiSettingsWithPermission = async <T = unknown>(
+      type: string,
+      id: string,
+      attributes: Partial<T>,
+      options: SavedObjectsUpdateOptions = {}
+    ): Promise<SavedObjectsUpdateResponse<T>> => {
+      const hasPermission = await this.checkPermission(wrapperOptions);
+      if (type === 'config' && !hasPermission && isGlobalScope(id)) {
```
FriedhelmWS: I think this permission check needs to be after we confirmed the incoming saved object is config, otherwise we do the check for every saved object type, which is unnecessary 😉

src/core/server/ui_settings/utils.ts
```diff
@@ -15,6 +15,9 @@ export const PERMISSION_CONTROLLED_UI_SETTINGS_WRAPPER_ID = 'permission-control-
 // other wrappers, therefore the priorty can be any number and the priority of 100 is trival
 export const PERMISSION_CONTROLLED_UI_SETTINGS_WRAPPER_PRIORITY = 100;
 
+export const DYNAMIC_CONFIG_CONTROLLEDUI_SETTINGS_WRAPPER_ID = 'dynamic-config-control-ui-settings';
+export const DYNAMIC_CONFIG_CONTROLLEDUI_SETTINGS_WRAPPER_PRIORITY = 101;
```
FriedhelmWS: Not related to this PR, but we think we can have some documentations on the orders of all the wrappers 🙂

src/plugins/advanced_settings/public/management_app/components/field/field.tsx
Qxisylolo: will update, thanks

src/core/server/ui_settings/saved_objects/dynamic_config_controlled_ui_settings_wrapper.ts
Qxisylolo: The registered capabilities are used on the public side to retrieve dynamic config, and in this case they are consumed inside the server-side wrapper. Since CapabilitiesService is registered after the UiSettingsService, we will register capabilities in the Advanced Settings plugin.

I don’t think an additional cache is needed here for two reasons:
1. When should the cache be updated, since dynamic settings can be modified ?
2. The dynamic config service already maintains its own cache, so adding another layer of caching would not provide much benefit, i think, although it is an async call.

src/core/server/ui_settings/saved_objects/dynamic_config_controlled_ui_settings_wrapper.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Wrapper for reading dynamic feature flag to decide if global UI settings
+ * are editable or not for non-admin users
+ * @param isPermissionControlEnabled
+ * @param dynamicConfig
+ */
+export class DynamicConfigControlledUiSettingsWrapper {
+  /**
+   * @param isPermissionControlEnabled
```
yubonluo: I think we can skip pass the `isPermissionControlEnabled` param to the class, because `getWorkspaceState(wrapperOptions.request).isDashboardAdmin !== false` contains the situation where  permission control is disabled.

src/plugins/advanced_settings/public/management_app/advanced_settings.tsx
Qxisylolo: We could merge isPermissionControlled with isGlobalScopeEditable, but my concern is that they control different kinds of scopes.

src/core/server/ui_settings/saved_objects/dynamic_config_controlled_ui_settings_wrapper.ts
Qxisylolo: yes, updated. thanks

src/plugins/advanced_settings/public/management_app/advanced_settings.tsx
FriedhelmWS: Yep, but I was thinking for a UI component, it shouldn't care about what kind of permission control condition makes it disabled for the user edit and an additional parameter here seems a bit redundant. Though, this is quite trivial so feel free to keep it here.

src/core/server/ui_settings/saved_objects/dynamic_config_controlled_ui_settings_wrapper.ts
Qxisylolo: thanks!!

src/plugins/advanced_settings/public/management_app/advanced_settings.tsx
Qxisylolo: ok i c, will combine them, thanks!

src/core/server/ui_settings/saved_objects/dynamic_config_controlled_ui_settings_wrapper.ts
```diff
@@ -0,0 +1,97 @@
...
+    const client = dynamicConfigServiceStart.getClient();
+
+    try {
+      const dynamicConfig = await client.getConfig(
+        { pluginConfigPath: 'uiSettings' },
+        { asyncLocalStorageContext: store! }
+      );
+
+      //  If saved object permission is disabled, getWorkspaceState will return undefined
+      const isDashboardAdmin = getWorkspaceState(wrapperOptions.request).isDashboardAdmin !== false;
```
SuZhou-Joe: Shall we return true earlier when `isDashboardAdmin` is true?

Qxisylolo: sure. thanks

src/core/server/ui_settings/saved_objects/dynamic_config_controlled_ui_settings_wrapper.ts
```diff
@@ -0,0 +1,97 @@
...
+      deleteByWorkspace: wrapperOptions.client.deleteByWorkspace,
+    };
+  };
+
+  private async checkPermission(
+    wrapperOptions: SavedObjectsClientWrapperOptions
+  ): Promise<boolean> {
+    // If saved object permission is disabled, everyone should be treated as admin here
+
+    const dynamicConfigServiceStart = await this.dynamicConfig.getStartService();
```
yubonluo: Shall we switch these lines to try/catch block？

yubonluo (dismissed):

SuZhou-Joe (dismissed):

yubonluo (approved):

SuZhou-Joe (approved):

PR #10968, opened 2025-11-24
Create saved object for prometheus data-connection

joshuali925:
### Description

Prometheus will be stored as data-connection saved objects. Only connection id and MDS datasource id is saved.

```json
        {
          "data-connection": {
            "connectionId": "my_prometheus",
            "type": "Prometheus"
          },
          "type": "data-connection",
          "references": [
            {
              "id": "02ced1b0-c95f-11f0-a5f2-dfc60f85a7e3",
              "type": "data-source",
              "name": "dataSource"
            }
          ],
          "updated_at": "2025-11-24T18:06:37.234Z"
        }
```

Underlying they are still in SQL, but OSD will use saved objects API to get Prometheus metadata. Users should only use OSD to create/delete prometheus connections if they use UI.

Additionally, this PR supports "No Auth" option when creating prometheus connections.

## Screenshot

image

## Changelog

- feat: Create saved object for prometheus data-connection

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_data_connection/manage_direct_query_data_connections_table.tsx
```diff
@@ -391,7 +391,9 @@ export const ManageDirectQueryDataConnectionsTable = ({
 
         const indentStyle =
           featureFlagStatus &&
-          record.connectionType !== DataSourceConnectionType.OpenSearchConnection
+          record.connectionType !== DataSourceConnectionType.OpenSearchConnection &&
+          // Prometheus data-connections are shown on top level
+          record.type !== 'Prometheus'
```
ps48: Can we use `DataConnectionType.Prometheus` enum?

src/plugins/data_source/common/data_connections/data_connection_saved_object_attributes.ts
```diff
@@ -22,5 +22,6 @@ export interface DataConnectionSavedObjectAttributes extends SavedObjectAttribut
 export enum DataConnectionType {
   CloudWatch = 'AWS CloudWatch',
   SecurityLake = 'AWS Security Lake',
+  Prometheus = 'Prometheus',
```
ps48, line 25: Do we need Amazon Prometheus here?

src/plugins/data_source_management/server/routes/data_connections_router.ts
```diff
@@ -317,34 +339,63 @@ export function registerDataConnectionsRoute(router: IRouter, dataSourceEnabled:
...
       } catch (error: any) {
-        console.error('Issue in deleting data sources:', error);
         const statusCode = error.statusCode || error.body?.statusCode || 500;
-        const errorBody = error.body ||
-          error.response || { message: error.message || 'Unknown error occurred' };
 
+        if (statusCode !== 404) {
+          console.error('Issue in deleting data connection from backend:', error);
+          const errorBody = error.body ||
+            error.response || { message: error.message || 'Unknown error occurred' };
```
ps48, line 366: why is only 404 handled here?

src/plugins/data_source_management/server/routes/data_connections_router.ts
```diff
@@ -145,74 +146,95 @@ export function registerNonMdsDataConnectionsRoute(router: IRouter) {
...
+            },
+            {
+              references: dataSourceMDSId
+                ? [{ id: dataSourceMDSId, type: 'data-source', name: 'dataSource' }]
+                : [],
+            }
+          );
+        }
+
+        return response.ok({ body: dataConnectionsresponse });
```
ps48, line 230: May be we can check if ppl datasource was created successfully and only then creation the prometheus saved object?

src/plugins/data_source_management/public/components/utils.ts
```diff
@@ -443,11 +448,24 @@ export async function deleteDataSourceById(
...
+          selectedDataSource.id
+        );
+        const dataSourceMDSId = savedObject.references.find((r) => r.type === 'data-source')?.id;
+        return http.delete(
+          `${DATACONNECTIONS_BASE}/${selectedDataSource.title}/dataSourceMDSId=${
+            dataSourceMDSId || ''
+          }`
+        );
+      }
+      return deleteDataSourceById(selectedDataSource.id, savedObjectsClient);
```
ps48, line 468: Do we need to handle scenarios where one deletion fails either saved object or SQL and mention it to the user explicitly?

ps48: Some CTA for users as toast to understand what's the inconsistent state.

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_data_source_configuration/prometheus/configure_prometheus_data_source.tsx
```diff
@@ -146,6 +155,30 @@ export const ConfigurePrometheusDatasourcePanel = (props: ConfigurePrometheusDat
...
+              id="selectDataSource"
+              options={[
+                ...(!hideLocalCluster
+                  ? [{ value: LocalCluster.id, text: LocalCluster.label }]
+                  : []),
+                ...dataSources.map((ds) => ({ value: ds.id, text: ds.title || ds.id })),
+              ]}
+              value={currentDataSourceId}
+              onChange={(e) => setDataSourceIdForRequest(e.target.value)}
+              data-test-subj="dataSourceSelect"
```
ps48: Thanks!

ps48 (commented):
Have some minor comments. 
One Question: How permissions are enforced on saved object operations with Datasource backend roles associated to the prometheus PPL datasource?

src/plugins/data_source_management/public/components/direct_query_data_sources_components/direct_query_data_connection/manage_direct_query_data_connections_table.tsx
joshuali925: good point, thanks

src/plugins/data_source/common/data_connections/data_connection_saved_object_attributes.ts
joshuali925: let's leave it till later, this PR only modifies the existing prometheus type

src/plugins/data_source_management/server/routes/data_connections_router.ts
joshuali925: It means SQL side data source is gone, and user is deleting the orphaned data-connection saved object. We should continue

If SQL is not working (5xx) i feel we should directly show error to user and not touch saved objects

src/plugins/data_source_management/server/routes/data_connections_router.ts
joshuali925: I believe if creation in SQL plugin failed, it goes into catch and skips the saved object creation

src/plugins/data_source_management/public/components/utils.ts
joshuali925: That should be existing logic when calling this utils. The utils only deletes and throws error up https://github.com/opensearch-project/OpenSearch-Dashboards/blob/f129a7e48194aba1cb059c67a68a817a73a4139d/src/plugins/data_source_management/public/components/data_source_table/data_source_table.tsx#L451-L474

PR #10973, opened 2025-11-25
Fix saved queries for 3rd-party plugins by honoring wildcard supportedAppName

### LINKED ISSUE #10972: [BUG] Saved queries are filtered out for 3rd-party plugins
**Describe the bug**
If enabled (using `showSaveQuery`), the `TopNavMenu` offers the possibility to save queries. We are using this feature in a third-party plugin, which we recently updated to OpenSearch Dashboards 3.2.0. Since then, saving queries still works (e.g., the query shows up in Discover), but the Saved Queries Flyout remains empty.

The culprit seems to have been introduced with #8325, which filters saved queries by `supportedAppNames`. For DQL, which we are using, `LanguageConfig.supportedAppNames` contains an asterisk ("*"; declared at https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/query/query_string/language_service/lib/dql_language.ts#L34), but the filter function (https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/query/saved_query/saved_query_service.ts#L153) does not honor this as a wildcard, causing all saved queries to be filtered out.

**Expected behavior**
I'd expect that the wildcard returned from the `LanguageConfig`'s `supportedAppNames` is treated as such permitting apps not explicitly declared in the list of `supportedAppNames` to retrieve previously saved queries.

**OpenSearch Version**
3.2.0

**Dashboards Version**
3.2.0

**Fix**
I have prepared a fix that I will submit as a pull request shortly.

ginkel:
### Description

This PR brings back Saved Queries to 3rd-party plugins 

### Issues Resolved

fixes #10972

## Screenshot

image

(after the fix)

## Testing the changes

The fix can be most easily tested by creating a global saved query from a third-party plugin or a plugin that does not explicitly appear in the allowlist at https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data/public/query/query_string/language_service/lib/dql_language.ts#L34. Without the fix, the Saved Queries Flyout will remain empty, with the fix applied it will show the saved queries.

## Changelog
- fix: Restore saved queries for 3rd-party plugins by honoring wildcard supportedAppName

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

Tests do not pass, but they also don't pass on `main`.

### DIFF
--- a/src/plugins/data/public/query/saved_query/saved_query_service.test.ts
+++ b/src/plugins/data/public/query/saved_query/saved_query_service.test.ts
@@ -32,6 +32,7 @@ import { createSavedQueryService } from './saved_query_service';
 import { FilterStateStore } from '../../../common';
 import { SavedQueryAttributes } from './types';
 import { applicationServiceMock, uiSettingsServiceMock } from '../../../../../core/public/mocks';
+import { of } from 'rxjs';
 
 const savedQueryAttributes: SavedQueryAttributes = {
   title: 'foo',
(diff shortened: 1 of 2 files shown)


LDrago27:
Thanks @ginkel . I seemed to have missed out , updating the conditional check to honor the wildcard in supportedAppName section. Do you mind adding an unit test specific to the above scenario, so we are accounting for it. Thanks. Let me know if you need any help for me.

cc @ruanyl

ginkel:
Can you point me to the unit tests added for the original change so I can extend them?

LDrago27:
@ginkel Sure. The changes pertaining to your change (`findSavedQueries`) have their corresponding unit tests here. 
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/7928916297b7edcdc0b203ae82fbf130cc8a1da1/src/plugins/data/public/query/saved_query/saved_query_service.test.ts#L186

To account for your use case, lets mock the currentAppId To return a random APP Id and then have a test case ,such that we are still able to all the DQL saved queries with the random APP Id since for DQL Saved Queries have * in supported App names.

ginkel:
@LDrago27 Thanks for getting back to me and sorry that it took me a while myself to respond.

I have added a test, rebased the branch against main and force-pushed the result. Hopefully we're good to go now. ;-)

PR #10976, opened 2025-11-25
fix(explore): clear query state when unmounting app

joshuali925:
### Description

Clear query state before switching flavors. Some dataset might not be supported in all flavors, but the dataset object gets persisted in query string manager. When constructing redux store before loading a new flavor, it might read the object from there and use it.

It would be better to clear all states when switching between flavors, because flavors might be very different and we cannot assume all states still apply.

## Changelog

- fix: clear query state before switching flavors in explore

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed query state not being properly cleared when switching flavors in the Explore feature.


zhongnansu (dismissed):

angle943 (dismissed):

joshuali925:
my bad didn't test this much, it breaks if user goes to URL directly. will fix tmr

angle943 (approved):

PR #10984, opened 2025-11-27
[Explore Vis] adapt deprecated styles in embeddable

Qxisylolo:
## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for visualizations created with deprecated styling by automatically adapting legacy style data during retrieval and spec generation.
* **Tests**
  * Added tests ensuring legacy style adaptation runs when older visualizations are fetched.

### DIFF
--- a/src/plugins/explore/public/embeddable/explore_embeddable.tsx
+++ b/src/plugins/explore/public/embeddable/explore_embeddable.tsx
@@ -411,6 +411,8 @@ export class ExploreEmbeddable
             axesMapping: visualization.axesMapping,
           })?.styles;
 
+          this.searchProps.styleOptions = styles;
+
           const spec = matchedRule.toSpec(
             visualizationData.transformedData,
             numericalColumns,


ruanyl (approved):

src/plugins/explore/public/embeddable/explore_embeddable.tsx
```diff
@@ -401,12 +404,19 @@ export class ExploreEmbeddable
           };
           this.searchProps.searchContext = searchContext;
           const styleOptions = visualization.params;
+
+          const styles = adaptLegacyData({
```
ruanyl, line 408: Perhaps we need to update `searchProps.styleOptions` as well?

Qxisylolo: only table needs searchProps.styleOptions when rendering embeddable currently which doesn't need to adapt legacy data. but I will include this as table will use value mapping in the future and we may need to migrate some data. thanks!

PR #11030, opened 2025-12-04
Discover:Traces Red Metrics

TackAdam:
### Description
1. Add red metrics to the Discover:Trace experience by making 3 query calls and generating the visualizations, adjust the components in logs to support this new behavior of having 3 charts and reducing the number of buckets for it to be useable.
* Only compatible with data-prepper schema currently.
RedMetrics
Error handling:
ErrorHandling

2. Fix bug where legacy url navigation would sometimes not show the default columns for the trace page when switching between the two.
UrlIssue

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

## Summary by CodeRabbit

* **New Features**
  * RED metrics charts (Request, Error, Latency) in Discover:Traces with per‑chart headers, per‑chart cache keys, improved interval options, and configurable bucket targeting.
  * Per‑chart loading/error UI, clear field‑missing messages, and option to hide the interval selector.

* **Bug Fixes / UX**
  * Traces charts render reliably in Discover and avoid stuck/incomplete loading states.

* **Tests**
  * Extensive unit and integration tests for trace aggregation, queries, and error scenarios.

* **Style**
  * New trace chart styling for consistent layout and sizing.


kamingleung:
@TackAdam 
1. For the Error count chart, how do we determine something is an error? Which field are we getting that information? Can the chart show Errors grouped by status codes?
2. FnF feedback: In the Error handling screenshot you posted, is there a reason why Error count chart has a red banner while the Latency chart has a yellow banner? Should they be consistent?

src/plugins/explore/public/application/utils/state_management/actions/processors/trace_aggregation_processor.ts
```diff
@@ -0,0 +1,461 @@
...
+export const processTraceAggregationResults = (
+  requestAggResults: ISearchResult | null,
+  errorAggResults: ISearchResult | null,
+  latencyAggResults: ISearchResult | null,
+  dataset: DataView,
+  interval: string,
+  timeField: string = 'endTime',
+  dataPlugin?: DataPublicPluginStart,
+  rawInterval?: string,
+  uiSettings?: IUiSettingsClient
```
joshuali925: might be better to use object as argument, so order doesn't matter when adding new args in the future

src/plugins/explore/public/application/utils/state_management/actions/processors/trace_aggregation_processor.ts
```diff
@@ -0,0 +1,461 @@
...
+
+  if (!requestAggResults && !errorAggResults && !latencyAggResults) {
+    return result;
+  }
+
+  try {
+    let minTime: number;
+    let maxTime: number;
+
+    if (dataPlugin) {
```
joshuali925: why wouldn't dataPlugin be always available?

src/plugins/explore/public/application/utils/state_management/actions/processors/trace_aggregation_processor.ts
```diff
@@ -0,0 +1,461 @@
...
+  // Handle unit-only intervals (m, s, h, d) - transform to 1 unit
+  if (/^[smhd]$/.test(interval)) {
+    const transformedInterval = `1${interval}`;
+    return convertIntervalToMs(transformedInterval);
+  }
+
+  const match = interval.match(/^(\d+)([smhd])$/);
+  if (!match) {
+    if (interval === 'auto') {
+      return 30000; // 30 seconds for auto intervals
```
joshuali925: why is auto always 30s?

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -728,6 +813,122 @@ export const executeDataTableQuery = createAsyncThunk<
   );
 });
 
+export const executeTraceAggregationQueries = createAsyncThunk<
```
joshuali925: would be better to move them out from query_actions, so if traces query ever change, we don't need to change the shared query_actions file

TackAdam:
> @TackAdam
> 
> 1. For the Error count chart, how do we determine something is an error? Which field are we getting that information? Can the chart show Errors grouped by status codes?
> 2. FnF feedback: In the Error handling screenshot you posted, is there a reason why Error count chart has a red banner while the Latency chart has a yellow banner? Should they be consistent?

1. The chart currently just shows status.code = 2 for errors, a more complicated ppl request could be made to do groupings by http status code, but data-prepper examples we are currently go off do not always have that information on the spans. Will consider this for future follow-up after investigating how to make that as reliable as the status.code field.

2. The unavailable warning (yellow) is if they manually changed the name of the field durationInNanos indicates to the user they have re-named the field and that it is not compatible with the PPL request required to generate the visualization.
The red warning indicates the PPL request to go through in which case we just display the direct error message output ie out of resources.

src/plugins/explore/public/application/utils/state_management/actions/processors/trace_aggregation_processor.ts
TackAdam: Created `ProcessTraceAggregationParams` interface and updated `processTraceAggregationResults` to use a single object parameter. This makes parameter order irrelevant and future parameter additions easier. All call sites (production and tests) have been updated to use the named parameter syntax.

src/plugins/explore/public/application/utils/state_management/actions/processors/trace_aggregation_processor.ts
TackAdam: Auto intervals are now dynamically calculated based on the time range using histogram configs (same approach as logs). The hardcoded 30s fallback was removed. When interval is "auto", we now use `createHistogramConfigs` to calculate the optimal interval based on the selected time range, then convert that to milliseconds. The interval selection is disabled at the moment pending bugfix but will work when re-enabled if we decide to do that now.

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
TackAdam: Moved all trace-specific query actions (`executeTraceAggregationQueries`, `executeRequestCountQuery`, `executeErrorCountQuery`, `executeLatencyQuery`, and `prepareTraceCacheKeys`) to a new dedicated file `trace_query_actions.ts`. This ensures future trace query changes won't affect the shared query_actions file.

src/plugins/explore/public/application/utils/state_management/actions/processors/trace_aggregation_processor.ts
TackAdam: Made `dataPlugin` a required parameter since it's always available in production. Removed unnecessary `if (dataPlugin)` checks and simplified the time range logic. Tests were updated to always provide the mock dataPlugin.

kamingleung:
> 2. The unavailable warning (yellow) is if they manually changed the name of the field durationInNanos indicates to the user they have re-named the field and that it is not compatible with the PPL request required to generate the visualization.
>    The red warning indicates the PPL request to go through in which case we just display the direct error message output ie out of resources.

@TackAdam I would recommend to display both as red error messages. In either scenario, it cannot generate a visualization. A yellow warning may be suitable for situations like a visualization with partial data.

TackAdam:
> > 2. The unavailable warning (yellow) is if they manually changed the name of the field durationInNanos indicates to the user they have re-named the field and that it is not compatible with the PPL request required to generate the visualization.
> >    The red warning indicates the PPL request to go through in which case we just display the direct error message output ie out of resources.
> 
> @TackAdam I would recommend to display both as red error messages. In either scenario, it cannot generate a visualization. A yellow warning may be suitable for situations like a visualization with partial data.

Updated both are now red.

src/plugins/explore/public/application/utils/state_management/actions/processors/trace_aggregation_processor.ts
```diff
@@ -0,0 +1,471 @@
...
+    case 'h':
+      return num * 60 * 60 * 1000;
+    case 'd':
+      return num * 24 * 60 * 60 * 1000;
+    default:
+      return fallbackMs;
+  }
+}
+
+function processRequestCountData(
```
ps48: The three processing functions (processRequestCountData, processErrorCountData, processLatencyData) in trace_aggregation_processor.ts share a lot of code. 

Does it make sense to refactor them into a single function:

```
Refactor into a single generic function:

  interface ChartProcessConfig {
    metricField: string | string[];
    yAxisLabel: string;
    transformer?: (value: number) => number;
  }

  function processChartData(
    results: ISearchResult,
    config: ChartProcessConfig,
    dataset: DataView,
    minTime: number,
    maxTime: number,
    intervalMs: number,
    timeField: string,
    xAxisFormat: any
  ): ChartData { /* ... */ }
```

src/plugins/explore/public/components/chart/_histogram.scss
```diff
@@ -59,3 +63,40 @@
     }
   }
 }
+
+// Traces charts container styles
+.exploreTracesChart {
+  &__container {
+    min-height: 200px;
```
ps48: nit. may be we can use eui token:
```
 $traces-chart-height: $euiSize * 25;  // Using EUI tokens
 ```

File: src/plugins/explore/public/application/utils/state_management/actions/processors/trace_aggregation_processor.ts [hunk 7]
```diff
@@ -0,0 +1,471 @@
...
+          dataMap.set(bucketKey, (dataMap.get(bucketKey) || 0) + requestCount);
+        }
+      }
+    }
+
+    const startBucket = Math.floor(minTime / actualIntervalMs) * actualIntervalMs;
+    const endBucket = Math.floor(maxTime / actualIntervalMs) * actualIntervalMs;
+
+    for (let bucketKey = startBucket; bucketKey <= endBucket; bucketKey += actualIntervalMs) {
+      const value = dataMap.get(bucketKey) || 0;
```
ps48: Is there a possibility that the time range is large and interval is small, this could generate excessive buckets. Should we have a maxBuckets size?

ps48 (commented):
This is great addition thanks @TackAdam! have some minor comments

src/plugins/explore/public/application/utils/state_management/actions/processors/trace_aggregation_processor.ts
TackAdam: Currently the interval selection behavior is disabled. It defaults to auto which has the safeguards for excessive buckets. Once the reported bug is fixed if we decide to re-enable user manual interval selection I will further test this and verify no issues.

src/plugins/explore/public/application/utils/state_management/actions/processors/trace_aggregation_processor.ts
TackAdam: Moved the shared logic into a shared function
``` 
function processChartData(
```
But let diverged logic in the separate wrappers for read-ability and adjustments in the future.
ie
```
function processRequestCountData(
  results: ISearchResult,
  dataset: DataView,
  minTime: number,
  maxTime: number,
  intervalMs: number,
  timeField: string,
  xAxisFormat: { id: string; params: { pattern: string } }
): ChartData {
  return processChartData(
    results,
    {
      metricField: ['request_count', 'count()'],
      yAxisLabel: 'Request Count',
      aggregationType: 'sum',
    },
    dataset,
    minTime,
    maxTime,
    intervalMs,
    timeField,
    xAxisFormat
  );
}
```

src/plugins/explore/public/components/chart/_histogram.scss
TackAdam: Updated too 
    min-height: clamp(200px, 20vh, 400px);

ps48 (approved):

joshuali925 (approved):

PR #11057, opened 2025-12-11
[Explore] Update PatternsTableFlyout to render without timefield

sejli:
### Description

For datasets that don't have a designated timefield, the patterns flyout does not work since it displays the results with two columns: the timefield and the pattern. This PR removes that restriction, and for datasets that don't have timefields, the results will display regardless.

## Screenshot

Before:
image

After:

image

## Changelog

- fix: Update PatternsTableFlyout to render without timefield

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the Patterns Table Flyout to render without requiring a time field. The event table now displays only the necessary columns based on available dataset fields, improving flexibility when time field data is unavailable.

### DIFF
--- a/changelogs/fragments/11057.yml
+++ b/changelogs/fragments/11057.yml
@@ -1,0 +1,2 @@
+fix:
+- Update PatternsTableFlyout to render without timefield ([#11057](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11057))


src/plugins/explore/public/components/patterns_table/utils/utils.ts
```diff
@@ -62,21 +62,27 @@ export const createSearchPatternQueryWithSlice = (
   patternsField: string,
   usingRegexPatterns: boolean,
   patternString: string,
-  timeField: string,
+  timeField: string | undefined,
   pageSize: number,
   pageOffset: number
 ) => {
   // TODO: switch this logic back to adding onto the createSearchPatternQuery
```
paulstn, line 69: Completely optional for this PR but if you would be able to fulfill this TODO and switch the query here, that would be really nice! I believe the associated backend bug that this workaround was required for has been fixed.

PR #11069, opened 2025-12-16
[Explore] Memoize patterns components to fix pagination in tables

sejli:
### Description

The `OuiBasicTable` components in patterns had broken pagination due to a React rerendering race condition. The pagination DOM was being destroyed and recreated before click events completed, causing the `onClick()` handler to never fire. This issue would be reproducible on Chrome, but not on Firefox (most likely due to a slower rendering engine).

This PR adds React memoization to the following patterns components to prevent unnecessary rerenders:

1. patterns_flyout_context.tsx: Wrapped callbacks in useCallback and context value in useMemo to stabilize context and prevent consumer rerenders
2. patterns_container.tsx: Memoized items array with useMemo to prevent recreation when flyout state changes
3. patterns_table.tsx: Wrapped component in React.memo to prevent prop-based re-renders
4. patterns_flyout_event_table.tsx: Wrapped component in React.memo to prevent prop-based re-renders

## Screenshot

Before: 

https://github.com/user-attachments/assets/eabb2115-ccbd-4b8d-8ad0-1596161ec603

After:

https://github.com/user-attachments/assets/2a653d96-0004-4cc4-a27d-e5964d8a3d67

## Changelog

- fix: [Explore] Memoize patterns components to fix pagination in tables

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed pagination issues in Explore patterns tables.
  * Added validation for required pattern fields to surface errors instead of silent failures.

* **Performance**
  * Introduced memoization to reduce unnecessary re-renders and improve table/flyout responsiveness.

* **UI**
  * Improved flyout behavior and pagination responsiveness for a smoother browsing experience.

### DIFF
--- a/changelogs/fragments/11115.yml
+++ b/changelogs/fragments/11115.yml
@@ -1,0 +1,2 @@
+fix:
+- [Chat] Fix tool call positioning in conversation timeline ([#11115](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11115))
--- a/changelogs/fragments/11116.yml
+++ b/changelogs/fragments/11116.yml
@@ -1,0 +1,2 @@
+feat:
+- Migrate pie chart ([#11116](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11116))
--- a/changelogs/fragments/11129.yml
+++ b/changelogs/fragments/11129.yml
@@ -1,0 +1,2 @@
+feat:
+- Add Client-Side Math Expression Evaluation for TSVB ([#11129](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11129))
--- a/changelogs/fragments/11131.yml
+++ b/changelogs/fragments/11131.yml
@@ -1,0 +1,2 @@
+feat:
+- [chat] expose action register method to allow plugins to register permanent actions ([#11131](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11131))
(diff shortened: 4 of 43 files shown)


src/plugins/explore/public/components/patterns_table/patterns_container.tsx
```diff
@@ -31,6 +31,30 @@ const PatternsContainerContent = () => {
 
   const usingRegexPatterns = useSelector(selectUsingRegexPatterns);
 
+  const logsTotal = histogramResults?.hits.total || 0;
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
paulstn: nit (from the ai reviewer): Is this eslint-disable needed here, the next line seems to be defining a local variable and the disablement is for a hook dependency array

angle943: @sejli could you look at this?

src/plugins/explore/public/components/patterns_table/patterns_container.tsx
```diff
@@ -58,10 +82,6 @@ const PatternsContainerContent = () => {
   }
 
   try {
```
Maosaic: Do we still need this try block?

src/plugins/explore/public/components/patterns_table/patterns_container.tsx
sejli: Good callout, initially I had it disabled because memoizing the `hits` would have some issues. Seemed to have resolved itself after some iterations of my implementation, so I'll wrap this back in a `useMemo()`.

src/plugins/explore/public/components/patterns_table/patterns_container.tsx
sejli: Don't think so, removed in my most recent commit.

PR #11074, opened 2025-12-18
[CSP] Create and provide nonces for style-src-elem in CSP-RO

paulstn:
## Description

This PR implements CSP (Content Security Policy) nonces for the `style-src-elem` directive within the CSP Report-Only feature. Nonces provide a more secure alternative to `'unsafe-inline'` by allowing only styles with a matching cryptographic token to be applied.

### Background

CSP nonces are cryptographically random values unique to each HTTP response. When a nonce is specified in the CSP header, only inline styles containing the matching `nonce` attribute will be allowed. This significantly improves security by preventing injection attacks while still allowing legitimate inline styles.

The key insight behind nonces is that an attacker cannot predict what value will be generated for a given request. Even if they manage to inject a `<style>` tag into the page, their injected element won't have the correct nonce attribute, so the browser will refuse to apply those styles.

### Why This Approach

OpenSearch Dashboards has multiple code paths that create inline styles, and each required a different solution:

**Server-rendered styles** are the simplest case. The nonce is generated at request time and passed through the rendering pipeline. These styles exist in the initial HTML before any JavaScript executes, so we simply add the `nonce` attribute during server-side rendering. A `<meta name="csp-nonce">` tag is also rendered to make the nonce available to client-side code.

**Client-side dynamic styles** are created by our own code after the page loads. Services like `styles_service.ts` (for accessibility settings) and `inject_header_style.ts` (for truncation settings) create style elements at runtime. These retrieve the nonce from the meta tag using a shared `getNonce()` utility and apply it when creating the element.

**Webpack-bundled styles** are handled by webpack's style-loader, which has built-in nonce support through the `__webpack_nonce__` global variable. We set this variable early in the bootstrap process, before any webpack bundles load, so all dynamically injected CSS from our bundles automatically receives the nonce.

**Third-party libraries** like nano-css (used by react-use) and the ace/brace editor create style elements directly through DOM APIs without any awareness of CSP nonces. To handle these, we intercept `document.createElement` and `document.createElementNS` early in the page lifecycle. When any code creates a `<style>` element, our interceptor automatically adds the nonce attribute. This approach is necessary because we cannot modify these libraries' source code, and they don't provide hooks for customizing how they create DOM elements.

### CSP Directive Strategy

Rather than using a single `style-src` directive, we split it into more granular directives to maximize security while maintaining compatibility:

`style-src-elem` receives the nonce and covers all `<style>` elements. This is where nonces provide real protection, as injected style tags are the primary attack vector for CSS-based attacks.

`style-src-attr` must allow `'unsafe-inline'` due to Monaco editor. Monaco heavily relies on inline style attributes like `style="color: red"`. These cannot be secured with nonces, CSP provides no mechanism for nonce-protecting style attributes. This is a known limitation: ([microsoft/monaco-editor#271](https://github.com/microsoft/monaco-editor/issues/271)) shared by any application that embeds Monaco.

`style-src` itself remains at `'self'` as a catch all for legacy browsers that do not understand the newer directives.

### How Nonces Protect Against XSS

```mermaid
sequenceDiagram
    participant A as Attacker
    participant B as Victim Browser
    participant S as Server

    A->>B: 1. Sends malicious link:<br/>example.com/search?q=<script>evil()</script>

    Note over B: Victim clicks link

    B->>S: 2. GET /search?q=<script>evil()</script>
    S->>S: 3. Generate nonce "x7f9k2"
    S-->>B: 4. Response

    Note over B: Header: script-src 'nonce-x7f9k2'<br/><script nonce="x7f9k2">legit()</script><br/><script>evil()</script>

    B->>B: 5. Execute legit()
    B->>B: 6. Block evil()
```

```mermaid
sequenceDiagram
    participant A as Attacker
    participant S as Server
    participant DB as Database
    participant B as Victim Browser

    A->>S: 1. Post comment:<br/><script>evil()</script>
    S->>DB: 2. Store comment

    Note over A,B: After some time

    B->>S: 3. Request page with comments
    S->>DB: 4. Fetch comments
    DB-->>S: Returns stored content (includes evil script)
    S->>S: 5. Generate nonce "m3p8q1"
    S-->>B: 6. Response

    Note over B: Header: script-src 'nonce-m3p8q1'<br/><script nonce="m3p8q1">legit()</script><br/><script>evil()</script>

    B->>B: 7. Execute legit()
    B->>B: 8. Block evil()
```

### Implementation Details

The nonce lifecycle begins in `http_resources_service.ts` and `ui_render_mixin.js`, which are the two entry points for serving HTML in OpenSearch Dashboards. A 16-byte (128-bit) random value is generated using Node's `crypto.randomBytes()` for each request—this provides sufficient entropy to be cryptographically unpredictable.

The `CspReportOnlyConfig` class was refactored to support dynamic nonce insertion. Rather than storing a static header string, it now exposes a `buildHeaderWithNonce(nonce)` method that constructs the header at request time. The `nonceDirectives` configuration array determines which directives receive the nonce (defaulting to `['style-src-elem']`).

The DOM interception for third-party libraries runs before any application code. By patching `document.createElement` in the bootstrap script, we ensure that even the earliest style element creations are covered. The original methods are preserved and called normally—we simply add the nonce attribute to any `<style>` element before returning it.

### Configuration

The feature is configured under `csp-report-only` in `opensearch_dashboards.yml`:

```yaml
csp-report-only.isEmitting: true
```

The `nonceDirectives` array can be customized if needed, though the default of `['style-src-elem']` should be appropriate for most deployments.

### Testing

Unit tests were added for the `CspReportOnlyConfig` class covering nonce application, the `buildHeaderWithNonce` method, and interaction with other configuration options. Rendering service tests and template snapshots were updated to account for the new nonce parameter.

## Changelog

- feat: style-src-elem nonces for CSP report-only

## Summary by CodeRabbit

* **New Features**
  * CSP enhancements: per-request nonces are generated and injected into rendered pages and inline/dynamically created styles in report-only mode, improving inline-style security without changing behavior when nonces are absent.
* **Public API**
  * Exposes a utility to read the CSP nonce so client code can apply it when available.
* **Tests**
  * Added tests covering nonce extraction and propagation to rendering and style injection.

### DIFF
--- a/src/core/public/integrations/styles/styles_service.test.ts
+++ b/src/core/public/integrations/styles/styles_service.test.ts
@@ -31,9 +31,13 @@
 import { BehaviorSubject } from 'rxjs';
 
 jest.mock('!!raw-loader!./disable_animations.css', () => 'MOCK DISABLE ANIMATIONS CSS');
+jest.mock('../../utils', () => ({
+  getNonce: jest.fn(),
+}));
 
 import { StylesService } from './styles_service';
 import { uiSettingsServiceMock } from '../../ui_settings/ui_settings_service.mock';
+import { getNonce } from '../../utils';
 
 describe('StylesService', () => {
   const flushPromises = () => new Promise((resolve) => setTimeout(resolve, 100));
@@ -70,5 +74,35 @@ describe('StylesService', () => {
     disableAnimations$.next(false);
     await flushPromises();
     expect(getDisableAnimationsTag().textContent).toEqual('');
+  });
+
+  test('sets nonce attribute on style tag when getNonce returns a value', async () => {
+    const mockGetNonce = getNonce as jest.Mock;
+    mockGetNonce.mockReturnValue('test-nonce-abc123');
+
+    const disableAnimations$ = new BehaviorSubject(false);
+    const uiSettings = uiSettingsServiceMock.createSetupContract();
+    uiSettings.get$.mockReturnValueOnce(disableAnimations$);
+
+    new StylesService().start({ uiSettings });
+    await flushPromises();
+
+    const styleTag = getDisableAnimationsTag();
+    expect(styleTag.getAttribute('nonce')).toBe('test-nonce-abc123');
+  });
+
+  test('does not set nonce attribute on style tag when getNonce returns empty string', async () => {
+    const mockGetNonce = getNonce as jest.Mock;
+    mockGetNonce.mockReturnValue('');
+
+    const disableAnimations$ = new BehaviorSubject(false);
+    const uiSettings = uiSettingsServiceMock.createSetupContract();
+    uiSettings.get$.mockReturnValueOnce(disableAnimations$);
+
+    new StylesService().start({ uiSettings });
+    await flushPromises();
+
+    const styleTag = getDisableAnimationsTag();
+    expect(styleTag.getAttribute('nonce')).toBeNull();
   });
 });
--- a/src/core/public/utils/csp.test.ts
+++ b/src/core/public/utils/csp.test.ts
@@ -1,0 +1,46 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { getNonce } from './csp';
+
+describe('getNonce', () => {
+  afterEach(() => {
+    // Clean up any meta tags created during tests
+    const meta = document.querySelector('meta[name="csp-nonce"]');
+    if (meta) {
+      meta.remove();
+    }
+  });
+
+  it('returns nonce value when meta tag exists with content', () => {
+    const meta = document.createElement('meta');
+    meta.setAttribute('name', 'csp-nonce');
+    meta.setAttribute('content', 'test-nonce-123');
+    document.head.appendChild(meta);
+
+    expect(getNonce()).toBe('test-nonce-123');
+  });
+
+  it('returns empty string when meta tag does not exist', () => {
+    expect(getNonce()).toBe('');
+  });
+
+  it('returns empty string when meta tag has no content attribute', () => {
+    const meta = document.createElement('meta');
+    meta.setAttribute('name', 'csp-nonce');
+    document.head.appendChild(meta);
+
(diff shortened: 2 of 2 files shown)


src/core/public/utils/csp.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Retrieves the CSP nonce from the meta tag in the document head.
+ * The nonce is used for inline styles and scripts to comply with Content Security Policy.
+ *
+ * @returns The nonce value as a string, or an empty string if not found
+ * @public
+ */
+export function getNonce(): string {
+  const meta = document.querySelector('meta[name="csp-nonce"]');
+  return meta?.getAttribute('content') ?? '';
+}
```
angle943, line 16: can we add some unit tests this for code coverage? (not a blocker)

angle943 (approved):

src/core/public/utils/csp.ts
paulstn: Will add

joshuali925 (approved):
This is good information. PR looks good just some questions:
1. The config is report only, any reason we don't actually block non-matching nonce styles?
2. If there's a XSS vulnerability, wouldn't the attacker be able to read `window.__webpack_nonce__` and know what the nonce is to add to their payload?

paulstn:
To your questions @joshuali925 
> The config is report only, any reason we don't actually block non-matching nonce styles?

We will eventually move this system over to the real CSP system and block non-matching nonce styles, but the goal with this is to see if there are any remaining violations, as to not block functioning code in production environments in case it hasn't been covered.

> If there's a XSS vulnerability, wouldn't the attacker be able to read `window.__webpack_nonce__` and know what the 
nonce is to add to their payload?

`window.__webpack_nonce__` exists to propagate the nonce to webpack related modules (and other libraries which create style elements), and it gets set by just reading the meta tag with the nonce, same way as client side code does. By the time an attacker reads `window.__webpack_nonce__`, any XSS attack being put into the page would be delivered as an http response alongside a new nonce. Do you see any way in particular in which webpack nonce wouldn't work?

PR #11084, opened 2025-12-18
[CSP] Replace handlebars package with safe alternative

paulstn:
### Description

Replace the `handlebars` package with `kbn-handlebars` to eliminate CSP `unsafe-eval` violations in browser-side template compilation.

### Problem

Two files in `vis_type_timeseries` use Handlebars in the browser:
- `replace_vars.js` - Markdown template variable replacement
- `tick_formatter.js` - Axis tick value formatting

The standard `handlebars.compile()` method internally uses `new Function()` to generate optimized JavaScript from templates. This requires the CSP directive `script-src 'unsafe-eval'`, which weakens security policy and is prohibited by many enterprise security standards.

### Solution

Use https://www.npmjs.com/package/kbn-handlebars, a drop-in replacement that provides a `compileAST()` method which interprets the template AST directly instead of generating JavaScript code:

`handlebars`: Template → AST → JavaScript Function (uses eval) → Execute → Output
`kbn-handlebars`: Template → AST → Direct AST interpretation → Output

### Changes

Client-side (CSP fix):
- `replace_vars.js`: `handlebars.compile()` → `Handlebars.compileAST()`
- `tick_formatter.js`: `handlebars.compile()` → `Handlebars.compileAST()

Server-side (no CSP concern, uses compile() for better performance):
- `app_bootstrap.js`: Import changed to `kbn-handlebars`
- `visual_regression.js`: Import changed to `kbn-handlebars`

Build configuration:
- Updated `webpack` exclude regex to transpile `kbn-handlebars` (contains ES2020 nullish coalescing syntax)

Dependencies:
- Added `kbn-handlebars@1.0.0`
- Removed direct `handlebars` dependency

### Why it works

- `compileAST()` supports all options currently in use (`strict`, `knownHelpersOnly`, `noEscape`)
- Returns the same `TemplateDelegate` function signature - no changes needed to calling code
- Server-side files can continue using `compile()` since Node.js has no CSP restrictions
- `kbn-handlebars` extends the original Handlebars object, so `registerHelper()` and other APIs remain available

### Trade-off

AST interpretation is slower than compiled JavaScript execution. This may impact rendering performance on dashboards with many TSVB visualizations or large datasets, but eliminates the `unsafe-eval` requirement entirely.

### Additional

Also removes the: 
```
^ A single error about an inline script not firing due to content security policy is expected!
```
message because the inline script not firing issue was handled previously, so this message is not needed anymore.

## Testing the changes

- Created a TSVB that would use handlebars for a custom label, verifying that no CSP violation reports were generated
- Verified that server side package replacements would still work, e.g. verified that bootstrap still worked

## Changelog

- infra: Replace handlebars package with kbn-handlebars

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

## Summary by CodeRabbit

* **Dependencies**
  * Replaced handlebars with kbn-handlebars, an optimized internal package providing improved performance and compatibility.

* **Improvements**
  * Removed unnecessary Content Security Policy-related console warning during application initialization.

* **Build System**
  * Updated build configuration to properly transpile the new handlebars package.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -25496,7 +25496,7 @@ string-length@^4.0.1:
     char-regex "^1.0.2"
     strip-ansi "^6.0.0"
 
-"string-width-cjs@npm:string-width@^4.2.0":
+"string-width-cjs@npm:string-width@^4.2.0", "string-width@^1.0.2 || 2 || 3 || 4", string-width@^4.0.0, string-width@^4.1.0, string-width@^4.2.0, string-width@^4.2.2, string-width@^4.2.3:
   version "4.2.3"
   resolved "https://registry.yarnpkg.com/string-width/-/string-width-4.2.3.tgz#269c7117d27b05ad2e536830a8ec895ef9c6d010"
   integrity sha512-wKyQRQpjJ0sIp62ErSZdGsjMJWsap5oRNihHhu6G7JVO/9jIB6UyevL+tXuOqrng8j/cxKTWyWUwvSTriiZz/g==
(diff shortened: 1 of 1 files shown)


SuZhou-Joe (approved):

paulstn:
> to see if performance will take a heavy hit

Ultimately I don't think it would matter much, since the client handlebar usage would only be to specify customized labels and annotations for TSVBs, which is not a use case that would scale to any noticeable degree.

PR #11108, opened 2025-12-26
[Explore vis] Refactor vis data processing

ruanyl:
### Description
1. refactor pivot and aggregate function
2. fix empty pivot field not displayed correctly
3. refactor createBarSeries to unified the stacked bar series creation

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

## Release Notes

* **New Features**
  * Bar charts now support multiple data series with dynamic axis configuration
  * Enhanced data aggregation and pivoting capabilities for improved data grouping
  * Improved time-based data handling and bucketing

* **Refactor**
  * Streamlined visualization data transformation pipeline
  * Enhanced legend display on chart components

### DIFF
--- a/src/plugins/explore/public/components/visualizations/utils/data_transformation.ts
+++ b/src/plugins/explore/public/components/visualizations/utils/data_transformation.ts
@@ -237,33 +237,6 @@ const roundToTimeUnit = (timestamp: Date, unit: TimeUnit): Date => {
 };
 
 /**
- * Aggregate data by time intervals
- * This is a convenience wrapper around aggregate with timeUnit parameter
- *
- * @param options - Aggregation configuration options
- * @param options.groupBy - Field name for time/date values (ISO 8601 format)
- * @param options.field - Field name for numerical values to aggregate
- * @param options.timeUnit - Time interval unit (YEAR, MONTH, DATE, HOUR, MINUTE, SECOND, AUTO)
- * @param options.aggregationType - Type of aggregation to apply (SUM, MEAN, MAX, MIN, COUNT, NONE)
- *
- * @returns Function that takes data array and returns array of aggregated objects
- *
- * @example
- * const aggregateByDay = aggregateByTime({
- *   groupBy: 'timestamp',
- *   field: 'sales',
- *   timeUnit: TimeUnit.DATE,
- *   aggregationType: AggregationType.SUM
- * });
- */
-export const aggregateByTime = (options: {
-  groupBy: string;
-  field: string;
-  timeUnit: TimeUnit;
-  aggregationType: AggregationType;
-}) => aggregate({ ...options });
-
-/**
  * Helper function to normalize empty/null/undefined values
  * @param value - Value to normalize
  * @param defaultValue - Default value to use for empty values
--- a/src/plugins/explore/public/components/visualizations/bar/to_expression.ts
+++ b/src/plugins/explore/public/components/visualizations/bar/to_expression.ts
@@ -38,13 +38,7 @@ import {
   assembleSpec,
   buildVisMap,
 } from '../utils/echarts_spec';
-import {
-  aggregate,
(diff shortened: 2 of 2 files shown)


Qxisylolo (approved):

src/plugins/explore/public/components/visualizations/utils/data_transformation.ts
```diff
@@ -27,60 +28,179 @@ const aggregateValues = (aggregationType: AggregationType, values?: number[]) =>
...
+export const transform = (...fns: TransformFn[]) => (state: EChartsSpecState) => {
+  const { data } = state;
+  const transformedData: Array<Array<Record<string, any>>> = [data];
+
+  for (const fn of fns) {
+    const transformed = fn(transformedData[transformedData.length - 1]);
+    transformedData.push(transformed);
+  }
+
+  return { ...state, transformedData };
```
Qxisylolo, line 68: Is there any particular reason to store every stage of the transformed data? I'm wondering if we can just return the last element as transformedData instead. In such way, we wouldn't need to write transformedData[transformedData?.length - 1] in other places.

src/plugins/explore/public/components/visualizations/utils/data_transformation.ts
```diff
@@ -345,81 +374,46 @@ export const pivotDataWithTime = <T extends BaseChartStyle>({
...
 
-export const formatData = <T extends BaseChartStyle>(
-  state: EChartsSpecState<T>
-): EChartsSpecState<T> => {
-  const { data, axisColumnMappings } = state;
+    // Add a column for each pivot value
+    pivotValues.forEach((pv) => {
+      row[pv] = aggregationType
+        ? aggregateValues(aggregationType, pd[pv]) ?? null
+        : pd[pv]?.[0] ?? null;
```
Qxisylolo, line 407: In previous line and area charts, no aggregation was supported, so one single timestamp could contain multiple data points. Currently, when aggregation is not applied, only the first data point is returned. Would this still align with the original goal, or could it result in data loss? 

截屏2025-12-29 11 50 27

src/plugins/explore/public/components/visualizations/utils/data_transformation.ts
ruanyl: @Qxisylolo Actually no particular reason, I was just thinking if this might be helpful for debug

src/plugins/explore/public/components/visualizations/utils/data_transformation.ts
ruanyl: > Currently, when aggregation is not applied, only the first data point is returned. 

If no data transformation, I would expect it to not change the raw data, you should see the two data points. Do you mean the current implementation breaks this behavior?

Qxisylolo: In my understanding, in current implementation, if no aggregationType, it will return the first element. But what we are expecting is returning all the entries that under group pd[pv].
```
  pivotValues.forEach((pv) => {
      row[pv] = aggregationType
        ? aggregateValues(aggregationType, pd[pv]) ?? null
        : pd[pv]?.[0] ?? null;

```

ruanyl: I see your point, let me check

ruanyl: I think you raised a good point, it's better to return `pd[pv]` which is an array, and then flatten it for line chart when `aggregationType` is not provided. I will submit another PR to handle the case.

It will be like:
```js
transform(
  pivot(
    groupBy: 'time',
    pivot: 'type',
    field: 'value'
  ),
  flatten(),
  convertTo2DArray(),
)
```

PR #11112, opened 2025-12-29
[chat] Fix ppl execution tool incorrect status

wanglam:
### Description

This PR addresses a bug in the PPL query execution tool where an incorrect function was being used to generate the cache key for query execution status tracking. The issue was identified in the `ppl_execute_query_action.tsx` file where `defaultPrepareQueryString` was being used instead of the more appropriate `prepareQueryForLanguage` function.

The fix replaces the deprecated `defaultPrepareQueryString` function with `prepareQueryForLanguage`, which properly processes the query object and returns the correctly formatted query string via its `.query` property. This ensures that the cache key generation aligns with the standardized query preparation logic used throughout the Explore plugin, preventing potential mismatches in query execution status tracking and improving the reliability of the PPL query execution workflow.

## Changelog

- fix: [chat] Fix ppl execution tool incorrect status

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect status reporting in the PPL execution tool for chat functionality.

### DIFF
--- a/src/plugins/explore/public/components/query_panel/actions/ppl_execute_query_action.test.tsx
+++ b/src/plugins/explore/public/components/query_panel/actions/ppl_execute_query_action.test.tsx
@@ -12,7 +12,7 @@ import {
 import { QueryExecutionStatus } from '../../../application/utils/state_management/types';
 import { useOpenSearchDashboards } from '../../../../../opensearch_dashboards_react/public';
 import { loadQueryActionCreator } from '../../../application/utils/state_management/actions/query_editor/load_query';
-import { defaultPrepareQueryString } from '../../../application/utils/state_management/actions/query_actions';
+import { prepareQueryForLanguage } from '../../../application/utils/languages';
 import { useDispatch } from 'react-redux';
 
 // Mock dependencies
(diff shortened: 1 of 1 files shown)


Hailong-am (approved):

src/plugins/explore/public/components/query_panel/actions/ppl_execute_query_action.tsx
```diff
@@ -145,7 +145,7 @@ export function usePPLExecuteQueryAction(
             ...query,
             query: args.query,
           };
-          const cacheKey = defaultPrepareQueryString(queryObject);
+          const cacheKey = prepareQueryForLanguage(queryObject).query;
```
ruanyl, line 148: `prepareQueryForLanguage` and `defaultPrepareQueryString` works differently, `defaultPrepareQueryString` will remove `stats` command from PPL query which I think is intended for discover ppl query.

wanglam: image
This is for fixing the issue above. When the PPL includes a `stats` command, it will switch to the visualization tab and show a syntax error. The chatbot displays `Query updated and executed successfully` which has different results from the visualization tab. The root cause is that the `execute_ppl_query` tool uses the modified PPL query (with `stats` removed) as the cache key, and this modified query can be executed successfully. So I use the original query here to keep it consistent with the visualization tab.

PR #11115, opened 2025-12-30
[Chat] Fix tool call positioning in conversation timeline

wanglam:
### Description

This PR improves the reliability of tool call placement in the chat conversation timeline by implementing a multi-strategy approach for determining where tool calls should be attached. Previously, tool calls could be misplaced in the timeline when users sent new messages after an assistant's response, leading to incorrect conversation flow and potential confusion.

The fix introduces three strategies for tool call placement: (1) using an explicit parent message ID when provided by the backend, (2) analyzing timeline positions to determine if the last assistant message is still the current conversation turn, and (3) creating a placeholder assistant message when the LLM responds with only tool calls and no text message. This ensures tool calls are always correctly associated with their corresponding assistant responses, maintaining proper chronological ordering in the chat interface. Comprehensive unit tests have been added to validate all three placement strategies.

### Screenshots

- Before
image

- After
image

## Changelog

- fix: [Chat] Fix tool call positioning in conversation timeline

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tool call positioning in chat conversations to display at the correct location in the timeline. Tool calls are now properly positioned based on conversation context, with improved handling when explicit parent message information is not provided.

### DIFF
--- a/changelogs/fragments/10909.yml
+++ b/changelogs/fragments/10909.yml
@@ -1,0 +1,2 @@
+feat:
+- Simplify threshold logic ([#10909](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10909))
--- a/changelogs/fragments/11103.yml
+++ b/changelogs/fragments/11103.yml
@@ -1,0 +1,2 @@
+fix:
+- ML router error response format ([#11103](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11103))
--- a/changelogs/fragments/11112.yml
+++ b/changelogs/fragments/11112.yml
@@ -1,0 +1,2 @@
+fix:
+- [chat] Fix ppl execution tool incorrect status ([#11112](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11112))
--- a/changelogs/fragments/11113.yml
+++ b/changelogs/fragments/11113.yml
@@ -1,0 +1,2 @@
+feat:
+- Migrate line and facet bar chart to echart ([#11113](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11113))
--- a/package.json
+++ b/package.json
@@ -204,7 +204,7 @@
     "@elastic/eui": "npm:@opensearch-project/oui@1.22.1",
     "@elastic/good": "^9.0.1-kibana3",
     "@elastic/numeral": "npm:@amoo-miki/numeral@2.6.0",
-    "@elastic/request-crypto": "2.0.0",
+    "@elastic/request-crypto": "2.0.2",
     "@elastic/safer-lodash-set": "0.0.0",
     "@hapi/accept": "^5.0.2",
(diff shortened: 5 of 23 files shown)


src/plugins/chat/public/services/chat_event_handler.ts
```diff
@@ -235,12 +246,46 @@ export class ChatEventHandler {
     // Add to pending map for args accumulation
     this.pendingToolCalls.set(toolCallId, toolCall);
 
-    // Use the last TEXT_MESSAGE_START message ID for association
-    const targetMessageId = parentMessageId || this.lastTextMessageStartId;
+    // Strategy 1: Use explicitly provided parent message ID
+    // This is the most reliable approach when the backend provides it
+    if (parentMessageId) {
```
xluo-aws, line 251: Under what situation the parentMessageId will not available?

src/plugins/chat/public/services/chat_event_handler.ts
```diff
@@ -235,12 +246,46 @@ export class ChatEventHandler {
...
+      // If the last assistant message appears after the last user message,
+      // it means this tool call belongs to the current conversation turn
+      if (lastAssistantTextMessageIndex > lastUserMessageIndex) {
+        this.addToolCallToMessage(this.lastTextMessageStartId, toolCall);
+        return;
+      }
     }
+
+    // Strategy 3: Create a new assistant message placeholder
+    // This handles the case where the LLM responds with tool calls but without any text message.
```
xluo-aws, line 276: is this an abnormal case because LLM didn't follow our prompt?

src/plugins/chat/public/services/chat_event_handler.ts
wanglam: This `parentMessageId` comes from the chat agent event stream and depends on the chat agent implementation. Typically, it has a valid value when the agent responds with a message that initiates tool calls. All tool calls from the same response share the same `parentMessageId`.

Previously, the implementation would fall back to the latest chat agent message when `parentMessageId` wasn't provided. In this PR, I added logic to check if there are any user input messages after the latest chat agent message. If so, it creates a virtual chat agent message (frontend-only) to properly group these tool calls.

src/plugins/chat/public/services/chat_event_handler.ts
wanglam: This depends on the chat agent implementation. From my perspective, if the user asks the chat agent to do something straightforward, having no text message before the tool calling section makes sense to me.

Hailong-am:
can you paste a screenshot of before and after?

wanglam:
Hi @Hailong-am , I've added the screenshots of before and after. Feel free to help me review it. Thank you.

Hailong-am (approved):

PR #11157, opened 2026-01-12
Add thinking message in chat conversation

arjunkumargiri:
### Description

- This PR adds a "Thinking..." loading message that appears immediately when users send a message in the chat interface, providing better UX feedback before the AI response arrives.

### Core Implementation
- **Chat Window (`chat_window.tsx`)**: Added loading message logic to `handleSend` and `handleResendMessage` functions
  - Creates temporary loading assistant message with `loading-` ID prefix
  - Adds to timeline immediately after user message
  - Removes when first response arrives, on error, or completion

- **Chat Messages (`chat_messages.tsx`)**: Updated message rendering to handle loading states
  - Detects loading messages by ID prefix
  - Displays "Thinking..." for loading messages
  - Prevents rendering content, suggestions, and tool calls for loading messages

### Testing & Debugging
- **Unit Tests**: Added comprehensive tests for loading message functionality in `chat_messages.test.tsx`

## User Experience Improvement

**Before**: Users send message → wait with no feedback → response appears
**After**: Users send message → "Thinking..." appears immediately → response replaces loading message

## Summary by CodeRabbit

* **New Features**
  * Shows a dedicated loading/thinking message after sending (including resend) to indicate assistant streaming.

* **Bug Fixes**
  * Hides assistant content, suggestions, and tool/result rows while the loading message is present.
  * Ensures loading entries are removed on first response, error, or completion.

* **Tests**
  * Expanded async test coverage for loading lifecycle, send/resend flows, streaming behavior, input handling, and cleanup.


src/plugins/chat/public/components/chat_window.tsx
wanglam, line 115: Nit: Do we need to filter out the loading message here?

wanglam:
Hi @arjunkumargiri , could you help to fix this changelog workflow?

SuZhou-Joe (approved):
I am thinking if we should use `activity message` for such requirement: https://docs.ag-ui.com/concepts/messages#activity-messages . The benefit is that we can remove those `.filter` lines and only render the loading message when it is the last message of the timeline.

arjunkumargiri:
@SuZhou-Joe good suggestion. I believe we need this supported in AG UI backend agent, we can check with ML commons plugin to support the feature. 

+ @jiapingzeng

jiapingzeng:
I think using `ActivityMessage` makes sense for thinking message. At the moment, when parsing chat history, ml-commons only parses user, assistant, and tool messages. Messages with other roles are dropped: [here](https://github.com/opensearch-project/ml-commons/blob/feature/3.4-release-prep/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAGUIAgentRunner.java#L197-L236) and [here](https://github.com/opensearch-project/ml-commons/blob/feature/3.4-release-prep/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLAGUIAgentRunner.java#L305-L321). So using `ActivityMessage` should be safe without code change in ml-commons.

Did a quick test with the following input and verified that it still works:
```
{
    "threadId": "thread-1760484425669-pj44w6ive",
    "runId": "run-1760484493143-gu67vx1jl",
    "messages": [
        {
            "id": "msg-1760484430561-epujp4bi6",
            "role": "user",
            "content": "what tools do you have"
        },
        {
            "id": "12345",
            "role": "activity",
            "activityType": "PLAN",
            "content": "thinking"
        }
    ],
    "tools": [...],
    "context": [...],
    "state": {},
    "forwardedProps": {}
}
```

PR #11172, opened 2026-01-14
[Explore vis] Migrate scatter to ECharts

lunalzm:
### Description

This PR migrates the scatter chart visualization in the explore plugin from vega-lite to echarts.

### TwoMetricScatter
image

### TwoMetricOneCateScatter
image

### ThreeMetricOneCateScatter
image

## Changelog

- feat: Migrate scatter chart to echarts

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/11164.yml
+++ b/changelogs/fragments/11164.yml
@@ -1,0 +1,2 @@
+refactor:
+- Echarts implementation of histogram visualization ([#11164](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11164))
--- a/changelogs/fragments/11192.yml
+++ b/changelogs/fragments/11192.yml
@@ -1,0 +1,2 @@
+feat:
+- Migrate heatmap to echarts ([#11192](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11192))
--- a/src/plugins/explore/public/components/visualizations/bar/bar_chart_utils.ts
+++ b/src/plugins/explore/public/components/visualizations/bar/bar_chart_utils.ts
@@ -22,6 +22,7 @@ import { BarChartStyle } from './bar_vis_config';
 import { getColors, DEFAULT_GREY } from '../theme/default_colors';
 import { BaseChartStyle, PipelineFn, EChartsSpecState } from '../utils/echarts_spec';
 import { getSeriesDisplayName } from '../utils/series';
+import { HistogramChartStyle } from '../histogram/histogram_vis_config';
 
 export const inferTimeIntervals = (data: Array<Record<string, any>>, field: string | undefined) => {
   if (!data || data.length === 0 || !field) {
@@ -195,18 +196,29 @@ export const buildThresholdColorEncoding = (
   return colorLayer;
 };
 
+type Options =
+  | {
+      kind: 'bar';
+      styles: BarChartStyle;
+      categoryField: string;
+      seriesFields: string[] | ((headers?: string[]) => string[]);
+    }
+  | {
+      kind: 'histogram';
+      styles: HistogramChartStyle;
+      categoryField: string;
+      seriesFields: string[] | ((headers?: string[]) => string[]);
+    };
+
 /**
  * Create bar series configuration
  */
-export const createBarSeries = <T extends BaseChartStyle>({
-  styles,
-  categoryField,
-  seriesFields,
-}: {
-  styles: BarChartStyle;
-  categoryField: string;
-  seriesFields: string[] | ((headers?: string[]) => string[]);
-}): PipelineFn<T> => (state) => {
+export const createBarSeries = <T extends BaseChartStyle>(options: Options): PipelineFn<T> => (
+  state
+) => {
+  const { styles, categoryField } = options;
+  let seriesFields = options.seriesFields;
+
   const { axisColumnMappings, transformedData = [] } = state;
   const newState = { ...state };
 
@@ -214,7 +226,24 @@ export const createBarSeries = <T extends BaseChartStyle>({
     seriesFields = seriesFields(transformedData[0]);
   }
 
-  const thresholdLines = generateThresholdLines(styles?.thresholdOptions, styles?.switchAxes);
+  const thresholdLines =
+    options.kind === 'bar'
+      ? generateThresholdLines(options.styles?.thresholdOptions, options.styles?.switchAxes)
+      : generateThresholdLines(options.styles?.thresholdOptions);
+
+  const encodeX =
+    options.kind === 'bar' ? adjustOppositeSymbol(options.styles?.switchAxes, 'x') : 'x';
+  const encodeY =
+    options.kind === 'bar' ? adjustOppositeSymbol(options.styles?.switchAxes, 'y') : 'y';
+
+  let barWidth: string | undefined;
+  if (styles.barSizeMode === 'manual') {
+    barWidth = `${(styles.barWidth || 0.7) * 100}%`;
+  } else {
+    if (options.kind === 'histogram') {
+      barWidth = '99%';
+    }
+  }
 
   const series = seriesFields.map((seriesField, index) => {
     const name = getSeriesDisplayName(seriesField, Object.values(axisColumnMappings));
@@ -226,11 +255,11 @@ export const createBarSeries = <T extends BaseChartStyle>({
       },
       name,
       encode: {
-        [adjustOppositeSymbol(styles?.switchAxes, 'x')]: categoryField,
-        [adjustOppositeSymbol(styles?.switchAxes, 'y')]: seriesField,
+        [encodeX]: categoryField,
+        [encodeY]: seriesField,
       },
       // TODO: barWidth and barCategoryGap seems are exclusive, we need to revise the current UI for this config
-      barWidth: styles.barSizeMode === 'manual' ? `${(styles.barWidth || 0.7) * 100}%` : undefined,
+      barWidth,
       barCategoryGap:
         styles.barSizeMode === 'manual' ? `${(styles.barPadding || 0.1) * 100}%` : undefined,
       ...(index === 0 && thresholdLines),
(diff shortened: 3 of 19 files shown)


src/plugins/explore/public/components/visualizations/scatter/scatter_chart_utils.ts
```diff
@@ -0,0 +1,450 @@
...
+      ...thresholdLines,
+    },
+  ] as any[];
+
+  const legendPosition = styles.legendPosition || 'bottom';
+  newState.visualMap = {
+    show: styles.addLegend === true,
+    type: 'piecewise',
+    categories,
+    dimension: 2,
```
ruanyl: What does `2` refer to?

src/plugins/explore/public/components/visualizations/scatter/scatter_chart_utils.ts
```diff
@@ -0,0 +1,450 @@
...
+              '#ee6666',
+              '#73c0de',
+              '#3ba272',
+              '#fc8452',
+              '#9a60b4',
+              '#ea7ccc',
+              '#c2c2c2',
+            ].slice(0, categories.length)
+          : undefined,
+    },
```
ruanyl: We should not hard code colors, they should be colors in theme, if you remove these color, can it still work?

src/plugins/explore/public/components/visualizations/scatter/scatter_chart_utils.ts
```diff
@@ -0,0 +1,450 @@
...
+    textStyle: {
+      color: '#333',
```
ruanyl: The same here, I think we can remove it

src/plugins/explore/public/components/visualizations/scatter/scatter_chart_utils.ts
```diff
@@ -0,0 +1,450 @@
...
+      emphasis: {
+        focus: 'self',
+        scale: 1.2,
+      },
+      ...thresholdLines,
+    },
+  ] as any[];
+
+  const legendPosition = styles.legendPosition || 'bottom';
+  newState.visualMap = {
```
ruanyl: Implement category with multi-series? It's simple based on my understanding

src/plugins/explore/public/components/visualizations/scatter/to_expression.ts
```diff
@@ -42,6 +55,36 @@ export const createTwoMetricScatter = (
...
+    const yField = axisConfig.yAxis?.column;
+
+    if (!xField || !yField) throw Error('Missing axis config for scatter chart');
+
+    const allColumns = [...Object.values(axisColumnMappings ?? {}).map((m) => m.column)];
+
+    const result = pipe(
+      transform(convertTo2DArray(allColumns)),
+      createBaseConfig({ title: `${axisConfig.xAxis?.name} vs ${axisConfig.yAxis?.name}` }),
+      buildAxisConfigs,
```
Qxisylolo, line 72: This simple scatter also supports mapping color to thresholds, let's add buildVisMap to implement it.

src/plugins/explore/public/components/visualizations/scatter/scatter_chart_utils.ts
Qxisylolo: What about writing spec like:
By such way, we can use pivot and you don't have to hard code colors.
(field 'y' is no longer needed, we can delete it in pivoting)
```
const data = [
  ['x', 'y', 'A', 'B', 'C'],
  [3.275154, 2.957587, 2.957587, null, null],
  [-3.344465, 2.603513, null, 2.603513, null],
  [0.355083, -3.376585, null, null, -3.376585],
  [1.852435, 3.547351, 3.547351, null, null],
  [-2.078973, 2.552013, 2.552013, null, null],
];


option = {
  dataset: [{ source: data }],

  xAxis: {},
  yAxis: {},
  tooltip: { trigger: 'item' },

  series: [
    {
      name: 'A',
      type: 'scatter',
      encode: {
        x: 'x',
        y: 'A'
      },
    },
    {
      name: 'B',
      type: 'scatter',
      encode: {
        x: 'x',
        y: 'B'
      },
    },
    {
      name: 'C',
      type: 'scatter',
      encode: {
        x: 'x',
        y: 'C'
      },
    }
  ]
};
```

src/plugins/explore/public/components/visualizations/scatter/to_expression.ts
```diff
@@ -109,6 +160,43 @@ export const createTwoMetricOneCateScatter = (
...
+    const yField = axisConfig.yAxis?.column;
+    const colorField = axisColumnMappings?.[AxisRole.COLOR]?.column;
+
+    if (!xField || !yField || !colorField)
+      throw Error('Missing axis config for colored scatter chart');
+
+    const allColumns = [...Object.values(axisColumnMappings ?? {}).map((m) => m.column)];
+
+    const result = pipe(
+      transform(convertTo2DArray(allColumns)),
```
Qxisylolo: Let's add pivot here, I assume you can do something like: 
```
        pivot({
          groupBy: xField,
          pivot: colorField,
          field: yField,
        }),
```

src/plugins/explore/public/components/visualizations/scatter/scatter_chart_utils.ts
```diff
@@ -0,0 +1,283 @@
...
+        focus: 'self',
+        scale: 1.2,
+      },
+      ...thresholdLines,
+    },
+  ] as ScatterSeriesOption[];
+
+  newState.series = series;
+
+  (newState as any).scatterTooltipConfig = {
```
ruanyl: `scatterTooltipConfig` is added to `newState`, but I couldn't find how it's used, did I miss anything?

src/plugins/explore/public/components/visualizations/scatter/scatter_chart_utils.ts
lunalzm: Dual visualMap mapping:
visualMap[0]: piecewise type, dimension: 2 → controls color
visualMap[1]: continuous type, dimension: 3 → controls size

src/plugins/explore/public/components/visualizations/scatter/scatter_chart_utils.ts
```diff
@@ -0,0 +1,480 @@
...
+    trigger: 'item',
+    formatter: (params: any) => {
+      if (params.value && Array.isArray(params.value) && params.value.length >= 3) {
+        const xValue = params.value[0];
+        const yValue = params.value[1];
+        const sizeValue = params.value[2];
+        const seriesName = params.seriesName;
+        return `${xAxisName}: ${xValue}<br/>${yAxisName}: ${yValue}<br/>${colorAxisName}: ${seriesName}<br/>${sizeAxisName}: ${sizeValue}`;
+      }
+      return `${xAxisName}: ${params.value || 'N/A'}<br/>${yAxisName}: ${params.name || 'N/A'}`;
```
Hailong-am: will this has XSS (Cross-Site Scripting) vulnerability if parameters container dangerous html? should we escape it?

src/plugins/explore/public/components/visualizations/scatter/scatter_chart_utils.ts
```diff
@@ -0,0 +1,480 @@
...
+      if (params.value && Array.isArray(params.value) && params.value.length >= 2) {
+        const xValue = params.value[0];
+        const seriesName = params.seriesName;
+        const yValueIndex = pivotHeader.indexOf(String(seriesName));
+        const yValue = yValueIndex >= 0 ? params.value[yValueIndex] : null;
+
+        return `${xAxisName}: ${xValue}<br/>${yAxisName}: ${yValue}<br/>${colorAxisName}: ${seriesName}`;
+      }
+      return `${xAxisName}: ${params.value || 'N/A'}<br/>${yAxisName}: ${params.name || 'N/A'}`;
+    },
```
Hailong-am: same here, could you double confirm this won't have Xss vulnerability

src/plugins/explore/public/components/visualizations/scatter/scatter_chart_utils.ts
```diff
@@ -0,0 +1,480 @@
...
+  categories.forEach((category) => {
+    seriesData[String(category)] = [];
+  });
+
+  // Group data points by color category
+  dataRows.forEach((row) => {
+    const x = row[xFieldIndex];
+    const y = row[yFieldIndex];
+    const category = String(row[colorFieldIndex] || 'undefined');
+    const size = Number(row[sizeFieldIndex]) || 0;
```
Hailong-am: if the size is 0, will it still show up?

src/plugins/explore/public/components/visualizations/scatter/to_expression.ts
```diff
@@ -234,9 +359,9 @@ export const createThreeMetricOneCateScatter = (
           ? {
               title: styles?.legendTitleForSize,
               orient: styles?.legendPosition,
-              symbolLimit: 10,
+              offset: 40,
```
Hailong-am: nit: magic number 40, what does it mean?

src/plugins/explore/public/components/visualizations/scatter/scatter_chart_utils.ts
lunalzm: Good question. After checking, we decided not to customize the tooltip for now and to stick with the default ECharts tooltip behavior. So I'll remove this part.

src/plugins/explore/public/components/visualizations/scatter/scatter_chart_utils.ts
lunalzm: I’ll update the logic to explicitly handle this, for example:

```
const size = Number(row[sizeFieldIndex]);
if (isNaN(size) || size <= 0) return;
```

This way, points with invalid values or size <= 0 will be ignored and won’t be rendered.

src/plugins/explore/public/components/visualizations/scatter/to_expression.ts
lunalzm: Good catch! The `40` here is the vertical offset used to separate the size legend from the color legend, so they don’t overlap when both are enabled. I’ll extract it into a named constant (e.g. `LEGEND_VERTICAL_OFFSET`)

PR #11183, opened 2026-01-16
make single-version=ignore by default

angle943:
### Description

- follow up to #11179 , this will make single-version=ignore as default.

## Changelog
- infra: make single-version=ignore as default

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/11212.yml
+++ b/changelogs/fragments/11212.yml
@@ -1,0 +1,2 @@
+fix:
+- Visualizations should hide ailias type ([#11212](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11212))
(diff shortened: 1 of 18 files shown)


LDrago27:
@angle943 A QQ with this in place do we still need this setting. It is also doing something similar for plugins and ignoring the version mismatch. 

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/5a4b9146286b5a6251154e63c41d6ee7ef6f9dd5/src/core/server/plugins/discovery/plugin_manifest_parser.ts#L50

angle943:
I don't think we need that one, but let's remove it later, as removing it is a breaking change

PR #11184, opened 2026-01-16
[Explore vis] add theme config for echarts

ruanyl:
### Description
1. add theme config for echarts
2. explore_embeddable to support echarts rendering

## Changelog

- chore: add echarts theme
- fix: explore embeddable not able to render echarts

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/visualizations/area/area_chart_utils.ts
```diff
@@ -122,7 +123,7 @@ export const createFacetAreaSeries = <T extends BaseChartStyle>({
       stack: `Total_${index}`, // Use unique stack name for each facet
       connectNulls: true,
       areaStyle: {
-        opacity: styles.areaOpacity || 0.3,
+        opacity: styles.areaOpacity || DEFAULT_OPACITY3,
```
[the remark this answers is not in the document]
SuZhou-Joe: @ruanyl , I think this comment makes sense. Guess a typo.

src/plugins/explore/public/components/visualizations/theme/default.ts
```diff
@@ -3,11 +3,82 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import * as echarts from 'echarts';
```
SuZhou-Joe, line 6: Not sure if it will impact performance, and this kind of import may make tree shaking useless I am afraid.

src/plugins/explore/public/embeddable/explore_embeddable.tsx
```diff
@@ -418,8 +419,10 @@ export class ExploreEmbeddable
             dateColumns,
             styles || styleOptions,
             selectedChartType,
-            axesMapping
+            axesMapping,
+            searchContext.timeRange
```
SuZhou-Joe, line 423: A little bit weird why we put timeRange into spec params, and no consumer found for this change.

src/plugins/explore/public/components/visualizations/echarts_render.tsx
```diff
@@ -53,6 +54,7 @@ export const EchartsRender = ({ spec }: Props) => {
         },
         { notMerge: true } // this is a must to update compulsorily otherwise will merge with previous option
       );
+      instance.setTheme(DEFAULT_THEME);
```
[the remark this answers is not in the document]
ruanyl: `setTheme` is an API since `v6.0.0`

src/plugins/explore/public/embeddable/explore_embeddable.tsx
ruanyl: It will be used by rendering the full time axis, I think this is a bug needs to be fixed in a separate PR. 
For example, if user selected last two years, but search result contains only data of last year, this timeRange is used to show the full time range(2 years) on chart

src/plugins/explore/public/components/visualizations/theme/default.ts
ruanyl: I think you have good point, I will look into this separetely, I think we can shrink the size based on the the official doc: https://echarts.apache.org/handbook/en/basics/import

wanglam (approved):

src/plugins/explore/public/embeddable/explore_embeddable.tsx
SuZhou-Joe: Get it, thanks.

PR #11194, opened 2026-01-19
feat: Add slash command system with autocomplete and user confirmations for trigger investigation from chatbot

Hailong-am:
### Description

This PR adds a slash command system to the OpenSearch Dashboards chatbot, enabling users to trigger investigations and execute actions directly from the chat interface. It introduces a confirmation workflow for tool executions and improves the chat UX with autocomplete and command hints.

  **Key Features Added**

  1. Slash Command System

  - Command Registry: Extensible registry allowing plugins to register custom slash commands
  - Autocomplete UI: Dropdown menu appears when typing / with fuzzy matching
  - Ghost Text Hints: Inline command completion suggestions with Tab/Space to accept
  - Keyboard Navigation: Arrow keys to navigate, Enter to select, Escape to dismiss

  2. User Confirmation Workflow

  - ConfirmationService: New service to manage user approval for sensitive tool executions
  - Confirmation UI: Sticky confirmation banner with approve/reject buttons
  - Observable Pattern: Uses RxJS BehaviorSubject for reactive confirmation state

  3. Tool Call Restoration

  - Resume Unfinished Tools: Automatically restores and executes incomplete tool calls on page reload
  - Seamless UX: Users can close/reopen chat without losing in-progress investigations

  4. Message Display Improvements

  - Raw Message Tracking: Separates display message (user sees /investigate) from LLM message (processed command)
  - Better Context: Users see what they typed, not internal command transformations

## Testing the changes

https://github.com/user-attachments/assets/d2bb71f7-1c81-4a16-8267-14905f2194db

## Changelog

- feat: Add slash command system with autocomplete and user confirmations for chatbot

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/chat/public/components/message_row.test.tsx
+++ b/src/plugins/chat/public/components/message_row.test.tsx
@@ -235,4 +235,143 @@ describe('MessageRow', () => {
       expect(cursor).not.toBeInTheDocument();
     });
   });
+
+  describe('resend button visibility', () => {
+    const mockOnResend = jest.fn();
+
+    beforeEach(() => {
+      jest.clearAllMocks();
+    });
+
+    it('should show resend button for regular user messages without rawMessage', () => {
+      const message: Message = {
+        id: 'msg-15',
+        role: 'user',
+        content: 'Hello, how are you?',
+      };
+
+      render(<MessageRow message={message} onResend={mockOnResend} />);
+
+      const resendButton = screen.getByLabelText('Resend message');
+      expect(resendButton).toBeInTheDocument();
+    });
+
+    it('should show resend button even if content starts with "/" (no rawMessage)', () => {
+      const message: Message = {
+        id: 'msg-16',
+        role: 'user',
+        content: '/help',
+      };
+
+      render(<MessageRow message={message} onResend={mockOnResend} />);
+
+      const resendButton = screen.getByLabelText('Resend message');
+      expect(resendButton).toBeInTheDocument();
+    });
+
+    it('should NOT show resend button when rawMessage is a slash command', () => {
+      const message: Message = {
+        id: 'msg-17',
+        role: 'user',
+        content: 'processed content',
+        rawMessage: '/help',
+      };
+
+      render(<MessageRow message={message} onResend={mockOnResend} />);
+
+      const resendButton = screen.queryByLabelText('Resend message');
+      expect(resendButton).not.toBeInTheDocument();
+    });
+
+    it('should NOT show resend button for slash command with arguments in rawMessage', () => {
+      const message: Message = {
+        id: 'msg-18',
+        role: 'user',
+        content: 'processed: search results',
+        rawMessage: '/search query string',
+      };
+
+      render(<MessageRow message={message} onResend={mockOnResend} />);
+
+      const resendButton = screen.queryByLabelText('Resend message');
+      expect(resendButton).not.toBeInTheDocument();
+    });
+
+    it('should NOT show resend button for slash command with leading whitespace in rawMessage', () => {
+      const message: Message = {
+        id: 'msg-19',
+        role: 'user',
+        content: 'processed content',
+        rawMessage: '  /help',
+      };
+
+      render(<MessageRow message={message} onResend={mockOnResend} />);
+
+      const resendButton = screen.queryByLabelText('Resend message');
+      expect(resendButton).not.toBeInTheDocument();
+    });
+
+    it('should show resend button when rawMessage contains "/" but not at the start', () => {
+      const message: Message = {
+        id: 'msg-20',
+        role: 'user',
+        content: 'What is the path/to/file?',
+        rawMessage: 'What is the path/to/file?',
+      };
+
+      render(<MessageRow message={message} onResend={mockOnResend} />);
+
+      const resendButton = screen.getByLabelText('Resend message');
+      expect(resendButton).toBeInTheDocument();
+    });
+
+    it('should NOT show resend button for assistant messages', () => {
+      const message: Message = {
+        id: 'msg-21',
+        role: 'assistant',
+        content: 'Here is my response',
+      };
+
+      render(<MessageRow message={message} onResend={mockOnResend} />);
+
+      const resendButton = screen.queryByLabelText('Resend message');
+      expect(resendButton).not.toBeInTheDocument();
+    });
+
+    it('should NOT show resend button when onResend is not provided', () => {
+      const message: Message = {
+        id: 'msg-22',
+        role: 'user',
+        content: 'Hello',
+      };
+
+      render(<MessageRow message={message} />);
+
+      const resendButton = screen.queryByLabelText('Resend message');
+      expect(resendButton).not.toBeInTheDocument();
+    });
+
+    it('should show resend button for multimodal content (not slash commands)', () => {
+      const message: Message = {
+        id: 'msg-23',
+        role: 'user',
+        content: [
+          {
+            type: 'text',
+            text: 'Check this image',
+          },
+          {
+            type: 'binary',
+            data: 'base64data',
+            mimeType: 'image/png',
+          },
+        ],
+      };
+
+      render(<MessageRow message={message} onResend={mockOnResend} />);
+
+      const resendButton = screen.getByLabelText('Resend message');
+      expect(resendButton).toBeInTheDocument();
+    });
+  });
 });
--- a/src/plugins/chat/public/components/message_row.tsx
+++ b/src/plugins/chat/public/components/message_row.tsx
@@ -28,6 +28,15 @@ export const MessageRow: React.FC<MessageRowProps> = ({
     }
   };
 
+  // Check if the message is a slash command
+  // A message is a slash command if rawMessage exists and starts with "/"
+  const isSlashCommand = () => {
+    if (message.role === 'user' && 'rawMessage' in message && message.rawMessage) {
+      return typeof message.rawMessage === 'string' && message.rawMessage.trim().startsWith('/');
+    }
+    return false;
+  };
+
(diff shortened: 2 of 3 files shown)


src/plugins/chat/public/components/chat_input.tsx
```diff
@@ -26,18 +28,150 @@ export const ChatInput: React.FC<ChatInputProps> = ({
...
+  const [showCommandMenu, setShowCommandMenu] = useState(false);
+  const [commandSuggestions, setCommandSuggestions] = useState<SlashCommand[]>([]);
+  const [selectedCommandIndex, setSelectedCommandIndex] = useState(0);
+  const [ghostText, setGhostText] = useState('');
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  // Update command suggestions when input changes
+  useEffect(() => {
+    if (input.startsWith('/')) {
+      const hasSpace = input.includes(' ');
```
SuZhou-Joe: Will it match `/something something`?

src/plugins/chat/public/components/chat_input.tsx
```diff
@@ -26,18 +28,150 @@ export const ChatInput: React.FC<ChatInputProps> = ({
...
+        }
+      }
+    } else {
+      setShowCommandMenu(false);
+      setCommandSuggestions([]);
+      setGhostText('');
+    }
+  }, [input]);
+
+  const handleKeyDown = (e: React.KeyboardEvent) => {
```
SuZhou-Joe: Shall we create a new component and handle the arrow up / down inside that component? The logic is kind of complicated here for chat_input.

src/plugins/chat/public/components/chat_window.tsx
```diff
@@ -309,9 +344,20 @@ const ChatWindowContent = React.forwardRef<ChatWindowInstance, ChatWindowProps>(
         timeline={timeline}
         isStreaming={isStreaming}
         onResendMessage={handleResendMessage}
+        onApproveConfirmation={handleApproveConfirmation}
+        onRejectConfirmation={handleRejectConfirmation}
         {...enhancedProps}
       />
 
+      {/* Sticky confirmation message - positioned above chat input */}
+      {pendingConfirmation && (
```
SuZhou-Joe, line 375: Seems we can leverage https://docs.ag-ui.com/concepts/messages#activity-messages for such frontend message.

src/plugins/chat/public/components/chat_header_button.tsx
```diff
@@ -47,6 +55,14 @@ export const ChatHeaderButton = React.forwardRef<ChatHeaderButtonInstance, ChatH
     const chatWindowRef = useRef<ChatWindowInstance>(null);
     const flyoutMountPoint = useRef(null);
 
+    // Instantiate confirmation service
+    const confirmationService = useMemo(() => new ConfirmationService(), []);
```
SuZhou-Joe: Since it is a service, shall we manage the lifecycle in chat/plugin.ts instead of a component?

src/plugins/chat/public/services/chat_event_handler.ts
```diff
@@ -47,12 +47,13 @@ export class ChatEventHandler {
 
   constructor(
     private assistantActionService: AssistantActionService,
-    private chatService: ChatService | null,
+    private chatService: ChatService,
     private onTimelineUpdate: (updater: (prev: Message[]) => Message[]) => void,
     private onStreamingStateChange: (isStreaming: boolean) => void,
-    private getTimeline: () => Message[]
+    private getTimeline: () => Message[],
+    confirmationService: any
```
SuZhou-Joe: We should have one type definition for this field, no need for any I think.

src/plugins/chat/public/components/chat_header_button.tsx
Hailong-am: good catch, we should move it to plugin.ts

src/plugins/chat/public/services/confirmation_service.ts
```diff
@@ -0,0 +1,146 @@
...
+   * Check if there are any pending confirmations
+   */
+  hasPendingConfirmations(): boolean {
+    return this.pendingConfirmations$.getValue().length > 0;
+  }
+
+  /**
+   * stop all confirmations (e.g. on unmount)
+   */
+  cleanAll(): void {
```
SuZhou-Joe, line 123: Do we want to clean up `responseCallbacks`?

src/plugins/chat/public/services/tool_executor.ts
```diff
@@ -4,13 +4,16 @@
  */
 
 import { AssistantActionService } from '../../../context_provider/public';
+import { ConfirmationService } from './confirmation_service';
```
SuZhou-Joe, line 7: Is it possible to use a userConfirmation tool for the confirm action? I am thinking if ConfirmationService is too much for a confirmation message.

src/plugins/chat/public/components/chat_input.tsx
Hailong-am: Thanks for suggestion, will move it to a hook/component

src/plugins/chat/public/components/chat_window.tsx
Hailong-am: It's not part of ag-ui messages. It's a separate component to display confirm message before running some of front-end tools

src/plugins/chat/public/services/tool_executor.ts
Hailong-am: Using `userConfirmation` tool could be an option too,  but it depends on LLM to decide when to call it. For my case like create investigation, it always need user confirmation to proceed the creating and it's a fix flow. For such fixed workflow, we can do it with code instead of relay on LLM to decide the flow.

src/plugins/chat/public/services/confirmation_service.ts
Hailong-am: It should have the clean up already. It will resolve all pending callbacks(promise)

```ts
this.pendingConfirmations$.getValue().forEach((request) => {
      const callback = this.responseCallbacks.get(request.id);
      if (callback) {
        callback({
          id: request.id,
          approved: false,
          cancelled: true,
        });
      }
    });
```

src/plugins/chat/public/components/chat_window.tsx
```diff
@@ -104,52 +115,94 @@ const ChatWindowContent = React.forwardRef<ChatWindowInstance, ChatWindowProps>(
...
+          // restore unfinished tool call by triggering events
+          const unfinishedToolCalls = lastMessage.toolCalls.filter(toolCall => {
+            // Check if there's no corresponding tool result message
+            const hasToolResult = currentMessages.some(msg =>
+              msg.role === 'tool' && msg.toolCallId === toolCall.id
+            );
+            return !hasToolResult;
+          });
+
+          // Trigger tool call events for unfinished tool calls
```
wanglam, line 133: Q: Is this restore for agent tools? Will front end tools have tool results?

Hailong-am: Front-end tools also have tool results, like investigation tool will have tool result for the investigation creation result. 

The restore will be applied for all tools include agent tool/front end tool. The reason to have this to restore the state of UI like show confirm message for unconfirmed tool call

src/plugins/chat/public/services/chat_event_handler.ts
```diff
@@ -328,8 +330,35 @@ export class ChatEventHandler {
...
+      // Check if tool execution was cancelled (e.g., due to cleanup)
+      if (result.cancelled) {
+        this.pendingToolCalls.delete(toolCallId);
+        return;
+      }
+
+      if (result.userRejected) {
+        // User rejected the tool execution
+        this.assistantActionService.updateToolCallState(toolCallId, {
+          status: 'failed',
```
wanglam: Nit: Do we need to have a separate status for user reject actions? Then LLM may know user doesn't want to do this instead of failed to execute this tool

src/plugins/chat/public/services/chat_service.ts
```diff
@@ -55,6 +55,9 @@ export class ChatService {
   // Subscription to assistant action service for tool updates
   private toolSubscription?: Subscription;
 
+  // Cache for datasourceId to avoid repeated lookups
+  private cachedDataSourceId?: string;
```
wanglam, line 59: Are there any places to clear this cachedDataSourceId? For example, if the data source id changed in the context provider. Will we clean the cachedDataSourceId?

src/plugins/chat/public/services/slash_commands.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  getSuggestions(input: string): SlashCommand[] {
+    if (!input.startsWith('/')) return [];
+
+    const query = input.slice(1).toLowerCase();
+    return this.getAll().filter((cmd) => cmd.command.toLowerCase().startsWith(query));
+  }
+
+  async execute(input: string): Promise<{ handled: boolean; message?: string }> {
```
wanglam, line 45: Q: How about fire a new command execute during previous command execution?

src/plugins/chat/public/services/chat_service.ts
Hailong-am: For each LLM call, it will get latest datasource either from context provider or default datasource of workspace/application.
This cache been used for front-end tool call, it will passed to the tool and re-use whatever last LLM call used. Based on this, it will always pick latest value.

src/plugins/chat/public/services/slash_commands.ts
Hailong-am: That's not doable as chat input is disabled when a command was executing

PR #11135, opened 2026-01-06
feat: support gzip for OpenSearch responses

joshuali925:
### Description

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11130

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11130

## Changelog

- feat: support gzip for OpenSearch responses

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

## Summary by CodeRabbit

* **New Features**
  * Enabled gzip compression for OpenSearch responses, reducing bandwidth usage and improving data transfer performance between OpenSearch and clients.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #11201, opened 2026-01-20
[Explore] Change PromQL generation to use ag-ui agent

joshuali925:
### Description

This PR makes PromQL generation to use the `chat` plugin's proxy route and calls an ag-ui agent. This allows LLM to call tools and get metrics metadata containing specific keywords on demand.

It makes the following changes

- partially reverts 077ba23 to remove API to directly call query assist agent
- add `chat.observabilityAgentId` config. If present, then chat proxy API will be enabled
- if the agent request from browser includes `body.forwardedProps.language` (`PROMQL` for prometheus, can expand to support other languages), the handler will use configured `observabilityAgentId`. Otherwise, it preserves original behavior and uses `configuredAgentId`
- handles gzip response streaming in chat proxy API, possibly from #11135 
- refactors resources API to support time range

Regarding query generation, 

- Currently system prompts are passed from browser to the chat proxy API as user prompt. This is because we want to use the same Observability agent for different purposes. Ag-UI agent implementation currently ignores system prompts
- a `search_prometheus_metadata` tool will be included in the agent. It takes a regex pattern to filter metrics by name and returns the list of matched metrics (default max 20), their labels (default max 20), and sample values (default max 5). Max is configurable by the agent. Currently Ag-UI agent implementation only includes the latest tool call result in the prompt, we cannot use multiple tools to answer a question
   ```json5
  {
    name: PromQLToolName.SEARCH_PROMETHEUS_METADATA,
    description:
      'Search Prometheus metadata. Returns matched metrics with their labels and sample label values. Use this to discover available metrics before writing a PromQL query.',
    parameters: {
      type: 'object',
      properties: {
        query: {
          type: 'string',
          description: 'Regex pattern to filter metrics by name.',
        },
        metricsLimit: {
          type: 'number',
          description: 'Maximum number of metrics to return. Defaults to 20.',
        },
        labelsLimit: {
          type: 'number',
          description: 'Maximum number of labels per metric. Defaults to 20.',
        },
        valuesLimit: {
          type: 'number',
          description: 'Maximum number of sample values per label. Defaults to 5.',
        },
      },
      required: [],
    },
  },
   ```

## Screenshot

image

## Testing the changes

added unit tests, tested locally

## Changelog

- feat: Change PromQL generation to use ag-ui agent

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/utils/query_assist/promql_tool_handlers.ts
+++ b/src/plugins/explore/public/application/utils/query_assist/promql_tool_handlers.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { DataPublicPluginStart } from '../../../../../data/public';
+import { DataPublicPluginStart, TimeRange } from '../../../../../data/public';
 import { PromQLToolName } from './promql_tools';
 
 /** Default limit for metrics */
@@ -36,12 +36,22 @@ interface SearchPrometheusMetadataResult {
 }
 
 interface PrometheusResourceClient {
-  getMetrics: (dataConnectionId: string) => Promise<string[]>;
+  getMetrics: (dataConnectionId: string, timeRange?: TimeRange) => Promise<string[]>;
   getMetricMetadata: (
-    dataConnectionId: string
+    dataConnectionId: string,
+    metric?: string,
+    timeRange?: TimeRange
   ) => Promise<Record<string, Array<{ type: string; help: string }>>>;
-  getLabels: (dataConnectionId: string, metric?: string) => Promise<string[]>;
-  getLabelValues: (dataConnectionId: string, label: string, metric?: string) => Promise<string[]>;
+  getLabels: (
+    dataConnectionId: string,
+    metric?: string,
+    timeRange?: TimeRange
+  ) => Promise<string[]>;
+  getLabelValues: (
+    dataConnectionId: string,
+    label: string,
+    timeRange?: TimeRange
+  ) => Promise<string[]>;
 }
 
 interface MetricMetadata {
@@ -55,6 +65,7 @@ interface MetricMetadata {
 export class PromQLToolHandlers {
   private prometheusClient: PrometheusResourceClient;
   private dataSourceName: string;
+  private data: DataPublicPluginStart;
 
   constructor(data: DataPublicPluginStart, dataSourceName: string) {
     const client = data.resourceClientFactory.get<PrometheusResourceClient>('prometheus');
@@ -63,6 +74,7 @@ export class PromQLToolHandlers {
     }
     this.prometheusClient = client;
     this.dataSourceName = dataSourceName;
+    this.data = data;
   }
 
   public async executeTool(
@@ -84,8 +96,9 @@ export class PromQLToolHandlers {
       const metricsLimit = args.metricsLimit ?? DEFAULT_METRICS_LIMIT;
       const labelsLimit = args.labelsLimit ?? DEFAULT_LABELS_LIMIT;
       const valuesLimit = args.valuesLimit ?? DEFAULT_VALUES_LIMIT;
+      const timeRange = this.data.query.timefilter.timefilter.getTime();
 
-      let metricNames = await this.prometheusClient.getMetrics(this.dataSourceName);
+      let metricNames = await this.prometheusClient.getMetrics(this.dataSourceName, timeRange);
 
       if (args.query) {
         try {
@@ -106,9 +119,9 @@ export class PromQLToolHandlers {
       }
 
       const [metadata, ...labelsResults] = await Promise.all([
-        this.prometheusClient.getMetricMetadata(this.dataSourceName),
+        this.prometheusClient.getMetricMetadata(this.dataSourceName, undefined, timeRange),
         ...metricNames.map((metric) =>
-          this.prometheusClient.getLabels(this.dataSourceName, metric).catch(() => [])
+          this.prometheusClient.getLabels(this.dataSourceName, metric, timeRange).catch(() => [])
         ),
       ]);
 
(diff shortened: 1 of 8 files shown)


src/plugins/explore/public/application/utils/query_assist/promql_tool_handlers.ts
```diff
@@ -0,0 +1,156 @@
...
+  labelValues: Record<string, string[]>;
+}
+
+interface PrometheusResourceClient {
+  getMetrics: (dataSourceId: string) => Promise<string[]>;
+  getMetricMetadata: (
+    dataSourceId: string
+  ) => Promise<Record<string, Array<{ type: string; help: string }>>>;
+  getLabels: (dataSourceId: string, metric?: string) => Promise<string[]>;
+  getLabelValues: (dataSourceId: string, label: string, metric?: string) => Promise<string[]>;
```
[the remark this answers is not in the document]
joshuali925: updated to be dataConnectionId, which is the dataSourceName for prometheus

src/plugins/explore/public/application/utils/query_assist/agui_agent.ts
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  private executeRequest(requestBody: RunAgentInput, input: AgUiRunInput): Observable<BaseEvent> {
+    this.abort();
+
+    const basePath = '/api/chat/proxy';
+    const url = input.dataSourceId
+      ? `${basePath}?dataSourceId=${encodeURIComponent(input.dataSourceId)}`
+      : basePath;
+
```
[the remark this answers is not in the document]
joshuali925: existing pattern doesn't seem to use basepath prepend https://github.com/opensearch-project/OpenSearch-Dashboards/blob/1d78029963d0aef70c837e7f484a8d56086b390c/src/plugins/chat/public/services/ag_ui_agent.ts#L50-L50

src/plugins/explore/public/application/utils/query_assist/promql_tool_handlers.ts
```diff
@@ -0,0 +1,156 @@
...
+  private dataSourceName: string;
+
+  constructor(data: DataPublicPluginStart, dataSourceName: string) {
+    const client = data.resourceClientFactory.get<PrometheusResourceClient>('prometheus');
+    if (!client) {
+      throw new Error('Prometheus resource client not found');
+    }
+    this.prometheusClient = client;
+    this.dataSourceName = dataSourceName;
+  }
```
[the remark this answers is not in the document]
joshuali925: dataSourceName and dataConnectionId are the same and correct, dataSourceId shouldn't be used to call prometheus

src/plugins/explore/public/application/utils/query_assist/promql_generator.ts
```diff
@@ -0,0 +1,268 @@
...
+  const toolHandlers = new PromQLToolHandlers(data, dataSourceName);
+
+  let streamingText = '';
+  let query: string | undefined;
+  let error: Error | undefined;
+
+  const messages: Message[] = [
+    // 'system' role messages seem to be ignored by ag-ui agent
+    { id: agent.generateMessageId(), role: 'user', content: PROMQL_SYSTEM_PROMPT },
+    { id: agent.generateMessageId(), role: 'user', content: question },
```
ps48, line 87: Passing system prompts as user content could be a prompt injection vector. Isn't ML commons blocking this?

src/plugins/explore/public/application/utils/query_assist/promql_utils.ts
```diff
@@ -0,0 +1,63 @@
...
+): string | undefined {
+  const MAX_STRING_LENGTH = 1000;
+  for (const [key, value] of Object.entries(args)) {
+    if (typeof value === 'string' && value.length > MAX_STRING_LENGTH) {
+      return `Argument '${key}' exceeds maximum length of ${MAX_STRING_LENGTH}`;
+    }
+  }
+
+  switch (toolName) {
+    case 'search_prometheus_metadata':
```
ps48, line 50: Should use PromQLToolName.SEARCH_PROMETHEUS_METADATA ?

src/plugins/explore/public/application/utils/query_assist/promql_generator.ts
```diff
@@ -0,0 +1,268 @@
...
+   */
+  async function processStream(observable: ReturnType<typeof agent.runAgent>): Promise<void> {
+    const pendingOperations: Array<Promise<void>> = [];
+
+    return new Promise((resolve, reject) => {
+      observable.subscribe({
+        next: (event: BaseEvent) => {
+          const operation = handleEvent(event);
+          if (operation) {
+            pendingOperations.push(operation);
```
ps48, line 218: Question: Should we put await on operations. The pendingOperations are collected but only awaited in complete. If an error occurs during tool execution, it may not be caught properly until stream completion.

ps48 (approved):
Added some minor comments

src/plugins/explore/public/application/utils/query_assist/promql_generator.ts
joshuali925: ml-commons accepts user prompts, and prompts are just texts, they won't block it

TackAdam (approved):

src/plugins/explore/public/application/utils/query_assist/promql_generator.ts
joshuali925: this is processing the response stream, I think we should finish processing it regardless of whether previous events succeeded? since events might not be related

src/plugins/explore/public/application/utils/query_assist/promql_utils.ts
joshuali925: yes, will update

PR #11208, opened 2026-01-21
Update the OTel sample data index names

ps48:
### Description

This PR fixes a naming conflict in the OTEL sample data. The sample data currently uses the same index naming convention as OpenSearch Data Prepper, which could conflict with actual customer-ingested data. The updated naming convention ensures OTEL sample data indexes are distinct and won't interfere with production data. 

## Changelog

- fix: Update the OTel sample data index names

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/11244.yml
+++ b/changelogs/fragments/11244.yml
@@ -1,0 +1,2 @@
+fix:
+- Address bugs for echarts migration ([#11244](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11244))
--- a/changelogs/fragments/11250.yml
+++ b/changelogs/fragments/11250.yml
@@ -1,0 +1,2 @@
+fix:
+- Bump less to 4.1.3 to use disablePluginRule when in Markdown panel ([#11250](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11250))
--- a/changelogs/fragments/11251.yml
+++ b/changelogs/fragments/11251.yml
@@ -1,0 +1,2 @@
+fix:
+- Add stricter sanitization on axis label and name in visualizations ([#11251](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11251))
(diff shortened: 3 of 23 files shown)


src/plugins/home/server/services/sample_data/data_sets/otel/index.ts
```diff
@@ -60,40 +60,40 @@ export const otelSpecProvider = function (): SampleDatasetSchema {
...
       {
-        id: 'ss4o_metrics-otel-sample',
+        id: 'ss4o_metrics-otel-opensearch_dashboards-sample',
         dataPath: path.join(__dirname, './sample_metrics.json.gz'),
         fields: metricsFieldMappings,
         timeFields: ['@timestamp', 'exemplar.time', 'startTime', 'time', 'observedTimestamp'],
         currentTimeMarker: '2024-10-16T19:00:01',
         preserveDayOfWeekTimeOfDay: false,
-        indexName: 'ss4o_metrics-otel-sample',
+        indexName: 'ss4o_metrics-otel-opensearch_dashboards-sample',
```
Hailong-am, line 87:
```suggestion
        indexName: 'opensearch_dashboards_sample_data_otel_metrics',
```

Hailong-am: guess we need to make the index name consistent

ps48: @Hailong-am, the dashboards-observability plugin is still hardcoded to use `ss4o_metrics-*` and there is no way for users to configure it. For all the others users change the index using settings. Hence, not modifying this one.

Hailong-am: Got it, that make sense. we might support user change the index name/pattern for metrics too.

PR #11219, opened 2026-01-22
Apply data table columns filter to download csv

Maosaic:
### Issues Resolved

Previously, download csv on Explore page does not filter `columns` state like data table, which can casue the down loaded csv table miss match with the data table render columns. 

This CR extract columns filter from data table to helper function and use it in both data table and down loaded csv workflow.  

## Changelog

- fix: Apply data table columns filter to download csv

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/11219.yml
+++ b/changelogs/fragments/11219.yml
@@ -1,2 +1,2 @@
 fix:
-- Apply data atble columns filter to download csv ([#11219](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11219))
+- Apply data table columns filter to download csv ([#11219](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11219))


src/plugins/explore/public/helpers/use_displayed_columns.ts
```diff
@@ -0,0 +1,127 @@
...
+  includeFieldCounts?: boolean;
+}
+
+/**
+ * Core function for processing displayed columns with proper filtering.
+ * This is the single source of truth for column processing logic.
+ */
+export const processDisplayedColumns = (
+  rawColumns: string[],
+  dataset: any,
```
TackAdam, line 37: NIT:
  dataset: any,  //  Should use IndexPattern type
  uiSettings: any,  //  Should use IUiSettingsClient type
  processedResults?: any  //  Could use more specific type

PR #11230, opened 2026-01-26
[Explore vis] explore visualization bug fixes

ruanyl:
### Description
This PR contains bunch of bug fixes:

1. Responsive grid size based on legend and visualMap to avoid legend overlap with grid
2. Fix panel styles to align with the new page density
3. Fix gauge title switch not working
4. Fix vis crash when switching to table
5. Fix reuse axes current mapping when switching chart type not work as expected
6. Area opacity changed to 0.5
7. Removed legend for single series charts
8. Show line dot only when hover
9. Fix unit tests

## Changelog

- fix: discover visualization fix and ux improvements

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/echarts_render.tsx
+++ b/src/plugins/explore/public/components/visualizations/echarts_render.tsx
@@ -5,9 +5,10 @@
 
 import * as echarts from 'echarts';
 import React, { useMemo, useRef, useEffect, useState } from 'react';
+import { isEqual } from 'lodash';
+
 import { TimeRange } from '../../../../data/public';
 import { DEFAULT_THEME } from './theme/default';
-import { isEqual } from 'lodash';
 
 interface Props {
   spec: echarts.EChartsOption;


src/plugins/explore/public/embeddable/explore_embeddable.tsx
```diff
@@ -4,6 +4,7 @@
  */
 
 import { isEqual } from 'lodash';
+import moment from 'moment';
 import { merge, Subscription } from 'rxjs';
```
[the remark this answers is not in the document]
ruanyl: It uses `*` by intention, `gte` and `lte` can be moment object

src/plugins/explore/public/embeddable/explore_embeddable_component.tsx
```diff
@@ -119,7 +119,9 @@ export const ExploreEmbeddableComponent = ({ searchProps }: ExploreEmbeddablePro
     }
 
     if (searchProps.spec && !searchProps.spec.$schema) {
-      return <EchartsRender spec={searchProps.spec} />;
+      return (
+        <EchartsRender spec={searchProps.spec} onSelectTimeRange={searchProps.onSelectTimeRange} />
```
Qxisylolo, line 123: Should we update this here, since the updated EchartsRender now only accepts onSelectTimeRange, data, config, and timeRange?

ruanyl: Good catch, I forgot this, will take a look

src/plugins/explore/public/components/visualizations/echarts_render.tsx
```diff
@@ -46,15 +58,136 @@ export const EchartsRender = ({ spec }: Props) => {
...
+            } else if (v.bottom) {
+              grid.bottom = 80;
+            }
+          }
+        }
+        instance.setOption({ grid, legend: legendConfig });
+        gridConfigRef.current = grid;
+      }
+    }, 300);
+
```
Hailong-am, line 139: can we use echart public apis like `nstance.on('rendered', handleRendered);` to avoid arbitrary timeout

src/plugins/explore/public/components/visualizations/pie/pie_chart_utils.ts
```diff
@@ -18,14 +18,14 @@ export const createPieSeries = <T extends BaseChartStyle>({
...
-      ? (params: any) => params.value[1]
-      : '{b}';
+  let formatter = '';
+  if (styles?.exclusive?.showValues && styles?.exclusive?.showLabels) {
+    formatter = `{b}: {@${valueField}}`;
+  } else if (styles?.exclusive?.showLabels) {
+    formatter = '{b}';
+  } else if (styles?.exclusive?.showValues) {
+    formatter = `{@${valueField}}`;
+  }
```
Hailong-am: would empty formatter breaks rendering? The original logic have `{b}` as default value

src/plugins/explore/public/components/visualizations/metric/metric_utils.ts
```diff
@@ -200,19 +200,19 @@ export const createMetricChartSeries = ({
...
   const xAxis = Array.isArray(state.spec?.xAxis)
-    ? state.spec.xAxis.map((a) => ({ ...a, show: false }))
-    : { ...state.spec?.xAxis, show: false };
+    ? state.spec.xAxis.map((a) => ({ ...a, show: false, silent: true }))
+    : { ...state.spec?.xAxis, show: false, silent: true };
   const yAxis = Array.isArray(state.spec?.yAxis)
-    ? state.spec.yAxis.map((a) => ({ ...a, show: false }))
-    : { ...state.spec?.yAxis, show: false };
+    ? state.spec.yAxis.map((a) => ({ ...a, show: false, silent: true }))
+    : { ...state.spec?.yAxis, show: false, silent: true };
```
Hailong-am, line 209: Setting silent: true disables all mouse interactions on axes, which might break tooltip or zoom functionality.

src/plugins/explore/public/components/visualizations/pie/pie_chart_utils.ts
ruanyl: Good catch, I will give it default `{b}`

src/plugins/explore/public/components/visualizations/echarts_render.tsx
ruanyl: Sounds a good idea, let me have a try

src/plugins/explore/public/components/visualizations/metric/metric_utils.ts
ruanyl: Metric sparkline should not have tooltip or zoom functionality, so this should be good.

src/plugins/explore/public/components/visualizations/echarts_render.tsx
ruanyl: Fixed, now listening `finished` event

src/plugins/explore/public/embeddable/explore_embeddable_component.tsx
ruanyl: I reverted the changes and implemented it in another way so the interface won't change

src/plugins/explore/public/components/visualizations/pie/pie_chart_utils.ts
```diff
@@ -18,14 +18,14 @@ export const createPieSeries = <T extends BaseChartStyle>({
...
 
-  // In dataset + pie mode, params.value is an array like [name, value],
-  // so we use params.value[1] to extract only the numeric value for the label.
-  const formatter =
-    styles?.exclusive?.showValues && styles?.exclusive?.showLabels
-      ? (params: any) => params.value
-      : styles?.exclusive?.showValues
-      ? (params: any) => params.value[1]
-      : '{b}';
+  let formatter = '{b}';
```
wanglam, line 21: Nit: Shall we use empty string as default? Seems if `showLabels` is false, it's still use `{b}` as formatter. It's the same as showLabels equal true.

ruanyl: If `label.show` is false, the formatter won't take effect, so it doesn't matter what's the default value

SuZhou-Joe (approved):

PR #10688, opened 2025-10-08
[CVE] update axios to ^1.12 for CVE-2025-58754

angle943:
### Description

- updates axios to v^1.12.0, closes #10523 and closes #10685
- axios v1.12.0 introduced a breaking typescript change: https://github.com/axios/axios/pull/6982/files#r2385695994
- There is an open issue regarding this: https://github.com/axios/axios/issues/7059
- This was causing `yarn osd bootstrap` to fail:

```
ERROR Error: Command failed with exit code 1: /Users/jungkm/.cache/node/corepack/v1/yarn/1.22.19/bin/yarn.js run osd:bootstrap
      error Command failed with exit code 1.
      $ scripts/use_node scripts/build_ts_refs && scripts/use_node scripts/register_git_hook
      ERROR Error: Command failed with exit code 1: /Users/jungkm/workplace/explore/OpenSearch-Dashboards/node_modules/typescript/bin/tsc -b tsconfig.refs.json --pretty
            node_modules/axios/index.d.ts:428:3 - error TS2416: Property 'cause' in type 'AxiosError<T, D>' is not assignable to the same property in base type 'Error'.
              Type 'unknown' is not assignable to type 'Error | undefined'.

            428   cause?: unknown;
                  ~~~~~


            Found 1 error.
```

- to remedy this, i am using `patch-package` to patch the type file

## Changelog
- fix: Upgrade Axios CVE

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [x] Commits are signed per the DCO using --signoff

PR #11233, opened 2026-01-26
[chore] bump axios

angle943:
### Description

- bump axios to 1.13.3, which was released [a day ago](https://www.npmjs.com/package/axios) and allows us to get rid of the monkey patch that was required previously with 1.13.2.
- See axios commit that is included in 1.13.3: https://github.com/axios/axios/pull/7327
- See our monkey-patch solution: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/10688

## Changelog
- chore: bump axios

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -10937,7 +10937,7 @@ axe-core@^4.0.2, axe-core@^4.10.3, axe-core@^4.3.5:
   resolved "https://registry.yarnpkg.com/axe-core/-/axe-core-4.10.3.tgz#04145965ac7894faddbac30861e5d8f11bfd14fc"
   integrity sha512-Xm7bpRXnDSX2YE2YFfBk2FnF0ep6tmG7xPh8iHee8MIcrgq762Nkce856dYtJYLkuIoYZvGfTs/PbZhideTcEg==
 
-axios@^1.6.5, axios@^1.8.2:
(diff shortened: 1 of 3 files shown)


LDrago27:
@angle943 we should bump axios within the osd-dev utils, osd-ui-shared-deps and chromedriver if possible.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/packages/osd-dev-utils/package.json
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/packages/osd-ui-shared-deps/package.json

It seems from the yarn lock as if we are forcing all internal version to use 1.13.3

yarn.lock
```diff
@@ -10938,9 +10938,9 @@ axe-core@^4.0.2, axe-core@^4.10.3, axe-core@^4.3.5:
   integrity sha512-Xm7bpRXnDSX2YE2YFfBk2FnF0ep6tmG7xPh8iHee8MIcrgq762Nkce856dYtJYLkuIoYZvGfTs/PbZhideTcEg==
 
 axios@^1.6.5, axios@^1.8.2:
```
LDrago27: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11233#issuecomment-3801624292

angle943: good call, let me update

Issue #11234, opened 2026-01-26
[BUG] Discover saved search now takes precedence over URL params


joshuali925:
**Describe the bug**

If user opens a saved search and modified the filter, their shared URL will not contain the modification they made.

**To Reproduce**
1. in discover, search with filter `clientip: 246.8.245.40` using sample logs data
2. save the saved search
3. modify the filter to be `NOT clientip: 246.8.245.40`, run query. Notice URL contains `negate:!t`
4. click share > snapshot > copy link
5. See clipboard, link contains `negate:!t`. But when opening the link, filter shows previous value `clientip: 246.8.245.40`, and URL becomes `negate:!f`

**Expected behavior**
Discover state should match filters present at the time it got shared, instead of matching the original saved search.

**OpenSearch Version**
main

**Dashboards Version**
main

**Plugins**

Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

This is introduced by https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9529 as a feature, but I'm not sure if 
> When loading saved search, saved filters will apply on top of existing filters

should be the behavior, because if user should be able to share modifications on top of a saved search. @Maosaic do you have more context on why #9529 was done?

PR #11239, opened 2026-01-26
[Discover] use URL param filters with saved search as fallback

joshuali925:
### Description

See https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11234. This PR partially modifies #9529 and instead of 
> When loading saved search, saved filters will apply on top of existing filters

we'll preserve the filters defined in URL params, since user may have modified them on top of saved search content. @Maosaic please take a look and call out any concerns, thanks

With this PR, this will be the behavior
1. saved search with URL params -> load filters **only** from URL params. It still opens saved search but does not use filters in it
1. saved search with no URL params -> load filters from saved search

### Issues Resolved

closes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11234

## Testing the changes

UT

## Changelog

- fix: use URL param filters with saved search as fallback

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/11239.yml
+++ b/changelogs/fragments/11239.yml
@@ -1,2 +1,2 @@
 fix:
-- Prefer URL param filters over saved search content in Discover ([#11239](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11239))
+- Use URL param filters with saved search as fallback ([#11239](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11239))


src/plugins/discover/public/application/view_components/utils/use_search.ts
```diff
@@ -580,18 +580,23 @@ export const useSearch = (services: DiscoverViewServices) => {
...
       } else if (filters !== undefined) {
         const result = typeof filters === 'function' ? filters() : filters;
         if (result !== undefined) {
-          actualFilters.push(...(Array.isArray(result) ? result : [result]));
+          savedSearchFilters.push(...(Array.isArray(result) ? result : [result]));
         }
       }
 
+      const urlFilters = cloneDeep(filterManager.getAppFilters());
+      const actualFilters = [...savedSearchFilters, ...urlFilters];
```
joshuali925: discussed with Joey, this won't work well if user removed filter from saved search, because the removed filter will still show up from `savedSearchFilters`

needs a better implementation to use URL filters if they exist, otherwise load from saved search

Issue #11130, opened 2026-01-06
[RFC] Enable HTTP Response Compression for OpenSearch Clients in Dashboards


ahkcs:
## Summary
Enable gzip/deflate HTTP response compression for all OpenSearch clients used by OpenSearch Dashboards to reduce payload size and significantly improve query latency, especially for Discover / Explore workloads.

---

## Problem Statement
OpenSearch Dashboards does not currently request compressed HTTP responses from OpenSearch clusters. This results in:

- Large uncompressed response payloads  
- Increased end-to-end query latency  
- Poor user experience for large result sets (e.g., Discover)

---

## Proposed Solution
Add the `Accept-Encoding: gzip, deflate` HTTP header to all OpenSearch client configurations in OpenSearch Dashboards.

### Affected Locations
1. `src/core/server/opensearch/default_headers.ts` (default client)  
2. `src/plugins/data_source/server/legacy/configure_legacy_client.ts` (MDS legacy client) (PPL usage)  
3. `src/plugins/data_source/server/client/client_config.ts` (MDS modern client)

This ensures consistent compression behavior across all client paths.

---

## Test Environment

| Component           | Value |
|---------------------|-------|
| OpenSearch Version  | AWS OpenSearch Service 2.19 |
| Query               | `source = osi-otel-logs` |
| Dataset             | 40,000 documents, ~54 MB total |
| Schema              | 100+ fields per document |
| Result Set          | 10,000 rows |

---

## Performance Impact

| Metric                | Before   | After    | Improvement |
|-----------------------|----------|----------|-------------|
| End-to-End Query Time | 9,625 ms | 1,835 ms | -81%        |
| Response Payload Size | 15.8 MB  | 355 KB   | -98%        |

---

## Compatibility & Risk
- OpenSearch supports gzip compression natively  
- Node.js HTTP client automatically handles decompression  
- No breaking changes  
- Header is ignored if compression is not supported  
- No user configuration required  

---

## Testing
- Validated in Discover with large result sets  
- Tested against AWS OpenSearch Service 2.19  
- No functional or behavioral changes observed  

---

## Conclusion
Enabling response compression by default is a low-risk, high-impact improvement that dramatically reduces payload size and improves query performance across OpenSearch Dashboards.

joshuali925:
[...]

PR #11240, opened 2026-01-26
feat: Add gzip compression support for PPL queries

ahkcs:
## Description

This PR adds **gzip compression support for PPL queries only**, reducing payload size and improving query latency when fetching large result sets from OpenSearch.

Compression is explicitly scoped to PPL search paths and does **not** affect SQL queries, other plugins, or the global OpenSearch client configuration.

---

## Key Changes

* Added `requestCompression` option to the `Facet` class in the `query_enhancements` plugin
* Enabled request compression (`accept-encoding: gzip, deflate`) for **PPL search strategies only**:

  * `ppl_search_strategy.ts` 
  * `ppl_async_search_strategy.ts` 
  * `ppl_raw_search_strategy.ts` 

---

## What Is **NOT** Affected

* SQL queries (no compression enabled)
* Other plugins (chat, ML, data_source, etc.) — completely untouched
* Global OpenSearch client configuration — unchanged

The `Facet` class is only used within the `query_enhancements` plugin, ensuring the change is fully isolated to PPL queries.

---

## Issues Resolved

* [https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11130](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11130)

---

### Manual Testing

1. Start OpenSearch Dashboards:

   ```bash
   yarn start --no-base-path
   ```
2. Open Explore (Discover) and select a PPL-compatible data source
3. Run a PPL query (e.g., `source=<index>`)
4. Open browser DevTools → Network tab
5. Inspect the request headers:

   * Request contains `accept-encoding: gzip, deflate`
   * Response contains `content-encoding: gzip` (when OpenSearch applies compression)

---

## Unit Tests

```bash
node scripts/jest.js src/plugins/query_enhancements/server/utils/facet.test.ts
```

---

## Changelog

 - feat: Add gzip compression support for PPL queries


src/plugins/query_enhancements/server/search/ppl_search_strategy.ts
```diff
@@ -30,6 +30,7 @@ export const pplSearchStrategyProvider = (
     endpoint: 'enhancements.pplQuery',
     useJobs: false,
     shimResponse: true,
+    requestCompression: true,
```
joshuali925, line 33: there are a few other SQL/PPL search strategies, do they need this?

src/plugins/query_enhancements/server/utils/facet.ts
```diff
@@ -51,6 +66,7 @@ export class Facet {
           ...(lang && { lang }),
         },
         ...(format !== 'jdbc' && { format }),
+        ...(Object.keys(compressionHeaders).length > 0 && { headers: compressionHeaders }),
```
joshuali925, line 69: this should work, but could you test with https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11205 `opensearch.requestCompression` being false or true just to make sure behavior is expected

src/plugins/query_enhancements/server/search/ppl_search_strategy.ts
ahkcs: We intentionally only enabled compression for the main PPL search strategy (ppl_search_strategy.ts)
The scope of this PR was specifically for PPL queries. We can extend compression to other strategies in follow-up PRs if needed.

src/plugins/query_enhancements/server/utils/facet.ts
ahkcs: Added temporary server-side logging to verify headers are sent between OSD ↔ OpenSearch:                                                 
[OSD->OpenSearch] endpoint=enhancements.pplQuery, headers={"accept-encoding":"gzip, deflate"}                                            
                                                                                                                                           
The PPL-specific compression via Facet works correctly in both cases.

PR #11231, opened 2026-01-26
enhance chatbot ui

yubonluo:
### Description

- __Input box upgrade__: EuiFieldText → EuiTextArea, supports multi line input and automatic height adjustment 
- __Welcome interface enhancement__: Added AI assistant welcome text, gradient halo effect, and 3 quick suggestion cards 
- __Interaction optimization__: Click on the suggestion card to automatically fill in the input box

https://github.com/user-attachments/assets/7088a49e-da3d-49f6-a1d9-070d93a37b40

## Changelog

- feat: enhance chatbot ui
### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #11266, opened 2026-01-29
UX improvement for chat window: Stop button, scroll bar fix and text box height increase

arjunkumargiri:
### Description

Chat box UX improvements:
* Added stop button to stop response streaming
* Scroll will not scroll down to bottom by default
* Modified text box to text area with max height of 4 lines allowing users to view more text during input.

## Screenshot
Screenshot 2026-01-28 at 5 31 49 PM
Screenshot 2026-01-28 at 5 32 13 PM

## Testing the changes

Local testing

## Changelog

- skip: UX improvement for chat window: Stop button, scroll bar fix and text box height increase

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/chat/public/components/chat_window.tsx
+++ b/src/plugins/chat/public/components/chat_window.tsx
@@ -353,20 +353,21 @@ const ChatWindowContent = React.forwardRef<ChatWindowInstance, ChatWindowProps>(
   const handleStop = useCallback(() => {
     // Abort the current streaming request
     chatService.abort();
-    
+
     // Remove loading message if it exists
     if (loadingMessageIdRef.current) {
       setTimeline((prev) => prev.filter((msg) => msg.id !== loadingMessageIdRef.current));
       loadingMessageIdRef.current = null;
     }
-    
+
     // Unsubscribe from current observable if exists
     if (currentSubscriptionRef.current) {
       currentSubscriptionRef.current.unsubscribe();
       currentSubscriptionRef.current = null;
     }
-    
-    // Update streaming state
+
+    // Update streaming state (both ref and state for React 18 compatibility)
+    isStreamingRef.current = false;
     setIsStreaming(false);
   }, [chatService]);
 
--- a/src/plugins/chat/public/components/chat_messages.test.tsx
+++ b/src/plugins/chat/public/components/chat_messages.test.tsx
@@ -47,7 +47,7 @@ describe('ChatMessages', () => {
     it('should render empty state when no messages', () => {
       const { getByText } = render(<ChatMessages {...defaultProps} />);
 
-      expect(getByText('Start a conversation')).toBeTruthy();
+      expect(getByText("Hi, I'm your AI Assistant")).toBeTruthy();
     });
 
     it('should render messages from timeline', () => {


src/plugins/chat/public/components/chat_input.scss
```diff
@@ -1,3 +1,5 @@
+@import "@elastic/eui/src/global_styling/variables";
+
```
Hailong-am: we don't need to import this

src/plugins/chat/public/components/chat_input.scss
```diff
@@ -7,17 +9,42 @@
...
-    align-items: center;
+    align-items: flex-end;
   }
 
   &__fieldWrapper {
     position: relative;
     width: 100%;
 
-    input {
+    textarea {
```
Hailong-am, line 17: @yubonluo have same change for this https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11231

src/plugins/chat/public/components/chat_input.tsx
```diff
@@ -57,15 +71,17 @@ export const ChatInput: React.FC<ChatInputProps> = ({
           />
         )}
         <div className="chatInput__fieldWrapper">
-          <EuiFieldText
+          <EuiTextArea
             inputRef={inputRef}
-            placeholder="Ask anything. Type / for actions"
+            placeholder="How can I help you today?"
```
Hailong-am, line 64: why we changed this placeholder?

src/plugins/chat/public/components/chat_window.tsx
```diff
@@ -328,6 +340,26 @@ const ChatWindowContent = React.forwardRef<ChatWindowInstance, ChatWindowProps>(
...
+    
+    // Unsubscribe from current observable if exists
+    if (currentSubscriptionRef.current) {
+      currentSubscriptionRef.current.unsubscribe();
+      currentSubscriptionRef.current = null;
+    }
+    
+    // Update streaming state
+    setIsStreaming(false);
+  }, [chatService]);
```
Hailong-am, line 372: what's the behavior stop the response in middle? should we clean the incomplete message for next round of llm call or keep it whatever it is? can we have a test about this?

src/plugins/chat/public/components/chat_input.test.tsx
```diff
@@ -0,0 +1,326 @@
...
+    it('should handle empty onSend callback', () => {
+      const { getByLabelText } = render(
+        <ChatInput {...defaultProps} input="test" onSend={undefined as any} />
+      );
+
+      const button = getByLabelText('Send message');
+
+      // Should not throw when clicking with undefined callback
+      expect(() => fireEvent.click(button)).not.toThrow();
+    });
```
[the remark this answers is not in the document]
Hailong-am: can you fix this

src/plugins/chat/public/components/chat_input.scss
arjunkumargiri: Nice! will remove it from my PR

src/plugins/chat/public/components/chat_input.tsx
arjunkumargiri: Existing placeholder seems incorrect.  I don't think '/' for actions is suggested yet.

src/plugins/chat/public/components/chat_window.tsx
arjunkumargiri: If stop is pressed in the middle of streaming, then the content is retained as is with incomplete message. This is the existing behavior in other LLM chat UI such as claude. There is existing test case to cover this scenario.

src/plugins/chat/public/components/chat_input.tsx
Hailong-am: we have added a `/` command system for investigation  with `/investigation` and there will more command next, i think we need to keep it

src/plugins/chat/public/components/chat_window.tsx
Hailong-am: so for next round LLM call, we will send the incomplete message to LLM? For now, it works. When we support store the conversation history and persist the conversation data, there will have a inconsistency between UI and the storage side. And agent side need to support stop as well

kamingleung:
@sicheng-lu, can you work with @arjunkumargiri to review the UX improvements before merging. Thanks.

sicheng-lu:
Stop button looks good. Is the textfield still accessible in the stop button state?

arjunkumargiri:
> Stop button looks good. Is the textfield still accessible in the stop button state?

No, it could be accessed only after streaming is stopped.

wanglam (approved):

PR #11285, opened 2026-02-02
Remove no need global variables ch

wanglam:
### Description

Change this file the same as old version:
https://github.com/wanglam/OpenSearch-Dashboards/blob/d7004dc5b0392477fdd54ac66b29d231975a173b/src/dev/mocha/run_mocha_cli.js

## Changelog

- fix: Remove no need global variables ch in mocha

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/11285.yml
+++ b/changelogs/fragments/11285.yml
@@ -1,0 +1,2 @@
+fix:
+- Remove no need global variables ch in mocha ([#11285](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11285))


src/dev/mocha/run_mocha_cli.js
```diff
@@ -49,7 +49,7 @@ export function runMochaCli() {
   // check that we aren't leaking any globals
   process.argv.push('--check-leaks');
   // prevent globals injected from canvas plugins from triggering leak check
-  process.argv.push('--globals', '__core-js_shared__,core,_,ch');
+  process.argv.push('--globals', '__core-js_shared__,core,_, ');
```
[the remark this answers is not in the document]
wanglam: Just want to keep the same code with the version before node.js 22 upgrade here. It should be the same with https://github.com/wanglam/OpenSearch-Dashboards/blob/d7004dc5b0392477fdd54ac66b29d231975a173b/src/dev/mocha/run_mocha_cli.js

Hailong-am, line 52: is it ok to keep last `,`

wanglam: Just want to keep the same implementation with before. I'm open to change to `process.argv.push('--globals', '__core-js_shared__,core,_');` here, the last `,` can be removed as well.

Hailong-am: as long as it won't throw any error, it should be good.

PR #11291, opened 2026-02-02
Add starter CLAUDE.md

Maosaic:
### Description

Add starter a `CLAUDE.md` which covers: 

 - Common commands — bootstrap, dev server, running single/all tests, linting, type checking, Cypress tests, and API review                                                        
  - Architecture — monorepo layout (src/core, src/plugins, packages, examples, test, cypress), the plugin system lifecycle (setup/start/stop), manifest format, and standard plugin 
  file structure                                                                                                                                                                    
  - Code conventions — snake_case filenames, TypeScript-first, named exports, ES module syntax, Prettier settings, API route conventions, SASS/React patterns, and saved object type
   definitions                                                                                                                                                                      
  - Testing conventions — Jest unit tests with 80% coverage requirement, react-testing-library over snapshots, Cypress as the preferred functional test framework (not Selenium),   
  and Cypress best practices                                                                                                                                                        
  - Core API changes — the yarn docs:acceptApiChanges workflow                                                                                                                      
  - PR requirements — DCO sign-off and test coverage expectations    

## Changelog

- chore: Add starter CLAUDE.md

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/11291.yml
+++ b/changelogs/fragments/11291.yml
@@ -1,0 +1,2 @@
+chore:
+- Add starter CLAUDE.md ([#11291](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11291))


CLAUDE.md
```diff
@@ -0,0 +1,132 @@
...
+
+# Pre-commit checks (lint + type check)
+node scripts/precommit_hook.js --fix
+```
+
+## Architecture
+
+### Monorepo Structure
+
+- **`src/core/`** — Core platform providing HTTP, plugin system, saved objects, logging, config, and both public (browser) and server APIs. Plugins depend on Core's `setup` and `start` contracts.
```
seraphjiang, line 62: could we describe the osd code  structure like we did for plugin?

Standard plugin file structure:

CLAUDE.md
```diff
@@ -0,0 +1,132 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+OpenSearch Dashboards is a browser-based data visualization and exploration tool for OpenSearch. It is a large-scale Node.js + React monorepo with a plugin-based architecture. The project was forked from Kibana and adapted for the OpenSearch ecosystem.
+
+## Common Commands
```
seraphjiang, line 9: could we be more specific about the commond?

1 build & release
2 local development
3 test
3.1 unit test
3.2 cypress test
3.3 selenium test

CLAUDE.md
```diff
@@ -0,0 +1,132 @@
...
+└── server/
+    ├── index.ts                  # Exports plugin + config
+    ├── plugin.ts                 # Server plugin class
+    ├── routes/                   # HTTP API routes (must start with /api/)
+    ├── saved_objects/            # Saved object type definitions + migrations
+    └── services/                 # Server-side services
+```
+
+Use `core.getStartServices()` in application mount handlers rather than storing `start` references as class fields.
+
```
seraphjiang, line 96: should we add a section for api?

https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/docs/openapi

Maosaic: It's useful for someone manually exploring the API in a browser, but Claude     
  doesn't use Swagger UI. When Claude needs to understand an API endpoint, it reads the route handler code directly in server/routes/. Adding this to CLAUDE.md would point Claude  
  toward external hosted docs when the source of truth (the actual route definitions) is already in the repo. If anything, a more useful addition would be noting that API routes   
  live in server/routes/ within each plugin and must follow the `/api/<plugin>/ prefix` — which the current file already covers.

CLAUDE.md
Maosaic: The plugin structure is in CLAUDE.md because it's a convention Claude needs to follow when writing code — it dictates where to put files, what index.ts should export, etc. The   
  top-level OSD directory structure (`src/`, `packages/`, `test/`, `cypress/`) is different: it's discoverable. Claude can run a quick ls or glob to see what's there. The current          
  Architecture section already describes what each top-level directory contains at a high enough level to orient Claude. Adding a full tree diagram would bulk up the file with     
  information that's both easy to discover and rarely actionable — Claude doesn't need to create new top-level directories, it needs to navigate existing ones.                     
                                                                                                                                                                                    
  The exception would be if there's a non-obvious structural pattern (like "functional tests for plugin X live in `test/plugin_functional/` not next to the plugin"). That kind of    
  surprising convention is worth calling out, and the current file already does.

PR #11314, opened 2026-02-06
Ruchi/discover explore

ruchidh:
### Description
- Showing only new discover in observability workspace.
- Enable both new and old discover in analytics workspace in case user want to use old discover.
- Remove all banners from old and new discover.
- When an old saved object is loaded from new discover, have it navigate to the classic discover page

Screenshot 2026-02-24 at 12 04 33 PM
Screenshot 2026-02-24 at 12 04 13 PM

## Changelog
- feat: Enable both old discover and new discover in analytics workspace and remove banners

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/11418.yml
+++ b/changelogs/fragments/11418.yml
@@ -1,0 +1,2 @@
+feat:
+- [Chat] Hide conversation delete button for agentic memory ([#11418](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11418))
--- a/src/plugins/chat/public/components/conversation_history_panel.test.tsx
+++ b/src/plugins/chat/public/components/conversation_history_panel.test.tsx
@@ -10,6 +10,7 @@ import {
   ConversationHistoryService,
   SavedConversation,
 } from '../services/conversation_history_service';
+import { AgenticMemoryProvider } from '../services/agentic_memory_provider';
 
 describe('ConversationHistoryPanel', () => {
   let mockService: jest.Mocked<ConversationHistoryService>;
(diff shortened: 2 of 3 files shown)


kamingleung:
@anirudha @brijos @ruchidh What is the migration plan from old to new Discover? It's not clear what we want accomplish here.

Why are we showing both version at the same time? If users opt to use the new Discover, we should replace the old experience, why are we creating another page and name it Explore?

ruchidh:
> **[kamingleung](/kamingleung) **

We have made new changes. observability will have only new discover, while if the user want to use old, it can switch to analytics workspace, where it will see both new and old. Updated pr description too.

kamingleung:
> We have made new changes. observability will have only new discover, while if the user want to use old, it can switch to analytics workspace, where it will see both new and old. Updated pr description too.

@anirudha @brijos Why do we need to show both new and old discover in analytics workspace? Can we default to the new experience and let users switch to the old one? Showing both experiences is rather confusing to users. 

I don't think users can tell what is Discover and Explore, the naming feels like an implementation detail. If we must show both experiences, I would suggest naming the old experience to like "Discover (classic)" to reflect its intention.

seraphjiang:
> @anirudha @brijos Why do we need to show both new and old discover in analytics workspace? Can we default to the new experience and let users switch to the old one? Showing both experiences is rather confusing to users.

we want to have both Discover (for general analytic use case) and Explorer (for log analytic and time series data)
we are not going to deprecate Discover with Explorer , hope that help calrify, let me know if any question

cc: @ruchidh

anirudha:
This PR should not modify the behavior of the Observability workspace. The Observability workspace already has its own Discover experience through the Explore plugin, and that routing/navigation logic should remain untouched.

please review 2 issues/
Specifically, the changes in 
1. >plugin.ts
 that check for isObservability and redirect to explore/logs introduce coupling between the Discover plugin and workspace-specific routing that doesn't belong there. The Discover plugin shouldn't need to be aware of which workspace it's in or make navigation decisions based on that. That responsibility belongs to the Explore plugin and the workspace framework itself.

2. >Similarly, registering Explore nav links into DEFAULT_NAV_GROUPS.all broadens the scope beyond what this PR describes. If the goal is to enable both old and new Discover in the Analytics workspace, that should be done without side effects on Observability.

Suggested approach:
Keep the Observability workspace routing as-is (Explore plugin already handles it)
Scope the Analytics workspace changes to only add the classic Discover nav link there, without altering Observability behavior
Banner removal in the new Discover pages is fine, but the redirect logic in the classic Discover plugin should not be workspace-aware

Is there an issue to understand the scope of this change ?

ruchidh:
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11444 - This is done in another pr . Closing this

PR #11331, opened 2026-02-12
feat: add time range tool to Chatbot

SuZhou-Joe:
### Description

This PR adds a new frontend tool for Chatbot to enable users to update time range within chatbot by using natural language.

## Screenshot

https://github.com/user-attachments/assets/64b9a1d5-8e3e-40e5-bf44-8bdc3a080c61

### Enhance ppl tool with time range

https://github.com/user-attachments/assets/95107bcf-291f-4561-9a8e-390d0fad06d4

## Changelog

- feat: add time range tool to Chatbot

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/query_panel/actions/ppl_execute_query_action.test.tsx
+++ b/src/plugins/explore/public/components/query_panel/actions/ppl_execute_query_action.test.tsx
@@ -121,13 +121,14 @@ describe('usePPLExecuteQueryAction', () => {
       mockUseAssistantAction.mock.calls[mockUseAssistantAction.mock.calls.length - 1][0];
     expect(latestCall).toMatchObject({
       name: 'execute_ppl_query',
-      description: 'Update the query bar with a PPL query and optionally execute it',
+      description:
+        'Update the query bar with a PPL query, optionally set the time range, and execute it. The query should NOT contain time filters - use the from/to parameters to specify the time range instead.',
       parameters: {
         type: 'object',
         properties: {
           query: {
             type: 'string',
-            description: 'The PPL query to set in the query bar',
+            description: 'The PPL query to set in the query bar (without time filters)',
           },
           autoExecute: {
             type: 'boolean',
@@ -174,7 +175,8 @@ describe('usePPLExecuteQueryAction', () => {
       success: true,
       executed: true,
       query: 'source=logs | head 10',
-      message: 'Query updated and executed successfully',
+      message: 'Query updated and executed successfully.',
+      timeRange: undefined,
     });
   });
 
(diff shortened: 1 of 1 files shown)


src/plugins/explore/public/application/utils/hooks/use_timefilter_subscription.ts
```diff
@@ -20,15 +22,17 @@ export const useTimefilterSubscription = (services: ExploreServices) => {
...
-      .subscribe(() => {
-        if (query.dataset) {
-          dispatch(clearResults());
-          dispatch(clearQueryStatusMap());
-          dispatch(executeQueries({ services }) as unknown);
-        }
-      });
+    const timefilter = services.data.query.timefilter.timefilter;
+    const subscription = merge(
+      timefilter.getTimeUpdate$(),
```
SuZhou-Joe: @abbyhu2000 Seems legacy explore will subscribe to time filter update, do you know why we remove the subscription in discover?

ananzh: I understand this adds a trigger mechanism and the data flow remains the same. However, I have a concern about the behavioral change this introduces:

In current explore, time picker changes update Redux state via setDateRange(), but queries only execute when user explicitly clicks "Run". This is intentional - Explore was designed to give users control over when expensive queries execute.

After this change, any timefilter.setTime() call will auto-execute queries, which changes the UX for all time picker interactions, not just the chatbot tool.

So for the chatbot time range tool, should we consider following the same pattern as ppl_execute_query_action.tsx:
* Dispatch setDateRange() to update Redux state
* Explicitly dispatch runQueryActionCreator() or onEditorRunActionCreator() to execute the query

src/plugins/explore/public/components/query_panel/actions/ppl_execute_query_action.tsx
```diff
@@ -135,6 +149,18 @@ export function usePPLExecuteQueryAction(
         // Check if we should auto-execute
         const shouldExecute = args.autoExecute !== false; // Default to true
 
+        // Update time range if both from and to are provided
```
wanglam, line 152: Q: Do we still need the update time range tool if the execute_ppl_query tool can update the time range by self?

wanglam (approved):

src/plugins/explore/public/components/query_panel/actions/ppl_execute_query_action.tsx
SuZhou-Joe: No, the update time range tool will be disabled in new discover page.

Issue #10883, opened 2025-11-07
[BUG] Blank Screen in OpenSearch Discover Due to Malformed Saved Query


ruchidh:
**Describe the bug**

OSD 3.1.0, 2.19  environment  is experiencing a critical UI rendering issue in the Discover page.

Current Behavior
Discover page displays a blank screen when certain saved queries are present
Issue persists across all production indices
Problem occurs regardless of SAML authentication status

**To Reproduce**
1. Add saved query with below format .
{
  "type": "query",
  "id": "80278d12-79df-4045-a3a5-12212324343434",
  "attributes": {
    "title": "DB: Unique Constraint Errors",
    "query": {
      "query": "{\"size\":0,\"query\":{\"bool\":{\"must\":[{\"match_phrase\":{\"message\":\"violates unique constraint\"}}]}}}",
      "language": "lucene"
    }
  }
}

   
2. Access Discover page
3. Load index pattern
4. Apply filter on index-pattern field
5. Screen goes blank

**Expected behavior**
 The discover page should show the  "No results found" message, when applying filter and got no data.

**Impact**
Affects all users accessing the Discover interface
Disrupts observability operations
Prevents proper data exploration and analysis

**Temporary Fix Applied**
 1. Identified and removed the problematic saved query
2. System functionality restored after query removal

**Action Items**
[ ] Add type checking for query objects before rendering
[ ] Implement input validation for saved queries
[ ] Add error handling for malformed queries
[ ] Diagnose how this wrong query format entered.

**Screenshots**

<img width="1673" height="1091" alt="Image" src="https://github.com/user-attachments/assets/d8d68901-a852-496e-b33b-054e008f7eb1" />

**Additional context**
Environment: Production
Version: OpenSearch 3.1.0, 2.19
Authentication: Both SAML and non-SAML affected

cwperks:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/cwperks) [3](https://github.com/imdhemy)

@ruchidh are you actively working on this. Can I assign this to you?

cwperks:
Please reply on this thread with any code deep dives in case anyone in the community would like to contribute a fix.

ruchidh:
i will look into this immediately

ruchidh:
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11347 - pr

PR #11347, opened 2026-02-13
Fix blank discover bug for json format saved queries

ruchidh:
### Description

Discover page displays blank screen when saved queries contain object-type query values (OSD 3.1.0, 2.19).

**Fixed:** Critical UI rendering issue in Discover page when saved queries contain object-type query values. 
Added type guards and error handling to prevent blank screen errors.

Screenshot 2026-03-16 at 11 03 59 AM

### Problematic Saved Query Format:

Screenshot 2026-03-11 at 11 28 36 AM

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/10883

## Changelog

- fix: Fix blank discover because of json type queries

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


LDrago27:
@ruchidh Should we add some tests to validate the fix. We can either do a cypress test or an unit test to account for the bug

ruchidh:
> @ruchidh Should we add some tests to validate the fix. We can either do a cypress test or an unit test to account for the bug

thanks suchit, added cases

src/plugins/data/public/query/saved_query/saved_query_service.ts
Maosaic: One gap worth noting: parseSavedQueryObject still has an unguarded property access at the top of the function:
```
const queryString = savedQuery.attributes.query.query as string;
```
If attributes.query is missing entirely (e.g. a corrupted or externally-created saved object), this throws a TypeError before the try/catch is reached.

A guard like the following would cover this:
```
const queryString = savedQuery.attributes.query?.query as string | undefined;
```

ruchidh: noted,

Issue #11091, opened 2025-12-23
[BUG] Embed mode causes incorrect height calculation for Discover layout, leaving empty space at bottom


untitledds:
**Describe the bug**

In the embedded mode (`app/data-explorer/discover#?embed`), the Discover panel's main container (`.euiPage.deLayout`) does not calculate its height correctly, resulting in a large empty space at the bottom of the page. This is caused by a CSS specificity conflict where the `.headerIsExpanded .deLayout` rule (`height: calc(100vh - 98px)`) overrides the base `.deLayout` rule (`height: calc(100vh - 49px)`). Neither calculation is accurate for the embedded context, leading to incorrect layout.

**To Reproduce**
1. Navigate to the Discover page in embedded mode: `https://<your-opensearch-dashboards-url>/app/data-explorer/discover#?embed`
2. Observe the bottom of the page; a significant empty area will be visible.
3. Open DevTools and inspect the element with class `.euiPage.deLayout`.
4. Note that the applied `height` property is `calc(100vh - 98px)` (or sometimes incorrectly forced to `100vh` as a temporary workaround).
5. Verify that this value does not account for the actual height of fixed elements in the embedded view.

**Expected behavior**
The `.euiPage.deLayout` container should dynamically calculate its height to perfectly fill the available vertical space within the iframe or embedded context, without leaving any empty gaps at the top or bottom. The calculation must be specific to the `embed` mode and should not rely on rules intended for the full Dashboards UI (e.g., those accounting for the expanded header).

**OpenSearch Version**
[Please specify your OpenSearch version, e.g., 2.19.4]

**Dashboards Version**
2.19.4

**Plugins**
- alertingDashboards@2.19.4.0
- anomalyDetectionDashboards@2.19.4.0
- assistantDashboards@2.19.4.0
- customImportMapDashboards@2.19.4.0
- flowFrameworkDashboards@2.19.4.0
- ganttChartDashboards@2.19.4.0
- indexManagementDashboards@2.19.4.0
- mlCommonsDashboards@2.19.4.0
- notificationsDashboards@2.19.4.0
- observabilityDashboards@2.19.4.0
- queryInsightsDashboards@2.19.4.0
- queryWorkbenchDashboards@2.19.4.0
- reportsDashboards@2.19.4.0
- searchRelevanceDashboards@2.19.4.0
- securityAnalyticsDashboards@2.19.4.0
- securityDashboards@2.19.4.0

**Screenshots**
[...]

PR #11350, opened 2026-02-15
Fix Discover embed mode layout height calculation

drtootsie:
Addresses #11091

## Problem
In embed mode (`?embed`), the Discover panel's main container (`.euiPage.deLayout`) incorrectly calculates its height by subtracting header offsets that don't exist in the embedded context, resulting in a large empty space at the bottom of the page.

## Solution
Detect embed mode and apply a dedicated CSS class that uses full viewport height without header compensation.

## Changes
```diff
+ const isEmbedded = new URLSearchParams(params.history.location.search).has('embed');

  className={classNames(
    'deLayout',
    isEnhancementsEnabled && !showActionsInGroup ? 'dsc--next' : undefined,
+   isEmbedded ? 'deLayout--embed' : undefined
  )}
```

```scss
+ &--embed {
+   height: 100vh;
+ }
```

## Testing
- Navigate to `/app/data-explorer/discover#?embed`
- Verify no empty space at bottom
- Verify content fills viewport correctly
- Test with different screen sizes

## Impact
- Embedded Discover views now use full viewport height
- No impact on standard (non-embedded) UI
- Fixes layout issue reported in #11091

### DIFF
--- a/src/plugins/data_explorer/public/components/app_container.scss
+++ b/src/plugins/data_explorer/public/components/app_container.scss
@@ -17,6 +17,10 @@ $osdHeaderOffset: $euiHeaderHeightCompensation;
     height: calc(100vh - #{$osdHeaderOffset * 2});
   }
 
+  &--embed {
+    height: 100vh;
+  }
+
   &__canvas {
     height: 100%;
   }
--- a/src/plugins/data_explorer/public/components/app_container.tsx
+++ b/src/plugins/data_explorer/public/components/app_container.tsx
@@ -28,6 +28,7 @@ import { DISCOVER_LOAD_EVENT, NEW_DISCOVER_LOAD_EVENT, trackUiMetric } from '../
 export const AppContainer = React.memo(
   ({ view, params }: { view?: View; params: AppMountParameters }) => {
     const isMobile = useIsWithinBreakpoints(['xs', 's', 'm']);
+    const isEmbedded = new URLSearchParams(params.history.location.search).has('embed');
 
     const opensearchDashboards = useOpenSearchDashboards<IDataPluginServices>();
     const { uiSettings } = opensearchDashboards.services;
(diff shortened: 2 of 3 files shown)


src/plugins/data_explorer/public/components/app_container.tsx
```diff
@@ -28,6 +28,7 @@ import { DISCOVER_LOAD_EVENT, NEW_DISCOVER_LOAD_EVENT, trackUiMetric } from '../
 export const AppContainer = React.memo(
   ({ view, params }: { view?: View; params: AppMountParameters }) => {
     const isMobile = useIsWithinBreakpoints(['xs', 's', 'm']);
+    const isEmbedded = new URLSearchParams(params.history.location.search).has('embed');
```
zengyan-amazon, line 31: This JS-based embed detection is unnecessary. The core rendering service already sets the `hidden-chrome` class on `#app-wrapper` when embed mode is active (see `src/core/public/rendering/app_containers.tsx` line 60). You can use that as a CSS-only ancestor selector instead, which avoids adding embed-detection logic to this component entirely.

Suggested approach — delete this line and the `classNames` change, and use a pure SCSS fix instead (see my comment on the SCSS file).

src/plugins/data_explorer/public/components/app_container.scss
```diff
@@ -17,6 +17,10 @@ $osdHeaderOffset: $euiHeaderHeightCompensation;
     height: calc(100vh - #{$osdHeaderOffset * 2});
   }
 
+  &--embed {
```
zengyan-amazon, line 20: Two issues with this rule:

1. **Missing `dsc--next` override.** The `.deLayout.dsc--next` variant uses `calc(100vh - 3*header)` (and `2*header` in non-expanded mode). This embed rule won't override those because `.deLayout.dsc--next` is more specific. So embed mode is still broken when `dsc--next` is active.

2. **Can be done without JS.** The core already applies `hidden-chrome` to `#app-wrapper` in embed mode. A CSS-only fix that handles all variants:

```scss
// In embed mode the header is hidden, so no offset is needed
// stylelint-disable-next-line @osd/stylelint/no_modifying_global_selectors
.hidden-chrome .deLayout {
  height: 100vh;

  &.dsc--next {
    height: 100vh;
  }
}
```

This goes after the `.headerIsExpanded .deLayout` block. It overrides all height variants, requires zero JS changes, and works because `.hidden-chrome` is already on an ancestor element in embed mode.

Issue #11070, opened 2025-12-17
[BUG] Duplicate Shortcut "focus_query_bar" from plugin "data" and "explore"


DdangJin:
**Describe the bug**

When attempting to re-register a focus_query_bar shortcut in expansion plugin while there is a focus_query_bar registered from the data plugin, an error screen is output.

**To Reproduce**

0. Set OSD Config `explore.enabled: true`
1. Go to 'Visualize' and create 'Data Table'
2. Add 'Filter' in Bucket
3. See error

**Expected behavior**
There should be no error

**OpenSearch Version**
Please list the version of OpenSearch being used.

- 3.3.2

**Dashboards Version**
Please list the version of OpenSearch Dashboards being used.

- 3.3.0

**Plugins**

Please list all plugins currently enabled.

**Screenshots**

<img width="1232" height="324" alt="Image" src="https://github.com/user-attachments/assets/561b653f-7997-4a9e-8995-97b014f768f7" />

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

When I only looked at the code, I think it was caused by trying to register in duplicate in the codes below.

- https://github.com/opensearch-project/OpenSearch-Dashboards/blob/0ff9af49200afe9e6c84888512e23558a925fe8f/src/plugins/data/public/ui/query_string_input/query_string_input.tsx#L537
- https://github.com/opensearch-project/OpenSearch-Dashboards/blob/0ff9af49200afe9e6c84888512e23558a925fe8f/src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_query_panel_editor.ts#L115

hyeyoon23:
I can take a look at this issue.
It seems the error is caused by both data and explore plugins registering the same shortcut id (`focus_query_bar`).
I’ll work on a fix to avoid the duplication.

abodehq:
When this can be download in which release currently we face this issue in using any filter inside visualization.

PR #11351, opened 2026-02-15
Fix duplicate focus_query_bar shortcut ID

drtootsie:
Addresses #11070

## Problem
When `explore.enabled: true` is set, both the `data` and `explore` plugins attempt to register a keyboard shortcut with the same ID (`focus_query_bar`), causing an error screen when creating visualizations with filters.

## Root Cause
- `src/plugins/data/public/ui/query_string_input/query_string_input.tsx` registers `focus_query_bar`
- `src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_query_panel_editor.ts` also registers `focus_query_bar`

## Solution
Rename the explore plugin's shortcut ID to `focus_explore_query_bar` to avoid collision.

## Changes
```diff
keyboardShortcut?.useKeyboardShortcut({
-   id: 'focus_query_bar',
+   id: 'focus_explore_query_bar',
    pluginId: 'explore',
```

## Testing
1. Set `explore.enabled: true` in config
2. Go to Visualize and create Data Table
3. Add Filter in Bucket
4. Verify no error occurs
5. Test that '/' shortcut still focuses query bar

## Impact
- No functional change - shortcut still works the same
- Fixes error when using explore plugin
- No breaking changes

### DIFF
--- a/src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_query_panel_editor.ts
+++ b/src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_query_panel_editor.ts
@@ -123,7 +123,7 @@ export const useQueryPanelEditor = (): UseQueryPanelEditorReturnType => {
   }, [queryLanguage]);
 
   keyboardShortcut?.useKeyboardShortcut({
-    id: 'focus_query_bar',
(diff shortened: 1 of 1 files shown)


abbyhu2000:
Fix is not sufficient and does not resolve the specific issue.

PR #11353, opened 2026-02-17
Change overload worker url in devtools for CSP violation

sumukhswamy:
### Description

The dev tools plugin was generating a Content Security Policy (CSP) violation warning in the browser:

image

This occurred because the Ace editor's creates Web Workers by:

- Taking worker JavaScript source as a string (via raw-loader)
- Creating a Blob from that source
- Generating a blob URL with 
- Instantiating new Worker(blobUrl)
- The CSP policy [worker-src 'self'] only permits workers loaded from the same origin (server), not from blob URLs.

Implemented a minimal URL override pattern inspired by [PR #10860] that fixed the same issue for Monaco editor.

## Changelog
- feat: Fix CSP violation for Console worker by loading from URL instead of blob

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/11359.yml
+++ b/changelogs/fragments/11359.yml
@@ -1,0 +1,2 @@
+feat:
+- Add default fetch_size=500 for PPL queries in Discover ([#11359](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11359))
--- a/src/plugins/query_enhancements/server/search/ppl_search_strategy.test.ts
+++ b/src/plugins/query_enhancements/server/search/ppl_search_strategy.test.ts
@@ -26,7 +26,7 @@ describe('pplSearchStrategyProvider', () => {
   let logger: Logger;
   let client: ILegacyClusterClient;
   let usage: SearchUsage;
-  const emptyRequestHandlerContext = ({} as unknown) as RequestHandlerContext;
+  let mockRequestHandlerContext: RequestHandlerContext;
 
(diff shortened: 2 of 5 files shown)


src/plugins/console/public/plugin.ts
```diff
@@ -33,12 +33,22 @@ import { Plugin, CoreSetup } from 'src/core/public';
...
 export class ConsoleUIPlugin implements Plugin<void, void, AppSetupUIPluginDependencies> {
+  private originalCreateObjectURL?: typeof URL.createObjectURL;
+
   public setup(
     { notifications, getStartServices, http }: CoreSetup,
     { devTools, home, usageCollection }: AppSetupUIPluginDependencies
   ) {
+    this.originalCreateObjectURL = URL.createObjectURL;
+    URL.createObjectURL = (obj: Blob | MediaSource) => {
+      if (obj instanceof Blob && obj.type === 'application/javascript') {
```
LDrago27, line 47: Why are we creating the workerURL only for a specific type.

sumukhswamy: Because I only want to intercept the worker blob that Ace creates (JavaScript). The CSP issue is specifically for workers other [URL.createObjectURL] uses (images, docs, downloads) shouldn’t be redirected to the worker file. The other implications would be for ex reporting plugin as Justin and I discussed which would break unrelated functionality.

Issue #9385, opened 2025-02-13
[BUG] Remove limit size in default PPL query once fetch_size is working for PPL requests


abbyhu2000:
**Describe the bug**

Current fetch_size param is not working when sending PPL requests due to: https://github.com/opensearch-project/sql/issues/3314

In order to get responses with sample size defined in advanced setting, we interpolate the default PPL query string to contain the limit. 
ex. `source = dataset | head 500`,

We should remove the limit in the default PPL query string once fetch_size is working and just uses the fetch_size to send the request.

**Expected behavior**
PPL default query should not contain size limit; but use the param fetch_size when sending request to define the size. Just like how SQL query behaves.

**OpenSearch Version**
Please list the version of OpenSearch being used.

**Dashboards Version**
Please list the version of OpenSearch Dashboards being used.

**Plugins**

Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

kavilla:
hey @abbyhu2000 is this a bug? or just a reminder in the future we should remove the limit once fetch size is supported?

PR #11359, opened 2026-02-18
Onboard fetch_size API for PPL Queries on OSD

ahkcs:
### Description

Adopt the `fetch_size` API for PPL queries in Discover/Explore. Every PPL query executed through the sync PPL search strategy now includes `fetch_size=500` in the request body sent to `POST /_plugins/_ppl`. This leverages the new `fetch_size` parameter introduced in the SQL plugin ([opensearch-project/sql#5109](https://github.com/opensearch-project/sql/pull/5109)), which injects a `Head` AST node to limit the number of rows returned at the OpenSearch level rather than fetching the full result set.

**Changes:**
- Added `PPL_DEFAULT_FETCH_SIZE = 500` constant in `query_enhancements/common/constants.ts`
- PPL search strategy injects `fetchSize` into the request body before calling the Facet
- Facet maps `fetchSize` (camelCase) to `fetch_size` (snake_case) in the OpenSearch API request body
- Only applies to the sync PPL path (`/_plugins/_ppl`); async PPL queries for S3 datasets (`/_plugins/_async_query`) are unaffected

### Issues Resolved

closes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9385

## Screenshot

Screenshot 2026-02-18 at 1 24 39 PM

## Testing the changes

1. Start OpenSearch with the SQL plugin that includes [sql#5109](https://github.com/opensearch-project/sql/pull/5109)
2. Start OpenSearch Dashboards: `yarn start`
3. Navigate to Discover, select a PPL dataset, and run queries:
   - `source=<index>` — verify at most 500 rows returned; OpenSearch log shows `| head 500`
   - `source=<index> | head 10` — verify 10 rows returned (smaller limit wins); log shows `| head 10 | head 500`
   - `source=<index> | head 1000` — verify 500 rows returned (fetch_size wins); log shows `| head 1000 | head 500`
   - `source=<index> | stats count() by span(@timestamp, 1d)` — verify aggregation query also receives `fetch_size`
4. Run unit tests:
   ```bash
   yarn test:jest src/plugins/query_enhancements/server/utils/facet.test.ts
   yarn test:jest src/plugins/query_enhancements/server/search/ppl_search_strategy.test.ts
   ```

## Changelog

- feat: Add default fetch_size=500 for PPL queries in Discover

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/common/constants.ts
+++ b/src/plugins/query_enhancements/common/constants.ts
@@ -70,8 +70,6 @@ export const UI_SETTINGS = {
   STATE_STORE_IN_SESSION_STORAGE: 'state:storeInSessionStorage',
 };
 
-export const PPL_DEFAULT_FETCH_SIZE = 500;
-
 export const ERROR_DETAILS = { GUARDRAILS_TRIGGERED: 'guardrails triggered' };
 
 export const S3_PARTITION_INFO_COLUMN = '# Partition Information';
--- a/src/plugins/query_enhancements/server/search/ppl_search_strategy.test.ts
+++ b/src/plugins/query_enhancements/server/search/ppl_search_strategy.test.ts
@@ -15,7 +15,6 @@ import { SearchUsage } from '../../../data/server';
 import * as utils from '../../common/utils';
 import * as facet from '../utils/facet';
 import { pplSearchStrategyProvider } from './ppl_search_strategy';
-import { PPL_DEFAULT_FETCH_SIZE } from '../../common';
 
 jest.mock('../../common/utils', () => ({
   ...jest.requireActual('../../common/utils'),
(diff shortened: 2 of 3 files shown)


src/plugins/query_enhancements/common/constants.ts
LDrago27, line 1: Instead of setting this size as default why don't we combine it with the following setting. This setting used to regulate the fetch size in DQL/Lucene and we can probably use it with for PPL/SQL. 

Image

ahkcs: Great suggestion! I have updated the fetch_size to read from `discover:sampleSize`
So that it's easily configurable now

LDrago27: Thanks. looks great.

LDrago27 (approved):

src/plugins/query_enhancements/common/constants.ts
LDrago27: nit: maybe we can update the setting description to reflect it but we can handle it later.

PR #11395, opened 2026-02-26
[BUG] Fix duplicate shortcut registration for focus_query_bar

### LINKED ISSUE #11070: [BUG] Duplicate Shortcut "focus_query_bar" from plugin "data" and "explore"

scuba3198:
### Description
This PR fixes the duplicate keyboard shortcut registration for the `focus_query_bar` shortcut. It resolves the crash reported in issue #11070 by ensuring that each instance of the query bar in the `data` plugin registers a unique shortcut ID. It also renames the shortcut in the `explore` plugin to prevent cross-plugin conflicts.

### Issues Resolved
Closes #11070

## Screenshot
(No UI changes, logic only)

## Testing the changes
Verified with a reproduction script simulating the shortcut registration logic. The script confirms that multiple component instances can now register shortcuts for the same key combination without throwing a "Duplicate Shortcut" error.

## Changelog

- skip

### Check List
- [ ] All tests pass
  - [ ] `yarn test:jest`


scuba3198:
I noticed PR #11351 also addresses the cross-plugin conflict. However, this PR (#11395) goes further — in addition to the explore plugin rename, it adds a `registerKeyboardShortcut` prop to  **QueryStringInput** (defaulting to `false`) so the shortcut is only registered by the top-level  **QueryBarTopRow**, preventing any duplicate registration, even when multiple query bar instances are on the same page.

src/plugins/data/public/ui/query_string_input/query_string_input.tsx
```diff
@@ -532,9 +532,9 @@ export default class QueryStringInputUI extends Component<Props, State> {
 
     // Register keyboard shortcut for focusing query input using direct service registration
     const { keyboardShortcut } = this.services;
-    if (keyboardShortcut) {
+    if (keyboardShortcut && this.textareaId) {
       keyboardShortcut.register({
-        id: 'focus_query_bar',
+        id: `focus_query_bar_${this.textareaId}`,
```
zengyan-amazon: **Behavioral concern:** Making the ID unique per instance fixes the crash, but now every `QueryStringInput` instance (main search bar + each filter bucket) registers a separate shortcut on the same `/` key.

The `KeyboardShortcutService` executes the **last registered** shortcut for a given key (`shortcuts[shortcuts.length - 1]`), so pressing `/` will focus whichever `QueryStringInput` mounted last — likely a filter input, not the primary search bar.

Consider only registering the shortcut when this component is the primary query bar. For example, add a prop like `registerKeyboardShortcut` (default `false`) and only pass `true` from the top-level `SearchBar`/`QueryBarTopRow`. This would also avoid polluting the keyboard shortcuts help modal with duplicate "Focus query bar" entries.

src/plugins/data/public/ui/query_string_input/query_string_input.tsx
```diff
@@ -532,9 +532,9 @@ export default class QueryStringInputUI extends Component<Props, State> {
 
     // Register keyboard shortcut for focusing query input using direct service registration
     const { keyboardShortcut } = this.services;
-    if (keyboardShortcut) {
+    if (keyboardShortcut && this.textareaId) {
```
zengyan-amazon: The `this.textareaId` guard is fine defensively, but `textareaId` is a class property initialized inline at line 521 via `htmlIdGenerator()()`, so it is always set before `componentDidMount` runs. The guard will never be false.

src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_query_panel_editor.ts
```diff
@@ -123,7 +123,7 @@ export const useQueryPanelEditor = (): UseQueryPanelEditorReturnType => {
   }, [queryLanguage]);
 
   keyboardShortcut?.useKeyboardShortcut({
-    id: 'focus_query_bar',
+    id: 'focus_explore_query_bar',
```
zengyan-amazon, line 188: Good rename — this prevents the cross-plugin namespaced ID conflict.

Separate issue: the `execute` callback at line 135 is an inline arrow function. Since `useKeyboardShortcut` includes `shortcut.execute` in its `useEffect` dependency array, every re-render creates a new function reference which triggers unregister + re-register. Consider wrapping it with `useCallback`.

scuba3198: Thanks for the feedback @zengyan-amazon! I've updated the PR to address both points:
1. In the data plugin, the ID was reverted to `focus_query_bar` and I added a `registerKeyboardShortcut` prop to `QueryStringInput` so only the top-level `QueryBarTopRow` registers the shortcut.
2. In the explore plugin, the `execute` callback was wrapped in `useCallback` to prevent re-registration cycles.

abbyhu2000 (approved):

LDrago27 (approved):

PR #11413, opened 2026-02-27
[Explore] Optimize initial workflow and remove redundant saved_objects calls

Maosaic:
## Summary

This PR significantly improves the initial page load performance of the Explore application by optimizing DataView fetching and eliminating redundant saved object requests. The optimization reduces network requests by 60-80% and improves load times through client-side caching and bulk fetching.

## Problem Statement

**Before this change:**
- Multiple components were independently fetching the same DataView objects, causing duplicate network requests
- Each DataView fetch required a separate HTTP round trip, even when requesting the same object
- Redux persistence was unnecessarily fetching DataViews just to read properties that were already available
- No client-side caching meant repeated requests for the same objects throughout the page lifecycle
- Sequential individual `get()` calls instead of batched `bulkGet()` operations

## Solution

This PR implements a multi-layered optimization strategy:

### 1. Bulk Fetching with `getMultiple()`

- Added new `DataViewsService.getMultiple()` method that uses `bulkGet()` API
- Fetches multiple DataViews in a single HTTP request
- Checks cache first, only fetches uncached items
- Processes all DataViews in parallel

**Benefits:**
- Single network request instead of N sequential requests
- Reduced network latency and server processing overhead
- Better utilization of OpenSearch's bulk operations

### 2. Enhanced Request Batching

- Improved batch queue deduplication in `SavedObjectsClient`
- Multiple concurrent requests for the same object are deduplicated within 100ms window
- Each caller receives their own cloned instance (prevents shared mutable state)

**Benefits:**
- Eliminates duplicate network requests during component initialization
- Maintains data isolation between components
- Simpler architecture without persistent caching complexity

### 3. Removed Redundant DataView Fetches

**a) Redux Persistence Optimization:**
- **Before:** Fetched entire DataView just to check `signalType` property
- **After:** Reads `signalType` directly from Dataset object (already in Redux state)
- **Impact:** Eliminated 1 unnecessary fetch per page load

**b) DatasetSelectWidget Optimization:**
- **Before:** Had redundant `useEffect` that fetched DataView on every render
- **After:** Removed the effect; batching now handled by DatasetSelect component
- **Impact:** Eliminated early individual `get()` calls that prevented bulk fetching

**c) DatasetSelect Fallback Enhancement:**
- **Before:** Had no fallback for non-index-pattern datasets (e.g., PROMETHEUS)
- **After:** Added fallback logic to handle datasets from URL/saved queries
- **Impact:** Supports all dataset types while maintaining performance

## Screenshot

Before:
Screenshot 2026-02-27 at 2 27 06 PM

After:
Screenshot 2026-02-27 at 2 27 54 PM

## Changelog

- fix: [Explore] Optimize initial workflow and remove redundant saved_objects calls

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/agent_traces/public/components/query_panel/query_panel_widgets/dataset_select/dataset_select.test.tsx
+++ b/src/plugins/agent_traces/public/components/query_panel/query_panel_widgets/dataset_select/dataset_select.test.tsx
@@ -189,13 +189,6 @@ describe('DatasetSelectWidget', () => {
     expect(screen.getByTestId('dataset-select')).toBeInTheDocument();
   });
 
-  // NOTE: The following tests were removed because the redundant useEffect that fetched
-  // DataView on component mount was removed as part of the bulkGet optimization.
-  // The DatasetSelect component now handles all dataset fetching internally via getMultiple().
-  // Tests removed:
-  // - "attempts to get dataView on component mount"
-  // - "caches dataset if dataView does not exist"
-
   it('handles dataset selection correctly', async () => {
     renderWithStore();
 
--- a/src/plugins/agent_traces/public/components/query_panel/query_panel_widgets/dataset_select/dataset_select.tsx
+++ b/src/plugins/agent_traces/public/components/query_panel/query_panel_widgets/dataset_select/dataset_select.tsx
@@ -32,14 +32,6 @@ export const DatasetSelectWidget = () => {
     },
   } = services;
 
-  // REMOVED: Redundant useEffect that was fetching DataView on every render
-  // This was causing early individual get() calls before the DatasetSelect component's
-  // getMultiple() optimization could batch fetch all DataViews together.
-  // The dataset caching is now handled by:
-  // 1. DatasetSelect.fetchDatasets() - fetches all datasets via bulkGet
-  // 2. DatasetProvider - fetches current dataset if needed
-  // 3. Query execution - fetches dataset when executing queries
-
   const handleDatasetSelect = useCallback(
     async (dataset: Dataset | undefined) => {
       try {
--- a/src/plugins/explore/public/components/query_panel/query_panel_widgets/dataset_select/dataset_select.tsx
+++ b/src/plugins/explore/public/components/query_panel/query_panel_widgets/dataset_select/dataset_select.tsx
@@ -30,14 +30,6 @@ export const DatasetSelectWidget = () => {
     },
   } = services;
 
-  // REMOVED: Redundant useEffect that was fetching DataView on every render
-  // This was causing early individual get() calls before the DatasetSelect component's
-  // getMultiple() optimization could batch fetch all DataViews together.
-  // The dataset caching is now handled by:
-  // 1. DatasetSelect.fetchDatasets() - fetches all datasets via bulkGet
-  // 2. DatasetProvider - fetches current dataset if needed
-  // 3. Query execution - fetches dataset when executing queries
-
   const handleDatasetSelect = useCallback(
     async (dataset: Dataset | undefined) => {
       try {
--- a/cypress/utils/commands.explore.js
+++ b/cypress/utils/commands.explore.js
@@ -555,7 +555,7 @@ cy.explore.add(
   }
 );
 
-cy.explore.add('createVisualizationWithQuery', (query, chartType, datasetName) => {
+cy.explore.add('createVisualizationWithQuery', (query, chartType, datasetName, options) => {
   cy.explore.clearQueryEditor();
   cy.explore.setDataset(datasetName, DATASOURCE_NAME, 'INDEX_PATTERN');
   setDatePickerDatesAndSearchIfRelevant('PPL');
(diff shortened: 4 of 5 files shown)


src/core/public/saved_objects/saved_objects_client.ts
```diff
@@ -228,24 +228,42 @@ export class SavedObjectsClient {
...
+
+      queue.forEach((item) => {
+        const key = `${item.type}:${item.id}`;
+        if (!seenObjects.has(key)) {
+          seenObjects.set(key, [item]);
+          deduplicatedQueue.push(item);
+        } else {
+          seenObjects.get(key)!.push(item);
+        }
+      });
```
joshuali925, line 244: this is ok but looks like a workaround to duplicated requests. why were there duplicated requests and would they cause other problems?

src/core/public/saved_objects/saved_objects_client.ts
```diff
@@ -228,24 +228,42 @@ export class SavedObjectsClient {
       const queue = [...this.batchQueue];
       this.batchQueue = [];
 
+      // Deduplicate requests in the queue - if multiple requests for the same object exist,
+      // keep only the first one and resolve all duplicates with the same result
+      const deduplicatedQueue: BatchQueueEntry[] = [];
+      const seenObjects = new Map<string, BatchQueueEntry[]>();
```
joshuali925, line 234: would this introduce cache validation problems?

src/plugins/agent_traces/public/components/query_panel/query_panel_widgets/dataset_select/dataset_select.test.tsx
```diff
@@ -189,35 +189,12 @@ describe('DatasetSelectWidget', () => {
...
-        false
-      );
-    });
-  });
+  // NOTE: The following tests were removed because the redundant useEffect that fetched
+  // DataView on component mount was removed as part of the bulkGet optimization.
+  // The DatasetSelect component now handles all dataset fetching internally via getMultiple().
+  // Tests removed:
+  // - "attempts to get dataView on component mount"
+  // - "caches dataset if dataView does not exist"
```
joshuali925: remove the comment

src/plugins/agent_traces/public/components/query_panel/query_panel_widgets/dataset_select/dataset_select.tsx
```diff
@@ -29,50 +29,16 @@ export const DatasetSelectWidget = () => {
...
-      isMounted = false;
-    };
-  }, [currentQuery, dataViews, queryString, services]);
+  // REMOVED: Redundant useEffect that was fetching DataView on every render
+  // This was causing early individual get() calls before the DatasetSelect component's
+  // getMultiple() optimization could batch fetch all DataViews together.
+  // The dataset caching is now handled by:
+  // 1. DatasetSelect.fetchDatasets() - fetches all datasets via bulkGet
+  // 2. DatasetProvider - fetches current dataset if needed
+  // 3. Query execution - fetches dataset when executing queries
```
joshuali925: same here, remove the comment if it's no longer relevant

src/plugins/explore/public/components/query_panel/query_panel_widgets/dataset_select/dataset_select.tsx
```diff
@@ -21,58 +20,23 @@ export const DatasetSelectWidget = () => {
...
-
-    const handleDataset = async () => {
-      if (currentQuery.dataset) {
-        const dataView = await dataViews.get(
-          currentQuery.dataset.id,
-          currentQuery.dataset.type !== DEFAULT_DATA.SET_TYPES.INDEX_PATTERN
-        );
-
-        if (!dataView) {
-          await queryString.getDatasetService().cacheDataset(
```
joshuali925, line 47: is it ok to remove this?

src/core/public/saved_objects/saved_objects_client.ts
Maosaic: One common senario, we fetch data source information when creating `dataView`,  when loading the list of indexPattern/dataSet on Disover/Explore page, different indexPattern/dataSet can point to the same data source. But now this senario is handle by `getMultiple` separately.

This deduplication code here act like last defence resort, but since our default batch window is only 100ms, for most cases this will not be triggered.

src/core/public/saved_objects/saved_objects_client.ts
Maosaic: Not likely, this map is only being referenced during processing queue, GC should take care of it after.

src/plugins/explore/public/components/query_panel/query_panel_widgets/dataset_select/dataset_select.tsx
Maosaic: Yes, this dataset_select widgets is a wrapper in explore for the dataset_select component in data plugin, fetching DataView is handle by dataset_select component directly, so doing the same in the wrapper is redundent.

src/core/public/saved_objects/saved_objects_client.ts
```diff
@@ -228,24 +228,42 @@ export class SavedObjectsClient {
       const queue = [...this.batchQueue];
       this.batchQueue = [];
 
+      // Deduplicate requests in the queue - if multiple requests for the same object exist,
+      // keep only the first one and resolve all duplicates with the same result
+      const deduplicatedQueue: BatchQueueEntry[] = [];
```
ruchidh, line 233: do we not have changes in unit test for saved_object_client ?

src/core/public/saved_objects/saved_objects_client.ts
```diff
@@ -228,24 +228,42 @@ export class SavedObjectsClient {
...
+              queueItem.resolve(this.createSavedObject(cloneDeep(foundObject)));
+            } else {
+              queueItem.resolve(
+                this.createSavedObject(pick(queueItem, ['id', 'type']) as SavedObject)
+              );
+            }
+          });
         });
       } catch (err) {
         queue.forEach((queueItem) => {
```
ruchidh, line 269: Nit: queue is the original undeduped list?, i guess this will work. 
So this means all  queue (not deduplicatedQueue) will recieve the rejection ?

worth clarifying a comment if doing intentionally.

Maosaic: The error catch behaviour remain unchange compare to the original undeduped list.

PR #11354, opened 2026-02-17
fix(autocomplete): remove registering providers in render fn

joshuali925:
### Description

Sometimes there are duplicated entries for each suggestion candidate in the query editor. The component is registering providers in the render function, which might be the root cause.

## Testing the changes

UT

## Changelog

- fix: fix duplicated autocomplete suggestions

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #11428, opened 2026-03-03
[PPL Autocomplete] Adopt backend grammar bundle as runtime source of truth with safe fallback

mengweieric:
### Description

This PR adopts the backend PPL grammar bundle as the runtime source of truth for the **simplified PPL autocomplete and validation path**, while keeping the compiled grammar path as the fallback where runtime grammar is unavailable or not used.

### Key Changes

- Added a datasource-aware PPL grammar bundle route:
  - `GET /api/enhancements/ppl/grammar`
  - forwards to backend `/_plugins/_ppl/_grammar`
  - supports optional `dataSourceId` for remote datasource requests

- Added client-side runtime grammar caching and warmup support:
  - caches backend grammar artifacts for the active datasource
  - retries failed fetches after a cooldown
  - resets cache state when the runtime grammar feature flag is disabled
  - notifies listeners when the active grammar changes

- Switched the **simplified PPL autocomplete path** to try runtime grammar suggestions first:
  - uses backend grammar bundle metadata with `LexerInterpreter` / `ParserInterpreter` + C3
  - falls back to the existing compiled simplified grammar path when runtime grammar is unavailable or produces no actionable result

- Kept the **default/compiled autocomplete path** intact:
  - `getDefaultSuggestions()` continues to use the compiled grammar path unchanged

- Added runtime validation support:
  - validates PPL using the cached backend grammar when runtime grammar is enabled for the active model/context
  - falls back to the compiled validator when runtime validation is unavailable
  - supports pipe-first queries by stripping the leading pipe and remapping errors back to the original editor positions

- Fixed SPA navigation provider loss:
  - Monaco `onLanguage` is one-shot and won't fire for already-encountered languages
  - after unmount/remount, providers are now registered directly in `render()` as well as via `onLanguage`

- Gated by `query:enhancements:runtimePplGrammar` advanced setting:
  - enabled by default (`true`)
  - when disabled, clears cached grammar and falls back to the compiled grammar path
  - requires page reload to take effect
  - listed under the Search category in Advanced Settings as "Enable runtime PPL grammar"

### Runtime Correctness / Safety Improvements

- Isolated C3 follow-set cache entries by `grammarHash` to avoid cross-grammar token pollution
- Added rerun logic when preferred rules hide token candidates
- Filtered invalid token candidates (for example EOF / unresolved vocab entries)
- Added pipe-first start-rule selection using `pipeStartRuleIndex` with rule-name fallback
- Made parsing non-fatal for partial input and used a synthetic completion context when needed
- Removed compiled-ID coupling from the runtime path where possible by preferring rule/token names
- Preserved supported symbolic-only runtime tokens so snippet-style suggestions can still render
- Added retry cooldown for failed grammar fetches instead of permanently disabling runtime grammar for the session
- Fixed validation-context freshness and editor subscription cleanup

### Compatibility / Scope

- The compiled/default autocomplete path remains available and unchanged
- The runtime path is designed to tolerate backend grammar token/rule renumbering by resolving rules/tokens by name where possible
- Runtime autocomplete behavior is currently applied to the simplified PPL suggestion path; the default PPL suggestion path remains compiled-based

### Testing

- Unit tests cover:
  - runtime autocomplete helper behavior
  - simplified autocomplete runtime + fallback scenarios
  - compiled/default autocomplete behavior
  - grammar cache behavior
  - runtime validation and validation-context wiring

- Manual validation performed for:
  - local cluster flow
  - datasource-backed flow
  - autocomplete fallback behavior
  - runtime validation behavior

### related PR
- https://github.com/opensearch-project/sql/pull/5162

## Changelog

- feat: Adopt backend grammar bundle as runtime source for PPL autocomplete

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/antlr/opensearch_ppl/ppl_grammar_cache.ts
+++ b/src/plugins/data/public/antlr/opensearch_ppl/ppl_grammar_cache.ts
@@ -9,6 +9,7 @@ import {
   SavedObjectsClientContract,
 } from 'opensearch-dashboards/public';
 import { ATN, ATNDeserializer, Vocabulary } from 'antlr4ng';
+import semver from 'semver';
 import { PPLGrammarBundle } from './ppl_bundle_loader';
 import { TokenDictionary } from '../opensearch_sql/table';
 
@@ -160,14 +161,8 @@ class PPLGrammarCache {
    */
   shouldFetchFromBackend(version?: string): boolean {
     if (!version) return false;
-
-    const match = version.trim().match(/^(\d+)\.(\d+)/);
-    if (!match) return false;
-    const major = Number(match[1]);
-    const minor = Number(match[2]);
-    if (!Number.isFinite(major) || !Number.isFinite(minor)) return false;
-
-    return major > 3 || (major === 3 && minor >= 6);
+    const coerced = semver.coerce(version);
+    return coerced ? semver.satisfies(coerced.version, '>=3.6.0') : false;
   }
 
   getCachedGrammar(datasourceId?: string): CachedGrammar | null {
--- a/src/plugins/data/public/antlr/opensearch_ppl/ppl_grammar_cache.test.ts
+++ b/src/plugins/data/public/antlr/opensearch_ppl/ppl_grammar_cache.test.ts
@@ -57,15 +57,40 @@ describe('ppl_grammar_cache', () => {
   });
 
   it('should gate backend grammar fetch by OpenSearch version', () => {
-    expect(pplGrammarCache.shouldFetchFromBackend(undefined)).toBe(false);
-    expect(pplGrammarCache.shouldFetchFromBackend('3.5.9')).toBe(false);
+    // --- supported (>= 3.6.0) ---
     expect(pplGrammarCache.shouldFetchFromBackend('3.6.0')).toBe(true);
-    expect(pplGrammarCache.shouldFetchFromBackend('3.6.0-SNAPSHOT')).toBe(true);
-    expect(pplGrammarCache.shouldFetchFromBackend('3.5.9-SNAPSHOT')).toBe(false);
+    expect(pplGrammarCache.shouldFetchFromBackend('3.6.1')).toBe(true);
     expect(pplGrammarCache.shouldFetchFromBackend('3.10.1')).toBe(true);
     expect(pplGrammarCache.shouldFetchFromBackend('4.0.0')).toBe(true);
-    expect(pplGrammarCache.shouldFetchFromBackend('v3.6.0')).toBe(false);
+
+    // pre-release / snapshot builds
+    expect(pplGrammarCache.shouldFetchFromBackend('3.6.0-SNAPSHOT')).toBe(true);
+    expect(pplGrammarCache.shouldFetchFromBackend('3.7.0-SNAPSHOT')).toBe(true);
+    expect(pplGrammarCache.shouldFetchFromBackend('3.6.0-rc1')).toBe(true);
+    expect(pplGrammarCache.shouldFetchFromBackend('3.6.0-alpha1')).toBe(true);
+    expect(pplGrammarCache.shouldFetchFromBackend('3.6.0-beta1')).toBe(true);
+
+    // four-part versions (OpenSearch build metadata)
+    expect(pplGrammarCache.shouldFetchFromBackend('3.6.1.0')).toBe(true);
+    expect(pplGrammarCache.shouldFetchFromBackend('4.0.0.0')).toBe(true);
+
+    // prefixed strings
+    expect(pplGrammarCache.shouldFetchFromBackend('v3.6.0')).toBe(true);
+    expect(pplGrammarCache.shouldFetchFromBackend('OpenSearch 3.6.0')).toBe(true);
+
+    // --- unsupported (< 3.6.0) ---
+    expect(pplGrammarCache.shouldFetchFromBackend('3.5.9')).toBe(false);
+    expect(pplGrammarCache.shouldFetchFromBackend('3.5.9-SNAPSHOT')).toBe(false);
+    expect(pplGrammarCache.shouldFetchFromBackend('3.0.0')).toBe(false);
+    expect(pplGrammarCache.shouldFetchFromBackend('2.17.0')).toBe(false);
+    expect(pplGrammarCache.shouldFetchFromBackend('2.17.0-SNAPSHOT')).toBe(false);
+    expect(pplGrammarCache.shouldFetchFromBackend('1.3.14')).toBe(false);
+
+    // --- invalid / missing ---
+    expect(pplGrammarCache.shouldFetchFromBackend(undefined)).toBe(false);
+    expect(pplGrammarCache.shouldFetchFromBackend('')).toBe(false);
     expect(pplGrammarCache.shouldFetchFromBackend('invalid')).toBe(false);
+    expect(pplGrammarCache.shouldFetchFromBackend('abc.def.ghi')).toBe(false);
   });
 
   it('should fetch and cache grammar via warmUp when version is supported', async () => {


src/plugins/query_enhancements/server/utils/plugins.ts
```diff
@@ -157,4 +157,9 @@ export const OpenSearchEnhancements = (client: any, config: any, components: any
     method: 'POST',
     needBody: true,
   });
+
+  enhancements.pplBundle = createAction(client, components, {
+    endpoint: URI.PPL_BUNDLE,
```
joshuali925, line 162: why bundle and not grammar?

src/plugins/query_enhancements/server/routes/index.ts
```diff
@@ -113,6 +113,62 @@ export function defineSearchStrategyRouteProvider(logger: Logger, router: IRoute
...
 }
 
+/**
+ * Defines route for PPL grammar bundle endpoint
+ * Forwards requests to OpenSearch /_plugins/_ppl/_grammar
+ */
+export function definePPLBundleRoute(logger: Logger, router: IRouter, client: any) {
+  router.get(
+    {
+      path: '/api/enhancements/ppl/grammar',
```
joshuali925: nit. use const to be consistent with other routes

src/plugins/query_enhancements/server/routes/index.ts
```diff
@@ -113,6 +113,62 @@ export function defineSearchStrategyRouteProvider(logger: Logger, router: IRoute
   };
 }
 
+/**
+ * Defines route for PPL grammar bundle endpoint
+ * Forwards requests to OpenSearch /_plugins/_ppl/_grammar
+ */
+export function definePPLBundleRoute(logger: Logger, router: IRouter, client: any) {
```
joshuali925: unused `client`?

src/plugins/query_enhancements/server/routes/index.ts
```diff
@@ -113,6 +113,62 @@ export function defineSearchStrategyRouteProvider(logger: Logger, router: IRoute
...
+        const result = await opensearchClient.transport.request({
+          method: 'GET',
+          path: URI.PPL_BUNDLE,
+        });
+        const body = result?.body ?? result;
+
+        // The result is the grammar bundle JSON
+        // Forward headers from OpenSearch if available
+        const responseHeaders: Record<string, string> = {
+          'content-type': 'application/json',
```
joshuali925: why hard code here, is SQL plugin response not including the right header?

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
```diff
@@ -183,25 +190,53 @@ export class CodeEditor extends React.Component<Props, {}> {
...
+    // Cancel any pending onLanguage listener from a previous render to prevent
+    // listener accumulation.
+    this._onLanguageDisposable?.dispose();
+
+    // Listen for the language's first encounter so providers are registered
+    // when a model for this language is created for the very first time.
+    // For SPA remounts (language already encountered), _editorDidMount handles
+    // registration directly since onLanguage won't fire again.
+    this._onLanguageDisposable = monaco.languages.onLanguage(languageId, () => {
+      this._registerProviders(languageId);
```
joshuali925: is it required to put onLanguage on render function? i was seeing duplicated suggestion candidates and sent a PR to avoid keep registering providers on each render, but not sure if it breaks other functionality, since CI was failing #11354

src/plugins/data/public/plugin.ts
```diff
@@ -314,6 +320,32 @@ export class DataPublicPlugin
...
+    // Subscribe to dataset changes to pre-fetch PPL grammar.
+    // The handler fires when the query language is PPL and the dataset changes.
+    // The initial fire with the current query covers the page-load case.
+    // the subscription covers subsequent dataset switches.
+    // Gated by query:enhancements:runtimePplGrammar setting (default: true).
+    const isRuntimePplGrammarEnabled = (() => {
+      try {
+        return uiSettings.get(UI_SETTINGS.QUERY_ENHANCEMENTS_RUNTIME_PPL_GRAMMAR) !== false;
+      } catch {
+        return true;
```
joshuali925: is try catch needed? I thought ui settings doesn't throw

src/plugins/data/public/query_snippet_suggestions/ppl/suggestions.ts
```diff
@@ -24,6 +24,9 @@ export const convertQueryToMonacoSuggestion = (queries: QuerySnippetItem[]): Que
     const snippets = extractSnippetsFromQuery(query.query.query as string);
     snippets.forEach((snippet) => {
       const trimmedSnippet = snippet.trim().toLowerCase();
+      if (!trimmedSnippet || !/[a-z0-9_]/i.test(trimmedSnippet)) {
```
joshuali925, line 27: why is this needed?

src/plugins/data/public/antlr/opensearch_ppl/ppl_grammar_cache.ts
```diff
@@ -0,0 +1,408 @@
...
+  shouldFetchFromBackend(version?: string): boolean {
+    if (!version) return false;
+
+    const match = version.trim().match(/^(\d+)\.(\d+)/);
+    if (!match) return false;
+    const major = Number(match[1]);
+    const minor = Number(match[2]);
+    if (!Number.isFinite(major) || !Number.isFinite(minor)) return false;
+
+    return major > 3 || (major === 3 && minor >= 6);
```
joshuali925: there is `semver.satisfies()`, something like this should work better https://github.com/opensearch-project/OpenSearch-Dashboards/blob/15cb9b02cf466cdb072ea97f0dfb78a544426fb8/src/plugins/dashboard/public/application/utils/migrate_app_state.ts#L88-L88

joshuali925:
### Code review

Found 1 issue:

1. Provider registration churn in `render()` -- `monaco.languages.onLanguage(languageId, callback)` is called inside `render()`, which runs on every React re-render (including every keystroke). When the language has already been encountered, `onLanguage` fires the callback **immediately and synchronously**, causing `_registerProviders()` to dispose and re-register all completion/hover/signature providers on every render cycle. The comment on line 236-237 states "onLanguage won't fire again" for SPA remounts, but this is incorrect: Monaco's `onLanguage` fires immediately for already-known languages. While the old providers are disposed before new ones are created (so there are no true duplicates), the constant dispose/re-register churn on every keystroke is a React anti-pattern (side effects in `render()`) and is wasteful. Consider moving provider registration out of `render()` entirely -- e.g., handle it only in `_editorDidMount` (for SPA remounts) and `componentDidUpdate` (for language changes).

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/d9254d295a3ad3633e8d219a8c703e96f90684d0/src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx#L227-L240

mengweieric:
> ### Code review
> Found 1 issue:
> 
> 1. Provider registration churn in `render()` -- `monaco.languages.onLanguage(languageId, callback)` is called inside `render()`, which runs on every React re-render (including every keystroke). When the language has already been encountered, `onLanguage` fires the callback **immediately and synchronously**, causing `_registerProviders()` to dispose and re-register all completion/hover/signature providers on every render cycle. The comment on line 236-237 states "onLanguage won't fire again" for SPA remounts, but this is incorrect: Monaco's `onLanguage` fires immediately for already-known languages. While the old providers are disposed before new ones are created (so there are no true duplicates), the constant dispose/re-register churn on every keystroke is a React anti-pattern (side effects in `render()`) and is wasteful. Consider moving provider registration out of `render()` entirely -- e.g., handle it only in `_editorDidMount` (for SPA remounts) and `componentDidUpdate` (for language changes).
> 
> https://github.com/opensearch-project/OpenSearch-Dashboards/blob/d9254d295a3ad3633e8d219a8c703e96f90684d0/src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx#L227-L240
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

 Verified against the Monaco source in this version: `onLanguage()` is a one-shot listener for a future `onDidRequestRichLanguageFeatures` event  
and does not replay for already-encountered languages. So the per-render churn described here does not happen.                                     
                                                                                                                                                   
On top of that, the current branch routes the callback through `_ensureProvidersRegistered()`, which no-ops when providers already exist for that  
language.                                                                                                                                          
                                                                                                                                                   
The actual duplicate-suggestion bug was that both `editorDidMount` and `onLanguage` could register providers on first encounter. That is fixed in  
`f024b39617` with the `_ensureProvidersRegistered()` guard and a regression test.

src/plugins/data/public/query_snippet_suggestions/ppl/suggestions.ts
mengweieric: Used this guard because `extractSnippetsFromQuery()` splits saved queries on `|`, so malformed or punctuation-only segments can show up as candidate snippets. This filter prevents empty / symbol-only segments from becoming Monaco suggestions.

src/plugins/data/public/plugin.ts
mengweieric: Removed.

src/plugins/query_enhancements/server/routes/index.ts
mengweieric: Removed.

src/plugins/data/public/antlr/opensearch_ppl/ppl_grammar_cache.ts
mengweieric: Done. Using semver now.

src/plugins/query_enhancements/server/routes/index.ts
mengweieric: Removed.

src/plugins/opensearch_dashboards_react/public/code_editor/code_editor.tsx
mengweieric: This was addressed in [f024b39](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11428/commits/f024b39617924a32a29085fa0d5daf10069adb04)

src/plugins/data/public/antlr/opensearch_ppl/code_completion.ts
```diff
@@ -192,195 +247,404 @@ export const getSimplifiedPPLSuggestions = async ({
...
     const isInBackQuote = suggestions.isInBackQuote || false;
+    const isRuntimeGrammar = runtimeSuggestions === suggestions;
+
+    // Runtime-specific context detection
+    if (isRuntimeGrammar) {
+      const isCommandPosition = isCommandPositionInCurrentSegment(queryTillCursor);
+
+      if (suggestions.suggestColumns && (isInBackQuote || !isInQuotes)) {
+        const initialFields = indexPattern.fields;
+        const cursorPosition = queryTillCursor.length;
```
RyanL1997, line 287: I saw there is diff for `cursorPosition` between runtime/compiled paths. For multi-line queries, will these produce different values, leading to incorrect field lookups on the runtime path?
- Runtime: `const cursorPosition = queryTillCursor.length; (byte offset from query start)`
- Compiled: `const cursorPosition = position?.column || selectionEnd; (column within current line)`

src/plugins/data/public/plugin.ts
```diff
@@ -367,6 +394,10 @@ export class DataPublicPlugin
   }
 
   public stop() {
+    this.pplGrammarWarmupSubscription?.unsubscribe();
```
RyanL1997, line 397: The warmup subscription and provider are cleaned up in `stop()`, but `pplGrammarCache` still holds its cached grammar and listeners. Could this cause stale data in HMR/test scenarios? Might be worth adding a `pplGrammarCache.reset()` here?

src/plugins/data/public/antlr/opensearch_ppl/runtime_ppl_grammar/opensearch_ppl_autocomplete.ts
```diff
@@ -0,0 +1,1090 @@
...
+  'SPAN',
+  'MATCH',
+  'MATCH_PHRASE',
+  'MATCH_BOOL_PREFIX',
+  'MATCH_PHRASE_PREFIX',
+  'SQUOTA_STRING',
+];
+
+const _runtimeC3BucketsByGrammarHash = new Map<string, unknown>();
+let _activeRuntimeC3GrammarHash: string | undefined;
```
RyanL1997, line 125: `_runtimeC3BucketsByGrammarHash` and `_operatorTokenCache` (line 642) grow with each new grammar hash and never evict. But I think this is fine cause it will be bounded by distinct OpenSearch versions per session correct?

PR #11430, opened 2026-03-03
Skip fetch_size when PPL query ends with explicit head command

ahkcs:
### Description

When a PPL query ends with an explicit `head` command (e.g. `source=t | head 600`), OSD unconditionally injects `fetch_size` (from `discover:sampleSize`, default 500) into the API request. Because the SQL plugin honors the smaller of `fetch_size` and `head`, a query like `| head 600` with `fetch_size=500` silently returns only 500 rows — the user has no way to retrieve the remaining 100.

This PR detects when `head` is the **last command** in the user's query and skips injecting `fetch_size`, letting the user's explicit `head` value be the sole row limiter. Queries without a trailing `head` continue to send `fetch_size` as before.

Key behaviors:
- `source=t | head 800` — skips `fetch_size` (head is the final command)
- `source=t | head 800 | where ts >= ...` — skips `fetch_size` (trailing `where` is an OSD-appended time filter)
- `source=t | head 800 | sort name ASC` — sends `fetch_size` as normal (head is not the final command)
- Join subqueries like `[source=t | head 3]` are masked and do not trigger detection
Screenshot 2026-03-03 at 2 52 58 PM
Screenshot 2026-03-03 at 2 53 04 PM

### Issues Resolved

Fixes https://github.com/opensearch-project/sql/issues/5191 (Bug 2)

## Testing the changes

1. Start OSD with default `discover:sampleSize = 500`
2. Run `source=<index> | head 600` — should return 600 rows (previously capped at 500)
3. Run `source=<index> | head 600 | sort name ASC` — should still send `fetch_size` and return 500 rows
4. Run `source=<index> | fields host` — should still send `fetch_size=500` and return 500 rows
5. Run a join query with `head` only in the subquery — should still send `fetch_size`
6. Run queries with field names like `header`, `headline` — no false positives

## Changelog

- skip

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/query_enhancements/common/utils.ts
+++ b/src/plugins/query_enhancements/common/utils.ts
@@ -141,6 +141,16 @@ export const buildQueryStatusConfig = (response: any) => {
 };
 
 /**
+ * Detects whether a PPL query ends with a `head` command in the main query,
+ * ignoring any trailing `| where ...` clauses (appended time-range filters)
+ * and any `head` commands inside subquery brackets [...].
+ */
+export const queryEndsWithHead = (queryString: string): boolean => {
+  const masked = queryString.replace(/\[.*?\]/g, (match) => '\0'.repeat(match.length));
+  return /\|\s*head\b(\s+\d+)?(\s+from\s+\d+)?\s*(\|\s*where\b.*)?\s*$/i.test(masked);
+};
+
+/**
  * Test if a PPL query is using search command
  * https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/cmd/search.md
  */
--- a/src/plugins/query_enhancements/common/utils.test.ts
+++ b/src/plugins/query_enhancements/common/utils.test.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { isPPLSearchQuery, throwFacetError, formatDate } from './utils';
+import { isPPLSearchQuery, queryEndsWithHead, throwFacetError, formatDate } from './utils';
 import { Query } from 'src/plugins/data/common';
 
 describe('throwFacetError', () => {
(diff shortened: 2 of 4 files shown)


src/plugins/query_enhancements/common/utils.ts
```diff
@@ -140,6 +140,16 @@ export const buildQueryStatusConfig = (response: any) => {
...
 };
 
+/**
+ * Detects whether a PPL query ends with a `head` command in the main query,
+ * ignoring any trailing `| where ...` clauses (appended time-range filters)
+ * and any `head` commands inside subquery brackets [...].
+ */
+export const queryEndsWithHead = (queryString: string): boolean => {
+  const masked = queryString.replace(/\[.*?\]/g, (match) => '\0'.repeat(match.length));
+  return /\|\s*head\b(\s+\d+)?\s*(\|\s*where\b.*)?\s*$/i.test(masked);
```
joshuali925: i think this doesn't detect `head 1 from 1`?

i'm ok with regex if it works, but would it be more reliable to detect on the SQL backend, if the expectation is that `head` in query takes priority over `fetch_size` param?

Also could you explain a bit why user gets 600 without `sort`, but 500 with `sort`? the behavior appears a bit inconsistent
> Run `source=<index> | head 600 | sort name ASC` — should still send fetch_size and return 500 rows

ahkcs: This was actually debated in the companion sql PR ([opensearch-project/sql#5194](https://github.com/opensearch-project/sql/pull/5194#discussion_r2880021092)). @penghuo's feedback there was that the fix belongs in the client:

fetch_size is an explicit, optional parameter the caller chooses to send. If the caller sends fetch_size=5, they're requesting at most 5 rows — that's the API contract. Adding server-side logic to silently ignore that parameter based on query content breaks the principle of least surprise at the API layer.

That feedback directly motivated opening this OSD-side PR. The sql PR remains open as a potential alternative, but the reviewer consensus there pointed to fixing it where the default injection happens.

ahkcs: >Also could you explain a bit why user gets 600 without sort, but 500 with sort? the behavior appears a bit inconsistent

You're right that this looks inconsistent, and it's a known limitation of this approach. A more complete fix would detect head anywhere in the main pipeline (not just at the end), but that's more aggressive and would require more careful scoping to avoid breaking cases where head is genuinely intermediate.

After discussing, we think that the current approach can cover more than 95% of the use cases and putting head command at the end should be the correct user behavior for most cases, for example this query:

```
source=<index> | head 600 | sort name ASC
```
The correct use case for users would be to use `sort` command first and then use `head` command in the end to get the first 600 names in ascending order

ahkcs: > i think this doesn't detect head 1 from 1?

Updated PR, now the regex handles head N from M syntax as well
Screenshot 2026-03-09 at 2 53 47 PM

joshuali925: got it, fetch_size taking priority is fine, as long as the behavior is documented

PR #11437, opened 2026-03-04
fix(Discover): breadcrumb from saved search navigates to root and clears state

JaswanthKumarBL:
### Description
When viewing a saved search in Discover, clicking the Discover breadcrumb did not return to Discover root with a clean state: the saved search stayed in Redux, query and filters were not cleared, and the previous index pattern was not restored.
This change:
Redux: Adds a clearSavedSearch reducer so the saved search can be cleared when navigating back to Discover root.
Breadcrumbs: Updates the Discover breadcrumb onClick to clear app filters, clear the query string, restore the last index pattern from sessionStorage (when available), dispatch clearSavedSearch, and replace the URL to /discover#/ when store and scopedHistory are available (e.g. from the Discover canvas). When they are not (e.g. embedded doc views), it falls back to full app navigation.
Services: Passes services into getRootBreadcrumbs(services) from top_nav.tsx and get_top_nav_links.tsx so the breadcrumb handler has access to store and scopedHistory.
Persistence: In use_index_pattern, persists the current index pattern id to sessionStorage when on Discover root (no saved search) so the breadcrumb can restore it when returning from a saved search.
Constants: Introduces a shared LAST_INDEX_PATTERN_KEY in application/helpers/constants.ts used by both the breadcrumb and use_index_pattern.
Loading: After dispatching updateIndexPattern(newId), the hook fetches the index pattern by id and sets it so the UI reflects the chosen pattern immediately when returning from a saved search.

### Issues Resolved

No existing issue. This fixes the behavior where clicking the Discover breadcrumb from a saved search did not navigate back to Discover root or clear query/filters/saved search state.

## Screenshot
**Issue (before fix):** When viewing a saved search, clicking the Discover breadcrumb did not return to Discover root with a clean state — saved search, query, and filters persisted.
image

**Fix (after):** Clicking the Discover breadcrumb now navigates to Discover root, clears query and filters, clears the saved search state, and restores the previous index pattern.

Attached: Screen Recording 2026-03-04 at 3.27.04 PM — demonstrates the fix (navigating back from saved search to Discover via breadcrumb).

## Testing the changes
1. **Prerequisites:** OpenSearch and OpenSearch Dashboards running; at least one index pattern and one saved search in Discover.

2. **Reproduce (without fix):** Go to Discover → select index pattern → add query/filter → save as saved search → open that saved search. Click the **Discover** breadcrumb. Previously: URL/state did not reset; query/filters or saved search could persist.

3. **Verify fix:** With the fix, from the same saved search view, click the **Discover** breadcrumb. Confirm:
   - URL is `/discover#/`
   - View is Discover root (no saved search)
   - Query and filters are cleared
   - Index pattern is the one used on Discover root before opening the saved search

4. **Optional:** Run `yarn test:jest` and `yarn test:jest_integration` from repo root to ensure no regressions.

## Changelog
- fix: Discover breadcrumb from saved search now navigates to root and clears query, filters, and saved search state

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/discover/public/application/helpers/breadcrumbs.ts
```diff
@@ -31,15 +31,48 @@
...
+  const isOnDiscoverRoot = store && scopedHistory && !store.getState().discover?.savedSearch;
+
+  const discoverBreadcrumb: EuiBreadcrumb = {
+    text: i18n.translate('discover.rootBreadcrumb', {
+      defaultMessage: 'Discover',
+    }),
+    ...(isOnDiscoverRoot
+      ? {}
+      : {
+          onClick: () => {
```
ruchidh: Nit: can we add  analytics tracker too, for ux insights 

eg. 
    core.analytics?.reportEvent('discover_breadcrumb_navigation', {
      from_saved_search: !!store.getState().discover?.savedSearch,
      restored_index_pattern: !!savedId
    });

src/plugins/discover/public/application/view_components/utils/use_index_pattern.ts
```diff
@@ -78,7 +81,10 @@ export const useIndexPattern = (services: DiscoverViewServices) => {
           );
           if (isMounted && newId) {
             store!.dispatch(updateIndexPattern(newId));
-            handleIndexPattern();
+            const ip = await fetchIndexPatternDetails(newId);
+            if (isMounted && ip) {
```
ruchidh: Nit: i avoid braces when not multi line, just to save lines

if (isMounted && ip)   setIndexPattern(ip);

src/plugins/discover/public/application/helpers/breadcrumbs.ts
```diff
@@ -31,15 +31,48 @@
...
+            filterManager.setAppFilters([]);
+            data.query.queryString.setQuery({ query: '' }, false, true);
```
ruchidh, line 64:
```
  filterManager.setAppFilters([]);
  data.query.queryString.setQuery({ query: '' }, false, true);
  if (savedId) store.dispatch(updateIndexPattern(savedId));
  store.dispatch(clearSavedSearch());
  scopedHistory.replace('/discover#/');
```
just for discussion, , Is this the optimal order for state cleanup? Should clearSavedSearch() happen before or after index pattern restoration?

src/plugins/discover/public/application/helpers/breadcrumbs.ts
```diff
@@ -31,15 +31,48 @@
...
+            if (store && scopedHistory) {
+              const savedId = sessionStorage.getItem(LAST_INDEX_PATTERN_KEY);
+              if (savedId) store.dispatch(updateIndexPattern(savedId));
```
ruchidh: Should we validate the restored index pattern ID before dispatching updateIndexPattern()?

also, What happens if user clicks breadcrumb while a search is in progress?

src/plugins/discover/public/application/helpers/breadcrumbs.ts
```diff
@@ -31,15 +31,48 @@
...
+              store.dispatch(clearSavedSearch());
+              scopedHistory.replace('/discover#/');
```
ruchidh, line 85: If we're on an invalid saved search, should the breadcrumb behavior be different? 
i am assuming a scenario where the url has invalid param or suppose user updated it with some other field, in that scenario how should the breadcrumb behave ?

src/plugins/discover/public/application/helpers/breadcrumbs.ts
JaswanthKumarBL: We kept the order as: clear filters/query → restore index pattern (if valid) → clearSavedSearch() → scopedHistory.replace(). We set the index pattern before calling clearSavedSearch() so that when the view re-renders after the URL replace, the state is already in the right shape (restored pattern + no saved search), and we avoid an intermediate state where the saved search is cleared but the index pattern is not yet updated. I don’t think this order will cause any issues, and we’re happy to adjust if you see a reason to do it differently.

src/plugins/discover/public/application/helpers/breadcrumbs.ts
JaswanthKumarBL: We don’t treat invalid differently; when they click Discover we always clear and go to the main Discover page, so they always land in a good, clean state. There are no edge cases in this breadcrumb logic for invalid URLs or params.

src/plugins/discover/public/application/helpers/breadcrumbs.ts
JaswanthKumarBL: 1)Should we validate the restored index pattern ID before dispatching updateIndexPattern()?
**agreed and added validation**. We use sessionStorage as intended (per-tab, to retain the user’s last index pattern on Discover root). The check before dispatch is for the rare edge case where that pattern was deleted in the meantime (e.g. in another tab). We only dispatch the restored id if it still exists (data.indexPatterns.get(savedId)); otherwise we skip restore and the user still lands on Discover root with the previous id (e.g. the saved search’s pattern).

2)What happens if user clicks breadcrumb while a search is in progress?

The Discover root breadcrumb is only clickable when we’re not on the root, i.e. when we’re viewing a saved search. That’s controlled by:

const isOnDiscoverRoot = store && scopedHistory && !store.getState().discover?.savedSearch;

So the “Discover” crumb gets an onClick only when there’s a saved search (and the saved-search breadcrumb appears next to it). If the user clicks it while the saved search’s request is still in progress, we clear state and navigate to /discover#/. The view then updates to the root context, and any in-flight search from the previous saved search is effectively superseded (and the existing AbortController usage in the search flow handles new requests). We didn’t add special handling for “search in progress” because leaving the saved-search context is the intended behavior and the current flow already handles it.

src/plugins/discover/public/application/view_components/utils/use_index_pattern.ts
JaswanthKumarBL: updated

src/plugins/discover/public/application/helpers/breadcrumbs.ts
JaswanthKumarBL: Added the analytics call as suggested. Core doesn’t expose analytics.reportEvent yet, so it’s wrapped in optional chaining and is a no-op for now. Left a comment in the code so it’s clear. Happy to remove it if you’d rather not have it until the API exists.

ananzh (commented):
@JaswanthKumarBL Thanks a lot for the contribution. I don't think we have to implement a complete clean out. The "New" button (discoverNewButton) here in discover https://github.com/opensearch-project/OpenSearch-Dashboards/blob/f2f3098e08dafab258301d90ef0929f724926ac0/src/plugins/discover/public/application/components/top_nav/get_top_nav_links.tsx#L56-L58 uses `navigateToApp('discover', { path: '#/' })` to reset state - and it works correctly. The breadcrumb should simply do the same thing. No explicit state clearing needed since the URL-driven state initialization handles everything. 

When URL is `#/` (clean, no _a parameter):
  - serializedState is null
  - Falls back to getPreloadedState() which returns fresh initial state

Then the query and filter state is synced FROM the URL via osdUrlStateStorage. A clean URL means:
  - No _a app state → no filters, no custom query
  - No _g global state → default time range

So I think updating https://github.com/opensearch-project/OpenSearch-Dashboards/blob/f2f3098e08dafab258301d90ef0929f724926ac0/src/plugins/discover/public/application/helpers/breadcrumbs.ts#L42 to 
```onClick: () => core.application.navigateToApp('discover', { path: '#/' }),```
should be enough and we could avoid manual clearing + navigation + cleaning session storage.

ruchidh (approved):

PR #9787, opened 2025-05-12
Fix the bug of open field link in a new tab

abbyhu2000:
### Description
In table vis, a url field with the setting of open in a new tab did not open itself in a new tab. More detailed can be found in the below issue.

### Issues Resolved

#3500 

## Changelog
- fix: Open field link in a new tab

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

PR #11458, opened 2026-03-06
Fix open link in new tab from table field

### LINKED ISSUE #3500: [BUG] open in a new tab broken in 2.5.0

sejli:
### Description

Building off #9787

Since the `table_vis_cell.tsx` component was rewritten to `table_vis_dynamic_table.tsx`, this PR updates it such that the `dompurify` hooks are in the correct component. For 2.19, which does not have the rewritten `table_vis_dynamic_table.tsx`, #11456 was opened to maintain the same behavior for opening links in a new tab.

### Issues Resolved

#3500

## Changelog

- fix: Fix open link in new tab from table field

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/vis_type_table/public/components/table_vis_cell.tsx
+++ b/src/plugins/vis_type_table/public/components/table_vis_cell.tsx
@@ -23,17 +23,6 @@ export const getTableVisCellValue = (
   const rawContent = row[columnId];
   const colIndex = columns.findIndex((col) => col.id === columnId);
   const htmlContent = columns[colIndex].formatter.convert(rawContent, 'html');
-  dompurify.addHook('uponSanitizeAttribute', (node, event) => {
-    if (event.attrName === 'target') {
-      event.forceKeepAttr = true;
-    }
-  });
-
-  dompurify.addHook('afterSanitizeElements', (node) => {
-    if (node instanceof Element && node.tagName?.toUpperCase() === 'A') {
-      node.setAttribute('rel', 'noopener noreferrer');
-    }
-  });
   const formattedContent = (
     /*
      * Justification for dangerouslySetInnerHTML:
--- a/src/plugins/vis_type_table/public/components/table_vis_cell.test.tsx
+++ b/src/plugins/vis_type_table/public/components/table_vis_cell.test.tsx
@@ -62,23 +62,4 @@ describe('getTableVisCellValue', () => {
     const { container } = render(<TableCell rowIndex={0} columnId="nonexistent" />);
     expect(container.firstChild).toBeNull();
   });
-
-  test('should sanitize HTML content using dompurify', () => {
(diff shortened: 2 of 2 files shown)


src/plugins/vis_type_table/public/components/table_vis_cell.tsx
```diff
@@ -23,6 +23,17 @@ export const getTableVisCellValue = (
...
+    if (event.attrName === 'target') {
+      event.forceKeepAttr = true;
+    }
+  });
+
+  dompurify.addHook('afterSanitizeElements', (node) => {
+    if (node instanceof Element && node.tagName?.toUpperCase() === 'A') {
+      node.setAttribute('rel', 'noopener noreferrer');
+    }
+  });
```
angle943: is this safe? The only thing this is checking for is that the node is an anchor tag. wouldn't this run if `afterSanitizeElements` ran for a completely different, unrelated anchor tag, say like an internal nav link (now why would we run dompurify on that who knows)

is it possible to be more specific then just targetting all anchor tags?

Maosaic: AI suggesting to move these to module scope like `able_vis_dynamic_table.tsx`, any reason doing it differently here?

```
dompurify.addHook() calls are placed inside the getTableVisCellValue render function body rather than at module scope. This causes new hooks to be registered on every single cell render, leading to unbounded hook accumulation. Over time this produces unpredictable sanitization behavior as stacked duplicate hooks all execute, and could degrade or circumvent the intended sanitization logic. Unlike the parallel change in table_vis_dynamic_table.tsx (which correctly places the same hooks at module level), this placement is anomalous and warrants investigation.
```

sejli: Just removed this, it's not needed since `table_vis_cell.tsx` is no longer used post rewrite.

sejli: Reverted the original changes since that component is no longer used.

PR #11478, opened 2026-03-09
[Explore] Move raw hits into a module-level cache and keep only lightweight metadata in Redux for reactivity

Maosaic:
## Problem

The Redux `results` slice stored full `ISearchResult` objects — raw OpenSearch hits containing up to 10,000 documents, each potentially with hundreds of fields.

On every `setResults` dispatch, Immer's `produce()` **recursively `Object.freeze()`s the entire new state subtree**, traversing all properties synchronously on the main thread. For large result sets this means potentially millions of property accesses per query execution, causing noticeable UI jank and blocking rendering.

Two of the three dev-mode walker costs were already mitigated: `ImmutableStateInvariantMiddleware` and `SerializableStateInvariantMiddleware` were excluded via `ignoredPaths: ['results']`. The Immer freeze pass inside the reducer itself was the remaining bottleneck and could not be opted out of without removing the data from Redux entirely.

The underlying architectural mismatch: results are always **fully replaced** and **never partially mutated**, so Redux's mutation tracking adds no safety value for them. The data belongs in a cache, not a state store.

## Solution

Move raw hits out of Redux into a **module-level `Map` cache**, keyed by `cacheKey`. Redux stores only lightweight **metadata** per key, used purely as a reactivity signal. A new **cache middleware** in `store.ts` intercepts `setResults` actions and writes the full `ISearchResult` to the cache **before** calling `next(action)` — ensuring that when React re-renders from the Redux state change, `resultsCache.get(key)` already returns fresh data.

### New Redux metadata shape

```ts
interface ResultMetadata {
  total: number;
  elapsedMs: number;
  fieldSchema?: Array<Partial<IFieldType>>;
  instantFieldSchema?: Array<Partial<IFieldType>>;  // Prometheus results
  hasResults: boolean;
}

type ResultsState = Record<string, ResultMetadata>;
```

### Cache + middleware

```ts
// results_slice.ts
export const resultsCache = new Map<string, ISearchResult>();
export const clearResultsCache = () => resultsCache.clear();
```

```ts
// store.ts — createResultsCacheMiddleware (registered first, before all other middleware)
if (setResults.match(action)) {
  resultsCache.set(action.payload.cacheKey, action.payload.results);   // before next(action)
} else if (clearResultsByKey.match(action)) {
  resultsCache.delete(action.payload);
} else if (clearResults.match(action) || resetState || hydrateState) {
  clearResultsCache();
}
return next(action);
```

**Critical ordering invariant:** cache writes happen _before_ `next(action)`. `next(action)` runs the reducer and notifies React subscribers. If a component's `useSelector` fires during that notification and calls `resultsCache.get()`, the cache must already contain fresh data. Moving the writes after `next(action)` would cause the first re-render to read stale results.

### Hook pattern (unchanged public API)

```ts
// Before
const results = useSelector(state => state.results[cacheKey]);

// After — Redux metadata drives reactivity, cache provides hits
const metadata = useSelector(state => cacheKey ? state.results[cacheKey] : null);
const results = metadata ? resultsCache.get(cacheKey) ?? null : null;
```

### Thunk pattern

```ts
// Before
const results = state.results[cacheKey];

// After — thunks read directly from cache, no hooks
const results = resultsCache.get(cacheKey);
```

### Core refactor

| File | Change |
|------|--------|
| `slices/results/results_slice.ts` | Replace `ResultsState` (full hits) with `ResultMetadata`; add `resultsCache` Map and `clearResultsCache()`; update `setResults` reducer to store extracted metadata only; remove `setResultsState` action |
| `state_management/store.ts` | Add `createResultsCacheMiddleware` (registered first); remove now-unnecessary `immutableCheck`/`serializableCheck` `ignoredPaths` for `results` |
| `actions/reset_explore_state/reset_explore_state.ts` | Replace `dispatch(setResultsState(state.results))` with `dispatch(clearResults())` — the old call was always equivalent to a clear since results are not URL-persisted |

### Hooks

| File | Change |
|------|--------|
| `hooks/use_tab_results.ts` | Subscribe to `state.results[cacheKey]` metadata for reactivity; read hits from `resultsCache.get(cacheKey)` |
| `hooks/use_histogram_results.ts` | Same pattern |

### Thunks

| File | Change |
|------|--------|
| `actions/export_actions.ts` | Replace two `state.results[key]` reads with `resultsCache.get(key)` |
| `actions/detect_optimal_tab/detect_optimal_tab.ts` | Replace `state.results[key]` read with `resultsCache.get(key)` |

### Components updated to use hooks

`discover_chart_container.tsx`, `explore_data_table.tsx`, `explore_metrics_data_table.tsx`, `explore_metrics_raw_table.tsx`, `fields_selector_panel.tsx`, `results_summary.tsx`, `use_displayed_columns.ts`, patterns table utilities — all updated to go through `useTabResults` / `useHistogramResults` rather than accessing `state.results[key]` directly.

### Tests (26 files)

- Seed `resultsCache` directly instead of Redux state for tests that need hits data
- Call `clearResultsCache()` in `afterEach` to prevent cross-test bleed
- Assert on `ResultMetadata` shape rather than full `ISearchResult` where appropriate

## Cache lifecycle

The cache is kept in sync with Redux metadata by the middleware — cache entries are removed whenever their corresponding Redux metadata is removed:

| Event | Cache effect |
|-------|-------------|
| `setResults` | `resultsCache.set(cacheKey, fullResult)` |
| `clearResultsByKey` | `resultsCache.delete(cacheKey)` |
| `clearResults` | `clearResultsCache()` |
| `app/resetState` | `clearResultsCache()` |
| `app/hydrateState` | `clearResultsCache()` |

## What did NOT change

- Public API of `useTabResults` and `useHistogramResults` — call sites need no updates
- The `setResults` action payload shape — `{ cacheKey, results: ISearchResult }` is unchanged; the middleware intercepts it before the reducer strips it down to metadata
- `query_actions.ts` and `trace_query_actions.ts` — no changes required

## Verification

```bash
yarn test:jest results_slice
yarn test:jest use_tab_results
yarn test:jest use_histogram_results
yarn test:jest export_actions
yarn test:jest detect_optimal_tab
yarn test:jest explore
yarn typecheck
```

Manual checks:
- Run a query in Explore — results render correctly
- Histogram displays
- CSV export works
- Tab switching preserves results
- Browser back/forward clears results (cache cleared by `hydrateState`/`resetState`)
## Changelog

- chore: Move raw hits into a module-level cache and keep only lightweight metadata in Redux for reactivity

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/11478.yml
+++ b/changelogs/fragments/11478.yml
@@ -1,0 +1,2 @@
+chore:
+- Move raw hits into a module-level cache and keep only lightweight metadata in Redux for reactivity ([#11478](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11478))


src/plugins/explore/public/application/utils/state_management/store.ts
```diff
@@ -66,23 +101,17 @@ export const configurePreloadedStore = (
     reducer: rootReducer,
     preloadedState,
     middleware: (getDefaultMiddleware) => {
-      // Exclude results from dev-mode checks: the results slice holds raw OpenSearch hits which
-      // can be very large. ImmutableStateInvariantMiddleware walks the entire state tree on every
-      // dispatch — including stored results — making every user interaction slow when results are
-      // loaded. Results are always fully replaced, never partially mutated, so these checks add
-      // no safety value there.
-      const middlewareOptions = {
```
ananzh, line 74: nit: why remove these two? I think we would not need them due to tiny object, but also no harm to keep them. So this is a deliberate choice, not an accidental omission right?

Maosaic: These settings were added to mitigate the large query result performance issue in dev environment before this change, since we are not storing the query result in redux any more, they are nolonger needed.

PR #11479, opened 2026-03-10
[Infra] handle logs and warnings in observeCompiler, and allow ContextModules

angle943:
### Description

- Running `yarn build` in a plugin would sometimes lead to a non-helpful failure:

```
jungkm@80a9972c0081 SomePlugin % yarn build        
yarn run v1.22.19
$ yarn plugin-helpers build
$ node ../../scripts/plugin_helpers build
 info deleting the build and target directories
 info running @osd/optimizer
 │ info initialized, 0 bundles cached
 │ info starting worker [1 bundle]
 │ERROR webpack compile errors
     │ERROR [storage] build
         │ERROR Optimizations failure.
         │      42 assets
         │          2475 modules
         │          Rspack 1.6.4 compiled with 1 warning in 30.86 s
 │ERROR webpack issue
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

- This is very unhelpful and wrong, since Rspack compiled with just 1 warning, so the end result should've been a success.
- So i added a change to 1) log warnings & errors for developers debugging experience, and 2) do not count warnings as failures

- But when i try to run `yarn build` again, it resulted in this (separate) error:

```
ERROR Error: Unexpected module type: ContextModule
```

- ContextModules are valid modules, so we should allow for these

## Changelog
- infra: add logging to compiler step, allow warnings, and allow contextModules

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/11479.yml
+++ b/changelogs/fragments/11479.yml
@@ -1,0 +1,2 @@
+infra:
+- Add logging to compiler step, allow warnings, and allow contextModules ([#11479](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11479))


LDrago27 (approved):

packages/osd-optimizer/src/worker/webpack_helpers.ts
```diff
@@ -31,13 +31,29 @@
...
-    return true;
+  const { warnings, errors } = stats.toJson({ all: false, warnings: true, errors: true });
+
+  // Log warnings if present
+  if (warnings && warnings.length > 0) {
+    // eslint-disable-next-line no-console
+    console.warn(`[Webpack] ${warnings.length} warning(s) found:`);
+    warnings.forEach((warning, index) => {
+      // eslint-disable-next-line no-console
+      console.warn(`  Warning ${index + 1}: ${warning.message || warning}`);
```
ananzh, line 42: I think before Rspack, we used to filter warnings `'(export .* was not found in)'`, should we following the previous way to expand the filtered warnings like
```
export const STATS_WARNINGS_FILTER = new RegExp(
    [
      '(export .* was not found in)',
      '(Critical dependency: require function is used in a way in which dependencies cannot be statically extracted)',
    ].join('|')
  );
```

ananzh: So we know exactly what we're allowing (Monaco's warning), which is a common warning according to

https://www.google.com/search?q=monaco-editor+Critical+dependency%3A+require+function+is+used+in+a+way&rlz=1C5GCEM_enUS1175US1175&oq=monaco-editor+Critical+dependency%3A+require+function+is+used+in+a+way&gs_lcrp=EgZjaHJvbWUyBggAEEUYOdIBCDQ3MTJqMGo3qAIAsAIA&sourceid=chrome&ie=UTF-8

but we still break for other warnings. I think my biggest concern is that if we allow passing all warnings then we might miss some actual issues. Though we log them, I feel no one will actually research/understand/fix if we allow passing. I would prefer a more conservative approach. What do you think?

angle943: i think we should treat warnings as warnings. We do not have such a strict process for installation of packages, running unit/cypress tests, TS warnings, and so on.

If a warning is actually production-breaking and our series of tests do not capture it, then we will address the fix and up the coverage.

I'm against the filter because as soon as we add the filter, it can be argued why not just provide the fix? The link you shared shows that we might be able to fix it by monaco-editor-webpack-plugin. And if we are not so eager to provide the fixes now, what are we going to do if we (or plugin teams) in the future encounters other warnings that results in failures? We will most likely do the same thing, add to our whitelist filter, and not action upon it.

Its a mechanism that doesn't produce better quality, its just a mechanism that produces more churn

Issue #11487, opened 2026-03-10
chore: Improve CI Experience for Forks


davidshq:
**Is your feature request related to a problem? Please describe.**

To get OpenSearch Dashboards to run through GitHub Actions on a fork one has to edit the `build_and_test_workflow.yml` and add ones' branch to it. After successfully running CI one then has to revert the change to the workflow before making a PR to OSD.

This is a fairly minor nuisance but is amplified in a few ways:
1. The tests now run fairly frequently and its a pretty hefty set of tests.
2. If one needs to make changes to the PR for any reason one has to re-edit the `build_and_test_workflow.yml` file and then revert it again (or one doesn't and crosses fingers that it passes CI).
3. If one isn't using codecov then you need to also make the codecov conditional or the tests will fail.

This is especially painful for new contributors because the workflows don't kick off automatically on a new PR. They require manual intervention by someone with permissions to start the workflows. This can create a long delay between when a fix is submitted and when CI actually runs on that fix.

**Describe the solution you'd like**

1. If `workflow_dispatch:` is added to the yml file then one doesn't need to edit the file to run on one's branch.
2. If a conditional is added for codecov then the codecov won't run if it isn't configured and the workflow will not fail.

PR #11488, opened 2026-03-10
chore: Enable manual workflow runs and fix codecov for forks

davidshq:
## Summary
Enables manual workflow runs for the build and test workflow and fixes codecov to only run when the token is configured and the project is opensearch-project/OpenSearch-Dashboards (so forks don't fail on missing codecov token).

- This PR addresses #11487 

## Changes
- [chore] Make manual runs possible for the build and test workflow
- [chore] Don't run codecov if token isn't configured or project isn't opensearch-project/OpenSearch-Dashboards

## Changelog
- skip


davidshq:
I believe the prefix issue is fixed? I updated to skip and I see the label was updated to Skip-Changelog, so this should be ready to run through GitHub workflows.

PR #11489, opened 2026-03-10
Add delimiter for comma separated terms

abbyhu2000:
### Description

For copy paste comma separated terms, we should be able to add each terms to the inbox. Previously this feature is broken.

Before:

https://github.com/user-attachments/assets/5a2ce845-4df1-4a51-8e63-6f7c619c1b41

After:
https://github.com/user-attachments/assets/c4ea3e73-aebf-47d1-bf23-bdb603195c3d

Note: we need to add manual onPaste wrapper here since the delimiter = ',' prop does not work in this situation. In the component we are calling onChange directly which doesn't use the previous state, so each call overwrites the last. That's why in the before video, we only see female being pasted but not male. We could fix the onChange prop to use the previous state, but a simpler approach would be adding a onPaste wrapper to handle the delimiter.
## Changelog
- fix: Add delimiter for comma separated terms

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/ui/filter_bar/filter_editor/phrases_values_input.tsx
+++ b/src/plugins/data/public/ui/filter_bar/filter_editor/phrases_values_input.tsx
@@ -56,37 +56,23 @@ class PhrasesValuesInputUI extends PhraseSuggestorUI<Props> {
           defaultMessage: 'Values',
         })}
       >
-        <div
-          onPaste={(e: React.ClipboardEvent) => {
-            const pasted = e.clipboardData.getData('text');
-            if (pasted.includes(',')) {
-              e.preventDefault();
-              const newValues = pasted
-                .split(',')
-                .map((v) => v.trim())
-                .filter((v) => v.length > 0);
-              onChange([...(values || []), ...newValues]);
-            }
-          }}
-        >
-          <StringComboBox
-            placeholder={intl.formatMessage({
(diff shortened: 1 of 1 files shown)


src/plugins/data/public/ui/filter_bar/filter_editor/phrases_values_input.tsx
```diff
@@ -55,22 +56,36 @@ class PhrasesValuesInputUI extends PhraseSuggestorUI<Props> {
...
-          getLabel={(option) => option}
-          selectedOptions={values || []}
-          onSearchChange={this.onSearchChange}
-          onCreateOption={(option: string) => onChange([...(values || []), option])}
-          onChange={onChange}
-          isClearable={false}
-          data-test-subj="filterParamsComboBox phrasesParamsComboxBox"
-          delimiter=","
-        />
+        <div
```
LDrago27: Can we move the onPaste handler into the ComboBox itself. Currently the onPaste handler seems tacked on a div. We can enhance the ComboBox to handle this and pass this behaviour to customize it.

abbyhu2000: Thanks and good idea, moved into combo box

LDrago27 (approved):

PR #11492, opened 2026-03-10
Dataset Selection and Query Clearing Issues

TackAdam:
### Description
  This PR fixes several critical issues related to dataset selection and query persistence when navigating between different Explore flavors (Logs, Traces,Metrics), particularly after loading saved visualizations and refreshing the page.
  
  

https://github.com/user-attachments/assets/e623a8ed-a926-45d3-b887-b43d2222bb99

  ## Issues Fixed

  ### 1. **Query Language Support Crash**
  **Problem**: Application crashed when loading saved visualizations with query languages other than PPL
  **Error**: `defaultPrepareQueryString encountered unhandled language: kuery`

  **Fix**: Added support for additional query languages: `kuery`, `DQL`, `SQL`, `lucene`, `PROMQL`

  ### 2. **Query Persistence When Switching Flavors**
  **Problem**: When navigating from Traces to Logs (or vice versa), the query from the previous flavor persisted, causing syntax errors and incorrect results

  **Fix**:
  - Added dataset compatibility detection in `redux_persistence.ts`
  - Clears incompatible queries from URL state when switching flavors
  - Explicitly clears editor text when query state is empty

  ### 3. **Wrong Dataset Selected for Logs**
  **Problem**: Logs page selected datasets without explicit `signalType: "logs"` instead of prioritizing datasets with the correct signal type

  **Fix**: Enhanced `fetchFirstAvailableDataset` to prefer datasets with explicit signal type matches over compatible datasets without signal type

  ### 4. **Default Dataset Not Prioritized**
  **Problem**: Default datasets (marked with "Default" badge in UI) were not being selected even when available and compatible

  **Fix**: Changed dataset resolution priority order:
  1. Preferred dataset (explicitly passed)
  2. **Default dataset** (from DatasetService, if compatible) ← NEW
  3. Cached dataset (last used, if compatible)
  4. Fetch first available dataset

  ### 5. **Saved Visualization Query Persists After Refresh and Flavor Switch**
  **Problem**: Critical UX issue - After loading a saved Traces visualization, refreshing the page, then switching to Logs tab, the incompatible Traces query
  remained visible in the editor with wrong dataset

  **Fix**:
  - Added flavor mismatch detection in `use_page_initialization.ts`
  - When saved visualization's flavor doesn't match current page flavor, redirects to base page
  - Clears URL state parameters (`_q`, `_a`, `_v`) and query from global state
  - Removes query field from `_g` parameter before redirect

  ## Files Changed

  ### Core Logic Changes

  #### 1. `src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts`
  - Added cases for `kuery`, `DQL`, `SQL`, `lucene`, `PROMQL` in `defaultPrepareQueryString()`
  - These languages now return the query string as-is instead of throwing an error

  #### 2. `src/plugins/agent_traces/public/application/utils/state_management/actions/query_actions.ts`
  - Same language support added for agent_traces plugin

  #### 3. `src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts`
  **Major changes:**
  - Added dataset fetching via `dataViews.get()` to populate `signalType` before compatibility checks
  - Enhanced `fetchFirstAvailableDataset()` to:
    - Check if dataset matches default dataset ID and return immediately
    - Prefer datasets with explicit signal type match
    - For Logs: prefer `signalType: "logs"` over datasets without signal type
  - Modified `resolveDataset()` to:
    - Fetch full dataset from dataViews to get signal type
    - Check default dataset first before cached dataset
    - Only accept perfect matches for default dataset on Logs flavor
  - Added incompatibility detection in `loadReduxState()`:
    - Fetches dataset signal type from dataViews if not present
    - Detects when URL dataset is incompatible with current flavor
    - Clears query when dataset changes or is replaced

  #### 4. `src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts`
  - Added flavor mismatch detection when loading saved visualizations
  - If saved explore flavor doesn't match current page flavor:
    - Clears URL state storage (`_q`, `_a`, `_v`)
    - Removes query from global state (`_g`)
    - Redirects to base page of current flavor with clean URL
  - Uses `window.location.href` for redirect since `application` service may not be available

  #### 5. `src/plugins/explore/public/application/pages/logs/logs_page.tsx`
  - Added effect to explicitly clear editor text when query state is empty
  - Ensures editor reflects the empty query after flavor switch redirects

  ### Test Updates

  #### 6. `src/plugins/explore/public/application/utils/state_management/actions/query_actions.test.ts`
  - Added test cases for new query languages: `kuery`, `DQL`, `SQL`, `lucene`

  #### 7. `src/plugins/agent_traces/public/application/utils/state_management/actions/query_actions.test.ts`
  - Same test updates for agent_traces plugin

  #### 8. `src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.test.ts`
  - Added `getDefault()` method to all mock DatasetService objects
  - Updated `setQuery` call expectation to include `forceUpdate` parameter

  #### 9. `src/plugins/explore/public/application/pages/logs/logs_page.test.tsx`
  - Added mock for `useSetEditorText` hook to prevent context errors in tests

  ## Testing

  ### Manual Testing Scenarios

  ✅ **Scenario 1: Normal Navigation**
  - Navigate from Traces to Logs without saved visualization
  - **Result**: Clean Logs page with empty query and correct dataset

  ✅ **Scenario 2: Saved Visualization on Same Flavor**
  - Open saved Traces visualization
  - **Result**: Visualization loads correctly with saved query and dataset

  ✅ **Scenario 3: Saved Visualization After Refresh** (Main fix)
  - Open saved Traces visualization
  - Refresh page
  - Switch to Logs tab
  - **Result**: Lands on clean Logs page with empty query and logs dataset selected

  ✅ **Scenario 4: Default Dataset Selection**
  - Navigate to page with default dataset configured
  - **Result**: Default dataset is selected and displayed

  ✅ **Scenario 5: Multiple Query Languages**
  - Load saved visualizations with kuery, DQL, SQL, lucene, PROMQL
  - **Result**: No crashes, queries load correctly

  ### Automated Tests

  - ✅ **redux_persistence.test.ts**: 32/32 tests passing
  - ✅ **logs_page.test.tsx**: 5/5 tests passing
  - ✅ **query_actions.test.ts**: All tests passing

  ## Dataset Selection Logic

  ### Priority Order (After Changes)

  1. preferredDataset (passed to resolveDataset)
  2. defaultDataset (user's default, if compatible) ← Enhanced
  3. cachedDataset (last used, if compatible) ← Fetches full dataset to check signalType
  4. fetchFirstAvailableDataset (search for compatible)
    - For required signalType: Exact match only
    - For Logs: Prefer signalType='logs', fallback to no signalType

  ### Compatibility Checks

  - **Traces page**: Requires `signalType: "traces"`
  - **Metrics page**: Requires `signalType: "metrics"`
  - **Logs page**: Accepts `signalType: "logs"` OR no signalType (compatible fallback)
  - **Incompatible**: Traces/Metrics datasets on Logs page are rejected

  ## Breaking Changes

  None. This is a bug fix that improves existing behavior without changing APIs.

  ## Additional Notes

  - The fix handles all query languages supported by OpenSearch Dashboards
  - URL state management properly cleans up incompatible parameters
  - Dataset resolution is more robust with proper signalType validation
  - Editor state is synchronized with query state across flavor switches

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts
```diff
@@ -24,74 +24,132 @@ import { ExploreFlavor } from '../../../../common';
...
-        if (visualization) {
-          const { chartType, params, axesMapping } = JSON.parse(visualization);
-          visualizationBuilder.setVisConfig({ type: chartType, styles: params, axesMapping });
+        // Parse and clean _g parameter to remove query
+        if (gParam) {
+          try {
+            // Decode the RISON-encoded parameter
+            const decodedG = decodeURIComponent(gParam);
+            // Remove the query field from _g by rebuilding without it
+            const cleanedG = decodedG
```
yubonluo: Should we consider the situation of nested parentheses？

src/plugins/explore/public/application/pages/logs/logs_page.tsx
```diff
@@ -78,6 +80,16 @@ export const LogsPage: React.FC<Partial<Pick<AppMountParameters, 'setHeaderActio
     execute: () => dispatch(setActiveTab(EXPLORE_VISUALIZATION_TAB_ID)),
   });
 
+  // Explicitly clear editor text if query is empty (e.g., after flavor switch redirect)
+  const queryText = useSelector((state: RootState) => state.query.query);
+  const setEditorText = useSetEditorText();
+
+  useEffect(() => {
+    if (queryText === '' || !queryText) {
```
yubonluo: Nit: we can simplify it to `!queryText`

src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts
```diff
@@ -24,74 +24,132 @@ import { ExploreFlavor } from '../../../../common';
...
+              .replace(/,?query:\([^)]*\),?/g, ',')
+              .replace(/,,/g, ',')
+              .replace(/\(,/g, '(')
+              .replace(/,\)/g, ')');
```
ps48: This manually parses RISON with regex, which is inherently fragile. The codebase already uses rison-node (see encode_decode_state.ts)
```
  // Add at top of file:
  import rison from 'rison-node';

  // Replace the regex block with:
  if (gParam) {
    try {
      const decoded = rison.decode(decodeURIComponent(gParam)) as Record<string, unknown>;
      delete decoded.query;
      gParam = encodeURIComponent(rison.encode(decoded));
    } catch (e) {
      // Failed to clean _g parameter, use as-is
    }
  }
```

src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts
```diff
@@ -24,74 +24,132 @@ import { ExploreFlavor } from '../../../../common';
...
+
+          services.osdUrlStateStorage.flush();
         }
 
-        // Update savedSearch to store just the ID (like discover)
-        // TODO: remove this once legacy state is not consumed any more
-        dispatch(setSavedSearch(savedExplore.id));
+        // Use window.location to navigate since application service might not be available
+        const currentPath = window.location.pathname;
+        const currentHash = window.location.hash;
```
ps48: Not sure will there be a case when application service might not be available? 
  
```
  // Before (forces full page reload):
  window.location.href = baseUrl;
  return;

  // After (SPA navigation):
  const basePath = services.http?.basePath?.get() ?? '';
  const flavorPath = `/app/explore/${currentFlavor}`;
  const newHash = gParam ? `#/?_g=${gParam}` : '#/';

  application?.navigateToUrl(`${basePath}${flavorPath}${newHash}`);
  
  return;
```

This preserves SPA state and avoids a full browser reload. The application service is already in the deps array, so it should be available.

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
```diff
@@ -67,10 +67,129 @@ export const loadReduxState = async (services: ExploreServices): Promise<RootSta
...
+        const baseUrl = services.application?.getUrlForApp(`explore/${currentFlavor}`, {
+          path: '#/',
+        });
+        if (baseUrl && services.application) {
+          services.application.navigateToUrl(baseUrl);
+          // Return a basic state to prevent further processing
+          return await getPreloadedState(services);
+        }
+      }
+    }
```
ps48, line 204: The view-route incompatibility check in `loadReduxState` duplicates the flavor-mismatch logic in `useInitPage`. Since `useInitPage` already handles the redirect at the React hook level, remove the navigation from `loadReduxState` and have it only handle state cleanup:

```
  // just return clean preloaded state — let useInitPage handle the redirect
  if (isIncompatibleForView) {
    // Clear incompatible URL state so useInitPage gets a clean slate
    services.osdUrlStateStorage.set('_q', null, { replace: true });
    services.osdUrlStateStorage.set('_a', null, { replace: true });
    return await getPreloadedState(services);
  }
```

This way `redux_persistence.ts` is responsible only for state, and `use_page_initialization.ts` is responsible for navigation. No race.

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
ps48, line 158: Minor: can we extract helpers functions from loadReduxState to reduce complexity. 

```
  // In loadReduxState, replacement block with and further chunk :
  if (!isViewRoute && queryState?.dataset) {
    const urlSignalType = await resolveSignalType(services, queryState.dataset);

    const shouldReplace =
      isDatasetIncompatible(urlSignalType, currentRequiredSignalType) ||
      shouldPreferDefault(defaultDataset, queryState.dataset, currentRequiredSignalType);

    if (shouldReplace) {
      replacementDataset = await findReplacementDataset(
        services, currentFlavor, currentRequiredSignalType, defaultDataset
      );
      services.osdUrlStateStorage.set('_q', null, { replace: true });
      queryState = null;
    }
  }

File: src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts [hunk 6]
```diff
@@ -67,10 +67,129 @@ export const loadReduxState = async (services: ExploreServices): Promise<RootSta
...
+
+    // Get the default dataset to check if we should prefer it over URL cached dataset
```
ps48: Currently `getDatasetService()?.getDefault()` is called in loadReduxState, fetchFirstAvailableDataset, resolveDataset, and the fallback at the end of resolveDataset. Should we call it once at the top of each entry point and pass it down?

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
```diff
@@ -262,43 +438,103 @@ const resolveDataset = async (
...
+        ?.getDefault();
+      if (fallbackDefaultDataset) {
+        return fallbackDefaultDataset;
+      }
+    } catch (error) {
+      // Continue to return undefined
+    }
+  }
+
+  return compatibleDataset;
```
ps48: Will this be always undefined here? As we are checking for `compatibleDataset` and `!compatibleDataset` in the lines above?

src/plugins/agent_traces/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -44,6 +44,12 @@ export const defaultPrepareQueryString = (query: Query): string => {
   switch (query.language) {
     case 'PPL':
       return defaultPreparePplQuery(query).query;
+    case 'PROMQL':
+    case 'kuery':
+    case 'DQL':
+    case 'SQL':
+    case 'lucene':
+      return query.query as string;
```
ps48: Let's use constants here. `export const PASSTHROUGH_LANGUAGES = ['kuery', 'DQL', 'SQL', 'lucene', 'PROMQL'] as const;`

ps48 (commented):
Should we add some more jest tests for `loadReduxState` 
```
  it('should return preloaded state when view route has incompatible dataset', async () => {
      // Setup: window.location.hash = '#/view/some-id'
      // queryState has dataset with signalType: 'traces'
      // currentFlavor is 'logs'
      // Expect: returns getPreloadedState result, clears _q and _a
    });

    it('should replace dataset when URL dataset is incompatible with current flavor', async () => {
      // Setup: NOT a view route, queryState has traces dataset, currentFlavor is logs
      // Expect: replacementDataset is fetched, queryState is cleared
    });

    it('should prefer default dataset over URL dataset when default is a perfect match', async () => {
      // Setup: defaultDataset has signalType: 'logs', URL dataset has no signalType
      // Expect: defaultDataset is used as replacement
    });

    it('should clear query when dataset changes due to replacement', async () => {
      // Setup: queryState has query text, dataset gets replaced
      // Expect: finalQueryState.query is ''
    });

    it('should keep query when dataset is compatible and unchanged', async () => {
      // Setup: queryState has query and compatible dataset
      // Expect: finalQueryState.query preserves original
    });

    it('should call setQuery with forceUpdate=true when dataset was replaced', async () => {
      // Expect: setQuery called with (state, true)
    });
```

src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts
TackAdam: Replaced the fragile regex pattern with proper rison-node library for parsing. The original regex [^)]* would fail on nested structures. We now use:
```
  import rison from 'rison-node';

  if (gParam) {
    try {
      const decoded = rison.decode(decodeURIComponent(gParam)) as Record<string, unknown>;
      delete decoded.query;
      gParam = encodeURIComponent(rison.encode(decoded));
    } catch (e) {
      // Failed to clean _g parameter, use as-is
    }
  }
```
  This properly handles arbitrary nesting levels in RISON-encoded parameters.

src/plugins/explore/public/application/pages/logs/logs_page.tsx
TackAdam: Good catch, updated!

src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts
TackAdam: Replaced manual regex parsing with the proper rison-node library that's already available in the codebase. Added import and used library methods for encoding/decoding instead of string manipulation.
```
  import rison from 'rison-node';

  const decoded = rison.decode(decodeURIComponent(gParam)) as Record<string, unknown>;
  delete decoded.query;
  gParam = encodeURIComponent(rison.encode(decoded));
```

src/plugins/explore/public/application/utils/hooks/use_page_initialization.ts
TackAdam: Replaced window.location.href (which causes full page reload) with proper SPA navigation using application.navigateToUrl(). Added proper basePath construction:
```
  const basePath = http?.basePath?.get() ?? '';
  const flavorPath = `/app/explore/${currentFlavor}`;
  const newHash = gParam ? `#/?_g=${gParam}` : '#/';

  application?.navigateToUrl(`${basePath}${flavorPath}${newHash}`);
```
  This preserves SPA state and avoids unnecessary browser reloads.

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
TackAdam: Removed navigation logic from loadReduxState and kept it responsible only for state management. Navigation is now handled exclusively by useInitPage:
```
  if (isIncompatibleForView) {
    // Clear incompatible URL state so useInitPage gets a clean slate
    // useInitPage will handle the redirect - redux_persistence.ts is only responsible for state
    services.osdUrlStateStorage.set('_q', null, { replace: true });
    services.osdUrlStateStorage.set('_a', null, { replace: true });
    return await getPreloadedState(services);
  }
```
  This eliminates race conditions and maintains clear separation of concerns.

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
TackAdam: Extracted 4 helper functions from loadReduxState to improve readability and maintainability:

  1. resolveSignalType() - Fetches signal type from dataset if not present
  2. isDatasetIncompatible() - Checks if dataset signal type is incompatible with current flavor
  3. shouldPreferDefault() - Determines if default dataset should be preferred over URL dataset
  4. findReplacementDataset() - Finds replacement dataset when URL dataset is incompatible

  This reduced the main logic block from ~73 lines to ~38 lines, making the code much more maintainable.

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
TackAdam: Refactored to call getDefault() only once at the top of loadReduxState and pass it as a parameter through all helper functions:
```
  // Fetch default dataset once at the top
  let defaultDataset: Dataset | undefined;
  defaultDataset = currentFlavor === ExploreFlavor.Metrics
    ? undefined
    : services.data?.query?.queryString?.getDatasetService()?.getDefault();
```
  Updated function signatures:
  - fetchFirstAvailableDataset() - Added defaultDataset parameter
  - resolveDataset() - Added defaultDataset parameter
  - getPreloadedQueryState() - Added defaultDataset parameter
  - getPreloadedState() - Added defaultDataset parameter

  This eliminated 3+ redundant calls to getDefault().

src/plugins/explore/public/application/utils/state_management/utils/redux_persistence.ts
TackAdam: Removed redundant logic that checked if (compatibleDataset) return compatibleDataset followed by return compatibleDataset. Simplified to directly return the result:
```
  // Before
  if (compatibleDataset) {
    return compatibleDataset;
  }
  return compatibleDataset;

  // After
  return await fetchFirstAvailableDataset(
    services,
    flavorFromAppId,
    requiredSignalType,
    defaultDataset
  );
  ```

src/plugins/agent_traces/public/application/utils/state_management/actions/query_actions.ts
TackAdam: Updated,
Kept agent constant separate to avoid required explore as app dependency.

TackAdam:
After Updates:

https://github.com/user-attachments/assets/38970ad7-fab0-435d-b129-55e3fb88ab88

Navigation working:

https://github.com/user-attachments/assets/5c6407d7-4442-4f08-830a-23ad003cc642

ps48 (approved):

TackAdam:
Switched to fix https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11521

Issue #11491, opened 2026-03-10
[BUG] shortURL does not work properly with storeInSessionStorage


abbyhu2000:
**Describe the bug**

The root cause is a fundamental incompatibility between OpenSearch Dashboards short URL functionality and the setting when enabled. When storeInSessionStorage is ON, dashboard state is stored in browser sessionStorage with URLs containing only hash references. Opening a short URL in a different browser, session, or tab causes hashedItemStore.getItem(stateHash) to return null, triggering the error in the retrieveState() function. 

**To Reproduce**

**Expected behavior**

**OpenSearch Version**
Please list the version of OpenSearch being used.

**Dashboards Version**
Please list the version of OpenSearch Dashboards being used.

**Plugins**

Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

PR #11506, opened 2026-03-12
Add _q _v flag to hash and unhash logics

abbyhu2000:
### Description

We miss _q and _v in the URL hash/unhash logics, thus causing share link feature to break when we turn on session storage. It was broken for both explorer and classic discover since both uses the flags to store states.

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11491

## Screenshot

previous:

https://github.com/user-attachments/assets/f1b56d0a-8faa-457f-bb5b-aac4

https://github.com/user-attachments/assets/07cf264a-9ecf-4b12-ae40-cb694981bb02

4abb7852

after:

new explorer
https://github.com/user-attachments/assets/8d7f5fcc-0493-4ef2-9d7f-263c1a594701

classic discover

https://github.com/user-attachments/assets/07c37daa-ce07-43ee-bd78-f50236cd494d

visualization

https://github.com/user-attachments/assets/9aa21023-7d17-40cf-bc83-1203b3e957f4

## Changelog
- fix: Fix share links when session storage is on

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/11506.yml
+++ b/changelogs/fragments/11506.yml
@@ -1,0 +1,2 @@
+fix:
+- Fix share links when session storage is on ([#11506](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11506))


ruchidh:
can we test it like large number of filters, as that's what cx reported ?
also if we can double check the viz page too

abbyhu2000:
> can we test it like large number of filters, as that's what cx reported ? also if we can double check the viz page too

will do, but can we merge this in first? this is a bug that needs to be fixed for explorer

Maosaic (approved):

LDrago27:
@abbyhu2000 The test of 

> share link feature to break when we turn on session storage

Seems a bit of niche use case. Do we want to add in a cypress test to capture this .

abbyhu2000:
> @abbyhu2000 The test of
> 
> > share link feature to break when we turn on session storage
> 
> Seems a bit of niche use case. Do we want to add in a cypress test to capture this .

good point, will add a task to add a cypress test for this specific scenario

PR #11510, opened 2026-03-12
[Chat] Fix chat input disabled state during tool result sending

wanglam:
### Description

This PR adds a polling mechanism to wait for tool call results to sync to agentic memory before sending tool results, and disables the chat input during this period to prevent user input while tool results are being processed.

Key changes:
- Added `isSendingToolResult` state to track when tool results are being sent
- Added `onSendToolResultStateChange` callback to `ChatEventHandler` to notify state changes
- Disabled chat input and send button when `isSendingToolResult` is true
- Added `waitForToolCallSync` method in `ChatService.sendToolResult` that polls conversation history to verify tool call sync before proceeding
  - Configurable max attempts (default: 10) and interval (default: 1000ms)
  - Only waits when `includeFullHistory` is false (when full history is included, messages are passed directly so no sync wait needed)
  - Gracefully handles timeout with warning log

## Screenshot

N/A - No visual UI changes, only behavioral change to disable input during tool result sending.

## Testing the changes
1. Start a chat conversation with the AI assistant
2. Trigger a tool call (e.g., ask the assistant to perform an action that requires tool execution)
3. Observe that the chat input is disabled while the tool result is being sent
4. After the tool call is synced and tool result is sent, the input should be re-enabled
5. Run the new unit tests:
   - `yarn test:jest src/plugins/chat/public/components/chat_input.test.tsx`
   - `yarn test:jest src/plugins/chat/public/services/chat_event_handler.test.ts`
   - `yarn test:jest src/plugins/chat/public/services/chat_service.test.ts`

## Changelog

- fix: [Chat] Disable chat input during tool result sending to prevent race conditions

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
- [ ] Update [[CHANGELOG.md](vscode-webview://0iain2m3os03u49kt4kui01tjdt820nffld08qlu0g26b9i5h9hk/CHANGELOG.md)](./../CHANGELOG.md)

### DIFF
--- a/src/plugins/chat/public/services/chat_service.ts
+++ b/src/plugins/chat/public/services/chat_service.ts
@@ -269,17 +269,16 @@ export class ChatService {
     observable: any;
     userMessage: UserMessage;
   }> {
+    // Start new thread first to avoid restoring from latest conversation when window opens
+    if (options?.clearConversation) {
+      this.newThread();
+    }
     // Ensure window is open and get the window instance
     const chatWindowInstance = await this.openWindow();
 
-    // Clear conversation if requested (create new thread)
+    // Reset chat window UI to a fresh chat panel
     if (options?.clearConversation) {
-      this.newThread();
-
-      // If we have ChatWindow instance, also clear its conversation
-      if (chatWindowInstance) {
-        chatWindowInstance.startNewChat();
-      }
+      chatWindowInstance.startNewChat();
     }
 
     await chatWindowInstance.sendMessage({ content, messages });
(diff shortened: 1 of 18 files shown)


src/plugins/chat/public/services/chat_service.ts
```diff
@@ -516,6 +516,11 @@ export class ChatService {
       forwardedProps: {},
     };
 
+    // Add 2 seconds wait for tool call been synced to the agentic memory
+    await new Promise((resolve) => {
+      setTimeout(resolve, 2000);
+    });
+
```
Hailong-am: By default index refresh interval is 1s, do we need to set 2s here?

wanglam: Let's keep the current implementation. Maybe we can do some optimizations to only add delay for agentic memory provider. What do you think about it?

ruanyl (approved):

src/plugins/chat/public/services/chat_service.ts
jiapingzeng: Per internal doc, AOSS index refresh interval 1s does not translate to read-after-write consistency in 1s. From testing, after applying the new refresh interval, read-after-write consistency generally hovers around 0.8s-1.5s. So I think 2s would be safer. 

If need arise in the future, maybe we can make this delay configurable.

PR #11524, opened 2026-03-13
[chat] Integrate "On Behalf Of" Bearer Token usage with AG-UI authorization flow AG-UI

epugh:
### Description

Users need to make calls to OpenSearch using their current logged in identity via the Chat plugin.

The flow is OSD --> Chat --> OpenSearch Agent Server --> OpenSearch MCP Tools Service --> OpenSearch.

This specifically integrates the OBO token flow by requesting this secure token via:

OSD --> Chat --> Security Plugin OBO --> Chat --> OpenSearch Agent Server --> OpenSearch MCP Tools Service --> OpenSearch

### Issues Resolved

n/a

## Screenshot

n/a

## Testing the changes

Set up a secure OpenSearch server and OpenSearch Dashboards.  Configure Agent Server and MCP Tool Server and send tool chain through.   The `forwardCredentials` is the key thing:

opensearch_dashboards.yml:
```
chat:
  enabled: true
  forwardCredentials: true
  agUiUrl: "http://localhost:8001/runs"
```

We did End 2 End testing using the Chorus reference implementation: https://github.com/o19s/chorus-opensearch-edition/pull/175 to validate all the data flows.

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/chat/server/routes/index.test.ts
+++ b/src/plugins/chat/server/routes/index.test.ts
@@ -887,7 +887,7 @@ describe('generateOboToken', () => {
     );
     expect(mockTransportRequest).toHaveBeenCalledWith({
       method: 'POST',
-      path: '/_plugins/_security/api/generateonbehalfoftoken',
+      path: '/_plugins/_security/api/obo/token',
       body: { description: 'OBO token for AG-UI credential forwarding' },
     });
   });
--- a/src/plugins/chat/server/routes/index.ts
+++ b/src/plugins/chat/server/routes/index.ts
@@ -31,7 +31,7 @@ export async function generateOboToken(
     const client = context.core.opensearch.client.asCurrentUser;
     const { body } = await client.transport.request({
       method: 'POST',
-      path: '/_plugins/_security/api/generateonbehalfoftoken',
+      path: '/_plugins/_security/api/obo/token',
       body: {
         description: 'OBO token for AG-UI credential forwarding',
       },


SuZhou-Joe:
Actually OSD is using session based authentication, 
image

No sure if this change will work. And seems there is a lint error.

epugh:
> Actually OSD is using session based authentication, image
> 
> No sure if this change will work. And seems there is a lint error.

So that is what I thought as well..   And I actually really struggled to understand how things would work.   I too looked at the Browser inspector and thought, where am I getting these headers from?

Fortunately, @mingshl had a clearer understanding..   What we see in the browser is NOT what goes to Agent Server.  The proxy, IIUC, which is running on the backend, actually adds the appropriate headers.  And that is where we get what we need.   So it's Browser --> React App --> Backend Node App (adding headers) --> Agent Server....

seraphjiang:
@epugh @SuZhou-Joe @mingshl  

authorization header is generated from security dasboards plugion today ,we treat it as credential. 

the use case seem clear , the question is how could we ensure the credential is only being scoped to invoke opensearch? 

> The flow is OSD --> Chat --> OpenSearch Agent Server --> OpenSearch MCP Tools Service --> OpenSearch.

while the code might be straight and simple, i would suggest a security review. 

At least, add config to allow OSD admin to decide if they want to allow forward credential to downstream chat's backend. 
also for minimal audit purpose, shoud we log the activity (whose credential forward to which endponint, not the credential itself)


-
related code in security plugin. 

Basic auth

https://github.com/opensearch-project/security-dashboards-plugin/blob/353df7d75461e48ddbfd2e879162ac79cc2f3aa3/server/auth/types/basic/basic_auth.ts#L35

SAML auth
https://github.com/opensearch-project/security-dashboards-plugin/blob/353df7d75461e48ddbfd2e879162ac79cc2f3aa3/server/auth/types/saml/saml_auth.ts#L129

DEVELOPER_GUIDE.md
```diff
@@ -438,7 +438,7 @@ remove it, don't simply comment it out.
 
 We are gradually moving the OpenSearch Dashboards code base over to Prettier. All TypeScript code
 and some JavaScript code (check `.eslintrc.js`) is using Prettier to format code. You
-can run `node script/eslint --fix` to fix linting issues and apply Prettier formatting.
+can run `yarn lint --fix` to fix linting issues and apply Prettier formatting.
```
seraphjiang: @SuZhou-Joe  @SuZhou-Joe  @abbyhu2000 should this be seperate PR as small improvement on guideline not related to the change?

SuZhou-Joe:
```suggestion
can run `yarn run lint:es --fix && yarn run lint:style --fix` to fix linting issues and apply Prettier formatting.
```
Recommend to change like this, and agree we should move this change to a standalone PR.

SuZhou-Joe:
> 
> So that is what I thought as well.. And I actually really struggled to understand how things would work. I too looked at the Browser inspector and thought, where am I getting these headers from?
> 
> Fortunately, @mingshl had a clearer understanding.. What we see in the browser is NOT what goes to Agent Server. The proxy, IIUC, which is running on the backend, actually adds the appropriate headers. And that is where we get what we need. So it's Browser --> React App --> Backend Node App (adding headers) --> Agent Server....

Right, actually the Node.js server should have all the auth information(For the backend opensearch cluster), and we can pass down auth information in that layer, instead of in React App -> Backend Node App server.

DEVELOPER_GUIDE.md
epugh: Makes sense...  I'll open up a new PR with this.  I'm just learning the OSD code base, so this is great to see some of these suggestions.

epugh: I have opened https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11531 for this, and will back this change out.

jiapingzeng:
@epugh could we implement this suggestion to address security concerns:

> At least, add config to allow OSD admin to decide if they want to allow forward credential to downstream chat's backend.
also for minimal audit purpose, should we log the activity (whose credential forward to which endponint, not the credential itself)

Seems that a feature flag that's disabled by default could be a fast way to move forward on this.

You can find some of the existing flags here: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/chat/common/chat_capabilities.ts

seraphjiang:
> @epugh could we implement this suggestion to address security concerns:
> 
> > At least, add config to allow OSD admin to decide if they want to allow forward credential to downstream chat's backend.
> > also for minimal audit purpose, should we log the activity (whose credential forward to which endponint, not the credential itself)
> 
> Seems that a feature flag that's disabled by default could be a fast way to move forward on this.
> 
> You can find some of the existing flags here: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/chat/common/chat_capabilities.ts

thanks @jiapingzeng for sharing the feature flag for chat, i believe we need seperate feature flag to control credential forward. 
so admin could enable chat but disable forward.

seraphjiang:
besides the feature control credential forward. 

@SuZhou-Joe raise a good point not all authentication we support use Authorization headers. ideally we should solution to cover all. otherwise we will go to situtation. only certain authentication type suppport the feature. 

it is good take incremental approaach, as long as we evaluate and likely have a plan.

mingshl:
I do agree on @searaphjiang 's point that, we should make passing auth header configurable, so that it won't impact the existing behaviors. Now that the next puzzle is which configuration that we should use to allow passing auth header 

### Approach 1, new config `chat.forwardCredentials` : 
since there are several new configuration settings under `chat.`, maybe we can follow the similar pattern, adding `chat.forwardCredentials: true`, in this case, only admin can decide if dashboard can forward auth header or now. At default, it's false and it won't pass through the auth header.

```
  Usage in opensearch_dashboards.yml:
  chat.enabled: true
  chat.agUiUrl: "http://agent-server:3000"
  chat.forwardCredentials: true   # admin explicitly opts in
```

  Add a simple boolean to the existing chat config schema:
```
  // src/plugins/chat/server/config.ts
  export const configSchema = schema.object({
    enabled: schema.boolean({ defaultValue: false }),
    agUiUrl: schema.maybe(schema.string()),
    forwardCredentials: schema.boolean({ defaultValue: false }),  // NEW
    mlCommonsAgentId: schema.maybe(schema.string()),
    observabilityAgentId: schema.maybe(schema.string()),
  });
```

### Approach 2: reuse `opensearch.requestHeadersAllowlist`
but I am thinking an even more secure approach if we can reuse a security configuration: `opensearch.requestHeadersAllowlist`

I am not very familiar with this flag, but in my understanding, if   `opensearch.requestHeadersAllowlist` (default: ['authorization']) that controls which headers are forwarded to OpenSearch. We could follow the same pattern:

```
  // src/plugins/chat/server/config.ts
  export const configSchema = schema.object({
    enabled: schema.boolean({ defaultValue: false }),
    agUiUrl: schema.maybe(schema.string()),
    agUiRequestHeadersAllowlist: schema.arrayOf(schema.string(), { defaultValue: [] }),  // NEW
    mlCommonsAgentId: schema.maybe(schema.string()),
    observabilityAgentId: schema.maybe(schema.string()),
  });
```

  Usage:
  chat.agUiRequestHeadersAllowlist: ["authorization"]


@seraphjiang @cwperks let me know this is a good use of the security setting `opensearch.requestHeadersAllowlist`, or approach 1 is more straightforward. Opinion wanted.

cwperks:
Ideally, I think we would need to do token vending and give mechanisms for the user to control what the tokens can do.

cwperks:
The AI review is spot on here: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11524#issuecomment-4067606973

We should not be relaying credentials outside of the application. The user should be prompted to allow OpenSearch to do things on their behalf.

We actually had something built into the security plugin for the extensions project called on-behalf-of tokens, but the functionality is currently in a very basic state. Essentially its a short-lived token that's connected to the user's identity and can perform actions on behalf of the authenticated user. Maybe we can scope out exactly what's required and see if the feature can be extended.

I also wanted to point out that [API Keys](https://github.com/opensearch-project/security/issues/1504) are the #1 feature request on the security repo and I would love to see them implemented. Those could be another thing to consider as well where authenticated user can request a token, then go to setup for AG UI and supply the token which is an act of authorizing AG UI to act on their behalf.

We would certainly want to do some sort of security review for a feature like this. Let me check what was done for Pull-Based Ingestion.

cwperks:
btw the obo documentation is here: https://docs.opensearch.org/latest/security/access-control/authentication-tokens/

at the very minimum, I would think we could require an admin to configure that feature and have OSD request a token and then forward it.

I would also support extending the feature to make it simpler for this use case. For instance, currently it requires both a `signing_key` and `encryption_key`. The signing key is needed to issue the jwts and the `encryption_key` was added for extensions to make roles and backend roles opaque to extensions to discourage using them directly as some plugins do today.

mingshl:
@cwperks make sense, instead of passing through the entire auth header, user can request, then using the on-behalf-of(obo) token seems more secure approach. 

the opensearch agent sever and opensearch mcp sever are remaining the same way as passing through header and the obo token will be validated when it reaches opensearch security plugin. 


 The full flow with OBO tokens would be:
```
  OSD (mints OBO token via security plugin)
    → Authorization: Bearer <OBO-token>
      → AG-UI (forwards the header)
        → Authorization: Bearer <OBO-token>
          → MCP Server (OPENSEARCH_HEADER_AUTH=true, --transport stream)
            → Authorization: Bearer <OBO-token>
              → OpenSearch (validates OBO token, applies user's permissions)
```

This address @SuZhou-Joe's concern that not all auth types use Authorization headers (SAML, session cookies, etc.). OBO tokens normalize them all into a JWT. 

In this PR, we can just need to do

1. add new config `chat.forwardCredentials`, default is false
2. in forwardToAgUI(), when config is enabled:
    - Use context.core.opensearch.client.asCurrentUser (which already has the user's auth) to call `POST /_plugins/_security/api/generateonbehalfoftoken`
    - Extract the authenticationToken from the response
    - Forward it as Authorization: Bearer <token> to AG-UI
 3. if failed, handle token errors gracefully — if OBO isn't configured on the cluster, log a clear error. 
 4. if success, add audit logging — log that a credential was forwarded to the agui_url, which can be read from another config. 
 
```
  Before:  request.headers.authorization → AG-UI (raw credential)

  After:   request → call OpenSearch OBO API using scoped client
             → get short-lived JWT
             → forward JWT as "Authorization: Bearer <obo-token>" → AG-UI
```

But before this plan, I have several questions, 
  1. Is the OBO token feature ready for this use case today?
  Our plan would be: the chat plugin's server-side route calls POST /_plugins/_security/api/generateonbehalfoftoken using the user's scoped OpenSearch client (asCurrentUser), then forwards the
  returned JWT as a Authorization: Bearer <obo-token> header to AG-UI. Is the current OBO implementation sufficient for this, or are there gaps we'd need to address first?

  2. How much admin setup burden does the OBO configuration require?
  Enabling OBO requires configuring on_behalf_of.enabled: true, signing_key, and encryption_key in the security config. You mentioned the encryption_key was originally added for extensions. Is it
  feasible to simplify the setup (e.g., auto-generate keys, or make encryption_key optional)? Or is the current setup straightforward enough that we just need to document it?

  3. Can we start with OBO tokens carrying full user roles (no scope-down) and add scoping later?
  For the initial implementation, we'd forward the OBO token with the user's full mapped roles. Scoped-down tokens (e.g., restricting to specific indices or read-only access) would be a future
  enhancement. Does that sound like a reasonable incremental approach?

  4. How should security:obo/create permission be granted to users?
  Should security:obo/create be added to a default role (like all_access), or should admins always explicitly grant it? For example, @epugh has sample users rumi and pete — would the admin need to
  create a role with security:obo/create and map it to every user who uses the chat feature? If so, that could be significant adoption friction we should think about.

  5. Will excluding backend roles from OBO tokens cause issues?
  The current OBO implementation sets includeBackendRoles=false, so backend roles are not embedded in the token. If a user's permissions depend on backend role mappings (common in SAML/OIDC setups),
   would those permissions be lost when using the OBO token through the AG-UI → MCP → OpenSearch flow?

cwperks:
@mingshl I raised this [PR on the security repo](https://github.com/opensearch-project/security/pull/6017) to create a simpler setup that helps in this use-case (and makes the route name more sensible). Note that it would require setting that feature up which requires securityconfig modification (something that people who self-manage can use, but not AOS). 

I guess it really depends here on how trusted ag ui is. Is it to be treated as a external service or a trusted entity? Forward the authorization header is forwarding credentials (including pw in the case of basic auth). Tokens also provide access, but the main advantage is that they are time bound and don't reveal the raw credentials.

kkhatua:
@epugh  Could you create a GitHub issue to link this to? 
A lot of the above conversation is valuable and should be sitting in a GH issue with other PRs (e.g. like the one from @cwperks ) making references to it.

epugh:
> @epugh Could you create a GitHub issue to link this to? A lot of the above conversation is valuable and should be sitting in a GH issue with other PRs (e.g. like the one from @cwperks ) making references to it.

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11537

mingshl:
The CI failures are not related to the code changes, @seraphjiang @SuZhou-Joe @cwperks would you review the new commit using OBO token?

Hailong-am:
Another thing in my mind is we need to consider about multiple data source. With [MDS](https://docs.opensearch.org/latest/dashboards/management/multi-data-sources/#step-2-create-a-new-data-source-connection), it supports different authentication method. It will not use the token for current login user instead of the credential for datasource.

1. Username & Password: A basic username and password are used to connect to the data source.
2. AWS SigV4: An AWS Signature Version 4 authenticating request is used to connect to the data source. AWS Signature Version 4 requires an access key and a secret key. 


@cwperks  will OBO support cross cluster which means OBO token issued at cluster A and it could use in cluster B?

cwperks:
> @cwperks will OBO support cross cluster which means OBO token issued at cluster A and it could use in cluster B?

If they share the same `signing_key` then yes. This is a symmetric key and used for both issuance and verification of the tokens.

src/plugins/chat/server/routes/index.test.ts
```diff
@@ -741,3 +844,108 @@ describe('Chat Proxy Routes', () => {
...
+
+    const token = await generateOboToken(mockContext, mockLogger, 'http://agui:3000');
+
+    expect(token).toBe('obo-jwt-token-123');
+    expect(mockLogger.info).toHaveBeenCalledWith(
+      expect.stringContaining('OBO token generated for credential forwarding to AG-UI endpoint')
+    );
+    expect(mockTransportRequest).toHaveBeenCalledWith({
+      method: 'POST',
+      path: '/_plugins/_security/api/generateonbehalfoftoken',
```
cwperks: nit: can we use the endpoint `/_plugins/_security/api/obo/token`. Its the same endpoint, without all the verbiage of this and introduced in the same release (3.6.0)

cwperks:
@mingshl the latest changes lgtm. Is there any way to e2e test this on the agent server side in an automated way?

mingshl:
@cwperks Thanks for the review! I added [use /_plugins/_security/api/obo/token instead](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11524/commits/4ea08017df4a41c39c8fa2e3c11b7a51cfbd43f9) . @eric has a chorus open source dashboard that can allow you to test the new changes with E2E workflow. @epugh Let us know when it's ready to test. Thank you!!

epugh:
I am going to go through https://docs.opensearch.org/latest/security/access-control/authentication-tokens/ and I might ping you on Friday @cwperks before my working session with @mingshl.   Thanks team!

cwperks:
> I am going to go through https://docs.opensearch.org/latest/security/access-control/authentication-tokens/ and I might ping you on Friday @cwperks before my working session with @mingshl. Thanks team!

Works for me. FYI I am making slight alterations to those docs for 3.6: https://github.com/opensearch-project/documentation-website/pull/12109

mingshl:
We had an offline section that pull up this PR along with some new commits on security plugin and we can verify some test cases that the OBO token is working, Would you review and approve if no major concerns? We are hoping to get in the 3.6 release by EOD. 

@cwperks @seraphjiang @kkhatua

epugh:
THANKYOU SO MUCH!

PR #11536, opened 2026-03-17
[Security] update csp configuration

angle943:
### Description

- This is a precursor to enable the strict set of CSP rules we want to ultimately enforce
- The strict CSP rules can be enforced via `csp.strict: true` in the opensearch yml file
- You can "relax" any of these rules via adding `csp.loosenCspDirectives: string[]` in the opensearch yml file. IE suppose you want to get rid of the connect-src directive, then add `csp.loosenCspDirectives: ['connect-src']`

## Changelog
- infra: update csp configuration to allow for strictness

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/server/http_resources/http_resources_service.ts
+++ b/src/core/server/http_resources/http_resources_service.ts
@@ -135,7 +135,7 @@ export class HttpResourcesService implements CoreService<InternalHttpResourcesSe
 
       const modifications = cspModificationsDynamicConfig?.modifications;
       if (modifications && modifications.length > 0) {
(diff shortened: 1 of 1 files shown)


sumukhswamy (commented):
LGTM! one clarification
In strict mode, nonces are now applied via header-building logic? do we need to validate that all nonce-dependent styles/scripts still render correctly

angle943:
@sumukhswamy the csp report has been on with the nonce values, so as long as we don't see violations there, we should be good

PR #11546, opened 2026-03-18
Improve "Build and test" CI test group

Maosaic:
### Summary

- **Faster checkouts:** Add `fetch-depth: 1` + `filter: blob:none` to all `actions/checkout` steps (all workflows)
- **Split integration tests** into a dedicated `integration-tests` job (Linux + Windows matrix) with Java 21, yarn cache, and bootstrap retry; remove them from `build-test` ciGroup1
- **Pre-build plugin bundles once** in a new `build-plugins-artifact` job; `functional-tests` downloads the artifact instead of rebuilding per-runner
- **Skip Windows FTR ciGroups 2–14:** Browser rendering tests are platform-agnostic and covered by Linux; only ciGroup1 (smoke test) runs on Windows
- **Add Jest ciGroup5** and rebalance groups 3/4/5 to keep Windows wall-clock time under 30 min
- **Split TSVB functional tests** into new ciGroup14 (moved out of ciGroup12)
- **Fix `use_node.bat` exit code bug:** `ENDLOCAL` was resetting `ERRORLEVEL` to 0, silently swallowing all Node.js test failures on Windows
- **Fix mocha path-separator regex** for Windows stack traces (`fixtures.{1,2}project.{1,2}test\.js`)
- **Add `--forceExit`** to `test:jest_integration:ci` to prevent hung workers from blocking CI
- **PR comment bot** (`post_ci_results.yml`): now includes integration test failures alongside unit test failures

### Details

**`build-plugins-artifact` job:** Runs once on Linux, uploads `src/core/target/public`, `src/plugins/*/target/public`, and `packages/*/target/public` as the `plugin-bundles` artifact. Each `functional-tests` runner downloads it via `actions/download-artifact@v4` instead of re-running the ~10-minute plugin build step. `functional-tests` now has `needs: [build-plugins-artifact]`.

**`integration-tests` job:** Separate Linux+Windows matrix job. Sets `OSD_OPTIMIZER_MAX_WORKERS: ''` to unset the global CI override so integration tests that check worker count (e.g., `basic_optimization.test.ts`) see the value passed in code (`maxWorkerCount: 1`). Requires Java 21 (OpenSearch won't start on Temurin 17). Uploads `junit-jest-integration-{Linux,Windows}` artifacts; the PR comment bot now downloads these alongside `junit-jest-group*`.

**Jest group rebalancing:**
| Group | Before | After |
|-------|--------|-------|
| 3 | `plugins/[a-d]` | `plugins/[a-c]` + `plugins/[q-u]` |
| 4 | `plugins/[e-u]` + cli/dev/legacy | `plugins/[e-k]` + cli/dev/legacy |
| 5 | _(none)_ | `plugins/[d]` + `plugins/[l-p]` |

`plugins/[d]` (data, dashboard, discover, data_source*) is the heaviest group and was causing Windows group 4 to exceed 30 minutes.

**`use_node.bat` exit code fix:** `SETLOCAL`/`ENDLOCAL` in Windows CMD resets `ERRORLEVEL` to 0. The fix captures the exit code before `ENDLOCAL` and restores it using the `ENDLOCAL & EXIT /B %VAR%` trick (variable expands before `ENDLOCAL` executes). This was masking all test failures on Windows — unit tests, integration tests, and mocha — making them always appear green.

**Mocha regex fix:** Windows Node.js stack traces produce `fixtures\\/test.js` (backslash + forward slash = 2 chars) between path components. The old regex used `.` (exactly 1 char). Changed to `.{1,2}` to handle both platforms. This bug was hidden by the `use_node.bat` exit-code bug above.

## Changelog

- chore: Improve CI test group

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/dev/mocha/__tests__/junit_report_generation.js
+++ b/src/dev/mocha/__tests__/junit_report_generation.js
@@ -105,7 +105,9 @@ describe('dev/mocha/junit report generation', () => {
     });
 
     expect(testFail.$.time).to.match(DURATION_REGEX);
-    expect(testFail.failure[0]).to.match(/Error: FORCE_TEST_FAIL\n.+fixtures.project.test.js/);
+    expect(testFail.failure[0]).to.match(
+      /Error: FORCE_TEST_FAIL\n.+fixtures.{1,2}project.{1,2}test\.js/
+    );
     expect(testFail).to.eql({
       $: {
         classname: sharedClassname,
@@ -120,7 +122,7 @@ describe('dev/mocha/junit report generation', () => {
     expect(beforeEachFail.$.time).to.match(DURATION_REGEX);
     expect(beforeEachFail.failure).to.have.length(1);
     expect(beforeEachFail.failure[0]).to.match(
-      /Error: FORCE_HOOK_FAIL\n.+fixtures.project.test.js/
+      /Error: FORCE_HOOK_FAIL\n.+fixtures.{1,2}project.{1,2}test\.js/
     );
     expect(beforeEachFail).to.eql({
       $: {


.github/workflows/build_and_test_workflow.yml
```diff
@@ -57,6 +57,9 @@ jobs:
 
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+          filter: blob:none
```
joshuali925, line 62: nit. i remember depth is 1 by default, and filter blob:none probably doesn't make a difference for depth 1

Maosaic: I think you are right, what I'm trying to solve the occasional git server congestion causes individual jobs to take **20–25 minutes** just for checkout, I will come up a different solution in my next PR.

PR #11551, opened 2026-03-19
[Dashboard] fix state transfer not added to dashboard vis create dropdown

ruanyl:
### Description
Fix stateTransfer not added when creating visualization from dashboard create visualization dropdown

Previously, when navigating to visualization editors from the vis dropdown menu, it didn't integrate with `stateTransfer`, which make it not able to navigate back to dashboard after editing. This PR fixed the issue.

Screenshot 2026-03-19 at 15 46 51

## Changelog

- fix: stateTransfer not added when creating visualization from dashboard create visualization dropdown

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/visualizations/public/wizard/new_vis_actions.test.tsx
+++ b/src/plugins/visualizations/public/wizard/new_vis_actions.test.tsx
@@ -332,7 +332,7 @@ describe('createNewVisActions', () => {
       await aliasActionCall![1].execute({} as ActionExecutionContext<{}>);
 
       expect(stateTransfer.navigateToEditor).toHaveBeenCalledWith('otherApp', {
-        path: 'otherApp',
+        path: '/path',
         state: { originatingApp: 'dashboard' },
       });
     });


src/plugins/visualizations/public/wizard/new_vis_actions.tsx
```diff
@@ -99,20 +117,22 @@ export const createNewVisActions = (services: {
           ...actionConfig,
           order: 10 * (visTypes.length - i),
           execute: async () => {
-            application.navigateToApp(visType.aliasApp, { path: visType.aliasPath });
+            const currentAppId = await application.currentAppId$.pipe(take(1)).toPromise();
+            await navigateTo(visType.aliasApp, visType.aliasApp, currentAppId);
```
wanglam: Shall we use `visType.aliasPath` as second parameter here?

ruanyl: you're right, good catch! updated

PR #11560, opened 2026-03-20
Add prepare-source and prepare-cypress jobs for CI reliability

Maosaic:
## Summary

- Replace per-job `git checkout` with a single source archive artifact, eliminating redundant checkouts and reducing exposure to GitHub git server outliers (20s–25min observed)
- Add dedicated cache-warming jobs for Linux and Windows yarn caches, converting downstream jobs to restore-only so no job wastes time on cache saves
- Pre-build plugin bundles once in the Cypress workflow instead of rebuilding in every parallel test job
- Split `build-min-artifact-linux-x64` into its own job so BWC tests start immediately when the Linux x64 artifact is ready, without waiting for Windows/macOS/ARM64 builds

## Motivation

The `Build and test` workflow runs 30+ parallel jobs. On `main`, every job independently runs `actions/checkout` and `actions/cache` (restore + save). This creates two reliability and performance issues:

1. **Checkout outliers**: GitHub's git server occasionally takes 5–25 minutes for a single checkout. With 30+ jobs each doing their own checkout, the probability of at least one outlier hitting the critical path is high.
2. **Redundant cache saves**: Every job attempts to save the yarn cache even when it's already warm, wasting time and creating race conditions.

#### New `prepare-source` job
- Single `actions/checkout` at the top of the DAG
- Creates a compressed source archive (~123MB) excluding `.git/`, `node_modules/`, `target/`, `data/`, and other non-essential directories
- Uploads as a GitHub Actions artifact (CDN-backed, ~3s download for downstream jobs)
- Restores the Linux yarn cache; conditionally runs bootstrap and saves the cache only on cache miss

#### New `prepare-windows-cache` job
- Downloads the source archive and runs bootstrap on a Windows runner
- Saves the Windows yarn cache only on cache miss
- Downstream Windows jobs use `cache/restore` (restore-only), eliminating redundant saves

#### New `build-min-artifact-linux-x64` job
- Extracted from the `build-min-artifact-tests` matrix to run independently
- BWC tests depend only on this job, not the full 5-platform matrix
- Unblocks BWC tests as soon as the Linux x64 release artifact is built

#### Downstream job changes
- All jobs (except `lint-and-validate` and `bwc-tests`) replace `actions/checkout` with artifact download + extract
- A minimal git repo is initialized after extraction (`git init` + empty commit) to satisfy tools that call `git ls-files` or `git log` during bootstrap
- Linux jobs use `cache/restore` (restore-only); Windows jobs use `cache/restore` (restore-only); macOS jobs use `cache` (restore + auto-save, no warming job)
- `lint-and-validate` retains `actions/checkout` because it runs `git status` checks that require a real git history
- `bwc-tests` retains `actions/checkout` because it needs `cypress/test_data/` (652MB of pre-generated OpenSearch data snapshots), which is too large to include in the shared source archive

#### Cache key fix
- `lint-and-validate` cache key changed from `yarn-${{ hashFiles('**/yarn.lock') }}` to `yarn-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}` to match the OS-prefixed keys used by all other jobs

#### New `prepare-cypress` job
- Single checkout, bootstrap, and plugin build (`build_opensearch_dashboards_platform_plugins`)
- Uploads two artifacts:
  - `cypress-source-archive`: source code without `node_modules/` or pre-built plugin bundles
  - `cypress-plugin-bundles`: pre-built `target/public` directories for all plugins
- Saves the yarn cache with a cypress-specific key on cache miss

#### Downstream `cypress-tests` changes
- Downloads both artifacts instead of running checkout + bootstrap + plugin build
- Restores yarn cache (restore-only) and runs bootstrap (fast, cache-hit path)
- Eliminates the ~9min plugin build step from each of the 13+ parallel Cypress jobs

### Source archive design decisions

**Exclusion-based archiving**: The archive uses `--exclude` patterns with `./`-prefixed paths to ensure only top-level directories are excluded. This prevents patterns like `--exclude=data` from accidentally removing nested directories like `src/plugins/data/`.

**Test fixture preservation**: The `node_modules` exclusion uses path-specific patterns (`./node_modules`, `./src/plugins/*/node_modules`, etc.) instead of a blanket `--exclude=node_modules` to preserve git-tracked test fixtures under `__fixtures__/node_modules/`.

## Job DAG (before → after)

**Before:**
```
checkout ──→ each job independently
cache    ──→ each job independently (restore + save)
build-min-artifact-tests (5 platforms) ──→ bwc-tests
```

**After:**
```
prepare-source ──→ prepare-windows-cache ──→ downstream Windows jobs (restore-only)
      │
      ├──→ downstream Linux jobs (restore-only)
      ├──→ build-min-artifact-linux-x64 ──→ bwc-tests (own checkout)
      ├──→ build-min-artifact-tests (4 remaining platforms)
      └──→ build-plugins-artifact ──→ build-test jobs
```

## Changelog

- chore: CI Source Archive & Cache Optimization

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/11560.yml
+++ b/changelogs/fragments/11560.yml
@@ -1,2 +1,2 @@
 chore:
-- Add prepare-source and prepare-cypress jobs for CI reliability ([#11560](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11560))
+- CI Source Archive & Cache Optimization ([#11560](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11560))


.github/workflows/build_and_test_workflow.yml
```diff
@@ -24,8 +24,152 @@ env:
...
+            --exclude='**/target' \
+            .
+          mv /tmp/source-archive.tar.gz .
+          ls -lh source-archive.tar.gz
+
+      - name: Upload source archive
+        uses: actions/upload-artifact@v4
+        with:
+          name: source-archive
+          path: source-archive.tar.gz
```
joshuali925, line 102: nice, this should reduce bootstrap failures

.github/workflows/build_and_test_workflow.yml
```diff
@@ -24,8 +24,152 @@ env:
...
+            sleep 10
+          done
+          exit 1
+
+      - name: Save Yarn Cache
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v4
+        with:
+          path: ${{ env.YARN_CACHE_LOCATION }}
+          key: yarn-Linux-${{ hashFiles('**/yarn.lock') }}
```
joshuali925, line 78: is this existing pattern? I'm thinking if code inside `packages/` changed, will the previous cache be used since `yarn.lock` didn't change, and cause issues since bootstrap didn't run after package code change?

.github/workflows/cypress_workflow.yml
```diff
@@ -50,7 +50,117 @@ env:
...
+
+      - name: Save Yarn Cache
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v4
+        with:
+          path: ${{ env.YARN_CACHE_LOCATION }}
+          key: yarn-Linux-cypress-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Build plugins
+        run: node scripts/build_opensearch_dashboards_platform_plugins --no-examples --workers 12
```
joshuali925, line 127: i think this can be optimized to only build one version of the theme
```suggestion
        run: env OSD_OPTIMIZER_THEMES=v9light,v9dark node scripts/build_opensearch_dashboards_platform_plugins --no-examples --workers 12
```

but i'm not sure if we have any cypress tests depending on specific theme version

.github/workflows/cypress_workflow.yml
```diff
@@ -212,8 +321,7 @@ jobs:
           done
           exit 1
 
-      - name: Build plugins
-        run: node scripts/build_opensearch_dashboards_platform_plugins --no-examples --workers 12
+      # Build plugins step REMOVED — downloaded from prepare-cypress artifact
```
joshuali925, line 324: nit. this comment can be removed

.github/workflows/build_and_test_workflow.yml
Maosaic: The yarn cache (yarn cache dir) stores downloaded package tarballs — it's a download cache, not a build cache. It only contains `.tgz` files fetched from the npm registry. It does NOT contain:                                                                                                                                                          
  - `Built/compiled` output from `packages/ `                                                                                                                                      
  - `node_modules/` contents                                                                                                                                                     
  - Any bootstrap artifacts
                                                                                                                                                                               
  When `yarn.lock` hasn't changed, the same packages need to be downloaded, so the cached tarballs are still correct. Every downstream job still runs yarn osd bootstrap which:  
  1. Installs packages from the cache (fast, no network)                                                                                                                       
  2. Builds/links internal `packages/` from source (always runs, uses current code)                                                                                              
                                                                                                                                                                               
  So a change to `packages/` code with no `yarn.lock` change will still be built correctly — bootstrap runs in every downstream job regardless of cache status. The `prepare-source` job only skips its own bootstrap (which is solely to populate the yarn download cache on miss).

PR #11567, opened 2026-03-23
add react grab for development

joshuali925:
### Description

This PR adds https://github.com/aidenybai/react-grab to opensearch dashboards. This helps for AI agent based development to easily target specific page elements in the prompt.

The library is added to devdependencies, and will only be enabled in development mode.

An alternative is [agentation](https://github.com/benjitaylor/agentation), but i found it to break some css styling in OSD.

React grab UI is hidden by default, so there won't be any difference in OSD visually. To trigger it, press and hold Cmd + C.

## Screenshot

(The blur effect comes from a chrome extension, it is not relevant to this PR.)

image

after clicking, it copies the following prompt

```
@<EuiHealth>

<div class="euiFlexGroup eu..." data-react-grab-frozen="">
  <div ...>
  <div ...>
  Success
</div>
  in FlyoutDetailPanel (at /plugin:agentTraces/src/plugins/agent_traces/public/application/pages/traces/flyout/flyout_detail_panel.tsx)
  in TraceDetailsFlyout (at /plugin:agentTraces/src/plugins/agent_traces/public/application/pages/traces/flyout/trace_details_flyout.tsx)
  in /plugin:agentTraces/src/plugins/agent_traces/public/application/pages/traces/flyout/trace_flyout_context.tsx
```

You can also do other actions like add comment before copying

image

## Changelog

- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/public/osd_bootstrap.ts
+++ b/src/core/public/osd_bootstrap.ts
@@ -32,6 +32,9 @@ import { i18n } from '@osd/i18n';
 import { CoreSystem } from './core_system';
 import { ApmSystem } from './apm_system';
 
+// React Grab lets developers select context for coding agents directly from
+// website. It will only be active for development mode. Its UI is hidden to
+// not affect all developers, hold Command-C or Control-C to invoke.
 if (process.env.NODE_ENV === 'development') {
   import('react-grab').then(({ registerPlugin }) => {
     registerPlugin({ name: 'osd-grab-config', theme: { toolbar: { enabled: false } } });


joshuali925:
> react-grab is registered in the application bootstrap with 'toolbar: { enabled: false }', meaning it runs silently with no visible UI indicator... Legitimate dev tools do not deliberately suppress their own UI indicators.

i disabled its UI because i don't want to modify OSD UI for developers who won't need react-grab

src/core/public/osd_bootstrap.ts
```diff
@@ -32,6 +32,12 @@ import { i18n } from '@osd/i18n';
 import { CoreSystem } from './core_system';
 import { ApmSystem } from './apm_system';
 
+if (process.env.NODE_ENV === 'development') {
```
ruchidh, line 38: can we add comment here about its purpose ?

joshuali925: sure, added e14a2d8

PR #11572, opened 2026-03-24
[CSP Security] Rename CSP strict config flag to csp.enable

sumukhswamy:
### Description

Renames the CSP configuration key from [csp.strict] to [csp.enable]
Updates CSP config parsing/runtime wiring so strict CSP behavior is controlled by [csp.enable]
Updates related config mocks/types to align with the new key and resolve bootstrap/typecheck mismatch.

when false and flag not present
image

when flag set to true

image

## Changelog
- skip

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_string/dataset_service/lib/index_data_structure_creator/use_index_fetcher.test.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/lib/index_data_structure_creator/use_index_fetcher.test.ts
@@ -62,7 +62,7 @@ describe('useIndexFetcher', () => {
       expect(mockHttpGet).toHaveBeenCalledWith(
         '/internal/index-pattern-management/resolve_index/test-*',
         {
-          query: { expand_wildcards: 'all', data_source: 'test-datasource' },
+          query: { data_source: 'test-datasource' },
         }
       );
 
@@ -199,7 +199,6 @@ describe('useIndexFetcher', () => {
         '/internal/index-pattern-management/resolve_index/test-*',
         {
           query: {
-            expand_wildcards: 'all',
             data_source: 'test-datasource',
           },
         }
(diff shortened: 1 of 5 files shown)


src/core/server/csp/csp_config.test.ts
```diff
@@ -82,9 +84,10 @@ describe('CspConfig', () => {
     `);
   });
 
-  test('strict true gives strict csp rules', () => {
-    expect(new CspConfig({ strict: true, warnLegacyBrowsers: false })).toMatchInlineSnapshot(`
+  test('enable true gives strict csp rules', () => {
+    expect(new CspConfig({ enable: true, warnLegacyBrowsers: false })).toMatchInlineSnapshot(`
```
LDrago27, line 88: nit: Lets avoid using snapshot tests. They don;t really validate much and need to be modified for tiny changes.

src/core/server/csp/csp_config.ts
```diff
@@ -125,12 +133,14 @@ export class CspConfig implements ICspConfig {
     const source = { ...DEFAULT_CONFIG, ...rawCspConfig };
 
     this.rules = source.rules;
+    this.enable = source.enable;
+    this.strict = source.enable;
+    const isCspEnabled = source.enable || source.strict;
```
LDrago27, line 138: nit: Do we need to have strict flag here for backwards compatibility. If not lets clean it up to only use the enable config

sumukhswamy: this is for backward compatibility as the csp.strict has been the norm for OSD

src/core/server/csp/csp_config.test.ts
sumukhswamy: same here old test modifying it so it does not fail due to name change

LDrago27: No issues. Lets clean up in a later PR

PR #11589, opened 2026-03-25
refactor(explore): colocate visualization rules with chart configs and simplify registry

ruanyl:
### Description
Refactors the Explore plugin's visualization system to give each chart type ownership of its own matching rules and render logic, replacing the centralized rule_repository with colocated VisRule definitions inside each chart config file.

#### Motivation
The previous architecture maintained a large centralized rule_repository.ts (~1,000 lines) that mapped column types to chart types and defined rendering logic separately from the chart configurations. This created tight coupling, made it hard to reason about individual chart types, and required updating multiple files when adding or modifying a visualization.

#### Key changes
**New VisRule interface** — each chart config now defines its own rules with priority, mappings (axis-role-to-field-type), and a render function:
```
interface VisRule<T extends ChartType> {
  priority: number;
  mappings: AxisTypeMapping[];
  render: (props: VisRenderProps<T>) => React.ReactNode;
}
```

**VisualizationRegistry simplified** — instead of maintaining separate rule lists, the registry derives available chart types and performs rule matching directly from registered

## Changelog

- refactor: colocate visualization rules with chart configs and simplify registry

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/.claude/skills/README.md
+++ b/.claude/skills/README.md
@@ -1,0 +1,36 @@
+# Claude Skills Directory
+
+This directory contains Claude Code skills for OpenSearch Dashboards development. 
+
+**Multi-Platform Support**: Skills are available for both Claude Code and Kiro users through corresponding prompt formats.
+
+## Available Skills
+
+### `resolve-cve` ([resolve_cve.md](resolve_cve.md))
+Automatically identify and resolve security vulnerabilities (CVEs) in project dependencies.
+
+- **Documentation**: [resolve_cve_readme.md](resolve_cve_readme.md)
+- **Claude Usage**: `/resolve-cve [--cve_id CVE-2024-12345]`
+- **Kiro Usage**: Also available as Kiro prompt at [.kiro/prompts/resolve_cve.md](../../.kiro/prompts/resolve_cve.md)
+- **Purpose**: Scans GitHub issues for CVEs, verifies presence in codebase, attempts automated fixes
+
+## Adding New Skills
+
+When adding new skills to this directory:
+
+1. **Claude skill file**: `skill-name.md` (the actual skill definition)
+2. **Kiro prompt file**: `../../.kiro/prompts/skill-name.md` (Kiro-format prompt that delegates to Claude skill)
+3. **Documentation**: `skill_name_readme.md` (comprehensive user guide)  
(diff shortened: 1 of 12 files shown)


src/plugins/explore/public/components/visualizations/visualization_registry.ts
```diff
@@ -3,212 +3,296 @@
...
+  ): FindRulesByColumnsResult {
+    const allMap = new Map<string, Array<VisRule<any>>>();
+    const exactMap = new Map<string, Array<VisRule<any>>>();
+    const numCount = numericalColumns.length;
+    const catCount = categoricalColumns.length;
+    const dateCount = dateColumns.length;
+
+    for (const [type, config] of this.visualizations) {
+      if (chartType && chartType !== type) {
+        continue;
```
wanglam, line 214: Nit: Since we will skip the not matched visualizations here. What about just call `this.visualizations.get(chartType)` here?

src/plugins/explore/public/components/visualizations/visualization_registry.ts
```diff
@@ -3,212 +3,296 @@
...
-    rules.forEach((rule) => this.registerRule(rule));
+  public registerVisualization(input: VisualizationType<any> | Array<VisualizationType<any>>) {
+    const visualizations: Array<VisualizationType<any>> = [];
+    if (Array.isArray(input)) {
+      visualizations.push(...input);
+    } else {
+      visualizations.push(input);
+    }
+    for (const visConfig of visualizations) {
+      if (!this.visualizations.has(visConfig.type)) {
```
wanglam, line 288: Nit: Do we need to throw error when vis type already exists?

wanglam (approved):

src/plugins/explore/public/components/visualizations/visualization_registry.ts
ruanyl: right, that's a good call, I will tweak in next PR

src/plugins/explore/public/components/visualizations/visualization_registry.ts
ruanyl: Also thought about this, will come back to when necessary

SuZhou-Joe (approved):

PR #11593, opened 2026-03-25
Add PPL query cancellation via Discover cancel button

ahkcs:
## Summary
- Generate a client-side UUID (`queryId`) for each synchronous PPL query and pass it through the full request pipeline to the OpenSearch `/_plugins/_ppl` endpoint
- Add `POST /api/enhancements/ppl/cancel` server route that queries `/_tasks?actions=*ppl*&detailed=true`, matches tasks by `queryId` in the task description, and cancels all matching tasks via `/_tasks/{taskId}/_cancel`
- Wire the existing Discover cancel button (abort flow) to call the new cancel route, enabling users to cancel long-running PPL queries from the UI
- Each PPL search (data query and histogram aggregation) gets its own `queryId`, and both are cancelled independently when the user clicks cancel

### Backend dependency
Depends on [opensearch-project/sql#5254](https://github.com/opensearch-project/sql/pull/5254) (merged) which registers PPL queries as `CancellableTask` with the OpenSearch task framework.

### Files changed
| File | Change |
|------|--------|
| `common/constants.ts` | Added `PPL_CANCEL` API path |
| `common/types.ts` | Added `queryId` to `EnhancedFetchContext` |
| `common/utils.ts` | Pass `queryId` in request body; call cancel route on abort |
| `public/search/ppl_search_interceptor.ts` | Generate UUID per PPL search |
| `server/routes/index.ts` | Accept `queryId` in route schema; register cancel route |
| `server/routes/ppl_cancel.ts` | New cancel route implementation |
| `server/routes/ppl_cancel.test.ts` | Unit tests (7 cases) |
| `server/utils/facet.ts` | Forward `queryId` to OpenSearch PPL API |

## Changelog
- skip

### Cancel both data + histogram queries (double join)

**OSD server log** — both tasks found and cancelled:
```
server log [18:29:34.131] [info][plugins][queryEnhancements] PPL query cancelled: queryId=ccfec0d5-9113-4489-8bd5-77ff64ae3c7f, taskId=cwL2x0XeS-i-g-kClRr0ew:2594
server log [18:29:34.134] [info][plugins][queryEnhancements] PPL query cancelled: queryId=0f863cc9-aab3-4589-9e88-191268f9f375, taskId=cwL2x0XeS-i-g-kClRr0ew:2593
```

**OpenSearch backend log** — both queries throw `TaskCancelledException` from cooperative `isCancelled()` check in `moveNext()`:
```
[2026-03-25T11:29:34,131][ERROR][o.o.s.p.r.RestPPLQueryAction] Error happened during query handling
Caused by: org.opensearch.core.tasks.TaskCancelledException: The task is cancelled.
        at ...OpenSearchIndexEnumerator.moveNext(OpenSearchIndexEnumerator.java:122)

[2026-03-25T11:29:34,261][ERROR][o.o.s.p.r.RestPPLQueryAction] Error happened during query handling
Caused by: org.opensearch.core.tasks.TaskCancelledException: The task is cancelled.
        at ...OpenSearchIndexEnumerator.moveNext(OpenSearchIndexEnumerator.java:122)
```

**OSD receives error responses** confirming cancellation:
```json
{
  "error": {
    "reason": "Error occurred in OpenSearch engine: The task is cancelled.",
    "type": "TaskCancelledException"
  },
  "status": 500
}
```

## Test plan
- [x] Unit tests pass (7 tests for cancel route, all existing tests unaffected)
- [x] E2E tested: submit long-running PPL query, click cancel, verify both data and histogram tasks are cancelled via server logs
- [x] Both queries cancelled simultaneously — backend confirms `TaskCancelledException` from cooperative `isCancelled()` polling in `moveNext()`
- [x] Verified compatibility with backend PR sql#5254 (`queryId` field name, task description format, task action wildcard)

### DIFF
--- a/src/plugins/query_enhancements/server/routes/ppl_cancel.ts
+++ b/src/plugins/query_enhancements/server/routes/ppl_cancel.ts
@@ -30,7 +30,7 @@ export function registerPPLCancelRoute(router: IRouter, logger: Logger) {
           method: 'GET',
           path: '/_tasks',
           querystring: {
-            actions: 'cluster:admin/opensearch/ppl',
+            actions: '*ppl*',
             detailed: 'true',
           },
         });
--- a/src/plugins/query_enhancements/server/routes/ppl_cancel.test.ts
+++ b/src/plugins/query_enhancements/server/routes/ppl_cancel.test.ts
@@ -102,7 +102,7 @@ describe('registerPPLCancelRoute', () => {
     expect(transportRequest).toHaveBeenCalledWith({
       method: 'GET',
       path: '/_tasks',
-      querystring: { actions: 'cluster:admin/opensearch/ppl', detailed: 'true' },
+      querystring: { actions: '*ppl*', detailed: 'true' },
     });
     expect(transportRequest).toHaveBeenCalledWith({
       method: 'POST',


src/plugins/query_enhancements/common/utils.ts
```diff
@@ -89,20 +90,37 @@ export const fetch = (context: EnhancedFetchContext, query: Query, aggConfig?: Q
...
+                },
+              });
+            } catch (cancelError) {
+              // eslint-disable-next-line no-console
+              console.error('Failed to cancel async query:', cancelError);
+            }
+          } else if (context.body?.queryId) {
+            // Cancel synchronous PPL query via task cancellation
+            try {
+              await http.fetch({
```
joshuali925: is it better to not await this request? since frontend can't do much if cancel failed, it's job is to notify backend to cancel query, and it can move on. no need to block here in case PPL cancel API is stuck

src/plugins/query_enhancements/server/routes/ppl_cancel.ts
```diff
@@ -0,0 +1,99 @@
...
+        const client = dataSourceId
+          ? await context.dataSource.opensearch.getClient(dataSourceId)
+          : context.core.opensearch.client.asCurrentUser;
+
+        // List all PPL tasks with detailed descriptions
+        const tasksResponse = await client.transport.request({
+          method: 'GET',
+          path: '/_tasks',
+          querystring: {
+            actions: '*ppl*',
```
joshuali925: can it be `PPL *` or something more specific, like searching for queryId?

src/plugins/query_enhancements/common/utils.ts
ahkcs: Good call — updated to fire-and-forget. The cancel request now uses a .catch() chain instead of await, so the AbortError propagates immediately and the UI isn't blocked

src/plugins/query_enhancements/server/routes/ppl_cancel.ts
ahkcs: The _tasks API doesn't support filtering by description or `queryId` — it only supports `taskId, actions and parentTaskId`. The queryId is embedded in the task description string (e.g. PPL [queryId=<uuid>]: ...), so we have to list PPL tasks with detailed=true and match client-side. This is the only approach available without a backend change to expose a cancel-by-queryId endpoint.

mengweieric:
General question:
> {
  "error": {
    "reason": "Error occurred in OpenSearch engine: The task is cancelled.",
    "type": "TaskCancelledException"
  },
  "status": 500
}

Should the API return 4xx error code instead of 500? as this is not the backend error at all.

ahkcs:
> Should the API return 4xx error code instead of 500? as this is not the backend error at all.

Good call. I will open a follow-up PR on the backend side to change the classification of `TaskCancelledException` to 4xx

ahkcs:
https://github.com/opensearch-project/sql/pull/5273
Opened PR on the backend side to change the classification of `TaskCancelledException`
cc @mengweieric 

Updated error message:
```
server    log   [19:02:02.468] [error][plugins][queryEnhancements] pplSearchStrategy: {
  "error": {
    "reason": "Query cancelled",
    "details": "The task is cancelled.",
    "type": "TaskCancelledException"
  },
  "status": 400
```

ahkcs:
CI failure seems to be unrelated to our changes — this is an Explore app UI issue (autocomplete widget visibility)

Example:

https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/23563713788/job/68611665458

PR #11596, opened 2026-03-26
[Fix][Explore Vis] Fix scope history is out of scope for in-context editor

Qxisylolo:
### Description
This pr fixes:
1. refactor useInitialSaveExplore and init logic
2. fix: Error restoring state from URL
3. fix the race condition between previous visData and visConfig
4. rename In context editor to visualization editor
5. after save a save explore and query is not displayed when reload page

## Changelog
- fix: scope history is out of scope for in-context editor

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
- [ ]

### DIFF
--- a/src/plugins/explore/public/application/in_context_vis_editor/component/save_vis_button.tsx
+++ b/src/plugins/explore/public/application/in_context_vis_editor/component/save_vis_button.tsx
@@ -191,6 +191,7 @@ export const SaveVisButton = () => {
       const url =
         exploreId !== undefined && savedExplore ? `#/edit/${encodeURIComponent(exploreId)}` : '#/';
       services.scopedHistory?.push(url);
+      window.location.reload();
     }
   }, [stateTransfer, services.scopedHistory, exploreId, savedExplore, originatingApp]);
 
--- a/src/plugins/explore/public/application/in_context_vis_editor/hooks/use_query_panel_editor.ts
+++ b/src/plugins/explore/public/application/in_context_vis_editor/hooks/use_query_panel_editor.ts
@@ -58,7 +58,11 @@ export const useQueryPanelEditor = (): UseQueryPanelEditorReturnType => {
   const promptModeIsAvailableRef = useRef(promptModeIsAvailable);
   const queryLanguageRef = useRef(queryLanguage);
 
-  const { switchEditorMode, setEditorRef: setEditor } = useEditorOperations();
+  const {
+    switchEditorMode,
+    setEditorRef: setEditor,
+    setEditorText: setText,
+  } = useEditorOperations();
 
   // Keep the refs updated with latest context
   useEffect(() => {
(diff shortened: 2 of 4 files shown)


src/plugins/explore/public/components/visualizations/style_panel_render.tsx
```diff
@@ -22,24 +22,25 @@ import { VisData } from './visualization_builder.types';
 import { getAxisConfigByColumnMapping } from './utils/axis';
 
 interface StylePanelProps<T> {
-  data$: Observable<VisData | undefined>;
-  config$: Observable<RenderChartConfig | undefined>;
+  // data$: Observable<VisData | undefined>;
+  // config$: Observable<RenderChartConfig | undefined>;
```
ruanyl: Q: why update this? I see you combine the two observable

Qxisylolo: inside axes selector, there are two useEffect. one will receive new coming axes mapping and set state. the other will perform a updating-axes-mapping logic, that will eventually make visConfig$ emit a new value.

This is where the problem comes, when new axes mapping and data coming, the second effect will run and write the old axes mapping back which will cause mismatching between old mapping and new data and  trigger error when generating spec.

combining 2 observables together may not be a good idea, what I want to do is adding a debounce time to ensure that component only receives the final settled axes mapping and use them to create vis spec.

ruanyl: How to reproduce this issue?

Qxisylolo: Will add a validation to axes selector which seems to be more straightforward

Qxisylolo: To reproduce the issue:
1. input query,  for example `source=opensearch_dashboards_sample_data_flights| stats count() as num_flights by `OriginCityName` `
2. just replace the categorical field, for example from OriginCityName to Carrier
3. then error will be displayed

截屏2026-03-27 10 09 35

src/plugins/explore/public/components/visualizations/style_panel_render.tsx
```diff
@@ -24,6 +24,7 @@ import { getAxisConfigByColumnMapping } from './utils/axis';
 interface StylePanelProps<T> {
   data$: Observable<VisData | undefined>;
   config$: Observable<RenderChartConfig | undefined>;
+
```
ruanyl:
```suggestion
```

ruanyl (approved):

src/plugins/explore/public/components/visualizations/style_panel_render.tsx
Qxisylolo: will fix it in the next pr

src/plugins/explore/public/components/visualizations/metric/to_expression.ts
```diff
@@ -34,10 +34,6 @@ export const createSingleMetric = (
     throw Error('Missing value for metric chart');
   }
 
-  if (!numericField) {
-    throw Error('Missing value for metric chart');
-  }
-
   if (!dateField) {
```
Hailong-am, line 37: is this removal by intention?

Qxisylolo: yes, this statement is duplicate.

PR #11631, opened 2026-03-31
[Workspace] Allow custom ID to be specified on workspace creation

nicklas-dohrn:
### Description

Adds an optional `id` field to the workspace creation flow — both via the API and the creation UI.

- **API**: `POST /api/workspaces` now accepts `attributes.id`. When provided, it is used as the workspace ID instead of the auto-generated 6-char base64url value. Must be a UUID or 6–20 alphanumeric/`_`/`-` characters. Returns an error if the ID is already taken.
- **UI**: A new optional "Workspace ID" field is shown on the creator form with a tooltip explaining the allowed formats. Validation mirrors the server-side rules.

### Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/12626

## Screenshot of the UI
Screenshot 2026-06-29 at 17 37 01

## Related documentation:

- https://github.com/opensearch-project/documentation-website/pull/12713

## Testing the changes

- [x] Creates workspace with valid custom ID via UI and API
- [x] Creates workspace without custom ID (falls back to auto-generated)
- [x] Returns error when workspace with provided ID already exists
- [x] Returns 400 for invalid ID (too short, too long, invalid chars)

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


KarstenSchnitter:
Thanks for providing this PR. The rational is to allow provisioning of pre-defined workspaces with deterministic ids. This allows to create links to the workspace contents in external systems without scanning OpenSearch for the randomly generated workspace id.

src/plugins/workspace/server/routes/index.ts
```diff
@@ -102,6 +102,19 @@ const workspaceNameSchema = schema.string({
 });
 
 const createWorkspaceAttributesSchema = schema.object({
+  id: schema.maybe(
```
SuZhou-Joe, line 111: Thanks for the PR, but I'd recommend we using `PUT /api/workspaces/${workspaceId}` to better align with REST API pattern.

The API returns 201 Created for newly-created workspace while 200 / 409 for updating workspace.

nicklas-dohrn: I will rework this, to give the correct responses.

changelogs/fragments/11631.yml
Hailong-am, line 1: we don't this change log now. let's. remove this file

src/plugins/workspace/server/routes/index.ts
nicklas-dohrn: Should be good now.

nicklas-dohrn:
@SuZhou-Joe and @Hailong-am,
I would really appreciate another look here, Added both docs and UI changes if wanted,
all is tested and working.
If we only want to have the api change, that would be fine for me, just thought this would be nice as full integration.

ruanyl:
Hey @nicklas-dohrn, thanks for the PR, I rebased with `main` branch and resolved the conflicts. Also I update the code with the following changes:

1. Create the workspace before data-source associations. Prevents duplicate-ID races from associating data with the wrong workspace.
2. Treat association failures as partial success. The workspace remains successfully created, failed associations are returned by the API, and the success toast tells users they can add them later.
3. Centralize workspace ID validation. UI and server now share one rule: 6–36 letters, numbers, underscores, or hyphens.
4. More test coverage

Hailong-am:
## Code Review Findings

All findings are introduced by commit `faf60b3f` ("Report partial association failures").

---

### 🐛 Bug — Trace detection uses unfiltered data source list after partial association failure

**File:** `src/plugins/workspace/public/components/workspace_creator/workspace_creator.tsx` ~line 176

`dataSourceConnections` is built from the original `selectedDataSourceConnections` list, not filtered to only the successfully associated subset. As a result, `detectTraceData` / `createAutoDetectedDatasets` run inside the new workspace context for data sources that were never actually attached to it, creating broken/orphaned index patterns that silently fail when the user tries to use them.

**Fix:** before building `dataSourceConnections`, filter `selectedDataSourceConnections` to exclude any entry whose connection ID appears in `failedAssociations`.

---

### 🐛 Bug — Workspace survives with zero data sources when all associations fail

**File:** `src/plugins/workspace/server/workspace_client.ts` ~line 187

Switching to `Promise.allSettled` is the right call for partial failures, but there is no floor: if *every* requested association fails, the server still returns `success: true` with a workspace ID. The workspace is permanently created with no data sources and no automated cleanup path. The client shows a failure-count toast but has no way to delete or repair the stranded workspace.

**Fix:** consider returning an error (or deleting the newly created workspace) when `failedAssociations.length === total requested`, or surface a distinct error code so the client can offer a retry/cleanup path.

---

### Minor notes

- **`Error | any` collapses to `any`** (`workspace_client.ts:110`): `Error | any` is equivalent to `any`. Both `getErrorMessage` and `formatError` should use `unknown` and narrow before accessing `.message` / `.error`.
- **ID constraint string duplicated 3×**: the "6–36 characters, letters, numbers, underscores, hyphens" description appears independently in the form tooltip, the client-side validation error, and the server-side route error. A shared constant would prevent drift if the regex changes.
- **Double `flatMap` over `associationResults`** (lines 190–208): `failedAssociations` and `associatedDataSources` are derived in two separate passes; a single `reduce`/`forEach` would suffice.

ruanyl:
> ## Code Review Findings
> All findings are introduced by commit `faf60b3f` ("Report partial association failures").
> 
> ### 🐛 Bug — Trace detection uses unfiltered data source list after partial association failure
> **File:** `src/plugins/workspace/public/components/workspace_creator/workspace_creator.tsx` ~line 176
> 
> `dataSourceConnections` is built from the original `selectedDataSourceConnections` list, not filtered to only the successfully associated subset. As a result, `detectTraceData` / `createAutoDetectedDatasets` run inside the new workspace context for data sources that were never actually attached to it, creating broken/orphaned index patterns that silently fail when the user tries to use them.
> 
> **Fix:** before building `dataSourceConnections`, filter `selectedDataSourceConnections` to exclude any entry whose connection ID appears in `failedAssociations`.
> 
> ### 🐛 Bug — Workspace survives with zero data sources when all associations fail
> **File:** `src/plugins/workspace/server/workspace_client.ts` ~line 187
> 
> Switching to `Promise.allSettled` is the right call for partial failures, but there is no floor: if _every_ requested association fails, the server still returns `success: true` with a workspace ID. The workspace is permanently created with no data sources and no automated cleanup path. The client shows a failure-count toast but has no way to delete or repair the stranded workspace.
> 
> **Fix:** consider returning an error (or deleting the newly created workspace) when `failedAssociations.length === total requested`, or surface a distinct error code so the client can offer a retry/cleanup path.
> 
> ### Minor notes
> * **`Error | any` collapses to `any`** (`workspace_client.ts:110`): `Error | any` is equivalent to `any`. Both `getErrorMessage` and `formatError` should use `unknown` and narrow before accessing `.message` / `.error`.
> * **ID constraint string duplicated 3×**: the "6–36 characters, letters, numbers, underscores, hyphens" description appears independently in the form tooltip, the client-side validation error, and the server-side route error. A shared constant would prevent drift if the regex changes.
> * **Double `flatMap` over `associationResults`** (lines 190–208): `failedAssociations` and `associatedDataSources` are derived in two separate passes; a single `reduce`/`forEach` would suffice.

@Hailong-am thanks!
For "🐛 Bug — Trace detection uses unfiltered data source list after partial association failure", I will fix it.

For "🐛 Bug — Workspace survives with zero data sources when all associations fail", I actually changed this intentionally, we allow to create workspace without data source initially associated. For the existing flow, it associate first, then create workspace, if workspace create failed, data source will be associated with invalid workspace id. This PR changed to create workspace first, then associate data source. Failed data association won't break workspace creation, user can always associate data source in workspace setting page afterwards.

Hailong-am:
> Failed data association won't break workspace creation, user can always associate data source in workspace setting page afterwards.

make sense, this is much better to avoid datasource with invalid workspace.

SuZhou-Joe (dismissed):

ruanyl:
> ID constraint string duplicated 3×

The regex is already shared. Tooltip, client error, and server error have different presentation needs. Sharing a format-description constant could help, but this is not blocking.

> Double flatMap

A single loop would be clearer, but the current implementation is correct, I'd keep it as it is.

Hailong-am (approved):

ruanyl (approved):

nicklas-dohrn:
Thank you so much @ruanyl for taking this over and making it fit.
Will try out your changes this week hopefully, to see if this now works with all our requirements, but honestly,
I do not see why this should not be the case.
Greetings,
Nicklas

nicklas-dohrn:
one remark:
I also created documentation for this feature,
to not be a bad contributor by just pushing half cooked things over the fence:
https://github.com/opensearch-project/documentation-website/pull/12713
hope this helps.

ruanyl:
> one remark: I also created documentation for this feature, to not be a bad contributor by just pushing half cooked things over the fence: [opensearch-project/documentation-website#12713](https://github.com/opensearch-project/documentation-website/pull/12713) hope this helps.

@nicklas-dohrn The current implementation enforce 6-32 valid characters, UUID is not validated, perhaps you need to update the documentation PR as well, let me know if that makes sense, thanks :)

PR #11633, opened 2026-03-31
[Fix][Explore vis] Error when faceting data into one group

Qxisylolo:
### Description

Root cause:
When facetTransform results in only a single facet group (facetNumbers <= 1), it fell back to calling `transform(...fns)(state)` , returning transformedData in the same shape as a non-faceted transform result. but createFacetLineSeries always assumes the multi-facet shape, so it iterated over wrong data.

To reproduce:

1. source=opensearch_dashboards_sample_data_flights| where FlightDelay = true 
2. set facet field to FlightDelay

## Screenshot

before

截屏2026-03-31 15 14 36

after

截屏2026-03-31 15 15 10

## Changelog
- fix: Error when faceting data into one group

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/utils/data_transformation/transform.test.ts
+++ b/src/plugins/explore/public/components/visualizations/utils/data_transformation/transform.test.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { transform } from './transform';
+import { transform, facetTransform } from './transform';
 
 describe('transform', () => {
   it('applies single transformation function', () => {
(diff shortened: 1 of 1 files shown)


src/plugins/explore/public/components/visualizations/utils/data_transformation/transform.ts
```diff
@@ -45,7 +45,10 @@ export const facetTransform = (facetColumn: string, ...fns: TransformFn[]) => (
 
   const facetNumbers = Object.keys(grouped).length;
 
-  if (facetNumbers <= 1) return transform(...fns)(state);
+  if (facetNumbers <= 1) {
+    const result = transform(...fns)(state);
+    return { ...result, transformedData: [result.transformedData] };
```
SuZhou-Joe, line 50: Does it make sense to add UT for this branch?

Qxisylolo: ok, I will add some

SuZhou-Joe: Not quite familiar with the data flow, but if facetNumbers are 0, is it expected that we return something like 
```
{
  ...result,
  transformedData: [undefined]
}
```

Qxisylolo: in my understanding, facetNumbers cannot be zero, since it is a data field you chose in your query. 
facetNumbers is 0 indicates that there is no search result, in such case, visualization won't be generated.

PR #11635, opened 2026-03-31
feat: introduce query-based tab auto switching to discover

ruanyl:
### Description
This PR introduced a new "Statistic" tab for dedicated tabular query results.

#### Tab Selection Rules
When a user runs a query, the tab behavior depends on two factors: the currently selected tab and the query content.

#### Rule 1 — Current tab is Logs (or no manual selection)

Logs is the default tab. When the user is on Logs (whether by default or manual selection), the system applies query-based auto-selection:

|Query Pattern	|Result	|
|---	|---	|
|Contains `stats` or `table`	|→ Switch to **Statistic**	|
|Contains `chart` or `timechart`	|→ Switch to **Visualization**	|
|None of the above	|→ Stay on **Logs**	|

#### Rule 2 — Current tab is Statistic

The user has expressed interest in tabular/aggregated data. Respect that intent:

|Query Pattern	|Result	|
|---	|---	|
|Contains `stats` or `table`	|→ Stay on **Statistic**	|
|Contains `chart` or `timechart`	|→ Stay on **Statistic** (interchangeable)	|
|None of the above	|→ Switch to **Logs**	|

#### Rule 3 — Current tab is Visualization

The user has expressed interest in visual/analytical data. Respect that intent:

|Query Pattern	|Result	|
|---	|---	|
|Contains `stats` or `table`	|→ Stay on **Visualization** (interchangeable)	|
|Contains `chart` or `timechart`	|→ Stay on **Visualization**	|
|None of the above	|→ Stay on **Visualization** (data-driven viz is valid without explicit commands)	|

### Decision Matrix

|Current Tab ↓ \ Query →	|stats / table	|chart / timechart	|Other cases	|
|---	|---	|---	|---	|
|**Logs**	|→ Statistic	|→ Visualization	|→ Logs (stay)	|
|**Statistic**	|→ Statistic (stay)	|→ Statistic (stay)	|→ Logs	|
|**Visualization**	|→ Visualization (stay)	|→ Visualization (stay)	|→ Visualization (stay)	|

## Screenshot

https://github.com/user-attachments/assets/c52f96cd-f112-4267-953a-e9acc166a290

https://github.com/user-attachments/assets/7359cdfb-1d9a-4525-a450-f72808517819

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/23/auto_tab_detect.spec.js
+++ b/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/23/auto_tab_detect.spec.js
@@ -1,0 +1,307 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+  DATASOURCE_NAME,
+  INDEX_WITH_TIME_1,
+  START_TIME,
+  END_TIME,
+} from '../../../../../../utils/apps/constants';
+import {
+  getRandomizedWorkspaceName,
+  getRandomizedDatasetId,
+} from '../../../../../../utils/apps/explore/shared';
+import {
+  prepareTestSuite,
+  createWorkspaceAndDatasetUsingEndpoint,
+} from '../../../../../../utils/helpers';
+
+const workspace = getRandomizedWorkspaceName();
+const datasetId = getRandomizedDatasetId();
+const indexPattern = `${INDEX_WITH_TIME_1}*`;
+
+const autoTabDetectTestSuite = () => {
+  describe('auto tab detection based on query', { scrollBehavior: false }, () => {
+    before(() => {
+      cy.osd.setupEnvAndGetDataSource(DATASOURCE_NAME);
+
+      createWorkspaceAndDatasetUsingEndpoint(
+        DATASOURCE_NAME,
+        workspace,
+        datasetId,
+        indexPattern,
+        'timestamp',
+        'logs',
+        ['use-case-observability']
+      );
+    });
+
+    beforeEach(() => {
+      cy.osd.navigateToWorkSpaceSpecificPage({
+        workspaceName: workspace,
+        page: 'explore/logs',
+        isEnhancement: true,
+      });
+
+      cy.explore.setDataset(indexPattern, DATASOURCE_NAME, 'INDEX_PATTERN');
+      cy.explore.setTopNavDate(START_TIME, END_TIME);
+      cy.osd.waitForLoader(true);
+    });
+
+    after(() => {
+      cy.osd.cleanupWorkspaceAndDataSourceAndIndices(workspace, [INDEX_WITH_TIME_1]);
+    });
+
+    // =========================================================================
+    // Rule 1: Current tab is Logs (default)
+    // =========================================================================
+    describe('Rule 1 - current tab is Logs (default)', () => {
+      it('switches to Statistics tab when query contains stats command', () => {
+        const statsQuery = `source = ${indexPattern} | stats count() by category`;
+        cy.explore.setQueryEditor(statsQuery);
+        cy.osd.waitForLoader(true);
+
+        cy.getElementByTestId('exploreTab-explore_statistics')
+          .should('be.visible')
+          .and('have.attr', 'aria-selected', 'true');
+      });
+
+      it('switches to Statistics tab when query contains table command', () => {
+        const tableQuery = `source = ${indexPattern} | table category, bytes_transferred`;
+        cy.explore.setQueryEditor(tableQuery);
+        cy.osd.waitForLoader(true);
+
+        cy.getElementByTestId('exploreTab-explore_statistics')
+          .should('be.visible')
+          .and('have.attr', 'aria-selected', 'true');
+      });
+
+      it('switches to Visualization tab when query contains chart command', () => {
+        const chartQuery = `source = ${indexPattern} | chart count() by category`;
+        cy.explore.setQueryEditor(chartQuery);
+        cy.osd.waitForLoader(true);
+
+        cy.getElementByTestId('exploreTab-explore_visualization_tab')
+          .should('be.visible')
+          .and('have.attr', 'aria-selected', 'true');
+      });
+
+      it('switches to Visualization tab when query contains timechart command', () => {
+        const timechartQuery = `source = ${indexPattern} | timechart count()`;
+        cy.explore.setQueryEditor(timechartQuery);
+        cy.osd.waitForLoader(true);
+
+        cy.getElementByTestId('exploreTab-explore_visualization_tab')
+          .should('be.visible')
+          .and('have.attr', 'aria-selected', 'true');
+      });
+
+      it('remains on Logs tab when query has no special command', () => {
+        const plainQuery = `source = ${indexPattern} | where bytes_transferred > 1000`;
+        cy.explore.setQueryEditor(plainQuery);
+        cy.osd.waitForLoader(true);
+
+        cy.getElementByTestId('exploreTab-logs')
+          .should('be.visible')
+          .and('have.attr', 'aria-selected', 'true');
+      });
+    });
+
+    describe('tab preservation on same-query refresh', () => {
+      it('preserves Visualization tab when re-running the same query', () => {
+        const statsQuery = `source = ${indexPattern} | stats count() by category`;
+        cy.explore.setQueryEditor(statsQuery);
+        cy.osd.waitForLoader(true);
+
+        // Auto-detected to Statistics tab
+        cy.getElementByTestId('exploreTab-explore_statistics')
+          .should('be.visible')
+          .and('have.attr', 'aria-selected', 'true');
+
+        // Manually switch to Visualization tab
+        cy.getElementByTestId('exploreTab-explore_visualization_tab').click();
+        cy.wait(500);
+
+        // Re-run the same query via the execution button (time-only refresh)
+        cy.getElementByTestId('exploreQueryExecutionButton').click({ force: true });
+        cy.osd.waitForLoader(true);
+
+        // Visualization tab should remain selected (same query = no tab switch)
+        cy.getElementByTestId('exploreTab-explore_visualization_tab')
+          .should('be.visible')
+          .and('have.attr', 'aria-selected', 'true');
+      });
+    });
+
+    describe('Statistics tab hides action bar buttons', () => {
+      it('hides Add to Dashboard and Export CSV buttons on Statistics tab', () => {
+        const statsQuery = `source = ${indexPattern} | stats count() by category`;
+        cy.explore.setQueryEditor(statsQuery);
+        cy.osd.waitForLoader(true);
+
+        // Verify we are on Statistics tab
+        cy.getElementByTestId('exploreTab-explore_statistics')
+          .should('be.visible')
+          .and('have.attr', 'aria-selected', 'true');
+
+        // Add to Dashboard button should not be visible
+        cy.get('[data-test-subj="dscResultsActionBar"]').within(() => {
+          cy.get('[data-test-subj="addToDashboardButton"]').should('not.exist');
+          cy.get('[data-test-subj="downloadCsvButton"]').should('not.exist');
+        });
+      });
+    });
+
+    describe('Statistics tab renders table with results', () => {
+      it('displays a table with data when stats query is executed', () => {
+        const statsQuery = `source = ${indexPattern} | stats count() by category`;
+        cy.explore.setQueryEditor(statsQuery);
+        cy.osd.waitForLoader(true);
+
+        // Verify Statistics tab is selected
+        cy.getElementByTestId('exploreTab-explore_statistics')
+          .should('be.visible')
+          .and('have.attr', 'aria-selected', 'true');
+
+        // Verify the statistics table is rendered with data
+        cy.get('.exploreStatisticTable').should('be.visible');
+        cy.get('.exploreStatisticTable table').should('exist');
+        cy.get('.exploreStatisticTable tbody tr').should('have.length.greaterThan', 0);
+      });
+    });
+
+    // =========================================================================
+    // Rule 2: Current tab is Statistic
+    // =========================================================================
+    describe('Rule 2 - current tab is Statistic', () => {
+      it('stays on Statistics when query contains stats (stats/table → stay)', () => {
+        // Get to Statistics tab via a stats query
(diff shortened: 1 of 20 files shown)


seraphjiang:
could we have 

notification badge (a small colored dot) showing you have something new you haven’t checked yet.

we always stay in first tab(log)

if other tab has info worth cx to read 

make it bold or use dot/number badge

ruanyl:
> could we have
> 
> notification badge (a small colored dot) showing you have something new you haven’t checked yet.
> 
> we always stay in first tab(log)
> 
> if other tab has info worth cx to read
> 
> make it bold or use dot/number badge

The tab content always determined by the current query, there seems no "notification" need to info cx based on my understanding, not sure if I fully get it.

anirudha:
We are missing field stats and patterns tab; That makes it too many tabs. Earlier we decided to fold stats into viz earlier with a viz table. Then table got changed to viz not a table. 


Let's enable those 2 tabs and test the experience with all tabs.

anirudha:
without auto switch/ the aggregation ppl query will show a only logs and no aggregation results. - this is bad / a bug as we should incorrect results.

a better option would be to add a new stats/pivot table view and not force viz.

Hailong-am:
what about PPL command contains table and chart like  
```
| table `time` `resource.attributes.service.name` | stats count() by span(time, 1d) | chart count()
```
, will it render as a chart?

ruanyl:
> what about PPL command contains table and chart like
> 
> ```
> | table `time` `resource.attributes.service.name` | stats count() by span(time, 1d) | chart count()
> ```
> 
> , will it render as a chart?

No, it will be `Statistic` tab, it checks `| table` first, what do you think?

Hailong-am:
> > what about PPL command contains table and chart like
> > ```
> > | table `time` `resource.attributes.service.name` | stats count() by span(time, 1d) | chart count()
> > ```    
> >   
> > , will it render as a chart?
> 
> No, it will be `Statistic` tab, it checks `| table` first, what do you think?

I'm OK with it as long as it doesn't show Logs. As long as the query have aggregation, show either `Statistic` or `Visualization` both works

Hailong-am (dismissed):

ps48 (commented):
### Review Summary

The core logic (`detect_optimal_tab.ts`) is well-structured with a clear decision matrix and thorough unit tests. Nice addition of the Statistics tab. Main concerns:

| # | Issue | Severity | Type |
|---|-------|----------|------|
| 1 | `selectActiveTab` vs `selectActiveTabId` naming | Nit | Readability |
| 2 | Regex missing word boundary — false positives on `stats_count` etc. | Medium | Bug |
| 3 | `activeTabId` in effect deps — fragile re-fire coupling | Medium | Design |
| 4 | Missing `await` + ordering asymmetry in reset path | Low-Medium | Design |
| 5 | String literal `'explore_patterns_tab'` should use constant | Nit | Consistency |
| 6 | `setActiveTab('')` uses implicit sentinel value | Nit | Readability |
| 7 | Cypress `cy.wait(1000)` instead of `cy.intercept()` | Medium | Testing |
| 8 | Portal target via `document.getElementById` in useEffect | Low | Design |
| 9 | Index-based row IDs in statistics table | Low | Design |
| 10 | Magic number `100` for page size | Nit | Readability |

CI note: The `ciGroup10Explore` failure is an infrastructure issue (permission denied on the runner), unrelated to this PR — the spec glob (`explore/10/*.spec.js`) doesn't match the new test files (`explore/23/`).

src/plugins/explore/public/application/utils/hooks/use_tab_results.ts
```diff
@@ -19,7 +19,7 @@ import { resultsCache } from '../state_management/slices';
 export const useTabResults = () => {
   const { services } = useOpenSearchDashboards<ExploreServices>();
   const query = useSelector((state: RootState) => state.query);
-  const activeTabId = useSelector((state: RootState) => state.ui.activeTabId);
+  const activeTabId = useSelector(selectActiveTab);
```
ps48: **Severity: Nit**

`selectActiveTab` and `selectActiveTabId` are functionally identical selectors — both return `string`. But the variable is named `activeTabId` while the selector is `selectActiveTab`, which misleadingly suggests it returns a tab object. Use `selectActiveTabId` to match the variable name (and update the import on line 12 accordingly).

```suggestion
  const activeTabId = useSelector(selectActiveTabId);
```

src/plugins/explore/public/application/utils/state_management/actions/detect_optimal_tab.ts
```diff
@@ -0,0 +1,70 @@
...
+import { setActiveTab } from '../slices';
+import { ExploreServices } from '../../../../types';
+import {
+  EXPLORE_LOGS_TAB_ID,
+  EXPLORE_STATISTICS_TAB_ID,
+  EXPLORE_VISUALIZATION_TAB_ID,
+} from '../../../../../common';
+
+const hasStatsOrTable = (q: string) => /\|\s*(stats|table)/i.test(q);
+const hasChartOrTimechart = (q: string) => /\|\s*(chart|timechart)/i.test(q);
```
ps48: **Severity: Medium — Potential Bug**

These regexes match partial command names. For example, `| stats_count()` would match `stats` even though `stats_count` isn't a PPL command. Adding a word boundary `\b` after each keyword prevents false positives on creative field/function names.

```suggestion
const hasStatsOrTable = (q: string) => /\|\s*(stats|table)\b/i.test(q);
const hasChartOrTimechart = (q: string) => /\|\s*(chart|timechart)\b/i.test(q);
```

src/plugins/explore/public/application/utils/hooks/use_initial_query_execution.ts
```diff
@@ -65,6 +71,7 @@ export const useInitialQueryExecution = (services: ExploreServices) => {
   }, [
     isInitialized,
     queryState,
+    activeTabId,
```
ps48: **Severity: Medium — Design**

Adding `activeTabId` to the dependency array means the effect re-fires whenever `detectAndSetOptimalTab` changes the tab. Today the `!isInitialized` guard prevents double execution — the effect dispatches `setIsInitialized(true)` in the same synchronous batch, so the re-fire is a no-op.

However, this coupling is fragile. If an `await` is ever introduced between `detectAndSetOptimalTab` and `setIsInitialized(true)`, React could re-render between the two dispatches, and the effect would re-run before initialization is marked complete — causing a double query execution.

Since `activeTabId` is only read inside a `!activeTabId` guard (i.e., "detect tab only on first load when no tab is selected"), consider using a ref instead to avoid the dependency:

```ts
const activeTabIdRef = useRef(activeTabId);
activeTabIdRef.current = activeTabId;
// Then inside effect: if (!activeTabIdRef.current) { ... }
```

This satisfies the intent without adding `activeTabId` to the dep array.

src/plugins/explore/public/application/utils/state_management/actions/reset_explore_state/reset_explore_state.ts
```diff
@@ -35,4 +36,5 @@ export const resetExploreStateActionCreator = (
   dispatch(setQueryState(state.query));
   dispatch(setQueryEditorState(state.queryEditor));
   await dispatch(executeQueries({ services }));
+  dispatch(detectAndSetOptimalTab({ services }));
```
ps48: **Severity: Low-Medium — Design**

Two concerns here:

1. **Missing `await`**: `detectAndSetOptimalTab` is an async thunk. Without `await`, the caller can't know when tab detection finishes. In `run_query.ts` the same call is awaited — this should be consistent.

2. **Ordering difference**: In `run_query.ts`, tab detection runs *before* `executeQueries`. Here it runs *after*. If `executeQueries` uses `activeTabId` to decide which tab-specific query to execute, this ordering means the query runs with a reset/empty `activeTabId` and the tab switches afterward. Is this intentional? A comment explaining the rationale would help.

```suggestion
  await dispatch(detectAndSetOptimalTab({ services }));
```

src/plugins/explore/public/components/tabs/action_bar/results_action_bar/results_action_bar.tsx
```diff
@@ -52,8 +52,10 @@ export const DiscoverResultsActionBar = ({
   const currentTab = useSelector(selectActiveTabId);
   const wrapCellText = useSelector(selectWrapCellText);
   const isLogsTab = currentTab === EXPLORE_LOGS_TAB_ID;
-  const shouldShowAddToDashboardButton = currentTab !== 'explore_patterns_tab';
-  const shouldShowExportButton = currentTab !== 'explore_patterns_tab';
+  const shouldShowAddToDashboardButton =
+    currentTab !== 'explore_patterns_tab' && currentTab !== EXPLORE_STATISTICS_TAB_ID;
+  const shouldShowExportButton =
+    currentTab !== 'explore_patterns_tab' && currentTab !== EXPLORE_STATISTICS_TAB_ID;
```
ps48: **Severity: Nit — Consistency**

The string literal `'explore_patterns_tab'` is used here alongside the constant `EXPLORE_STATISTICS_TAB_ID`. Since `EXPLORE_PATTERNS_TAB_ID` is already exported from `common/index.ts`, use it for consistency (and to prevent silent breakage if the ID ever changes). Note: `EXPLORE_PATTERNS_TAB_ID` would need to be added to the import on line 23.

The same applies to `showTabSpecificSettings` on line 59.

```suggestion
  const shouldShowAddToDashboardButton =
    currentTab !== EXPLORE_PATTERNS_TAB_ID && currentTab !== EXPLORE_STATISTICS_TAB_ID;
  const shouldShowExportButton =
    currentTab !== EXPLORE_PATTERNS_TAB_ID && currentTab !== EXPLORE_STATISTICS_TAB_ID;
```

src/plugins/explore/public/application/utils/state_management/middleware/dataset_change_middleware.ts
```diff
@@ -51,6 +52,7 @@ export const createDatasetChangeMiddleware = (
 
       currentDataset = dataset;
 
+      store.dispatch(setActiveTab(''));
```
ps48: **Severity: Nit — Readability**

Resetting `activeTabId` to `''` works because `detect_optimal_tab.ts` does `state.ui.activeTabId || EXPLORE_LOGS_TAB_ID`, treating the empty string as "no tab selected". But this is implicit — a named constant (e.g. `const NO_TAB_SELECTED = ''`) would make the intent clearer and prevent someone from accidentally changing the fallback behavior.

cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/23/auto_tab_detect.spec.js
```diff
@@ -0,0 +1,328 @@
...
+        const statsQuery = `source = ${indexPattern} | stats count() by category`;
+        cy.explore.setQueryEditor(statsQuery);
+        cy.osd.waitForLoader(true);
+        cy.wait(1000);
+
+        cy.getElementByTestId('exploreTab-explore_statistics')
+          .should('be.visible')
+          .and('have.attr', 'aria-selected', 'true');
+      });
+
```
ps48, line 70: **Severity: Medium — Testing**

`cy.wait(1000)` appears ~20 times across the two new test files. Per Cypress best practices (and the repo's own conventions), `cy.intercept()` should be preferred over hard-coded delays. These waits are brittle (may fail in slow CI environments or pass spuriously) and add ~20s of wall-clock time to the suite.

Consider intercepting the search/query API call and waiting on it instead:

```js
cy.intercept('POST', '**/search').as('searchRequest');
cy.explore.setQueryEditor(statsQuery);
cy.osd.waitForLoader(true);
cy.wait('@searchRequest');
```

This applies throughout both `auto_tab_detect.spec.js` and `statistics_tab.spec.js`.

src/plugins/explore/public/components/tabs/statistics_tab.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const [itemIdToExpandedRowMap, setItemIdToExpandedRowMap] = useState<Record<string, JSX.Element>>(
+    {}
+  );
+  const [slot, setSlot] = useState<HTMLElement | null>(null);
+  const [pageIndex, setPageIndex] = useState(0);
+  const { results } = useTabResults();
+
+  useEffect(() => {
+    setSlot(document.getElementById(EXPLORE_ACTION_BAR_SLOT_ID));
+  }, []);
```
ps48, line 32: **Severity: Low — Design**

This runs once on mount with `[]` deps. If the portal target element doesn't exist yet at mount time (e.g., due to render ordering), `slot` will be `null` and the action bar portal will silently fail. If this matches the pattern used by other tabs in the codebase, it's fine to keep. Otherwise, consider a retry or `MutationObserver` approach to handle late-rendering targets.

src/plugins/explore/public/components/tabs/statistics_tab.tsx
```diff
@@ -0,0 +1,136 @@
...
+    if (results) {
+      const hits = results.hits?.hits || [];
+      const schema = results.fieldSchema || [];
+
+      const transformedData = hits.map((row, i) => {
+        const transformedRow: Record<string, any> = {};
+        transformedRow.id = i;
+        for (const { name } of schema) {
+          const source = row._source as Record<string, any>;
+          if (name) {
```
ps48, line 97: **Severity: Low — Design**

Using the array index as `id` means `itemIdToExpandedRowMap` entries will point to the wrong row if results are re-ordered or paginated differently (same index → different data). For a stats result set that's stable until the next query execution, this is fine today. But if sorting or filtering is ever added to this table, expanded rows will silently reference stale data.

Consider deriving a stable key from the row content (e.g., a hash of field values) if this table may gain interactive sorting later.

src/plugins/explore/public/components/tabs/statistics_tab.tsx
```diff
@@ -0,0 +1,136 @@
...
+      return transformedData;
+    }
+    return [];
+  }, [results]);
+
+  const pagination = useMemo(
+    () => ({
+      pageIndex,
+      pageSize: 100,
+      totalItemCount: rows.length,
```
ps48, line 112: **Severity: Nit — Readability**

The page size `100` appears three times (here, and twice on line 115 in the `pageOfItems` slice). Extract to a constant to keep them in sync:

```ts
const STATISTICS_PAGE_SIZE = 100;
```

ps48 (approved):

Hailong-am (approved):

PR #11638, opened 2026-03-31
chore(deps): reduce node_modules size by removing duplicates from yarn.lock

### LINKED ISSUE #11637: Cleans up yarn.lock by removing duplicates
Because yarn v1 is used in this project, after some time We have many duplicate packages currently.

A duplicate package is when two dependencies are resolved to a different version, even when a single version matches the range specified in the dependencies

The idea is to use tools e.g. yarn-deduplicate to fix this issue, and adjust if need it.

I was able to reduce node_modules size for 1.86 GB to 1.69 GB, ~170MB

I will open PR soon for this

tomkdgun:
### Description

Because yarn v1 is used in this project, We have many duplicate packages currently.

A duplicate package is when two dependencies are resolved to a different version, even when a single version matches the range specified in the dependencies

I was able to reduce node_modules size for 1.86 GB to 1.69 GB, ~170MB

Most of duplicates were around @babel, @parcel/watcher, ws

The root cause of adjustment in unit test is that @hapi/mimos@6.0.0 now utilise mime-db@1.54.0 vs mime-db@1.52.0. Since `text/javascript` is standard vs obsolete `application/javascript`, starting from mime-db@1.53.0 the text/javascript is returned for `.js` and `.mjs`. Changelog: 
https://github.com/jshttp/mime-db/blob/master/HISTORY.md#1530--2024-07-12

### Issues Resolved

Closes #11637

## Changelog

- skip

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/cypress/utils/apps/explore/autocomplete.js
+++ b/cypress/utils/apps/explore/autocomplete.js
@@ -33,8 +33,8 @@ export const typeAndVerifySuggestion = (input, expectedSuggestion) => {
   if (input) {
     cy.get('.inputarea').type(input, { force: true });
   }
-  // Wait for resizable button to be in stable position
-  cy.get('[data-test-subj="euiResizableButton"]').should('exist');
+  // Wait for UI to settle and suggestion widget to be fully rendered
+  cy.wait(100);
   cy.get('.suggest-widget')
     .should('be.visible')
     .within(() => {


tomkdgun:
@angle943 Is there any reason integration tests are failing in PR pipeline, while they are passing on my local environment ?

src/optimize/bundles_route/bundles_route.test.ts
```diff
@@ -138,10 +138,7 @@ describe('js file', () => {
 
     expect(response.statusCode).toBe(200);
     expect(response.headers).not.toHaveProperty('content-length');
-    expect(response.headers).toHaveProperty(
-      'content-type',
-      'application/javascript; charset=utf-8'
-    );
+    expect(response.headers).toHaveProperty('content-type', 'text/javascript; charset=utf-8');
```
ruanyl, line 141: Hi @tomkdgun, thanks for the PR! Is this change intentionally? seems unrelated

tomkdgun: Hi @ruanyl The pull request description contains the reason why this was adjusted. It's expected.

ruanyl: I wasn't expecting behavior changes by removing duplicate dependencies. I love these cleanups, just a bit concern would there be other side effects like this? btw, how did you do the deductions? any tools used?

tomkdgun: @ruanyl Base on this (screenshot below) + changelog check. I can potentially to previous behaviour, but `text/javascript` is I think correct value according to standard
image

Issue #11550, opened 2026-03-19
[RFC] Dashboard Variables for OpenSearch Dashboards


yubonluo:
## Abstract

This RFC proposes a Dashboard-level Variables system for OpenSearch Dashboards that enables users to define reusable parameters and dynamically inject them into queries through explicit template syntax. Variables eliminate the need for dashboard duplication when users need to modify query structure (field names, aggregation functions, dimensions, or metrics), enabling template-based dashboards where a single configuration can serve multiple business contexts.

### POC
dashboard variable POC which contains:

- create variable
- manage variable
- use variable interact with explore visualization
- some use cases
```
# as aggregation function
source=opensearch_dashboards_sample_data_flights| where `FlightDelay` = true | stats ${function}(`FlightDelayMin`) as ${function}_delay by `OriginWeather`, `DestWeather`

# as value
source=opensearch_dashboards_sample_data_flights|where DestCityName = "${DestCityName}" | stats count() as num_flights by `OriginCityName`

# as time interval
source=opensearch_dashboards_sample_data_logs| stats AVG(`bytes`) as avg_bytes, MAX(`bytes`) as max_bytes by span(`timestamp`, $Interval)

# as dimension
source=opensearch_dashboards_sample_data_ecommerce| stats count() as order_count by `$order_count_dimension`

# as metrics
source=opensearch_dashboards_sample_data_logs
| stats avg($log_metric) as avg_${log_metric} by span(`timestamp`, 1h) as timestamp, response
```

https://github.com/user-attachments/assets/a1bce058-55c9-4919-9429-7918be2aec36

---

### 1.1 Current Limitations

OpenSearch Dashboards introduced the [Explore](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9826) experience, enabling users to build visualizations directly from queries. This approach allows users to express complex analytical logic in a familiar query syntax and quickly generate visual insights without requiring deep knowledge of visualization configuration.
<img width="664" height="216" alt="Image" src="https://github.com/user-attachments/assets/8c483558-aecc-4ed7-8748-54aaf2da88ee" />

However, users currently have no way to parameterize queries at the dashboard level. To change a field name, aggregation function, or dimension, users must either:

1. Copy and edit each visualization's query/configuration, or
2. Duplicate entire dashboards to support different analytical perspectives

This approach leads to significant operational challenges:

* Dashboard Sprawl: Multiple copies of the same business logic
[...]

PR #11646, opened 2026-04-01
[Dashboards] feat: Add Dashboard Variables Support

yubonluo:
### Description

This PR proposes a Dashboard-level Variables system for OpenSearch Dashboards that enables users to define reusable parameters and dynamically inject them into queries through explicit template syntax. Variables eliminate the need for dashboard duplication when users need to modify query structure (field names, aggregation functions, dimensions, or metrics), enabling template-based dashboards where a single configuration can serve multiple business contexts.

### Description

Introduces dashboard-level variables that allow users to create dynamic, parameterized dashboards. Variables can be referenced in queries using `$varName` or `${varName}` syntax and are interpolated at query execution time.

### Changes

**Core Variable System** (`src/plugins/dashboard/public/variables/`)
- `VariableService` — manages variable CRUD, options refresh, and state through DashboardContainer's input/output mechanism
- `VariableInterpolationService` — replaces `$var` / `${var}` placeholders in queries with actual values, with language-aware formatting for multi-select (PPL `IN ('a','b')`, PromQL `(a|b)`)
- `variable_query_utils.ts` — executes queries via SearchSource and parses response hits into deduplicated string option arrays
- Type definitions for `QueryVariable` (options fetched from a query) and `CustomVariable` (manually defined comma-separated options)

**Dashboard Integration**
- `DashboardContainer` — instantiates VariableService/InterpolationService, connects them to container input, subscribes to time range and reload changes to auto-refresh query variable options
- `DashboardContainerInput` — extended with optional `variables` field
- App state sync — variables are persisted in URL state (with query variable options stripped to keep URLs small) and in saved objects via `variablesJSON`
- Server-side saved object schema updated with `variablesJSON` text field

**UI Components** (`dashboard_variables/`)
- `VariablesBar` — displays variable dropdowns at the top of the dashboard with single/multi-select support, "All" option, collapse/expand toggle
- `VariableEditorFlyout` — form for creating/editing variables (name, type, query, custom values, multi-select, include-all)
- `VariableManagementFlyout` — lists all variables with drag-to-reorder, hide/show toggle, edit, and delete actions
- `VariableQueryPanel` — Monaco-based query editor with autocomplete, language toggle (PPL/PromQL), dataset selector, and query preview

**Explore Plugin Integration**
- `ExploreEmbeddable` — subscribes to parent DashboardContainer's variable changes, interpolates saved query, and re-fetches data. Skips re-execution while variables are in loading state to avoid stale value interpolation
- `VisualizationEditorContext` — creates a standalone VariableService (backed by BehaviorSubject) when editing a visualization from a dashboard with variables
- `useVariableInterpolation` hook — wires variable refresh on time changes, sets interpolation service on QueryBuilder, and triggers query re-execution when variable values change (with loading-state guard)
- `QueryBuilder.executeQuery` — interpolates variables in query string before execution

**Cross-Plugin Wiring**
- `EmbeddableEditorState` — extended with `variables` so variables are passed through when navigating to visualization editors
- `EditPanelAction` — reads variables from root container and includes them in editor state
- `DatasetSelect` — added `controlledSelectedDataset` prop for controlled mode (avoids mutating global query state when used in variable editor)

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11550

## Screenshot
Manage variable
image

edit variable
image

## Changelog

- feat: [Dashboards] feat: Add Dashboard Variables Support

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/dashboard/public/application/components/dashboard_editor.tsx
```diff
@@ -75,21 +76,32 @@ export const DashboardEditor = () => {
     execute: handleFullScreen,
   });
 
+  const theme = localStorage.getItem('__DEVELOPMENT__.discover.vis.theme') || 'echarts';
```
ruanyl: @yubonluo you can use a new flag, `__DEVELOPMENT__.dashboard.variable.enabled`

src/plugins/dashboard/public/application/components/dashboard_variables/variable_editor_flyout.tsx
```diff
@@ -0,0 +1,419 @@
...
+              </>
+            )}
+
+            <EuiFormRow
+              label={i18n.translate('dashboard.variableEditor.nameLabel', {
+                defaultMessage: 'Name',
+              })}
+              helpText="Use this name to reference the variable: $var or ${var}"
+            >
+              <EuiFieldText
```
ruanyl: Let's use `compressed` variant of form input components `EuiFieldText, EuiSuperSelect, EuiSwitch`, etc

src/plugins/dashboard/public/application/components/dashboard_variables/variables_bar.tsx
```diff
@@ -0,0 +1,330 @@
...
+                Add Variable
+              </EuiSmallButtonEmpty>
+            </EuiFlexItem>
+          )}
+        </>
+      )}
+      {isCollapsed && visibleVariables.length > 0 && (
+        <EuiFlexItem grow={false}>
+          <EuiText size="xs" color="subdued">
+            {visibleVariables.length} variable{visibleVariables.length !== 1 ? 's' : ''} hidden
```
ruanyl: Nit: i18n, we can fix all i18n issues in a follow up PR

src/plugins/dashboard/public/application/components/dashboard_variables/variables_bar.tsx
```diff
@@ -0,0 +1,330 @@
...
+            }}
+            height={300}
+            singleSelection={variable.multi ? false : 'always'}
+          >
+            {(list, search) => (
+              <div>
+                <div style={{ padding: '8px' }}>{search}</div>
+                <EuiText size="xs" color="subdued" style={{ padding: '0 12px 4px' }}>
+                  {variable.options.length} options
+                </EuiText>
```
ruanyl: remove this to keep the UI clean?

src/plugins/dashboard/public/application/components/dashboard_variables/variables_bar.tsx
```diff
@@ -0,0 +1,330 @@
...
+            searchProps={{
+              placeholder: 'Contains...',
+              compressed: true,
+            }}
+            height={300}
+            singleSelection={variable.multi ? false : 'always'}
+          >
+            {(list, search) => (
+              <div>
+                <div style={{ padding: '8px' }}>{search}</div>
```
ruanyl: Nit:
```suggestion
                <div style={{ padding: '8px 8px 0 8px' }}>{search}</div>
```

src/plugins/dashboard/public/variables/types.ts
```diff
@@ -0,0 +1,75 @@
...
+  /** Variable name (used in queries as $name) */
+  name: string;
+  /** Display label */
+  label?: string;
+  /** Variable type */
+  type: VariableType;
+  /** Available options (simple string array) */
+  options: string[];
+  /** Currently selected value */
+  current?: string;
```
ruanyl: I see `options` is persisted to saved object, but I feel it shouldn't.  For custom type variable, `options` can be derived from `customOptions`. For query type variable, `options` always based on the current query result.

I suggest to model the variable in this way:

```ts
// These are the common variable metadata which are the same across different type of variables
interface VariableMeta {
  id: string;
  name: string;
  label?: string;
  // ... all the rest common meta
}

// Custom variable metadata
interface CustomVariable extends VariableMeta {
  type: 'custom';
  customOptions: string;
}

// Query variable metadata
interface QueryVariable extends VariableMeta {
  type: 'query';
  query: string;
  language: string;
  dataset?: {
    id: string;
    title: string;
    type: string;
    timeFieldName?: string;
  };
  refresh?: 'never' | 'onDashboardLoad' | 'onTimeRangeChange';
}

// UI state for Query variable, should never persisted
interface QueryVariableState {
  loading?: boolean;
  error?: string;
}

type Variable = CustomVariable | QueryVariable & QueryVariableState
```

src/plugins/dashboard/public/variables/types.ts
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Query type variable - options fetched from a query
+ */
+export interface QueryVariable extends BaseVariable, VariableQueryParams {
+  /** Refresh interval */
+  refresh?: 'never' | 'onDashboardLoad' | 'onTimeRangeChange';
+  /** Loading state */
+  loading?: boolean;
+  /** Error message */
+  error?: string;
```
ruanyl: It seems `loading` and `error` are also persisted, I believe they should not, let's exclude them before saving

src/plugins/dashboard/public/variables/active_variables.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Module-level storage for the currently active dashboard's variables.
+ * Set by DashboardContainer when rendered, read by DashboardStart.getActiveVariables().
+ */
+let activeVariables: Variable[] | undefined;
+
+export function setActiveVariables(variables: Variable[] | undefined) {
+  activeVariables = variables;
+}
+
+export function getActiveVariables(): Variable[] | undefined {
```
ruanyl, line 18: If I understood correctly, this will be changed to use url to send dashboard id right?

src/plugins/dashboard/public/variables/types.ts
ruanyl: What about we store `current` as an array for both single select and multi-select?
```
current: string[]
```

Then you don't need to split the value by `,` later, this could be causing bug potentially, because the value itself may contain `,`.

src/plugins/dashboard/public/application/components/dashboard_editor.tsx
yubonluo: Got it, i will update the key

src/plugins/dashboard/public/application/components/dashboard_variables/variable_editor_flyout.tsx
yubonluo: Got it, I will update all of the input compoments

src/plugins/dashboard/public/application/components/dashboard_variables/variables_bar.tsx
yubonluo: This ui is used to show 
image

src/plugins/dashboard/public/application/components/dashboard_variables/variables_bar.tsx
yubonluo: It will cause the search box to stick closely to the options number text

src/plugins/dashboard/public/variables/active_variables.ts
yubonluo: Yes, if the dashboard id can be got from the url, this change will be removed

src/plugins/dashboard/public/variables/types.ts
yubonluo: Got it, I will update it.

src/plugins/dashboard/public/variables/types.ts
yubonluo: Got it, will optimize this logic

src/plugins/dashboard/public/variables/types.ts
yubonluo: Thanks yulong, has updated this type

src/plugins/dashboard/public/application/components/dashboard_variables/variables_bar.tsx
ruanyl: I see, but not an issue, just I see OUI selectable examples which doesn't have padding-bottom which looks fine to me

src/plugins/dashboard/public/application/components/dashboard_variables/variables_bar.tsx
ruanyl: Right, maybe we don't need it? It doesn't see necessary

src/plugins/dashboard/public/variables/types.ts
```diff
@@ -0,0 +1,87 @@
...
+ */
+export interface VariableMeta {
+  /** Unique identifier */
+  id: string;
+  /** Variable name (used in queries as $name) */
+  name: string;
+  /** Display label */
+  label?: string;
+  /** Variable type */
+  type: VariableType;
```
ruanyl, line 25: This `type` seems not needed here, because you have `type` defined for each type of variable

src/plugins/dashboard/public/variables/types.ts
```diff
@@ -0,0 +1,87 @@
...
+  customOptions: string[];
+}
+
+/**
+ * Query type variable — options fetched from a query
+ */
+export interface QueryVariable extends VariableMeta, VariableQueryParams {
+  type: VariableType.Query;
+  /** Refresh interval (Not be used) */
+  refresh?: 'never' | 'onDashboardLoad' | 'onTimeRangeChange';
```
ruanyl: Nit: If `(Not be used)`, let's remove it for now?

src/plugins/dashboard/public/application/embeddable/dashboard_container.tsx
```diff
@@ -139,6 +155,58 @@ export class DashboardContainer extends Container<InheritedChildInput, Dashboard
     );
     this.embeddablePanel = options.embeddable.getEmbeddablePanel(stateTransfer);
     this.logos = options.chrome.logos;
+
+    this.variableService = new VariableService();
+
+    if (options.data) {
+      this.variableService.setupDataPlugin(options.data);
```
ruanyl: Nit: could simply pass `options.data` as `VariableService` constructor?

```
this.variableService = new VariableService({data});
```

src/plugins/dashboard/public/variables/variable_query_utils.ts
```diff
@@ -0,0 +1,84 @@
...
+  if (Array.isArray(value)) {
+    // Flatten arrays — each element becomes a separate option
+    return value.flatMap((item) => extractValues(item));
+  }
+
+  if (typeof value === 'string') {
+    return value ? [value] : [];
+  }
+
+  if (typeof value === 'number' || typeof value === 'boolean') {
```
ruanyl, line 26: Can you add a comment to elaborate why should it converts number/boolean to string?

src/plugins/dashboard/public/application/components/dashboard_variables/variables_bar.tsx
yubonluo: Yes, has removed

src/plugins/dashboard/public/application/components/dashboard_variables/variables_bar.tsx
yubonluo: Got it, updated

src/plugins/dashboard/public/application/utils/update_saved_dashboard.ts
```diff
@@ -47,6 +47,10 @@ export function updateSavedDashboard(
   savedDashboard.timeRestore = appState.timeRestore;
   savedDashboard.panelsJSON = JSON.stringify(appState.panels);
   savedDashboard.optionsJSON = JSON.stringify(appState.options);
+  savedDashboard.variablesJSON =
+    appState.variables && appState.variables.length > 0
+      ? JSON.stringify(appState.variables)
```
ruanyl: We can store it as `variablesJSON: {values: appState.variables}` for better extensibility, in case we want to add more attributes to variablesJSON object

src/plugins/dashboard/public/variables/types.ts
```diff
@@ -0,0 +1,87 @@
...
+  /** Unique identifier */
+  id: string;
+  /** Variable name (used in queries as $name) */
+  name: string;
+  /** Display label */
+  label?: string;
+  /** Variable type */
+  type: VariableType;
+  /** Currently selected value(s) */
+  current?: string[];
```
ruanyl, line 27: I thinking this scenario:
1. User created a multi variable
2. Select multiple values ["s1," "s2", "s3"]
3. Edit the variable to disable multi
4. How will it handle this?

src/plugins/dashboard/public/application/components/dashboard_variables/dashboard_variables.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { Variable } from '../../../variables/types';
+import { VariableService } from '../../../variables/variable_service';
+import { IVariableInterpolationService } from '../../../variables/variable_interpolation_service';
+import { DashboardAppState } from '../../../types';
+import { ViewMode } from '../../../../../embeddable/public';
+
+export interface DashboardVariablesProps {
+  variableService: VariableService;
+  interpolationService?: IVariableInterpolationService;
+  currentAppState: DashboardAppState;
```
ruanyl: Can we pass what's needed instead of the entire `DashboardAppState`? We only need `ViewMode`, then we can have a prop `mode: 'view' | 'edit'`

src/plugins/dashboard/public/variables/types.ts
yubonluo: You are right, but I kept it here so that `VariableMeta` is self-descriptive as a base interface — anyone reading it can immediately see that every variable has a type field, without needing to check the sub-interfaces.

src/plugins/dashboard/public/variables/types.ts
yubonluo: got it, removed it

src/plugins/dashboard/public/variables/variable_query_utils.ts
yubonluo: yes, will add some comment

src/plugins/explore/public/application/in_context_vis_editor/visualization_editor_page.tsx
```diff
@@ -26,13 +25,17 @@ import { useQueryBuilderState } from './hooks/use_query_builder_state';
 import { useVisualizationBuilder } from './hooks/use_visualization_builder';
 import { syncQueryStateWithUrl } from '../../../../data/public';
 import { VISUALIZATION_EDITOR_APP_ID } from '../../../common';
+import { VariablesBar } from '../../../../dashboard/public';
```
ruanyl: What do you think if we expose `VariablesBar` from dashboard plugin lifecycle method `setup/start`, then consumer won't rely on static import

src/plugins/dashboard/public/application/utils/update_saved_dashboard.ts
yubonluo: got it, will update the variablesJSON

src/plugins/dashboard/public/variables/types.ts
yubonluo: If user switch from multi to single, the current should be the first value ["s1"], the code missing this logic and will add it

src/plugins/dashboard/public/application/components/dashboard_variables/dashboard_variables.tsx
yubonluo: Yes, let me update it

src/plugins/dashboard/public/application/embeddable/dashboard_container.tsx
yubonluo: got it, will update the constructor

src/plugins/explore/public/application/in_context_vis_editor/visualization_editor_page.tsx
ruanyl: Not important, just want to discuss

yubonluo: `VariablesBar` is a pure React component with no runtime dependencies — it only needs variableService as a prop. I think Static import is enough?

ruanyl: I'm actually thinking if it needs to a build time deps or we can make it a runtime deps

src/plugins/dashboard/public/application/components/dashboard_variables/variables_bar.tsx
yubonluo: Had updated the missing i18n

src/plugins/explore/public/application/in_context_vis_editor/visualization_editor_page.tsx
yubonluo: Yes, We could expose it as a runtime dependency through DashboardStart:
```
// dashboard plugin.tsx start()
return {
  VariableService,
  ...
};
```
Then explore consumes it via dashboard.VariablesBar instead of a static import. This would decouple the build-time dependency.

yubonluo: updated

ruanyl (dismissed):

wanglam (dismissed):

ruanyl (dismissed):

wanglam (approved):

ruanyl (approved):

PR #11649, opened 2026-04-01
Ruchi/session timeout 401

ruchidh:
### Description

When a user's session cookie expires, API calls (e.g., from the Discover/Explore page) now return 401 Unauthorized instead of the previous 303 See Other redirect. Since XHR/fetch calls follow redirects silently, the old 303 behavior meant the browser received HTML instead of JSON, couldn't parse it, and the user got stuck — never logged out, never redirected.

The backend has been updated to return 401 with an X-Auth-Redirect-URL response header containing the login page URL. This PR adds the frontend handling for that.

### Changes
A global HTTP response error interceptor is registered during CoreSystem.setup() that:

- Detects 401 responses across all HTTP calls made through the OSD HTTP service
- Reads the X-Auth-Redirect-URL header from the response
- Validates the redirect URL is safe (relative path or same-origin only) to prevent open redirect attacks
- Shows a warning toast notifying the user their session has expired
- Redirects to the login page after 5 seconds
- Halts the interceptor chain to prevent duplicate error toasts from other plugins
- Uses a deduplication flag to prevent multiple toasts when several API calls fail simultaneously, with automatic reset after the redirect timeout

Screenshot 2026-04-02 at 2 08 18 PM

Files changed
session_expired_interceptor.ts
 — New interceptor with URL validation and redirect logic
core_system.ts
 — Registers the interceptor during setup
session_expired_interceptor.test.ts
 — 11 test cases covering: toast display, redirect timing, non-401 passthrough, missing header, deduplication, undefined response, open redirect protection (external domains, protocol-relative URLs), same-origin URL acceptance, flag reset, and halt ordering
core_system.test.ts
 — Verifies the interceptor is called with correct arguments during setup
core_system.test.mocks.ts
 — Mock for the interceptor module
### Testing
Expire session cookie → click any UI element that triggers an API call → toast appears for 5 seconds → browser redirects to login page
Verified on the Explore page with the filter refresh triggering POST /api/enhancements/search/pplRequest returning 401
### Security
Redirect URL is validated before use — only relative URLs (/path) and same-origin absolute URLs are allowed. Protocol-relative URLs (
evil.com
) and cross-origin URLs are rejected to prevent open redirect attacks.

## Changelog
- fix: Interceptor to handle 401 unauthorized access

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/changelogs/fragments/11649.yml
+++ b/changelogs/fragments/11649.yml
@@ -1,0 +1,2 @@
+fix:
+- Interceptor to handle 401 unauthorized access ([#11649](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11649))


src/core/public/core_system.ts
```diff
@@ -176,6 +177,7 @@ export class CoreSystem {
       const http = this.http.setup({ injectedMetadata, fatalErrors: this.fatalErrorsSetup });
       const uiSettings = this.uiSettings.setup({ http, injectedMetadata });
       const notifications = this.notifications.setup({ uiSettings });
+      setupSessionExpiredInterceptor(http, notifications);
```
Maosaic, line 180: Should we add this into `HttpService` instead of directly inject into `core_system`?

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/http/http_service.ts

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/public/core_system.ts#L43

ruchidh: i can check this

ruchidh: Checked this, Good suggestion but http does not have access to notification.
The problem with putting it in HttpService is that the interceptor needs notifications.toasts to show the toast.

HttpService.setup() only receives injectedMetadata and fatalErrors — it doesn't have access to the notifications service.If we add notifications as a dependency of HttpService, 

That creates a circular-ish coupling: HTTP depends on notifications, but notifications depends on uiSettings, which depends on HTTP.

That's http → notifications → uiSettings → http — a circular dependency.

The alternative would be to register it in HttpService.start() instead, but that means any 401s during the setup/start phase wouldn't be caught (probably fine in practice since the UI isn't interactive yet).

Keeping it in CoreSystem.setup() is the simplest approach because — it's the one place that has access to all services and can wire them together. 

Let me know what do you think.

Maosaic (approved):

PR #11653, opened 2026-04-02
fix: Remove legacy write-on-read behavior from DataViewsService and IndexPatternsService

Maosaic:
### Description

The `get()` and `getMultiple()` methods in `DataViewsService` and `IndexPatternsService` currently perform writes on the read path. When loading a data view, `isFieldRefreshRequired()` checks whether stored field specs are missing `aggregatable`, `searchable`, or `readFromDocValues` properties. If any are missing, it fetches field metadata via `_fields_for_wildcard` and writes the updated fields back to the saved object — all as a side effect of a `get()` call.

This logic was a migration shim introduced in [elastic/kibana#8421](https://github.com/elastic/kibana/pull/8421) (September 2016) to handle the Kibana 4.x to 5.0 upgrade, when Elasticsearch added the `field_capabilities` API. OpenSearch Dashboards was forked from Kibana 7.10.2 (2021), by which point every index pattern already had these properties as standard.

### Problem

The write-on-read behavior causes several issues:

- **409 conflict storms**: When multiple data views with incomplete field metadata are loaded in parallel (e.g., via `getMultiple()` using `Promise.all`), each triggers an independent save. These saves race on the saved object version, producing 409 conflicts.
- **Recursive amplification**: The `updateSavedObject()` conflict retry handler calls `this.get()` to re-fetch the data view, which itself can trigger another `isFieldRefreshRequired` check, another `_fields_for_wildcard` call, and another save — creating a recursive loop bounded only by `MAX_ATTEMPTS_TO_RESOLVE_CONFLICTS` (3).
- **Unnecessary network calls**: Every `get()` call evaluates `isFieldRefreshRequired()` on every field of every data view, and when triggered, issues a `_fields_for_wildcard` request that is not needed for the read path.

### Changes

**Removed from `DataViewsService`** (`data_views.ts`):
- `isFieldRefreshRequired()` — legacy migration check
- `refreshFieldSpecMap()` — field refresh helper only used by the read path
- Write-on-read logic in `get()` and `getMultiple()` (`isFieldRefreshRequired` / `isSaveRequired` / `refreshFieldSpecMap` / `updateSavedObject` block)

**Removed from `IndexPatternsService`** (`index_patterns.ts`):
- Same three removals as above

**Updated tests** (`data_views.test.ts`, `index_patterns.test.ts`):
- Updated the `"should handle version conflicts"` test to reflect that `get()` no longer triggers a save. The test now verifies conflicts arise from explicit `updateSavedObject()` calls rather than implicit saves during `get()`.

### What is preserved

- `refreshFields()` — the explicit field refresh used by the **"Refresh field list"** button in Index Patterns management is untouched
- `updateSavedObject()` with conflict retry — still works for intentional saves (editing data views, creating new ones)
- `getFieldsForWildcard()` — still available for on-demand field fetching

### Impact

- `get()` and `getMultiple()` are now **purely read-only**
- No `_fields_for_wildcard` calls on the read path
- No writes to saved objects during page load
- The 409 conflict amplification loop can no longer be triggered by loading data views
- If a data view has incomplete field metadata, the field simply won't appear in aggregation dropdowns until the user clicks "Refresh field list" in Index Patterns management

### Testing

- All 66 `data_views` tests pass
- All 52 `index_patterns` tests pass

## Changelog

- fix: Remove legacy write-on-read behavior from DataViewsService and IndexPatternsService

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


changelogs/fragments/11653.yml
```diff
@@ -0,0 +1,2 @@
+fix:
```
seraphjiang: no change log file, please :)

PR #11684, opened 2026-04-06
[Explore] Add resizable query panel with drag handle

ruchidh:
### Description

Adds a resizable vertical split between the query panel and the content area in the Explore plugin (Logs, Metrics, Traces pages). Users can now drag to resize the query editor panel height.

###  Changes
ResizableQueryContainer component (resizable_query_container.tsx)

- Wraps the query panel and content area in an EuiResizableContainer with direction="vertical"
- Dynamically measures the query panel's initial height from Monaco's rendered .view-lines height plus the widgets bar, so the panel starts with single line height
-  Initial panel size is computed from the viewport height to always fit one editor line (~72px) regardless of screen size, clamped between 5-15
- Dispatches resize events on panel change and mount so Monaco recalculates its layout
- Resize handle styles (resizable_query_container.scss)
- `onDidContentSizeChange` in both explore and agent_traces `use_query_panel_editor.ts` now reads the resizable panel's allocated height via `closest()` instead of the editor's immediate parent, preventing multi-line queries from locking the panel height on page reload

- Custom styles for the EuiResizableButton to work around an oui/eui CSS class prefix mismatch where the default OUI styles don't apply to the EUI-prefixed button element
- Matches the existing OUI resizable button pattern: transparent background, two centered grab indicator lines, full-width line expansion on hover/focus with primary color highlight

### Screenshot

https://github.com/user-attachments/assets/adbf9f80-91a5-432d-9de5-f82d8708a314

###  Testing

- Verified on Logs, Metrics, and Traces pages
- Empty query: panel starts at single-line editor height
- Pre-filled query: panel starts at multi-line height
- Drag handle visible and functional, consistent with bottom container's resize handle
- Monaco editor relayouts correctly after resize

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/config/opensearch_dashboards.yml
+++ b/config/opensearch_dashboards.yml
@@ -419,9 +419,9 @@
 
 # @experimental Set the enabled value to true and provide an AG-UI agent to enable chat
 # chat:
-# enabled: true
-# agUiUrl: "http://your-ai-agent-url:port"
-# mlCommonsAgentId: "olly-chat-agent-id"
+  # enabled: true
+  # agUiUrl: "http://your-ai-agent-url:port"
+  # mlCommonsAgentId: "olly-chat-agent-id"
 
 # @experimental Set the value to true to enable context provider
 # contextProvider:
--- a/config/opensearch_dashboards.yml
+++ b/config/opensearch_dashboards.yml
@@ -384,7 +384,7 @@
 # opensearchDashboards.dashboardAdmin.users: ["dashboard_admin"]
 
 # Set the value to true to enable the new UI for savedQueries in Discover
-data.savedQueriesNewUI.enabled: true
+# data.savedQueriesNewUI.enabled: true
 
 # Set the value to true to enable the banner plugin
 # banner.enabled: true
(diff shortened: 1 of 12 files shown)


LDrago27:
@ruchidh Should we have a button/hot key area to reset the query editor size to default. I feel currently once you resize your editor even by mistake, there is no way to rest back to default without actually dragging it back.

ruchidh:
> @ruchidh Should we have a button/hot key area to reset the query editor size to default. I feel currently once you resize your editor even by mistake, there is no way to rest back to default without actually dragging it back.

good suggestion, what do you think if we just double click the resize button and it resets ? or may be can add a small icon next to it for reset ?

ruchidh:
Fixing cypress test

LDrago27:
> > @ruchidh Should we have a button/hot key area to reset the query editor size to default. I feel currently once you resize your editor even by mistake, there is no way to rest back to default without actually dragging it back.
> 
> good suggestion, what do you think if we just double click the resize button and it resets ? or may be can add a small icon next to it for reset ?

Yeah that might be neat. having a tinny button near the resize drag handle to reset it might be great. However it might get overlooked if it is too small, a tooltip might help .

Maosaic (dismissed):

Maosaic (approved):

LDrago27 (approved):

PR #11693, opened 2026-04-07
[Chat] Register disabled PPL execute query action as fallback on plugin start and restore on unmount

wanglam:
### Description

Refactors the `execute_ppl_query` assistant action registration in the Explore plugin to use the imperative `registerAssistantAction` API instead of the React hook `useAssistantAction`, and adds proper cleanup when the plugin stops.

- Extracts the tool definition (`EXECUTE_PPL_QUERY_TOOL_DEFINITION`) into a shared constant to avoid duplication between the enabled and disabled variants.
- Adds `registerDisabledPPLExecuteQueryAction` — a helper that registers a "disabled" version of the action which returns a structured stop instruction to the AI agent when the query panel is not mounted.
- Registers the disabled action as a placeholder at plugin `start()` time, so the agent always has a registered handler even before the query panel mounts.
- Replaces `useAssistantAction` (React hook) with direct calls to `registerAssistantAction` via `useMount`, making the registration lifecycle explicit and decoupled from React's render cycle.
- Uses `useUnmount` in `usePPLExecuteQueryAction` to restore the disabled version when the query panel component unmounts, preventing stale tool calls after navigation.
- Calls `unregisterAssistantAction` in the plugin's `stop()` lifecycle to clean up the registered action when the Explore plugin is torn down.
- Updates unit tests to remove `mockUseAssistantAction` and assert against `mockRegisterAssistantAction` instead.
- Adds tests to verify the action is unregistered on plugin stop, including the case where `contextProvider` was not available at start.

## Screenshot

N/A — no UI changes.

## Testing the changes

1. Open the Explore plugin (Logs/Traces/Metrics view).
2. Trigger an AI assistant action that calls `execute_ppl_query` — it should execute normally.
3. Trigger another `execute_ppl_query` and navigate away from the query panel.
4. Refresh page and open the chatbot
5. It should send a tool not found to the agent 
6. Run unit tests:
```
yarn test:jest src/plugins/explore/public/components/query_panel/actions/ppl_execute_query_action.test.tsx
yarn test:jest src/plugins/explore/public/plugin.test.ts
yarn test:jest src/plugins/context_provider/public/services/assistant_action_service.test.ts
```

### Check List

- [ ] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/query_panel/actions/ppl_execute_query_action.tsx
+++ b/src/plugins/explore/public/components/query_panel/actions/ppl_execute_query_action.tsx
@@ -4,7 +4,7 @@
  */
 
 import { useDispatch } from 'react-redux';
-import { useUnmount } from 'react-use';
+import { useMount, useUnmount } from 'react-use';
 import { AppDispatch } from '../../../application/utils/state_management/store';
 import { useOpenSearchDashboards } from '../../../../../opensearch_dashboards_react/public';
 import { ExploreServices } from '../../../types';
@@ -12,16 +12,6 @@ import { loadQueryActionCreator } from '../../../application/utils/state_managem
 import { useSetEditorTextWithQuery } from '../../../application/hooks';
 import { setDateRange } from '../../../application/utils/state_management/slices/query_editor/query_editor_slice';
 import { QueryExecutionStatus } from '../../../application/utils/state_management/types';
-
-interface PPLExecuteQueryArgs {
-  query: string;
-  autoExecute?: boolean;
-  description?: string;
-  from?: string;
-  to?: string;
-}
-
-const NOOP_ASSISTANT_ACTION_HOOK = (_action: any) => {};
 
 // Shared tool definition for execute_ppl_query action
 export const EXECUTE_PPL_QUERY_TOOL_DEFINITION = {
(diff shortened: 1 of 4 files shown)


src/plugins/explore/public/components/query_panel/actions/ppl_execute_query_action.tsx
```diff
@@ -22,6 +23,65 @@ interface PPLExecuteQueryArgs {
...
+
+// Helper function to register the disabled version of the action
+export function registerDisabledPPLExecuteQueryAction(
+  registerAction: (action: any) => void | undefined
+) {
+  if (!registerAction) return;
+
+  registerAction({
+    ...EXECUTE_PPL_QUERY_TOOL_DEFINITION,
+    available: 'disabled',
```
Hailong-am, line 59: with the available is disabled, we won't send this tool to LLM, right?

src/plugins/explore/public/components/query_panel/actions/ppl_execute_query_action.tsx
```diff
@@ -142,4 +172,13 @@ export function usePPLExecuteQueryAction(
...
   });
+
+  // Cleanup: restore the disabled version when component unmounts
+  useUnmount(() => {
+    // Re-register the disabled version on unmount
+    const registerAction = services.contextProvider?.actions?.registerAssistantAction;
+    if (registerAction) {
+      registerDisabledPPLExecuteQueryAction(registerAction);
+    }
+  });
```
Hailong-am, line 174: `usePPLExecuteQueryAction` will remove `execute_ppl_query` when unmount, will it have race condition for the case it disabled been registered and removed immediately

src/plugins/explore/public/components/query_panel/actions/ppl_execute_query_action.tsx
wanglam: Yes. It will be filter out in the createToolDefinition method here.
https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/context_provider/public/services/assistant_action_service.ts#L175

src/plugins/explore/public/components/query_panel/actions/ppl_execute_query_action.tsx
wanglam: Good catch. Let me refactor this to use `registerAssistantAction` directly.

wanglam:
Hi @Hailong-am @SuZhou-Joe , could you take a look this PR again? I've replaced the `useAssistantAction` with `registerAssistantAction` to avoid action been unregistered.

Issue #11439, opened 2026-03-04
[RFC] Time Series Analysis & Monitoring in Discover Metrics


jackiehanyang:
## Summary
Bring **Anomaly Detection (AD)**, **Forecasting preview**, and **Monitoring/Alerting** into the Discover - Metrics workflow so users can analyze and operationalize Prometheus/AMP metrics without leaving the page.

## Background
Today users can query Prometheus/AMP metrics in Discover - Metrics, but they still need to switch contexts to preview AD/forecasting or create alerting artifacts. This breaks flow and increases setup friction.

## Goals (P0)
1. Preview anomalies on Discover - Metrics for Prometheus queries.
2. Preview forecasting on Discover - Metrics in on-demand mode (no persistent forecaster required for preview).
3. Configure alert monitor creation on Discover - Metrics.
4. On create, automatically create AD detector first, then create monitor.
5. Allow saving the Metrics view to Dashboard so query + analysis/monitoring context is persisted.
6. P0 supports anomaly-based alerts only (no forecast-based alerts yet).

## User Workflow
1. User opens Discover - Metrics and runs a PromQL query
2. User sees anomaly/forecast preview overlays on the metric chart
3. User evaluates overlay quality
4. User configures the monitor in Discover - Metrics (a new tab next to Settings)
5. User clicks Create, system first creates required AD detector config using Prometheus source, then creates the monitor using that detector
6. Users can save the configured Metrics view to Dashboard and recover the same context after reload

### 1. Forecasting Preview API
- Add a new preview API for forecasting that supports on-demand execution
- Preview should not require creating a persistent forecaster document
- Return chart-friendly forecast points over requested time window

**API Contract**
`POST /_plugins/_forecast/forecasters/_preview`
```
{
  "source_type": "PROMETHEUS",
  "prometheus_source": {
    "query_language": "PROMQL",
    "query": "rate(go_gc_heap_allocs_bytes_total{instance=\"localhost:9090\"}[5m])",
    "data_connection_id": "local"
  },
  "period_start": 1772576400000,
  "period_end": 1772580000000,
  "forecast_interval": { "period": { "interval": 1, "unit": "Minutes" } },
  "horizon": 12,
  "history": 40
}
```
```
{
  "forecast_result": [
    { "timestamp": 1772580060000, "value": 0.44, "lower": 0.38, "upper": 0.49 }
  ],
  "metadata": {
    "source_type": "PROMETHEUS"
  }
}
```

[...]

PR #11729, opened 2026-04-10
[Metrics] Add Prometheus detector/forecaster/monitor flows and previe…

jackiehanyang:
## Summary

This PR adds Prometheus-backed anomaly preview and alert creation flows to the Metrics experience in Explore.

With this change, users can:

- preview anomaly and forecast overlays directly on Metrics charts
- create anomaly detectors and alerting monitors from the Metrics `Alerts` panel
- preserve the created detector/monitor association when saving the metric or adding it to a dashboard
- rediscover an existing detector/monitor for the same metric and avoid re-creating duplicates
- handle multi-series Prometheus queries more intentionally for both preview and detector creation
### Related Issue
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11439

### Metrics preview overlays

- Added anomaly preview and forecast preview routes for Metrics
- Rendered preview overlays on top of Metrics visualizations
- For multi-series Prometheus queries:
  - auto-preview overlays for all visible series when series count is small
  - otherwise allow choosing which series to preview
- Forecast overlays now follow the base metric series color instead of using a single fallback color
- Preview behavior is dependency-aware:
  - if AD UI is unavailable, preview requests are suppressed
  - if the backend preview dependency is missing, dead-end preview errors are suppressed

### Metrics Alerts panel

- Added an `Alerts` panel in Metrics for detector/monitor creation
- Supports Prometheus-backed detector creation from the current metric query
- For multi-series queries:
  - detect whether the result is a `single_stream` or `high_cardinality` candidate
  - infer candidate entity fields from varying Prometheus labels
  - prefill detector mode and entity field accordingly
- Persists created detector/monitor association in Explore saved state
- Rehydrates that association when reopening the same saved metric
- If no saved association exists, rediscovery checks for an existing detector/monitor for the same query/data connection and shows it instead of the create flow

### Saved metric / dashboard behavior

- Saving the metric or adding it to a dashboard now preserves the monitor/detector association
- Reopening the same saved Metrics view shows the existing monitor state instead of offering duplicate creation

### Dependency handling

- Hide the `Alerts` tab when Alerting UI is unavailable
- Do not show preview overlays when AD UI is unavailable

### Navigation / deep links

- Fixed `View monitor` deep link behavior for datasource-enabled environments
- Added the required Alerting URL params so classic monitor details open correctly

## Why

The Metrics experience already allows users to explore Prometheus-backed metrics, but the alerting/anomaly workflow was disconnected from that context.

This PR makes the Metrics page a first-class entry point for:

- previewing anomalies and forecasts
- creating detectors and monitors from the current metric query
- preserving that relationship when the metric is saved and reopened

## Multi-series behavior

For multi-series Prometheus queries, this PR intentionally distinguishes between:

- `single_stream`
- `high_cardinality`

Entity field candidates are inferred from the returned Prometheus label sets rather than being hardcoded to `instance`.

Example:

- if only `instance` varies, `instance` is prefilled
- if `instance` and `quantile` both vary, `instance` is preferred as the default

## Duplicate prevention

When a detector/monitor already exists for the same metric query and data connection, the `Alerts` panel will:

- restore the saved association if available, or
- rediscover the existing resources from the backend

This prevents showing the create flow again for the same persisted metric.

### Manual

- Verified anomaly preview and forecast preview on Metrics
- Verified multi-series preview behavior
- Verified detector + monitor creation from Metrics
- Verified saving the metric / adding to dashboard preserves the association
- Verified reopening the same metric shows the existing monitor state
- Verified `View detector` and `View monitor` links
- Verified datasource-enabled deep-link behavior

### Automated

- `prettier --check` on touched files
- focused lint/test passes where repo tooling allowed
- some existing repo-level Jest / optimizer / duplicate-mock noise is still present and not introduced by this PR

## Notes

This PR is the OpenSearch Dashboards side of the feature. Full end-to-end behavior depends on the companion AD backend support for Prometheus-backed AD flows. Will link the pr here soon

https://github.com/user-attachments/assets/b782e231-0135-4103-b015-0405f77066b6

### DIFF
--- a/src/plugins/explore/public/application/pages/metrics/metrics_bottom_container/bottom_right_container/metrics_alerts_helpers.ts
+++ b/src/plugins/explore/public/application/pages/metrics/metrics_bottom_container/bottom_right_container/metrics_alerts_helpers.ts
@@ -1,0 +1,380 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+  PROMETHEUS_DETECTOR_MODES,
+  PrometheusDetectorMode,
+} from '../../../../utils/metrics_feature_constants';
+
+const DEFAULT_MONITOR_INTERVAL_MINUTES = 1;
+const DEFAULT_ANOMALY_THRESHOLD = 0.7;
+const DEFAULT_TRIGGER_SEVERITY = '1';
+const DEFAULT_AD_RESULT_INDEX = '.opendistro-anomaly-results*';
+
+const PROMQL_RANGE_SELECTOR_REGEX = /\[\s*([^\]:]+?)\s*(?::[^\]]*)?\]/;
+const PROMQL_DURATION_SEGMENT_REGEX = /(\d+(?:\.\d+)?)(ms|s|m|h|d|w|y)/g;
+
+const DURATION_UNIT_SECONDS: Record<string, number> = {
+  ms: 0.001,
+  s: 1,
+  m: 60,
+  h: 3600,
+  d: 86400,
+  w: 604800,
+  y: 31536000,
+};
+
+const PROMQL_RESERVED_WORDS = new Set([
+  'bool',
+  'by',
+  'without',
+  'on',
+  'ignoring',
+  'group_left',
+  'group_right',
+  'and',
+  'or',
+  'unless',
+  'sum',
+  'avg',
+  'min',
+  'max',
+  'count',
+  'stddev',
+  'stdvar',
+  'topk',
+  'bottomk',
+  'quantile',
+  'rate',
+  'irate',
+  'increase',
+  'delta',
+  'idelta',
+]);
+
+export interface MetricsAlertFormValues {
+  monitorName: string;
+  scheduleIntervalMinutes: number;
+  triggerName: string;
+  severity: string;
+  anomalyGradeThreshold: number;
+  anomalyConfidenceThreshold: number;
+  detectorMode: PrometheusDetectorMode;
+  selectedSeriesId: string;
+  selectedEntityField: string;
+}
+
+export interface MetricsAlertMonitorMetadata {
+  detectorId: string;
+  detectorName: string;
+  promqlQuery: string;
+  dataConnectionId: string;
+  dataSourceId?: string;
+  detectorMode?: PrometheusDetectorMode;
+  selectedSeriesId?: string;
+  selectedEntityField?: string;
+  selectedSeriesLabels?: Record<string, string>;
+}
+
+const sanitizeNamePart = (rawName: string): string =>
+  (rawName || '')
+    .trim()
+    .replace(/[^a-zA-Z0-9_-]+/g, '-')
+    .replace(/^-+|-+$/g, '')
+    .slice(0, 48);
+
+const parsePromqlDurationToSeconds = (durationText: string): number | undefined => {
+  const trimmed = (durationText || '').trim();
+  if (!trimmed) {
+    return undefined;
+  }
+
+  let totalSeconds = 0;
+  let matchedLength = 0;
+  PROMQL_DURATION_SEGMENT_REGEX.lastIndex = 0;
+  let matchedSegment = PROMQL_DURATION_SEGMENT_REGEX.exec(trimmed);
+
+  while (matchedSegment) {
+    if (matchedSegment.index !== matchedLength) {
+      return undefined;
+    }
+
+    const value = Number(matchedSegment[1]);
+    const unit = matchedSegment[2];
+    if (!Number.isFinite(value) || !DURATION_UNIT_SECONDS[unit]) {
+      return undefined;
+    }
+
+    totalSeconds += value * DURATION_UNIT_SECONDS[unit];
+    matchedLength += matchedSegment[0].length;
+    matchedSegment = PROMQL_DURATION_SEGMENT_REGEX.exec(trimmed);
+  }
+
+  if (matchedLength !== trimmed.length || totalSeconds <= 0) {
+    return undefined;
+  }
+
+  return totalSeconds;
+};
+
+export const extractPromqlMetricName = (queryText: string): string | undefined => {
+  const query = (queryText || '').trim();
+  if (!query) {
+    return undefined;
+  }
+
+  const metricSelectorRegex = /([a-zA-Z_:][a-zA-Z0-9_:]*)\s*(?=\{|\[)/;
+  const selectorMatch = metricSelectorRegex.exec(query);
+  if (selectorMatch && selectorMatch[1]) {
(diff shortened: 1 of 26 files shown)


src/plugins/explore/public/application/pages/metrics/metrics_bottom_container/bottom_right_container/metrics_alerts_panel.tsx
```diff
@@ -0,0 +1,1556 @@
+/*
```
joshuali925, line 1: nit. maybe put alert/AD related files in a new sub folder?

src/plugins/explore/public/application/pages/metrics/metrics_bottom_container/bottom_right_container/metrics_alerts_panel.tsx
```diff
@@ -0,0 +1,1556 @@
...
+  const serializedError =
+    typeof error === 'string'
+      ? error
+      : typeof error?.body === 'string'
+      ? error.body
+      : error?.body
+      ? JSON.stringify(error.body)
+      : error
+      ? JSON.stringify(error)
+      : '';
```
joshuali925, line 84: do we actually have these many possibilities when handling error?

src/plugins/explore/server/routes/anomaly_preview.ts
```diff
@@ -0,0 +1,394 @@
...
+          bucketSizeSeconds: schema.maybe(schema.number()),
+          shingleSize: schema.maybe(schema.number()),
+          promqlQuery: schema.maybe(schema.string()),
+          dataConnectionId: schema.maybe(schema.string()),
+          seriesValue: schema.maybe(schema.string()),
+        }),
+      },
+    },
+    async (context, request, response) => {
+      const client = context.core.opensearch.client.asCurrentUser;
```
joshuali925: does this need MDS support?

src/plugins/explore/public/application/pages/metrics/metrics_bottom_container/bottom_right_container/metrics_alerts_panel.tsx
```diff
@@ -0,0 +1,1556 @@
...
+    () =>
+      String((actionDeps as any)?.queryInEditor ?? (actionDeps.query as any)?.query ?? '').trim(),
+    [actionDeps]
+  );
+  const dataConnectionId = useMemo(
+    () => String((actionDeps.query as any)?.dataset?.id || '').trim(),
+    [actionDeps.query]
+  );
+  const explicitDataSourceId = useMemo(
+    () => String((actionDeps.query as any)?.dataset?.dataSource?.id || '').trim(),
```
joshuali925, line 374: does prometheus dataset have datasource id?

src/plugins/explore/public/application/pages/metrics/metrics_bottom_container/bottom_right_container/metrics_alerts_panel.tsx
```diff
@@ -0,0 +1,1556 @@
...
+  const requiresManagedDataSource = services.dataSourceEnabled && services.hideLocalCluster;
+  const missingManagedDataSource =
+    requiresManagedDataSource && !isResolvingDataSourceId && !dataSourceId;
+  const isActionAvailable =
+    supportsPrometheusAlerts &&
+    isStatusAllowed &&
+    !isAOSSCollection &&
+    missingUiDependencies.length === 0 &&
+    !isResolvingDataSourceId &&
+    !missingManagedDataSource;
```
joshuali925, line 404: the UI is available if current prometheus dataset's datasource id or workspace default datasource is a supported opensearch cluster?

If workspace default datasource is AOSS, is there a way for customer to use this feature without changing default datasource?

joshuali925 (commented):
thanks for adding this

i didn't get to review everything yet, will look more

src/plugins/explore/public/application/pages/metrics/metrics_bottom_container/bottom_right_container/metrics_alerts_panel.tsx
jackiehanyang: Good point. We don't actually expect that many error shapes here.

In this flow the error is typically either:
- `HttpFetchError` from `services.http`, where the useful extra detail lives in `error.body`
- a plain `Error` from our own code, which is already covered by `rawMessage`

I simplified this block to only inspect `error.body` when normalizing dependency-related backend errors, and otherwise rely on `rawMessage`.

src/plugins/explore/public/application/pages/metrics/metrics_bottom_container/bottom_right_container/metrics_alerts_panel.tsx
jackiehanyang: Not by default, no. 

By checking the existing code, current Prometheus dataset shape: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/query_enhancements/public/datasets/prometheus_type.ts#L20
- `dataset.id` = Prometheus `dataConnectionId`
- `dataset.dataSource` only contains `meta`
- it does not set `dataset.dataSource.id`

So for Prometheus in Explore today:
- use `query.dataset.id` for the Prometheus connection
- do not assume `query.dataset.dataSource.id` exists

that means there are two different ids:
- `dataConnectionId`: prometheus connection, from `dataset.id`.
- `dataSourceId`: OpenSearch target domain for AD/Alerting, resolved separately.

So Prometheus dataset has a connection id, it does not inherently have an OpenSearch datasource id

src/plugins/explore/server/routes/anomaly_preview.ts
jackiehanyang: Yes it needs MDS support, missed it here. Updated it with MDS support, in forecast_preview as well, in the new revision

src/plugins/explore/public/application/pages/metrics/metrics_bottom_container/bottom_right_container/metrics_alerts_helpers.ts
```diff
@@ -0,0 +1,380 @@
...
+
+import {
+  PROMETHEUS_DETECTOR_MODES,
+  PrometheusDetectorMode,
+} from '../../../../utils/metrics_feature_constants';
+
+const DEFAULT_MONITOR_INTERVAL_MINUTES = 1;
+const DEFAULT_ANOMALY_THRESHOLD = 0.7;
+const DEFAULT_TRIGGER_SEVERITY = '1';
+const DEFAULT_AD_RESULT_INDEX = '.opendistro-anomaly-results*';
```
lezzago, line 14: should this information be in a sub-folder/module for AD and alerting specific pieces?

PR #11753, opened 2026-04-15
[Explore Vis]separate and expose query panel 

Qxisylolo:
### Description

This PR introduces a generic query editor designed for reuse across multiple plugins

## Screenshot

截屏2026-07-22 16 08 37

截屏2026-07-22 16 10 57

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/types.ts
```diff
@@ -83,6 +84,9 @@ export interface ExplorePluginStart {
   savedSearchLoader: SavedExploreLoader;
   savedExploreLoader: SavedExploreLoader;
   slotRegistry: SlotRegistryServiceStart;
+  ui: {
+    QueryEditor: React.ComponentType<QueryEditorProps>;
```
ruanyl: Do we need to expose it now? If not, I'd remove it util there is a clear use case.

src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_query_panel_editor.ts
```diff
@@ -345,7 +338,6 @@ export const useQueryPanelEditor = (): UseQueryPanelEditorReturnType => {
           'explore'
         );
 
-        // Get the current dataset from Query Service to avoid stale closure values
         const currentDataset = queryString.getQuery().dataset;
```
ruanyl: Should it use `queryState.dataset` here as well? Otherwise, the dataset of autocomplete might be different from the `queryState.dataset`.

src/plugins/explore/public/types.ts
Qxisylolo: ok, I c, will delete it

src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_query_panel_editor.ts
Qxisylolo: I think the reason they use queryString.getQuery().dataset is:
```
  componentDidUpdate(prevProps: Props) {
    // Re-register providers when the language changes on an already-mounted
    // editor. react-monaco-editor handles language switches via
    // setModelLanguage without re-calling editorDidMount, so neither
    // editorDidMount nor onLanguage (one-shot) would fire in that case.
    // Note: we intentionally do NOT compare provider prop identity here
    // because callers pass fresh object literals on every render, which
    // would cause dispose/re-register churn on every keystroke.
    if (prevProps.languageId !== this.props.languageId) {
      this._registerProviders(this.props.languageId);
      this._setLanguageConfiguration(this.props.languageId, true);
    }
  }
  ```
dataset change（without language change） will not cause provider to re-register. Thus code editor will keep the stale queryState.dataset closure. We can safely use queryString.getQuery() since any dataset change syncs to global.

Comment (ruanyl) [reply src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_query_panel_editor.ts]: Then perhaps it should use `datasetRef`? 
My concern is now `queryState` is passed as component props, essentially, it could be from any source, then there is possibility that the `queryState.dataset` is different from the `queryString.getQuery().dataset`

File: src/plugins/explore/public/application/in_context_vis_editor/component/query_editor.tsx [hunk 3]
```diff
@@ -3,38 +3,11 @@
...
-      data-test-subj="exploreQueryPanelEditor"
-      onClick={onEditorClick}
-    >
-      <CodeEditor {...editorProps} />
-      {showPlaceholder ? (
-        <div className={`exploreQueryPanelEditor__placeholder`}>{placeholder}</div>
-      ) : null}
-    </div>
-  );
+  return <InnerQueryEditor {...props} />;
```
joshuali925, line 12: i guess this is not a real problem, but do we still need this wrapper around inner query editor? it confused me for a bit to find the difference between `ExploreQueryPanelEditor` and this component

joshuali925 (commented):
there are some conflicts probably due to the query builder PR merged https://github.com/opensearch-project/OpenSearch-Dashboards/pull/12398

src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_query_panel_editor.ts
Qxisylolo: yes, updated, thanks

src/plugins/explore/public/application/in_context_vis_editor/component/query_editor.tsx
Qxisylolo: These two components are processed with different parameters. explore log query editor reads states from redux. and this one reads states from query builder, a self-contained store.

ruanyl (dismissed):

ruanyl (approved):

src/plugins/explore/public/application/in_context_vis_editor/query_builder/query_builder.ts
```diff
@@ -357,22 +356,9 @@ export class QueryBuilder {
...
-    if (promptMode.status === 'fulfilled') {
-      updates.promptModeIsAvailable = promptMode.value;
-    }
-    if (summaryAgent.status === 'fulfilled') {
-      updates.summaryAgentIsAvailable = summaryAgent.value;
-    }
-    if (Object.keys(updates).length > 0) {
-      this.updateQueryEditorState(updates);
-    }
+  private async checkAgentAvailability(_datasourceId?: string) {
```
yubonluo, line 359: Should we add a try/catch for this?

Qxisylolo: getPromptModeIsAvailable will handle the error and return false if error.

yubonluo (approved):

PR #11759, opened 2026-04-15
Ruchi/hide-ai-editor_Drag

ruchidh:
### Description

In AI/prompt mode, the EuiResizableButton (resize handle) overlaps the generated query bar's "Replace query" button, making it unclickable. This happens because the resizable panel's initial height is too small to contain both the editor and the generated query bar, causing the content to overflow. The resize handle sits on top (z-index: 1000) and blocks interaction.

The fix conditionally renders a simple stacked layout in prompt mode instead of the resizable container like how it was earlier.

In prompt mode the editor is a single-line input, so resizing adds no value. 
The resize requirement is for PPL/query mode which keeps the resizable behavior unchanged.

## Testing the changes

- Existing unit tests pass (PPL/resizable)
- New unit tests cover prompt mode: stacked layout renders, no resize handle
- Fixes Cypress ai_editor.spec.js failures: should be able to query via AI mode for PPL-INDEX_PATTERN and should be able to query via AI mode for PPL-INDEXES

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


LDrago27:
@ruchidh So we are kind of disabling the drag handles all together for the AI/Prompt mode.  The resizing feature is only available for the normal query editor. 

yes, it is  a requirement only for large queries, currently ai or natural lagnguage fixed to a single text input so keep it as it is. In future if requirement comes then it would require change for generated ppl too at the bottom.

LDrago27 (approved):

Maosaic (approved):

PR #1234, opened 2022-02-10
Removes backport and release-notes scripts

tmarkley:
### Description
* Deletes the unnecessary `@osd/release-notes` package.
* Backport is unnecessary since we're adding a backport action in #1233.
* Release notes are generated with `git-release-notes`, and there's no need to maintain a package dedicated to release notes for now. Future work can be done to integrate existing release notes repositories instead of reinventing the wheel.
* Removes unused `@types/graphql` dependency.
 
### Issues Resolved
N/A
 
### Check List

- [x] Commits are signed per the DCO using --signoff

PR #11770, opened 2026-04-16
Added workspace acl check in workspacePlugin to be consumed by other plugins for serverless support

nishtham-amazon:
### Description                                                                                                                                                                                                    
                                                                                                                                                                                                                     
  Add `authorizeWorkspace` method to `WorkspacePluginStart` contract, enabling any plugin to check if a calling principal has the required permissions on one or more workspaces.                                    
                                                                                                                                                                                                                     
  The method:                                                                                                                                                                                                        
  - Accepts a request, workspace IDs, principal, and optional permission modes (defaults to `['read']`)                                                                                                              
  - Fetches workspace details via the workspace client and checks the principal against the workspace's permission map                                                                                               
  - Uses OR logic — access is granted if the principal matches **any** of the requested permission modes                                                                                                             
  - Skips the ACL check entirely if the caller is a dashboard admin (`isDashboardAdmin`), consistent with the existing pattern in `WorkspaceSavedObjectsClientWrapper`                                               
  - Returns a `WorkspaceAuthResult` with an `authorized` flag and a list of `unauthorizedWorkspaces` if any failed                                                                                                   
                                                                                                                                                                                                                     
  This authorizer can then be used by any plugin (e.g., Alerting, Anomaly Detection, Notifications) to enforce workspace-level access control on their APIs without directly depending on workspace internals.       
                                                                                                                                                                                                                     
  **Changes:**                                                                                                                                                                                                       
  - `server/types.ts` — Added `WorkspaceAuthResult` interface and `authorizeWorkspace` method to `WorkspacePluginStart`                                                                                              
  - `server/plugin.ts` — Implemented `authorizeWorkspace` in the plugin's `start()` return, including dashboard admin bypass                                                                                         
  - `server/index.ts` — Exported `WorkspaceAuthResult`                                                                                                                                                               
                                                                                                                                                                                                                     
  ### Issues Resolved                                                                                                                                                                                                
                                                                                                                                                                                                                     
  <!-- List any issues this PR will resolve. Prefix the issue with the keyword closes, fixes, fix -->                                                                                                                
  <!-- Example: closes #1234 or fixes <Issue_URL> -->                                                                                                                                                                
                                                                                                                                                                                                                     
  ## Screenshot                                                                                                                                                                                                      
                                                                                                                                                                                                                     
  N/A — server-side only change, no UI impact.                                                                                                                                                                       
                                                                                                                                                                                                                     
  ## Testing the changes                                                                                                                                                                                             
                                                                                                                                                                                                                     
  1. Unit tests added in `server/plugin.test.ts` covering:                                                                                                                                                           
     - Empty workspace IDs returns unauthorized                                                                                                                                                                      
     - Principal with matching permission is authorized                                                                                                                                                              
     - Principal without matching permission is unauthorized                                                                                                                                                         
     - Wildcard `*` in users grants access                                                                                                                                                                           
     - OR logic across multiple permission modes                                                                                                                                                                     
     - Workspace not found returns unauthorized                                                                                                                                                                      
     - Workspace with no permissions returns unauthorized                                                                                                                                                            
     - Multiple workspaces with partial access returns unauthorized with list                                                                                                                                        
     - Default permission mode is `read`                                                                                                                                                                             
     - Dashboard admin skips ACL check entirely                                                                                                                                                                      
     - Non-admin performs full ACL check                                                                                                                                                                             
                                                                                                                                                                                                                     
  2. Manual validation:                                                                                                                                                                                              
     - Start OpenSearch Dashboards with workspace and permission control enabled                                                                                                                                     
     - Create a workspace with specific user permissions                                                                                                                                                             
     - From a consuming plugin, call `workspace.authorizeWorkspace(request, [workspaceId], principal, ['library_write', 'library_read'])` and verify the result matches the workspace's permission configuration     
     - Verify dashboard admin users bypass the check                                                                                                                                                                 
                                                                                                                                                                                                                     
  ### Check List                                                                                                                                                                                                     
                                                                                                                                                                                                                     
  - [x] All tests pass                                                                                                                                                                                               
    - [x] `yarn test:jest`                                                                                                                                                                                           


src/plugins/workspace/server/plugin.ts
```diff
@@ -314,6 +314,65 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
 
     return {
       client: this.client as IWorkspaceClientImpl,
+      authorizeWorkspace: async (
+        request: any,
```
huyaboo: isn't this supposed to be of type `OpenSearchDashboardsRequest` from the interface definition provided in this PR?

src/plugins/workspace/server/plugin.ts
```diff
@@ -314,6 +314,65 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
...
+        for (const workspaceId of workspaceIds) {
+          const result = await (this.client as IWorkspaceClientImpl).get({ request }, workspaceId);
+
+          if (!result.success) {
+            this.logger.warn(`Workspace authorization: workspace ${workspaceId} not found`);
+            unauthorizedWorkspaces.push(workspaceId);
+            continue;
+          }
+
+          this.logger.info(`Workspace object fetched: ${JSON.stringify(result)}`);
```
huyaboo: We shouldn't be logging saved objects information

src/plugins/workspace/server/plugin.ts
```diff
@@ -314,6 +314,65 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
...
+
+          const permissions = (result.result as any).permissions as
+            | Record<string, { users?: string[]; groups?: string[] }>
```
huyaboo: https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/core/server/saved_objects/permission_control/acl.ts#L16

Consider importing this type def instead

src/plugins/workspace/server/plugin.ts
```diff
@@ -314,6 +314,65 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
 
     return {
       client: this.client as IWorkspaceClientImpl,
+      authorizeWorkspace: async (
+        request: any,
+        workspaceIds: string | any[],
```
huyaboo: again, isn't this supposed to be just a `string[]`?

src/plugins/workspace/server/plugin.ts
```diff
@@ -314,6 +314,65 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
...
+
+          if (!result.success) {
+            this.logger.warn(`Workspace authorization: workspace ${workspaceId} not found`);
+            unauthorizedWorkspaces.push(workspaceId);
+            continue;
+          }
+
+          this.logger.info(`Workspace object fetched: ${JSON.stringify(result)}`);
+
+          const permissions = (result.result as any).permissions as
```
huyaboo: We should avoid casting to `any`. If the `permissions` field is needed consider asserting as type `WorkspaceAttributeWithPermission`

https://github.com/nishtham-amazon/OpenSearch-Dashboards/blob/898be6bf766eb61d7a40c132ffb4cf8cc7c89899/src/core/types/workspace.ts#L26-L29

src/plugins/workspace/server/types.ts
```diff
@@ -167,6 +167,27 @@ export interface WorkspacePluginSetup {
   client: IWorkspaceClientImpl;
 }
 
+export interface WorkspaceAuthResult {
+  authorized: boolean;
+  unauthorizedWorkspaces?: string[];
+}
```
huyaboo:
```suggestion
export type WorkspaceAuthResult =
  | { authorized: true }
  | { authorized: false; unauthorizedWorkspaces: string[] };
```

src/plugins/workspace/server/plugin.ts
```diff
@@ -314,6 +314,65 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
...
+          if (!hasAccess) {
+            unauthorizedWorkspaces.push(workspaceId);
+          }
+        }
+
+        const authorized = unauthorizedWorkspaces.length === 0;
+        return {
+          authorized,
+          ...(unauthorizedWorkspaces.length > 0 ? { unauthorizedWorkspaces } : {}),
+        };
```
huyaboo:
```suggestion
return unauthorizedWorkspaces.length === 0
  ? { authorized: true }
  : { authorized: false, unauthorizedWorkspaces };
```

src/plugins/workspace/server/plugin.test.ts
```diff
@@ -513,6 +513,207 @@ describe('Workspace server plugin', () => {
...
+            read: { users: ['*'] },
+          },
+        },
+      });
+      const request = httpServerMock.createOpenSearchDashboardsRequest();
+<<<<<<< HEAD
+      const result = await workspaceStart.authorizeWorkspace(request, ['ws-1'], 'any-user', [
+=======
+      const result = await workspaceStart.authorizeWorkspace(request, ['ws-1'], [
+>>>>>>> 7307cb0d42 (Fix PR comments)
```
FriedhelmWS: Merge conflict here, could you resolve these please.

src/plugins/workspace/server/plugin.ts
```diff
@@ -314,6 +316,56 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
...
+        }
+
+        const unauthorizedWorkspaces: string[] = [];
+
+        for (const workspaceId of workspaceIds) {
+          const result = await (this.client as IWorkspaceClientImpl).get({ request }, workspaceId);
+
+          if (!result.success) {
+            this.logger.warn(`Workspace authorization: workspace ${workspaceId} not found`);
+            unauthorizedWorkspaces.push(workspaceId);
```
huyaboo, line 365: Nit: this is a non-existent workspace, not an unauthorized. Consider returning a `404`

huyaboo (dismissed):

wanglam:
Hi @nishtham-amazon , could you fix the failed CI checks?

src/plugins/workspace/server/plugin.ts
```diff
@@ -314,6 +316,56 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
...
+          return { authorized: true as const };
+        }
+
+        if (!workspaceIds.length) {
+          return { authorized: false as const, unauthorizedWorkspaces: [] as string[] };
+        }
+
+        const unauthorizedWorkspaces: string[] = [];
+
+        for (const workspaceId of workspaceIds) {
```
yubonluo: Could we use `Promise.all` to replace a separate call to `this.client.get()` for each workspace ID.

src/plugins/workspace/server/plugin.ts
```diff
@@ -314,6 +316,56 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
 
     return {
       client: this.client as IWorkspaceClientImpl,
+      authorizeWorkspace: async (
+        request: OpenSearchDashboardsRequest,
+        workspaceIds: string[],
+        permissionModes: string[] = ['read']
+      ) => {
+        const principals = this.permissionControl!.getPrincipalsFromRequest(request);
```
yubonluo: Could we not use `!` to forcefully assert that permissionControl is not empty

Hailong-am: should we check isPermissionControlEnabled before perform permission check?

wanglam: `this.permissionControl` would be `undefined` if permission control was disabled. Can we remove the `!` after permissionControl, and return all workspaceIds if the permission control is `undefined`

src/plugins/workspace/server/plugin.ts
```diff
@@ -314,6 +316,56 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
...
+        const { isDashboardAdmin } = getWorkspaceState(request);
+        if (isDashboardAdmin) {
+          this.logger.debug(
+            `Workspace authorization skipped: caller is dashboard admin`
+          );
+          return { authorized: true as const };
+        }
+
+        if (!workspaceIds.length) {
+          return { authorized: false as const, unauthorizedWorkspaces: [] as string[] };
```
yubonluo: `authorized: false` means there are unauthorized workspaces, but `unauthorizedWorkspaces: []` says there are none. Maybe cause confused?

src/plugins/workspace/server/plugin.ts
```diff
@@ -314,6 +316,56 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
...
+        }
+
+        const unauthorizedWorkspaces: string[] = [];
+
+        for (const workspaceId of workspaceIds) {
+          const result = await (this.client as IWorkspaceClientImpl).get({ request }, workspaceId);
+
```
Hailong-am, line 360: can we use Promise.all to fetch workspace detail in parallel

src/plugins/workspace/server/types.ts
```diff
@@ -167,6 +167,24 @@ export interface WorkspacePluginSetup {
...
+   *
+   * @param request - the incoming request (used to scope the saved objects client and extract principals)
+   * @param workspaceIds - workspace IDs to check
+   * @param permissionModes - permission modes to check (defaults to ['read'])
+   * @returns WorkspaceAuthResult with authorized flag and any unauthorized workspace IDs
+   */
+  authorizeWorkspace: (
+    request: OpenSearchDashboardsRequest,
+    workspaceIds: string[],
+    permissionModes?: string[]
```
Hailong-am: nit: we can use `PermissionModeId` instead of string

src/plugins/workspace/server/plugin.ts
```diff
@@ -314,6 +316,56 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
...
+          if (!result.success) {
+            this.logger.warn(`Workspace authorization: workspace ${workspaceId} not found`);
+            unauthorizedWorkspaces.push(workspaceId);
+            continue;
+          }
+
+          const { permissions } = result.result as { permissions?: Permissions };
```
Hailong-am, line 369: if the permissionModes is read, we don't need to verify it again. workspace client get will perform read permission check

src/plugins/workspace/server/plugin.ts
nishtham-amazon: Treating not-found as unauthorized is intentional — returning 404 would leak workspace existence to callers who don't have access. This follows the same pattern as AWS APIs which return 403 for resources the caller can't see.

src/plugins/workspace/server/plugin.ts
nishtham-amazon: this.permissionControl is only set when isPermissionControlEnabled is true (in setupPermission()), so checking !this.permissionControl is effectively the same as checking !isPermissionControlEnabled. If         
  permission control is disabled, this.permissionControl is undefined, and the method returns unauthorized.

src/plugins/workspace/server/plugin.ts
nishtham-amazon: Can add a log but it does serve the intent where if ws is missing, we unauthorize the call downstream. Let me know if this a blocking experience.

src/plugins/workspace/server/plugin.ts
```diff
@@ -314,6 +324,74 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
...
+          this.logger.warn('Workspace authorization called with empty workspaceIds');
+          return { authorized: false, unauthorizedWorkspaces: [] };
+        }
+
+        if (!this.permissionControl) {
+          this.logger.warn(
+            'Workspace authorization: permissionControl not initialized, denying access'
+          );
+          return { authorized: false, unauthorizedWorkspaces: workspaceIds };
+        }
```
Hailong-am, line 349: For case isPermissionControlEnabled disabled in yml, should we bypass and return true as there is no permission check needed?

Hailong-am (approved):

src/plugins/workspace/server/plugin.ts
nishtham-amazon: I can test and add this in followup pr

yubonluo (approved):

src/plugins/workspace/server/plugin.ts
```diff
@@ -314,6 +324,74 @@ export class WorkspacePlugin implements Plugin<WorkspacePluginSetup, WorkspacePl
 
     return {
       client: this.client as IWorkspaceClientImpl,
+      aclEnforceEndpointPatterns: this.aclEnforceEndpointPatterns,
```
SuZhou-Joe, line 327: What is `aclEnforceEndpointPatterns` used for?

PR #11771, opened 2026-04-17
feat(saved-objects): Add pluggable SQLite storage backend

seraphjiang:
## Description

This PR implements a pluggable storage backend for saved objects with SQLite as the first alternative to OpenSearch.

**RFC:** https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11772

### Before / After

|  | Before (OpenSearch) | After (SQLite) |
|--|-------------------|----------------|
| **Disk space** | ~1.4 GB (`.opensearch/` snapshots) | 4 KB (`osd-metadata.db`) |
| **Prerequisites** | Java runtime + OpenSearch binary | None |
| **Startup** | Download snapshot + start cluster + wait for green + run migrations | `yarn start --savedObjects.storage.backend=sqlite` |
| **Time to server ready** | ~30-60s (OpenSearch boot + migrations) | ~3s (immediate, no migrations) |
| **Unit test speed** | Needs running OpenSearch for integration tests | In-memory SQLite (`:memory:`), ~1s for 24 tests |
| **Offline/air-gapped** | ❌ Cannot start without OpenSearch | ✅ Fully standalone |
| **Memory (RSS)** | ~2-3 GB (OSD ~500 MB + OpenSearch ~1.5-2.5 GB) | ~500 MB (OSD only, no JVM) |
| **Desktop app** | ❌ Cannot run standalone | ✅ Enables [dashboards-desktop](https://github.com/opensearch-project/dashboards-desktop/issues/51) — local-first, no server dependency |
| **CI pipelines** | Needs OpenSearch service container | No external dependencies |

### What's in this PR

- **`SavedObjectsStorageBackend` interface** — clean abstraction for pluggable backends (`src/core/server/saved_objects/storage/types.ts`)
- **`SQLiteStorageBackend`** — embedded backend using better-sqlite3 with WAL mode
- **`StorageBackendRepositoryAdapter`** — bridges any backend to `ISavedObjectsRepository`
- **Wired via `setRepositoryFactoryProvider`** — uses the existing repository factory hook, zero changes to `createRepository` or the saved objects client
- **Config support** — `savedObjects.storage.backend: opensearch | sqlite`
- **24 unit tests** covering CRUD, bulk ops, find, namespaces, and workspaces
- **Documentation** — updated [DEVELOPER_GUIDE.md](DEVELOPER_GUIDE.md), [TESTING.md](TESTING.md), added [design doc](docs/saved_objects/pluggable_storage_backend.md), and [module README](src/core/server/saved_objects/storage/README.md)
- **Zero changes** to default OpenSearch behavior

## When to Use SQLite vs OpenSearch

**Works with SQLite (no OpenSearch needed):** UI/frontend development, workspace features, saved object CRUD, plugin scaffolding, application chrome/navigation, management pages, theming/a11y/i18n, unit tests with in-memory SQLite.

**Still needs OpenSearch:** Data exploration (Discover), visualizations with real data, index pattern field discovery, security plugin, alerting, observability, performance testing.

## How to Use

```bash
# Start OSD with SQLite backend — no OpenSearch needed
yarn start --savedObjects.storage.backend=sqlite --workspace.enabled=true --no-base-path
```

## Demo

**Workspace creation form (SQLite backend, no OpenSearch):**

Workspace create form

**After workspace created successfully:**

Workspace created

## Revision History

| Rev | Description |
|-----|-------------|
| 1 | Initial implementation with SQLite backend and StorageBackend interface |
| 2 | Fix prettier formatting, import order, and ESLint errors |
| 3 | Move better-sqlite3 to optionalDependencies for CI compatibility |
| 4 | Add `(MIT OR WTFPL)` to license allowlist for expand-template transitive dep |
| 5 | **Refactor:** Wire SQLite via `setRepositoryFactoryProvider` instead of modifying `createRepository` directly (feedback from @SuZhou-Joe) |
| 6 | **Bug fixes** from @joshuali925 review: fix bare catch in `checkConflicts`, paginate `deleteByNamespace/Workspace`, preserve version on create-overwrite |
| 7 | **Self-review fixes:** fix bare catch in `incrementCounter`, fix `bulkCreate` version hardcode, handle `type='*'` wildcard in `find()` |
| 8 | Remove RFC from PR (posted separately as #11772), add demo screenshots |
| 9 | Add module README convention to DEVELOPER_GUIDE.md, add storage README.md |
| 10 | Fix delete pagination (always fetch page 1 after deletes), document namespace no-ops |

## Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11772

### CI Notes

**Code-Diff-Analyzer:** Flags `better-sqlite3` for downloading prebuilt native binaries. This is the same pattern used by `lmdb`, `msgpackr-extract`, `@parcel/watcher`, and `sass-embedded` already in the dependency tree. Requires `skip-diff-analyzer` label from a maintainer.

### Check List
- [x] All tests pass locally (24/24)
- [x] New functionality has been documented
- [ ] Commits are signed


seraphjiang:
## Demo: OSD running with SQLite backend (no OpenSearch)

### How it was run
```bash
node scripts/opensearch_dashboards --dev \\
  --savedObjects.storage.backend=sqlite \\
  --workspace.enabled=true \\
  --no-base-path
```

All saved objects (workspace metadata) stored in `data/osd-metadata.db` via SQLite. Zero OpenSearch dependency.

### Workspace creation form
Workspace create form

### After workspace created
Workspace created

### Video demo
A [1 MB webm recording](https://github.com/user-attachments/assets/97a5fc30-9ac9-458b-a798-721ba59ab4af) is available showing the full flow: open create page → fill name/description → select use
 case → create workspace → redirected to new workspace.

src/core/server/saved_objects/saved_objects_service.ts
SuZhou-Joe, line 1: Should we use customized repository factory instead of modifying saved objects client? It is more like a storage change, saved objects client is not necessary to be aware of that.

seraphjiang: Great point — agreed. The saved objects client shouldn't need to know about the storage backend.

Refactored in the latest commit to use `setRepositoryFactoryProvider` instead. The SQLite backend is now wired through the existing repository factory provider hook during `setup()`:

```typescript
if (this.config.storage.backend === 'sqlite') {
  const sqliteBackend = new SQLiteStorageBackend(this.config.storage.sqlite.path);
  this.respositoryFactoryProvider = () => new StorageBackendRepositoryAdapter(sqliteBackend);
}
```

The `createRepository` function is completely untouched — it naturally picks up the factory provider. The only remaining changes in the service are:
- Initializing the SQLite backend in `start()`
- Setting status to available (no OpenSearch dependency)
- Skipping OpenSearch migrations
- Shutting down in `stop()`

ZilongX (dismissed):

docs/saved_objects/img/sqlite-workspace-create-form.png
joshuali925, line 1: do we need images in version control? seems unused and they will permanently increase `.git` size

RFC-alternative-storage-backends.md
```diff
@@ -0,0 +1,382 @@
+# [RFC] Pluggable Storage Backend for OpenSearch Dashboards Saved Objects
```
joshuali925: good content but do we need rfc as source code?

joshuali925:
### Code review

Found 3 issues:

1. `checkConflicts()` uses a bare `catch` and treats **any** failure (DB unavailable, I/O error, permission issue) as "not found = no conflict". Real failures are silently masked and creates can proceed despite an existing object. Only `NotFound` should be swallowed; other errors should propagate.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/55c348df4cacb3f8b2ef0eda31521435d8b6e5fc/src/core/server/saved_objects/storage/repository_adapter.ts#L104-L110

2. `deleteByNamespace()` and `deleteByWorkspace()` hard-code `perPage: 10000` with no pagination loop and never consult `result.total`. If a namespace/workspace holds more than 10k objects, only the first page is deleted — silent data retention after a delete-everything call.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/55c348df4cacb3f8b2ef0eda31521435d8b6e5fc/src/core/server/saved_objects/storage/repository_adapter.ts#L118-L132

3. `create()` (and `createSync()`) uses `INSERT OR REPLACE` with `version` hardcoded to `1`. When `overwrite: true` is used to update an existing row, the version is reset to 1 and the prior `updated_at`/version history is discarded, breaking the optimistic-concurrency contract that `update()` honors via `newVersion = existing.version + 1`.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/55c348df4cacb3f8b2ef0eda31521435d8b6e5fc/src/core/server/saved_objects/storage/sqlite_backend.ts#L148-L158

RFC-alternative-storage-backends.md
```diff
@@ -0,0 +1,382 @@
...
+// Plugin authors don't need to understand OpenSearch mappings.
+// Just register a type and use the client — the backend handles the rest.
+core.savedObjects.registerType({
+  name: 'my-plugin-config',
+  hidden: false,
+  namespaceType: 'single',
+  mappings: {
+    properties: {
+      setting1: { type: 'keyword' },
+      setting2: { type: 'integer' },
```
joshuali925: > // Plugin authors don't need to understand OpenSearch mappings.
// Just register a type and use the client — the backend handles the rest.

this is still opensearch mapping?

src/core/server/saved_objects/saved_objects_service.ts
```diff
@@ -460,12 +483,16 @@ export class SavedObjectsService
...
      */
-    const skipMigrations = this.config.migration.skip || !pluginsInitialized;
+    const skipMigrations = useSqliteBackend || this.config.migration.skip || !pluginsInitialized;
 
     if (skipMigrations) {
-      this.logger.warn(
-        'Skipping Saved Object migrations on startup. Note: Individual documents will still be migrated when read or written.'
-      );
+      if (useSqliteBackend) {
+        this.logger.info('Skipping OpenSearch migrations — using SQLite storage backend.');
```
joshuali925, line 499: we don't need migrations because sqlite table has fixed schema, and column types will never change? since attributes is text

src/core/server/saved_objects/storage/sqlite_backend.ts
```diff
@@ -0,0 +1,478 @@
...
+      options.permissions !== undefined
+        ? JSON.stringify(options.permissions)
+        : existing.permissions;
+    const newVersion = existing.version + 1;
+
+    db.prepare(
+      `UPDATE saved_objects
+       SET attributes = ?, "references" = ?, version = ?, updated_at = ?, workspaces = ?, permissions = ?
+       WHERE type = ? AND id = ? AND namespace = ?`
+    ).run(JSON.stringify(merged), refs, newVersion, now, ws, perms, type, id, namespace);
```
joshuali925: unlikely for sqlite use case, but i think this can have race condition. asked claude it suggested
```ts
const expectedVersion = options.version ? Number(options.version) : existing.version;

const result = db.prepare(
  `UPDATE saved_objects
      SET attributes = ?, "references" = ?, version = version + 1, updated_at = ?, ...
    WHERE type = ? AND id = ? AND namespace = ? AND version = ?`
).run(..., expectedVersion);

if (result.changes === 0) {
  throw SavedObjectsErrorHelpers.createConflictError(type, id);
}
```

seraphjiang:
Thanks for the review @joshuali925 — all 3 issues are valid. Fixed in a5f0826:

1. **checkConflicts**: Now only swallows `NotFoundError` — other errors (DB unavailable, I/O, permissions) propagate correctly.

2. **deleteByNamespace/deleteByWorkspace**: Replaced hardcoded `perPage: 10000` with a pagination loop (`perPage: 1000`) that continues until all objects are deleted.

3. **create with overwrite**: Now reads the existing version and increments it (`version + 1`) instead of resetting to 1, preserving the optimistic concurrency contract consistent with `update()`.

seraphjiang:
Thanks again @joshuali925 — wanted to share the reasoning on how I triaged the 3 issues:

**#1 `checkConflicts` bare catch — Critical ✅**
Real bug. Silently masking DB errors (I/O, permissions) could allow duplicate objects. Fixed immediately.

**#2 `deleteByNamespace/Workspace` 10k limit — Nice-to-have, but fixed ✅**
In practice, no namespace/workspace will have 10k+ saved objects in a SQLite dev environment. That said, the fix was small so I included it. (Interestingly, the same `perPage: 9999` pattern without pagination exists in other repository implementations in production.)

**#3 `create` overwrite resets version — Nice-to-have, but fixed ✅**
`create` with `overwrite: true` is rarely used — most updates go through `update()` which already handled versioning correctly. But the fix was a few lines, so included.

**Additional self-review fixes (3a4ccb0):**
While reviewing against other repository implementations, I found 3 more issues:
- `incrementCounter` had the same bare-catch bug as #1
- `bulkCreate`'s internal `createSync` had the same version=1 hardcode as #3
- `find()` with `type='*'` (used by deleteByNamespace/Workspace) matched nothing because it was treated as a literal type string

All fixed. This is Phase 1 / proof of concept — the interface and abstraction layer are the important parts. Edge case hardening will continue as the interface stabilizes.

joshuali925 (dismissed):
nice

seraphjiang:
### Code-Diff-Analyzer: `better-sqlite3` supply chain flag

The diff analyzer flags `better-sqlite3` as high severity because it uses `prebuild-install` to download prebuilt native binaries. This is the same pattern used by native addons **already in the OSD dependency tree**:

| Package | How it got in | Binary download mechanism |
|---------|--------------|--------------------------|
| `lmdb@2.8.5` | `@osd/optimizer` | `node-gyp-build-optional-packages` |
| `msgpackr-extract@3.0.2` | `@osd/optimizer → lmdb → msgpackr` | `prebuildify` |
| `@parcel/watcher@2.5.6` | `@tailwindcss/cli` | `prebuildify` |
| `sass-embedded` | direct dep | Downloads Dart binary |
| **`better-sqlite3@12.9.0`** | **this PR (optionalDependencies)** | `prebuild-install` |

These existing deps were never flagged because they entered as transitive dependencies in larger PRs.

`better-sqlite3` is a widely-used package (5M+ weekly npm downloads), maintained actively, and declared as `optionalDependencies` so install succeeds even if compilation fails.

**Requesting maintainer to add the `skip-diff-analyzer` label** to unblock this PR.

src/core/server/saved_objects/saved_objects_service.ts
```diff
@@ -460,12 +483,16 @@ export class SavedObjectsService
      * We also cannot safely run migrations if plugins are not initialized since
      * not plugin migrations won't be registered.
      */
-    const skipMigrations = this.config.migration.skip || !pluginsInitialized;
+    const skipMigrations = useSqliteBackend || this.config.migration.skip || !pluginsInitialized;
```
ruanyl, line 495: I thoguht migration is storage backend-agnostic, no?
Look at the current migrations we have in codebase, the vast majority of migrations fix the internal structure of the JSON blobs stored in attributes like visState, panelsJSON, searchSourceJSON, these are application-level data schema evolution.

ruanyl (commented):
The RFC mentioned:
> Existing hook: OSD has setRepositoryFactoryProvider() which allows plugins to provide a custom repository. However, ISavedObjectsRepository is a structural type derived from the concrete class — not a clean, documented interface. No reference implementation exists.

It seem `ISavedObjectsRepository` already exists as a clean interface (`Pick<SavedObjectsRepository, keyof SavedObjectsRepository>`), so everything above the repository (routes, client, wrappers) only depends on this interface.

In the proposed architecture, it mentions `SavedObjectsRepository` will be "refactored: orchestration only Delegates storage to backend", but I didn't see it in this implementation, I think this is a **critical** architecture change(which I believe is the main point of the this PR, otherwise, `setRepositoryFactoryProvider` perhaps is the way to go):
```
┌─────────────────────────────────────────┐
│         SavedObjectsClient              │  ← No changes (stable plugin API)
├─────────────────────────────────────────┤
│    SavedObjectsClientProvider           │  ← No changes
├─────────────────────────────────────────┤
│      SavedObjectsRepository             │  ← Refactored: orchestration only
│      (validation, serialization, ACL)   │     Delegates storage to backend
├──────────┬──────────┬───────────────────┤
│OpenSearch│  SQLite  │  Community        │  ← StorageBackend implementations
│ Backend  │  Backend │  Backends         │     (PostgreSQL, MySQL, DynamoDB...)
│ (default)│(built-in)│  (via plugins)    │
└──────────┴──────────┴───────────────────┘
```

**More importantly**, I think:

###  The abstraction boundary is at the wrong level
The PR introduces `SavedObjectsStorageBackend` as a new interface, then wraps it with `StorageBackendRepositoryAdapter` to conform to `ISavedObjectsRepository`. This creates a two-layer indirection:
```
Routes → Client → Adapter → StorageBackend → SQLite
```
The problem is that `ISavedObjectsRepository` already is the storage abstraction. It's the interface that everything above it depends on. Introducing a second, thinner interface (`StorageBackend`) means you have to maintain semantic parity between two abstractions that represent the same thing. 

So my question here remains: why not implement `ISavedObjectsRepository` directly as `SqliteSavedObjectsRepository`?

seraphjiang:
Thanks @ruanyl — both points are valid. Implemented the refactor.

### Abstraction boundary — fixed

Replaced `StorageBackend` + `Adapter` with `SqliteSavedObjectsRepository implements ISavedObjectsRepository` directly. This gives full access to `_migrator`, `_registry`, `_serializer`, `_allowedTypes` — same pattern as the OpenSearch repository.

**Removed:** `types.ts`, `repository_adapter.ts`, `sqlite_backend.ts` (~400 lines of adapter/interface code)
**Added:** `sqlite_repository.ts` (direct implementation)

### Migrations — fixed

`_migrator.migrateDocument()` is now called on all read/write paths:
- `create()`, `bulkCreate()` — migrate before writing
- `get()`, `bulkGet()`, `find()` — migrate on read
- `update()`, `incrementCounter()` — migrate merged attributes before writing

5 dedicated migration tests verify transforms work on each path.

### Updated docs

- Design doc rewritten for current architecture
- TESTING.md updated (removed references to deleted classes)
- Storage README updated
- RFC issue updated with architecture revision

### Test coverage

32 tests total covering CRUD, bulk ops, find (with hasReference, workspaces, pagination, search), checkConflicts, incrementCounter, bulkUpdate, workspaces, namespaces, deleteByNamespace, deleteByWorkspace, and 5 document migration tests.

src/core/server/saved_objects/saved_objects_service.ts
```diff
@@ -341,6 +343,27 @@ export class SavedObjectsService
       .toPromise();
     this.config = new SavedObjectConfig(savedObjectsConfig, savedObjectsMigrationConfig);
 
+    // Wire up SQLite backend via the existing repository factory provider hook
+    if (this.config.storage.backend === 'sqlite') {
+      this.logger.info('Configuring SQLite storage backend for saved objects');
+      const dbPath = this.config.storage.sqlite.path;
+      const typeRegistry = this.typeRegistry;
+      const logger = this.logger;
+      this.respositoryFactoryProvider = ({ migrator, includedHiddenTypes }) => {
```
ruanyl, line 353: Better to call `setRepositoryFactoryProvider` instead assign directly to `this.respositoryFactoryProvider`

src/core/server/saved_objects/saved_objects_service.ts
```diff
@@ -460,12 +495,16 @@ export class SavedObjectsService
      * We also cannot safely run migrations if plugins are not initialized since
      * not plugin migrations won't be registered.
      */
-    const skipMigrations = this.config.migration.skip || !pluginsInitialized;
+    const skipMigrations = useSqliteBackend || this.config.migration.skip || !pluginsInitialized;
```
ruanyl, line 495: Do we still need to skip migrations for SQLite backend?

src/core/server/saved_objects/saved_objects_service.ts
```diff
@@ -409,6 +432,13 @@ export class SavedObjectsService
 
     this.logger.debug('Starting SavedObjects service');
 
+    const useSqliteBackend = this.config.storage.backend === 'sqlite';
+
+    // SQLite backend is initialized lazily via the factory provider
+    if (useSqliteBackend) {
+      this.logger.info('Using SQLite storage backend for saved objects');
+    }
```
ruanyl: Nit: Move these log output to `SqliteSavedObjectsRepository` class since it's SQLite-specific log?
```suggestion
    const useSqliteBackend = this.config.storage.backend === 'sqlite';

    // SQLite backend is initialized lazily via the factory provider
    if (useSqliteBackend) {
      this.logger.info('Using SQLite storage backend for saved objects');
    }
```

src/core/server/saved_objects/saved_objects_service.ts
```diff
@@ -341,6 +343,27 @@ export class SavedObjectsService
...
+        const serializer = new SavedObjectsSerializer(typeRegistry);
+        const repo = new SqliteSavedObjectsRepository({
+          dbPath,
+          migrator,
+          typeRegistry,
+          serializer,
+          includedHiddenTypes,
+          logger,
+        });
+        this.sqliteRepository = repo;
```
ruanyl: Should `SqliteSavedObjectsRepository` be a singleton? because currently, every time `createRepository` is called, a new `SqliteSavedObjectsRepository` instance is create and `this.sqliteRepository` is update with the new instance, however, in `stop()` function, `this.sqliteRepository.shutdown();` always call `showdown` on latest instance, would there be a problem?

ruanyl (dismissed):

src/core/server/saved_objects/saved_objects_service.ts
seraphjiang: Good point. Since the wiring happens inside `setup()` before the return, we can't call the public method. But it's the same internal assignment. Added a comment clarifying this.

src/core/server/saved_objects/saved_objects_service.ts
seraphjiang: We skip **index-level** migrations (`migrator.runMigrations()` — creating `.kibana` index, mapping changes) which are OpenSearch-specific. Document-level migrations (`_migrator.migrateDocument()`) still run on every read/write path inside `SqliteSavedObjectsRepository`. So the `skipMigrations` flag is correct here — it only controls the index migration, not document migration.

src/core/server/saved_objects/saved_objects_service.ts
seraphjiang: Good nit — moved the SQLite-specific log into the repository constructor where it belongs.

ruanyl (dismissed):

src/core/server/saved_objects/storage/sqlite_repository.ts
```diff
@@ -0,0 +1,803 @@
...
+
+  // --- CRUD ---
+
+  async create<T = unknown>(
+    type: string,
+    attributes: T,
+    options: SavedObjectsCreateOptions = {}
+  ): Promise<SavedObject<T>> {
+    const id = options.id || uuid.v4();
+    const namespace = options.namespace || '';
```
SuZhou-Joe, line 163: Not related to this PR, but we should start an issue to remove those namespace code, they are not needed anymore and is complicating repository code.

src/core/server/saved_objects/storage/sqlite_repository.ts
```diff
@@ -0,0 +1,803 @@
...
+      // Atomic upsert: increment version if exists, else insert with version 1
+      this.db
+        .prepare(
+          `INSERT INTO saved_objects
+           (type, id, namespace, attributes, "references", version, updated_at, migration_version, origin_id, workspaces, permissions)
+           VALUES (?, ?, ?, ?, ?, 1, ?, ?, ?, ?, ?)
+           ON CONFLICT(type, id, namespace) DO UPDATE SET
+             attributes = excluded.attributes,
+             "references" = excluded."references",
+             version = saved_objects.version + 1,
```
SuZhou-Joe, line 200: Not quite understand how the version prevents race condition for update.

src/core/server/saved_objects/storage/sqlite_repository.ts
```diff
@@ -0,0 +1,803 @@
...
+    const refs = JSON.stringify(migrated.references || []);
+    const migVer = migrated.migrationVersion ? JSON.stringify(migrated.migrationVersion) : null;
+    const ws = options.workspaces ? JSON.stringify(options.workspaces) : null;
+    const perms = options.permissions ? JSON.stringify(options.permissions) : null;
+
+    if (options.overwrite) {
+      // Atomic upsert: increment version if exists, else insert with version 1
+      this.db
+        .prepare(
+          `INSERT INTO saved_objects
```
SuZhou-Joe, line 194: I suppose this repository may not support multi tenant, could we add some content in the markdown file?

src/core/server/saved_objects/storage/sqlite_repository.ts
```diff
@@ -0,0 +1,803 @@
...
+
+          let version = 1;
+          const refs = JSON.stringify(migrated.references || []);
+          const migVer = migrated.migrationVersion
+            ? JSON.stringify(migrated.migrationVersion)
+            : null;
+          const ws = obj.workspaces ? JSON.stringify(obj.workspaces) : null;
+
+          if (options.overwrite) {
+            this.db
```
SuZhou-Joe, line 321: Seems we are duplicating code here, shall we reuse single create method or having a new atomic create method?

SuZhou-Joe (dismissed):

src/core/server/saved_objects/storage/sqlite_repository.ts
seraphjiang: Agreed — that's a good cleanup. Filed as a separate issue would make sense. Out of scope for this PR.

src/core/server/saved_objects/storage/sqlite_repository.ts
seraphjiang: The duplication exists because `bulkCreate` runs inside `db.transaction()` (all-or-nothing for the batch), while `create()` is a standalone operation. Calling `create()` from inside the transaction would open a nested transaction which SQLite doesn't support. Could extract a shared private helper, but kept it 
  explicit for readability

src/core/server/saved_objects/storage/sqlite_repository.ts
seraphjiang: The `WHERE version = ?` makes the UPDATE atomic — if two concurrent updates both read version=5, the first one succeeds (changes version to 6), the second one's `WHERE version = 5` matches zero rows (`result.changes === 0`), so it throws a conflict error instead of silently overwriting. It's optimistic        
  concurrency control — the same pattern OpenSearch uses with `_seq_no` / `_primary_term`.        

Here's a clearer explanation with a concrete example:                                                                                                                                                                                                                                                                    
   The `WHERE version = ?` makes the UPDATE atomic via optimistic concurrency control.                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                           
  **Example:** A dashboard exists with `version = 5`. Two requests update it simultaneously:                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                           
  **With `AND version = ?` (current implementation):**                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                           
  | Step | Request A | Request B |                                                                                                                                                                                                                                                                                         
  |------|-----------|-----------|                                                                                                                                                                                                                                                                                         
  | 1 | `SELECT * WHERE id='dash-1'` → reads version=5 | |                                                                                                                                                                                                                                                                 
  | 2 | | `SELECT * WHERE id='dash-1'` → reads version=5 |                                                                                                                                                                                                                                                                 
  | 3 | `UPDATE SET version=6 WHERE id='dash-1' AND version=5` → ✅ 1 row changed | |                                                                                                                                                                                                                                      
  | 4 | | `UPDATE SET version=6 WHERE id='dash-1' AND version=5` → ❌ 0 rows changed |                                                                                                                                                                                                                                     
  | 5 | | `result.changes === 0` → throws `ConflictError` |                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                           
  Request B fails safely because the version guard detects the row was already modified.                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                           
  **Without `AND version = ?` (unsafe):**                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                           
  | Step | Request A | Request B |                                                                                                                                                                                                                                                                                         
  |------|-----------|-----------|                                                                                                                                                                                                                                                                                         
  | 3 | `UPDATE SET version=6 WHERE id='dash-1'` → ✅ | |                                                                                                                                                                                                                                                                  
  | 4 | | `UPDATE SET version=6 WHERE id='dash-1'` → ✅ silently overwrites A's changes 💥 |                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                           
  Same concept as OpenSearch's `_seq_no` / `_primary_term` and HTTP's `If-Match` / ETag.

src/core/server/saved_objects/storage/sqlite_repository.ts
seraphjiang: Good point — will add a note in seperate PR. this one getting larger

src/core/server/saved_objects/saved_objects_config.ts
```diff
@@ -65,6 +65,14 @@ export const savedObjectsConfig = {
     permission: schema.object({
       enabled: schema.boolean({ defaultValue: false }),
     }),
+    storage: schema.object({
+      backend: schema.oneOf([schema.literal('opensearch'), schema.literal('sqlite')], {
```
huyaboo: For other plugins to register a backend implementation for a storage service of their choice, this wouldn't be extensible right? It would be better to resolve the backend service at service bootstrap to enable plugins to register their own custom repositories

ZilongX:
@copilot resolve the merge conflicts in this pull request

ZilongX:
@copilot resolve the merge conflicts in this pull request

seraphjiang:
> @copilot resolve the merge conflicts in this pull request

copolit not working, asked kiro agent to fix  ~

ruchidh (approved):

LDrago27 (approved):

PR #11775, opened 2026-04-17
[Explore] add metrics builder to discover metrics

joshuali925:
### Description

  This PR adds builder mode (default) to discover metrics, where user can switch between builder and code.

  This PR also contains other minor UI updates.

## Screenshot

image

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/pages/metrics/explore/components/loading_state.tsx
+++ b/src/plugins/explore/public/application/pages/metrics/explore/components/loading_state.tsx
@@ -24,7 +24,7 @@ export function ErrorCallout({ error }: { error: string }): JSX.Element {
       iconColor="danger"
       title={
         <h2>
-          {i18n.translate('explore.metricsExplore.errorTitle', {
+          {i18n.translate('explore.metricsExplore.unableToLoadMetrics', {
             defaultMessage: 'Unable to load metrics',
           })}
         </h2>
--- a/src/plugins/explore/public/application/pages/metrics/explore/components/loading_state.tsx
+++ b/src/plugins/explore/public/application/pages/metrics/explore/components/loading_state.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { EuiCallOut, EuiFlexGroup, EuiFlexItem, EuiLoadingChart } from '@elastic/eui';
+import { EuiEmptyPrompt, EuiFlexGroup, EuiFlexItem, EuiLoadingChart, EuiText } from '@elastic/eui';
 import { i18n } from '@osd/i18n';
 
 export function LoadingIndicator(): JSX.Element {
@@ -17,13 +17,29 @@ export function LoadingIndicator(): JSX.Element {
 }
 
 export function ErrorCallout({ error }: { error: string }): JSX.Element {
+  const message = error.replace(/^Error:\s*/, '');
   return (
-    <EuiCallOut
-      title={i18n.translate('explore.metricsExplore.errorTitle', { defaultMessage: 'Error' })}
-      color="danger"
+    <EuiEmptyPrompt
       iconType="alert"
-    >
-      {error}
-    </EuiCallOut>
+      iconColor="danger"
+      title={
+        <h2>
+          {i18n.translate('explore.metricsExplore.errorTitle', {
+            defaultMessage: 'Unable to load metrics',
+          })}
+        </h2>
+      }
+      body={
+        <EuiText size="s" color="subdued">
+          <p>
+            {i18n.translate('explore.metricsExplore.errorBody', {
+              defaultMessage:
+                'We encountered a problem fetching data from the Prometheus data source. Check your connection and try again.',
+            })}
+          </p>
+          <p style={{ fontFamily: 'monospace', wordBreak: 'break-word' }}>{message}</p>
+        </EuiText>
+      }
+    />
   );
 }
--- a/src/plugins/explore/public/application/pages/metrics/explore/components/metric_card.tsx
+++ b/src/plugins/explore/public/application/pages/metrics/explore/components/metric_card.tsx
@@ -13,6 +13,8 @@ import {
   EuiToolTip,
   EuiLoadingChart,
   EuiButtonEmpty,
+  EuiIcon,
+  EuiText,
 } from '@elastic/eui';
 import { darkMode } from '@osd/ui-shared-deps/theme';
 import { i18n } from '@osd/i18n';
@@ -24,6 +26,7 @@ interface MetricCardProps {
   name: string;
   metadata?: MetricMetadata;
   sparkline: Array<[number, string]> | null;
+  sparklineError?: string | null;
   isSelected: boolean;
   colorIndex: number;
   onToggleSelect: () => void;
@@ -35,6 +38,7 @@ export const MetricCard: React.FC<MetricCardProps> = ({
   name,
   metadata,
   sparkline,
+  sparklineError,
   isSelected,
   colorIndex,
   onToggleSelect,
@@ -67,6 +71,30 @@ export const MetricCard: React.FC<MetricCardProps> = ({
   const chartColor = SERIES_COLORS[colorIndex % SERIES_COLORS.length];
 
   function renderSparklineContent() {
+    if (sparklineError) {
+      return (
+        <EuiToolTip content={sparklineError}>
+          <EuiFlexGroup
+            direction="column"
+            gutterSize="xs"
+            alignItems="center"
+            responsive={false}
+            style={{ textAlign: 'center' }}
+          >
+            <EuiFlexItem grow={false}>
+              <EuiIcon type="alert" color="danger" size="m" />
+            </EuiFlexItem>
+            <EuiFlexItem grow={false}>
+              <EuiText size="xs" color="danger">
+                {i18n.translate('explore.metricsExplore.sparklineError', {
+                  defaultMessage: 'Query failed',
+                })}
+              </EuiText>
+            </EuiFlexItem>
+          </EuiFlexGroup>
+        </EuiToolTip>
+      );
+    }
     if (sparkline)
       return (
         <SparklineChart
@@ -158,7 +186,7 @@ export const MetricCard: React.FC<MetricCardProps> = ({
             minHeight: 160,
             display: 'flex',
             alignItems: 'center',
-            justifyContent: !sparkline ? 'center' : undefined,
+            justifyContent: !sparkline || sparklineError ? 'center' : undefined,
             width: '100%',
           }}
         >
--- a/src/plugins/explore/public/application/pages/metrics/explore/components/metric_card.tsx
+++ b/src/plugins/explore/public/application/pages/metrics/explore/components/metric_card.tsx
@@ -112,10 +112,11 @@ export const MetricCard: React.FC<MetricCardProps> = ({
   return (
     <div ref={cardRef}>
       <EuiPanel
-        paddingSize="s"
+        paddingSize="none"
         hasBorder
         onClick={onToggleSelect}
         style={{
+          padding: 6,
           cursor: 'pointer',
           height: '100%',
           display: 'flex',
--- a/src/plugins/explore/public/application/pages/metrics/explore/components/metric_detail.tsx
+++ b/src/plugins/explore/public/application/pages/metrics/explore/components/metric_detail.tsx
@@ -166,7 +166,16 @@ const BreakdownPanel: React.FC<{
             <EuiTitle size="xxs">
               <h3>{s.labelValue}</h3>
             </EuiTitle>
-            <div style={{ marginTop: 4 }}>
+            <div
+              style={{
+                marginTop: 4,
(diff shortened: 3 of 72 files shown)


joshuali925:
### Code review

Found 2 issues:

1. Hard-coded `cy.wait()` delays in Cypress tests (CLAUDE.md says "Use `cy.intercept()` instead of hard-coded delays")

Multiple instances of `cy.wait(1000)` used after page navigation instead of intercepting network requests or using Cypress retry-ability:

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a3dcb78ec0eb341c0593917028ce58fe35b6a4b1/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/14/prometheus_dataset.spec.js#L183-L185

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a3dcb78ec0eb341c0593917028ce58fe35b6a4b1/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/14/prometheus_dataset.spec.js#L417-L419

2. Use of `any` type in production code (CLAUDE.md says "Avoid `any` (use `unknown` or generics)")

Several instances across new files:

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a3dcb78ec0eb341c0593917028ce58fe35b6a4b1/src/plugins/explore/public/application/pages/metrics/explore/services/prometheus_client.ts#L57-L59

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a3dcb78ec0eb341c0593917028ce58fe35b6a4b1/src/plugins/explore/public/application/pages/metrics/explore/services/prometheus_client.ts#L201-L203

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a3dcb78ec0eb341c0593917028ce58fe35b6a4b1/src/plugins/explore/public/application/pages/metrics/explore/components/metric_detail.tsx#L243-L245

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a3dcb78ec0eb341c0593917028ce58fe35b6a4b1/src/plugins/explore/public/application/pages/metrics/query_panel/suggestion_provider.ts#L39-L41

ps48 (commented):
Nice work on this substantial feature PR — the ANTLR-based PromQL parser/builder round-trip is well designed, and the test coverage for the new modules is solid. Left a few inline comments, mostly nits with code suggestions.

src/plugins/explore/public/application/utils/query_assist/promql_tool_handlers.ts
```diff
@@ -153,8 +153,12 @@ export class PromQLToolHandlers {
       );
 
       if (args.query) {
+        // Split multi-word queries into individual terms joined by OR (|)
```
ps48, line 156: **Bug:** When `terms.length <= 1`, the raw unsanitized `args.query` is passed to `new RegExp()` instead of the escaped version from `escapeRegexForPrometheus`. A crafted ReDoS pattern (e.g. `(a+)+$`) in the query argument could hang the event loop.

```suggestion
        const terms = args.query.split(/\s+/).filter(Boolean).map(escapeRegexForPrometheus);
        const pattern = terms.length > 1 ? terms.join('|') : terms[0] || args.query;
```

src/plugins/explore/public/application/pages/metrics/explore/components/metrics_empty_state.tsx
```diff
@@ -0,0 +1,125 @@
...
+          <EuiText size="s" color="subdued">
+            <p>{body}</p>
+          </EuiText>
+          <EuiSpacer size="m" />
+          <EuiPanel paddingSize="m" color="subdued" hasShadow={false} hasBorder={false}>
+            <EuiTitle size="xxs">
+              <h3>
+                {i18n.translate('explore.metricsExplore.sampleQueriesTitle', {
+                  defaultMessage: 'Sample PromQL queries',
+                })}
```
ps48, line 72: **Nit (a11y):** `<div role="button" tabIndex={0}>` with manual `onKeyDown` is an anti-pattern — a native `<button>` provides Enter/Space handling, focus ring, and screen reader semantics for free.

```suggestion
                    <button
                      type="button"
                      className="metricsEmptyState__sampleTrigger"
                      onClick={() => onSelectQuery(sample.query)}
                      aria-label={i18n.translate('explore.metricsExplore.useSampleQuery', {
                        defaultMessage: 'Use sample query: {label}',
                        values: { label: sample.label },
                      })}
                      style={{ all: 'unset', display: 'block', width: '100%', cursor: 'pointer' }}
                    >
```

src/plugins/explore/public/application/pages/metrics/explore/components/metric_card.tsx
```diff
@@ -84,10 +112,11 @@ export const MetricCard: React.FC<MetricCardProps> = ({
   return (
     <div ref={cardRef}>
       <EuiPanel
-        paddingSize="s"
+        paddingSize="none"
         hasBorder
         onClick={onToggleSelect}
         style={{
```
ps48, line 118: **Nit:** `paddingSize="none"` + inline `style={{ padding: 6 }}` bypasses the EUI spacing scale. 6px doesn't correspond to any EUI token. Consider keeping `paddingSize="s"` (8px) which is on-scale, or at minimum using the `$euiSizeXS` (4px) token via a CSS class.

```suggestion
        paddingSize="s"
        hasBorder
        onClick={onToggleSelect}
        style={{
```

src/plugins/explore/public/application/pages/metrics/promql_builder/promql_builder.tsx
```diff
@@ -0,0 +1,193 @@
...
+        <EuiFlexItem grow={false}>
+          <div className="pqbGroup">
+            <span className="pqbGroup__label">
+              {i18n.translate('explore.promqlBuilder.range', { defaultMessage: 'Range' })}
+            </span>
+            <input
+              value={state.range}
+              placeholder="5m"
+              onChange={(e) => dispatch({ type: 'SET_RANGE', range: e.target.value })}
+              className="pqbParamInput"
```
ps48, line 131: **Nit:** Using array index as React key for label filters. Since filters can be added/removed from any position, index keys can cause React to re-use DOM nodes in the wrong order (e.g. combo box focus/selection state getting swapped). Consider giving each `LabelFilter` a stable `id` field, similar to how `QueryRow` has `row.id`.

src/plugins/explore/public/application/pages/metrics/promql_builder/use_metric_data.ts
```diff
@@ -0,0 +1,132 @@
...
+    return () => {
+      clearTimeout(metricSearchTimerRef.current);
+    };
+  }, []);
+
+  const onMetricFocus = useCallback(() => {
+    if (metricNamesLoadedRef.current) return;
+    metricNamesLoadedRef.current = true;
+    setMetricSearchLoading(true);
+    client
```
ps48, line 33: **Nit:** `metricNamesLoadedRef` is never reset when `client` changes. If the user switches Prometheus data sources, the ref stays `true` and the metric dropdown never reloads for the new source.

```suggestion
  const metricSearchTimerRef = useRef<ReturnType<typeof setTimeout>>();
  const metricNamesLoadedRef = useRef(false);
  const loadedLabelsRef = useRef<Set<string>>(new Set());

  // Reset cached flags when the client (data source) changes so the
  // metric dropdown and label cache refresh for the new source.
  useEffect(() => {
    metricNamesLoadedRef.current = false;
    loadedLabelsRef.current = new Set();
  }, [client]);
```

src/plugins/explore/public/application/pages/metrics/promql_builder/promql_parser.ts
```diff
@@ -0,0 +1,446 @@
...
+  [PromQLParser.DEQ]: 'eq',
+  [PromQLParser.NE]: 'neq',
+  [PromQLParser.GT]: 'gt',
+  [PromQLParser.LT]: 'lt',
+  [PromQLParser.GE]: 'gte',
+  [PromQLParser.LE]: 'lte',
+};
+
+const emptyState = (): BuilderState => ({
+  metric: '',
```
ps48, line 108: **Nit:** The eviction here is FIFO (first-inserted key), not true LRU. A frequently parsed query that was inserted early will be evicted before rarely-used later entries. At 64 entries this is unlikely to matter, but a short comment clarifying it's FIFO would prevent a future reader from assuming LRU semantics.

```suggestion
  // FIFO eviction — simple and sufficient at this cache size.
  if (parseCache.size >= PARSE_CACHE_MAX) {
```

src/plugins/explore/public/application/pages/metrics/explore/components/loading_state.tsx
```diff
@@ -17,13 +17,29 @@ export function LoadingIndicator(): JSX.Element {
 }
 
 export function ErrorCallout({ error }: { error: string }): JSX.Element {
+  const message = error.replace(/^Error:\s*/, '');
   return (
-    <EuiCallOut
-      title={i18n.translate('explore.metricsExplore.errorTitle', { defaultMessage: 'Error' })}
-      color="danger"
+    <EuiEmptyPrompt
```
ps48, line 22: **Nit:** The i18n key `explore.metricsExplore.errorTitle` previously had `defaultMessage: 'Error'` and now has `defaultMessage: 'Unable to load metrics'`. If the key has existing translations in any locale catalog, translators won't see that the meaning changed. Consider using a new key (e.g. `explore.metricsExplore.unableToLoadMetrics`) to force re-translation.

src/plugins/explore/public/application/pages/metrics/explore/components/metric_detail.tsx
```diff
@@ -85,10 +115,23 @@ const BreakdownPanel: React.FC<{
     };
```
ps48, line 115: **Nit:** Same pattern as metric_card — `paddingSize="none" style={{ padding: 8 }}` appears in several places in this file. 8px matches `paddingSize="s"` exactly, so you can just use:

```suggestion
        <EuiPanel paddingSize="s" hasBorder>
```

This applies to lines 162, 205, and 210 as well.

src/plugins/explore/public/application/pages/metrics/promql_builder/promql_builder.scss
```diff
@@ -0,0 +1,244 @@
...
+  border: 1px solid $euiBorderColor;
+  border-radius: $euiSizeM;
+  padding: 0 2px;
+  background: $euiColorEmptyShade;
+  margin-top: 2px;
+
+  &__label {
+    position: absolute;
+    top: -0.6em;
+    left: $euiSizeS;
```
ps48, line 97: **Nit:** There are 6 `!important` overrides in this file. You can achieve the same override effect without `!important` by doubling the class selector for higher specificity:

```suggestion
  // Remove all inner borders/shadows — the group border is the only outline
  &.pqbGroup .euiFormControlLayout,
  &.pqbGroup .euiFormControlLayout__childrenWrapper,
  &.pqbGroup .euiComboBox-isOpen .euiFormControlLayout,
  &.pqbGroup .euiComboBox__inputWrap {
    box-shadow: none;
    border: none;
    background: transparent;
  }

  &.pqbGroup .euiSuperSelectControl {
    box-shadow: none;
    border: none;
    background: transparent;
  }
```

src/plugins/explore/public/application/pages/metrics/query_panel/suggestion_provider.ts
```diff
@@ -0,0 +1,71 @@
...
+          // ExploreServices storage type incompatible with IDataPluginServices.DataStorage
+          services: services as any,
+        });
+        const wordUntil = model.getWordUntilPosition(position);
+        const range = new monaco.Range(
+          position.lineNumber,
+          wordUntil.startColumn,
+          position.lineNumber,
+          wordUntil.endColumn
+        );
```
ps48, line 51: **Nit:** The `as any` cast here silently disables all type checking on the services argument. Consider a more targeted cast or a brief adapter that satisfies the expected interface, e.g.:

```ts
services: (services as unknown) as IDataPluginServices,
```

That way you still get compile errors if `IDataPluginServices` changes shape in a way that matters.

src/plugins/explore/public/application/pages/metrics/metrics_page.scss
```diff
@@ -3,7 +3,44 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+// Metrics page: allow full page scrolling instead of internal scroll regions
+$osdHeaderOffset: $euiHeaderHeightCompensation;
```
ps48: **Nit:** `$osdHeaderOffset` is only used once — consider inlining to reduce indirection:

```suggestion
$osdHeaderOffset: $euiHeaderHeightCompensation; // used once below
```

Or just inline `$euiHeaderHeightCompensation` directly in the `calc()` on line 12.

ps48 (dismissed):

lezzago (approved):

ps48 (approved):

PR #9137, opened 2024-12-31
Add identity source service for customizing identity APIs

yubonluo:
### Description

The identity source service enables dynamic identity management by allowing the registration of customized `getIdentityEntries` and `getIdentityEntriesByIds` APIs for different identity sources and types. By specifying a source (e.g., `local_cluster`) and type (e.g., `roles`, `internalusers`), the system intelligently routes API calls to the appropriate implementation, ensuring that identity operations are seamlessly adapted to the selected source. This flexibility simplifies the integration of diverse identity providers, streamlining user and role management across varying environments.

By default, users and roles are retrieved through the identity source registered within the `security-dashboard-plugin`. This ensures a seamless integration with the default identity management system, providing a consistent and secure mechanism for handling user and role data.

### Issues Resolved
N/A

## Screenshot

Only backend changes

## Testing the changes
Assuming that the identity source has already been registered in the security plugins.

```
core.security.registerIdentitySourceHandler('local_cluster', {
      getIdentityEntries: async ({type}, request, context) => {
        try {
          const typeMap: Record<string, string> = {
            internalusers: '/_plugins/_security/api/internalusers',
            roles: '/_plugins/_security/api/roles',
          };

          if (!typeMap[type]) {
            throw new Error(`This type "${type}" does not exist.`);
          }

          const client = context.core.opensearch.client.asCurrentUser;
          const response = await client.transport.request({
            method: 'GET',
            path: typeMap[type],
          });
          const result = Object.keys(response.body).map((name) => ({ name: name, id: name }));
          return result;
        } catch (error) {
          throw new Error(`Failed to get ${type}: ${error.message}`)
        }
      }
    });
```

Get all users
```zsh
curl -X GET -H "osd-xsrf: string" -H "Content-Type: application/json" -d '{}' "http://localhost:5601/api/security/identity/_entries?source=local_cluster&type=internalusers" -u 'admin:myStrongPassword123!' --insecure

[
    {
        "name": "logstash",
        "id": "logstash"
    },
    {
        "name": "test",
        "id": "test"
    },
    {
        "name": "snapshotrestore",
        "id": "snapshotrestore"
    },
    {
        "name": "admin",
        "id": "admin"
    },
    {
        "name": "kibanaserver",
        "id": "kibanaserver"
    },
    {
        "name": "kibanaro",
        "id": "kibanaro"
    },
    {
        "name": "readall",
        "id": "readall"
    },
    {
        "name": "anomalyadmin",
        "id": "anomalyadmin"
    }
]
[...]

PR #11780, opened 2026-04-20
[Workspace] collaborators support name lookup

gaobinlong:
### Description

This is a followup PR for https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9137, after users register their own identity source APIs, they can use these API to search collaborators by their names and show collaborators' name in the list page.

## Screenshot

Add collaborators Dialog:
image

Workspace collaborators list page:

image

## Testing the changes

1. Register identity source APIs in server side,  following https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9137
2. In src/plugins/workspace/public/register_default_collaborator_types.tsx, add identitySource field into each workspaceCollaboratorType and AddCollaboratorModal.

### Check List

- [ ] All tests pass
  - [x] `yarn test:jest`


src/plugins/workspace/public/components/add_collaborators_modal/add_collaborators_modal.tsx
```diff
@@ -30,14 +31,14 @@ import { DuplicateCollaboratorError } from './duplicate_collaborator_error';
 const DUPLICATE_COLLABORATOR_ERROR_MESSAGE = i18n.translate(
   'workspace.addCollaboratorsModal.errors.duplicateCollaborator',
   {
-    defaultMessage: 'A collaborator with this ID already exists.',
+    defaultMessage: 'This identity already exists.',
```
gaobinlong, line 34: Now we can add collaborators by name, not only by ID, this change makes it compatible with both cases.

src/plugins/workspace/public/components/workspace_form/workspace_collaborator_table.tsx
```diff
@@ -160,8 +161,54 @@ export const WorkspaceCollaboratorTable = ({
...
+          groupMap.set(key, { ...identitySource, ids: [] });
+        }
+        groupMap.get(key)!.ids.push(id);
+      });
+
+      const newMap: Record<string, string> = {};
+      await Promise.all(
+        Array.from(groupMap.values()).map(async ({ source, type, ids }) => {
+          try {
+            const resp = await http.post<Array<{ id: string; name: string }>>(
```
wanglam: Can we add a abort controller here to abort the http request when component unmount.

src/plugins/workspace/public/components/workspace_form/workspace_collaborator_table.tsx
```diff
@@ -160,8 +161,54 @@ export const WorkspaceCollaboratorTable = ({
...
+            );
+            resp?.forEach(({ id, name }) => {
+              newMap[id] = name;
+            });
+          } catch {
+            // silently ignore, Name column will remain empty
+          }
+        })
+      );
+      setIdToNameMap(newMap);
```
wanglam: Can we add component unmount check here? I think we should cancel the `setIdToNameMap` after component unmount.

src/plugins/workspace/public/components/workspace_form/workspace_collaborator_table.tsx
```diff
@@ -160,8 +161,54 @@ export const WorkspaceCollaboratorTable = ({
...
+    };
+    fetchNames();
```
wanglam, line 223: The `fetchNames` will be called after every permissionSettings updated. Can we find some ways to call `fetchNames` only after the source, type and ids array modified?

Currently, even when just changing a collaborator's access level (without modifying any IDs), the effect will re-run and make unnecessary API calls to fetch the same names again.

src/plugins/workspace/public/components/workspace_form/workspace_collaborator_table.tsx
wanglam: Can we add component unmount check here? I think we should cancel the `setIdToNameMap` after component unmount.

src/plugins/workspace/public/components/add_collaborators_modal/workspace_collaborator_input.tsx
```diff
@@ -59,6 +82,42 @@ export const WorkspaceCollaboratorInput = ({
...
+              type: identitySource.type,
+              keyword: searchValue,
+            },
+            signal: abortController.current.signal,
+          });
+          setOptions(results.map((r) => ({ label: `${r.name} (${r.id})`, value: r.id })));
+        } catch (e) {
+          if ((e as Error).name !== 'AbortError') setOptions([]);
+        } finally {
+          setIsLoading(false);
```
wanglam: ### Race Condition in handleSearchChange
While the code properly cancels previous requests with AbortController, there's a subtle issue:

```
const handleSearchChange = useCallback(
  (searchValue: string) => {
    if (debounceTimer.current) clearTimeout(debounceTimer.current);
    abortController.current?.abort();

    if (!identitySource || !http || !searchValue || searchValue.length > 1024) {
      setOptions([]);
      return;
    }

    debounceTimer.current = setTimeout(async () => {
      abortController.current = new AbortController();
      setIsLoading(true);
      try {
        const results = await http.get('/api/security/identity/_entries', {
          query: { source: identitySource.source, type: identitySource.type, keyword: searchValue },
          signal: abortController.current.signal,
        });
        setOptions(results.map((r) => ({ label: `${r.name} (${r.id})`, value: r.id })));
      } catch (e) {
        if ((e as Error).name !== 'AbortError') setOptions([]);
      } finally {
        setIsLoading(false);  // ❌ Always sets loading to false
      }
    }, 300);
  },
  [http, identitySource]
);
```
Problem: If a user types rapidly, the finally block of an aborted request will set isLoading(false) even when a newer request is still pending.

Fix: Track the request ID or check if it's still the current request before setting loading state.

src/plugins/workspace/public/components/add_collaborators_modal/workspace_collaborator_input.tsx
```diff
@@ -59,6 +82,42 @@ export const WorkspaceCollaboratorInput = ({
     [index, onCollaboratorIdChange]
   );
 
+  const handleSearchChange = useCallback(
+    (searchValue: string) => {
+      if (debounceTimer.current) clearTimeout(debounceTimer.current);
+      abortController.current?.abort();
+
+      if (!identitySource || !http || !searchValue || searchValue.length > 1024) {
```
Hailong-am: nit: why 1024?

src/plugins/workspace/public/components/add_collaborators_modal/workspace_collaborator_input.tsx
```diff
@@ -59,6 +82,42 @@ export const WorkspaceCollaboratorInput = ({
...
+        setIsLoading(true);
+        try {
+          const results: Array<{
+            id: string;
+            name: string;
+          }> = await http.get('/api/security/identity/_entries', {
+            query: {
+              source: identitySource.source,
+              type: identitySource.type,
+              keyword: searchValue,
```
Hailong-am: should we trim the searchValue?

src/plugins/workspace/public/components/add_collaborators_modal/workspace_collaborator_input.tsx
```diff
@@ -75,15 +134,37 @@ export const WorkspaceCollaboratorInput = ({
...
-          data-test-subj={`workspaceCollaboratorIdInput-${index}`}
-          placeholder={collaboratorIdInputPlaceholder}
-          aria-labelledby={COLLABORATOR_ID_INPUT_LABEL_ID}
-          isInvalid={!!error}
-        />
+        {identitySource ? (
+          <EuiComboBox
+            compressed
+            singleSelection={{ asPlainText: true }}
+            placeholder={collaboratorIdInputPlaceholder}
```
Hailong-am, line 141: nit: For the case we can do search by input, should we update placeholder to mention the search capability? like `type to search by name ...`

src/plugins/workspace/public/components/workspace_form/workspace_collaborator_table.tsx
Hailong-am: +1

src/plugins/workspace/public/components/workspace_form/workspace_collaborator_table.tsx
gaobinlong: Added it, please help to review again, thanks!

src/plugins/workspace/public/components/workspace_form/workspace_collaborator_table.tsx
gaobinlong: Optimize here by only call APIs when new collaborators are added.

src/plugins/workspace/public/components/add_collaborators_modal/workspace_collaborator_input.tsx
gaobinlong: Change that to this: `if (!abortController.current?.signal.aborted) setIsLoading(false);`

src/plugins/workspace/public/components/add_collaborators_modal/workspace_collaborator_input.tsx
gaobinlong: This is not necessary, the limitation should be set in the API side.

src/plugins/workspace/public/components/add_collaborators_modal/workspace_collaborator_input.tsx
gaobinlong: The inputPlaceHolder can be override when registering a new collaborator type.

changelogs/fragments/11780.yml
```diff
@@ -0,0 +1,2 @@
+feat:
+- [Workspace] collaborators support name lookup ([#11780](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11780))
```
Hailong-am: oh, we don't need changelogs anymore

gaobinlong: Removed it, thanks!

gaobinlong:
@Hailong-am @wanglam please help to take a look into this PR again, thanks!

Hailong-am (approved):

wanglam (approved):

PR #11788, opened 2026-04-20
[Bug] Dashboard Index Visualization

### LINKED ISSUE #11779: [BUG] dashboard panel showing no data for explore visualization
**Describe the bug**

When a data table or visualization is created using a dataset of type INDEX, the dashboard panel appears empty after these objects are added to the dashboard

截屏2026-04-20 08 10 58

### Root cause
For datasets of type INDEX, it appears to bypass retrieving the full object when creating the search source.

```
export const createSearchSource = (
  indexPatterns: IndexPatternsContract,
  searchSourceDependencies: SearchSourceDependencies
) => async (searchSourceFields: SearchSourceFields = {}) => {
  const fields = { ...searchSourceFields };

  // hydrating index pattern
  if (
    fields.index &&
    typeof fields.index === 'string' &&
    (!fields.query?.dataset?.type ||
      fields.query.dataset.type === DEFAULT_DATA.SET_TYPES.INDEX_PATTERN)
  ) {
    fields.index = await indexPatterns.get(fields.index as string);
  }

  const searchSource = new SearchSource(fields, searchSourceDependencies);

  // todo: move to migration script .. create issue
  const query = searchSource.getOwnField('query');

  if (typeof query !== 'undefined') {
    searchSource.setField('query', migrateLegacyQuery(query));
  }

  return searchSource;
};
```

logging error:
since INDEX type is passed,  searchSource.index will be in type String

截屏2026-04-20 08 25 08

For example:
the successful query state:
```
{
    "query": "",
    "language": "PPL",
    "dataset": {
        "id": "ef2c7630-37db-11f1-a624-7501d27e8725",
        "title": "otel-v1-apm-span*",
        "type": "INDEX_PATTERN",
        "timeFieldName": "endTime",
        "dataSource": {
            "id": "7d03f9f5-b9c4-38d3-b888-751575bb49df",
            "title": "olly3-public",
            "type": "OpenSearch"
        }
    }
}
```
the failed query state:
```
{
    "query": "",
    "language": "PPL",
    "dataset": {
        "id": "7d03f9f5-b9c4-38d3-b888-751575bb49df::3f7a3f22-7637-4e89-897c-5206c5ca52a0",
        "title": "train-ticket-logs-104",
        "type": "INDEXES",
        "timeFieldName": "timestamp",
        "dataSource": {
            "id": "7d03f9f5-b9c4-38d3-b888-751575bb49df",
            "title": "olly3-public",
            "type": "OpenSearch",
            "version": "2.19.0"
        },
        "signalType": "logs"
    }
}
```

### Relevant pr
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8689

**To Reproduce**
1. Create a dataset in type INDEX and INDEX_PATTERN
2. Create visualizations or data tables using created datasets and add them to dashboard

TackAdam:
### Description

Fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11779 — dashboard panels render empty for explore visualizations built from a non-`INDEX_PATTERN` dataset (e.g. `INDEXES`).

  ## Root cause

  [PR #8689](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/8689) changed `createSearchSource` to skip index-pattern hydration entirely for any dataset whose type isn't `INDEX_PATTERN`. For those datasets, `searchSource.index` is left as a raw string id.

Discover's `use_search.ts` compensates for this by priming the dataset cache (`datasetService.cacheDataset`) and swapping the hydrated pattern back onto the search source. The dashboard path never hits that code — the explore embeddable factory loads the saved explore, passes the search source straight to the embeddable, and downstream consumers that assume `index` is a full `IndexPattern` (`dataset.flattenHit(...)`, `fields.index.fields.forEach(...)` in data-frame result conversion, etc.) crash.

  ## Changes

Hydration is now centralized in the search-source layer so Discover, the Explore embeddable, and any other consumer of searchSource.create() all benefit from the same fix — rather than each caller re-implementing the cacheDataset + setField('index', ...) dance.

  src/plugins/data/common/search/search_source/search_source.ts — added an optional hydrateDataset?: (dataset: Dataset) => Promise<void> hook to SearchSourceDependencies. Left undefined on the server and in tests that don't care about hydration.

  src/plugins/data/common/search/search_source/create_search_source.ts — restored index hydration for all dataset types that was previously bypassed by PR #8689. For INDEX_PATTERN (and unset) it uses full get(id); for other dataset types it does a cache-only get(id, true), and on a cache miss invokes hydrateDataset(dataset) and retries the cache-only lookup. Lookup errors are swallowed, leaving index as a string if hydration fails.

  src/plugins/data/public/search/search_service.ts — wires hydrateDataset on plugin start by calling
  datasetService.cacheDataset(dataset, { uiSettings, savedObjects, notifications, http }). This is the single place that connects the common-layer hook to the public-layer dataset service.

  ## Tests

  Added regression tests at the layer the fix now lives at:

  - create_search_source.test.ts — verifies indexPatterns.get is called with (id) for INDEX_PATTERN / no dataset type and with (id,true) for non-INDEX_PATTERN; that the index stays as a string when the initial lookup throws; that hydrateDataset is invoked on a cache miss and the second get(id, true) returns the freshly-cached pattern; and that hydrateDataset is not invoked for INDEX_PATTERN datasets or when the cache-only lookup already returns a pattern.

  ## Testing the changes

  - Reproduced empty panel on a dashboard containing a visualization built from an INDEXES dataset; panel renders rows after the
  fix.
  - Verified INDEX_PATTERN-based panels still render identically.
  - Verified Discover still works for INDEXES datasets (now routes through the same centralized hydration).
  - yarn test:jest src/plugins/data/common/search/search_source/create_search_source.test.ts
  - yarn test:jest src/plugins/explore/public/embeddable/explore_embeddable_factory.test.tsx

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11779 

## Screenshot
Before:
Before

After:
After

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/embeddable/explore_embeddable_factory.tsx
```diff
@@ -31,6 +33,39 @@ interface StartServices {
   isEditable: () => boolean;
 }
 
+/**
+ * For non-INDEX_PATTERN datasets (e.g. INDEXES), the saved search source stores the
+ * index as a string id and there is no backing saved-object to resolve it. Prime the
+ * dataset cache and swap the `index` field to a hydrated IndexPattern so downstream
+ * code that assumes a full IndexPattern (flattenHit, fields.forEach, etc.) works.
+ */
+async function hydrateSearchSourceIndex(
```
LDrago27: @TackAdam Why don;t we move the hydration to the search source or at that level itself. It should resolve the issue both for explore and discover. 

Currently this resolves the issue only for Explore.

LDrago27:
@TackAdam Btw on a tangent, We are currently saving all the newly created Datasets as INDEXES, which I don;t think adds any value. Historically INDEXES were meant for temporary data views which were not saved to .kibana index. With the newer Dataset flows the Datasets are anyways written to .kibana index. In that scenario why not save them as INDEX PATTERNS type itself. 

It might simplify the flow.

LDrago27 (approved):

Maosaic (approved):

PR #11790, opened 2026-04-21
[Explore Vis] fix: add to dashboard not working when save a vis the second time

Qxisylolo:
### Description

This pr fixes:
1. add to dashboard not working when save a vis the second time
2. Optimize tooltip size and overflow in Y direction

## Screenshot
1. duplication handler for explore title and dashboard title

https://github.com/user-attachments/assets/25ca7d44-188d-46a7-940f-f5ea8ff63c41

2. Optimize tooltip size and overflow in Y direction

https://github.com/user-attachments/assets/079c8d28-5bd0-4539-9fd8-77088d329dd8

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/visualizations/add_to_dashboard_modal.tsx
```diff
@@ -77,52 +112,54 @@ export const AddToDashboardModal: React.FC<AddToDashboardModalProps> = ({
...
-    setIsTitleOrDashboardTitleDuplicate(true);
+  const renderExploreDuplicateCallout = () => {
+    if (!isExploreDuplicate) {
+      return null;
+    }
+
+    return (
+      <EuiCallOut
+        title={i18n.translate('explore.addtoDashboardModal.exploreDuplicateWarning', {
+          defaultMessage: 'A search with this name already exists.',
```
Hailong-am: nit: should we say a visualization here?

Hailong-am (dismissed):

src/plugins/explore/public/components/visualizations/add_to_dashboard_modal.tsx
Qxisylolo: Yes, visualization makes more sense to me. I called it “search” mainly to align with the initial design. will change the name

yubonluo (dismissed):

wanglam:
## Review: [Explore Vis] fix: add to dashboard not working when save a vis the second time

Thanks for the fix, @Qxisylolo! The rework from callback-based duplicate detection to proactive inline checking is a cleaner pattern. Here are my findings:

### Concerns

**1. Missing debounce on duplicate checks**
`checkExploreDuplicate` and `checkDashboardDuplicate` are called on every keystroke via the `onChange` handler. These make `savedObjectsClient.find()` API calls each time. The component already uses `DebouncedFieldText` but the duplicate check fires on every `onChange` callback from it. This should be debounced to avoid unnecessary requests for fast typists.

**2. No error handling in duplicate check functions**
Both `checkExploreDuplicate` and `checkDashboardDuplicate` do `await savedObjectsClient.find(...)` with no try/catch. If the API call fails (network issue, permissions), it will throw an unhandled promise rejection. Should silently catch and default to `false`.

### Should Fix

**3. Typo in test IDs: `titleDupicateWarnMsg` and `dashboardTitleDupicateWarnMsg`**
"Dupicate" should be "Duplicate". The first is carried over from old code, but the new `dashboardTitleDupicateWarnMsg` perpetuates the typo. Consider fixing both while you're here.

**4. `onTitleDuplicate` made optional but never cleaned up**
In `add_to_dashboard.ts` and `save_explore.ts`, `onTitleDuplicate` is changed from required to optional. Since `isTitleDuplicateConfirmed` is now always `true`, `onTitleDuplicate` will never be called. Consider removing it entirely rather than making it optional — dead code paths are confusing.

**5. Race condition on duplicate checks**
If the user types quickly, multiple `find()` calls could resolve out of order, causing the duplicate state to reflect a stale query rather than the current input. Debouncing would help, but ideally you'd also cancel/ignore stale responses (e.g., check that the current title still matches when the response arrives).

Qxisylolo:
> ## Review: [Explore Vis] fix: add to dashboard not working when save a vis the second time
> Thanks for the fix, @Qxisylolo! The rework from callback-based duplicate detection to proactive inline checking is a cleaner pattern. Here are my findings:
> 
> ### Concerns
> **1. Missing debounce on duplicate checks** `checkExploreDuplicate` and `checkDashboardDuplicate` are called on every keystroke via the `onChange` handler. These make `savedObjectsClient.find()` API calls each time. The component already uses `DebouncedFieldText` but the duplicate check fires on every `onChange` callback from it. This should be debounced to avoid unnecessary requests for fast typists.
> 
> **2. No error handling in duplicate check functions** Both `checkExploreDuplicate` and `checkDashboardDuplicate` do `await savedObjectsClient.find(...)` with no try/catch. If the API call fails (network issue, permissions), it will throw an unhandled promise rejection. Should silently catch and default to `false`.
> 
> ### Should Fix
> **3. Typo in test IDs: `titleDupicateWarnMsg` and `dashboardTitleDupicateWarnMsg`** "Dupicate" should be "Duplicate". The first is carried over from old code, but the new `dashboardTitleDupicateWarnMsg` perpetuates the typo. Consider fixing both while you're here.
> 
> **4. `onTitleDuplicate` made optional but never cleaned up** In `add_to_dashboard.ts` and `save_explore.ts`, `onTitleDuplicate` is changed from required to optional. Since `isTitleDuplicateConfirmed` is now always `true`, `onTitleDuplicate` will never be called. Consider removing it entirely rather than making it optional — dead code paths are confusing.
> 
> **5. Race condition on duplicate checks** If the user types quickly, multiple `find()` calls could resolve out of order, causing the duplicate state to reflect a stale query rather than the current input. Debouncing would help, but ideally you'd also cancel/ignore stale responses (e.g., check that the current title still matches when the response arrives).

 Thanks for the review
Regarding: 
**1. Missing debounce on duplicate checks**, DebouncedFieldText already debounces onChange 
**4. `onTitleDuplicate` made optional but never cleaned up**, top_nav_save will use it, so can't delete.


For other comments, I have fixed, thanks again!

ruanyl (approved):

src/plugins/explore/public/components/visualizations/add_to_dashboard_modal.tsx
```diff
@@ -77,52 +132,54 @@ export const AddToDashboardModal: React.FC<AddToDashboardModalProps> = ({
...
-        newTitle: title,
-        isTitleDuplicateConfirmed: isTitleOrDashboardTitleDuplicate,
-        onTitleDuplicate: handleTitleDuplicate,
-        mode: selectedOption,
-        selectDashboard: selectedDashboard,
-        newDashboardName,
-      });
+    try {
+      if (savedExplore) {
+        await onConfirm({
```
yubonluo, line 137: do we need to consider error scenarios？

Qxisylolo: Thanks for the comment, we already handle error at onConfirm

yubonluo (approved):

PR #11791, opened 2026-04-21
add cross origin check for prod

ruchidh:
### Description

The session expired interceptor's isValidRedirectURL function was rejecting legitimate cross-origin AWS auth redirect URLs.
 When a 401 response includes an X-Auth-Redirect-URL header pointing to the AWS console (e.g., https://us-east-2.console.aws.amazon.com/...), the validation failed because window.location.origin is on a different AWS domain (e.g., https://...opensearch.amazonaws.com).
 
 isValidRedirectURL only allowed relative URLs and same-origin absolute URLs. In production, the auth flow redirects to a different AWS domain for re-authentication, which was being blocked as an open redirect.
 
 ###  Fix
Added trusted AWS domain patterns to the URL validation:

*.amazonaws.com
*.aws.amazon.com
*.console.aws.amazon.com
These are only allowed over HTTPS. Non-AWS external domains are still rejected to prevent open redirect attacks.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/core/public/http/session_expired_interceptor.ts
```diff
@@ -17,18 +17,34 @@ const SESSION_REDIRECT_DELAY_MS = 5000;
...
   try {
     const parsed = new URL(url);
-    return parsed.origin === window.location.origin;
+
+    if (parsed.origin === window.location.origin) {
+      return true;
+    }
+
+    // Trust AWS auth domains that the 401 response may redirect to
+    const trustedPatterns = [
```
LDrago27, line 38: nit: maybe we should move out all of this into a constants file. Easier to update and maintain in future.

LDrago27 (approved):

src/core/public/http/session_expired_interceptor.ts
ruchidh: noted

TackAdam (approved):

src/core/public/http/session_expired_interceptor.ts
Hailong-am: could place it in opensearch_dashboards.yml file so we can customize it easily

PR #11792, opened 2026-04-21
fix: upgrade uuid from 3.3.2 to 8.3.2

### LINKED ISSUE #7141: [BUG] The uuid dependency is on a deprecated version of uuid (3.3.2)
**Describe the bug**

OpenSearch-Dashboards uses uuidv4 for many identifiers used in the application. To work with UUIDs, OSD has a dependency on the [uuid node package](https://www.npmjs.com/package/uuid) which is currently on [version 3.3.2](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/package.json#L244).

This version is currently deprecated and should be upgraded to a [supported version](https://www.npmjs.com/package/uuid?activeTab=versions) of the dependency.

gittihub-jpg:
## What
Upgrades uuid dependency from 3.3.2 to 8.3.2.

## Why
Fixes #7141

uuid 3.3.2 has known security vulnerabilities (CVE). Version 8.3.2 is the current stable release with security fixes.

## API Compatibility
uuid v8 maintains backward compatibility for the commonly used functions (v4, v5, etc.). The main breaking changes in uuid v4-v8 affect:
- Named exports instead of default export (already handled by existing import style)
- No functional change for standard usage in this codebase

## Testing
- Dependency version bump only
- No functional code changes to application logic

### DIFF
--- a/package.json
+++ b/package.json
@@ -331,7 +331,7 @@
     "tslib": "^2.0.0",
     "type-detect": "^4.0.8",
     "use-sync-external-store": "^1.5.0",
-    "uuid": "3.3.2",
+    "uuid": "8.3.2",
     "whatwg-fetch": "^3.0.0",
     "yauzl": "^2.10.0"
   },


angle943:
please check in the yarn.lock changes

PR #11797, opened 2026-04-22
[Workspace] Fix checkbox still be checked after deletion or update in collaborator list page

gaobinlong:
### Description

In workspace collaborator list page, select one or more items to delete or update access level, after the operation, the checkbox are still be checked.

### Issues Resolved

No issue.

## Screenshot

image

image

https://github.com/user-attachments/assets/ede0ce40-e473-48e3-92c6-f6482815ad17

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/public/components/workspace_form/workspace_collaborator_table.tsx
+++ b/src/plugins/workspace/public/components/workspace_form/workspace_collaborator_table.tsx
@@ -158,6 +158,7 @@ export const WorkspaceCollaboratorTable = ({
   handleSubmitPermissionSettings,
 }: Props) => {
   const [selection, setSelection] = useState<PermissionSetting[]>([]);
+  const tableRef = useRef<EuiInMemoryTable<PermissionSettingWithAccessLevelAndDisplayedType>>(null);
   const {
     overlays,
     services: { notifications },
@@ -344,6 +345,7 @@ export const WorkspaceCollaboratorTable = ({
               }),
             });
             setSelection([]);
+            tableRef.current?.setSelection([]);
             modal.close();
           }
         },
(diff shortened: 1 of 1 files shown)


src/plugins/workspace/public/components/workspace_form/workspace_collaborator_table.tsx
```diff
@@ -344,6 +345,7 @@ export const WorkspaceCollaboratorTable = ({
               }),
             });
             setSelection([]);
+            tableRef.current?.setSelection([]);
```
Hailong-am, line 348: we have `setSelection([]);` already, why it doesn't work?

gaobinlong: I see with few total collaborators, it works, but with many collaborators like more than 10, it doesn't work, the reason maybe that with few items EUI cleans up naturally, but with many items, re-render doesn't fully flush EUI's internal state in the same way, so the stale checkboxes remain.

Hailong-am (approved):
LGTM. The fix is correct.

`setSelection([])` only clears the parent React state — it has no effect on `EuiInMemoryTable` internal checkbox state because selection is uncontrolled. The imperative `ref.setSelection([])` API is the only reliable way to programmatically clear checkboxes, especially:

1. **With pagination (10+ items)** — the table does not reliably reconcile its internal selection against updated `items` for non-visible pages
2. **For "change access level"** — items remain in the list (just modified), so the table has no reason to auto-clear selection

The `onSelectionReset` callback pattern is clean for propagating the reset from the `Actions` child component back up.

PR #11812, opened 2026-04-24
fix: restore UTF-8 BOM in CSV exports for Excel compatibility

### LINKED ISSUE #11807: [BUG] CSV downloads are missing UTF-8 BOM character
**Describe the bug**

Table visualizations contain a button to download the data as a CSV file. For other visualizations this can be done via the Inspect page. 
These downloads no longer contain the UTF-8 BOM character. As a result, Excel does not recognize the character encoding. All non-ascii characters are therefore not correctly displayed in Excel. 
This used to work still in OpenSearch Dashboard 2.19.x

**To Reproduce**
1. Go to a dashboard with a table visualization showing text with non-ascii characters.
2. Click on the Download CSV button. Select Raw or Formatted, it does not matter.
3. Open the generated CSV file in Excel or in Notepad++.
4. Notice that the non-ascii characters are not correctly rendered in Excel and that the file encoding in Notepad++ is simply UTF-8 not UTF-8-BOM.

**Expected behavior**
The CSV export should always include the BOM character, such that the file is automatically recognized as UTF-8 encoded by applications like Excel.

**OpenSearch Version**
Latest, currently 3.6.0

**Dashboards Version**
Latest, currently 3.6.0

**Plugins**
Only default plugins.

**Screenshots**

Image

Image

 - OS: Windows 11
 - Browser: Chrome 147

**Additional context**

This bug is likely caused by moving to node.js 20+ when a native implementation of the Blob class was introduced. Previously it would rely on the browser or polyfills. The Blob polyfill or browser implementation would sometimes automatically prepend a UTF-8 BOM (EF BB BF) when charset=utf-8 was specified in the MIME type. The node.js 20+ native Blob implementation follows the WHATWG spec more strictly — it encodes strings as UTF-8 without a BOM, regardless of the charset parameter. The fix would be to explicitly prepend the BOM character when exporting UTF-8 files.

Hailong-am:
### Description

PR #9484 replaced `@elastic/filesaver` with `file-saver`, which inadvertently removed the UTF-8 BOM from CSV exports. The old `@elastic/filesaver` had an `auto_bom()` function that automatically prepended the BOM (`U+FEFF` / `EF BB BF`) for `text/*;charset=utf-8` blobs. The replacement `file-saver` library defaults `autoBom` to `false` and only applies it in the legacy IE code path, so modern browsers never received the BOM.

Without the BOM, Excel does not recognize the file as UTF-8 and renders non-ASCII characters incorrectly.

**Changes:**
- Add a shared `autoBom()` utility in `@osd/std` that restores the `@elastic/filesaver` behavior
- Apply `autoBom()` to all CSV export `saveAs()` call sites across inspector, vis_type_table, discover, explore, and agent_traces plugins
- Add unit tests for the new utility

### Issues Resolved

fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11807

## Screenshot

image

## Testing the changes

1. Create a table visualization with non-ASCII text (e.g. Chinese, Japanese, accented characters)
2. Download CSV (both Raw and Formatted) and open in Excel — characters should render correctly
3. Verify the downloaded CSV file starts with bytes `EF BB BF` (UTF-8 BOM) using a hex editor or Notepad++
4. Download CSV from Discover page and verify the same
5. Download CSV from Inspector Data view and verify the same
6. Run `yarn test:jest packages/osd-std/src/auto_bom.test.ts` — all 6 tests pass

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/packages/osd-std/src/auto_bom.ts
+++ b/packages/osd-std/src/auto_bom.ts
@@ -1,0 +1,28 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+const UTF8_TEXT_MIME_RE = /^\s*(?:text\/\S*|application\/xml|\S*\/\S*\+xml)\s*;.*charset\s*=\s*utf-8/i;
+
+/**
+ * Prepend a UTF-8 BOM (U+FEFF) to a Blob whose MIME type indicates
+ * `charset=utf-8` text content (e.g. `text/csv;charset=utf-8`).
+ *
+ * This restores the behavior of the old `@elastic/filesaver` package,
+ * which ran this check automatically.  The replacement `file-saver`
+ * library defaults `autoBom` to `false` and only applies it in the
+ * legacy IE code-path, so callers must handle it themselves.
+ */
+export function autoBom(blob: Blob): Blob {
+  if (UTF8_TEXT_MIME_RE.test(blob.type)) {
+    return new Blob(['\uFEFF', blob], { type: blob.type });
+  }
+  return blob;
+}
--- a/packages/osd-std/src/auto_bom.test.ts
+++ b/packages/osd-std/src/auto_bom.test.ts
@@ -1,0 +1,66 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { autoBom } from './auto_bom';
+
+function blobToBytes(blob: Blob): Promise<Uint8Array> {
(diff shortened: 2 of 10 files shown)


yubonluo (approved):

packages/osd-std/src/auto_bom.ts
```diff
@@ -0,0 +1,28 @@
...
+ * `charset=utf-8` text content (e.g. `text/csv;charset=utf-8`).
+ *
+ * This restores the behavior of the old `@elastic/filesaver` package,
+ * which ran this check automatically.  The replacement `file-saver`
+ * library defaults `autoBom` to `false` and only applies it in the
+ * legacy IE code-path, so callers must handle it themselves.
+ */
+export function autoBom(blob: Blob): Blob {
+  if (UTF8_TEXT_MIME_RE.test(blob.type)) {
+    return new Blob(['\uFEFF', blob], { type: blob.type });
```
wanglam, line 25: Nit: is it safe to append BOM twice if BOM already exists in blob?

wanglam (approved):

packages/osd-std/src/auto_bom.ts
Hailong-am: A double BOM is harmless — the second U+FEFF is interpreted as a zero-width no-break space (ZWNBSP), which is an invisible
  character. Excel, Notepad++, and CSV parsers handle it without issue.

PR #11823, opened 2026-04-24
[Explore] Fill the update button with primary color if query changed

joshuali925:
### Description

In Explore logs, traces, metrics, the update (run query) button is far on the top right, and it's hard to notice that user needs to click Update, especially in query builder mode.

@kamingleung suggested a small change to make it fill with primary color. After UI is updated and reflects query result, it goes back to original "Refresh" without the fill.

This PR additionally avoids showing update on Metrics page Query tab switch

## Screenshot

image
image

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/agent_traces/public/components/top_nav/query_execution_button/query_execution_button.test.tsx
+++ b/src/plugins/agent_traces/public/components/top_nav/query_execution_button/query_execution_button.test.tsx
@@ -248,6 +248,17 @@ describe('QueryExecutionButton', () => {
     expect(screen.getByText('Update')).toBeInTheDocument();
     // Verify the button has the primary color (blue) for consistent theming
     expect(button).toHaveClass('euiButton--primary');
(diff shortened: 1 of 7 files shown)


angle943 (approved):

TackAdam (approved):

kamingleung:
@joshuali925 In the changed state, consider displaying the play icon with the text "Run". Maintain the same button width across the different states, so that UI elements are not jumping positions.

joshuali925:
@kamingleung thanks for the clarification. It appears to be an OUI change that will affect all update buttons (unless we remove usages of `SuperUpdateButton` in explore). Do we want this change in all places including old discover?

https://github.com/opensearch-project/oui/blob/24fc9f0b560bad366ff11776e49fac95240df595/src/components/date_picker/super_date_picker/super_update_button.tsx#L143

Also I'll merge this first since OUI release would take more time

PR #11831, opened 2026-04-25
feat(osd-pm): Skip bootstrap when possible

angle943:
### Description

Adds a fingerprint file at the repo root that captures the set of inputs bootstrap depends on: root yarn.lock, root package.json, every project's package.json, every external plugin's yarn.lock, the osd-pm bundle, and the Node version. On subsequent runs, if the fingerprint matches, node_modules is intact, and every per-project .bootstrap-cache is still valid, the whole pipeline is skipped.

Typical warm-unchanged runs go from ~30-50s to ~3s (~14x faster). Cold and warm-with-edits runs are unaffected within measurement noise.

The fast path is disabled when the user scopes the run (--include, --exclude, --skip-opensearch-dashboards-plugins, --oss), passes --no-cache, or passes --frozen-lockfile. yarn osd clean deletes the fingerprint file so stale matches can't happen after a clean.

Verbose logs show why the fast path did or did not trigger, making unexpected misses easy to diagnose.

https://github.com/user-attachments/assets/d2bc8fbf-b8cb-4bdf-ade6-7a0a2a7469c1

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


packages/osd-pm/src/utils/bootstrap_fingerprint.ts
```diff
@@ -0,0 +1,135 @@
...
+const FINGERPRINT_FILENAME = '.osd-bootstrap-fingerprint';
+
+export interface Fingerprint {
+  version: number;
+  rootLockfile: string;
+  rootPackageJson: string;
+  projectManifests: string;
+  externalPluginLockfiles: string;
+  osdPmDist: string;
+  nodeVersion: string;
```
joshuali925, line 23: should the yarn version also be part of the fingerprint?

packages/osd-pm/src/utils/bootstrap_fingerprint.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  const sortedProjects = Array.from(projects.values()).sort((a, b) => a.path.localeCompare(b.path));
+
+  const manifestParts: string[] = [];
+  const lockParts: string[] = [];
+  for (const project of sortedProjects) {
+    const rel = normalizeRelativePath(osd.getRelative(project.path));
+    manifestParts.push(`${rel}:${sha1(readOrEmpty(project.packageJsonLocation))}`);
+    // External plugins (./plugins/*, ../opensearch-dashboards-extra/*) run their
+    // own yarn install against their own yarn.lock. Workspace projects have
```
joshuali925: does this mean lock changes in external plugins won't trigger fingerprint change? For example, if a workflow clones OSD, bootstrap, clones plugins, bootstrap, all running under OSD root, does it correctly bootstrap plugins?

joshuali925:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

packages/osd-pm/src/commands/bootstrap.ts
```diff
@@ -38,12 +39,76 @@ import { getAllChecksums } from '../utils/project_checksums';
...
+              break;
+            }
+          }
+        }
+
+        if (!staleProject) {
+          log.success(
+            'bootstrap already up to date (fingerprint matched) — run with `--no-cache` to force a full bootstrap'
+          );
+          return;
```
joshuali925, line 121: should this still run https://github.com/opensearch-project/OpenSearch-Dashboards/blob/059518842a31bf3f6b3bc2172262d36c6f541810/packages/osd-pm/src/commands/bootstrap.ts#L137-L139

packages/osd-pm/src/utils/bootstrap_fingerprint.ts
angle943: Good catch — technically this is already covered indirectly: the root `package.json`'s `engines.yarn` field pins the yarn version, so a yarn upgrade that actually bumps that pin will flip `rootPackageJson` and invalidate the fingerprint. But:

(a) yarn 1 doesn't strictly enforce `engines`, so a local machine running a newer yarn than the pin wouldn't invalidate
(b) explicitly capturing it is essentially free — I'm reading `process.env.npm_config_user_agent` (which yarn sets when it runs the script, no subprocess) and parsing the `yarn/<version>` prefix. Added it for consistency with `nodeVersion`. Falls back to empty string when bootstrap is invoked outside a yarn script (e.g. directly via `scripts/osd`), so non-yarn invocations are unaffected.

packages/osd-pm/src/utils/bootstrap_fingerprint.ts
angle943: Traced through both carefully:

**`linkProjectExecutables`**: idempotent — re-linking produces the same symlinks, and self-heals if someone wipes `node_modules/.bin` out-of-band. Runs unconditionally on the fast path.
**`validateDependencies`**: pure validator in `strict`/`ignore`, but `loose`/`force`/`brute-force` can mutate `package.json`/`yarn.lock` while reconciling single-version conflicts. Running it unconditionally and logging "bootstrap already up to date" while silently writing files would be confusing, so on the fast path I only invoke it in the non-mutating modes and defer the reconciling modes to a full bootstrap. Added an `isMutatingSingleVersionMode` helper in `validate_dependencies.ts` so the "which modes write to disk" knowledge stays colocated with the enum itself rather than hardcoded in the bootstrap fast path.

packages/osd-pm/src/commands/bootstrap.ts
angle943: Traced through both carefully. Both are safe to skip on a fingerprint hit, but I think both are also cheap enough to just always run and the observability/self-heal is worth it:

**`validateDependencies`**: pure validator in `strict`/`ignore` modes, but can *mutate* lockfiles/package.jsons in `loose`/`force`/`brute-force` (those modes reconcile version conflicts by calling `installDependencyVersion`). Mutations on the fast path are fine: the next bootstrap invocation sees the changed manifests, misses the fingerprint, and runs the full pipeline to materialize them — same as today's non-fast-path behavior, which also relies on a second bootstrap to install mutated state.

**`linkProjectExecutables`**: idempotent — same inputs produce the same symlinks. Useful if someone wipes `node_modules/.bin` out-of-band between runs.

Cost: a few hundred ms each on a ~3s fast path. Moved them to run before the early return so the fingerprint hit still validates and self-heals.

packages/osd-pm/src/utils/bootstrap_fingerprint.ts
joshuali925: I'm thinking about the global `npm i -g yarn` version, it would be adding `yarn --version` command output string as part of the fingerprint. It does not necessarily mean package.json would have a change.

Because our CI runs bootstrap using specific yarn version, and compares updated lock file agains the one in PR

> I'm reading process.env.npm_config_user_agent (which yarn sets when it runs the script, no subprocess) and parsing the yarn/<version> prefix.

it's a fork subprocess call which shouldn't be a meaningful overhead, but no subprocess is better. I just saw it's already added in PR update, resolving this

packages/osd-pm/src/utils/bootstrap_fingerprint.ts
joshuali925: this reply is for a different comment?

joshuali925 (approved):
nice. curious when do you run into this, where bootstrap runs with no dependency change?

ruchidh (approved):

angle943:
"nice. curious when do you run into this, where bootstrap runs with no dependency change?" <- maybe its just me, but sometimes i run "yarn osd bootstrap" after lunch break, or starting a new day. but nothing changed from previous session. or you git pull and run it again

PR #11832, opened 2026-04-26
fix(visualization): preserve saved pie chart colors and legend visibility on dashboards

### LINKED ISSUE #9597: [BUG] Colors in Pie Chart resets on removing a filter
**Describe the bug**
When a dashboard filter is applied that results in no data (zero documents) And when that filter is removed, the pie chart doesn't retain its original predefined colors instead, it displays new/different colors for the segments.

**To Reproduce**
1. Create a Pie Chart Visualization
2. Click on each of the label and select a color of label that is different from default color
3. Save the visualization
4. Add the visualization to a Dashboard
5. Add a global dashboard filter that produces no results
6. Remove the dashboard filter , the Pie chart switches to default color. 

**Expected behavior**

**OpenSearch Version**
2.17, 3.0.0

**Dashboards Version**
2.17, 3.0.0

Video:

https://github.com/user-attachments/assets/3c9929d2-0e03-4ce5-839f-8ea703c5649e

seanthegeek:
### Description

Two related defects in how `VisualizeEmbeddable` and the visualization expression renderer manage `PersistedState`. Both manifest as the symptom in #9597 (pie chart resets to default colors and legend visibility), but they fire under different conditions, so addressing only one leaves the other in place.

#### Bug 1 — `transferCustomizationsToUiState` clobbers the entire `vis` subtree

`transferCustomizationsToUiState` previously applied dashboard panel customizations with:

```ts
this.vis.uiState.clearAllKeys();
this.vis.uiState.set('vis', visCustomizations.vis);
```

`PersistedState.set('vis', value)` replaces the **entire `vis` subtree** in `_mergedState` (the `mergeMethod` returns `sourceValue` whenever the merge key matches the requested keyPath at the root). Any saved key not present in `input.vis` is dropped from `_mergedState`. The Flight sample dashboard's pie panels store:

```json
"embeddableConfig":{"vis":{"legendOpen":false}}
```

— no `colors`. So a viz saved with custom slice colors via the Visualize editor renders on that dashboard with **default** colors, even on initial load with no filter applied.

`clearAllKeys` has the same shape: `set(key, null)` nulls `_mergedState[key]` via the merge, so subsequent sets cannot fall back to `_defaultState` either.

**Fix:** iterate `input.vis` and `set('vis.<subKey>', value)` per property; replace `clearAllKeys` with iterated `reset(key)`, which restores `_mergedState[key]` from `_defaultState[key]` before re-applying customizations.

#### Bug 2 — renderer creates a fresh `ExprVis` whose `uiState` is never wired up

The visualization expression renderer constructs a new `ExprVis` on every render call. `ExprVis`'s constructor initializes `this.uiState = new PersistedState()` (empty). The only thing that wired it to the embeddable's persisted state was `<Visualization>`'s React constructor — which fires once per mount.

When data goes to zero hits (filter or time range), `<Visualization>` swaps to `<VisualizationNoResults>` and unmounts `<VisualizationChart>`. When the data returns, `<VisualizationChart>` remounts; its `componentDidMount` builds a brand-new `VislibVisController` from the **latest** `props.vis` — whichever fresh `ExprVis` the renderer last produced. That `ExprVis` was never wired, so vislib's color lookup got an empty `vis.colors` and the legend's `vis.legendOpen` fell back to its `true` default. This is the symptom from the original bug report's filter-cycle reproduction.

**Fix:** call `vis.setUiState(uiState)` in the renderer immediately after constructing the `ExprVis`, so the persisted state is wired on every render, not just the first React mount.

### Issues Resolved

Fixes #9597

## Screenshot

No UI markup change — the visible difference is "your saved colors render correctly, on the initial load and across filter/time-range cycles." The video on the linked issue demonstrates the original behavior.

## Testing the changes

**Manual reproduction**

Both pre-existing repros for the issue are covered.

1. Install Sample flight data (Home → "Try sample data").
2. In the Visualize editor, edit `[Flights] Airline Carrier`. Click each legend entry → **Change color** → pick obviously non-default colors. Save.
3. **Initial-load case (Bug 1):** Open `[Flights] Global Flight Dashboard`. Without applying any filter, confirm the pie viz now renders with your saved colors. Before this PR it would render with defaults because the panel's `embeddableConfig.vis.legendOpen` clobbered the saved color subtree.
4. **Filter-cycle case (Bug 2):** Add a filter such as `Carrier : "DefinitelyNotARealAirline"`. The pie shows "No results found." Remove the filter. After this PR your custom colors persist; before, they reverted.
5. **Time-range variant of Bug 2:** Set the time picker to a window with no flights, then expand it back to a window with flights. Same expected behavior.
6. In the Visualize editor, toggle the legend closed and save; verify the legend stays closed across the same filter cycle on the dashboard.

**Automated**

Two new Jest unit tests, all passing locally:

```
yarn test:jest src/plugins/visualizations/public/embeddable/visualize_embeddable.test.ts
yarn test:jest src/plugins/visualizations/public/expressions/visualization_renderer.test.tsx
```

- [`visualize_embeddable.test.ts`](src/plugins/visualizations/public/embeddable/visualize_embeddable.test.ts) exercises `transferCustomizationsToUiState` against a real `PersistedState` and asserts that saved colors survive when the panel only customizes `legendOpen`, that dashboard color overrides take precedence on a per-slice basis, that no panel customizations leaves the saved defaults intact, and that removing a previous customization reverts to the saved default.
- [`visualization_renderer.test.tsx`](src/plugins/visualizations/public/expressions/visualization_renderer.test.tsx) calls the renderer three times against a single DOM node (simulating the unmount/remount cycle) and asserts every captured React tree's `vis.getUiState()` returns the persisted state including `vis.colors` and `vis.legendOpen`.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest` (new tests for both fixes)
  - [ ] `yarn test:jest_integration` (no integration coverage added; existing suites unaffected)
- [ ] New functionality has been documented. (bug fix; no new behavior)

### DIFF
--- a/src/plugins/visualizations/public/embeddable/visualize_embeddable.test.ts
+++ b/src/plugins/visualizations/public/embeddable/visualize_embeddable.test.ts
@@ -1,0 +1,138 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { VisualizeEmbeddable } from './visualize_embeddable';
+import { PersistedState } from '../persisted_state';
+
+// transferCustomizationsToUiState only depends on a small slice of `this`;
+// build a minimal stub and bind the prototype method onto it so we can
+// exercise the behavior against a real PersistedState without standing up
+// the full embeddable (which needs data plugin, timefilter, etc.).
+type CallableTransfer = () => void;
+type CallableReset = () => void;
+
+interface StubState {
+  input: { vis?: { [key: string]: unknown }; table?: unknown };
+  vis: { uiState: PersistedState };
+  parent?: unknown;
+  visCustomizations?: unknown;
+  uiStateChangeHandler: () => void;
+  resetUiStateChanges: CallableReset;
+}
+
+const proto = (VisualizeEmbeddable.prototype as unknown) as {
+  transferCustomizationsToUiState: CallableTransfer;
+  resetUiStateChanges: CallableReset;
+};
+
+const transfer = proto.transferCustomizationsToUiState;
+const resetChanges = proto.resetUiStateChanges;
+
+const buildStub = (savedUiState: Record<string, unknown>, input: StubState['input']): StubState => {
+  const stub: StubState = {
+    input,
+    vis: { uiState: new PersistedState(savedUiState) },
+    parent: { id: 'parent' },
+    uiStateChangeHandler: () => {},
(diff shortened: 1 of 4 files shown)


LDrago27 (dismissed):

angle943 (dismissed):

LDrago27 (dismissed):

sumukhswamy (dismissed):

ruanyl (dismissed):

ruanyl:
FYI: the failed tests seems related to this PR

LDrago27 (approved):

mengweieric (approved):

sumukhswamy (approved):

PR #11833, opened 2026-04-26
feat(build): parallelize release build (~15% faster wall time)

angle943:
### Description

Reduce OpenSearch Dashboards **release-build wall time by ~15%** on a 16-vCPU runner with no new `package.json` dependencies and no change to the produced artifacts' format or contents. `pigz` is used as an **optional runtime dependency** for parallel gzip; without it the refactor alone still delivers ~11%. The gains come from parallelizing CPU-bound stages that previously ran serially and from cutting redundant I/O passes:

**Parallelism**
- `CreateArchives` now compresses platform archives in parallel (bounded queue; concurrency derived from `pigz` availability and CPU count). **This is the dominant contributor** — parallel archives alone account for ~83 s of the 87 s "refactor-only" gain (~95%).
- `TranspileBabel` runs the server and client babel passes concurrently. Done for structural consistency with the other parallelizations; measured wall-clock impact is **under 1 s** (babel is effectively CPU-bound on one thread, and the two globs contend for disk cache).
- `CreateOsPackages` wraps the individual `fpm` tasks in a bounded-concurrency queue (opt-out via `OSD_BUILD_FPM_CONCURRENCY=1`).
- `compressTar` transparently uses `pigz` when it's on `PATH`; the output is standard gzip — any gunzip reads it back. Falls back to Node `zlib` otherwise. Opt-out via `OSD_BUILD_NO_PIGZ=1`.
- `compressZip` is **intentionally untouched** beyond a matching fail-fast on missing source: zip is Windows-only, has no parallel-compressor analog to `pigz`, and historically produces correctly-formed archives; adding the pigz/stall-watchdog machinery would be pure code debt with no win.

**I/O reduction**
- `WriteShaSums`: sha1 + sha256 are now computed in a single read-pass per artifact via a new `getFileHashes(path, algos)` helper (halves disk reads on this step).
- `CreateArchivesSources` / `scanCopy`: prefer hardlinks (same-FS only; opt out via `OSD_BUILD_NO_HARDLINK=1`). The hardlink contract — "destinations must not be mutated in-place" — is audited in code and pinned by a `[CONTRACT]` regression test. All 8 post-`CreateArchivesSources` tasks were reviewed; none mutate hardlinked files.

**Code clarity (not a perf win)**
- `CleanTypescript` + `CleanExtraFilesFromModules` consolidated into one `CleanExtraBuildFiles` task. The combined task walks `node_modules` **once** (merged regex set) instead of twice as in the old two-task sequence; the outer build-tree walk is unchanged. Both old tasks were < 1 s each; this change is for clarity, not speed.

**Dead code**
- Removed `tasks/install_chromium.js`. Wired up to nothing in the build pipeline; the in-file comment indicated it had been superseded since 2020.

**Incidental Windows fix**
- While adding tests for the consolidated `CleanExtraBuildFiles` task, CI surfaced a latent Windows bug in the *pre-existing* `CleanExtraFilesFromModules` task's regex handling: `minimatch.makeRe()` emits forward-slash-only regexes, but `scanDelete` walks paths via `path.join()` which uses the OS-native separator (backslash on Windows). Under the old code, the `**/__tests__`, `**/docs`, `**/CHANGELOG.md`, etc. patterns silently never matched on Windows CI runners — those files shipped in distributables. Fixed here by relaxing the emitted regexes to match either separator. See commit `844a15e4` and the `deletes extras patterns on Windows-style paths` regression test.

### Measured gains (n=3 per condition)

Environment: 16 vCPU Linux (8 cores × 2 threads), Node 22.22.0, yarn 1.22.19, pigz 2.3.4, NVMe.
Command: `yarn build --release --skip-os-packages` (5 platforms).
Clean `rm -rf build target .node_binaries` between runs.

Wall clock (seconds):

| Condition | n | mean | stdev | min | max |
|---|---|---:|---:|---:|---:|
| Baseline (main @ `346c8800`) | 3 | 825.0 | 8.00 | 815.9 | 831.0 |
| This PR, pigz absent | 3 | 737.5 | 1.43 | 735.9 | 738.5 |
| This PR, pigz present | 3 | 697.8 | 5.05 | 693.2 | 703.2 |

Per-task breakdown (mean seconds, n=3):

| Task | Baseline | This PR | +pigz | Δ baseline→pigz |
|---|---:|---:|---:|---:|
| BuildPlugins | 439.0 | 435.7 | 436.7 | −2.3 (−0.5%) |
| TranspileBabel | 16.3 | 16.0 | 16.0 | −0.3 (−2.0%) |
| CreateArchives | 222.0 | 139.3 | 97.7 | **−124.3 (−56.0%)** |

Δ wall clock vs baseline (mean):
- This PR, pigz absent: **−87.5 s (−10.6%)**
- This PR, pigz present: **−127.2 s (−15.4%)**

*Re-benchmarked post-review (n=3 pigz): wall clock mean improved to 676.87 s, ~20 s lower than the table above. Attributing to host-level variance rather than a code change; the 15.4% figure stands as the conservative claim.*

**GitHub Actions CI context (`Build release artifact sanity check` job, `ubuntu-latest`, n=9 baseline):**

| Metric | Baseline (n=9) | This PR (n=2) | Δ |
|---|---:|---:|---:|
| `Build release artifact` step | 1687.6 s (±18.1 s stdev) | **1635.5 s** (runs: 1637 s / 1634 s) | **−52.1 s (−3.1%, −2.9σ)** |
| `CreateArchives` sub-task | 198 s | **135 s** (runs: 137 s / 133 s) | **−63 s (−31.8%)** |
| `BuildOpenSearchDashboards…PlatformPlugins` | 1333 s | 1348 s | +15 s (noise) |

The CI runner is 4 vCPU where local bench was 16 vCPU. `BuildPlugins` dominates CI (~79% of step time, unchanged by this PR), so the archive-parallelism win is diluted to ~3% at the step level even though `CreateArchives` itself drops 32%. The headline −15.4% still applies to release-pipeline runners with more cores, as the parallel-archive savings scale with available CPU up to `platforms.length` (5 here). Inter-run variance on our PR is tight (<4 s stdev across two runs), well within the baseline's noise floor.

**Caveats:**
- Single machine, n=3. Variance is tight (stdev ≤ 8 s), so signal >> noise, but pipeline-specific numbers will differ with runner shape and disk.
- `--skip-os-packages` was used because `fpm` isn't installed on the bench machine, so **parallel-fpm savings are not in these numbers**. The `createOsPackagesTask` factory is behind `OSD_BUILD_FPM_CONCURRENCY` so ops can pin it to 1 on memory-constrained runners.
- On CI runners with ≥ 32 vCPU, the optimizer default (`cpuCount − 1`) already gives `OSD_OPTIMIZER_MAX_WORKERS≈31`; locally on this 16-CPU box, raising it to 30 produced an additional −13 s (not included in the table).

### Behaviour changes reviewers should note

- **Compression level**: release archives stay at gzip level 9 (historical default, unchanged). **Dev archives now default to level 6** — the Node/zlib default — down from the previous unconditional level 9. Dev archives will be ~5–15% larger; dev builds are not distributed so this is an intentional CPU-for-size trade. Both release and dev can be overridden via `OSD_BUILD_GZIP_LEVEL` (0-9). If a CI job validates dev artifact size or pinpoints compression ratio, set `OSD_BUILD_GZIP_LEVEL=9`.
- **Hash emission (`WriteShaSums`)**: sha1 and sha256 sidecars are still emitted per archive, byte-identical to the old implementation (verified by unit test). No new algorithms are introduced. sha1 is retained for consumer compatibility; flagging here for a future deprecation discussion.
- **`install_chromium.js` removed**: the task was neither referenced from `build_distributables.ts` nor from any pipeline config in this repo.
- **`pigz` output is standard gzip**: artifacts stay `.tar.gz` and `.zip`; any gunzip / unzip reads them. No change to sha1/sha256 values' format (content hashes depend on compressor output bytes, which will differ between pigz and zlib at the same level — **the sidecar files will update accordingly**; downstream consumers that pin a specific sha256 must re-record the new value after this change).
- **Security**: `pigz` is invoked via `spawn('pigz', …)` with `PATH` lookup, matching the existing trust model for `fpm`, `tar`, `yarn`, `node`, etc. Operators running hardened build images should pin `pigz` via absolute path (symlink into a trusted bin dir) and review the image layers that provide it. `spawn` uses a `pipe` stdio so the child cannot access the parent's TTY.
- **Log output ordering**: `CreateDebPackage`, `CreateDebArmPackage`, `CreateRpmPackage`, and `CreateRpmArmPackage` now start interleaved rather than strictly sequential. Each task emits explicit start (`▶`) and completion (`✓`) markers with per-task duration, so log slicers that key on those markers continue to work, but any scraper that relied on "deb before rpm" ordering must update. Set `OSD_BUILD_FPM_CONCURRENCY=1` to restore strictly sequential behaviour.
- **Error containment under concurrency**: the bounded worker pools in `CreateArchives` and `createOsPackagesTask` carry a shared `failed` sentinel. When one worker throws, the others stop pulling new items from the queue; in-flight work still completes (we can't safely interrupt a running archiver stream or fpm subprocess). Behaviorally this matches the old sequential "abort on first error" semantics under failure, while still delivering parallel speed on the happy path. Covered by a regression test (`stops dispatching new tasks once any worker throws`).
- **Windows distributables now exclude `**/__tests__`, `**/docs`, `**/CHANGELOG.md`, etc. from `node_modules`**: the old regex emitter produced forward-slash-only patterns that silently never matched backslash paths on Windows CI runners, so those files were being bundled into Windows `.zip` distributables. Linux and macOS builds are unchanged. See the "Incidental Windows fix" bullet above and commit `844a15e4`.

### Environment variables (all optional)

Fully documented in [`src/dev/build/README.md`](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/dev/build/README.md#environment-variables) (section: Environment variables). Summary:

| Var | Default | Purpose |
|---|---|---|
| `OSD_BUILD_NO_PIGZ` | unset | Force pure-Node gzip even if `pigz` is on `PATH` |
| `OSD_BUILD_PIGZ_THREADS` | pigz default | `pigz -p N` |
| `OSD_BUILD_PIGZ_STALL_MS` | `300000` | Stall-watchdog timeout for `pigz` |
| `OSD_BUILD_GZIP_LEVEL` | `6` dev / `9` release | Override gzip level |
| `OSD_BUILD_ARCHIVE_CONCURRENCY` | `2` w/ pigz else cpuCount | Parallel archive cap |
| `OSD_BUILD_FPM_CONCURRENCY` | `tasks.length` | Parallel `fpm` cap |
| `OSD_BUILD_NO_HARDLINK` | unset | Disable hardlink-preferred `scanCopy` |

### Testing the changes

Unit tests (new + existing):
```sh
yarn test:jest src/dev/build
```
Integration tests (including `scan_copy` hardlink contract + `compressTar` pigz roundtrip):
```sh
yarn test:jest_integration src/dev/build/lib/integration_tests
```

Full release build benchmark on a 16-vCPU Linux box (requires pigz for the +pigz column):
```sh
for i in 1 2 3; do
  rm -rf build target .node_binaries
  /usr/bin/time -v yarn build --release --skip-os-packages 2>&1 | tee /tmp/bench/run-$i.log
done
```

New tests added:
- `src/dev/build/lib/fs.test.ts` — pigz error paths (exit-nonzero, signal, stall, source-missing).
- `src/dev/build/tasks/clean_tasks.test.ts` — `node_modules`-only extras patterns preserved outside `node_modules`.
- `src/dev/build/tasks/create_archives_task.test.ts` — in-flight counter asserts the parallel-archives concurrency cap.
- `src/dev/build/tasks/os_packages/create_os_package_tasks_concurrency.test.ts` — `OSD_BUILD_FPM_CONCURRENCY` bounding (default parallel, explicit cap, serialize with `=1`).
- `src/dev/build/tasks/write_sha_sums_task.test.ts` — sha1 + sha256 sidecars match `createHash`.
- `scan_copy.test.ts` — 5 new cases: same-FS inode share, `OSD_BUILD_NO_HARDLINK=1` fallback, `time`-provided skip, `EEXIST` propagation, and a `[CONTRACT]` regression test.
- `fs.test.ts` integration — `compressTar` pigz roundtrip (loud-skips when `pigz` is absent).

### Check List

- [x] All tests pass
  - [x] `yarn test:jest` — 92/93 in `src/dev/build` (single failure is pre-existing `get_build_number.test.ts`, reproduces on clean `main`)
  - [x] `yarn test:jest_integration` — 40 pass + 2 pre-existing skips in the touched directories
- [x] New functionality has been documented (`src/dev/build/README.md`).


src/dev/build/README.md
```diff
@@ -49,3 +49,17 @@ The majority of this logic is extracted from the grunt build that has existed fo
 [lib/build.js]: ./lib/build.js
 [build_distributables.js]: ./build_distributables.js
 [../tooling_log/tooling_log.js]: ../tooling_log/tooling_log.js
+
+# Environment variables
+
+All vars below are **optional**; the build works with none of them set. They exist for CI runners, release engineers, and debugging. Values are read at task entry, so exporting them in a shell before invoking `node scripts/build` is sufficient.
```
joshuali925, line 55: why env instead of arguments?

src/dev/build/tasks/create_archives_task.ts
```diff
@@ -30,21 +30,54 @@
...
+import { mkdirp, compressTar, compressZip, hasPigz, Task } from '../lib';
 
 const asyncStat = promisify(Fs.stat);
 
 export const CreateArchives: Task = {
   description: 'Creating the archives for each platform',
 
   async run(config, log, build) {
-    const archives = [];
-    // archive one at a time, parallel causes OOM sometimes
```
joshuali925, line 46: is this a concern? seems it was intentional to make archive sequential

src/dev/build/lib/scan_copy.ts
```diff
@@ -32,15 +32,25 @@ import Fs from 'fs';
...
 
 const statAsync = promisify(Fs.stat);
 const mkdirAsync = promisify(Fs.mkdir);
 const utimesAsync = promisify(Fs.utimes);
 const copyFileAsync = promisify(Fs.copyFile);
+const linkAsync = promisify(Fs.link);
 const readdirAsync = promisify(Fs.readdir);
 
+// Hardlinks share the inode so they're ~instant and use no extra disk. We fall back to a
+// reflink-preferred copy if linking fails (cross-filesystem, permissions, etc.). Opt out
```
joshuali925: does this actually fallback to reflink? I feel reflink is safer than hardlink (our CI OS should be RHEL-based so xfs?), but probably ok if CI and normal dev activities don't modify the copied files

src/dev/build/README.md
angle943: Good question — three reasons I went with env:

**1. Precedent in this build.** `node scripts/build` already reads `OSD_OPTIMIZER_MAX_WORKERS`, `OSD_OPTIMIZER_NO_CACHE`, etc.from env. Adding CLI flags for just the parallelism knobs would make the interface inconsistent.
 
 **2. Subprocess transparency.** `node scripts/build` spawns nested processes (grunt → use_node → build tasks → fpm/pigz). Env vars flow through all of them automatically; CLI flags would need plumbing through every Task signature and `config` object just to reach e.g. `compressTar` deep in `lib/fs.ts`.

**3. Audience.** These are operator/debugging knobs, not user-facing features. A release engineer tuning a CI runner wants `OSD_BUILD_ARCHIVE_CONCURRENCY=1` in the pipeline's env block next to `OSD_OPTIMIZER_MAX_WORKERS`. Most users shouldn't touch them; the defaults are picked to Just Work.

All of them are also optional — the build runs correctly with none set. Happy to reconsider if you'd rather some of these (e.g. `--gzip-level`, `--archive-concurrency`) be first-class CLI flags; I kept them as env for consistency with the existing ones.

src/dev/build/tasks/create_archives_task.ts
angle943: Legitimate flag — the old comment ("parallel causes OOM sometimes") is exactly why I was careful here. Three points on why I think the concern no longer applies:

*1. Concurrency is bounded, not unbounded.**
 - With `pigz`: cap **2** (pigz itself already spawns cpuCount internal threads, so 2× invocations saturate all cores — no benefit going higher).
 - Without `pigz`: cap **cpuCount**, but also capped by `platforms.length` (= 5).
 - Both capped again by `OSD_BUILD_ARCHIVE_CONCURRENCY` if operators want lower.
 - Setting `OSD_BUILD_ARCHIVE_CONCURRENCY=1` restores the exact old sequential behaviour.

**2. The OOM-risk model has changed since that comment was written.**
Archiver streams the tar; it doesn't buffer the whole platform tree. Peak memory is dominated by `pigz`'s dictionary (a few MB × cpuCount threads) plus Node's heap, not the archive contents. Two concurrent pigz processes on a 16-core box is ~100–200 MB RSS overhead, not gigabytes. The old comment likely predates the pigz path entirely — the old sequential code ran Node-zlib which buffered more aggressively.

**3. Empirically tested.**
- Local: n=6 full release builds on 16 vCPU / 64 GB with no OOM, no swap pressure.
- CI: `Build release artifact sanity check` on 4 vCPU `ubuntu-latest` passed both of our runs — that's the smallest runner in the matrix and it's the one most likely to OOM if there were a real problem.

Happy to add a peak-RSS measurement to the PR description if that would help confirm it. And if a specific pipeline does see pressure, the opt-out is one env var away.

src/dev/build/lib/scan_copy.ts
angle943: Good catch — you're right on both counts. The old comment was stale ("reflink-preferred" was aspirational, not what the code did), and reflink is the better primitive here. Fixed in 3f59fc5.

The flow is now three-tier, first-win:
1. **Reflink** via `fs.copyFile(src, dest, COPYFILE_FICLONE_FORCE | COPYFILE_EXCL)` — on CoW filesystems (xfs w/ reflink=1 which is RHEL 8+ default, btrfs, APFS), each copy gets an independent inode with copy-on-write. Writes to dest don't affect source, so the hardlink contract is moot on those filesystems.
2. **Hardlink** — used when reflink returns `ENOTSUP` (ext4, NTFS, HFS+) or any other error. Same behavior as before.
3. **Byte copy** — final fallback, same as before. The fallback warning now surfaces both reflink and hardlink error codes for diagnosis.

`COPYFILE_FICLONE_FORCE` (not plain `FICLONE`) is deliberate — the `_FORCE` variant errors fast with `ENOTSUP` on non-CoW filesystems, so we fall through to hardlink without silently byte-copying. Plain `FICLONE` would mask that degradation.

Tier ordering is pinned deterministically by new unit tests in `src/dev/build/lib/scan_copy.test.ts` that mock `fs.copyFile` + `fs.link` — so the assertion holds on every runner regardless of filesystem. The existing integration test was updated to detect which tier ran (via `srcStat.nlink`) and assert the matching semantic, so it passes on both CoW and non-CoW filesystems without relying on the runner's FS.

joshuali925 (approved):

ruchidh (approved):

PR #11866, opened 2026-04-29
Remove application-level bundle cache and rely on Rspack persistent cache

ruanyl:
### Description

Removes the custom application-level bundle-skipping cache from @osd/optimizer and relies on rspack's built-in persistent cache instead. This eliminates ~1,400 lines of redundant cache invalidation logic (file hashing, git-based cache keys, cache key diffing) while maintaining equivalent build performance through rspack's module-level caching.

### Behavior change

The old cache machenism will invalide the entire bundle even there is only one file changed, Rspack persisted cache is at module level which mean only the changed file get recompiled, developers are expected to gain generally faster build speed locally.

### Bug fix

Also fixes a pre-existing bug where the optimizer logged two "bundles compiled successfully" messages

### Improvements

Added build progress bar shows ██████████░░░░░░░░░░░░░░░░░░░░ 33% (20/60 bundles) during compilation
See screen record

https://github.com/user-attachments/assets/bd3bba9c-9f85-45e4-945e-1c6537e95885

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


packages/osd-optimizer/src/optimizer/optimizer_state.ts
```diff
@@ -151,18 +138,9 @@ export function createOptimizerStateSummarizer(
...
-    ) {
+    if (event.type === 'changes') {
       const onlineBundles: Bundle[] = [...state.onlineBundles];
-      if (event.type === 'changes') {
-        onlineBundles.push(...event.bundles);
-      }
-      if (event.type === 'bundle not cached') {
-        onlineBundles.push(event.bundle);
-      }
+      onlineBundles.push(...event.bundles);
```
angle943: **Potential watch-mode regression: success may never be reported after the first rebuild.**

`state.onlineBundles` is now seeded with `[...config.bundles]` on `'optimizer initialized'` (line ~114), so it already contains all N bundles. Every subsequent `changes` event `push`es `event.bundles` on top **without dedup**, so `state.onlineBundles.length` grows to `N + K` on each rebuild.

Meanwhile `getStatePhase` now requires `states.length >= totalOnlineBundles` (line 95) to declare success, and `compilerStates` is deduped by `bundleId` so its length stays ≤ N. After the first `changes` event the check `N >= N+K` is false forever, the `states.length < totalOnlineBundles` branch returns `'running'`, and the summarizer never transitions back to `'success'`.

The old `getStatePhase` didn't compare against `onlineBundles.length`, which is why this was latent before. Suggested fixes:

```ts
// option A: dedup on merge
const onlineBundles = Array.from(new Set([...state.onlineBundles, ...event.bundles]));

// option B: track as Set<string> of bundleIds throughout

// option C: reset state.onlineBundles to [] on 'success' and only push changed bundles
```

Please also add a watch-mode integration test that performs change → rebuild → asserts `success` is reached. The existing `watch_bundles_for_changes.test.ts` only exercises the watcher, not the summarizer, so it wouldn't catch this.

packages/osd-optimizer/src/run_optimizer.ts
```diff
@@ -67,32 +65,23 @@ export function runOptimizer(config: OptimizerConfig) {
...
+      // watch all bundles for changes in watch mode
       const changeEvent$ = config.watch
-        ? watchBundlesForChanges$(bundleCacheEvent$, startTime).pipe(share())
+        ? watchBundlesForChanges$(config.bundles, startTime).pipe(share())
         : Rx.EMPTY;
 
-      // run workers to build all the online bundles, including the bundles turned online by changeEvent$
-      const workerEvent$ = runWorkers(config, cacheKey, bundleCacheEvent$, changeEvent$);
+      // run workers to build all bundles, plus any changed bundles from watch mode
+      const workerEvent$ = runWorkers(config, config.bundles, changeEvent$);
```
angle943, line 82: **No-op rebuild may be meaningfully slower than before.**

The PR description says "equivalent build performance through rspack's module-level caching." That's true for a warm cache on a single bundle, but the old `bundleCacheEvent$` path short-circuited entirely when all bundles were up-to-date — no workers were spawned. Now every bundle always goes through `runWorkers`, which means N × (fork node + load `@rspack/core` + eval config + load persistent cache) even on a pure no-op restart. At ~60 bundles and 1–3s of worker cold-start per bundle, that's a real wall-clock regression for the common case of restarting the dev server without code changes.

Could you share before/after numbers for:

- Cold build (no cache)
- No-op restart (full cache hit, nothing changed)
- One-file edit rebuild

If the no-op case is significantly worse, would a thin mtime-based "nothing changed on disk since last successful run" short-circuit over bundle outputs be acceptable? That keeps the code dramatically simpler than the old git/hash system while preserving the instant-restart property.

packages/osd-optimizer/src/integration_tests/basic_optimization.test.ts
```diff
@@ -219,15 +208,9 @@ it('uses cache on second run and exist cleanly', async () => {
...
-      "initializing",
-      "initializing",
-      "initializing",
-      "initialized",
-      "success",
-    ]
-  `);
+  // All bundles go through workers, but rspack persistent cache makes it fast
+  const successStates = msgs.filter((msg) => msg.state.phase === 'success');
+  expect(successStates.length).toBeGreaterThanOrEqual(1);
```
angle943: This assertion is a tautology — any successful second run passes it, whether the persistent cache actually kicks in or not. The old version of this test asserted the exact state sequence (`initializing → initialized → success` with no `running`), which proved no worker ran.

Since the new design does always run workers, that exact sequence no longer applies, but we still need to prove the cache hits. A few options:

- Enable `stats.loggingDebug: /^rspack\.persistentCache$/` in the test webpack config and assert a cache-restore log line is present.
- Measure second-run duration relative to first-run and assert at least some ratio (e.g. `secondRun < 0.5 * firstRun`).
- Compare output artifact mtimes — unchanged modules shouldn't be rewritten if the cache hit.

Without something like this, the test won't catch a regression where the cache silently stops working.

packages/osd-optimizer/src/worker/webpack.config.ts
```diff
@@ -104,7 +104,30 @@ export function getWebpackConfig(bundle: Bundle, bundleRefs: BundleRefs, worker:
...
+            cache: {
+              type: 'persistent',
+              buildDependencies: [
+                __filename,
+                require.resolve('./theme_loader'),
+                require.resolve('./entry_point_creator'),
+                require.resolve('@osd/optimizer/postcss.config.js'),
+              ],
+              // Isolate caches when config that affects build output changes
+              version: JSON.stringify({
```
angle943, line 122: **`cache.version` is missing a per-bundle discriminator, and all bundles share one cache storage directory.**

Per the rspack v1 docs, the cache folder under `storage.directory` is derived from `config.name`, `config.mode`, `buildDependencies`, and `version`. `config.name` isn't set here, and `storage.directory` defaults to `node_modules/.cache/rspack`. So all N bundle compilations share one cache folder.

Two concerns:

1. **Concurrent writers.** N worker processes may race on the same cache directory. The rspack docs don't clearly state whether parallel writers are safe in 1.7. Worth confirming with the rspack team before shipping without scoping per bundle — corruption here would be very hard to debug (silent bad builds).
2. **Invalidation amplification.** One bundle's `buildDependencies` change (or a version bump) invalidates the shared cache for all bundles.

Suggest either:

```ts
name: bundle.id,  // rspack uses this in the cache folder path
```

or

```ts
storage: {
  type: 'filesystem',
  directory: Path.join(/* ... */, 'rspack', bundle.id),
},
```

Also consider including `worker.dist` in `version`. Rspack does invalidate on `config.mode` change, so that one may be implicitly covered, but an explicit value makes the intent obvious and defends against future rspack changes.

packages/osd-optimizer/src/worker/webpack.config.ts
```diff
@@ -104,7 +104,30 @@ export function getWebpackConfig(bundle: Bundle, bundleRefs: BundleRefs, worker:
...
     mode: worker.dist ? 'production' : 'development',
     context: Path.normalize(bundle.contextDir),
-    cache: true,
+    cache: worker.cache,
+    ...(worker.cache
+      ? {
+          experiments: {
+            cache: {
+              type: 'persistent',
+              buildDependencies: [
```
angle943, line 114: **`buildDependencies` is missing `.browserslistrc`.**

`browserslist.loadConfig({ path: worker.repoRoot })` is called at the top of `getWebpackConfig` and its result influences the swc/lightningcss output. If a developer edits `.browserslistrc` the cache will happily serve stale bundles.

```ts
buildDependencies: [
  __filename,
  require.resolve('./theme_loader'),
  require.resolve('./entry_point_creator'),
  require.resolve('@osd/optimizer/postcss.config.js'),
  Path.resolve(worker.repoRoot, '.browserslistrc'),
],
```

The recursive JS/TS import tracking mentioned in the rspack docs covers `bundle_deps_check_plugin.ts` and friends (because they're imported from this file), so you don't need to list those — but non-JS inputs like `.browserslistrc` and `postcss.config.js` have to be explicit.

packages/osd-optimizer/src/worker/webpack.config.ts
```diff
@@ -104,7 +104,30 @@ export function getWebpackConfig(bundle: Bundle, bundleRefs: BundleRefs, worker:
...
+                require.resolve('./entry_point_creator'),
+                require.resolve('@osd/optimizer/postcss.config.js'),
+              ],
+              // Isolate caches when config that affects build output changes
+              version: JSON.stringify({
+                themeTags: worker.themeTags,
+                browserslistEnv: worker.browserslistEnv,
+              }),
+              snapshot: {
+                unmanagedPaths: [/[\\/]node_modules[\\/]@osd[\\/]/],
```
angle943: **`unmanagedPaths` only excludes `@osd/*`.** The default `managedPaths` is `[/[\\/]node_modules[\\/][^.]/]`, which matches any scoped or unscoped package in `node_modules`. Workspace-symlinked packages under other scopes (or unscoped) will be trusted based on their `package.json` version, so local edits inside them won't invalidate the cache.

A quick way to keep this correct going forward is to generate the list from the workspace manifest (e.g. read `package.json`'s `workspaces`/`resolutions` and include each symlinked package path). If `@osd/*` is genuinely the only scope for workspace-internal linked packages, a comment here documenting that invariant would help avoid future regressions.

packages/osd-plugin-helpers/src/integration_tests/build.test.ts
```diff
@@ -104,7 +104,7 @@ it('builds a generated plugin into a viable archive', async () => {
   expect(processProcOutput(buildProc.all)).toMatchInlineSnapshot(`
     " info deleting the build and target directories
      info running @osd/optimizer
-     │ info initialized, 0 bundles cached
+     │ info initialized, 1 bundles to build
```
angle943, line 107: Nit, since you're touching this snapshot: `1 bundles to build` is grammatically wrong (was also wrong as `0 bundles cached`). Given the snapshot is already being updated, could be a good opportunity to fix the pluralization in `log_optimizer_state.ts` (`${n} bundle${n === 1 ? '' : 's'} to build`).

angle943 (commented):
Thanks for this — big +1 on deleting the hand-rolled cache in favor of rspack's persistent cache. Most of the mechanical removals look clean (verified no dangling references to `createCacheKey` / `optimizerCacheKey` / `getHashes` / `getChanges` / `BundleCacheEvent` elsewhere in `osd-optimizer`).

A few concerns I'd like to flag before this ships, left as inline comments. The main ones:

1. **Watch-mode can get stuck in `running`** after the first rebuild (correctness regression) — see `optimizer_state.ts`.
2. **No-op rebuild likely slower** than before because every bundle now spawns a worker, even on full cache hit — see `run_optimizer.ts`.
3. **The "uses rspack persistent cache on second run" test doesn't actually assert the cache hits** — see `basic_optimization.test.ts`.
4. **All bundles share a single cache scope** (no `config.name`, default `storage.directory`) — see `webpack.config.ts`. Worth confirming rspack 1.7 is safe with N parallel worker processes writing the same cache folder.

Rest are smaller: `buildDependencies` is missing `.browserslistrc`, `unmanagedPaths` only covers `@osd/*` (what about other workspace-symlinked packages?), and a snapshot pluralization nit.

Benchmarks for cold vs. no-op rebuild would also help justify the tradeoff.

packages/osd-plugin-helpers/src/integration_tests/build.test.ts
angle943: Above was from kiro. From Justin: kiro u are being too nitpick :P

packages/osd-optimizer/src/run_optimizer.ts
ruanyl: This is right to some extent: the old getBundleCacheEvent$ + cacheKey path short-circuited entirely for up-to-date bundles (no workers spawned). The new code always sends all bundles through `runWorkers`.

However: Workers are pooled via assignBundlesToWorkers — it's not N separate forks, it's min(N, maxWorkerCount) workers.

As you can see in the video, there isn't signicant difference between the warm start with cache vs the old cache

packages/osd-optimizer/src/worker/webpack.config.ts
ruanyl: > 1. Concurrent writers. 

The `storage.directory` defaults to `node_modules/.cache/rspack` relative to `context`, and each bundle has its own `context: bundle.contextDir` (each plugin's directory). So they already resolve to different physical paths. No race condition exists.

> 2. Invalidation amplification.

Consider each bundle has its own context dir, this is not a problem either. Plus all bundle share the same config, so once the config changed, all bundle cache should be invalidated.

ruanyl:
Hi @angle943 , I went trough the feedback and pushed a commit, please take another look, thanks

angle943 (dismissed):

Hailong-am:
> Rspack persisted cache is at module level which mean only the changed file get recompiled, developers are expected to gain generally faster build speed locally

will Rspack handle module dependencies like plugin A depends on plugin B?

Hailong-am:
### `--no-cache` disables in-memory caching too, which hurts CI cold builds

When `--no-cache` is set, `worker.cache` becomes `false`, and the webpack config sets:

```typescript
cache: worker.cache,  // false
```

This disables **all** caching — both the new persistent disk cache and rspack's in-memory cache within a single build run. The in-memory cache is still useful during a single compilation (shared modules across chunks), and the old code actually hardcoded `cache: true` for this reason.

Suggest keeping in-memory cache always on and only gating persistent cache:

```typescript
cache: true,  // in-memory cache is always beneficial within a single run
...(worker.cache ? {
  experiments: {
    cache: { type: 'persistent', ... }
  }
} : {}),
```

This way CI can pass `--no-cache` (or set `OSD_OPTIMIZER_NO_CACHE`) to skip persistent cache writes (no disk I/O overhead from cache artifacts that get thrown away) while still benefiting from in-memory caching during the build.

ruanyl:
> ### `--no-cache` disables in-memory caching too, which hurts CI cold builds
> When `--no-cache` is set, `worker.cache` becomes `false`, and the webpack config sets:
> 
> ```ts
> cache: worker.cache,  // false
> ```
> 
> This disables **all** caching — both the new persistent disk cache and rspack's in-memory cache within a single build run. The in-memory cache is still useful during a single compilation (shared modules across chunks), and the old code actually hardcoded `cache: true` for this reason.
> 
> Suggest keeping in-memory cache always on and only gating persistent cache:
> 
> ```ts
> cache: true,  // in-memory cache is always beneficial within a single run
> ...(worker.cache ? {
>   experiments: {
>     cache: { type: 'persistent', ... }
>   }
> } : {}),
> ```
> 
> This way CI can pass `--no-cache` (or set `OSD_OPTIMIZER_NO_CACHE`) to skip persistent cache writes (no disk I/O overhead from cache artifacts that get thrown away) while still benefiting from in-memory caching during the build.

Sounds good, I will update

ruanyl:
> > Rspack persisted cache is at module level which mean only the changed file get recompiled, developers are expected to gain generally faster build speed locally
> 
> will Rspack handle module dependencies like plugin A depends on plugin B?

Plugins are compiled separately, so Rspack won't handle this. But I think this is not a problem, for example, if plugin `A` imports module `X` from plugin `B`, when building `A`, `X` will never be compiled again, instead `A` will reference `X` at runtime from `B`, so the imports among plugins are not actually build time dependency.

Hailong-am (approved):

yubonluo (approved):

SuZhou-Joe (approved):

Issue #11577, opened 2026-03-24
[META] Integrate with SQL/PPL Error Reports


Swiddis:
**Is your feature request related to a problem? Please describe.**

The current error reporting from the SQL plugin isn't very descriptive, which propagates to a messy experience on the frontend for people whose queries fail for nontrivial reasons (i.e. not syntax errors).

**Describe the solution you'd like**

https://github.com/opensearch-project/sql/issues/4919 describes an approach to adding more context to SQL errors, there will be a meta issue linking this too. Once the context is available, we can:
- Integrate different pieces of context with the frontend (e.g. query position)
- Provide suggestions on how to repair queries
- Throw error context in the chat flow to help LLMs debug easier
- Integrate with explain-analyze features to suggest auto-fixes for different query components (e.g. suggest changing time windows when working with data outside a 15 minute window)

**Describe alternatives you've considered**

Integrating with any particular error context is optional, any error enhancement approach fundamentally needs to be incremental since there's hundreds of error cases (147 in SQL when I counted a couple months ago).

**Additional context**

See https://github.com/opensearch-project/sql/issues/5261

PR #11871, opened 2026-04-29
Error context integration

Swiddis:
### Description

Redoes some of the rendering for Discover's PPL errors, integrates it with AI, and integrates it with error reports.

In draft while testing, tweaking, adding auto tests.

### Issues Resolved

Related: #11577
Related: https://github.com/opensearch-project/sql/issues/4919

## Screenshot

image
image

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/tabs/error_guard/error_guard.tsx
```diff
@@ -31,26 +51,161 @@ export interface ErrorGuardProps {
...
+  const query = useSelector(selectQuery);
+
+  const chatService = services.core?.chat;
+  const isChatAvailable = chatService?.isAvailable() ?? false;
+
+  const errorContextData = useMemo(
+    () =>
+      error
+        ? {
+            id: `query-error-${Date.now()}`,
```
mengweieric: `Date.now()` in the id means `useDynamicContext`'s `deepEqual` check fails every render, which churns `removeContextById` then `addContext` against the assistant store. Since this represents "the currently displayed error," a stable id works fine:

```ts
id: 'query-error',
```

If you want it to change when the error changes, hash something deterministic off `error.errorBody` instead of wall-clock time.

src/plugins/explore/public/components/tabs/error_guard/error_guard.tsx
```diff
@@ -31,26 +51,161 @@ export interface ErrorGuardProps {
...
+  const useDynamicContext =
+    services.contextProvider?.hooks?.useDynamicContext || noopDynamicContext;
+  useDynamicContext(errorContextData, false);
+
+  const handleAskAi = useCallback(async () => {
+    if (!chatService || !isChatAvailable) return;
+
+    setIsAskingAi(true);
+    try {
+      const question = 'Why did my query fail and how can I fix it?';
```
mengweieric: A few strings slipped past i18n: `'Query Error'`, the Ask-AI question, and the toast `'Failed to send message to AI'`. The ones inside the store config (`description`, `label`) probably don't need translation, but the question and toast are user-visible.

src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_error_decorations.ts
```diff
@@ -0,0 +1,138 @@
...
+type IStandaloneCodeEditor = monaco.editor.IStandaloneCodeEditor;
+type IEditorDecorationsCollection = monaco.editor.IEditorDecorationsCollection;
+
+const PPL_FIELD_CONTEXTS = ['fields', 'where', 'stats by', 'sort by', 'rename', 'eval', 'parse'];
+
+const findFieldInQuery = (
+  model: monaco.editor.ITextModel,
+  fieldName: string
+): monaco.Range | null => {
+  const matches = model.findMatches(fieldName, true, false, true, null, false);
```
mengweieric, line 22: The whole point of the sql-plugin side of this work (opensearch-project/sql#5266) was giving us `query_pos: { line, column }` so we wouldn't have to guess. The type carries it through and the tests assert it round-trips, but `findFieldInQuery` ignores it and falls back to a keyword heuristic. Would you mind wiring it up?

```ts
const { query_pos, requested_field } = errorContext ?? {};
if (query_pos) {
  const len = requested_field?.length ?? 1;
  return new monaco.Range(
    query_pos.line, query_pos.column,
    query_pos.line, query_pos.column + len,
  );
}
// fall back to findFieldInQuery only when query_pos is absent
```

Nice side effect: also sidesteps the case-sensitivity issue in `findMatches` since you bypass the heuristic entirely.

src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_error_decorations.ts
```diff
@@ -0,0 +1,138 @@
...
+        {
+          range,
+          options: {
+            className: 'ppl-error-highlight',
+            glyphMarginClassName: 'ppl-error-glyph',
+            hoverMessage: {
+              value: `**Query Error**\n\n${errorDetails}\n\nField: \`${requestedField}\``,
+            },
+            minimap: {
+              color: '#ff0000',
```
mengweieric: Minimap and overview-ruler red won't adapt to dark mode. If it's easy, `euiThemeVars.euiColorDanger` from `@osd/ui-shared-deps/theme` covers both themes. Happy to land as a follow-up if you'd rather not expand scope.

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -653,16 +654,17 @@ const executeQueryBase = async (
...
-    let parsedError;
+    let parsedError: OpenSearchErrorResponse | undefined;
     try {
-      parsedError = JSON.parse(error.body.message);
+      parsedError = JSON.parse(error.body.message) as OpenSearchErrorResponse;
     } catch (parseError) {
       parsedError = {
         error: {
           reason: error.body?.message || error.message || 'Unknown Error',
           details: error.body?.error || 'An error occurred',
```
mengweieric: `details: error.body?.error || 'An error occurred'`: in Hapi/Boom, `error.body.error` is the status name (`"Bad Request"`), not the detail message. The actual message is in `error.body.message` (which you're already using for `reason`). When the JSON.parse fallback fires, the UI ends up showing a class name where details should go:

```ts
details: error.body?.message || error.message || 'An error occurred',
```

mengweieric (dismissed):
Really solid work on this. The structured error type is well-shaped, the test coverage in `utils.test.ts` is thorough, and `query_pos`-driven highlighting is going to be a meaningful UX win.

A few notes inline, nothing structural. Approving so this isn't blocked while you triage them.

src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_error_decorations.ts
Swiddis: I'm not sure how to solve cursor positioning on the frontend yet, it's complicated because we technically send a modified query with `where` clauses inserted early and newlines stripped, so the backend gives invalid values. Leaving it for future me, this just puts a starting hook in the monaco editor.

Probably the stable version of this will track edits to the query as part of pre-processing, and then play them backward with the returned cursor position to return the right place. Vaguely: `transform(q) => (q', slices)`, then later `untransform(query_pos, slices)`

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
Swiddis: flipped them

mengweieric (approved):

mengweieric (approved):

RyanL1997 (approved):

joshuali925:
### Code review

Found 2 issues:

1. The Monaco hover decoration falls back to `'Unknown error'` when the backend error has a `reason` but no `details`. OpenSearch PPL syntax errors commonly carry only `reason` (see the "only reason (no details)" case in `query_enhancements` tests), so the inline hover will read "Unknown error" even though a meaningful message is available. The error guard panel falls back to `reason`, but this decoration does not. Consider `?? queryStatus.error?.errorBody?.error?.reason ?? 'Unknown error'`.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a812c41ff33cdc2e6fc126b29aa3cc482c4cfad8/src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_error_decorations.ts#L88-L90

2. `findFieldInQuery`'s clause-context filter is effectively a no-op for single-line queries. It selects a field match with `match.range.startLineNumber >= keywordPos.lineNumber`, comparing only line numbers and never columns. Since most PPL queries are a single line, the keyword and every field match are all on line 1, so the condition is always true and the function returns the first text occurrence of the field regardless of clause. Also, `fullText.indexOf(contextKeyword)` does an unbounded substring match (it can match inside other identifiers), and the precise `query_pos` (line/column) from the backend error context is ignored in favor of this text re-search.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a812c41ff33cdc2e6fc126b29aa3cc482c4cfad8/src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/use_error_decorations.ts#L27-L37

joshuali925 (approved):

PR #11879, opened 2026-04-30
Disable Data Source management for unsupported OS Versions

LDrago27:
### Description

With the feature of single OSD. We are aiming to run the single Version of OSD across all backend OS version. For 1.3,2.11 version we don;t support the Data Source Management page this PR aims to disable it for those versions. 

## Screenshot

1.3
Screenshot 2026-04-30 at 1 32 25 PM

2.11
Screenshot 2026-04-30 at 1 41 19 PM

2.19
Screenshot 2026-04-30 at 1 43 26 PM

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data_source_management/public/plugin.ts
+++ b/src/plugins/data_source_management/public/plugin.ts
@@ -18,6 +18,7 @@ import {
 import { toMountPoint } from '../../../../src/plugins/opensearch_dashboards_react/public';
 import { DashboardDirectQuerySyncBanner } from './components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner';
 import { parseUrlHash } from '../../opensearch_dashboards_utils/public';
+import semver from 'semver';
 
 import { PLUGIN_NAME } from '../common';
 import { createDataSourceSelector } from './components/data_source_selector/create_data_source_selector';
@@ -121,6 +122,7 @@ export class DataSourceManagementPlugin
   private core: CoreStart | null = null;
   private currentAppId: string | undefined = undefined;
   private config: ConfigSchema;
+  private managementApp: any = null;
 
   constructor(initializerContext: { config: { get: () => ConfigSchema } }) {
     this.config = initializerContext.config.get();
@@ -148,7 +150,7 @@ export class DataSourceManagementPlugin
 
     this.featureFlagStatus = !!dataSource;
 
-    opensearchDashboardsSection.registerApp({
+    this.managementApp = opensearchDashboardsSection.registerApp({
       id: DSM_APP_ID,
       title: PLUGIN_NAME,
       order: 1,
@@ -255,6 +257,17 @@ export class DataSourceManagementPlugin
   public start(core: CoreStart): DataSourceManagementPluginStart {
     this.started = true;
     this.core = core;
+
+    if (!this.featureFlagStatus && this.managementApp) {
+      core.http
+        .get<{ version: string }>('/internal/data-source-management/localClusterVersion')
+        .then(({ version }) => {
+          if (semver.satisfies(version, '~1.3.0 || ~2.11.0')) {
+            this.managementApp.disable();
+          }
+        })
+        .catch(() => {});
+    }
 
     setApplication(core.application);
     setWorkspaces(core.workspaces);
--- a/src/plugins/data_source_management/public/plugin.ts
+++ b/src/plugins/data_source_management/public/plugin.ts
@@ -6,6 +6,7 @@
 import React from 'react';
 import { i18n } from '@osd/i18n';
 import { DataSourcePluginSetup } from 'src/plugins/data_source/public';
+import semver from 'semver';
 import {
   AppMountParameters,
   CoreSetup,
@@ -18,8 +19,8 @@ import {
 import { toMountPoint } from '../../../../src/plugins/opensearch_dashboards_react/public';
 import { DashboardDirectQuerySyncBanner } from './components/direct_query_data_sources_components/direct_query_sync/direct_query_sync_banner';
 import { parseUrlHash } from '../../opensearch_dashboards_utils/public';
-import semver from 'semver';
 
+import * as pluginManifest from '../opensearch_dashboards.json';
 import { PLUGIN_NAME } from '../common';
 import { createDataSourceSelector } from './components/data_source_selector/create_data_source_selector';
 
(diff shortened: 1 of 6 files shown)


sumukhswamy (dismissed):

seraphjiang (commented):
## Review

### Summary
This PR disables the Data Source Management page for OS 1.3.x and 2.11.x when running in single-cluster mode (MDS disabled). It adds a new server route to fetch the local cluster version and client-side logic to disable the management app based on that version.

### Issues

**1. Hard-coded version denylist is fragile**

```js
if (semver.satisfies(version, '~1.3.0 || ~2.11.0')) {
  this.managementApp.disable();
}
```

This is a denylist — you enumerate versions to block. Every new unsupported version requires a code change. The [version decoupling design](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5877) established an allowlist pattern via `supportedOSDataSourceVersions` in the plugin manifest. Consider using an allowlist approach instead (e.g., `!semver.satisfies(version, '>=2.12.0')`) so new unsupported versions are automatically excluded.

**2. New route duplicates existing infrastructure**

OSD core already fetches the local cluster version during startup via `opensearch.client.asInternalUser.info()`. The `opensearch` service exposes cluster info internally. Adding a new HTTP route for this creates a round-trip from browser → server → OpenSearch that may not be necessary. Consider whether the version can be passed from server to client during plugin setup (e.g., via `core.getStartServices()` or injected config) instead of a separate API call.

**3. Race condition — async disable in `start()`**

The version fetch is async in `start()`. If the user navigates to Data Source Management before the HTTP response arrives, they will see the page briefly before `disable()` takes effect. Depending on how `managementApp.disable()` works, it may not remove an already-rendered page.

**4. Silent error swallowing**

Server-side:
```ts
} catch (e) {
  return response.ok({ body: { version: '' } });
}
```

Client-side:
```ts
.catch(() => {});
```

Both silently swallow errors. If the version fetch fails, the management page stays enabled for potentially unsupported versions (fail-open). This may be intentional but should have a comment explaining the design choice. At minimum, log the error server-side.

**5. `validate: false` should be `validate: {}`**

```ts
validate: false,
```

`validate: false` disables all validation. Since this is a simple GET with no parameters, use `validate: {}` to explicitly declare no input expected. This avoids the security analyzer flag and is the recommended pattern.

**6. `managementApp` typed as `any`**

```ts
private managementApp: any = null;
```

Use the return type of `opensearchDashboardsSection.registerApp()` for type safety.

**7. Missing linked issue**

The PR description has an empty "Issues Resolved" section. A behavioral change like disabling a management page should have a tracking issue.

### Minor
- The semver range `~1.3.0 || ~2.11.0` is correct for matching 1.3.x and 2.11.x patch versions. No issue there.
- Tests cover the happy path and error case for the server route. Good.
- The `!this.featureFlagStatus` guard is correct — this only applies in single-cluster mode.

seraphjiang:
## Suggestion: Use the Version Decoupling Pattern Instead

The current approach (new route + hard-coded version denylist) works but doesn't scale. The [version decoupling design](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5877) already solved this problem for MDS — the local cluster should be treated the same way.

### Core Issue

This is a denylist:
```js
if (semver.satisfies(version, '~1.3.0 || ~2.11.0')) {
  this.managementApp.disable();
}
```
Every new unsupported version requires a code change. With the "single OSD" initiative, every plugin will face this same problem.

### Recommended: Manifest-Driven Allowlist

**Step 1** — Add `supportedOSDataSourceVersions` to the manifest (same field the version decoupling design established):

```json
{
  "id": "dataSourceManagement",
  "supportedOSDataSourceVersions": ">=2.12.0"
}
```

**Step 2** — Use it in `start()`:

```ts
import * as pluginManifest from '../../opensearch_dashboards.json';

if (!this.featureFlagStatus && this.managementApp) {
  core.http
    .get<{ version: string }>('/internal/data-source-management/localClusterVersion')
    .then(({ version }) => {
      if (
        version &&
        pluginManifest.supportedOSDataSourceVersions &&
        !semver.satisfies(version, pluginManifest.supportedOSDataSourceVersions)
      ) {
        this.managementApp.disable();
      }
    })
    .catch(() => {});
}
```

### Why This Matters

- **Allowlist, not denylist** — new unsupported versions are automatically excluded without code changes
- **Consistent with existing pattern** — `index_pattern_management` and `dataset_management` already use `supportedOSDataSourceVersions`, and the [alerting plugin](https://github.com/opensearch-project/alerting-dashboards-plugin/pull/1003) adopted the same approach
- **Sets the precedent** — since `data_source_management` owns the version decoupling infrastructure, it should use its own pattern. Other plugins facing the same "single OSD" problem can follow this example
- **Single place to update** — version range lives in the manifest JSON, not buried in code logic

Ideally the local cluster version would also be injected server-side during setup (avoiding the new route entirely), but the manifest-driven allowlist is the minimum change that aligns with the existing architecture.

LDrago27:
@seraphjiang The idea of using the plugin manifest was really good. I have updated the code to address the comments.

src/plugins/data_source_management/public/plugin.ts
```diff
@@ -256,6 +259,23 @@ export class DataSourceManagementPlugin
...
+            version &&
+            pluginManifest.supportedOSDataSourceVersions &&
+            !semver.satisfies(version, pluginManifest.supportedOSDataSourceVersions)
+          ) {
+            this.managementApp!.disable();
+          }
+        })
+        // Fail-open: if version fetch fails, keep the management page enabled
+        // rather than blocking access when the version is unknown
+        .catch(() => {});
```
zhongnansu, line 276: this is not fail open on the version fetch, since you already handled that by always returning `response.ok()` when it fails. This is more related to errors elsewhere other than fetching, so I'd suggest add logging in the catch here, so you know what failed

zhongnansu (commented):
@LDrago27 I have some questions, so all these seems to disable data source management UI from rendering if version doesn't satisfy.
1) why not just disabling the DSM plugin from loading?
2) what about the other plugin, data source plugin, which handles the actual connection?

angle943 (approved):

src/plugins/data_source_management/public/plugin.ts
angle943: @LDrago27 it could be good to log an error here. Someone might complain that they see this seemingly useless feature on an older version, and we can look into the logs to see why they are seeing it

ruchidh (approved):

PR #11887, opened 2026-05-01
[Bug] Dataset Discover loading

TackAdam:
### Description
  Fixes a 500 error (Unsupported index pattern type INDEXES) on initial Discover load when the default index pattern was created via the dataset management "Save as" flow. DQL/Lucene was being issued against a dataset that only supports PPL/SQL, and the request routed through the default search strategy which rejects any non-empty indexType.

  Builds on #11788 (search-source hydration for non-INDEX_PATTERN datasets). That PR made dashboard panels render for these datasets; this one fixes the language selection on initial load so Discover doesn't blow up on the same datasets.

  Root cause

  Three compounding issues:

  1. fetchDefaultDataset() hardcoded the dataset type to INDEX_PATTERN regardless of what the saved object actually stored. So when an INDEXES-typed saved object was the default, the dataset returned to callers lied about its type and the language guard never fired.
  2. getDefaultQuery() did not clamp the default language to the dataset's supportedLanguages. Even if the dataset type had been correct, DQL/Lucene would still be chosen for an INDEXES dataset that only supports PPL/SQL.
  3. QueryStringManager seeded query$ before datasetService existed, and DatasetService.init() is async. The BehaviorSubject emitted a stale default (no dataset, kuery), consumers cached it, and the post-init default was never applied. Classic Discover's use_search.ts then used the stale value as the initial query.

  Fix

  - dataset_service.ts — fetchDefaultDataset(): read indexPattern.type from the loaded saved object (fall back to INDEX_PATTERN when absent), and pass the real type through to toDataset().
  - query_string_manager.ts — getDefaultQuery(): clamp defaultLanguageId to the dataset's supportedLanguages when the current default isn't in the list. Mirrors the guard already in setQuery().
  - query_string_manager.ts — constructor ordering: instantiate datasetService and languageService before seeding query$, so the initial default can run the dataset-aware clamp.
  - query_string_manager.ts — new refreshDefaultQuery(previousDefault): re-seeds query$ from getDefaultQuery() after DatasetService.init() resolves, but only when the user hasn't touched the query yet.
  - query_service.ts — start(): snapshots the pre-init default, awaits the init promise, then calls
  refreshDefaultQuery(preInitDefault). This is the hook that connects the async init to the BehaviorSubject.
  - use_search.ts (classic Discover): when there is no saved-search query and dataQuery equals the default, prefer the freshly computed defaultQuery over the stale dataQuery.

  Tests

  - dataset_service.test.ts — fetchDefaultDataset preserves non-INDEX_PATTERN saved-object type; defaults to INDEX_PATTERN when the saved object has none.
  - query_string_manager.test.ts —
    - getDefaultQuery clamps unsupported language to first supported entry, keeps it when supported, tolerates undefined supportedLanguages.
    - refreshDefaultQuery re-seeds query$ when untouched; does not overwrite a query the user has already set.
  - Full existing suite still passes: 52/52 across both files.

  Manual verification

  - Created an index pattern via Dashboards Management → Data Sources → "Save as" (stores type: "INDEXES" on the saved object).
  - Set it as the default index pattern.
  - Reloaded Discover — DQL/Lucene no longer 500s; language defaults to PPL, payload routes through a strategy that accepts indexType: "INDEXES".
  - PPL/SQL still work; traditional INDEX_PATTERN defaults still default to DQL.

  Pending / follow-up

  Still open for owner input: DatasetService.saveDataset() writes type: dataset.type (e.g. INDEXES) into the
  index-pattern saved object. This PR makes every read path tolerant of that, but the real question is whether INDEXES on the saved index-pattern object is intentional:

  - If yes → this PR is the right long-term fix and nothing else is needed.
  - If no → the save path should normalize to INDEX_PATTERN at write time, and the existing INDEXES-typed saved objects will need a migration. The read-path fixes here remain valuable either way (defense in depth, plus they unblock existing customer data).

  Also worth flagging for a future PR: the UI should restrict the language-switcher to only show languages the current dataset supports, so a user can't manually select DQL against an INDEXES dataset and hit the same 500 that way.

## Screenshot

Before:
Before

After:
After

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/data/public/query/query_service.ts
+++ b/src/plugins/data/public/query/query_service.ts
@@ -101,7 +101,13 @@ export class QueryService {
     indexPatterns,
     application,
   }: QueryServiceStartDependencies): IQueryStart {
-    this.queryStringManager.getDatasetService().init(indexPatterns);
+    const preInitDefault = this.queryStringManager.getDefaultQuery();
+    this.queryStringManager
+      .getDatasetService()
+      .init(indexPatterns)
+      .then(() => {
+        this.queryStringManager.refreshDefaultQuery(preInitDefault);
+      });
     return {
       addToQueryLog: createAddToQueryLog({
         storage,
--- a/src/plugins/data/public/query/query_string/dataset_service/dataset_service.test.ts
+++ b/src/plugins/data/public/query/query_string/dataset_service/dataset_service.test.ts
@@ -445,6 +445,61 @@ describe('DatasetService', () => {
     // expect(service.getRecentDatasets().length).toEqual(4);
   });
 
+  test('fetchDefaultDataset preserves the saved object type for non-INDEX_PATTERN datasets', async () => {
+    jest.clearAllMocks();
+    uiSettings = coreMock.createSetup().uiSettings;
+    uiSettings.get = jest.fn().mockImplementation((setting: string) => {
+      if (setting === UI_SETTINGS.SEARCH_MAX_RECENT_DATASETS) return 4;
+      if (setting === 'defaultIndex') return 'id';
+      if (setting === UI_SETTINGS.QUERY_ENHANCEMENTS_ENABLED) return true;
+    });
+    sessionStorage = new DataStorage(window.sessionStorage, 'opensearchDashboards.');
+    service = new DatasetService(uiSettings, sessionStorage);
+
+    indexPatterns = ({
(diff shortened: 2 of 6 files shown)


ruchidh (commented):
Lgtm..

ruchidh (approved):

LDrago27 (approved):

PR #11889, opened 2026-05-04
Add parameter validation to /api/timeline/run

sumukhswamy:
### Description

Add input validation to prevent heap OOM crash caused by sending thousands of comma-separated expressions in a single sheet string.

- Enforce maxSize(10) on the sheet array and maxLength(65536) on each sheet string at the route schema validation layer
- Cap total parsed expressions per request to 100 in chain_runner, rejecting oversized payloads before any series allocation

```The limit of 100 expressions per request provides a safe margin above realistic usage while still blocking the attack vector. The PoC requires 15,000 expressions to crash the server 100 is two orders of magnitude below that threshold. In typical usage, a Timeline visualization has 1-5 series per panel, and even a complex dashboard with 10 sheets of 10 expressions each would only hit 100. Setting the cap here avoids breaking any legitimate power-user workflows while completely neutralizing the DoS vector. Combined with the maxSize: 10 on the sheet array, the worst case becomes 10 sheets × ~10 expressions each, which is well within normal memory bounds.```

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### CODE REVIEW

**Comment by cwperks** [comment]
@sumukhswamy Can I unit test be added for this?

File: src/plugins/vis_type_timeline/server/routes/run.ts [hunk 1]
```diff
@@ -62,7 +62,7 @@ export function runRoute(
       path: '/api/timeline/run',
       validate: {
         body: schema.object({
-          sheet: schema.arrayOf(schema.string()),
+          sheet: schema.arrayOf(schema.string({ maxLength: 65536 }), { maxSize: 10 }),
```
LDrago27, line 65: We should probably house these limits in a constants file. Also it might be good to know the reasoning behind the limits

PR #11899, opened 2026-05-08
Remove trailing comma since cypress 13 dont support

yyfamazon:
### Description

cypress 13 don't support "," at the end of file path: cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/vis_builder/*.js,

So I [deleted "," ](https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/.github/workflows/cypress-workflow-bundle-snapshot-based-ci-groups.yml#L49) in FT repo. But didn't do the same in osd

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/25488324732/job/74791480774?pr=11897
Some OSD cypress tests CI groups start to fail, these groups are running FTR tests, related to recent Cypress update.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/.github/workflows/cypress_workflow.yml
+++ b/.github/workflows/cypress_workflow.yml
@@ -369,7 +369,7 @@ jobs:
           IFS="," read -a SPEC_ARRAY <<< $(yarn --silent osd:ciGroup${{ matrix.group }})
           FORMATTED_SPEC=''
           for i in "${SPEC_ARRAY[@]}"; do
-            FORMATTED_SPEC+="cypress/integration/core-opensearch-dashboards/opensearch-dashboards/${i},"
+            FORMATTED_SPEC+="cypress/integration/core-opensearch-dashboards/opensearch-dashboards/${i}"
           done
           echo "SPEC=${FORMATTED_SPEC}" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/cypress_workflow.yml
```diff
@@ -369,7 +369,7 @@ jobs:
           IFS="," read -a SPEC_ARRAY <<< $(yarn --silent osd:ciGroup${{ matrix.group }})
           FORMATTED_SPEC=''
           for i in "${SPEC_ARRAY[@]}"; do
-            FORMATTED_SPEC+="cypress/integration/core-opensearch-dashboards/opensearch-dashboards/${i},"
+            FORMATTED_SPEC+="cypress/integration/core-opensearch-dashboards/opensearch-dashboards/${i}"
```
ruanyl, line 372: What if there were multiple items in `SPEC_ARRAY`? We still need a way to separate the file list, right?

wanglam, line 372: The current implementation will remove "," between different specs. What about just remove the last "," or doesn't append "," for the last spec item.

PR #11934, opened 2026-05-11
[Variable] Fix explore visualization crashes when dashboard variables change due to multi axis

yubonluo:
### Description

 - Fixed explore visualization crashes when dashboard variables change (e.g., switching function from max to min)
 - Fixed reuseAxesMapping method to properly handle multi-axis fields with array values
 - Fixed when there is only one variable on the dashboard, deleting the variable and refreshing the page will cause the variable to reappear without actually being deleted.
 - Fixed: Variable order constraints are enforced everywhere

### Issues Resolved
When dashboard variables change and cause column names to change (e.g., ${function} changes from max to min, resulting in column name change from max(field) to min(field)), explore visualizations would crash with the error:
Cannot load saved visualization with id xxx at explore_embeddable.tsx:520

Root cause:
The reuseAxesMapping method in VisualizationRegistry incorrectly treated array values as map keys when checking field types:

// Before (broken for multi-axis arrays):
const fieldType = columnTypeByName.get(savedAxesMapping[role]);
// When savedAxesMapping[role] = ["max(field)", "count"], this would fail

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/components/visualizations/visualization_registry.ts
+++ b/src/plugins/explore/public/components/visualizations/visualization_registry.ts
@@ -179,9 +179,9 @@ export class VisualizationRegistry {
    */
   public reuseAxesMapping(
     chartType: string,
-    savedAxesMapping: Record<string, string>,
+    savedAxesMapping: AxisFieldNameMappings,
     allColumns: VisColumn[]
-  ): Record<string, string> | undefined {
+  ): AxisFieldNameMappings | undefined {
     const rules = this.getVisualization(chartType)?.getRules();
     if (!rules) return undefined;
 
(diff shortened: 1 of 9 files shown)


src/plugins/explore/public/components/visualizations/visualization_registry.ts
```diff
@@ -179,54 +179,104 @@ export class VisualizationRegistry {
...
           return false;
         }
+
         return savedRoles.every((role) => {
-          const fieldType = columnTypeByName.get(savedAxesMapping[role]);
-          return !fieldType || fieldType === (m as TypeMapping)[role].type;
+          const savedFieldNames = toFieldArray(savedAxesMapping[role]);
+          const expectedType = (ruleMapping as AxisTypeMapping)[role as AxisRole]?.type;
+          return savedFieldNames.every((savedFieldName) => {
+            const savedfieldType = columnTypeByName.get(savedFieldName);
```
ruanyl: Typo: `savedfieldType` -> `savedFieldType`

src/plugins/explore/public/components/visualizations/visualization_registry.ts
```diff
@@ -179,54 +179,104 @@ export class VisualizationRegistry {
...
+      if (!ruleConfig) continue;
+
+      const { type: expectedType, multi: isMulti } = ruleConfig;
+      const existingFields = toFieldArray(result[role]);
+
+      const availableColumns = allColumns.filter(
+        (c) => c.schema === expectedType && !used.has(c.name)
+      );
+
+      if (isMulti) {
```
ruanyl: Have you tested the cases with y_2 axis?

src/plugins/explore/public/components/visualizations/visualization_registry.ts
yubonluo: Got it, will update it

src/plugins/explore/public/components/visualizations/visualization_registry.ts
yubonluo: yes, It works well:
test with ```source=opensearch_dashboards_sample_data_flights| where `FlightDelay` = true | stats ${function}(`FlightDelayMin`) as ${function}_delay, count(`FlightDelayMin`) as delay_count by `OriginWeather` ```and change the variable
image

src/plugins/explore/public/components/visualizations/visualization_registry.ts
```diff
@@ -179,57 +179,123 @@ export class VisualizationRegistry {
...
+      const used = new Set<string>();
+      const missingRoles: string[] = [];
+
+      for (const role of savedRoles) {
+        const savedValue = savedAxesMapping[role];
+        const fieldNames = toFieldArray(savedValue);
+        const expectedType = matchedRule[role as AxisRole]?.type;
+        const survivingFields = fieldNames.filter((name) => isValidField(name, expectedType));
+
+        if (survivingFields.length === 0) {
```
Qxisylolo, line 242: nit: what about simplifying the code:
```
 for (const role of savedRoles) {
    const savedValue = savedAxesMapping[role];
    const fieldNames = toFieldArray(savedValue);
    const expectedType = matchedRule[role as AxisRole]?.type;
    const survivingFields = fieldNames.filter((name) => isValidField(name, expectedType));
  
    if (survivingFields.length > 0) {
      result[role] = survivingFields.length === fieldNames.length
        ? savedValue 
        : Array.isArray(savedValue) ? survivingFields : survivingFields[0];
      survivingFields.forEach((f) => used.add(f));
    }
  
    if (survivingFields.length < fieldNames.length) {
      missingRoles.push(role);
    }
  }
  
```

Qxisylolo (approved):

ruanyl (approved):

PR #11939, opened 2026-05-11
feat(data): auto-refresh index pattern fields in the background

### LINKED ISSUE #11938: [Feature] Auto-refresh index pattern fields in the background

When a new field is ingested into an index covered by an existing index pattern, that field does not appear in Discover, Visualize, Dashboard or the field selector until the user manually navigates to **Management → Index Patterns → \<pattern\> → Refresh field list** and waits for the round-trip. This is a recurring point of friction during data exploration: users see partial data, blame the visualization, and only later discover that the field list was stale.

Kibana solves this transparently by refreshing the field list in the background while the user navigates, so newly-discovered fields show up without any manual action. OpenSearch Dashboards still requires the explicit click.

A stale-while-revalidate refresh in `IndexPatternsService.get(id)`:

1. The cached `IndexPattern` is returned to the caller immediately (no added latency on the hot path).
2. A non-blocking background task refetches `_field_caps` if a configurable TTL has elapsed for that pattern.
3. The returned field list is diffed against the in-memory one. If anything changed, the saved object is persisted; otherwise nothing is written.
4. When *new* fields are discovered (additions, not just removals), a one-shot info toast surfaces the change once per session per pattern.
5. Refresh errors are swallowed — the manual *Refresh field list* button remains the channel for actionable error surfacing, so a transient cluster issue does not spam toasts.

Configuration is opt-out via three new Advanced Settings under the **data** category, defaulting to a sensible behaviour out of the box:

| Setting | Type | Default |
|---|---|---|
| `indexPattern:autoRefreshFields` | boolean | `true` |
| `indexPattern:autoRefreshFieldsIntervalMs` | number (min 30 000) | `300000` (5 min) |
| `indexPattern:notifyOnNewFields` | boolean | `true` |

To keep the cost predictable under contention, the implementation also adds:

- A short-lived (5 s TTL) server-side cache + in-flight deduplication on `_field_caps` in `IndexPatternsFetcher` so that simultaneous callers from different sessions flatten into a single cluster round-trip.
- A client-side coalescer that, before persisting a diff, checks whether another client has already written a superset of the local field list — and if so adopts that state instead of writing again. Under N concurrent observers of the same new field, this turns N writes + N-1 409 retries into 1 write + N-1 reads.

No backend changes are needed. The `/api/index_patterns/_fields_for_wildcard` endpoint and the saved-object schema are reused as-is, so the change is migration-free.

- **Periodic global polling** of every cached index pattern: simple but wasteful because it refreshes patterns nobody is actively looking at, and scales badly with the number of saved patterns.
- **Refresh on mount of Discover/Visualize/Dashboard**: less invasive at the service level but requires changes in every consumer plugin and misses scenarios like opening the field selector in a saved search.
- **\"Fields on-the-fly\" model (Kibana 8 data views)**: longer-term refactor where field lists are not persisted at all and are recomputed on each access with a server-side cache. Significantly larger surface area and saved-object migration; out of scope for an incremental fix.

The proposed approach is intentionally minimal — it builds on the existing `refreshFields` + `updateSavedObject` paths and only touches one core service plus the fetcher.

**Additional context**

- The cost added is bounded to *at most* one `_field_caps` request per index pattern per TTL window per active session (default 12/hour/pattern); the diff guarantees zero writes when nothing changed.
- Discover, Visualize, Dashboard and `index_pattern_management` need no changes: they pick up new fields transparently through the existing field-list reference. The well-known limitation that an already-rendered sidebar does not re-render until the next mount remains (a follow-up could add a `Subject` on `IndexPatternFieldList` to push updates), but the saved object and any next mount already reflect the new fields.
- Implementation, tests and rationale are in the companion PR.

dfradehubs:
## Description

Brings parity with Kibana for index pattern field freshness: newly-ingested fields appear in Discover, Visualize and Dashboard automatically, without making the user click **Management → Index Patterns → Refresh field list** manually.

Implements the design proposed in #11938.

### What this PR adds

**Stale-while-revalidate refresh in `IndexPatternsService.get(id)`**
- The cached `IndexPattern` is returned to the caller immediately — no added latency on the hot path.
- A non-blocking background task refetches mappings via the existing `/api/index_patterns/_fields_for_wildcard` endpoint when a configurable TTL has elapsed for that pattern.
- The new field list is diffed against the in-memory one; the saved object is persisted only when there is a real diff (zero writes otherwise).
- A one-shot info toast announces brand-new fields, once per session per pattern, with a separate setting to silence it.
- Background errors are swallowed so a transient backend failure never disrupts the UI flow that triggered `get()` — the manual *Refresh field list* button remains the channel for actionable error surfacing.
- An in-flight `Map` deduplicates simultaneous refreshes within the same session.

**Three new Advanced Settings (category: `data`)**

| Setting | Type | Default |
|---|---|---|
| `indexPattern:autoRefreshFields` | boolean | `true` |
| `indexPattern:autoRefreshFieldsIntervalMs` | number (min 30 000) | `300000` (5 min) |
| `indexPattern:notifyOnNewFields` | boolean | `true` |

**Server-side cache + in-flight dedupe on `_field_caps`** (`IndexPatternsFetcher`)
- Module-level cache with a deliberately short 5 s TTL — `_field_caps` reflects per-request OpenSearch permissions, so a wider window risks serving one user's view of the schema to another.
- Concurrent callers with the same cache key share a single in-flight Promise.
- Bounded memory: entries are swept past a 200-entry soft cap, in-flight entries are dropped on rejection so retries work.
- Net effect: 100 concurrent clients hitting the same pattern within the TTL window become a single `_field_caps` round-trip to the cluster.

**Client-side coalescer for saved-object persistence**
- Before `updateSavedObject`, the refresher reads the latest persisted version. If another client already wrote a field set that is a *superset* of the local one, we adopt that state (replace fields + bump local version) and skip the write entirely.
- If the remote write is not a superset (we still have something new), we proceed with the write, letting `updateSavedObject`'s existing 409 retry+merge handle the conflict.
- Under N concurrent observers of the same new field: 1 write + N-1 reads instead of N writes + N-1 409 retries.

### What this PR does **not** change

- No backend changes; the `_fields_for_wildcard` endpoint is reused.
- No saved-object schema or migration — the persisted format is identical; the last-refresh timestamp lives only in memory and resets cleanly on reload.
- No changes in `discover`, `visualize`, `dashboard` or `index_pattern_management`. The manual *Refresh field list* button still works exactly as before; they pick up the new fields transparently through the existing field-list reference.

### Known limitation (out of scope)

Once a Discover or Visualize page has already rendered its sidebar, it does not re-subscribe to changes on `IndexPatternFieldList.replaceAll`. The newly-detected fields are present in memory and persisted to the saved object, but appearance in the *current* page still requires a navigation or page reload. Any *new* mount over the same pattern shows them immediately. A follow-up could add a `Subject` on `IndexPatternFieldList` to push updates to live consumers; intentionally not bundled here to keep the surface area small.

## Issues Resolved

Closes #11938

## Testing

- 27 jest tests across two suites, all passing:
  - `src/plugins/data/common/index_patterns/index_patterns/index_patterns.test.ts` — 11 preexisting + 10 new (auto-refresh + coalescer scenarios).
  - `src/plugins/data/server/index_patterns/fetcher/index_patterns_fetcher.test.ts` — 6 new (cache hit, TTL expiry, concurrent dedupe, distinct keys not coalesced, rejection clears in-flight entry, cache shared across instances).
- `i18n_check` passes for the 7 new translation keys.
- Local pre-commit hook (i18n + eslint) green on all four touched files.

### Smoke test plan

- [ ] Lower `indexPattern:autoRefreshFieldsIntervalMs` to `15000`, ingest a new field into an existing index, switch index pattern in Discover and back — new field appears without clicking refresh, and an info toast is shown once.
- [ ] Disable `indexPattern:autoRefreshFields` — new fields do not appear until the manual *Refresh field list* button is pressed (regression direction).
- [ ] Verify the manual *Refresh field list* button still works.
- [ ] Verify Advanced Settings shows the three new entries under the *data* category with the documented defaults and validation (`autoRefreshFieldsIntervalMs` ≥ 30 000).

## Check List

- [x] All tests pass
- [x] New functionality has been documented in the code (jsdoc + setting descriptions)
- [x] Commits are signed per the [DCO](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/CONTRIBUTING.md#developer-certificate-of-origin)

The release-notes / changelog fragment will be added in a follow-up commit on this branch once GitHub assigns this PR a number.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


nicklas-dohrn:
this seems really good, and also solves a real issue we and our customers face.
having this possibly automated is really helpful to observability usecases that do not follow set schemata.

Issue #11928, opened 2026-05-09
[RFC]Data Transformations Panel


Qxisylolo:
# Introduction

### What it is

Data Transformation is a new feature added to Explore visualization editor that allows users to reshape query results before they reach the visualization layer—without changing the underlying query.

Transformations are applied through a user-defined pipeline, where each step builds on the output of the previous one. This creates a flexible, composable workflow for manipulating data directly within the editor.

### What problem it solves

Explore visualization is purely query-driven — any data manipulation required modifying and re-running the query.
Which creates Pain Points:

1. Each visualization is tightly coupled to a specific query, so even small chart changes requires rewriting and re-running the query before validating results, which may interrupts exploratory analysis.

Data Transformation decouples this — users can now transform query results in real-time, without touching the query itself.

```
   Query Results (Raw Data)
        ↓
┌───────────────────┐
│ Transformation 1  │  ← can be moved up/down
├───────────────────┤
│ Transformation 2  │  ← can be edited, removed, or reordered
├───────────────────┤
│ Transformation 3  │  ← new steps can be inserted
└───────────────────┘
        ↓
    Transformed Data
        ↓
    Visualization

```

# POC

https://github.com/user-attachments/assets/93ae36bb-f3f7-43cb-bad7-11c56835efb1

# Use Cases

###  Use Case 1 — Multiple Visualizations from a Single Query

A user is monitoring service health and wants to build three different visualizations from the same log data. Without transformations, they'd need to write and maintain a separate query for each chart. With the Transform tab, they write the query once and shape the data differently per visualization.

User provided: the Base Query 

```
source = service_logs
| where timestamp >= '2024-01-01' AND timestamp <= '2024-01-31'
| fields timestamp, status_code, response_time_ms, host, error_message
```

This returns the full month of logs — all fields, all rows, unsorted. 

#### Transformation 1 — Response Time Trend (Line Chart)

```
1. Sort By
   └─ field: timestamp
   └─ order: ascending

2. Filter Fields
   └─ mode: include
   └─ fields: [timestamp, response_time_ms, host]
```

The transform layer sorts the rows chronologically and filters out the irrelevant columns so the line chart only sees what it needs.

#### Transformation 2  — Top Slow Requests (Bar Chart)

```
1. Filter
   └─ field: response_time_ms
   └─ operator: greater_than
   └─ value: 500

2. Sort By
   └─ field: timestamp
   └─ order: descending

3. Limit
[...]

PR #11944, opened 2026-05-12
[Explore vis]add data transformation panel

Qxisylolo:
### Description
This pr adds data transformation panel:

Data Transformation is a new feature added to Explore visualization editor that allows users to reshape query results before they reach the visualization layer—without changing the underlying query.

Transformations are applied through a user-defined pipeline, where each step builds on the output of the previous one. This creates a flexible, composable workflow for manipulating data directly within the editor.

RFC:
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11928

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/application/in_context_vis_editor/hooks/use_transformation_service.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+import { useMemo, useEffect } from 'react';
+import { TransformationService } from '../data_transformations/transformation_service';
+import { useQueryBuilderState } from './use_query_builder_state';
+import { useVisualizationBuilder } from './use_visualization_builder';
+import { getServices } from '../../../services/services';
+import { registerAllTransformations, UrlTransformationState } from '../data_transformations';
+
+let globalTransformationService: TransformationService | undefined;
```
FriedhelmWS: Can we not putting the service at global scope?

src/plugins/explore/public/embeddable/explore_embeddable.tsx
```diff
@@ -189,6 +203,22 @@ export class ExploreEmbeddable
...
+      return;
+    }
+
+    try {
+      const savedPipeline: UrlTransformationState[] = JSON.parse(
+        JSON.parse(this.savedExplore.visualization).dataTransformationJSON
+      );
+      this.transformationService.restoreFromState(savedPipeline);
+    } catch (error) {
+      // skip failed pipeline
```
FriedhelmWS: Why we decide to swallow error sliently for failed piepline here?

src/plugins/explore/public/application/in_context_vis_editor/data_transformations/transformation_service.ts
```diff
@@ -0,0 +1,268 @@
...
+  toggleTransformationHide,
+  deriveSchemaFromRows,
+} from './transformation_utils';
+import { OpenSearchSearchHit } from '../../../types/doc_views_types';
+import { TRANSFORMATION_STATE_KEY } from '../types';
+
+export interface UrlTransformationState {
+  definitionId: string; // defination identifier
+  instanceId: string; // unique instance identifier
+  config: any;
```
FriedhelmWS: Can we declare a type for config here, and avoid any, similar to other place for config in this PR.

src/plugins/explore/public/application/in_context_vis_editor/hooks/use_transformation_service.ts
Qxisylolo: thanks for the comment. we need a global singleton to maintain pipelines to stay consistent among components. I assume it is fine, as 
1. in-context editor is a page-level application — only one transformation pipeline is ever active at a time. 
2. cleanupGlobalTransformationService() is called when VisualizationEditorPage unmounts, ensuring the service is destroyed.
I think using singleton is a more lightweight approach compared to alternatives (context or service registry)

src/plugins/explore/public/application/in_context_vis_editor/component/save_vis_button.tsx
```diff
@@ -125,6 +127,19 @@ export const SaveVisButton = () => {
       isTitleDuplicateConfirmed,
       onTitleDuplicate,
     }: OnSaveProps) => {
+      const pipeline = transformationService.pipeline$.getValue();
+      const serializedPipeline = pipeline.map((instance) => {
+        const definition = transformationService
+          .getDefinitions()
+          .find((def) => def.label === instance.label);
```
FriedhelmWS: Nit: is that possible to have a id for this look up? I think label are i18n strings that is intent for only displaying.

Qxisylolo: good catch, fixed

src/plugins/explore/public/application/in_context_vis_editor/hooks/use_transformation_service.ts
FriedhelmWS: I think from logic prespective putting the singleton here at module level is totally ok, as we already assume there will be only one editor and one pipline. 

What I was thinking is putting a service at this level looks bit suspicious and we are actually makes the lifecycle a bit implicit, someone else use this without properly clean up can cuase slient state leak. Also, for all the tests to be added, we need to properly call the `cleanupGlobalTransformationService` in `afterEach` and otherwise we can have flaky tests.

But anyway, I don't think this is a problem logically, feel free to go with it.

src/plugins/explore/public/components/data_transformations/transformation_service.ts
```diff
@@ -0,0 +1,268 @@
...
+  clearPipeline(): void {
+    this.pipeline$.next([]);
+    // Also clear URL state so it doesn't restore the old pipeline on next render
+    if (this.urlStateStorage) {
+      this.urlStateStorage.set(TRANSFORMATION_STATE_KEY, [], { replace: true });
+    }
+  }
+
+  restoreFromState(states: UrlTransformationState[]): void {
+    if (!states || !Array.isArray(states) || states.length === 0) return;
```
FriedhelmWS, line 106: Nit: what if the user cleared the pipeline and saved, result in empty `[]` here?

Qxisylolo: restoreFromState is only used in URL restoration and saved explore pipeline restoration. in both case, the pipeline has just been initialized (pipeline is already []) , so need to trigger unnecessary pipeline$.next([]) again

src/plugins/explore/public/application/in_context_vis_editor/component/save_vis_button.tsx
```diff
@@ -125,6 +126,14 @@ export const SaveVisButton = () => {
       isTitleDuplicateConfirmed,
       onTitleDuplicate,
     }: OnSaveProps) => {
+      const pipeline = transformationService.pipeline$.getValue();
+      const serializedPipeline = pipeline.map((instance) => ({
+        definitionId: instance.definition_id,
+        instanceId: instance.instance_id,
```
ruanyl: why need to persist `instance_id`? It seems it's purely a runtime identifier

src/plugins/explore/public/application/in_context_vis_editor/component/save_vis_button.tsx
```diff
@@ -137,6 +146,7 @@ export const SaveVisButton = () => {
         splitField: visConfig?.splitField,
         splitLayout: visConfig?.splitLayout,
         showSplitLabel: visConfig?.showSplitLabel,
+        dataTransformationJSON: JSON.stringify(serializedPipeline),
```
ruanyl: Curious why does it need `JSON.stringify()` for `serializedPipeline`?

src/plugins/explore/public/components/visualizations/visualization_container.tsx
```diff
@@ -42,64 +47,75 @@ export interface UpdateVisualizationProps {
...
+    useEffect(() => {
+      if (results) {
+        const rows = results.hits?.hits || [];
+        const fieldSchema = results.fieldSchema || [];
+        visualizationBuilder.handleData(rows, fieldSchema);
       }
-    },
-    [services, dispatch]
-  );
+    }, [visualizationBuilder, results, pipeline]);
```
ruanyl: `VisualizationContainer` should not known the exists of `pipeline`, better if we can decouple transformation pipeline from VisualizationContainer.

I guess you added `pipeline` to deps array because you want it to trigger `handleData` to apply data transformations. But I think this should be VisualizationBuilder internal logic.

Qxisylolo: Got it, VisualizationBuilder will do the subscription.

src/plugins/explore/public/application/in_context_vis_editor/component/save_vis_button.tsx
Qxisylolo: Yes, we don't need to save it, will delete it

src/plugins/explore/public/application/in_context_vis_editor/component/save_vis_button.tsx
Qxisylolo: We don’t need to double-stringify this, I assume it is the leftover code from the initial changes.

Hailong-am (commented):
## Review Comments

### 1. Mutation inside `applyPipeline` (`transformation_service.ts:148`)

```typescript
if (cleaned !== instance.config) {
  instance.config = cleaned;  // mutates the object inside the BehaviorSubject
  pipelineChanged = true;
}
```

This directly mutates an object that's currently held by `pipeline$`. While it won't cause an infinite loop (the debounced re-invocation self-terminates because the cleaned config becomes stable), it violates the otherwise-immutable contract established by `updateTransformationConfig`, `removeTransformation`, etc. (which all create new objects).

The practical risk: any code that captured a reference to `pipeline$.getValue()` before `applyPipeline` ran would see its instances mutated in place unexpectedly.

Suggested fix — use an index-based loop and clone the instance:

```typescript
let pipelineChanged = false;
const updatedInstances = [...instances];

for (let i = 0; i < updatedInstances.length; i++) {
  let instance = updatedInstances[i];
  schemaMap.set(instance.instance_id, currentSchema);

  if (instance.validateConfig) {
    const cleaned = instance.validateConfig(instance.config, currentSchema);
    if (cleaned !== instance.config) {
      instance = { ...instance, config: cleaned };
      updatedInstances[i] = instance;
      pipelineChanged = true;
    }
  }
  // ...rest of loop
}

if (pipelineChanged) {
  this.pipeline$.next(updatedInstances);
}
```

---

### 2. Unused `transformationService` prop on `VisualizationContainer`

In `visualization_editor_bottom_left_container.tsx`:

```typescript
return <VisualizationContainer transformationService={transformServices} />;
```

But `VisualizationContainer` is defined with `React.memo()` and uses hooks internally — it never declares or consumes a `transformationService` prop. This prop is silently ignored (React doesn't error on extra props). It should be removed to avoid confusion.

src/plugins/explore/public/components/data_transformations/transformation_service.ts
```diff
@@ -0,0 +1,274 @@
...
+  private debouncedPipeline$ = this.pipeline$.pipe(debounceTime(300));
+
+  private urlStateStorage?: IOsdUrlStateStorage;
+  private urlSyncSubscription?: Subscription;
+
+  /**
+   * transformation catalog management
+   */
+  registerDefinition<TConfig>(definition: TransformationDefinition<TConfig>): void {
+    this.definitions.set(definition.id, definition as TransformationDefinition);
```
Hailong-am, line 50: if id is duplicate, should we log a warning message here?

Hailong-am (dismissed):

ruanyl (approved):

ruanyl (dismissed):

src/plugins/explore/public/components/data_transformations/transformation_service.ts
Qxisylolo: Thanks for the suggestion，fixed

Hailong-am (approved):

ruanyl (approved):

PR #11949, opened 2026-05-13
[Chat] Add maturity-based sync polling and resend affordance for tool result dispatch

wanglam:
### Description

   This PR improves the reliability of tool result dispatch by addressing an instability in the sync polling mechanism and adding user-facing recovery when dispatch cannot proceed.

   **Problem:**

   When a tool call completes locally, the client polls agentic memory to confirm the tool call id exists before sending the tool result. Two issues existed:

   1. **Transient sync observations:** The tool call id would appear in one poll response but disappear in the next (due to memory rewrites). Sending the tool result based on a single observation could
   result in an orphaned tool result — one that references a tool call id the LLM does not recognize. This triggers a validation error on the LLM side and invalidates the entire session.

   2. **Timeout fallback sent anyway:** When polling exhausted all attempts without confirming sync, the old code sent the tool result regardless. But an orphaned tool result corrupts the session rather
   than just losing a single turn.

   **Changes:**

   - **Maturity-based sync polling:** Require 3 consecutive successful observations of the tool call id before treating it as stably synced. Any pending or error response resets the streak. Increased max
   attempts from 10 to 15. Polling sleep is now interruptible via `AbortSignal`.
   - **Skip dispatch on timeout:** When polling exhausts without confirming the tool call id, dispatch is skipped. A resendable system message with the tool result payload is appended to the timeline so
   the user can retry.
   - **Skip dispatch when thread id is missing:** Previously threw an error. Now returns a typed skip reason (`no_thread_id`) so callers can surface a user-visible message.
   - **Abort signal propagation:** `sendToolResult` accepts an `AbortSignal` that flows through sync polling → agent dispatch → response stream. `cancelToolResultDispatch()` aborts both pre-dispatch
   polling and post-dispatch agent stream. Starting a new send automatically aborts any in-flight previous send.
   - **Resend UI:** `ErrorRow` renders a "Resend tool result" button when the system message carries `canResend: true` and `toolResult`. The button is disabled during send. `ChatWindow` removes the error
   message and calls `sendToolResultToAssistant` directly.
   - **Type changes:** Added `toolCallId`, `canResend`, and `toolResult` fields to `SystemMessage`.

   ## Screenshot

[video.webm](https://github.com/user-attachments/assets/80cffece-9d5c-4c1d-8176-54c7c9fd1381)

   ## Testing the changes

   1. Trigger a tool call in chat that requires sync polling (non-full-history memory provider).
   2. Verify the tool result is sent successfully after 3 consecutive synced observations.
   3. Simulate a sync timeout (e.g., by disconnecting the backend briefly) — verify a "Resend tool result" button appears.
   4. Click the resend button — verify the error message is removed and the tool result is re-dispatched.
   5. Click "Stop" during tool result dispatch — verify the abort cancels immediately without leaving stale state.

   ### Check List

   - [x] All tests pass
     - [x] `yarn test:jest`

### DIFF
--- a/src/core/public/chat/types.ts
+++ b/src/core/public/chat/types.ts
@@ -64,6 +64,21 @@ export interface DeveloperMessage extends BaseMessage {
 export interface SystemMessage extends BaseMessage {
   role: 'system';
   content: string;
+  /**
+   * When set, this system message corresponds to a failed tool result
+   * submission for the given tool call. The UI may use this (together with
+   * `canResend`) to offer a resend affordance.
+   */
+  toolCallId?: string;
+  /**
+   * Whether the associated tool result submission can be retried by the user.
+   */
+  canResend?: boolean;
+  /**
+   * The tool result payload to resend. Stored directly on the timeline message
+   * so no external map is needed.
+   */
+  toolResult?: any;
 }
 
 /**
--- a/src/plugins/chat/public/components/chat_input.tsx
+++ b/src/plugins/chat/public/components/chat_input.tsx
@@ -15,7 +15,6 @@ import { SlashCommandMenu } from './slash_command_menu';
 import { ChatContextPopover } from './chat_context_popover';
 import { useOpenSearchDashboards } from '../../../opensearch_dashboards_react/public';
 import { CoreStart } from '../../../../core/public';
-import { ConfirmationRequest } from '../services/confirmation_service';
 
 import './chat_input.scss';
 
@@ -24,8 +23,8 @@ interface ChatInputProps {
   input: string;
   isCapturing: boolean;
   isStreaming: boolean;
-  isSendingToolResult?: boolean;
-  pendingConfirmation?: ConfirmationRequest | null;
+  disabled?: boolean;
+  placeholder?: string;
   onInputChange: (value: string) => void;
   onSend: () => void;
   onStop: () => void;
@@ -39,8 +38,8 @@ export const ChatInput: React.FC<ChatInputProps> = ({
   input,
   isCapturing,
   isStreaming,
-  isSendingToolResult = false,
-  pendingConfirmation,
+  disabled = false,
+  placeholder,
   onInputChange,
   onSend,
   onStop,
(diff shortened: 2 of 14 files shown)


src/core/public/chat/types.ts
```diff
@@ -64,6 +64,21 @@ export interface DeveloperMessage extends BaseMessage {
...
+  toolCallId?: string;
+  /**
+   * Whether the associated tool result submission can be retried by the user.
+   */
+  canResend?: boolean;
+  /**
+   * The tool result payload to resend. Stored directly on the timeline message
+   * so no external map is needed.
+   */
+  toolResult?: any;
```
yubonluo, line 81: Does the tool result payload may be large (such as API responses containing a large amount of data)

src/plugins/chat/public/components/chat_input.tsx
```diff
@@ -124,7 +120,7 @@ export const ChatInput: React.FC<ChatInputProps> = ({
             fullWidth
             resize="none"
             rows={2}
-            disabled={isSendingToolResult || !!pendingConfirmation}
+            disabled={disabled}
```
yubonluo, line 123: Even when streaming, the stop button will be disabled. This may not be the expected behavior - users should always be able to stop generating

yubonluo (approved):

Hailong-am (approved):

src/core/public/chat/types.ts
wanglam: For now, both result of create_investigation and execute_ppl_query tool won't too large. Should not be a big concern for now.

src/plugins/chat/public/components/chat_input.tsx
wanglam: Here is the logic for the chat input disabled:
```
disabled={isSendingToolResult || hasPendingResend || !!pendingConfirmation}
``` 
If the chat is in streaming, it won't be disabled and can be stoped. It should the sam as previous implementation.

PR #11493, opened 2026-03-10
feat(opensearch): Add Transport extension point to OpenSearch service for client customization

eemijuntunen:
### Description
Related RFC - https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11470

Add registerClientTransport(), hasClientTransport(), and getClientTransport() to the OpenSearch service contracts, allowing plugins to register a custom Transport class that intercepts all HTTP-level communication with the backend cluster.

The Transport class is a built-in extension point in the @opensearch-project/opensearch client library. When registered, the custom Transport is injected into ClientOptions.Transport for all ClusterClient instances created by core.

This enables plugins to transparently transform requests and responses at the HTTP serialization layer without modifying existing OSD plugins or UI code. When no Transport is registered, there is zero runtime overhead.

Core changes:
- types.ts: registerClientTransport, hasClientTransport on setup contract; getClientTransport on internal start contract
- opensearch_service.ts: Store Transport class, expose registration/check/getter, recreate data client on start
- configure_client.ts: Accept optional customTransport param
- cluster_client.ts: Thread customTransport through constructor
- plugin_context.ts: Expose to new plugin setup context
- legacy_service.ts: Expose to legacy plugin setup context

Includes unit tests for registration, double-registration guard, client recreation with transport, and getClientTransport retrieval.

## Testing the changes

UT

## Changelog
- feat: Add Transport extension point to OpenSearch service for client customization

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] Update [CHANGELOG.md](./../CHANGELOG.md)
- [ ] Commits are signed per the DCO using --signoff

Issue #11470, opened 2026-03-06
[RFC] Legacy Elasticsearch Backend Support for OpenSearch Dashboards


eemijuntunen:
## 1. Introduction

This RFC adds ES 6.x/7.x backend support to OSD using a core Transport extension point and a self-contained compatibility plugin.

1. The **core extension point** lets plugins register a custom Transport class for the OpenSearch client, enabling request/response interception at the HTTP layer
2. The **compatibility plugin** houses all Elasticsearch-specific translation logic in a Transport subclass with adapter modules organized by API domain

Every client API method (`search`, `bulk`, `index`, etc.) goes through `transport.request()`. The plugin subclasses `Transport`, overrides `request()`, and translates requests/responses based on the detected backend version. With the plugin enabled, core OSD functionality (saved objects, search, index patterns, sample data, visualizations) works against ES 6.0–6.8 and 7.0–7.10.2 without changes to other plugins. See section 7.3 for limitations. The feature is opt-in so when disabled, nothing changes.

**Status:** The `backend_compatibility` plugin is experimental. Its configuration and translation behavior may change in future releases based on community feedback. The core Transport extension point (`registerClientTransport`, `hasClientTransport`, `getClientTransport`) is a stable API.

**Scope:** The primary goal is enabling core OSD functionality against legacy Elasticsearch backends. This includes saved objects, search, index patterns, visualizations, and sample data. Expanding compatibility to support additional OSD plugins (alerting, security, observability, etc.) that communicate with backend-specific APIs is considered future work and out of scope for the initial implementation.

### Architecture Overview

```
┌─────────────────────────────────────────────────────────────┐
│                   OpenSearch Dashboards                      │
│                                                              │
│  Plugins call client.search(), client.bulk(), etc.          │
│                          │                                   │
│                          ▼                                   │
│              ┌───────────────────────┐                       │
│              │   OpenSearch Client   │                       │
│              └───────────┬───────────┘                       │
│                          │                                   │
│              ┌───────────▼───────────┐                       │
│              │ CompatibilityTransport │  ◄── Extension Point │
│              │  • Detect backend      │                       │
│              │  • Route by path       │                       │
│              │  • Adapt request       │                       │
│              │  • Normalize response  │                       │
│              └───────────┬───────────┘                       │
└──────────────────────────┼───────────────────────────────────┘
                           │
          ┌────────────────┼────────────────┐
          ▼                ▼                ▼
[...]

PR #11961, opened 2026-05-14
feat(plugins): Add backend_compatibility plugin for legacy ES support

eemijuntunen:
### Description

   Adds the `backend_compatibility` plugin that enables OpenSearch Dashboards to connect to legacy Elasticsearch 6.x and 7.x clusters.

   The plugin registers a custom `Transport` class with core (via `registerClientTransport` from #11493) that intercepts all HTTP-level communication at the single `transport.request()` convergence point. It applies version-appropriate translations:

   - **ES 6.x**: Full request/response translation — `_type` injection, typed mappings, `interval` syntax, `_seq_no` synthesis, field type downgrades, `/_resolve/index` synthesis
   - **ES 7.x**: Response-only normalization — strip `_type` from document hits
   - **OpenSearch**: Pass-through with zero overhead

   The plugin is experimental and disabled by default: `backendCompatibility.enabled: true`

   **Key design decisions:**
   - Single interception point via Transport subclass (not per-method Proxy wrapping)
   - Declarative route table for path-based dispatch
   - Lazy backend detection on first request via `GET /`
   - Zero core knowledge of Elasticsearch — all logic lives in the plugin

   **Scope:** This PR covers the new opensearch-js client path, which handles core OSD functionality (saved objects, search, index patterns, visualizations, dashboards). Code paths still using the legacy elasticsearch-js client (index patterns field
   discovery, autocomplete, usage collection) bypass the Transport and are not covered. Full legacy client compatibility is planned as a follow-up via either migrating remaining legacy client usage to opensearch-js (preferred) or adding a
   `legacyRequestInterceptor` hook (prototype exists).

   **References:**
   - RFC: #11470
   - Core Transport extension: #11493

   ### Issues Resolved

   closes #11492

   ## Screenshot

   N/A — server-only plugin with no UI changes.

   ## Testing the changes

   **Unit tests:**

   `yarn test:jest src/plugins/backend_compatibility --no-cache`

   **Manual validation against ES 6.8 / 7.10.2:**

   1. Start an ES cluster:

   `docker run -d -p 9200:9200 -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch-oss:6.8.23`

   Or for ES 7.10.2:

   `docker run -d -p 9200:9200 -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch-oss:7.10.2`

   2. Add to `opensearch_dashboards.yml`:
      - `opensearch.hosts: ["http://localhost:9200"]`
      - `backendCompatibility.enabled: true`

   3. Start OSD: `yarn start`

   4. Verify core functionality works: saved objects, index patterns, search, visualizations, dashboards.

   ### Check List

   - [x] All tests pass
     - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/backend_compatibility/README.md
+++ b/src/plugins/backend_compatibility/README.md
@@ -1,0 +1,87 @@
+# Backend Compatibility Plugin
+
+Enables OpenSearch Dashboards to connect to legacy Elasticsearch 6.x and 7.x clusters.
+
+## How It Works
+
+The plugin registers a custom `Transport` class with the OpenSearch client. Every client API call (`search`, `bulk`, `index`, etc.) passes through `transport.request()` — making it the single point where all cluster communication can be intercepted and adapted.
+
+```
+OSD Plugin calls client.search(params)
+    │
+    ▼
+CompatibilityTransport.request(params)
+    │
+    ├─ Detect backend (lazy, once per instance)
+    ├─ Match URL path against route table
+    ├─ Translate request for target backend
+    ├─ Execute request via underlying Transport
+    └─ Normalize response to OpenSearch format
+    │
+    ▼
+OpenSearch-compatible response returned to caller
+```
+
+## Backend Handling
+
+| Backend | Request | Response |
+|---------|---------|----------|
+| OpenSearch 1.x+ | Pass-through | Pass-through |
+| Elasticsearch 7.0–7.10.2 | Pass-through | Strip `_type` from documents |
+| Elasticsearch 6.0–6.8 | Full translation | Full normalization |
+
+When disabled, no Transport is registered — zero runtime overhead.
+
+## Key Design Decisions
+
+**Single interception point.** Instead of wrapping 15+ individual client methods with Proxies, the Transport intercepts at the HTTP serialization boundary where all methods converge into one `request()` call.
+
+**Lazy detection.** Backend version is detected on the first request via `GET /`. This naturally handles timing issues where saved object migrations run before the plugin's `start()` phase.
+
+**Declarative route table.** URL path patterns are matched against an ordered array of entries. Each entry maps to request/response translator functions. Adding support for a new API means adding one entry.
+
+**Zero core knowledge of Elasticsearch.** Core provides a generic hook (`registerClientTransport`). All version detection and translation logic lives in the plugin.
+
+## What Gets Translated (ES 6.x)
+
+| Area | Translations |
+|------|-------------|
+| Documents | `_type` injection, `if_seq_no`/`if_primary_term` stripping, `/_create` → `/_doc` rewrite, `/_update` path rewrite |
+| Search | `calendar_interval` → `interval`, unsupported agg/query removal, `hits.total` normalization |
+| Mappings | Typed ↔ typeless format, `include_type_name`, field type downgrades |
+| Field Caps | Strip unsupported parameters |
+| Scroll | Response normalization |
+| Index Resolution | Synthesize `/_resolve/index` from `/_cat/indices` + `/_cat/aliases` |
+
+## Configuration
+
+```yaml
+# opensearch_dashboards.yml
+backendCompatibility.enabled: true  # Default: false
+```
+
+## File Structure
+
+```
+server/
+├── plugin.ts                      # Registers Transport with core
+├── config.ts                      # Schema (enabled: boolean)
+└── transport/
+    ├── compatibility_transport.ts # Transport subclass, route table, request() override
+    ├── backend_detector.ts        # GET / → BackendInfo
+    ├── types.ts                   # BackendInfo, DEFAULT_DOCUMENT_TYPE
+    └── adapters/
+        ├── search_adapter.ts      # Search/msearch/aggregation transforms
+        ├── document_adapter.ts    # Bulk/CRUD/_type/seq_no handling
+        ├── mapping_adapter.ts     # Typed mappings, field type downgrades
+        ├── field_caps_adapter.ts  # Parameter filtering
+        ├── scroll_adapter.ts      # Scroll response normalization
+        └── normalization_utils.ts # isPlainObject, synthesizeSeqNo, normalizeTotalHits
+```
+
+## Limitations
+
+- Optimistic concurrency disabled on ES 6.x (concurrent writes can silently overwrite)
+- Single Transport class registration (sufficient but not extensible to multiple plugins)
+- Scripted field syntax differences not auto-translated
+- `track_total_hits` unavailable on ES 6.x (counts above 10,000 may be inexact)
--- a/src/plugins/backend_compatibility/server/plugin.ts
+++ b/src/plugins/backend_compatibility/server/plugin.ts
@@ -3,9 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { first } from 'rxjs/operators';
 import { CoreSetup, CoreStart, Logger, Plugin, PluginInitializerContext } from 'src/core/server';
-import { BackendCompatibilityConfig } from './config';
 import { CompatibilityTransport } from './transport/compatibility_transport';
 import { BackendInfo } from './transport/types';
 import { PLUGIN_NAME } from '../common';
@@ -28,17 +26,9 @@ export class BackendCompatibilityPlugin
   }
 
   public async setup(core: CoreSetup): Promise<BackendCompatibilityPluginSetup> {
-    const config = await this.initializerContext.config
-      .create<BackendCompatibilityConfig>()
-      .pipe(first())
-      .toPromise();
-
-    if (!config.enabled) {
-      this.logger.info(`${PLUGIN_NAME} is disabled`);
-      return {};
-    }
-
-    this.logger.info(`Setting up ${PLUGIN_NAME}`);
+    this.logger.warn(
+      `[EXPERIMENTAL] ${PLUGIN_NAME} is enabled. Legacy backend compatibility is experimental and may change in future releases.`
+    );
     core.opensearch.registerClientTransport(CompatibilityTransport);
     this.logger.info('CompatibilityTransport registered with core');
     return {};
--- a/src/plugins/backend_compatibility/server/plugin.ts
+++ b/src/plugins/backend_compatibility/server/plugin.ts
@@ -1,0 +1,56 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { first } from 'rxjs/operators';
+import { CoreSetup, CoreStart, Logger, Plugin, PluginInitializerContext } from 'src/core/server';
+import { BackendCompatibilityConfig } from './config';
+import { CompatibilityTransport } from './transport/compatibility_transport';
+import { BackendInfo } from './transport/types';
+import { PLUGIN_NAME } from '../common';
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface BackendCompatibilityPluginSetup {}
+
+export interface BackendCompatibilityPluginStart {
+  getBackendInfo: () => BackendInfo | undefined;
+}
+
+export class BackendCompatibilityPlugin
+  implements Plugin<BackendCompatibilityPluginSetup, BackendCompatibilityPluginStart> {
+  private readonly logger: Logger;
+  private readonly initializerContext: PluginInitializerContext;
+
+  constructor(initializerContext: PluginInitializerContext) {
+    this.logger = initializerContext.logger.get();
+    this.initializerContext = initializerContext;
+  }
+
+  public async setup(core: CoreSetup): Promise<BackendCompatibilityPluginSetup> {
+    const config = await this.initializerContext.config
+      .create<BackendCompatibilityConfig>()
+      .pipe(first())
+      .toPromise();
+
+    if (!config.enabled) {
+      this.logger.info(`${PLUGIN_NAME} is disabled`);
+      return {};
+    }
+
+    this.logger.info(`Setting up ${PLUGIN_NAME}`);
+    core.opensearch.registerClientTransport(CompatibilityTransport);
+    this.logger.info('CompatibilityTransport registered with core');
+    return {};
+  }
+
+  public async start(core: CoreStart): Promise<BackendCompatibilityPluginStart> {
+    return {
+      getBackendInfo: () => CompatibilityTransport.lastDetectedBackend ?? undefined,
+    };
+  }
+
+  public stop() {
+    this.logger.info(`Stopping ${PLUGIN_NAME}`);
+  }
+}
--- a/src/plugins/backend_compatibility/server/transport/adapters/document_adapter.ts
+++ b/src/plugins/backend_compatibility/server/transport/adapters/document_adapter.ts
@@ -4,7 +4,10 @@
  */
 
 import { BackendInfo, DEFAULT_DOCUMENT_TYPE } from '../types';
-import { isPlainObject, synthesizeSeqNo } from './normalization_utils';
+import { isPlainObject, synthesizeSeqNo, extractQueryString } from './normalization_utils';
+
+/** Bulk action verbs used in bulk request/response processing */
+const DOC_ACTIONS = ['index', 'create', 'update', 'delete'] as const;
 
 /**
  * ES 6.x doesn't support if_seq_no/if_primary_term — stripping disables OCC.
@@ -30,7 +33,7 @@ export function translateBulkRequest(params: any, backend: BackendInfo): any {
 
   const transformed = source.map((item: any) => {
     if (!item || typeof item !== 'object') return item;
-    for (const action of ['index', 'create', 'update', 'delete']) {
+    for (const action of DOC_ACTIONS) {
       if (action in item) {
         const meta = { ...item[action] };
         if (!meta._type && !meta.type) {
@@ -53,8 +56,7 @@ export function translateBulkRequest(params: any, backend: BackendInfo): any {
 
 export function translateCreateRequest(params: any, backend: BackendInfo): any {
   const newPath = params.path.replace('/_create', '/_doc');
-  const existing =
-    typeof params.querystring === 'object' && params.querystring !== null ? params.querystring : {};
+  const existing = extractQueryString(params);
   const qs = stripSeqNoFromQuerystring({ ...existing, op_type: 'create' });
   return { ...params, path: newPath, querystring: qs };
 }
@@ -95,13 +97,14 @@ export function translateDeleteByQueryRequest(params: any, backend: BackendInfo)
 
 export function translateResponse(response: any, backend: BackendInfo): any {
   const body = response?.body || response;
-  if (!body) return response;
+  if (!body || typeof body !== 'object') return response;
 
   if (body._id !== undefined) normalizeDocResponse(body);
 
   if (Array.isArray(body.items)) {
     body.items.forEach((item: any) => {
-      for (const action of ['index', 'create', 'update', 'delete']) {
+      if (!item || typeof item !== 'object') return;
+      for (const action of DOC_ACTIONS) {
         if (item[action]) normalizeDocResponse(item[action]);
       }
     });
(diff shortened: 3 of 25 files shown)


angle943:
## Kiro Review

**Overall:** Promising architecture — intercepting at the Transport layer is the right abstraction level. However, a few issues need addressing before this leaves draft.

---

### Critical

**1. Race condition in backend detection**

Multiple concurrent requests hitting the transport before detection completes will each see `this.backend === null` and potentially enter `performDetection()` simultaneously. The `detecting` promise should be a proper singleton:

```ts
// Should be:
if (!this.detecting) { this.detecting = this.performDetection(); }
await this.detecting;
```

**2. Silent fallback to OpenSearch on detection failure**

After 3 failed detection attempts, defaulting to `{ distribution: 'opensearch', version: '0.0.0' }` is dangerous. If the backend is actually ES 6.x but detection failed (network blip, auth issue), requests will be sent without translation — potentially corrupting data (missing `_type` fields, wrong mappings). This should either:
- Throw and prevent the client from operating (fail-closed)
- Retry indefinitely with backoff
- At minimum, log at ERROR level, not just warn

**3. No tests in the commit**

The PR description references unit tests but the automated reviewer flagged "No relevant tests." Tests should be included in the PR for the translation logic to be reviewable.

---

### Design Concerns

**4. `lastDetectedBackend` should be private** — currently leaks infrastructure topology (ES 6.x vs 7.x vs OpenSearch) to any code that imports the class.

**5. Scope gap needs explicit documentation** — the legacy `elasticsearch-js` client paths (index patterns field discovery, autocomplete, usage collection) are not covered. Users need a clear compatibility matrix of what works and what doesn't against ES 6.x.

**6. Experimental warning** — if `backendCompatibility.enabled: true` is set, the plugin should log a prominent startup warning that this is experimental.

---

### Questions

- How are conflicts handled in the declarative route table? First-match or most-specific?
- What happens with bulk requests, scroll contexts, and aliases in ES 6.x mode?
- Is there a plan to add integration tests against actual ES 6.8/7.10 Docker containers in CI?

---

Architecture is solid. Once tests are added and the race condition + fallback behavior are fixed, this will be in good shape for a full review.

ruchidh (dismissed):

src/plugins/backend_compatibility/README.md
```diff
@@ -0,0 +1,87 @@
...
+| Elasticsearch 7.0–7.10.2 | Pass-through | Strip `_type` from documents |
+| Elasticsearch 6.0–6.8 | Full translation | Full normalization |
+
+When disabled, no Transport is registered — zero runtime overhead.
+
+## Key Design Decisions
+
+**Single interception point.** Instead of wrapping 15+ individual client methods with Proxies, the Transport intercepts at the HTTP serialization boundary where all methods converge into one `request()` call.
+
+**Lazy detection.** Backend version is detected on the first request via `GET /`. This naturally handles timing issues where saved object migrations run before the plugin's `start()` phase.
```
huyaboo, line 39: Is the backend logic already existing logic or are we adding it in this plugin?

src/plugins/backend_compatibility/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+**Declarative route table.** URL path patterns are matched against an ordered array of entries. Each entry maps to request/response translator functions. Adding support for a new API means adding one entry.
```
huyaboo, line 41: Could this quickly balloon/become harder to scale with future releases? I guess since this only applies to legacy versions so new APIs aren't added all that often if at all.

src/plugins/backend_compatibility/server/plugin.ts
```diff
@@ -0,0 +1,56 @@
...
+  public async setup(core: CoreSetup): Promise<BackendCompatibilityPluginSetup> {
+    const config = await this.initializerContext.config
+      .create<BackendCompatibilityConfig>()
+      .pipe(first())
+      .toPromise();
+
+    if (!config.enabled) {
+      this.logger.info(`${PLUGIN_NAME} is disabled`);
+      return {};
+    }
```
huyaboo: BTW, the core plugin resolution logic will automatically handle this and the plugin lifecycle will never be started for this plugin if `enabled = false`. You don't need to handle this.

src/plugins/backend_compatibility/server/transport/types.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+export type BackendDistribution = 'opensearch' | 'elasticsearch';
+
+export interface BackendInfo {
+  distribution: BackendDistribution;
+  version: string;
+  majorVersion: number;
+  minorVersion: number;
+  patchVersion: number;
```
huyaboo, line 13: Whats the difference between `version` and major/minor/path versions? If `version` is just the stringified version, then we probably don't need the redundant field.

src/plugins/backend_compatibility/server/transport/adapters/mapping_adapter.ts
```diff
@@ -0,0 +1,240 @@
...
+  }),
+
+  histogram: (def) => ({
+    type: 'object',
+    enabled: false,
+  }),
+
+  wildcard: (def) => {
+    const { ...rest } = def;
+    return { ...rest, type: 'keyword' };
```
huyaboo, line 59:
```suggestion
    return { ...def, type: 'keyword' };
```

src/plugins/backend_compatibility/server/transport/adapters/document_adapter.ts
```diff
@@ -0,0 +1,122 @@
...
+    body.docs.forEach((doc: any) => {
+      if (doc && !doc.error) normalizeDocResponse(doc);
+    });
+  }
+
+  return response;
+}
+
+function normalizeDocResponse(doc: any): void {
+  delete doc._type;
```
huyaboo, line 123: Can we add null/undefined guards?

src/plugins/backend_compatibility/server/transport/adapters/document_adapter.ts
```diff
@@ -0,0 +1,122 @@
...
+function stripSeqNoFromQuerystring(
+  qs: Record<string, any> | undefined
+): Record<string, any> | undefined {
+  if (!qs) return qs;
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  const { if_seq_no, if_primary_term, ...rest } = qs;
+  return rest;
+}
+
+export function translateBulkRequest(params: any, backend: BackendInfo): any {
```
huyaboo, line 25: For a lot of these translators, they can be `Record<string, any>` since it's always an object. Let's avoid using `any` directly and maybe have a type `TransportRequest`/`TransportResponse`

src/plugins/backend_compatibility/server/transport/adapters/document_adapter.ts
```diff
@@ -0,0 +1,122 @@
...
+  const source = Array.isArray(params.bulkBody)
+    ? params.bulkBody
+    : Array.isArray(params.body)
+    ? params.body
+    : null;
+  if (!source) return params;
+
+  const transformed = source.map((item: any) => {
+    if (!item || typeof item !== 'object') return item;
+    for (const action of ['index', 'create', 'update', 'delete']) {
```
huyaboo: These actions are repeated multiple times. For clarity, can we move `['index', 'create', 'update', 'delete']` to an exported constant `DOC_ACTIONS`

src/plugins/backend_compatibility/server/transport/adapters/document_adapter.ts
```diff
@@ -0,0 +1,122 @@
...
+    body: undefined,
+    bulkBody: transformed,
+    querystring: stripSeqNoFromQuerystring(params.querystring),
+  };
+}
+
+export function translateCreateRequest(params: any, backend: BackendInfo): any {
+  const newPath = params.path.replace('/_create', '/_doc');
+  const existing =
+    typeof params.querystring === 'object' && params.querystring !== null ? params.querystring : {};
```
huyaboo: maybe have this as a utility function `extractQueryString(params: Record<string, any>)` as this has been repeated several times

src/plugins/backend_compatibility/server/transport/adapters/mapping_adapter.ts
```diff
@@ -0,0 +1,240 @@
...
+ */
+
+import { BackendInfo, DEFAULT_DOCUMENT_TYPE } from '../types';
+import { isPlainObject } from './normalization_utils';
+
+interface FieldDefinition {
+  type?: string;
+  properties?: Record<string, FieldDefinition>;
+  enabled?: boolean;
+  dynamic?: boolean | string;
```
huyaboo: Nit: we can be more specific:


```suggestion
  dynamic?: boolean | 'true' | 'false' | 'strict';
```

src/plugins/backend_compatibility/server/transport/adapters/search_adapter.ts
```diff
@@ -0,0 +1,230 @@
...
+    const qs = { ...v };
+    if (qs.default_field === '*' && !qs.all_fields) {
+      qs.all_fields = true;
+      delete qs.default_field;
+    }
+    return qs;
+  },
+};
+
+function transformQuery(query: any): any {
```
huyaboo, line 205: Wouldn't a query be a query string or object?

src/plugins/backend_compatibility/server/transport/adapters/search_adapter.ts
```diff
@@ -0,0 +1,230 @@
...
+}
+
+function transformGeotileToGeohash(gt: Record<string, any>): Record<string, any> {
+  const result: Record<string, any> = { field: gt.field };
+  if (typeof gt.precision === 'number') {
+    const p = gt.precision;
+    if (p <= 4) result.precision = Math.max(1, Math.ceil(p / 2));
+    else if (p <= 10) result.precision = Math.min(5, 3 + Math.floor((p - 5) / 2));
+    else if (p <= 20) result.precision = Math.min(9, 6 + Math.floor((p - 11) / 3));
+    else result.precision = Math.min(12, 10 + Math.floor((p - 21) / 3));
```
huyaboo, line 164: A brief comment over why the precision is computed the way it is would be helpful

src/plugins/backend_compatibility/server/transport/compatibility_transport.ts
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    const normalized = this.normalizeParams(params);
+    const route = this.matchRoute(normalized);
+
+    const translatedParams = route?.request ? route.request(normalized, this.backend) : normalized;
+
+    const response = await super.request(translatedParams, options);
+
+    return route?.response ? route.response(response, this.backend) : response;
```
huyaboo: For future extensibility and readability, consider extracting this to `applyES6Response`

src/plugins/backend_compatibility/server/transport/compatibility_transport.ts
```diff
@@ -0,0 +1,328 @@
...
+    if (this.detecting) {
+      await this.detecting;
+      return;
+    }
+    this.detecting = this.performDetection(options);
+    await this.detecting;
+    this.detecting = null;
+  }
+
+  private async performDetection(options?: any): Promise<void> {
```
huyaboo, line 233: Do we need a backoff strategy to prevent throttling?

huyaboo (commented):
Overall looks good; just have style comments and a couple questions

src/plugins/backend_compatibility/server/transport/compatibility_transport.ts
```diff
@@ -0,0 +1,328 @@
...
+        );
+      });
+
+      const response = await Promise.race([detectionPromise, timeoutPromise]);
+      const info = response?.body || response;
+      this.backend = detectBackend(info);
+      CompatibilityTransport.lastDetectedBackend = this.backend;
+    } catch (error) {
+      if (this.detectionAttempts >= MAX_DETECTION_ATTEMPTS) {
+        this.backend = {
```
ruanyl, line 257: It silently fallback to OpenSearch `0.0.0`, can we log a warn here?

src/plugins/backend_compatibility/server/transport/compatibility_transport.ts
```diff
@@ -0,0 +1,328 @@
...
+    this.detecting = null;
+  }
+
+  private async performDetection(options?: any): Promise<void> {
+    this.detectionAttempts++;
+    try {
+      const detectionPromise = super.request({ method: 'GET', path: '/' }, options);
+
+      const timeoutPromise = new Promise<never>((_, reject) => {
+        setTimeout(
```
ruanyl: Consider to clear the timeout after Promise.race, otherwise the timer remains if I understood correctly.

ruanyl (dismissed):

src/plugins/backend_compatibility/README.md
eemijuntunen: This is entirely new logic added by this plugin. OSD currently has no compatibility layer for legacy backend. It only speaks the OpenSearch API. The detection, translation, and normalization are all new code introduced here.

src/plugins/backend_compatibility/README.md
eemijuntunen: Good point. Since this targets frozen legacy versions (ES 6.x and 7.x), the route table is fixed. No new ES 6/7 APIs should be added. The only changes could be if we discover additional OSD code paths that hit untranslated endpoints, which would be additive entries to the existing table rather than structural changes.

src/plugins/backend_compatibility/server/plugin.ts
eemijuntunen: Thanks, you're right. The core plugin system already handles this. I'll remove the manual check.

src/plugins/backend_compatibility/server/transport/types.ts
eemijuntunen: Kept for logging convenience ```backend.version vs reconstructing ${major}.${minor}.${patch} everywhere```, but open to removing preferred.

src/plugins/backend_compatibility/server/transport/adapters/mapping_adapter.ts
eemijuntunen: Updated. Thanks!

src/plugins/backend_compatibility/server/transport/adapters/document_adapter.ts
eemijuntunen: Added. Thanks!

src/plugins/backend_compatibility/server/transport/adapters/document_adapter.ts
eemijuntunen: Will check on this.

src/plugins/backend_compatibility/server/transport/adapters/document_adapter.ts
eemijuntunen: Updated. Thanks

src/plugins/backend_compatibility/server/transport/adapters/document_adapter.ts
eemijuntunen: Put into function. Thanks!

src/plugins/backend_compatibility/server/transport/adapters/mapping_adapter.ts
eemijuntunen: Updated. Thanks

src/plugins/backend_compatibility/server/transport/adapters/search_adapter.ts
eemijuntunen: Will check this. Thanks!

src/plugins/backend_compatibility/server/transport/adapters/search_adapter.ts
eemijuntunen: Added. Thanks!

src/plugins/backend_compatibility/server/transport/compatibility_transport.ts
eemijuntunen: Extracted. Thanks!

src/plugins/backend_compatibility/server/transport/compatibility_transport.ts
eemijuntunen: Added a backoff. Thanks!

src/plugins/backend_compatibility/server/transport/compatibility_transport.ts
eemijuntunen: Added message. Thanks!

src/plugins/backend_compatibility/server/transport/compatibility_transport.ts
eemijuntunen: Updated. Thanks

eemijuntunen:
> ## Kiro Review
> **Overall:** Promising architecture — intercepting at the Transport layer is the right abstraction level. However, a few issues need addressing before this leaves draft.
> 
> ### Critical
> **1. Race condition in backend detection**
> 
> Multiple concurrent requests hitting the transport before detection completes will each see `this.backend === null` and potentially enter `performDetection()` simultaneously. The `detecting` promise should be a proper singleton:
> 
> ```ts
> // Should be:
> if (!this.detecting) { this.detecting = this.performDetection(); }
> await this.detecting;
> ```
> 
> **2. Silent fallback to OpenSearch on detection failure**
> 
> After 3 failed detection attempts, defaulting to `{ distribution: 'opensearch', version: '0.0.0' }` is dangerous. If the backend is actually ES 6.x but detection failed (network blip, auth issue), requests will be sent without translation — potentially corrupting data (missing `_type` fields, wrong mappings). This should either:
> 
> * Throw and prevent the client from operating (fail-closed)
> * Retry indefinitely with backoff
> * At minimum, log at ERROR level, not just warn
> 
> **3. No tests in the commit**
> 
> The PR description references unit tests but the automated reviewer flagged "No relevant tests." Tests should be included in the PR for the translation logic to be reviewable.
> 
> ### Design Concerns
> **4. `lastDetectedBackend` should be private** — currently leaks infrastructure topology (ES 6.x vs 7.x vs OpenSearch) to any code that imports the class.
> 
> **5. Scope gap needs explicit documentation** — the legacy `elasticsearch-js` client paths (index patterns field discovery, autocomplete, usage collection) are not covered. Users need a clear compatibility matrix of what works and what doesn't against ES 6.x.
> 
> **6. Experimental warning** — if `backendCompatibility.enabled: true` is set, the plugin should log a prominent startup warning that this is experimental.
> 
> ### Questions
> * How are conflicts handled in the declarative route table? First-match or most-specific?
> * What happens with bulk requests, scroll contexts, and aliases in ES 6.x mode?
> * Is there a plan to add integration tests against actual ES 6.8/7.10 Docker containers in CI?
> 
> Architecture is solid. Once tests are added and the race condition + fallback behavior are fixed, this will be in good shape for a full review.

@angle943  Couple answers to the questions brought up by the review.

###  Critical

   1. Race condition: Already uses the singleton pattern, but there was a bug where this.detecting wasn't cleared on failure. Fixed:
```
   this.detecting = this.performDetection(options);
   try {
     await this.detecting;
   } finally {
     this.detecting = null;
   }
```
   2. Silent fallback: Added error logging and exponential backoff (500ms → 1s → 2s):
```
   console.error(
     `[backend_compatibility] Failed to detect backend after ${MAX_DETECTION_ATTEMPTS} attempts. ` +
       `Falling back to OpenSearch pass-through. Last error: ${error}`
   );
```
   Kept fallback-to-OpenSearch rather than fail-closed. If the backend is OpenSearch and just slow, fail-closed would break OSD startup entirely.

### Design Concerns

   4. lastDetectedBackend: Can make private with a getter in a follow-up.

   5. Scope gap: Will add a compatibility matrix to the README.

   6. Added Experimental warning:
```
   this.logger.warn(
     `[EXPERIMENTAL] ${PLUGIN_NAME} is enabled. Legacy backend compatibility is experimental and may change.`
   );
```
### Questions

   - Route table: First-match. ```/_search/scroll``` matches before ```/_search``` etc.
   - Bulk/scroll/aliases: Bulk is fully translated ```_type injected, if_seq_no stripped```. Scroll works via scroll adapter. Aliases pass through unchanged.
   - Integration tests in CI: Exist locally (Docker matrix: ES 6.8.23, 7.0.0, 7.10.2, OS 1.3.18, 2.19.0). Can add as a CI workflow.
```

ruanyl (dismissed):

ZilongX (dismissed):

yujin-emma (approved):
Thanks

ZilongX (approved):

sumukhswamy (approved):

PR #11975, opened 2026-05-15
feat(vis): split chart with custom legend

ruanyl:
### Description
This PR introduces two major improvements to Explore visualizations:
1. Generalized split chart as a universal chart-level setting — Instead of each chart type implementing its own split logic, a common `splitField/splitLayout/showSplitLabel` configuration on `ChartConfig` works across all chart types. When configured, data is grouped by the split field and rendered as independent chart instances in a responsive grid container (`SplitContainer`). This removes ~1300 lines of duplicated per-chart split handling and replaces it with ~400 lines of shared infrastructure.
2. Replace `ECharts` built-in legend with a custom legend component — A new CustomLegend component renders outside the ECharts canvas, enabling shared legend state across split chart instances. 

#### Split any charts demo

https://github.com/user-attachments/assets/022a5fcb-67a8-4d1f-94e7-31d0ebab0214

https://github.com/user-attachments/assets/578cdc1e-bb6f-4034-bccf-54598724bb77

#### Scrollable legend

https://github.com/user-attachments/assets/202d33af-6059-4a53-a34d-2cf4fec17870

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/components/visualizations/custom_legend.tsx
```diff
@@ -0,0 +1,102 @@
...
+      });
+    return sorted;
+  }, [legendMap]);
+  const [selected, setSelected] = useState<Record<string, boolean>>({});
+  const isVertical = position === Positions.LEFT || position === Positions.RIGHT;
+
+  const handleToggle = useCallback(
+    (name: string) => {
+      setSelected((prev) => {
+        const next = { ...prev, [name]: prev[name] === undefined ? false : !prev[name] };
```
yubonluo: Could you add some comments to explain why
```
  - undefined → false                                                                                                                                                                     
  - false → true                                                                                                                                                                           
  - true → false
```

ruanyl: Thanks, added @yubonluo

src/plugins/explore/public/components/visualizations/custom_legend.tsx
```diff
@@ -0,0 +1,103 @@
...
+    Object.values(legendMap).forEach((map) => Object.assign(merged, map));
+    const sorted: ColorMap = {};
+    Object.keys(merged)
+      .sort()
+      .forEach((name) => {
+        sorted[name] = merged[name];
+      });
+    return sorted;
+  }, [legendMap]);
+  const [selected, setSelected] = useState<Record<string, boolean>>({});
```
yubonluo: The component maintains the local `selected` state, but only synchronizes unidirectionally to `legendSelected$` observable without subscribing to `legendSelected$`. If other split chart instances update the legend state through `legendSelected$`, this component will not respond

ruanyl: Good point, I think we don't even need local state of `selected`

ruanyl: Updated @yubonluo

SuZhou-Joe (approved):

yubonluo (approved):

PR #12006, opened 2026-05-20
[Variable] Dashboard panels reset when switching variables in VIEW mode

yubonluo:
### Description

After adding/deleting visualizations in a dashboard and saving, switching a variable value causes deleted visualizations to reappear and newly added visualizations to disappear. Refreshing the page restores the correct panel state.

#### Root Cause

`stateDefaults` is a closure variable computed once at page load time from `savedDashboard.panelsJSON` and never updated afterward. In VIEW mode, `toUrlState()` intentionally strips `panels` from the URL state (`_a` param) to reduce URL length. When variable value selection triggers a URL sync round-trip (state → URL → state), the custom `set()` callback in `create_dashboard_app_state.tsx` falls back to `stateDefaults.panels` because the URL state has no `panels` field:

```typescript
stateContainer.set({
  ...stateDefaults,  // panels from initial page load (stale after save)
  ...state,          // from URL — no panels in VIEW mode
});
```

This overwrites the current in-memory panels with the stale snapshot from page load, causing newly added panels to be removed and previously deleted panels to be restored.

The full trigger chain:
1. User adds/removes viz → saves dashboard → backend `panelsJSON` updated correctly → switches to VIEW mode
2. User selects a different variable option → `updateVariableValue()` → `variables$.next()` → container `updateInput({variables})` → `handleDashboardContainerChanges()` → `appState.transitions.setDashboard({variables})` → URL updated (without panels) → `history.listen` fires → `updateState()` → custom `set()` called with URL state (no panels) → `stateDefaults.panels` (stale) overwrites current panels → `refreshDashboardContainer()` applies stale panels to container → panels mismatch → `maybeUpdateChildren()` removes/adds wrong panels

#### Solution

In the URL sync `set()` callback, when URL state does not include `panels` (VIEW mode), preserve the current in-memory panels instead of falling back to the stale `stateDefaults`:

```typescript
stateContainer.set({
  ...stateDefaults,
  ...state,
  panels: state.panels ?? stateContainer.getState().panels,
});
```

This is safe because:
- In VIEW mode, panels are never stored in the URL, so all panel state changes go through direct `setDashboard()` calls (save, revert, add/remove panel) which update `stateContainer` correctly
- On fresh page load, `stateContainer.getState().panels` equals `stateDefaults.panels` (just initialized)
- When switching dashboards, the entire state sync is torn down and recreated with fresh `stateDefaults`
- Discard changes flow writes panels explicitly via `setDashboard({panels: dashboard.panels})`, not through the URL sync callback

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/dashboard/public/application/utils/create_dashboard_app_state.test.tsx
+++ b/src/plugins/dashboard/public/application/utils/create_dashboard_app_state.test.tsx
@@ -200,3 +200,188 @@ describe('updateStateUrl', () => {
     expect(replaceSpy).toHaveBeenCalledWith({ ...location, state: someState });
   });
 });
+
+describe('panels preservation logic in URL sync', () => {
+  /**
+   * This test suite verifies the fix for the bug where panels were being reset
+   * when switching variables in VIEW mode.
+   *
+   * The bug occurred because:
+   * 1. In VIEW mode, toUrlState() excludes panels from URL
+   * 2. When syncing back from URL, stateDefaults (with initial panels) was merged
+   * 3. This caused panels to revert to their initial state
+   *
+   * The fix: `panels: state.panels ?? stateContainer.getState().panels`
+   * This preserves current panels when URL state doesn't include them.
+   */
+
+  // Helper to simulate the set function logic from createDashboardGlobalAndAppState
+  const simulateSetFunction = (state: any, stateDefaults: any, currentPanels: any[]) => {
+    if (!state) {
+      return null; // Don't set anything for null state
+    }
+
+    // This simulates the logic in create_dashboard_app_state.tsx lines 105-109
+    return {
+      ...stateDefaults,
+      ...state,
+      panels: state.panels ?? currentPanels, // The bug fix
+    };
+  };
+
+  test('should preserve current panels when URL state has no panels field (VIEW mode)', () => {
+    const initialPanels = [{ panelIndex: 'old-panel', gridData: { x: 0, y: 0, w: 10, h: 10 } }];
+    const currentPanels = [{ panelIndex: 'new-panel', gridData: { x: 0, y: 0, w: 10, h: 10 } }];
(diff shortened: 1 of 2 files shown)


src/plugins/dashboard/public/application/utils/create_dashboard_container.tsx
```diff
@@ -413,6 +413,15 @@ const getChangesForContainerStateFromAppState = (
...
     if (key === 'filters') return;
+    // Skip panels comparison here - panels sync is handled by handleDashboardContainerChanges
+    // This prevents a circular update loop where:
+    // 1. Variable changes -> updateInput({ variables })
+    // 2. Triggers handleDashboardContainerChanges -> syncs panels to appState
+    // 3. Triggers refreshDashboardContainer -> would overwrite panels back from appState
+    // By skipping panels here, we ensure container remains the source of truth for panels
+    // Panels flow: Container -> appState (via handleDashboardContainerChanges) -> SavedObject (on save)
+    // Panels never flow back from appState to Container (except during initial load)
+    if (key === 'panels') return;
```
SuZhou-Joe: Nit: possible to add a UT? Since we have a large comment for this tiny change.

SuZhou-Joe (dismissed):

src/plugins/dashboard/public/application/utils/create_dashboard_container.tsx
ruanyl: This introduced a regression: "Discard Changes" will break.
When you edit a dashboard, added a visualization, you turn off edit mode and select "Discard changes"

ruanyl (changes requested):
@yubonluo see my comment, there is likely a regression.

ruanyl (approved):

Qxisylolo (approved):

PR #12014, opened 2026-05-20
Align input handling across saved-objects import and data_source routes

sumukhswamy:
Description
  

  
  ## Changes
  
  **1. Reject `config` type in saved-objects import**
  
  The import pipeline accepts any type with `importableAndExportable: true`, which today includes `config`. That type holds global Advanced Settings (`defaultRoute`, banners, theme, etc.) and is meant to be written via the Advanced Settings UI / `PUT /api/opensearch-dashboards/settings`, not bulk import. Reject it in both `importSavedObjectsFromStream` and
  `resolveSavedObjectsImportErrors` with `unsupported_type`, mirroring the existing guard for `data-source`.
  
  **2. Run NDJSON import lines through the existing `validateObject` helper**
  
  The body-path import already runs each parsed object through the platform `validateObject` helper. `createSavedObjectsStreamFromNdJson` did not, so the two entry points behaved differently for the same input. Apply the same helper to each NDJSON line; rejected input returns HTTP 400 before any persistence.
  
  **3. Validate endpoint URL in `fetchDataSourceMetaData`**
  
  `test_connection` validates `dataSourceAttr.endpoint` via `isValidURL()` against `endpointDeniedIPs` / `endpointAllowlistedSuffixes`. `fetchDataSourceMetaData` accepts the same shape but skipped the check. Plumb `logger`, `endpointDeniedIPs`, and `endpointAllowlistedSuffixes` into the route, call `isValidURL()` before `getDataSourceClient()`, and return HTTP
  400 with the validator's `userMessage` (or a generic fallback) on failure. Defaults for `endpointDeniedIPs` (loopback and link-local CIDR ranges) are applied at the plugin layer and shared between both routes.


src/core/server/saved_objects/routes/utils.ts
```diff
@@ -38,14 +38,23 @@ import {
...
 export async function createSavedObjectsStreamFromNdJson(ndJsonStream: Readable) {
   const savedObjects = await createPromiseFromStreams([
     ndJsonStream,
     createSplitStream('\n'),
     createMapStream((str: string) => {
       if (str && str.trim() !== '') {
-        return JSON.parse(str);
+        const parsed = JSON.parse(str);
+        try {
+          validateObject(parsed);
```
LDrago27, line 51: nit: Maybe we can enhance the `validateObject` to include this validation too       
`if (str && str.trim() !== '') {`
So the code would be more concise and just have the validateObject

LDrago27 (approved):

TackAdam (approved):

Issue #11882, opened 2026-04-30
New Engine Type Compatibility — Plugin Guidance Proposal


seraphjiang:
# New Engine Type Compatibility — Plugin Guidance Proposal

**Date:** 2026-04-29
**Status:** Draft
---

## Problem Statement

When adding a new OpenSearch-compatible engine that supports a subset of OpenSearch DSL (~70% at GA), the existing plugin version decoupling system handles **version** and **required backend plugin** compatibility, but has **no mechanism for engine type compatibility**. When a data source backed by the new engine is connected, plugins that rely on unsupported DSL features will fail with cryptic errors or return incorrect results.

The core question: **How should plugin teams declare and enforce compatibility with a new engine type, given that the current manifest system has no `supportedOSDataSourceEngineTypes` field?**

---

### What Exists Today

The version decoupling system provides two manifest-level declarations in `opensearch_dashboards.json`:

```json
{
  "supportedOSDataSourceVersions": ">=2.13.0",
  "requiredOSDataSourcePlugins": ["opensearch-alerting"]
}
```

And one imperative mechanism — the `dataSourceFilter` callback prop on `DataSourceMenu`:

```javascript
dataSourceFilterFn = (dataSource) => {
  const dataSourceVersion = dataSource?.attributes?.dataSourceVersion || "";
  const installedPlugins = dataSource?.attributes?.installedPlugins || [];
  return (
    semver.satisfies(dataSourceVersion, pluginManifest.supportedOSDataSourceVersions) &&
    pluginManifest.requiredOSDataSourcePlugins.every((plugin) =>
      installedPlugins.includes(plugin)
    )
  );
};
```

This pattern is the reference implementation for all plugin teams.

### The Engine Type Gap

The `DataSourceEngineType` enum is defined in `src/plugins/data_source/common/data_sources/types.ts`:

```typescript
export enum DataSourceEngineType {
  OpenSearch = 'OpenSearch',
  OpenSearchServerless = 'OpenSearch Serverless',
  Elasticsearch = 'Elasticsearch',
  NA = 'No Engine Type Available',
  OpenSearchCrossCluster = 'OpenSearch(Cross-cluster search)',
}
```

Engine type detection happens in `DataSourceConnectionValidator.fetchDataSourceInfo()` by calling `GET /` on the cluster and checking `response.version.distribution`:
- `"opensearch"` → `DataSourceEngineType.OpenSearch`
[...]

PR #12023, opened 2026-05-21
feat(data-source): detect AnalyticEngine, manifest-driven plugin opt-out

seraphjiang:
### Description

Implements **Option D** of RFC [#11882](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11882) (*New Engine Type Compatibility — Plugin Guidance Proposal*): an opt-out mechanism that lets plugins declare engine types they don't support via a new `unsupportedOSDataSourceEngineTypes` manifest field, paired with detection of the `AnalyticEngine` engine type at data-source registration time.
analytic-engine-demo

#### What it does

1. **New engine type detection.** `DataSourceConnectionValidator.fetchDataSourceInfo()` now also reads `cluster.pluggable.dataformat` from `_cluster/settings`. When an OpenSearch domain reports `dataformat=composite` and `dataformat.enabled=true`, the data source is tagged `DataSourceEngineType.AnalyticEngine` instead of `OpenSearch`. Detection failure (permission, network, older cluster) falls back to `OpenSearch` — fail-open.

2. **New manifest field: `unsupportedOSDataSourceEngineTypes`.** Added to `PluginManifest` in `src/core/server/plugins/types.ts` and parsed in `plugin_manifest_parser.ts`. Plugins declare engine types they don't support; default (omit) = supports all.

3. **Opt-in filter on `IndexPatternsService.getCache()`.** New optional `{ excludeEngineTypes }` parameter. Discover (legacy) reads its own manifest and the `data_explorer` sidebar (which hosts legacy Discover) filters picker options by the same list. Other consumers (Explore at `/app/explore/*`, Trace, Metric, Visualize, Map) pass nothing and see all engines.

4. **Discover declares the block:** `opensearch_dashboards.json` adds `"unsupportedOSDataSourceEngineTypes": ["AnalyticEngine"]`.

5. **Remote-cluster compatibility:** `index_type.ts` accepts `AnalyticEngine` for remote-cluster fetch — these domains are OpenSearch underneath.

#### Why opt-out (Option D) and not opt-in (Option A/C)

The RFC originally weighed three options and recommended a hybrid manifest + imperative override (Option C). Discussion during implementation highlighted that the *positive* form (`supportedOSDataSourceEngineTypes`) forces every plugin to enumerate every engine — and update on every new engine addition. Most plugins are mostly compatible with new engine types; only a handful have specific incompatibilities (TSVB pipeline aggs, Significant Terms aggs, alerting's `_monitors/_execute`).

Option D takes the negative form: only plugins that *break* against a specific engine declare it. The trade-offs are documented in the RFC update (commit on issue #11882 adds Option D as a parallel choice). Option D doesn't replace Option C — the imperative `dataSourceFilter` callback remains available for partial-compatibility cases.

### Issues Resolved

Refs #11882 (RFC implementation, Option D)

### Screenshot / video

A short screencast of the end-to-end behavior: legacy Discover excludes the AnalyticEngine-backed index pattern from its picker, then the global Data Sources management page shows ALL data sources (including the AnalyticEngine one tagged correctly).

#### Automated

- 4 new unit tests in `data_source_connection_validator.test.ts` covering AnalyticEngine detection, `dataformat.enabled=false` staying as OpenSearch, `getSettings` throwing, and Elasticsearch domains not promoted.
- 5 new unit tests in `index_patterns.test.ts` covering `excludeEngineTypes` (omitted, empty, blocked, no reference, missing data source).
- 3 new unit tests in `plugin_manifest_parser.test.ts` covering the new field's parsing (array passes through, missing → `[]`, non-array → `[]`).
- All 9 existing parser tests updated to include the new optional field in expected manifest shape.
- 62 tests pass across the 3 affected files.

#### Manual

1. Boot a local OpenSearch with `yarn opensearch snapshot`, then `yarn start`.
2. Register a data source pointed at an OpenSearch domain that has `cluster.pluggable.dataformat=composite` and `dataformat.enabled=true` set as a persistent setting. An AnalyticEngine-enabled playground is publicly reachable at `http://mustang.playground.nightly.opensearch.org` (the hostname is the public infra name; nothing in the code references it).

#### Reproduce via API (no UI needed)

Replace `/abc` with your dev basepath (the random suffix shown when `yarn start` boots).

```bash
# Create the data source — fetchDataSourceInfo runs server-side and auto-tags AnalyticEngine
curl -sS -X POST 'http://localhost:5601/abc/api/saved_objects/data-source' \
  -H 'osd-xsrf: true' \
  -H 'Content-Type: application/json' \
  -d '{
    "attributes": {
      "title": "AnalyticEngine Test",
      "endpoint": "http://mustang.playground.nightly.opensearch.org",
      "auth": { "type": "no_auth", "credentials": {} }
    }
  }'

# Verify the engine type tag
curl -sS 'http://localhost:5601/abc/api/saved_objects/_find?type=data-source' \
  -H 'osd-xsrf: true' | jq '.saved_objects[].attributes | {title, dataSourceEngineType}'
# expected: { "title": "AnalyticEngine Test", "dataSourceEngineType": "AnalyticEngine" }
```

3. Open legacy Discover at `/app/data-explorer/discover`. The data source / index pattern picker excludes any index pattern whose backing data source is AnalyticEngine.
4. Open the global Data Sources management page (`/app/management/opensearch-dashboards/dataSources`). All data sources are visible, including the AnalyticEngine one — the filtering is plugin-scoped, not global.

To toggle and re-test:

```bash
# Force back to OpenSearch (AnalyticEngine-tagged data sources reappear in Discover)
curl -sS -X PUT 'http://localhost:5601/abc/api/saved_objects/data-source/<id>' \
  -H 'osd-xsrf: true' -H 'Content-Type: application/json' \
  -d '{ "attributes": { "dataSourceEngineType": "OpenSearch" } }'
```

### Backport notes

Backport-safe by design:
- No SO migration; `dataSourceEngineType` is set on creation as before, no migration of existing rows.
- The new `PluginManifest` field is optional. Older OSD versions reading manifests with the new field simply ignore unknown keys.
- The new `DataSourceEngineType.AnalyticEngine` enum value is purely additive. Existing equality/switch checks against `OpenSearch` won't match `AnalyticEngine` — the conservative default.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest` (focused: 62 tests across the 3 affected files)
  - [ ] `yarn test:jest_integration` (not run locally — please verify in CI)
- [ ] New functionality has been documented. (RFC #11882 has been updated to include Option D and a Setup section with the API/curl example above. Once approved, the manifest field could land in plugin development docs.)
- [x] Commits are signed per the DCO using `--signoff`.


seraphjiang:
The Build & test failure flagged by the bot above is **not related to this PR** — it's a pre-existing broken test on `main`.

**Actual failing test (corrected from the bot's summary, which conflated parallel-runner names):**

```
HelloWorldContainer in edit mode hides disabledActions
src/plugins/embeddable/public/lib/panel/embeddable_panel.test.tsx:277
```

**Why it's not from this PR:**

1. **The failing file is not in this PR's diff.** The 12 files changed by #12023 are entirely under `src/core/server/plugins/`, `src/plugins/data/`, `src/plugins/data_explorer/`, `src/plugins/data_source/`, and `src/plugins/discover/`.

2. **The test fails on clean upstream `main`.** Verified locally by checking out `origin/main` (commit `fe2aecea8a`) in a separate worktree with this PR's changes entirely absent — same failure:
   ```
   FAIL  src/plugins/embeddable/public/lib/panel/embeddable_panel.test.tsx
     ✕ HelloWorldContainer in edit mode hides disabledActions (50 ms)
       expect(received).toBe(expected) // Object.is equality
       Expected: 1
       Received: 0
   ```

3. **Failing across multiple SHAs of this branch.** Same test has now failed on `0d9d86d8`, `ffb5bff2`, `ab0fcf02b2`, and `a6eeeea727` — different SHAs, same pre-existing failure.

4. The test is rendering an embeddable panel via Enzyme `mount()` and expecting `embeddablePanelAction-FOO` items to render synchronously. Likely a regression from the React 16→18 upgrade ([commit 3f7b577bb8](https://github.com/opensearch-project/OpenSearch-Dashboards/commit/3f7b577bb8)) — the test file also emits `[React Intl] Could not find required intl object` warnings, suggesting the test setup wasn't fully migrated. May be worth filing a separate tracking issue.

5. Locally on this PR's branch, the focused test suite passes:
   - `data_source_connection_validator.test.ts` — 14 tests
   - `index_patterns.test.ts` — 17 tests
   - `plugin_manifest_parser.test.ts` — 32 tests

   Total: 63/63 passing. Typecheck and lint clean.

In short: this PR's diff doesn't touch the embeddable plugin, and the failing test fails identically on `main`. Suggesting reviewers ignore this specific failure for #12023's review and consider filing a tracking issue for `embeddable_panel.test.tsx` separately.

src/plugins/data/common/index_patterns/index_patterns/index_patterns.ts
```diff
@@ -229,13 +229,65 @@ export class IndexPatternsService {
     }
   };
 
-  getCache = async () => {
+  getCache = async (options?: { excludeEngineTypes?: readonly string[] }) => {
```
huyaboo, line 232: Nit: we can leave options as optional but let `excludeEngineTypes` not-optional


```suggestion
  getCache = async (options?: { excludeEngineTypes: readonly string[] }) => {
```

seraphjiang:
**Update:** filed a separate fix PR for the embeddable panel test flake — [#12035](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/12035).

The flake was an earlier test in `embeddable_panel.test.tsx` calling `mount()` synchronously without `await act()`, leaking React 18 scheduler state into the next test. Verified locally: 15/15 consecutive runs pass with the fix vs ~30–60% before.

Once #12035 merges, the Merge Gate failure on this PR should resolve naturally on the next push (or a maintainer can merge upstream main back into this branch).

src/plugins/data/common/index_patterns/index_patterns/index_patterns.ts
```diff
@@ -229,13 +229,65 @@ export class IndexPatternsService {
...
+  // field (see core PluginManifest). Other consumers omit options and see everything.
+  private async applyEngineTypeFilter<T>(
+    savedObjects: Array<SavedObject<T>>,
+    excludeEngineTypes: readonly string[]
+  ): Promise<Array<SavedObject<T>>> {
+    const dataSourceIds = Array.from(
+      new Set(
+        savedObjects
+          // @ts-expect-error TS2339 references not typed on SavedObject<T> generic
+          .map((obj) => (obj.references || []).find((ref) => ref.type === 'data-source')?.id)
```
LDrago27: In some of the indexpattern object we save the datasourceID encoded in the id itself with a `::` operator.
We should probably add that logic in here or better have a util that does so we can use across multiple places. 

For reference: 

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/854efd5f6816b8ff6cbc36ea82440718b4c9dac6/src/plugins/data/public/query/query_string/dataset_service/lib/index_pattern_type.ts#L120

huyaboo (dismissed):

LDrago27:
@seraphjiang Also should we include the osEnginebased Filtering for Datasets too. For reference this is page i am refering to 

image

src/plugins/data/common/index_patterns/index_patterns/index_patterns.ts
huyaboo: While that may be true, isn't the `references` field the source of truth for `data-source` relationships?

seraphjiang: Good catch — addressed in the latest push. Two changes:

1. Extracted `getDataSourceIdFromIndexPattern` to `src/plugins/data/common/index_patterns/utils.ts` (where similar utilities like `getDataSourceReference` already live), with 6 unit tests covering both encodings, undefined references, the prefer-references-over-id precedence, and the no-data-source case.
2. Used the new util inside `applyEngineTypeFilter` so it correctly handles the namespaced-id format you flagged.

I'm intentionally **not** updating the existing call sites in `index_pattern_type.ts` (your linked file) in this PR — that file lives in the dataset_service / Explore path and isn't in this PR's filter scope. De-duplicating those inline blocks would expand the review surface to a different plugin's CODEOWNERS and risk introducing subtle regressions in new Explore. Suggesting a follow-up PR scoped to dataset_service to migrate to the new util.

Let me know if you'd rather see the consolidation in this PR — happy to add it back if the scope expansion is acceptable.

seraphjiang: Update on the previous reply: caught one more spot. The sidebar in `data_explorer/public/components/sidebar/index.tsx` (also part of this PR's diff) had the same references-only resolution bug. Now also using `getDataSourceIdFromIndexPattern` there.

Final state of the util usage in this PR:

| Site | In PR diff? | Uses util? |
|---|---|---|
| `index_patterns.ts:applyEngineTypeFilter` | yes | ✅ |
| `data_explorer/sidebar/index.tsx` | yes | ✅ |
| `index_pattern_type.ts` (dataset_service) | no | unchanged — proposed as follow-up |

So both filter sites added/modified by this PR now correctly resolve data source ids from both the modern `references[]` and the legacy `<dsId>::<patternId>` namespaced-id format. The pre-existing duplicated logic in `index_pattern_type.ts` is left alone to keep this PR's scope tight; happy to migrate it in a follow-up if you'd like.

seraphjiang:
> @seraphjiang Also should we include the osEnginebased Filtering for Datasets too. For reference this is page i am refering to
> 
> image

DataSet is intended out of scope, as it is used in New Discover, and New discover should support Analytic Engine

joshuali925:
### Code review

Found 1 issue:

1. Cross-plugin internal path import: `data_explorer` reaches into `discover`'s directory to import its manifest JSON directly (CLAUDE.md says "Only import top-level module APIs, never reach into internal paths."). Consider exposing the unsupported engine types via Discover's `public/index.ts` (or via a core registration mechanism) so consumers don't depend on Discover's manifest path.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a4346b09bc012bf09ccc67a98dde2aba9388e7aa/src/plugins/data_explorer/public/components/sidebar/index.tsx#L18-L22

joshuali925 (approved):

LDrago27 (approved):

PR #12028, opened 2026-05-21
fix: unnecessary re-render which causes performance issue

ruanyl:
### Description

Fixes chart re-rendering oscillation when switching chart types (e.g., line → bar) for queries with multiple numerical columns in the same axis type (e.g., `| stats AVG(bytes) as avg_bytes, MAX(bytes) as max_bytes by span(timestamp, 1h)`). The chart would flash continuously because the axes mapping alternated between two states (y/color fields swapping).

Root cause: Two issues combined after #11975 moved AxesSelectPanel to persist across chart type changes:

1. Format mismatch in `setAxesMapping: getAxesMappingByRule` returns string values (`"max_bytes"`) while `convertMappingsToStrings` returns arrays (`["max_bytes"]`). `isEqual("str", ["str"])` is false, causing spurious `visConfig$ ` re-emissions.
2. Stale effect trigger in `AxesSelectPanel`: The `useEffect` that syncs selections to the builder had `chartType` and `remainingMappings` in its dependency array, causing it to fire on chart type changes with a stale `currentSelections` whose key insertion order differed — feeding columns to `getAxesMappingByRule` in a different order and producing a swapped mapping.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


d-buckner (dismissed):
Nice catch. Seems like a unit test would be valuable to protect against against a future regression since this seems pretty impactful to the user experience.

yubonluo (dismissed):

yubonluo (approved):

Qxisylolo (approved):

PR #12041, opened 2026-05-22
Pin GitHub Actions to commit SHAs

Divyaasm:
### Description

Pin all GitHub Action tag references to their corresponding commit SHAs.

Tags are mutable references that can be force-pushed to point to different commits, making them vulnerable to supply chain attacks. Commit SHAs are immutable and guarantee that the exact reviewed code is executed in CI workflows. This change pins all third-party actions to their current commit SHAs to prevent potential tampering.


Hailong-am:
For actions start with `action/*` are GitHub-owned, do we still need to pin a specific id?

SuZhou-Joe (approved):

ruanyl (approved):

PR #12044, opened 2026-05-22
[Workspace] Refresh collaborators page from server on mount

xluo-aws:
## Description
The Workspace **Collaborators** page renders the permission list from the cached `workspaces.currentWorkspace$` BehaviorSubject. If a different user (e.g. an admin) revokes someone's access remotely, the affected user's UI keeps showing the old collaborator list until the page is hard reloaded.

This PR mirrors that pattern for the Collaborators page: on mount (and whenever the workspace id changes, plus after a successful permission update), call `workspaceClient.get(currentWorkspace.id)` and render the rendered permission list from the server response. The cached `currentWorkspace$.permissions` is used only as a fallback if the request fails.

## Testing the changes
Existing snapshot test for `WorkspaceCollaborators` retained. Added a new test that asserts:

- `workspaceClient.get(workspaceId)` is called on mount.
- The row for a collaborator that exists in the cached `currentWorkspace$` but **not** in the server response is no longer rendered.

Existing tests updated to provide `workspaceClient.get` in their mock services.

## Check List
- [x] All tests pass
  - [x] `yarn test:jest` (only the changed test file was run locally; full suite via CI)
- [x] New functionality has been documented in the code (inline comment explaining the staleness rationale)
- [ ] Update [CHANGELOG.md](./CHANGELOG.md) — will add a fragment under `changelogs/fragments/` once a PR number is assigned

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/src/plugins/workspace/public/components/workspace_collaborators/workspace_collaborators.tsx
+++ b/src/plugins/workspace/public/components/workspace_collaborators/workspace_collaborators.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { useCallback, useEffect, useState } from 'react';
+import { useEffect, useState } from 'react';
 import { EuiPage, EuiPanel, EuiSpacer } from '@elastic/eui';
 import { i18n } from '@osd/i18n';
 
@@ -73,23 +73,25 @@ export const WorkspaceCollaborators = () => {
 
   const isPermissionEnabled = application?.capabilities.workspaces.permissionEnabled;
 
-  const refreshCollaborators = useCallback(async () => {
-    if (!currentWorkspace?.id || !workspaceClient?.get) {
+  useEffect(() => {
+    if (!currentWorkspace?.id) {
       return;
     }
-    try {
-      const response = await workspaceClient.get(currentWorkspace.id);
-      if (response.success) {
+    let cancelled = false;
+    workspaceClient
+      .get(currentWorkspace.id)
+      .then((response) => {
+        if (cancelled || !response.success) {
+          return;
+        }
         setLatestPermissions(response.result.permissions ?? {});
-      }
-    } catch {
-      // Fall back to the cached permissions from `currentWorkspace$`.
-    }
+      })
+      // eslint-disable-next-line no-console
+      .catch((error) => console.warn('Failed to refresh workspace collaborators', error));
+    return () => {
+      cancelled = true;
+    };
   }, [currentWorkspace?.id, workspaceClient]);
-
-  useEffect(() => {
-    refreshCollaborators();
-  }, [refreshCollaborators]);
 
   const handleSubmitPermissionSettings = async (settings: WorkspacePermissionSetting[]) => {
     const showErrorNotification = (errorText?: string) => {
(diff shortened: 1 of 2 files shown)


xluo-aws:
Pushed `c893657c58` addressing the review feedback:

1. **Race condition on unmount** — accepted. Folded into a single `useEffect` with a `cancelled` flag so an in-flight `workspaceClient.get` can't overwrite a newly-mounted workspace's permissions if the user switches workspaces mid-fetch.
2. **Loading state** — declined. The same fetch-after-render pattern is used by `WorkspaceDetailApp` for data sources (no spinner), and the cached `currentWorkspace$` is used as the initial paint so the common case (cache == server) shows no flicker. Adding a spinner that flashes for a few hundred ms on every mount felt worse than the current UX.
3. **Error visibility** — accepted. Replaced the empty catch with `console.warn` so unexpected fetch failures show up in devtools, while still silently falling back to the cached value (no toast spam).
4. **`workspaceClient?.get` guard** — accepted. Dropped the optional chain; in production `workspaceClient` is always the concrete `WorkspaceClient` and the test mock now provides `get` explicitly.
5. **Double fetch after mutation** — accepted. `WorkspaceClient.update`'s response is just `IResponse<boolean>` so it doesn't carry the new permissions, but we already have them locally — `setLatestPermissions(convertPermissionSettingsToPermissions(settings))` after a successful update avoids the extra round-trip and the small chance of a read-after-write inconsistency.

All 5 unit tests in the file still pass.

xluo-aws:
Verified end-to-end against a security-enabled cluster

**Setup:** local OpenSearch 3.6.0-SNAPSHOT with security plugin (demo install), local OSD with security-dashboards-plugin cloned in, separate `admin` and `bar` users, workspace with admin (write) + bar (read).

**Manual repro:**
1. `admin` opens Collaborators → sees admin + bar.
2. `bar` opens Collaborators in incognito → sees admin + bar.
3. `admin` revokes `bar`.
4. `bar` navigates within the workspace (no hard reload) and back to Collaborators.
5. Before this commit → bar still saw the cached list. After → bar is redirected to the fatal-error app.

SuZhou-Joe (approved):

src/plugins/workspace/public/components/workspace_collaborators/workspace_collaborators.tsx
```diff
@@ -53,12 +56,52 @@ export const WorkspaceCollaborators = () => {
...
 
   const isPermissionEnabled = application?.capabilities.workspaces.permissionEnabled;
 
+  useEffect(() => {
+    if (!currentWorkspace?.id || !workspaces) {
+      return;
+    }
+    let cancelled = false;
+    workspaceClient
+      .get(currentWorkspace.id)
```
SuZhou-Joe, line 82: Thanks for the change, approved. But I am thinking if it would be better for workspaceClient to refresh the cache and update the observable, since this `get` should have most up-to-date metadata.

xluo-aws:
Per offline discussion with @SuZhou-Joe, he will provide a general solution to better handle this type of issue. close this PR.

PR #12050, opened 2026-05-25
[Explore Vis]Fix gauge unit unable to display

Qxisylolo:
### Description

This pr fixes：
1. bar gauge unit display by accessing the .label property from the display() function return value in showDisplayValue

before:
截屏2026-05-25 15 44 03

3. Change overflow: hidden to overflow: visible on vertical bar gauge layout so value text is not clipped

4. fails to show scatter with size due to invalid visualMap field

before :
截屏2026-05-26 15 07 21

5. add tooltip toggle for bar gauge

截屏2026-05-26 15 07 45

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ruanyl (dismissed):

ruanyl (approved):

src/plugins/explore/public/components/visualizations/scatter/scatter_chart_utils.ts
```diff
@@ -330,7 +329,7 @@ export const createSizeScatterSeries = <T extends BaseChartStyle>({
   newState.visualMap = {
     show: styles.addLegend === true,
     type: 'continuous',
-    dimension: sizeDimension,
+    dimension: 2, // data type is [x, y, size] format
```
ruanyl, line 332: why change to hard coded?

Qxisylolo: Because in this case, the data is processed again into groups, sizeDimension refers to the dimension before the grouping transformation.

yubonluo (approved):

PR #12061, opened 2026-05-26
[Explore] Enable SQL queries in Explore Logs

ritvibhatt:
#### Summary
Adds experimental SQL language support to the Explore plugin, allowing users to query their data using OpenSearch SQL syntax alongside the existing PPL support. (Only works with analytics angine)

#### Core Features
- **SQL Language Toggle**: Added SQL as a language option in Explore's query editor alongside PPL
- **Date Picker Integration**: Modified date picker visibility logic to show date picker for SQL queries in Explore
(overriding the global `hideDatePicker: true` setting)
- **Feature Flag**: Added `explore.sqlSupport.enabled` configuration option to gate the SQL feature
- **Tab Support**: Enabled SQL language support in Logs, Visualization, and Statistics tabs for logs datasets

##### Query Editor Integration
- **Date Picker**: Added special case in `shouldRenderDatePicker()` to show date picker for SQL queries in Explore app
- **Tab Registration**: Updated tab definitions to include SQL in `supportedLanguages` for appropriate tabs
- **Feature Gating**: SQL options are filtered out when `explore.sqlSupport.enabled` is `false`. SQL visualizations aren't visible in the visualizations app and not able to be selected to add to dashboard. In a dashboard that has a SQL visualization, the visualization won't render with an error saying that it needs SQL and to turn on SQL support in Explore.

#### Files Modified
- `src/plugins/data/public/ui/query_editor/query_editor_top_row.tsx` - Date picker logic for SQL in Explore
- `src/plugins/explore/public/application/register_tabs.ts` - Tab language support
- `src/plugins/explore/public/components/query_panel/query_panel_widgets/language_toggle/language_toggle.tsx` - Feature flag
 enforcement
- `config/opensearch_dashboards.yml` - Added SQL support configuration
- `docs/plugins/explore/sql-support.md` - Documentation for the feature

#### Testing
- ✅ SQL language appears in Explore language dropdown when feature is enabled
- ✅ Date picker shows for SQL queries in Explore
- ✅ Feature flag properly gates SQL availability
- ✅ SQL support limited to Explore (doesn't affect Discover or other apps)
- ✅ Works across Logs, Visualization, and Statistics tabs

#### Configuration
To enable this feature, add to `opensearch_dashboards.yml`:
```yaml
explore.sqlSupport.enabled: true
```

Limitations

- Experimental feature
- Only available in Explore for logs datasets
- Will only work with analytics engine backend

## Screenshot

https://github.com/user-attachments/assets/0afaed79-f450-4e11-a0e5-1efda8ea16ee

Block SQL visualization in dashboard
Screenshot 2026-05-28 at 12 50 59 PM

## Testing the changes

1. Ensure OpenSearch is running and accessible with analytics engine
2. Add to `opensearch_dashboards.yml`:
     ```yaml
     explore.sqlSupport.enabled: true
3. Start OpenSearch Dashboards
4. Navigate to Explore with a logs dataset
5. Query with OpenSearch SQL

### Check List

- [x] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -294,7 +296,7 @@ export const executeQueries = createAsyncThunk<
 
   // Execute histogram query in background (non-blocking)
   if (needsHistogramQuery) {
-    const interval = state.legacy?.interval;
+    const interval = state.legacy?.interval || 'auto';
```
joshuali925: is this needed?

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -83,7 +83,11 @@ export const registerBuiltInTabs = (
             }),
       flavor: [ExploreFlavor.Logs, ExploreFlavor.Traces],
       order: 10,
-      supportedLanguages: [EXPLORE_DEFAULT_LANGUAGE],
+      // SQL is only supported for Logs flavor, not Traces
+      supportedLanguages:
+        registryFlavor === ExploreFlavor.Traces
+          ? [EXPLORE_DEFAULT_LANGUAGE]
+          : [EXPLORE_DEFAULT_LANGUAGE, 'SQL'],
```
joshuali925: could you make this more explicit, otherwise metrics also falls into the else branch. it won't break now since metrics doesn't have logs tab, but it's fragile

src/plugins/explore/public/application/register_tabs.ts
```diff
@@ -186,10 +190,13 @@ export const registerBuiltInTabs = (
...
     flavor: [ExploreFlavor.Logs, ExploreFlavor.Metrics, ExploreFlavor.Traces],
     order: 20,
+    // SQL only for Logs, PROMQL for Metrics, PPL only for Traces
     supportedLanguages:
       registryFlavor === ExploreFlavor.Metrics
         ? [EXPLORE_DEFAULT_LANGUAGE, 'PROMQL']
-        : [EXPLORE_DEFAULT_LANGUAGE],
+        : registryFlavor === ExploreFlavor.Traces
+        ? [EXPLORE_DEFAULT_LANGUAGE]
+        : [EXPLORE_DEFAULT_LANGUAGE, 'SQL'],
```
joshuali925: same here, if you want only logs to have SQL, do the other way around. we may add agent traces back to explore later, and it will break

src/plugins/explore/public/components/query_panel/query_panel_widgets/language_toggle/language_toggle.tsx
```diff
@@ -98,7 +178,7 @@ export const LanguageToggle = () => {
         anchorPosition="downCenter"
         panelPaddingSize="none"
       >
-        <EuiContextMenuPanel size="s" items={items} />
+        {isPopoverOpen && <EuiContextMenuPanel size="s" items={items} />}
```
joshuali925: why is this needed? i thought Eui handles it

src/plugins/explore/public/embeddable/explore_embeddable.tsx
```diff
@@ -440,6 +440,12 @@ export class ExploreEmbeddable
     if (needFetch) {
       this.prevState = { filters, query, timeRange };
       this.searchProps = searchProps;
+
+      // Apply dashboard filters to the SearchSource (same as PPL)
+      if (this.filtersSearchSource) {
+        this.filtersSearchSource.setField('filter', filters || []);
```
joshuali925, line 446: how are filters handled?

src/plugins/query_enhancements/public/search/sql_search_interceptor.ts
```diff
@@ -50,13 +60,63 @@ export class SQLSearchInterceptor extends SearchInterceptor {
...
+      // Insert after WHERE keyword, wrapping existing expression with AND
+      const whereToken = existingWhere.start!;
+      const insertPos = whereToken.stop! + 1;
+      return sql.slice(0, insertPos) + ` ${whereClause} AND` + sql.slice(insertPos);
+    }
+
+    // Insert WHERE after the relation (table name + alias)
+    const relation = fromClause.relation();
+    const insertPos = relation.stop!.stop! + 1;
+    return sql.slice(0, insertPos) + ` WHERE ${whereClause}` + sql.slice(insertPos);
```
joshuali925: do we support sub query?

src/plugins/query_enhancements/public/plugin.tsx
```diff
@@ -180,8 +181,14 @@ export class QueryEnhancementsPlugin
...
       showDocLinks: false,
       editor: createEditor(SingleLineInput, null, sqlControls, DefaultInput),
-      editorSupportedAppNames: ['discover'],
-      supportedAppNames: ['discover', 'data-explorer'],
+      editorSupportedAppNames: ['discover', 'explore', 'agentTraces'],
+      supportedAppNames: [
+        'discover',
+        'data-explorer',
+        'explore',
+        'dataset_management',
```
joshuali925: why dataset management?

joshuali925:
### Code review

Found 6 issues:

1. Hardcoded `appName === 'explore'` inside the generic `data` plugin's `shouldRenderDatePicker`. The data plugin is meant to be app-agnostic; the documented extension point is `languageOverrides[lang].hideDatePicker` on the dataset type. Either drop `hideDatePicker: true` from the SQL language config and rely on dataset-level overrides, or have Explore set `languageOverrides.SQL.hideDatePicker = false` on its dataset types.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/afb4db71165de92bfa252062396b7984d80667d7/src/plugins/data/public/ui/query_editor/query_editor_top_row.tsx#L307-L313

2. Redundant `{isPopoverOpen && ...}` guard on `EuiContextMenuPanel`. `EuiPopover` already gates render via `isOpen`; no other popover usage in this plugin wraps its body conditionally. Already flagged inline by a reviewer (\"why is this needed? i thought Eui handles it\").

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/afb4db71165de92bfa252062396b7984d80667d7/src/plugins/explore/public/components/query_panel/query_panel_widgets/language_toggle/language_toggle.tsx#L178-L184

3. `dataset_management` added to SQL's `supportedAppNames` with no stated motivation. PR scope is SQL in Explore Logs; this enables SQL in the dataset management UI as a side effect, untested. Already flagged inline by a reviewer (\"why dataset management?\").

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/afb4db71165de92bfa252062396b7984d80667d7/src/plugins/query_enhancements/public/plugin.tsx#L184-L192

4. Double non-null assertion `relation.stop!.stop! + 1` (CLAUDE.md: \"Avoid non-null assertions (`!.`)\"). For complex SQL (subqueries, parenthesized relations) `relation.stop` is not guaranteed; this will throw at runtime instead of falling back. The same pattern repeats in `sql_search_interceptor.ts:116`.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/afb4db71165de92bfa252062396b7984d80667d7/src/plugins/query_enhancements/public/search/filters/sql_filter_utils.ts#L34-L40

5. Fragile `else`-defaults-to-SQL ternaries in `register_tabs.ts`. Today only Logs/Traces/Metrics flavors exist, but a future flavor would silently inherit SQL via the trailing `: [EXPLORE_DEFAULT_LANGUAGE, 'SQL']` branch. Already flagged inline by a reviewer (\"if you want only logs to have SQL, do the other way around\"). Same shape at L85-L90.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/afb4db71165de92bfa252062396b7984d80667d7/src/plugins/explore/public/application/register_tabs.ts#L194-L201

6. `insertWhereClause` is duplicated nearly verbatim in `SQLSearchInterceptor` (private static) and `SQLFilterUtils` (private static). Same ANTLR parsing, same fallback logic. Make `SQLFilterUtils.insertWhereClause` public static and call it from the interceptor so a fix to one applies to both.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/afb4db71165de92bfa252062396b7984d80667d7/src/plugins/query_enhancements/public/search/sql_search_interceptor.ts#L86-L120

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
ritvibhatt: Nope, default should be auto, removed

src/plugins/explore/public/application/register_tabs.ts
ritvibhatt: Makes sense updated to check logs instead

src/plugins/explore/public/application/register_tabs.ts
ritvibhatt: Updated to check logs instead

src/plugins/explore/public/components/query_panel/query_panel_widgets/language_toggle/language_toggle.tsx
ritvibhatt: Removed

src/plugins/query_enhancements/public/search/sql_search_interceptor.ts
ritvibhatt: Yes, not using for filters though because for example if timestamp isnt in inner query then adding the timestamp filter outside would fail

src/plugins/query_enhancements/public/plugin.tsx
ritvibhatt: Removed

src/plugins/explore/public/embeddable/explore_embeddable.tsx
ritvibhatt: Its inserting a where clause into the query in sql_search_interceptor (updated to add to the innermost subquery)

ritvibhatt: Updated to use wrap the query and use subquery instead so we don't need to parse the query

joshuali925:
i pulled it down but it's showing syntax error for me?
```
Query must start with SELECT, DELETE, SHOW or DESCRIBE: WITH `agent_traces_all_operations` AS (SELECT * FROM `agent_traces_all_operations` WHERE `startTime` >= '2025-05-27 22:44:58.514' AND `startTime` <= '2026-05-27 22:44:58.514') SELECT * FROM agent_traces_all_operations LIMIT 10
```

ritvibhatt:
> i pulled it down but it's showing syntax error for me?
> 
> ```
> Query must start with SELECT, DELETE, SHOW or DESCRIBE: WITH ``agent_traces_all_operations`` AS (SELECT * FROM ``agent_traces_all_operations`` WHERE ``startTime`` >= '2025-05-27 22:44:58.514' AND ``startTime`` <= '2026-05-27 22:44:58.514') SELECT * FROM agent_traces_all_operations LIMIT 10
> ```

It'll only work with the analytics engine backend since it needs the calcite engine for sql, updated docs to add that as a prereq

src/plugins/query_enhancements/public/search/filters/sql_filter_utils.ts
```diff
@@ -0,0 +1,48 @@
...
+   */
+  public static wrapWithTimeFilterCTE(sql: string, tableName: string, whereClause: string): string {
+    if (!tableName) return sql;
+
+    const escapedName = tableName.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+    const collisionRe = new RegExp(`\\bWITH\\b[\\s\\S]*?\\b${escapedName}\\b\\s+AS\\s*\\(`, 'i');
+    if (collisionRe.test(sql)) return sql;
+
+    const ourCte = `\`${tableName}\` AS (SELECT * FROM \`${tableName}\` WHERE ${whereClause})`;
+    const leadingWithRe = /^(\s*)WITH(\s+)/i;
```
joshuali925: i'm mainly concerned about the regex based query manipulations, otherwise PR looks good to me. is this robust enough to handle arbitrary user queries? we had antlr parser before, is it possible to use antlr?

joshuali925 (dismissed):
see comment

src/plugins/query_enhancements/public/search/filters/sql_filter_utils.ts
ritvibhatt: The regex is just a guard against the user already defining a CTE with the same name as their dataset table and if its found it'll degrade gracefully (skip wrap). Since this is for analytics engine we can't use the antlr grammar since we're using calcite sql parser

ritvibhatt: updated back to using where with antlr parser since routing to the analytics engine is using the v2 parser anyways

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
```diff
@@ -553,7 +555,7 @@ const executeQueryBase = async (
     if (includeHistogram) {
       // Histogram-specific: Get interval and create with aggregations
       const state = getState();
-      const effectiveInterval = interval || state.legacy?.interval || 'auto';
+      const effectiveInterval = interval || state.legacy?.interval;
```
angle943: Since needsHistogramQuery already excludes SQL (query.language !== 'SQL'), this code path only runs for PPL. If
  state.legacy?.interval is undefined (e.g., fresh state), effectiveInterval is undefined and gets passed to
  createSearchSourceWithQuery. The old code defaulted to 'auto'. This is a behavioral regression for PPL histogram queries
   that's unrelated to SQL support. Should be reverted or split into a separate PR with its own justification.

src/plugins/explore/public/components/query_panel/query_panel_widgets/language_toggle/language_toggle.tsx
```diff
@@ -44,24 +49,91 @@ export const LanguageToggle = () => {
...
+
+      // Get the default query string for the new language
+      // SQL needs a base query (SELECT * FROM ...) to be valid; PPL works with empty
+      const newQueryString =
+        newLanguage === 'PPL' ? '' : langConfig?.getQueryString?.(currentQuery) ?? '';
+
+      queryString.setQuery({ query: newQueryString, language: newLanguage, dataset });
+      languageSvc.setUserQueryLanguage(newLanguage);
+      dispatch(setQueryWithHistory({ ...queryString.getQuery() }));
+      setTimeout(focusOnEditor);
```
angle943: - No useEffect cleanup — if the component unmounts before the timeout fires, this dispatches against a potentially stale
  store
  - services is captured from getServices() inside the callback, which is fine, but dispatch from useDispatch() could be
  stale if the component tree re-renders between the click and the timeout

src/plugins/explore/public/application/utils/state_management/actions/query_actions.ts
ritvibhatt: Reverted back to original

src/plugins/explore/public/components/query_panel/query_panel_widgets/language_toggle/language_toggle.tsx
ritvibhatt: Fixed added a pendingTimeouts ref and useEffect cleanup that clears any pending timeouts on unmount. The dispatch from useDispatch() is stable across re-renders so it's not actually stale, but the cleanup ensures we don't dispatch post-unmount either way.

angle943 (approved):

joshuali925 (approved):

Maosaic (approved):

PR #12070, opened 2026-05-27
fix(build): remove redundant external file watcher from osd-optimizer

ruanyl:
### Background

The osd-optimizer originally had a bundle cache system where bundles with valid caches were "offline" (skipped, not sent to workers) and only bundles with stale caches were "online" (sent to workers for building). The external Watchpack-based watcher (`watchBundlesForChanges$`) existed to detect file changes in "offline" bundles and move them "online" by spawning new worker processes — this was the only way those cached bundles would ever get rebuilt.

After the bundle cache removal in #11866 (which switched to rspack's persistent cache), all bundles are always "online" and built on every startup. The online/offline distinction lost its meaning — `offlineBundles` is always empty and `onlineBundles` always equals all bundles.

#### Problem

With all bundles always built on startup, each worker runs `multiCompiler.watch()` for incremental rebuilds. But the external watcher remained, detecting the same file changes and spawning new worker processes for bundles that already had living workers watching — potentially causing double builds of the same bundles.

This commit removes Watchpack-based watcher.

Rspack's internal watcher within each worker process continues to handle file change detection and incremental recompilation in watch mode.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


yubonluo (approved):

wanglam (approved):
This PR stops future cache corruption, but what about users who already have bad caches from before? Should we mention that clearing `node_modules/.cache/rspack` fixes existing issues? Or add that path to `bootstrap clean` so it gets cleaned automatically?

ruanyl:
> This PR stops future cache corruption, but what about users who already have bad caches from before? Should we mention that clearing `node_modules/.cache/rspack` fixes existing issues? Or add that path to `bootstrap clean` so it gets cleaned automatically?

Yes, I have another PR to add the `.cache` path to `yarn osd clean`, will merge this PR now. 
Please also help check: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/12057

PR #12126, opened 2026-06-01
fix: dashboard admin user gets correct Owner permissionMode for workspaces

SuZhou-Joe:
### Description

Fix an issue where dashboard admin users get `permissionMode` as `read` when getting workspaces without permissions entry, even though they are effectively owners of all workspaces.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


SuZhou-Joe:
> Handle undefined workspace state safely

The getWorkspaceState function will always ensure there is result and won't be null.

Hailong-am (approved):
Thanks

yubonluo (approved):

Issue #11835, opened 2026-04-27
Console warnings in development mode


ruanyl:
Currently, there are quite a bit of console warnings nearly on every page, example:

<img width="1720" height="740" alt="Image" src="https://github.com/user-attachments/assets/50a89b48-a3b0-4660-8176-19a32a367ca0" />

PR #12140, opened 2026-06-01
chore(deps): upgrade react-intl from 2.9.0 to 6.8.9

### LINKED ISSUE #12166: Upgrade react-intl from 2.9.0 to 6.8.9, breaking changes in @osd/i18n
## Description

This react-intl upgrade is required to adjust to changes in React and address React warnings "legacy contextTypes API which is no longer supported and will be removed in the next major release. Use React.createContext() with static contextType instead." from react-intl module.
Each major version upgrade or react-intl has braking changes with code adjustments need it:

https://formatjs.github.io/docs/react-intl/upgrade-guide-3.x
https://formatjs.github.io/docs/react-intl/upgrade-guide-4.x
https://formatjs.github.io/docs/react-intl/upgrade-guide-5.x
https://formatjs.github.io/docs/react-intl/upgrade-guide-6.x

Because of breaking changes, the major version change for @osd/i18n will be required, from 1.0.0 to 2.0.0

Below summary of changes which will be need it in plugins which are utilising @osd/i18n during upgrade.

### Overview

The `@osd/i18n` package upgrade from version 1.0.0 to 2.0.0 includes a major upgrade of the underlying `react-intl` library from version 2.9.0 to 6.8.9. This upgrade introduces breaking changes that require updates in plugins code.

#### 1. **Apostrophe Escaping in ICU Messages** ⚠️ **CRITICAL**

**What Changed:**
In react-intl 3.x and later, the escaping rules for apostrophes (single quotes) in ICU message format have changed.

**Old Behavior (react-intl 2.x):**
- Single apostrophes around placeholders: `'{placeholder}'` 
- This would display placeholder value surrounded by quotes
- backslash escaping (`\{` and `\}`) is supported

**New Behavior (react-intl 6.x):**
- Single apostrophes are treated as escape characters
- To display a literal apostrophe, you must use double apostrophes: `''`
- Placeholders wrapped in single quotes `'{placeholder}'` will FAIL
- The legacy backslash escaping (`\{` and `\}`) is **no longer supported**, replace with single apostrophes `'`

**Required Changes:**

❌ **Before:**
```typescript
i18n.translate('myPlugin.message', {
  defaultMessage: "Error with '{fieldName}'",
  values: { fieldName: 'username' }
});
// Output: Error with 'username'
```

✅ **After:**
```typescript
i18n.translate('myPlugin.message', {
  defaultMessage: "Error with ''{fieldName}''",
  values: { fieldName: 'username' }
});
// Output: Error with 'username'
```

❌ **Before:**
```typescript
i18n.translate('myPlugin.message2', {
  defaultMessage: "Error with \\{var\\}"
});
// Output: Error with {var}
```

✅ **After:**
```typescript
i18n.translate('myPlugin.message2', {
  defaultMessage: "Error with '{var}'"
});
// Output: Error with {var}
```

#### 2. **TypeScript Type Changes**

**What Changed:**
The TypeScript types exported by `@osd/i18n/react` have been updated to match react-intl 6.x.

**Required Changes:**

❌ **Before:**
```typescript
import { InjectedIntlProps } from '@osd/i18n/react';

interface MyComponentProps extends InjectedIntlProps {
  // ...
}
```

✅ **After:**
```typescript
import { WrappedComponentProps } from '@osd/i18n/react';

interface MyComponentProps extends WrappedComponentProps {
  // ...
}
```

**Type Mapping:**
- `InjectedIntlProps` → `WrappedComponentProps`
- `InjectedIntl` → `IntlShape`

#### 3. **Rich Text Formatting Support**

**What's New:**
The package now supports rich text formatting with HTML-like tags in messages.

**Usage:**
```typescript
i18n.translate('myPlugin.richMessage', {
  defaultMessage: 'This is <strong>bold</strong> and this is <em>italic</em>',
});
```

**Supported Tags:**
- `<b>`, `<strong>` - Bold text
- `<i>`, `<em>` - Italic text
- `<code>` - Code formatting
- `<p>` - Paragraph
- `<ul>`, `<li>` - Lists

**Note:** These tags are automatically handled by the @osd/i18n engine. You can also pass custom formatting functions in the `values` object. Custom HTML handlers are need to for HTML tags not included above

#### Step 1: Update Message Strings

Search your plugin for all i18n message definitions and update apostrophe escaping
Replace single apostrophes around placeholders with double apostrophes:
- `'{placeholder}'` → `''{placeholder}''`
Replace backslash escaping with single apostrophes escaping:
- `\{text\}` → `'{placeholder}'`

### Step 2: Update TypeScript Imports

Replace deprecated type imports:

```typescript
// Find and replace in your codebase
- import { InjectedIntlProps } from '@osd/i18n/react';
+ import { WrappedComponentProps } from '@osd/i18n/react';

- import { InjectedIntl } from '@osd/i18n/react';
+ import { IntlShape } from '@osd/i18n/react';
```

#### Step 3: Update Translation Files

If you maintain translation JSON files, replace backslash escaping with single apostrophes escaping, and single apostrophes around placeholder values with double apostrophes

```json
{
  "myPlugin.message": "Error with ''{fieldName}''"
}
```

#### Step 4: Test Your Plugin

1. **Visual Testing:** Check all UI strings that contain apostrophes around dynamic values
2. **Unit Tests:** Update snapshots if they contain i18n strings
3. **Integration Tests:** Verify that messages display correctly in different locales

### Validation Tools

The OpenSearch Dashboards i18n extraction tools have been updated to detect and warn about:

1. **Single apostrophe wrapping:** Messages using `'{placeholder}'` instead of `''{placeholder}''` if placeholder value is provided. If `'` is not expected in message the `{placeholder}` without any apostrophe is correct and typical
2. **Legacy escaped braces:** Messages using `\{placeholder\}` instead of `'{placeholder}'`

Run the i18n checker to validate your changes:

```bash
node scripts/i18n_check
```

### Additional Resources

For more details on the react-intl upgrade, refer to the official upgrade guides:
- [react-intl 3.x Upgrade Guide](https://formatjs.github.io/docs/react-intl/upgrade-guide-3.x)
- [react-intl 4.x Upgrade Guide](https://formatjs.github.io/docs/react-intl/upgrade-guide-4.x)
- [react-intl 5.x Upgrade Guide](https://formatjs.github.io/docs/react-intl/upgrade-guide-5.x)
- [react-intl 6.x Upgrade Guide](https://formatjs.github.io/docs/react-intl/upgrade-guide-6.x)
- [ICU Message Format Documentation](https://formatjs.github.io/docs/core-concepts/icu-syntax)

## FAQ

**Q: Why do I need to use double apostrophes?**
A: In ICU message format (used by react-intl), a single apostrophe is an escape character. To display a literal apostrophe, you must escape it by doubling it.

**Q: What happens if I don't update my messages?**
A: Messages with single apostrophes around placeholders will fail

**Q: Do I need to update messages without placeholders?**
A: No. Simple apostrophes in regular text (not around placeholders) work fine: `"Don't worry"` remains `"Don't worry"`.

**Q: Can I use the old escaping style?**
A: No. The old `\{placeholder\}` style is no longer supported and will cause errors. You must use the new apostrophe-based escaping.

**Q: How do I escape curly braces that are not placeholders?**
A: Wrap them in single apostrophes: `"Use '{ and }' for objects"` will display as `"Use { and } for objects"`.

tomkdgun:
### Description

This PR upgrades react-intl from version 2.9.0 to 6.8.9 to address React warnings "legacy contextTypes API which is no longer supported and will be removed in the next major release. Use React.createContext() with static contextType instead." from react-intl module.
Each major version upgrade or react-intl has braking changes with code adjustments need it:
- https://formatjs.github.io/docs/react-intl/upgrade-guide-3.x
- https://formatjs.github.io/docs/react-intl/upgrade-guide-4.x
- https://formatjs.github.io/docs/react-intl/upgrade-guide-5.x
- https://formatjs.github.io/docs/react-intl/upgrade-guide-6.x
- https://github.com/formatjs/formatjs/issues/1845 - [escape](https://github.com/formatjs/formatjs/issues/1845#issuecomment-855919671) need it 

Since pull request is big anyway I've decided to stop on version 6.x. Updates to newer versions or potential removal of @osd/i18n can be done in future on separate PR.

The most noticeable braking change in @osd/i18n is replacing FormattedRelative with FormattedRelativeTime, and how messages with HTML tags are handled. I haven't found any place with FormattedRelative, but because of this breaking changes I've changed version from 1.0.0 to 2.0.0.
Also escape character changed so in many places one apostrophe was replaced by two, to have apostrophe included in message on UI https://formatjs.github.io/docs/react-intl/upgrade-guide-3.x/#escape-character-has-been-changed-to-apostrophe-

**Key Changes**

Dependency Updates
react-intl: 2.9.0 → 6.8.9 - breaking Changes in react-intl 3.x, 4.x, 5.x, 6.x
@osd/i18n: 1.0.0 → 2.0.0 - major version bump due to breaking changes
react: 18.2.0 → 18.3.1 - just package.json updates, yarn.lock already have 18.3.1
react-dom: 18.2.0 → 18.3.1 - just package.json updates, yarn.lock already have 18.3.1
react-is: 18.2.0 → 18.3.1 - just package.json updates, yarn.lock already have 18.3.1

Removed deprecated components and APIs:
- FormattedRelative replaced with FormattedRelativeTime
- Removed intl-relativeformat, intl-messageformat v2, and intl-format-cache dependencies
- Added new @formatjs/* scoped packages for modular functionality

API Changes:
- FormattedRelativeTime now requires explicit value and unit parameters
- Updated format options and removed legacy relative formats
- Error message format changes in validation

**Code Modifications:**

packages/osd-i18n/
- Updated README with new FormattedRelativeTime usage examples
- Removed legacy format definitions for relative time
- Updated type definitions to match react-intl 6.x API
- Modified test snapshots to reflect new error message formats

Test Utilities (src/test_utils/public/enzyme_helpers.tsx)
- Replaced legacy getChildContext() approach with createIntl() from react-intl 6.x
- Removed intlShape from childContextTypes (no longer needed)
- Updated wrapWithIntl to use __IntlProvider with required props

Component Tests
- Updated all test files to use wrapWithIntl() helper or <I18nProvider> wrapper
- Fixed snapshot tests to match new component output structure
- Added missing <I18nProvider> wrappers in workspace components
- Enzyme tests with new react-intl sometimes require special helper wrappers e.g. to set props on wrapped element instead of I18nProvider, with @testing-library/react it's much easier, but re-writting all enzyme tests would make this pull request too big, I've modified few test only this way

### Issues Resolved

partially address #11835
closes #12166

## Testing the changes

- Automated Testing, all existing jest tests updated and passing
- Snapshot tests regenerated for new component structure
- Manual Testing

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


tomkdgun:
@angle943 @ruanyl The failing ciGroup23Explore it's flaky, it was passing before, and I was able to have successful results during local run. Pull request is ready for review.

angle943:
Bug: Inconsistent apostrophe escaping in `create_edit_field.tsx`

In `src/plugins/index_pattern_management/public/components/edit_index_pattern/create_edit_field/create_edit_field.tsx` (~line 74):

```ts
defaultMessage:
  "'{indexPatternTitle}' index pattern doesn't have a scripted field called ''{fieldName}''",
```

`'{indexPatternTitle}'` uses single quotes (un-escaped), while `''{fieldName}''` correctly uses doubled quotes. The ICU MessageFormat parser will treat the single `'` around `{indexPatternTitle}` as an escape sequence start, suppressing the quotes in the rendered output.

Should be:
```ts
"''{indexPatternTitle}'' index pattern doesn''t have a scripted field called ''{fieldName}''",
```

The equivalent fix was correctly applied in `dataset_management/create_edit_field.tsx`.

angle943:
Nit: In `packages/osd-i18n/src/core/formats.ts`, the `Formats` type is imported directly from `intl-messageformat`:

```ts
import { Formats } from 'intl-messageformat';
```

`intl-messageformat` is a transitive dependency of `react-intl`, not a direct dependency of `@osd/i18n`. If `react-intl` ever bumps its `intl-messageformat` peer/dependency to a version with a different `Formats` shape, this import could break silently. Consider importing from `react-intl` instead (which re-exports the relevant types), or explicitly adding `intl-messageformat` as a direct dependency in `packages/osd-i18n/package.json` to pin the expected version.

src/plugins/index_pattern_management/public/components/edit_index_pattern/create_edit_field/create_edit_field.tsx
```diff
@@ -74,7 +74,7 @@ export const CreateEditField = withRouter(
         'indexPatternManagement.editIndexPattern.scripted.noFieldLabel',
         {
           defaultMessage:
-            "'{indexPatternTitle}' index pattern doesn't have a scripted field called '{fieldName}'",
+            "'{indexPatternTitle}' index pattern doesn't have a scripted field called ''{fieldName}''",
```
seraphjiang: 🔴 **Swallowed placeholder (partial apostrophe doubling).** `'{indexPatternTitle}'` was left single-quoted while `''{fieldName}''` was doubled. In intl-messageformat v10 a single `'` opens a quoted/escaped section, so `'{indexPatternTitle}'` renders as the literal text `{indexPatternTitle}` — the actual index-pattern title is lost.

Verified in a v10 harness:
```
input:  values={indexPatternTitle:'logs-*', fieldName:'foo'}
output: "{indexPatternTitle} index pattern doesn't have a scripted field called 'foo'"
```
The sibling `dataset_management/.../create_edit_field.tsx` doubled **both** correctly. Fix: `''{indexPatternTitle}''`.

**Suggested fix:**
```suggestion
            "''{indexPatternTitle}'' index pattern doesn't have a scripted field called ''{fieldName}''",
```

src/plugins/explore/public/embeddable/explore_embeddable.tsx
```diff
@@ -674,7 +675,11 @@ export class ExploreEmbeddable
   private renderComponent(node: HTMLElement, searchProps: SearchProps) {
     if (!this.searchProps || !this.root) return;
     const MemorizedExploreEmbeddableComponent = React.memo(ExploreEmbeddableComponent);
-    this.root.render(<MemorizedExploreEmbeddableComponent searchProps={searchProps} />);
+    this.root.render(
+      <I18nProvider>
```
seraphjiang, line 681: ✅ This `<I18nProvider>` wrapper is the right fix — but the same `createRoot(...).render(...)` pattern exists in **three sibling render roots that did _not_ get it**, and each renders a `FormattedMessage`/`useIntl` child. Under react-intl 6 those throw `Could not find required "intl" object. <IntlProvider> needs to exist in the component ancestry` (react-intl 2 silently fell back to `defaultMessage`, which is why this never crashed before). Verified the throw in a react-intl 6.8.9 harness.

Please add the same `<I18nProvider>` wrapper to:
- `src/plugins/agent_traces/public/embeddable/agent_traces_embeddable.tsx:374` — child `DataTable` uses `<FormattedMessage>` (`data_table.tsx:201,209`); behind a conditional footer so Cypress didn't trip it.
- `src/plugins/advanced_settings/public/register_nav_control.tsx:16` — mounts `HeaderUserThemeMenu`, which renders `<FormattedMessage>` (`header_user_theme_menu.tsx:208`); wrapped only in `OpenSearchDashboardsContextProvider` (not intl).
- `src/plugins/visualizations/public/embeddable/disabled_lab_embeddable.tsx:56` — child `DisabledLabVisualization` renders `<FormattedMessage>` (`disabled_lab_visualization.tsx:47`); crashes a lab-stage panel when `visualize:enableLabs` is off.

src/plugins/vis_type_timeseries/public/application/components/color_rules.js
```diff
@@ -72,28 +72,28 @@ class ColorRulesUI extends Component {
       {
         label: intl.formatMessage({
           id: 'visTypeTimeseries.colorRules.greaterThanLabel',
-          defaultMessage: '> greater than',
+          defaultMessage: '&gt; greater than',
```
seraphjiang: 🟠 **Literal `&gt;`/`&lt;` in a plain-text combobox.** These four operator labels feed `EuiCompressedComboBox` with `singleSelection={{ asPlainText: true }}` (line ~180), so the entities render verbatim — the dropdown shows `&gt; greater than`, `&gt;= greater than or equal`, `&lt; less than`, `&lt;= less than or equal`.

Lone `<`/`>` format fine in intl-messageformat v10 (verified), so this conversion was unnecessary. Please revert all four to plain `>`, `>=`, `<`, `<=`.

**Suggested fix:**
```suggestion
          defaultMessage: '> greater than',
```

src/plugins/vis_type_timeline/server/series_functions/holt/index.js
```diff
@@ -132,7 +132,7 @@ export default new Chainable('holt', {
         if (!sample || !args.byName.season || sample < 2) {
           throw new Error(
             i18n.translate('timeline.serverSideErrors.holtFunction.missingParamsErrorMessage', {
-              defaultMessage: 'Must specify a season length and a sample size >= 2',
+              defaultMessage: 'Must specify a season length and a sample size &gt;= 2',
```
seraphjiang: 🟠 **Literal `&gt;=` in a thrown `Error` message.** This string goes into `throw new Error(i18n.translate(...))` and surfaces as plain text, so the user sees `... a sample size &gt;= 2`. A lone `>=` formats fine in v10 (verified), so revert to `>= 2`.

**Suggested fix:**
```suggestion
              defaultMessage: 'Must specify a season length and a sample size >= 2',
```

src/plugins/explore/public/application/pages/metrics/query_panel/query_row.tsx
```diff
@@ -175,7 +175,7 @@ export const QueryRowComponent: React.FC<QueryRowProps> = React.memo(
                 {!row.query && (
                   <div className="exploreQueryPanelEditor__placeholder">
                     {i18n.translate('explore.metricsQueryPanel.codePlaceholder', {
-                      defaultMessage: 'Search using </> {language}',
+                      defaultMessage: 'Search using &lt;/&gt; {language}',
```
seraphjiang: 🟠 **Literal `&lt;/&gt;` in a plain-text `<div>`.** Rendered inside `<div className="exploreQueryPanelEditor__placeholder">`, so the user sees `Search using &lt;/&gt; PPL`.

This is the one case where a change *was* needed — `</>` is tag-shaped and throws `UNMATCHED_CLOSING_TAG` in v10 — but HTML entities are the wrong escape for a plain-text context. The correct ICU fix is apostrophe-quoting: `defaultMessage: "Search using '</>' {language}"`, which renders `Search using </> PPL` (verified).

**Suggested fix:**
```suggestion
                      defaultMessage: "Search using '</>' {language}",
```

src/plugins/vis_type_timeseries/public/application/components/index_pattern.js
```diff
@@ -226,7 +226,7 @@ export const IndexPattern = ({ fields, prefix, onChange, disabled, model: _model
                   })
                 : i18n.translate('visTypeTimeseries.indexPattern.searchByIndex', {
                     defaultMessage:
-                      'Use an asterisk (*) to match multiple indices. Spaces and the characters , /, ?, ", <, >, | are not allowed.',
+                      'Use an asterisk (*) to match multiple indices. Spaces and the characters , /, ?, ", &lt;, &gt;, | are not allowed.',
```
seraphjiang: 🟠 **Literal entities in `EuiFormRow` helpText (plain text).** Renders as `... the characters , /, ?, ", &lt;, &gt;, | are not allowed.` Lone `<`/`>` format fine in v10, so revert to `<, >`. (Same string is duplicated in `annotations_editor.js:140`.)

**Suggested fix:**
```suggestion
                      'Use an asterisk (*) to match multiple indices. Spaces and the characters , /, ?, ", <, >, | are not allowed.',
```

src/plugins/vis_type_timeseries/public/application/components/index_pattern.js
```diff
@@ -267,7 +267,7 @@ export const IndexPattern = ({ fields, prefix, onChange, disabled, model: _model
               defaultMessage: 'Interval',
             })}
             helpText={i18n.translate('visTypeTimeseries.indexPattern.intervalHelpText', {
-              defaultMessage: 'Examples: auto, 1m, 1d, 7d, 1y, >=1m',
+              defaultMessage: 'Examples: auto, 1m, 1d, 7d, 1y, &gt;=1m',
```
seraphjiang: 🟠 **Literal `&gt;=1m` in helpText.** Renders as `Examples: auto, 1m, 1d, 7d, 1y, &gt;=1m`. Note the `description` field right below still says `>=1m` — so the developer-facing note and the user-facing text now disagree. Revert to `>=1m`.

**Suggested fix:**
```suggestion
              defaultMessage: 'Examples: auto, 1m, 1d, 7d, 1y, >=1m',
```

src/plugins/vis_type_timeseries/public/application/components/annotations_editor.js
```diff
@@ -137,7 +137,7 @@ export class AnnotationsEditor extends Component {
                   }
                   helpText={i18n.translate('visTypeTimeseries.indexPattern.searchByIndex', {
                     defaultMessage:
-                      'Use an asterisk (*) to match multiple indices. Spaces and the characters , /, ?, ", <, >, | are not allowed.',
+                      'Use an asterisk (*) to match multiple indices. Spaces and the characters , /, ?, ", &lt;, &gt;, | are not allowed.',
```
seraphjiang: 🟠 **Literal entities in helpText (plain text).** Same regression as `index_pattern.js:229` — user sees `, &lt;, &gt;, |`. Revert to `<, >`.

**Suggested fix:**
```suggestion
                      'Use an asterisk (*) to match multiple indices. Spaces and the characters , /, ?, ", <, >, | are not allowed.',
```

packages/osd-i18n/src/react/provider.tsx
```diff
@@ -55,15 +60,25 @@ export class I18nProvider extends React.PureComponent<I18nProviderProps> {
...
+      <IntlProvider
         locale={i18n.getLocale()}
         messages={i18n.getTranslation().messages}
         defaultLocale={i18n.getDefaultLocale()}
         formats={i18n.getFormats()}
         textComponent={React.Fragment}
+        defaultRichTextElements={globalRichTextElements}
+        onError={(err) => {
+          // Suppress missing translation errors for en locale
+          if (err.code === ReactIntlErrorCode.MISSING_TRANSLATION && i18n.getLocale() === 'en') {
```
seraphjiang: 🟠 **`console.error` flood for every non-`en` locale.** This guard suppresses `MISSING_TRANSLATION` only when locale is `en`. Verified against react-intl 6.8.9: with any non-`en` locale and a partial translation bundle (the normal case — most locales don't translate 100% of keys), `formatMessage` fires `MISSING_TRANSLATION` for *every* key that falls back, and this handler `console.error`s each one. A single page can emit hundreds of error lines and trip any test asserting on `console.error`. Consider suppressing `MISSING_TRANSLATION` regardless of locale (it's expected with partial bundles), while still logging real `FORMAT_ERROR`s.

packages/osd-i18n/src/core/i18n.ts
```diff
@@ -28,21 +28,48 @@
...
+  ul: (chunks: any) => `<ul>${chunks}</ul>`,
+};
 const translationsForLocale: Record<string, Translation> = {};
-const getMessageFormat = memoizeIntlConstructor(IntlMessageFormat);
+
+// Memoization cache for IntlMessageFormat instances
+const messageFormatCache = new Map<string, IntlMessageFormat>();
+
+function getMessageFormat(message: string, locale: string, formats?: Formats): IntlMessageFormat {
+  const cacheKey = `${locale}:${message}`;
```
seraphjiang, line 55: 🟡 **Cache key omits `formats`.** The old `intl-format-cache` memoizer keyed on `(message, locale, formats)`; this key drops `formats`. After `setFormats(...)` changes formatting at runtime, any already-cached `(locale, message)` returns a formatter built with the *old* formats. Latent today (the only `setFormats` caller is `init()`, pre-format) but the invariant is silently lost. Separately, this `Map` has no eviction — it grows unbounded for the session, and the `@internal` `clearMessageFormatCache` is never called in production.

packages/osd-i18n/src/core/i18n.ts
```diff
@@ -220,14 +245,21 @@ export function translate(id: string, { values = {}, defaultMessage }: Translate
 
   if (message) {
     try {
+      // IntlMessageFormat natively supports functions in values for rich text formatting
       // We should call `format` even for messages without any value references
       // to let it handle escaped curly braces `\\{` that are the part of the text itself
       // and not value reference boundaries.
-      const formattedMessage = getMessageFormat(message, getLocale(), getFormats()).format(values);
+      const formattedMessage = getMessageFormat(message, getLocale(), getFormats()).format({
+        ...globalRichTextElements,
```
seraphjiang: 🟡 **Per-call allocation in the hottest function.** `{ ...globalRichTextElements, ...values }` allocates a fresh 8-function object on *every* `translate()` call — including the ~99% of messages with no tags — adding ~50% overhead vs the cached `format()` plus GC churn (`translate` runs thousands of times per render). Cheap guard: only spread the handlers when the message can contain a tag, e.g. `message.indexOf('<') === -1 ? values : { ...globalRichTextElements, ...values }`. (Same spread is duplicated in the default-message branch at line 273.)

packages/osd-i18n/src/react/pseudo_locale_wrapper.tsx
```diff
@@ -61,30 +61,37 @@ interface PseudoLocaleWrapperProps {
...
 
-  constructor(props: PseudoLocaleWrapperProps, context: any) {
-    super(props, context);
+    // Create a wrapped intl object with modified formatMessage
+    return {
+      ...intl,
+      formatMessage: ((descriptor: any, values?: any, opts?: any) => {
+        const result = intl.formatMessage(descriptor, values, opts);
+        // Only translate if result is a string (not ReactNode[])
+        if (typeof result === 'string') {
```
seraphjiang, line 82: 🟡 **Pseudo-locale silently skips rich-text messages.** Scrambling only runs when `typeof result === 'string'`, but react-intl 6 returns `ReactNode[]` for any message containing tags/rich-text — those hit the `return result` path unscrambled. So `en-xa` QA coverage regresses for every markup-bearing string (the old context-override wrapped all `formatMessage` calls). Dev-only, but it quietly narrows pseudo-localization coverage.

seraphjiang (commented):
## Review — react-intl 2→6 / intl-messageformat v2→v10 semantics

Thanks for tackling this upgrade 🙏 I verified every point below empirically against `intl-messageformat@10.7.7` and `react-intl@6.8.9` (the versions this PR pins). CI is green, but these are defects that unit/Cypress tests structurally can't catch: crashes behind feature-flags/conditional branches, literal text in widgets no snapshot asserts, and non-`en` locales.

Inline comments are attached to the relevant changed lines. A few items below are **pre-existing strings/files not in this diff** that the engine swap nonetheless breaks, so GitHub won't let me comment inline on them:

**🔴 Crash — `dashboard_variables/variable_editor_flyout.tsx:364` (not in diff).** The message `'Use this name to reference the variable: $var or $\{var\}'` uses v2 backslash brace-escaping, which v10 rejects with `MALFORMED_ARGUMENT`. `i18n.translate` rethrows, so the variable-editor flyout crashes when rendered (explore/observability workspaces). v10 uses apostrophe escaping — write the literal braces as `$'{'var'}'`. Verified throwing in a v10 harness. This is the only remaining `\{` in an i18n message (the other repo hits are in RegExps).

**🟠 Wrong output for `es-ES` — `src/translations/es-ES.json:2569` (not in diff).** `"visualizations.newVisWizard.resultsFound"` uses non-ICU plural keywords `uno {…} otro {…}`; v10 requires the literal `other`, so this throws `MISSING_OTHER_CLAUSE`. It's the only malformed message across all 14 translation files. Via `<FormattedMessage>` it falls back to English + console-floods; on the `i18n.translate` string path it would hard-crash. Fix: `one {type} other {types}`.

---

### Note on the two systematic patterns

1. **The `&lt;`/`&gt;` conversions are mostly unnecessary *and* wrong.** I confirmed that lone `<`, `>`, `>=`, `<=` format fine in v10 (`"< less than"`, `"sample size >= 2"`, `"Examples: >=1m"` all render correctly) — only tag-shaped tokens (`</>`, `<script>`, unclosed `<p>`) throw. So most of these strings needed no change, and HTML entities render *literally* in plain-text contexts (combobox labels, `EuiFormRow` helpText, `Error()`, plain `<div>`). For the one genuinely tag-shaped case (`</>`), the correct ICU fix is apostrophe-quoting (`'</>'`), not entities. (In `dangerouslySetInnerHTML` contexts like advanced-settings descriptions, the entities are harmless — those are fine.)

2. **The apostrophe doubling has no guard rail.** It's correct and necessary, but it was done as a manual ~40-file sweep and one instance slipped through (see the `create_edit_field.tsx` inline comment). Consider an `i18n_check` rule that parses each `defaultMessage` through intl-messageformat (or flags single-`'`-wrapped placeholders) so future strings can't regress.

Full methodology + the items I checked and *cleared* (other render roots, removed exports, `select`/`other` clauses, the `enzyme_helpers` rewrite) are in the summary I can share separately.

src/plugins/vis_type_timeseries/public/application/components/color_rules.js
```diff
@@ -72,28 +72,28 @@ class ColorRulesUI extends Component {
...
         }),
         value: 'gt',
       },
       {
         label: intl.formatMessage({
           id: 'visTypeTimeseries.colorRules.greaterThanOrEqualLabel',
-          defaultMessage: '>= greater than or equal',
+          defaultMessage: '&gt;= greater than or equal',
```
seraphjiang: 🟠 Same plain-text combobox label — revert the entity to `>=` (formats fine in v10).

**Suggested fix:**
```suggestion
          defaultMessage: '>= greater than or equal',
```

src/plugins/vis_type_timeseries/public/application/components/color_rules.js
```diff
@@ -72,28 +72,28 @@ class ColorRulesUI extends Component {
...
         }),
         value: 'gte',
       },
       {
         label: intl.formatMessage({
           id: 'visTypeTimeseries.colorRules.lessThanLabel',
-          defaultMessage: '< less than',
+          defaultMessage: '&lt; less than',
```
seraphjiang: 🟠 Same plain-text combobox label — revert the entity to `<` (formats fine in v10).

**Suggested fix:**
```suggestion
          defaultMessage: '< less than',
```

seraphjiang (commented):
Follow-up suggestion blocks for the remaining three `color_rules.js` operator labels (the first is on the comment above at line 75).

src/plugins/vis_type_timeseries/public/application/components/color_rules.js
```diff
@@ -72,28 +72,28 @@ class ColorRulesUI extends Component {
...
         }),
         value: 'lt',
       },
       {
         label: intl.formatMessage({
           id: 'visTypeTimeseries.colorRules.lessThanOrEqualLabel',
-          defaultMessage: '<= less than or equal',
+          defaultMessage: '&lt;= less than or equal',
```
seraphjiang: 🟠 Same plain-text combobox label — revert the entity to `<=` (formats fine in v10).

**Suggested fix:**
```suggestion
          defaultMessage: '<= less than or equal',
```

packages/osd-i18n/src/core/i18n.ts
tomkdgun: @seraphjiang What do you think about removing this cache completely. New version of react-intl has few solutions to improve performance, memorised format messages etc. ?

packages/osd-i18n/src/react/pseudo_locale_wrapper.tsx
tomkdgun: @seraphjiang Since pull request is big I think I will try to handle this on separate PR, together with improvements in test files

packages/osd-i18n/src/core/i18n.ts
seraphjiang: Great question, and thanks for being so responsive on this! 🙏

The subtlety is *which* path this cache serves. This `getMessageFormat` lives in the **core** `i18n.ts`, which builds `new IntlMessageFormat(...)` directly from `intl-messageformat` and never goes through react-intl. So react-intl 6's memoization (`createIntl`'s internal cache, etc.) only kicks in for the **React** path — `<FormattedMessage>`, `useIntl`, `injectIntl`, all already covered by `IntlProvider`. It doesn't cover the imperative `i18n.translate()` path, which is ~11k call sites across the codebase (toasts, validation errors, aria-labels, server-side strings). The pre-PR code also cached these (via `intl-format-cache`'s `memoizeIntlConstructor`), so dropping the cache with no replacement would re-parse every ICU message on each call — a regression on that path.

So I'd lean against removing it outright, but your instinct to lean on react-intl is the right direction. Two options:

1. **Idiomatic** — refactor `translate()` onto `createIntl()` + `createIntlCache()` from `@formatjs/intl` (already in the tree, and you're already using `createIntl` in `enzyme_helpers.tsx` in this PR). That lets us delete the hand-rolled `Map` *and* fixes the stale-`formats` issue from the comment above, since FormatJS keys its cache on `formats`. The catch is `createIntl` is `{ id, defaultMessage, messages }`-shaped, so `translate()` needs a modest rework.
2. **Minimal** — keep the `Map` but add `formats` to the cache key and clear it on locale/translation reload. Much smaller diff if you'd rather keep this PR focused and do the bigger refactor separately.

Either is fine by me — happy to pair on the `createIntl` approach if you want to go that route. 👍

angle943:
## 🔴 Must-Fix Issues

### 1. Crash: `variable_editor_flyout.tsx:364` (not in diff)

The message `'Use this name to reference the variable: $var or $\{var\}'` uses v2 backslash brace-escaping (`\{`), which intl-messageformat v10 rejects with `MALFORMED_ARGUMENT`. Since `i18n.translate` rethrows, the variable-editor flyout will crash when rendered in explore/observability workspaces.

**Fix:** Use apostrophe escaping instead: `$'{'var'}'`

### 2. Missing `<I18nProvider>` on 3 sibling render roots (not in diff)

The `<I18nProvider>` wrapper added to `explore_embeddable.tsx` is correct, but the same `createRoot(...).render(...)` pattern exists in three siblings that also render `FormattedMessage`/`useIntl` children. Under react-intl 6 these throw `"Could not find required intl object"` (v2 silently fell back to `defaultMessage`):

- `src/plugins/agent_traces/public/embeddable/agent_traces_embeddable.tsx:374`
- `src/plugins/advanced_settings/public/register_nav_control.tsx:16`
- `src/plugins/visualizations/public/embeddable/disabled_lab_embeddable.tsx:56`

### 3. `es-ES` translation crash (not in diff)

`src/translations/es-ES.json:2569` — `"visualizations.newVisWizard.resultsFound"` uses non-ICU plural keywords `uno/otro` instead of `one/other`, causing `MISSING_OTHER_CLAUSE` error in v10.

---

## 🟠 Should-Fix: Systematic HTML entity issue

Multiple places convert `<`/`>` to `&lt;`/`&gt;` unnecessarily. Lone `<`, `>`, `>=`, `<=` format fine in intl-messageformat v10 — only tag-shaped tokens (`</>`, `<script>`) throw. The entities render *literally* in plain-text contexts (combobox labels, Error messages, helpText props). The affected files have inline comments from @seraphjiang already — just confirming those all need reverting.

For the one genuinely tag-shaped case (`</>` in `query_row.tsx`), use ICU apostrophe-quoting: `"Search using '</>' {language}"` instead of HTML entities.

### `console.error` flood for non-`en` locales

In `packages/osd-i18n/src/react/provider.tsx`, the `onError` handler only suppresses `MISSING_TRANSLATION` when locale is `en`. With any non-`en` locale and a partial translation bundle, `formatMessage` fires `MISSING_TRANSLATION` for every key that falls back, flooding the console. Consider suppressing `MISSING_TRANSLATION` regardless of locale while still logging `FORMAT_ERROR`s.

src/plugins/index_pattern_management/public/components/edit_index_pattern/create_edit_field/create_edit_field.tsx
```diff
@@ -74,7 +74,7 @@ export const CreateEditField = withRouter(
         'indexPatternManagement.editIndexPattern.scripted.noFieldLabel',
         {
           defaultMessage:
-            "'{indexPatternTitle}' index pattern doesn't have a scripted field called '{fieldName}'",
+            "''{indexPatternTitle}'' index pattern doesn't have a scripted field called ''{fieldName}''",
```
angle943, line 77: 🔴 **Swallowed placeholder.** `'{indexPatternTitle}'` is still single-quoted while `''{fieldName}''` was correctly doubled. In v10, the single `'` opens an escape section, so the actual index-pattern title is lost (renders literal text `{indexPatternTitle}`).

Should be: `''{indexPatternTitle}''` — same fix already applied in the sibling `dataset_management/.../create_edit_field.tsx`.

packages/osd-i18n/src/react/provider.tsx
```diff
@@ -55,15 +60,25 @@ export class I18nProvider extends React.PureComponent<I18nProviderProps> {
...
+      <IntlProvider
         locale={i18n.getLocale()}
         messages={i18n.getTranslation().messages}
         defaultLocale={i18n.getDefaultLocale()}
         formats={i18n.getFormats()}
         textComponent={React.Fragment}
+        defaultRichTextElements={globalRichTextElements}
+        onError={(err) => {
+          // Suppress missing translation errors, it's common for OSD
+          if (err.code === ReactIntlErrorCode.MISSING_TRANSLATION) {
```
angle943, line 72: 🟠 **`console.error` flood for non-`en` locales.** This guard suppresses `MISSING_TRANSLATION` only when locale is `en`. With any non-`en` locale and a partial translation bundle (the normal case), `formatMessage` fires `MISSING_TRANSLATION` for *every* key that falls back, and this handler `console.error`s each one. A single page can emit hundreds of error lines and trip any test asserting on `console.error`.

Consider suppressing `MISSING_TRANSLATION` regardless of locale (it's expected with partial bundles), while still logging real `FORMAT_ERROR`s.

packages/osd-i18n/src/core/i18n.ts
```diff
@@ -28,21 +28,48 @@
...
+  ul: (chunks: any) => `<ul>${chunks}</ul>`,
+};
 const translationsForLocale: Record<string, Translation> = {};
-const getMessageFormat = memoizeIntlConstructor(IntlMessageFormat);
+
+// Memoization cache for IntlMessageFormat instances
+const messageFormatCache = new Map<string, IntlMessageFormat>();
+
+function getMessageFormat(message: string, locale: string, formats?: Formats): IntlMessageFormat {
+  const cacheKey = `${locale}:${message}`;
```
angle943, line 55: 🟡 **Cache key omits `formats`.** The old `intl-format-cache` memoizer keyed on `(message, locale, formats)`; this key drops `formats`. If `setFormats(...)` is ever called after init, any already-cached `(locale, message)` returns a formatter built with the *old* formats. Latent today but the invariant is silently lost.

Also, this `Map` has no eviction — it grows unbounded for the session. Consider either adding `formats` to the key + clearing on reload, or migrating to `createIntl()`/`createIntlCache()` from `@formatjs/intl` (which handles this correctly).

src/plugins/index_pattern_management/public/components/edit_index_pattern/create_edit_field/create_edit_field.tsx
tomkdgun: It's already addressed. I've also added checks, so such places are automatically discover when `yarn run i18n:check` is called, so lint for pull requests will find such places

packages/osd-i18n/src/react/provider.tsx
tomkdgun: Already addressed in current code base

packages/osd-i18n/src/core/i18n.ts
tomkdgun: In current code base, the clearMessageFormatCache is called in setFormats, so every time formats change, cache is cleared

packages/osd-i18n/src/core/i18n.ts
```diff
@@ -28,21 +28,48 @@
...
 const EN_LOCALE = 'en';
+const gRichTextElements = {
+  b: (chunks: any) => `<b>${chunks}</b>`,
+  code: (chunks: any) => `<code>${chunks}</code>`,
+  em: (chunks: any) => `<em>${chunks}</em>`,
+  i: (chunks: any) => `<i>${chunks}</i>`,
+  li: (chunks: any) => `<li>${chunks}</li>`,
+  p: (chunks: any) => `<p>${chunks}</p>`,
+  strong: (chunks: any) => `<strong>${chunks}</strong>`,
+  ul: (chunks: any) => `<ul>${chunks}</ul>`,
```
ruanyl, line 47: Is it possible to ingore the tags so it behaves the same as before? Just two mild concerns:
1. could this potentially be a breaking change?
2. since we injected raw html, could this cause XSS? might not be a problem if `chunks` from a managed source, not user input.

tomkdgun: @ruanyl Before such messages with HTML were present and handled before, so it's not new feature but different way of support e.g. https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a4ad56a70f3359a6b40ae88a06bb1876d8dda44b/src/plugins/data/server/ui_settings.ts#L235-L244

To make it work as before I have added these default handlers.

The HTML tags are discovered and handled safely by library. If message will contain tag which is not in this list, it needs to be handled in values in component code, so it's breaking change for such rare cases:
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/12140/changes#diff-cdcc8926f8a4b7c0664cccbc4c6e703aff4197fb0dd7e2d433e213d9f259f388
image
https://github.com/opensearch-project/OpenSearch-Dashboards/pull/12140/changes#diff-e5d22c2a14c399b4d0cc4c70f565ce936980f564af1b7cf752a697b6f72d425f
image

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/12166 has section "3. Rich Text Formatting Support".

ruanyl (dismissed):

PR #12146, opened 2026-06-02
[Workspace] Fix 500 error for identity source APIs and optimize collaborator page

gaobinlong:
### Description

Fix 500 errors from identity source APIs by returning proper 400 responses, and optimize the workspace collaborator page with batched requests and inline error display.

Server-side: Fix 500 errors

  - Wrap getIdentitySourceHandler() calls in try/catch in both GET and POST /identity/_entries routes, returning res.badRequest()
  instead of letting unhandled exceptions bubble up as 500s.
  - Use Object.create(null) for the identity source registry to prevent prototype pollution from keys like constructor or __proto__.

  Client-side: Collaborator table optimization

  - Batch identity lookup requests into chunks of 1000 IDs to avoid oversized payloads.
  - Show a toast notification when fetching collaborator names fails, with the error detail when available.

  Client-side: Add collaborators modal error handling

  - Surface API errors inline in the collaborator input fields via a new onSearchError callback chain (WorkspaceCollaboratorInput →
  WorkspaceCollaboratorsPanel → AddCollaboratorsModal).
  - Clear errors when a subsequent search succeeds or the input is emptied.

### Issues Resolved

No issue.

## Screenshot

Collaborators

Add Users

## Testing the changes

Register identity source APIs in server side, following https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9137
In src/plugins/workspace/public/register_default_collaborator_types.tsx, add identitySource field into each workspaceCollaboratorType and AddCollaboratorModal.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


gaobinlong:
@SuZhou-Joe @Hailong-am @wanglam , please help to review this PR, thanks!

src/plugins/workspace/public/components/workspace_form/workspace_collaborator_table.tsx
```diff
@@ -207,26 +209,41 @@ export const WorkspaceCollaboratorTable = ({
...
+                  '/api/security/identity/_entries',
+                  {
+                    body: JSON.stringify({ source, type, ids: slice }),
+                    signal: abortController.signal,
+                  }
+                );
+                resp?.forEach(({ id, name, alias }) => {
+                  newMap[id] = { name, alias };
+                });
+              }
```
Hailong-am, line 224: nit: we can group by `MAX_IDS_PER_REQUEST` first and call API in parallel. But need to make sure it won't get throttled.

src/core/server/security/router/index.ts
```diff
@@ -61,7 +66,12 @@ export function registerRoutes({
       },
     },
     router.handleLegacyErrors(async (context, req, res) => {
-      const handler = identitySourceService.getIdentitySourceHandler(req.body.source);
+      let handler;
+      try {
+        handler = identitySourceService.getIdentitySourceHandler(req.body.source);
+      } catch (e) {
+        return res.badRequest({ body: e.message });
```
Hailong-am, line 73: If the error comes from identity source, the response code is still 400, is that expected?

Hailong-am (approved):

src/core/server/security/router/index.ts
gaobinlong: For now the only one error here is the source received hasn't been registered so need to return 400, for other error comes from identity source, we return the raw status code directly.

src/plugins/workspace/public/components/workspace_form/workspace_collaborator_table.tsx
gaobinlong: Yeah, there maybe api throttling, in most cases the number of collaborators is not too big so I think this maybe safe.

yubonluo (approved):

PR #12148, opened 2026-06-02
[Chat] Support local-only responses from slash commands

wanglam:
### Description

Extend the slash command infrastructure to support local-only responses that display immediately in the chat without invoking the AI agent.

**Changes:**
- Add `SlashCommandResult` type: handlers can return `string` (sent to agent) or `{ localMessage: string; title?: string }` (displayed locally as a system message)
- Add optional `title` field to `SystemMessage` schema for custom message titles
- `ErrorRow` renders `message.title` when present, falls back to "Something went wrong"
- Export `SlashCommandResult` type from chat plugin public API

This enables slash commands to show informational messages (e.g., workspace incompatibility warnings) with custom titles without involving the AI agent.

### Issues Resolved

N/A - Infrastructure enhancement for slash command system

## Screenshot

N/A - No visual change without consuming plugins

## Testing the changes

1. Register a slash command returning `{ localMessage: "test", title: "Custom Title" }`
2. Type the command in chat
3. Verify the system message shows "Custom Title" instead of "Something went wrong"

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/chat/public/components/chat_window.tsx
+++ b/src/plugins/chat/public/components/chat_window.tsx
@@ -216,13 +216,13 @@ const ChatWindowContent = React.forwardRef<ChatWindowInstance, ChatWindowProps>(
     // Block AI features for unsupported data sources
     if (await isUnsupportedDataSource()) {
       const userMsg: UserMessage = {
-        id: `msg-${Date.now()}-${Math.random().toString(36).substring(2, 11)}`,
+        id: chatService.generateMessageId(),
         role: 'user',
         content: messageContent,
         rawMessage: rawMessage || messageContent,
       };
       const systemMsg: SystemMessage = {
-        id: `msg-${Date.now()}-${Math.random().toString(36).substring(2, 11)}`,
+        id: chatService.generateMessageId(),
         role: 'system',
         content: i18n.translate('chat.dataSourceUnsupported', {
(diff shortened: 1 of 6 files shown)


yubonluo (dismissed):

src/plugins/chat/public/components/chat_window.tsx
```diff
@@ -322,6 +322,21 @@ const ChatWindowContent = React.forwardRef<ChatWindowInstance, ChatWindowProps>(
...
+          rawMessage: messageContent,
+        };
+        const systemMsg: SystemMessage = {
+          id: chatService.generateMessageId(),
+          role: 'system',
+          content: commandResult.localMessage,
+        };
+        setTimeline((prev) => [...prev, userMsg, systemMsg]);
+        return;
+      }
```
Hailong-am, line 340: This is client side message and will lost when reload the session. is this expected?

Hailong-am (dismissed):

src/plugins/chat/public/components/chat_window.tsx
wanglam: Yes. It won't be stored in the agentic memory. It's an expected behavior.

Hailong-am (approved):

yubonluo (approved):

PR #12154, opened 2026-06-02
[Explore] Inline actions in Query Panel toolbar

TackAdam:
### Description

Refactors Explore's Query Panel "Actions" affordance so registered actions render directly in the toolbar instead of being buried behind an "Actions ▾" dropdown. With ≤ 5 actions, all actions are visible inline; with more than 5, the first 5 remain inline and the rest fall back into a "+N more action(s)" overflow popover. The action group is right-anchored within the toolbar's left section, visually separating it from the save, recent, and dataset controls without an explicit divider.

This is a small change affecting the Logs, Metrics, Traces, and Events query panel in Explore. There are no registry API changes, no behavior changes for action authors, and no migration required.

Motivation

Today every registered action — even when there is only one — sits behind a click. With the new "Create monitor" entry registered by dashboards-observability joining the existing alerting plugin entry, the dropdown was hiding important affordances. For a small but growing list of actions, inline display is the more appropriate pattern.

What changed

### src/plugins/explore/public/components/query_panel/query_panel_widgets/query_panel_actions/query_panel_actions.tsx

- Splits registry.getSortedActions() at INLINE_ACTION_LIMIT = 5 (defined as a named constant with rationale).
- First 5 actions render as EuiButtonEmpty size="xs" directly in the toolbar using the existing handleActionClick dispatch path, preserving behavior for both button and flyout action types.
- Remaining actions render inside the existing EuiPopover, triggered by a "+N more action(s) ▾" button with plural-aware i18n messaging (+1 more action, +3 more actions).
- Inline button labels are wrapped in `<EuiText size="xs">` so they visually match neighboring toolbar controls (Saved queries, Recent queries, Save).
- The inline actions and the overflow popover are wrapped in a single right-anchored `<div className="exploreQueryPanelActions__group">` so the entire group hugs the right edge of the toolbar's left section, regardless of count.

Test subject contract preserved:

- Popover trigger retains `data-test-subj="queryPanelFooterActionsButton"` so existing integrations continue to work.
- Inline buttons use `queryPanelActionInline-{action.id}`.
- Overflow buttons use `queryPanelActionOverflow-{action.id}`.
- The right-anchored wrapper uses `queryPanelActionsGroup`.

### query_panel_actions.scss

- Adds `.exploreQueryPanelActions__group` — `display: flex; margin-left: auto; align-items: center; gap: $euiSizeXS;`
- `margin-left: auto` pushes the action group to the far right of the toolbar's left section, creating visual separation without requiring a divider.

### query_panel_widgets.scss

- Adds `flex: 1` to `&__left` so the toolbar's left section consumes remaining horizontal space.
- Without this, `margin-left: auto` inside the left section had no room to push the action group right.
- Existing controls continue rendering at their natural widths and positions.

Behavior matrix

| Registered actions | Toolbar shows |
|-------------------|---------------|
| 0 | Nothing (unchanged — QueryPanelWidgets still gates rendering on registry.isEmpty()) |
| 1–5 | Each action inline as a small button, right-anchored as a group |
| 6+ | First 5 inline + a "+N more action(s) ▾" popover trigger, all right-anchored as a group |

Disabled actions (`getIsEnabled() === false`) render as disabled inline buttons or disabled overflow items, matching existing behavior.

Why these defaults

- INLINE_ACTION_LIMIT = 5: Balances visibility with toolbar space. Too many inline buttons crowd adjacent controls; too few make the overflow menu the common case.
- "+N more action(s)" with plural-aware copy: Communicates both the count and what's being counted in a compact format.
- Right-anchored via `margin-left: auto`: Keeps the existing toolbar structure intact while creating separation through layout alone.
- Shared dispatch path: Inline and overflow rendering are purely presentational. Both use the same handleActionClick implementation, preserving behavior across button and flyout action types.

Tests

query_panel_actions.test.tsx was rewritten to cover both rendering paths:

- Inline rendering: empty state, 3 actions inline, click dispatch behavior, disabled state, dependency propagation to getLabel/getIsEnabled/getIcon.
- Overflow rendering: 6 actions → 5 inline + 1 overflow, popover reveal, click dispatch, plural-aware copy (+1 more action and +3 more actions).

Results:

- 9 tests passing.
- Full query panel suite passing: 23 suites / 241 tests.
- No regressions observed.

How to test locally

1. Ensure at least one plugin registers a queryPanelAction (dashboards-observability and alerting-dashboards-plugin already do).
2. Navigate to Logs in Explore.
3. Verify actions appear directly in the toolbar, right-anchored after Saved queries.
4. To test overflow, temporarily register additional actions and reload. The first 5 should render inline and the remainder should appear under "+N more action(s) ▾".

Backwards compatibility

100% additive. The action registry contract is unchanged, the popover path remains intact (now used only for overflow), and the legacy `data-test-subj="queryPanelFooterActionsButton"` is preserved. Existing plugins continue to work without modification.

The `flex: 1` added to `__left` does not change the rendering of any existing widget. Flex children continue to render left-aligned by default, so the existing toolbar controls remain anchored at the left edge.

## Screenshot
Before:
Before

After:
After1

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ps48:
Can we drop the `Actions:` there's no need for it in the expanded UI. For more than 5 stuff we can keep text like `more actions`?

ps48 (commented):
#### Diff Assessment

Clean refactor of Explore's Query Panel Actions from a dropdown-only pattern to inline display (≤5 inline, overflow into popover). 4 files changed, well-decomposed with proper OUI usage, stable `data-test-subj` contracts preserved, plural-aware i18n, and comprehensive test rewrite (9 tests covering both render paths). No API changes, fully backwards-compatible.

---

#### No Blocking, High, or Medium Issues

This PR is architecturally sound and ready to merge.

---

#### LOW Priority

**L1.** `EuiListGroup` wrapping `EuiButtonEmpty` items (`query_panel_actions.tsx` ~line 145) — semantically `EuiListGroup` expects `EuiListGroupItem` children. Works visually but could produce accessibility warnings. Consider a plain `div` or switching to `EuiListGroupItem`.

**L2.** `className="exploreQueryPanelActions__inlineButton"` is referenced in the component but not styled in the SCSS — it's a no-op class. Either add styles or remove the className.

---

#### Positive Highlights

- **Fully backwards-compatible** — registry API unchanged, `data-test-subj="queryPanelFooterActionsButton"` preserved
- **Proper `useMemo` on action split** — stable identity unless registry changes
- **Shared `handleActionClick` dispatch** — inline and overflow are purely presentational
- **Plural-aware i18n** — correct ICU `{count, plural, one {action} other {actions}}`
- **`margin-left: auto` layout** — clean separation without extra wrappers, well-documented in SCSS
- **9 tests covering both paths** — empty, inline (3), click dispatch, disabled, deps propagation, overflow boundary (6), popover reveal, overflow click, plural copy
- **Well-documented `INLINE_ACTION_LIMIT = 5`** with rationale
- **Clean 175-line component** — single responsibility, no `any`, no console statements

TackAdam:
**L1 — `EuiListGroup` semantics:** Replaced the popover's `EuiListGroup` wrapper with a plain `<div
  className="exploreQueryPanelActions__overflowList">`. EuiListGroup expects EuiListGroupItem children; we use EuiButtonEmpty for consistent click/icon/disabled handling with the inline buttons, so the list semantics weren't pulling weight. Added a small flex-column SCSS rule to replicate the visual stacking. Dropped the now-unused `EuiListGroup` import.

  **L2 — dead className:** Removed `className="exploreQueryPanelActions__inlineButton"` from the inline button — it was never styled. The right-anchor + gap behavior lives on the parent `__group` class, which still does its job.

  All 9 query_panel_actions tests still pass, lint clean.

ps48 (approved):

Maosaic (approved):

Issue #9851, opened 2025-06-05
Allow queuing of searches when loading a dashboard.


alamzeeshan:
**Is your feature request related to a problem? Please describe.**

When we open a dashboards with lots of panel each panel triggers a search request. If my dashboard is having 50 panels then 50 concurrent search request will be sent to opensearch and sometime if the time duration selected is large then we started seeing circuit breaker errors in opensearch and many of my panels fails to load. 

**Describe the solution you'd like**

I would like to have a queuing mechanism which allows me to limit the concurrent search request coming from a dashboard or a user so that all my search request will be processed one after another or in small groups so that I will get all my search results at the end. I would prefer waiting a bit and getting all the results than getting partial results. 

**Describe alternatives you've considered**

There is no alternative as of now. 

**Additional context**

PR #12163, opened 2026-06-04
[Dashboard] Lazy load panel data queries based on viewport visibility

Qxisylolo:
### Description
Support dashboard lazy loading for 

- Defer data fetch (PPL/DSL /PromQl queries) for dashboard panels until they scroll into the viewport 
- Panels above the fold load immediately; panels below the fold show a loading indicator until visible
- Reduces concurrent query pressure on OpenSearch clusters for dashboards with many visualizations

## Test plan
- [ ] Open a dashboard with 10+ panels where some are below the fold
- [ ] Verify above-fold panels load immediately
- [ ] Verify below-fold panels show loading spinner and only fetch data when scrolled into view
- [ ] Check Network tab: PPL/DSL requests should fire incrementally as panels scroll into view, not all at once

## Screenshot

https://github.com/user-attachments/assets/7a0beee1-826c-4836-9d8a-0940162c2fc3

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/13/inspect.spec.js
+++ b/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/explore/13/inspect.spec.js
@@ -124,6 +124,8 @@ const inspectTestSuite = () => {
         'dashboardListingTitleLink-[Flights]-Global-Flight-Dashboard'
       ).click();
 
+      cy.scrollTo('bottom', { duration: 3000 });
+
       cy.getElementByTestId('visualizationLoader').should(
         'have.length',
         NUMBER_OF_VISUALIZATIONS_IN_FLIGHTS_DASHBOARD
--- a/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/13/inspect.spec.js
+++ b/cypress/integration/core_opensearch_dashboards/opensearch_dashboards/apps/query_enhancements/13/inspect.spec.js
@@ -130,6 +130,8 @@ const inspectTestSuite = () => {
         'dashboardListingTitleLink-[Flights]-Global-Flight-Dashboard'
       ).click();
 
+      cy.scrollTo('bottom', { duration: 3000 });
+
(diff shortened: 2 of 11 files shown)


yubonluo:
1. Race condition: this.mounted set after observer starts observing
  
```
  // componentDidMount
  const el = this.panelRef.current;
  if (el) {
    this.observer = new IntersectionObserver(([entry]) => {
      if (entry.isIntersecting && this.mounted) {  // ← reads mounted
        ...
      }
    }, { rootMargin: '200px' });
    this.observer.observe(el);
  }
  this.mounted = true;  // ← set AFTER observe()
```
  
  The IntersectionObserver callback fires asynchronously, so in practice
  this.mounted will be true by then. However, the ordering is fragile. Move
  this.mounted = true before this.observer.observe(el) to make the invariant
  explicit.
  
  
  2. Panels above the fold still show a brief loading spinner
  
  Because the IntersectionObserver callback is async (even if the element is
  already visible), there's at least one render cycle where visible: false and
  the loading spinner shows. For above-the-fold panels, consider initializing
  visible: true if the element is already intersecting at mount time, or use {
  threshold: 0 } with an immediate check:
  
```
  // After observe(), check if already in viewport:
  const rect = el.getBoundingClientRect();
  if (rect.top < window.innerHeight + 200) {
    this.observer.disconnect();
    this.setState({ visible: true });
  }
  ```

ruanyl:
Here is one related issue: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/9851

ruanyl (dismissed):

Qxisylolo:
Hi @yubonluo , thanks for the review

> 1. Race condition: this.mounted set after observer starts observing (already fixed)

> 2. Panels above the fold still show a brief loading spinner
For the brief loading spinner, here are two related callbacks, 1. IntersectionObserver, 2, untilEmbeddableLoaded. so By the time the embeddable resolves, the spinner is always loading. untilEmbeddableLoaded mostly takes longer since it requires an async fetch.

yubonluo (approved):

ruanyl (approved):

Issue #2712, opened 2022-11-01
[BUG] Dashboard fails to connect when using securityconfig


jonathon2nd:
**Describe the bug**

Dashboard does not connect when trying to set and use different passwords.

**To Reproduce**
Deploy cluster from operator.

`securityconfig-secret` is taken from https://github.com/Opster/opensearch-k8s-operator/blob/main/opensearch-operator/examples/opensearch-cluster-securityconfig.yaml
I have rehashed passwords by using the `hash.sh` I found on the opensearch pod, as getting it working locally failed.

`admin-credentials-secret`
```
---
apiVersion: v1
kind: Secret
metadata:
  name: admin-credentials-secret
  namespace: labs-opensearch
type: Opaque
stringData:
  username: admin
  password: PASSWORD
```

Values for cluster
```
---
apiVersion: opensearch.opster.io/v1
kind: OpenSearchCluster
metadata:
  name: labs-cluster
  namespace: labs-opensearch
spec:
  security:
    tls:  # Everything related to TLS configuration
      transport:  # Configuration of the transport endpoint
        generate: true  # Have the operator generate and sign certificates
        perNode: true  # Separate certificate per node
      http:  # Configuration of the HTTP endpoint
        generate: true  # Have the Operator generate and sign certificates
    config:
      securityConfigSecret:
        name: securityconfig-secret
      adminCredentialsSecret:
        name: admin-credentials-secret
  general:
    serviceName: labs-cluster
    version: 1.3.1
    setVMMaxMapCount: true
  dashboards:
    enable: true
    additionalConfig:
      opensearch_security.auth.type: "proxy"
      opensearch.requestHeadersWhitelist: |
        ["securitytenant","Authorization","x-forwarded-for","x-auth-request-access-token", "x-auth-request-email", "x-auth-request-groups"]
      opensearch_security.multitenancy.enabled: "true"
      opensearch.username: "admin"
      opensearch.password: "PASSWORD"
    tls:
      enable: true  # Configure TLS
      generate: true  # Have the Operator generate and sign a certificate
    version: 1.3.1
    replicas: 1
    resources:
      requests:
         memory: "512Mi"
         cpu: "200m"
      limits:
         memory: "512Mi"
         cpu: "200m"
  nodePools:
    - component: masters
      replicas: 3
      diskSize: "50Gi"
      jvm: -Xmx1024M -Xms1024M #Set to half of memory request
      persistence:
        pvc:
[...]

PR #12168, opened 2026-06-04
[Explore] Add Create alert rule entry point on Metrics page

riysaxen-amzn:
### Description

Adds a "Create alert rule" entry point at the top of the Explore Metrics page. When clicked, a flyout opens pre-populated with all current PromQL queries from the query panel, allowing users to configure thresholds and create alerting rules without leaving the Metrics exploration flow.

### Motivation

Users exploring metrics often discover patterns or anomalies they want to be alerted about. Previously, they had to manually copy PromQL expressions to the Alerting page and recreate the query context. This PR eliminates that friction — one click from exploring a metric to creating an alert rule for it.

This follows the same pattern as PR #2712 which adds a similar entry point from the Explore Logs page.

### What's new

- **"Create alert rule" button** (🔔) in the Metrics page top toolbar, visible in both Explore and Query tabs
- **`create_metrics_rule_flyout.tsx`** — self-contained flyout component that:
  - Reads all current PromQL queries from the QueryStringManager (no "Update" click required)
  - Shows each query as a configurable rule card with auto-derived name, operator, threshold, and for-duration
  - Shared evaluation interval setting at the top
  - Creates rules via `POST /api/alerting/prometheus/{dsId}/rules` (handled by the observability plugin)
  - Toast informs user about Cortex propagation delay

### User flow

1. User explores metrics on the Metrics page (Builder or Code mode)
2. Adds one or more PromQL queries
3. Clicks **"Create alert rule"** in the top toolbar
4. Flyout opens with all queries pre-populated as rule cards
5. Configures thresholds per rule
6. Clicks **"Create N rule(s)"**
7. Rules are created in Cortex and appear on the Alerting → Rules tab

### Companion PR

- dashboards-observability: https://github.com/opensearch-project/dashboards-observability/pull/2718

### Testing

- Add queries on Metrics page → click "Create alert rule" → verify flyout shows all queries
- Change queries → reopen flyout → verify it reflects latest queries (no Update click needed)
- Save rules → navigate to Alerting → Rules tab → verify rules appear after ~60s
- Verify alerts fire when metric conditions are met
- Unit tests included for flyout rendering, save behavior, and edge cases

https://github.com/user-attachments/assets/1e461e7b-ab8b-404b-b4d2-76ee0f03dc3f

### Check List

- [x] All tests pass


src/plugins/explore/public/application/pages/metrics/create_metrics_rule_flyout.tsx
```diff
@@ -0,0 +1,339 @@
...
+  // Try to extract the innermost metric name, skipping PromQL function wrappers
+  // e.g., rate(http_requests_total[5m]) → http_requests_total
+  const innerMatch = query.match(/\(\s*([a-zA-Z_:][a-zA-Z0-9_:]*)/);
+  if (innerMatch) return innerMatch[1];
+  // Fallback: first identifier that's followed by { ( [ or whitespace
+  const match = query.match(/\b([a-zA-Z_:][a-zA-Z0-9_:]*)\s*[{([\s]/);
+  if (match) return match[1];
+  const simpleMatch = query.match(/^([a-zA-Z_:][a-zA-Z0-9_:]*)/);
+  if (simpleMatch) return simpleMatch[1];
+  return 'metrics_alert';
```
joshuali925, line 88: can there be some unit test for the regex patterns? what input / output do we expect and how reliable are they?

riysaxen-amzn: Good call — added a dedicated `deriveRuleName` test block with parameterized cases covering:

- Function-wrapped metrics: `rate(http_requests_total[5m])` → `http_requests_total`
- Nested functions: `histogram_quantile(0.99, rate(http_duration_seconds_bucket[5m]))` → `http_duration_seconds_bucket`  
- Simple metrics: `up` → `up`, `node_load1` → `node_load1`
- Label selectors: `http_requests_total{method="GET"}` → `http_requests_total`
- Fallback: `123 + 456` → `metrics_alert`

The strategy is: look inside the first `(` for a metric name (handles function wrappers), fall back to the first identifier in the string, and default to `'metrics_alert'` if nothing matches. It's best-effort name derivation — the user can always edit the rule name in the flyout before saving.

joshuali925:
### Code review

Found 5 issues (confidence ~75 each — non-blocking but worth addressing before merge):

1. **`addToast` prop is never wired at the call site, so success/error toasts are silent in production.** The flyout declares `addToast` as optional and gates every notification with `addToast?.(...)`. The component is rendered without that prop, so users get zero feedback on success or failure. The matching toast assertions in the test only pass because the test passes a mock.

   - Use site (no `addToast` passed): https://github.com/riysaxen-amzn/OpenSearch-Dashboards/blob/6678d275a4fe68a99ae29fc032311edf46d72ddb/src/plugins/explore/public/application/pages/metrics/metrics_page_tabs.tsx#L114-L122
   - Optional-chained calls in flyout: https://github.com/riysaxen-amzn/OpenSearch-Dashboards/blob/6678d275a4fe68a99ae29fc032311edf46d72ddb/src/plugins/explore/public/application/pages/metrics/create_metrics_rule_flyout.tsx#L165-L191

   Suggested: forward `services.notifications.toasts.addSuccess` / `addDanger` from `metrics_page_tabs.tsx`, or call them directly inside the flyout via context.

2. **Bespoke multi-query splitter misses common patterns and contradicts the canonical helper used elsewhere on this same page.** The hand-rolled split only triggers on `;\n` or trailing `;`, so `up;node_cpu` (no newline) is treated as one query, and the regex is unaware of semicolons inside PromQL label values like `up{job="my;app"}`. The canonical [`splitMultiQueries`](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/0920c73800/src/plugins/data/common/query/multi_query.ts#L138) handles both correctly and is already used by [`row_state.ts`](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/0920c73800/src/plugins/explore/public/application/pages/metrics/query_panel/row_state.ts#L31) on the same Metrics page.

   - https://github.com/riysaxen-amzn/OpenSearch-Dashboards/blob/6678d275a4fe68a99ae29fc032311edf46d72ddb/src/plugins/explore/public/application/pages/metrics/metrics_page_tabs.tsx#L51-L67

3. **`Number.parseFloat(value) || 0` blocks decimals and negative thresholds in the controlled `EuiFieldNumber`.** Every intermediate keystroke that doesn't parse to a number (`-`, `.`, `1.`, empty string) is coerced to `0` and the field re-renders as `'0'`, dropping the keystroke. Users cannot type `1.5` or `-5` incrementally. Guard on `Number.isNaN` (preserve previous value or accept a string state).

   - https://github.com/riysaxen-amzn/OpenSearch-Dashboards/blob/6678d275a4fe68a99ae29fc032311edf46d72ddb/src/plugins/explore/public/application/pages/metrics/create_metrics_rule_flyout.tsx#L270-L278

4. **`parsedQueries` `useMemo` only re-runs on flyout open/close — the comment claiming live-keystroke sync is misleading and the eslint-disable hides a stale snapshot.** The dep array is `[showAlertRuleFlyout]`, so any query edited *after* the flyout opens is not reflected in the rule list. Either subscribe to `queryString.getUpdates$()` or read inside `handleSave` instead of memoizing.

   - https://github.com/riysaxen-amzn/OpenSearch-Dashboards/blob/6678d275a4fe68a99ae29fc032311edf46d72ddb/src/plugins/explore/public/application/pages/metrics/metrics_page_tabs.tsx#L51-L67

5. **`services.http as any` cast violates [CLAUDE.md](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/0920c73800/CLAUDE.md) ("Avoid `any` (use `unknown` or generics)"). `CoreStart['http']` already structurally satisfies the flyout's narrow `{ post, get }` prop type, so the cast is unnecessary. Drop the cast or widen the prop to `CoreStart['http']`.

   - https://github.com/riysaxen-amzn/OpenSearch-Dashboards/blob/6678d275a4fe68a99ae29fc032311edf46d72ddb/src/plugins/explore/public/application/pages/metrics/metrics_page_tabs.tsx#L117-L122

riysaxen-amzn:
> **AI-powered '_Code-Diff-Analyzer_' found issues on commit [9f96f3b](https://github.com/opensearch-project/OpenSearch-Dashboards/commit/9f96f3be5173b723a5b73b09c1771d831be0a9d5).**
> 
> Path	Line	Severity	Description
> src/plugins/explore/public/application/pages/metrics/create_metrics_rule_flyout.tsx	168	low	User-controlled rule.name is interpolated directly into the annotations.summary field (`Alert: ${rule.name}`) before being sent to the backend API. If the backend reflects this value into rendered output without sanitization it could be a stored XSS vector, though the risk depends entirely on how the backend renders annotations.

This is a false positive for our use case. The `annotations.summary` field is stored in Cortex as rule metadata and rendered in the OSD UI via React's `{value}` JSX interpolation, which auto-escapes HTML entities. There's no `dangerouslySetInnerHTML` or raw HTML rendering path for annotation values in our codebase.

Additionally, the `rule.name` is constrained by the server-side schema (`schema.string({ minLength: 1, maxLength: 256 })`) and the Cortex ruler itself validates rule names. The annotation value is purely informational metadata displayed in the Alerting UI as plain text.

joshuali925 (dismissed):

src/plugins/explore/public/application/pages/metrics/create_metrics_rule_flyout.tsx
```diff
@@ -0,0 +1,306 @@
...
+      for (const rule of rules) {
+        const payload = {
+          name: rule.name,
+          query: rule.query,
+          operator: rule.operator,
+          threshold: rule.threshold,
+          forDuration: rule.forDuration,
+          evaluationInterval,
+          labels: {},
+          annotations: { summary: `Alert: ${rule.name}` },
```
lezzago, line 142: **High: Duplicate rule names silently overwrite each other**

When multiple queries derive the same metric name (e.g., `rate(http_requests_total{method="GET"}[5m])` and `rate(http_requests_total{method="POST"}[5m])` both derive `"http_requests_total"`), both POSTs use the same `groupName`. Since Cortex upsert replaces the entire group, the second POST silently overwrites the first. The user sees "2 alert rule(s) created successfully" but only 1 survives.

This affects Query tab users who write 2+ semicolon-separated queries targeting the same metric.

**Suggested fix** — deduplicate names before save:

```typescript
const seenNames = new Map<string, number>();
const dedupedRules = rules.map((r) => {
  const count = seenNames.get(r.name) || 0;
  seenNames.set(r.name, count + 1);
  return count > 0 ? { ...r, name: `${r.name}-${count + 1}` } : r;
});
```

Alternatively, validate uniqueness in the UI and show a warning if names collide.

src/plugins/explore/public/application/pages/metrics/create_metrics_rule_flyout.tsx
```diff
@@ -0,0 +1,306 @@
...
+        'success'
+      );
+      onClose();
+    } catch {
+      setSaveStatus('idle');
+      addToast?.(
+        i18n.translate('explore.metricsRule.toast.failed', {
+          defaultMessage: 'Failed to create alert rules',
+        }),
+        'danger'
```
lezzago, line 168: **Medium: Flyout closeable during active save**

`EuiFlyout onClose={onClose}` allows Escape or X-click while POSTs are in flight. This unmounts the component mid-async-operation — the user loses feedback on which rules were created.

**Suggested fix:**

```typescript
<EuiFlyout onClose={isSaving ? undefined : onClose} ...>
```

Or use `hideCloseButton` + disable Cancel button during save.

joshuali925 (approved):

lezzago (approved):

PR #12183, opened 2026-06-08
[Explore] Fix date picker filtering for SQL visualizations

ritvibhatt:
### Description

- Update SQL visualizations on dashboards to respect time filter in the date picker component
- Before SQL visualizations would stay the same no matter what values were chosen on the date picker, now the visualizations have a time filter added based on the date picker

## Screenshot
[](https://github.com/user-attachments/assets/ab881769-e1bb-448d-810a-f948df96a93b)

## Testing the changes

- Start Dashboards with explore enabled
- Add ```explore.sqlSupport.endabled: true``` to opensearch_dashboards.yml
- Create a SQL visualization through explore and add to dashboard
- change time picker on dashboard

### Check List

- [x] All tests pass
  - [ ] `yarn test:jest`


joshuali925 (approved):

joshuali925:
### Code review

Found 1 issue:

1. The new time-filter allowlist drops time filtering for the standalone SQL Visualization Editor, which contradicts this PR's goal. The PR replaces the per-request `enableTimeFiltering` flag (which `createSearchSourceWithQuery` set for *all* SQL queries regardless of app) with an app-ID allowlist of `['dashboards', 'explore/logs']`. But the SQL Visualization Editor app (`VISUALIZATION_EDITOR_APP_ID = 'visualization-editor'`) also runs SQL through `createSearchSourceWithQuery` → `searchSource.fetch()` → this interceptor, and it renders a date picker for time-based datasets. Since `'visualization-editor'` is not in the allowlist, SQL previews in the editor will now ignore the date picker and return all-time data, while the same saved viz *is* time-filtered on the dashboard. (SQL is a raw string, so the search-source DSL filter doesn't constrain results — only this WHERE-clause injection does.) Consider adding `'visualization-editor'` to `timeFilterSupportedAppNames`.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a59bc101234878001c9418cf36de07c91c872f4e/src/plugins/query_enhancements/public/search/sql_search_interceptor.ts#L31-L33

The editor executes SQL through the same search source / interceptor path:

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a59bc101234878001c9418cf36de07c91c872f4e/src/plugins/explore/public/application/in_context_vis_editor/query_builder/utils.ts#L220-L226

TackAdam (approved):

ritvibhatt:
visualization-editor only supports ppl and promql for supported language so don't need to add to time filter support

Issue #11996, opened 2026-05-19
[BUG] Workspace API Permissions bug


oli1234554:
**Describe the bug**

Not able to create read only groups or users as collaborators via workspace api. 

**To Reproduce**
use the offical curl from the documentation:
`curl -k -u admin:admin \
  -H 'osd-xsrf: true' \
  -H 'Content-Type: application/json' \
  -X PUT 'https://localhost:5601/api/workspaces/B9Le1w' \
  -d '{
    "attributes": {},
    "settings": {
      "permissions": {
        "library_write": { "groups": ["obs-admins"] },
        "read": { "groups": ["obs-users"] }
      }
    }
  }'`

Nothing will get created.
You can create read_write like this for example:
`curl -k -u admin:admin \
  -H 'osd-xsrf: true' \
  -H 'Content-Type: application/json' \
  -X PUT 'https://localhost:5601/api/workspaces/B9Le1w' \
  -d '{
    "attributes": {},
    "settings": {
      "permissions": {
        "library_write": { "groups": ["obs-admins"] },
        "read": { "groups": ["obs-admins"] }
      }
    }
  }'`
this will create it but you cant add read only user or group!
**Expected behavior**

create the collabortation correctly. If i have a group in "read" permission then add this group with read permissions. 

**OpenSearch Version**
Please list the version of OpenSearch being used.
3.5 / 3.6

**Dashboards Version**
Please list the version of OpenSearch Dashboards being used.
3.5/3.6

**Plugins**

Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

oli1234554:
ok I think it maybe work but on gui you cant see the read only things you did over api

xluo-aws:
@yubonluo , @SuZhou-Joe , could any of you help to take a look? Could be an issue that doc is not matching code? To get read access it requires both library_read and read, right?

andrross:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/andrross)

yubonluo:
Thanks for the report, @oli1234554.

**Root cause**

A workspace collaborator role isn't a single permission mode — it's a *combination* of one workspace-level mode and one library-level mode:

| Access level | Required modes |
| --- | --- |
| Read only | `library_read` + `read` |
[...]

PR #12186, opened 2026-06-09
[Fix][Workspace] Validate collaborator permission combinations and reject invalid ones in workspace APIs

yubonluo:
### Description

The workspace create (`POST /api/workspaces`) and update (`PUT /api/workspaces/{id}`) APIs accepted any `permissions` map and persisted it verbatim. A workspace collaborator role, however, is defined by a *combination* of one workspace-level mode and one library-level mode:

| Access level   | Required modes            |
| -------------- | ------------------------- |
| Read only      | `library_read` + `read`   |
| Read and write | `library_write` + `read`  |
| Admin          | `library_write` + `write` |

- `read` / `write` apply to the **workspace object** (open it / manage it).
- `library_read` / `library_write` apply to the **saved objects inside it**.

When a principal was granted an incomplete combination (e.g. only `read`, or only `library_write`), the value was stored but did not form a usable role: it was silently dropped by the collaborators UI and granted no coherent access (for example, a `write`-only user can open the workspace shell but cannot read any of its saved objects and never appears as a collaborator). This is the cause of #11996 — and the official docs example used exactly such an incomplete combination.

This PR makes the workspace create/update APIs validate permissions:

- **Reject invalid combinations.** When the supplied `permissions` contain a principal whose mode combination does not map to a recognized access level, the request fails with `400 Bad Request` and an error message naming the offending principal and listing the valid combinations. This prevents the silent, unusable state described above.
- **Normalize redundant combinations.** A principal granted a superset of a role (e.g. all four modes) is collapsed to the single highest matching access level before persisting (e.g. `library_write` + `write` for admin). This keeps the stored ACL canonical and prevents the collaborators UI from rendering the same principal once per matching access level.

### Screenshot

Example response when an incomplete combination is supplied:
```
curl ...
  --data-raw '{"attributes":{},"settings":{"permissions":{"library_write":{"users":["admin", "wrongPermission1"]},"write":{"users":["admin"]},"library_read":{"users":["wrongPermission1"]},"read":{"users":["wrongPermission2"]}}}}'

{"statusCode":400,"error":"Bad Request","message":"Invalid workspace permissions for: user \"wrongPermission1\" ([library_write, library_read]); user \"wrongPermission2\" ([read]). Each user or group must be granted one of the following permission mode combinations: [\"library_read\", \"read\"] for read only, [\"library_write\", \"read\"] for read and write, or [\"library_write\", \"write\"] for admin."}% 
```

### Issues Resolved
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11996

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/server/routes/index.ts
+++ b/src/plugins/workspace/server/routes/index.ts
@@ -22,7 +22,7 @@ import { registerDuplicateRoute } from './duplicate';
 import { getPermissionMode, transferCurrentUserInPermissions } from '../utils';
 import {
   validateWorkspaceColor,
-  getWorkspacePermissionWarning,
+  getInvalidWorkspacePermissionsError,
   normalizeWorkspacePermissions,
 } from '../../common/utils';
 import { getUseCaseFeatureConfig } from '../../../../core/server';
@@ -226,14 +226,13 @@ export function registerRoutes({
         dataConnections?: string[];
       } = attributes;
 
-      // Non-blocking warning for permission combinations that do not map to a
-      // recognized collaborator access level. The request still succeeds to keep
-      // the API backward compatible.
-      const permissionWarning = isPermissionControlEnabled
-        ? getWorkspacePermissionWarning(settings.permissions)
-        : undefined;
-
+      // Reject permission combinations that do not map to a recognized
+      // collaborator access level (read only, read and write, or admin).
       if (isPermissionControlEnabled) {
+        const invalidPermissionsError = getInvalidWorkspacePermissionsError(settings.permissions);
+        if (invalidPermissionsError) {
+          return res.badRequest({ body: invalidPermissionsError });
+        }
         const normalizedPermissions = normalizeWorkspacePermissions(settings.permissions);
         createPayload.permissions = normalizedPermissions;
         if (!!principals?.users?.length) {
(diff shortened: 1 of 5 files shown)


SuZhou-Joe (dismissed):

SuZhou-Joe (commented):
Workspace IT failed, may need some fix.

yubonluo:
> Workspace IT failed, may need some fix.

Thanks, all of the IT passed.

src/plugins/workspace/server/routes/index.ts
```diff
@@ -222,12 +226,19 @@ export function registerRoutes({
         dataConnections?: string[];
       } = attributes;
 
+      // Reject permission combinations that do not map to a recognized
+      // collaborator access level (read only, read and write, or admin).
       if (isPermissionControlEnabled) {
-        createPayload.permissions = settings.permissions;
+        const invalidPermissionsError = getInvalidWorkspacePermissionsError(settings.permissions);
```
Hailong-am, line 232: nit: we can put the parameters validation logic into body schema 
```
body: schema.object({
          attributes: createWorkspaceAttributesSchema,
          settings: settingsSchema,
        }),
```

Hailong-am (approved):

SuZhou-Joe (approved):

PR #12194, opened 2026-06-09
Fix agent traces trace map latency bar always showing 50%

### LINKED ISSUE #2720: [BUG] agent traces trace map bar not showing correct latency
**What is the bug?**

in flyout trace map tab of agent traces, regardless of what latency is, the bar is always in the center

Image

**How can one reproduce the bug?**
go to https://observability.playground.opensearch.org/w/8xYRJ9/app/agentTraces/traces/ and click any trace, click trace map in flyout

**What is the expected behavior?**
bar should fill on where the latency is, like the gantt timeline chart

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
@joshuali925-osdbot take a look

### LINKED ISSUE #12196: [BUG] Agent traces trace map bar not showing correct latency
[**What is the bug?**](https://github.com/opensearch-project/dashboards-observability/issues/2720)

In the flyout trace map tab of agent traces, regardless of what latency is, the bar is always at 50%.

The root cause is in `packages/osd-apm-topology/src/components/nodes/agent_card_node.tsx` where `MetricBar` uses `max={data.duration! * 2}`, making the bar fill always `duration / (duration * 2) = 0.5`.

**How can one reproduce the bug?**
1. Go to agent traces and click any trace
2. Click "Trace Map" in the flyout
3. Observe all span latency bars are at 50% regardless of actual duration

**What is the expected behavior?**
Bar should fill proportionally based on the span duration relative to the root span duration (matching the gantt timeline chart).

**Screenshots**

Image

**Fix**
PR: https://github.com/opensearch-project/OpenSearch-Dashboards/pull/12194

joshuali925-osdbot:
### Description
Fix the agent traces flyout trace map tab where the latency bar for each span always shows at 50% regardless of actual duration, and add Gantt-style offset positioning.

**Root cause:** The `MetricBar` in `AgentCardNode` used `max={data.duration * 2}`, making the bar fill always `duration / (duration * 2) = 50%`.

**Fix:**
1. Pass the root span's duration as `maxDuration` through the flow transform, so each span's bar correctly shows its duration relative to the total trace duration.
2. Add an `offset` prop to `MetricBar` that pushes the bar fill to the right based on span start time relative to the root span (Gantt-style positioning).

### Issues Resolved
Resolves https://github.com/opensearch-project/dashboards-observability/issues/2720
Resolves https://github.com/opensearch-project/OpenSearch-Dashboards/issues/12196

### Screenshots

**Before (bug):** All bars at 50% regardless of actual span duration
Before - all bars at 50%

**After (fix):** Bars proportional to root span duration with Gantt-style offset (bar position shows when span starts, bar width shows duration)
After - Gantt-style bars with offset

### Testing
- Unit tests pass for `flow_transform.test.ts` (12/12 including new offset test)
- Unit tests pass for `metric_bar.test.tsx` (9/9 including new offset tests)
- Unit tests pass for `agent_card_node.test.tsx` (12/12)
- Lint and i18n checks pass

### Changes
- `packages/osd-apm-topology/src/shared/types/agent.types.ts`: Add `maxDuration` and `startOffset` fields to `AgentNodeData`
- `packages/osd-apm-topology/src/components/nodes/metric_bar.tsx`: Add `offset` prop for Gantt-style left margin positioning
- `packages/osd-apm-topology/src/components/nodes/agent_card_node.tsx`: Use `maxDuration` for bar max and pass `startOffset` as offset
- `src/plugins/agent_traces/public/services/flow_transform.ts`: Compute root span duration/start time, pass `maxDuration` and `startOffset` to all nodes
- `src/plugins/agent_traces/public/services/flow_transform.test.ts`: Add test for offset computation
- `packages/osd-apm-topology/src/components/nodes/metric_bar.test.tsx`: Add tests for offset prop

### DIFF
--- a/packages/osd-apm-topology/src/components/nodes/agent_card_node.tsx
+++ b/packages/osd-apm-topology/src/components/nodes/agent_card_node.tsx
@@ -85,7 +85,7 @@ export const AgentCardNode = ({ data }: NodeProps<AgentCardCustomNode>) => {
         {showStandaloneDuration && (
           <MetricBar
             value={data.duration!}
-            max={data.duration! * 2}
+            max={data.maxDuration ?? data.duration!}
             color={kindConfig.color}
             label={data.latency ?? `${data.duration}ms`}
           />
--- a/packages/osd-apm-topology/src/components/nodes/agent_card_node.tsx
+++ b/packages/osd-apm-topology/src/components/nodes/agent_card_node.tsx
@@ -86,6 +86,7 @@ export const AgentCardNode = ({ data }: NodeProps<AgentCardCustomNode>) => {
           <MetricBar
             value={data.duration!}
             max={data.maxDuration ?? data.duration!}
+            offset={data.startOffset}
             color={kindConfig.color}
             label={data.latency ?? `${data.duration}ms`}
           />
--- a/packages/osd-apm-topology/src/components/nodes/agent_card_node.tsx
+++ b/packages/osd-apm-topology/src/components/nodes/agent_card_node.tsx
@@ -86,7 +86,6 @@ export const AgentCardNode = ({ data }: NodeProps<AgentCardCustomNode>) => {
           <MetricBar
             value={data.duration!}
             max={data.maxDuration ?? data.duration!}
-            offset={data.startOffset}
             color={kindConfig.color}
             label={data.latency ?? `${data.duration}ms`}
           />
--- a/packages/osd-apm-topology/src/components/nodes/metric_bar.test.tsx
+++ b/packages/osd-apm-topology/src/components/nodes/metric_bar.test.tsx
@@ -36,6 +36,19 @@ describe('MetricBar', () => {
     const labelSpans = container.querySelectorAll('.osd\\:text-right');
     expect(labelSpans).toHaveLength(0);
   });
+
+  it('applies marginLeft when offset is provided', () => {
+    const { container } = render(<MetricBar value={50} max={100} offset={25} />);
+    const fillBar = container.querySelector('.osd\\:h-full') as HTMLElement;
+    expect(fillBar.style.marginLeft).toBe('25%');
+    expect(fillBar.style.width).toBe('50%');
+  });
+
+  it('defaults offset to 0 when not provided', () => {
+    const { container } = render(<MetricBar value={50} max={100} />);
+    const fillBar = container.querySelector('.osd\\:h-full') as HTMLElement;
+    expect(fillBar.style.marginLeft).toBe('0%');
+  });
 });
 
 describe('MetricBarGroup', () => {
--- a/packages/osd-apm-topology/src/components/nodes/metric_bar.test.tsx
+++ b/packages/osd-apm-topology/src/components/nodes/metric_bar.test.tsx
@@ -49,6 +49,21 @@ describe('MetricBar', () => {
     const fillBar = container.querySelector('.osd\\:h-full') as HTMLElement;
     expect(fillBar.style.marginLeft).toBe('0%');
   });
+
+  it('applies a minimum width so tiny nonzero values stay visible', () => {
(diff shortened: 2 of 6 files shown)


ps48:
@joshuali925-osdbot In PR #12194 (agent traces trace map latency bar), keep the core fix but change how
the bar conveys time, and add a minimum visible bar width.

Background: The agent card latency bar now does two things:
1. Scales the bar width to the root trace duration (max = maxDuration). This is the
   real fix for the "always 50%" bug and should stay.
2. Adds a Gantt-style horizontal offset (startOffset -> MetricBar marginLeft) so the
   bar is positioned along an implied timeline.

Problem with (2): these bars live inside cards in a node graph (react-flow topology),
where cards are positioned by graph layout, not by time, and there is no shared or
visible time axis and no alignment between one card's bar and another's. A
left-offset bar there is floating with no reference, so users cannot read the gap as
"started later in the trace." It looks like ambiguous empty space or a half-empty
progress bar. It also has bad edge cases: a short-but-late span becomes a tiny bar
pushed right (reads as "low"), and a late-and-long span overflows and gets clipped by
the track's overflow-hidden, silently dropping information. A Gantt offset only reads
when bars share a visible axis and aligned rows (a waterfall), which this view is not.

Change 1: drop the Gantt offset, keep the magnitude scaling
- Keep the maxDuration scaling: width = spanDuration / rootDuration, left-aligned.
- Drop the Gantt offset so the latency bar is a plain relative magnitude bar
  ("this step took X% of the whole trace"). The latency label already carries the
  absolute value, so bar length plus label is enough.
- packages/osd-apm-topology/src/components/nodes/agent_card_node.tsx: remove the
  offset={data.startOffset} prop from the MetricBar (keep
  max={data.maxDuration ?? data.duration!}).
- src/plugins/agent_traces/public/services/flow_transform.ts: stop computing and
  setting startOffset on the node data (keep maxDuration). Also drop the now-unused
  parseTimestampMs import and rootStartMs.
- packages/osd-apm-topology/src/shared/types/agent.types.ts: remove the startOffset
  field (keep maxDuration).
- MetricBar's generic `offset` prop can stay (default 0, harmless) for future
  waterfall use, or be removed if you prefer. If kept, keep its unit test; if removed,
  delete the offset tests.
- Update flow_transform.test.ts: remove the startOffset assertions (the "computes
  startOffset" test) and the startOffset field from expected node data.
- Keep the Gantt offset idea only for a true waterfall layout with a shared time axis
  (e.g., the Explore trace-details timeline), not for these node-graph cards.

Change 2: add a minimum visible bar width so tiny nonzero values do not look like 0
- In packages/osd-apm-topology/src/components/nodes/metric_bar.tsx, the fill uses
  width = (value/max)*100 %. A very small span (e.g. 0.003ms in a ~1s trace) computes
  to a sub-pixel width and disappears, which looks like 0 or a broken bar.
- Add a pixel-based minimum, applied only when value > 0, so tiny-but-nonzero values
  stay visible while true zeros remain empty:
  - On the fill div style, set minWidth to about 2px (e.g. '2px') when value > 0, and
    leave it unset (0) when value === 0.
  - Keep width as the real percentage so relative magnitude stays honest; the minWidth
    only guarantees a visible sliver. The label already shows the exact value, so this
    is not misleading.
- Add a unit test in metric_bar.test.tsx: a small nonzero value (value 1, max 10000)
  has a nonzero minWidth (visible), and value 0 has no minWidth floor (stays empty).

Note: the Explore trace-details waterfall already clamps to a minimum bar width, so
this brings the agent-traces bars in line with that behavior.

Run: yarn test:jest packages/osd-apm-topology and yarn test:jest
src/plugins/agent_traces, plus yarn typecheck and yarn lint. Keep commits DCO signed.

ps48 (approved):

joshuali925 (approved):

PR #12202, opened 2026-06-10
fix(explore): include data connection meta in metrics tab PromQL query

joshuali925:
### Description

Align the metrics tab with the query tab: reuse the live query state's dataset and override only the query string, so the data-connection meta reaches the server-side PromQL search strategy.

## Testing the changes

unit tests

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


joshuali925:
@joshuali925-osdbot pull down this PR and set up prometheus connection, use chrome to send requests that trigger this logic and verify `queryState` actually contains the right dataset with `meta` field. post your investigation results here

joshuali925:
bot is having problem setting up prometheus environment properly. i tested manually it works as expected

ps48 (approved):

lezzago (approved):

PR #5592, opened 2023-12-09
[Discover] Add long numerals support

AMoo-Miki:
### Description

* Add long numerals support to Discover tabular view, document view, inspector, hash-url, and share
* Use a fork of `numeral-js` that supports long numerals
* Patch `rison` to correctly serialize BigInts
Replace the uses of `withLongNumerals` with `withLongNumeralsSupport` to match the rest of the codebase and mark `withLongNumerals` for deprecation
* Add `withLongNumeralsSupport` as an option to `code/public/http/fetch`
* Add long numerals support to `UiSettingsClient`
* Add long numerals support to `core/server/http/router` response handler
* Add long numerals support to `RangeFilter` and `FilterBar`
* Add long numerals support to `kuery/ast`
* Introduce `OPENSEARCH_SEARCH_WITH_LONG_NUMERALS_STRATEGY` in `core/plugins/data/search`

### Check List

- [x] All tests pass
  - [ ] `yarn test:jest`
  - [ ] `yarn test:jest_integration`
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
- [x] Update [CHANGELOG.md](./../CHANGELOG.md)
- [X] Commits are signed per the DCO using --signoff

PR #12208, opened 2026-06-12
Replace @amoo-miki/numeral fork with stock @elastic/numeral + BigInt helper

Maosaic:
### Description

`@elastic/numeral` was aliased to a personal-scope fork, `npm:@amoo-miki/numeral@2.6.0`, in two places (`package.json` and `packages/osd-ui-shared-deps/package.json`). The fork exists for exactly **one** reason: BigInt (long-numeral) support added in #5592. A line-by-line diff confirms the fork is just **stock `@elastic/numeral@2.5.1` + a BigInt branch** — nothing else.

This PR removes the fork in favor of the unmodified upstream package plus a small, well-tested in-repo helper that reproduces the BigInt path **byte-for-byte**, so we no longer depend on a personal npm scope for a core formatting library.

**Changes**
- `package.json` + `packages/osd-ui-shared-deps/package.json`: alias both back to stock `@elastic/numeral@2.5.1` (the shared-deps one feeds the browser bundle).
- `field_formats/utils/format_bigint.ts` (new): faithful port of numeral's BigInt format path — number / currency / abbreviation / sign / grouping. The `percent` (`%`), `bytes` (`b`) and `ordinal` (`o`) format types perform the same Number arithmetic numeral does, which throws on a BigInt — **identical to the forked package's current behavior**.
- `converters/numeral.ts`: route the `isBigInt` branch through `formatBigInt`; the `Number` path still calls stock numeral.
- `format_bigint.test.ts` + `__fixtures__/numeral_bigint_golden.json` (new): a 576-case golden corpus **captured from the live fork** asserts bug-for-bug parity, including the throws.

**Why this is safe (no behavior change)**
- **Number path:** stock 2.5.1 is byte-identical to the fork for every Number input (verified across 1360 cases over en/fr/ja/de — the fork only adds a BigInt branch). Only 1 of the 5 numeral consumers in the repo ever receives a BigInt (the field-format converter); the other four are Number-only.
- **BigInt path:** the new helper matches the fork on all 576 captured cases (values up to int64 min/max `±9,223,372,036,854,775,808`), including locale-specific quirks (e.g. `en | 0o` throws while `fr | 0o` → `…e`).

### Issues Resolved

N/A — supply-chain/maintenance cleanup (removes dependency on a personal-scope npm fork).

## Screenshot

N/A — no UI change.

## Testing the changes

```
yarn test:jest src/plugins/data/common/field_formats/utils/format_bigint.test.ts
yarn test:jest src/plugins/data/common/field_formats/converters/number.test.ts \
               src/plugins/data/common/field_formats/converters/bytes.test.ts \
               src/plugins/data/common/field_formats/converters/percent.test.ts
```

`format_bigint.test.ts`: **584/584 pass** (576 golden-parity cases + 8 spot checks). Converter tests: **6/6 pass**.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest` (new + affected converter suites)
  - [ ] `yarn test:jest_integration` (n/a — no integration surface changed)
- [ ] New functionality has been documented. (n/a — internal refactor, no API change)

### DIFF
--- a/src/plugins/data/common/field_formats/converters/numeral.ts
+++ b/src/plugins/data/common/field_formats/converters/numeral.ts
@@ -36,6 +36,7 @@ import { OSD_FIELD_TYPES } from '../../osd_field_types/types';
 import { FieldFormat } from '../field_format';
 import { TextContextTypeConvert } from '../types';
 import { UI_SETTINGS } from '../../constants';
+import { formatBigInt } from '../utils/format_bigint';
 
 const numeralInst = numeral();
 
@@ -70,7 +71,12 @@ export abstract class NumeralFormat extends FieldFormat {
       (this.getConfig && this.getConfig(UI_SETTINGS.FORMAT_NUMBER_DEFAULT_LOCALE)) || 'en';
     numeral.language(defaultLocale);
 
-    const formatted = numeralInst.set(val).format(this.param('pattern'));
+    // Stock `@elastic/numeral` has no BigInt support, so long-numeral values are
+    // routed through a faithful port of the BigInt format path instead.
+    const pattern = this.param('pattern');
+    const formatted = isBigInt
+      ? formatBigInt(val, pattern, (numeral as any).languageData())
+      : numeralInst.set(val).format(pattern);
 
     numeral.language(previousLocale);
 
--- a/src/plugins/data/common/field_formats/converters/numeral.ts
+++ b/src/plugins/data/common/field_formats/converters/numeral.ts
@@ -71,8 +71,8 @@ export abstract class NumeralFormat extends FieldFormat {
       (this.getConfig && this.getConfig(UI_SETTINGS.FORMAT_NUMBER_DEFAULT_LOCALE)) || 'en';
     numeral.language(defaultLocale);
 
-    // Stock `@elastic/numeral` has no BigInt support, so long-numeral values are
-    // routed through a faithful port of the BigInt format path instead.
+    // `@elastic/numeral` has no BigInt support, so long-numeral values are routed
+    // through `formatBigInt` instead.
     const pattern = this.param('pattern');
     const formatted = isBigInt
       ? formatBigInt(val, pattern, (numeral as any).languageData())
--- a/src/plugins/data/common/field_formats/utils/format_bigint.ts
+++ b/src/plugins/data/common/field_formats/utils/format_bigint.ts
@@ -10,17 +10,14 @@
  */
 
 /*
- * This is a faithful port of the BigInt-handling code path that the forked
- * `@amoo-miki/numeral` package added on top of stock `@elastic/numeral`. It lets us
- * depend on the unmodified upstream `@elastic/numeral` (which has no BigInt support)
- * while preserving byte-for-byte identical output when an OpenSearch `long` field
- * value arrives as a BigInt.
+ * `@elastic/numeral` has no BigInt support, so OpenSearch `long` field values that
+ * arrive as a BigInt are formatted here. This mirrors numeral's own `formatNumber`/
+ * `formatCurrency` logic, specialized for BigInt — an integer, so there is no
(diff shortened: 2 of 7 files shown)


src/plugins/data/common/field_formats/utils/format_bigint.ts
```diff
@@ -0,0 +1,247 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+/*
+ * This is a faithful port of the BigInt-handling code path that the forked
+ * `@amoo-miki/numeral` package added on top of stock `@elastic/numeral`. It lets us
```
angle943: i don't think we need all these comments

huyaboo: +1

src/plugins/data/common/field_formats/converters/numeral.ts
```diff
@@ -70,7 +71,12 @@ export abstract class NumeralFormat extends FieldFormat {
       (this.getConfig && this.getConfig(UI_SETTINGS.FORMAT_NUMBER_DEFAULT_LOCALE)) || 'en';
     numeral.language(defaultLocale);
 
-    const formatted = numeralInst.set(val).format(this.param('pattern'));
+    // Stock `@elastic/numeral` has no BigInt support, so long-numeral values are
+    // routed through a faithful port of the BigInt format path instead.
+    const pattern = this.param('pattern');
```
huyaboo, line 76: We have performance tests on this right? When this was first introduced, there was latency spikes with large queries

angle943 (dismissed):

src/plugins/data/common/field_formats/converters/numeral.ts
Maosaic: Good callback — but this PR doesn't touch the path that caused those spikes, and the hot path here is unchanged.

The large-query latency from when long numerals first landed (#5592) came from the BigInt-aware **JSON parsing** layer (`withLongNumeralsSupport` in `core/.../http/fetch.ts`, `response_adapter.ts`, the opensearch client), which parses whole response bodies with a slower non-native parser. None of that is touched here — this PR only changes the *formatting* converter.

On the formatting side specifically:

- **Number path (the hot path) is byte-identical to the existing optimized code.** It still does `numeralInst.set(val).format(pattern)` against the module-level singleton `numeralInst = numeral()` (created once), not `numeral(val)` which would allocate a fresh instance per cell. That singleton-reuse is the existing optimization and is unchanged. Only the field-format converter ever receives a BigInt; the other numeral consumers are Number-only, so virtually every value in a large grid hits this unchanged path.
- **BigInt path is lighter than before, not heavier.** `formatBigInt(...)` is a pure, allocation-free function (string ops + truncating BigInt division) — the old fork did the same work *inside* numeral with a class instance. `languageData()` is only fetched on the (rare) BigInt branch.

So no new per-value allocation is introduced relative to `main`.

That said — there's no automated perf test asserting this in the repo today; the guarantee rests on the hot path being unchanged plus the 576-case golden-corpus parity test. Happy to add a micro-benchmark (format N BigInts vs Numbers) if you'd like a regression guardrail.

src/plugins/data/common/field_formats/utils/format_bigint.ts
Maosaic: Good call — trimmed in ddb4364720. I cut the verbose header/inline comments down and also dropped the references to the old `@amoo-miki/numeral` fork throughout this file (plus the test and the one mention in `numeral.ts`), since we're removing that dependency and a future reader would have no context for it. The comments now describe the helper on its own terms (numeral's BigInt format path). 584/584 still pass.

angle943 (approved):

ruchidh (approved):

PR #12209, opened 2026-06-12
[Workspace] Mark requested IDs not returned by the API as sentinel to avoid re-fetching

gaobinlong:
### Description

In the workspace collaborators page, if an user or a group is deleted, we cannot fetch its name by the identity entries API, we need to set the `name` field to a sentinel value for these deleted user or group, to avoid re-fetching its name in next render.

By the way, hide the user input `source` from the error message returned by the identity entries API, to improve security.

## Screenshot
image

### Check List

- [x] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/server/security/identity_source_service.test.ts
+++ b/src/core/server/security/identity_source_service.test.ts
@@ -45,7 +45,7 @@ describe('IdentitySourceService', () => {
 
       expect(() => {
         service.registerIdentitySourceHandler(sourceA, mockHandler2);
-      }).toThrow("Identity source 'sourceA' has already been registered");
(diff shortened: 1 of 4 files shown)


xluo-aws:
if it's deleted, why do we still want to show it on the page?

Hailong-am (approved):

gaobinlong:
> if it's deleted, why do we still want to show it on the page?

We may mark it as 'not found' or 'deleted', but let users to delete from the collaborator list? If not shown, users never get a chance to remove it from the backend storage by UI.

SuZhou-Joe (approved):

PR #12210, opened 2026-06-12
feat(chat): parse and display inline suggestions from agent responses

Hailong-am:
### Description

Add support for parsing `SUGGESTIONS:["..."]` from the last line of assistant messages and displaying them as selectable suggestion bubbles in the chat UI.

When the agent appends a `SUGGESTIONS:["action1","action2"]` line to its response, the chat plugin:
1. Strips it from displayed markdown (including during streaming)
2. Renders selectable suggestion bubbles below the message
3. Clicking a bubble toggles selection and appends/removes from the chat input (separated by `;`)
4. Only inline suggestions affect input; custom provider suggestions execute their action directly

### Issues Resolved

_N/A - New feature_

## Screenshot

image

image

## Testing the changes

1. Start the dev server with the chat plugin enabled
2. Ask the agent a question (e.g., "show me 10 documents")
3. Verify the `SUGGESTIONS:` line is not visible in the rendered message
4. Verify suggestion bubbles appear below the last assistant message
5. Click a suggestion — it should append to the input with `;` separator and show a check icon with green border
6. Click the same suggestion again — it should deselect and remove from input
7. Run unit tests: `yarn test:jest src/plugins/chat/common/parse_inline_suggestions.test.ts`

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


wanglam (dismissed):

src/plugins/chat/public/components/chat_suggestions.tsx
```diff
@@ -81,14 +86,39 @@ const SuggestionBubble: React.FC<SuggestionBubbleProps> = ({
...
 }) => {
   const { suggestedActionsService, chatService } = useChatContext();
 
   const [customSuggestions, setCustomSuggestions] = useState<SuggestedActions[]>([]);
   const [isLoadingCustomSuggestions, setIsLoadingCustomSuggestions] = useState(false);
+  const [selectedMessages, setSelectedMessages] = useState<Set<string>>(new Set());
+
+  // Reset selection state when the message changes
+  useEffect(() => {
+    setSelectedMessages(new Set());
```
ruanyl: If a user selects a suggestion (appending it to the input), then manually backspaces or clears the input field, will the bubble remains visually selected?

src/plugins/chat/public/components/chat_suggestions.tsx
```diff
@@ -117,27 +147,51 @@ export const ChatSuggestions = ({
     loadCustomSuggestions();
   }, [suggestedActionsService, chatService, messages, currentMessage]);
 
-  if (isLoadingCustomSuggestions || customSuggestions.length === 0) {
+  const hasUserTypedInput =
+    !!(inputValue && inputValue.trim().length > 0) && selectedMessages.size === 0;
+  const visibleInlineSuggestions = hasUserTypedInput ? [] : inlineSuggestionActions;
+  const allSuggestions = [...visibleInlineSuggestions, ...customSuggestions];
+
+  if (isLoadingCustomSuggestions || allSuggestions.length === 0) {
```
ruanyl, line 146: Is it expected to return `null` if is loading but has inline suggestions?

Hailong-am: `isLoadingCustomSuggestions` will finally changed to false in line [143](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/12210/changes#diff-9947e6794350735843222d4c6e8a33c542afcf37a12e57bd52557007f9a7726fR143)
```
   } finally {
        setIsLoadingCustomSuggestions(false);
      }
```

ruanyl: I see, so inline suggestion won't show if custom suggestion is loading, maybe not a big issue

src/plugins/chat/public/components/chat_suggestions.tsx
Hailong-am: It shouldn't be. Change the logic to match the input with suggestion text, if input contains suggestion text then it will show as selected, otherwise show as deselected.

SuZhou-Joe (approved):

SuZhou-Joe:
Have we considered using tool for suggestion generation?

Hailong-am:
> Have we considered using tool for suggestion generation?

How would that works? Is it something like a workflow, the last step is calling tool suggestion generation tool?

SuZhou-Joe:
May not be a workflow, more like a tool with instruction to include suggestions after the final answer.
```
tools = [{
      "name": "suggest_followups",
      "description": "Always call this after answering. Provide 2-3 natural follow-up questions.",
      "input_schema": {
          "type": "object",
          "properties": {
              "suggestions": {
                  "type": "array",
                  "items": {"type": "string"},
                  "minItems": 2,
                  "maxItems": 3
              }
          },
          "required": ["suggestions"]
      }
  }]
```

ruanyl (approved):

Hailong-am:
> May not be a workflow, more like a tool with instruction to include suggestions after the final answer.
> 
> ```
> tools = [{
>       "name": "suggest_followups",
>       "description": "Always call this after answering. Provide 2-3 natural follow-up questions.",
>       "input_schema": {
>           "type": "object",
>           "properties": {
>               "suggestions": {
>                   "type": "array",
>                   "items": {"type": "string"},
>                   "minItems": 2,
>                   "maxItems": 3
>               }
>           },
>           "required": ["suggestions"]
>       }
>   }]
> ```

Make sense. LLM may not always invoke this tool at the last, i could have some test with this approach to see how it works.

PR #12217, opened 2026-06-15
fix(vis): prevent DoS via chained datasource functions in Timeline

Qxisylolo:
### Description

A crafted Timeline expression like `.static(1).static(2).static(3)` can crash the OSD server through memory exhaustion. The chain runner passes the previous stage's entire seriesList object as the `value` parameter to subsequent datasource functions(Datasource functions (`static`, `opensearch`) generate data from scratch), creating exponential nesting that explodes during JSON serialization.

### Root Cause

`invokeChain()` unconditionally prepends the previous result to the next function's arguments via `result.concat(link.arguments)`. This works for modifier functions (whose first parameter is a seriesList input) but is incorrect for datasource functions (like `static`). The mismatch causes seriesList objects to be stored as y-values inside data points, nesting exponentially with each chained datasource.

return results for static(1) after fix:
截屏2026-06-15 11 27 55

return results for static(1) before fix:
截屏2026-06-15 11 33 06

### fix
The fix adds a `functionDef.datasource` check in `invokeChain()` so that datasource functions in non-first chain positions are executed independently (same as chain-head behavior) rather than receiving piped input. This is semantically correct — datasources generate their own data and never need pipeline input from a prior stage.

script results after fix:
截屏2026-06-15 11 16 19

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


yubonluo (approved):
LGTM

ruanyl (approved):

PR #12220, opened 2026-06-15
fix(saved-objects): gate config import on advancedSettings.save capabilities

### LINKED ISSUE #12201: [BUG] saved_object API call failing in OpenSearch Dashboards 3.7.0 (worked in 3.6.0 and earlier)
**Describe the bug**

We use the `saved_object` API to load prebuilt content into OpenSearch Dashboards instances.  Our code and content has not changed in over a year but now fails when we point to an OpenSearch Dashboards 3.7.0 instance.  This same code and content worked successfully with versions 3.6.0, 3.4.0, 2.19.4 and 2.19.3.  (And comments in the code indicate we updated it to deal with a change in response syntax when moving from ODFE 1.7.0 to ODFE 1.13.2.)  When the API call is made to OpenSearch Dashboards 3.7.0, it returns an `unsupported_type` error.

**To Reproduce**
I simplified our actual code to create a minimal example that demonstrates the problem.  The following console session excerpt shows `curl` commands to submit the API call and the responses received against two different OpenSearch Dashboards instances (one OpenSearch Dashboards 3.7.0 instance and one OpenSearch Dashboards 3.6.0 instance).  In addition, it also shows the content of the sample content file being imported (which suppresses a new feature announcement introduced in 3.4.0(?)).  However, we are seeing the same results for other content files containing dashboards, saved queries and other types of content as well.

This excerpt has been copied from my session and, other than obfuscating sensitive information (e.g. hostnames, passwords, etc.), is "real".  

```
$ curl  -XPOST "https://osd370.xxx.xxx/api/saved_objects/_import?overwrite=true" -H "osd-xsrf:true" --form file=@"sample_content.ndjson" --user "admin":"xxxx" --insecure
{"successCount":0,"success":false,"errors":[{"error":{"type":"unsupported_type"},"type":"config","id":"explore-experience-notice-dismissed","title":"explore-experience-notice-dismissed","meta":{"title":"explore-experience-notice-dismissed"}}]}


$ curl  -XPOST "https://osd670.xxx.xxx/dashboards/api/saved_objects/_import?overwrite=true" -H "osd-xsrf:true" --form file=@"sample_content.ndjson" --user "admin":"xxxx" --insecure
{"successCount":1,"success":true,"successResults":[{"type":"config","id":"explore-experience-notice-dismissed","meta":{"icon":"gear"},"overwrite":true}]}

$ cat sample_content.ndjson
{"attributes":{"dismissedAt":"2026-03-12T19:57:20.211Z"},"id":"explore-experience-notice-dismissed","migrationVersion":{"config":"7.9.0"},"references":[],"type":"config","updated_at":"2026-03-12T19:57:20.275Z","version":"Wzg5LDFd"}
```

**Expected behavior**
I expected OpenSearch Dashboard 3.7.0 to accept the same saved content successfully imported in earlier versions of OpenSearch Dashboards using the same API call.

**OpenSearch Version**
OpenSearch 3.7.0
We deploy using the publicly available container image, i.e. `docker.io/opensearchproject/opensearch:3.7.0`

**Dashboards Version**
OpenSearch Dashboard 3.7.0
We deploy using the publicly available container image, i.e. `docker.io/opensearchproject/opensearch-dashboards:3.7.0`

**Plugins**

Default set contained in the Docker image identified above.

**Screenshots**

See session output above

 - Kubernetes 1.33.2
 - Host nodes: Rocky Linux 9.6

sumukhswamy:
### Description

The config type block introduced in #12014 rejected all config-type objects unconditionally and failed the entire import when any were present.

This change:
- Gates config import on the advancedSettings.save capability: admin users with this capability can import config objects; non-admin users still receive unsupported_type errors for config objects.
- Changes from fail-all to filter-and-continue: when config objects are rejected, remaining valid objects in the same import are still processed. Config objects are reported as errors in the response alongside any successful imports.
### Issues Resolved

Resolves #12201

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/core/server/saved_objects/import/import_saved_objects.ts
+++ b/src/core/server/saved_objects/import/import_saved_objects.ts
@@ -63,6 +63,7 @@ export async function importSavedObjectsFromStream({
   workspaces,
   dataSourceEnabled,
   isCopy,
+  canImportConfig,
 }: SavedObjectsImportOptions): Promise<SavedObjectsImportResponse> {
   let errorAccumulator: SavedObjectsImportError[] = [];
   const supportedTypes = typeRegistry.getImportableAndExportableTypes().map((type) => type.name);
@@ -74,27 +75,26 @@ export async function importSavedObjectsFromStream({
     supportedTypes,
     dataSourceId,
   });
-  const configErrors: SavedObjectsImportError[] = collectSavedObjectsResult.collectedObjects
-    .filter((obj) => obj.type === 'config')
-    .map((obj) => ({
-      error: { type: 'unsupported_type' } as SavedObjectsImportUnsupportedTypeError,
-      type: obj.type,
-      id: obj.id,
-      title: obj.id,
-      meta: { title: obj.id },
-    }));
-  if (configErrors.length > 0) {
-    return {
-      successCount: 0,
-      success: false,
-      errors: configErrors,
-    };
+  let collectedObjects = collectSavedObjectsResult.collectedObjects;
+
+  if (!canImportConfig) {
+    const configErrors: SavedObjectsImportError[] = collectedObjects
+      .filter((obj) => obj.type === 'config')
+      .map((obj) => ({
+        error: { type: 'unsupported_type' } as SavedObjectsImportUnsupportedTypeError,
+        type: obj.type,
+        id: obj.id,
+        title: obj.id,
+        meta: { title: obj.id },
+      }));
+    errorAccumulator = [...errorAccumulator, ...configErrors];
+    collectedObjects = collectedObjects.filter((obj) => obj.type !== 'config');
   }
 
   // if dataSource is not enabled, but object type is data-source, or saved object id contains datasource id
   // return unsupported type error
   if (!dataSourceEnabled) {
-    const notSupportedErrors: SavedObjectsImportError[] = collectSavedObjectsResult.collectedObjects.reduce(
+    const notSupportedErrors: SavedObjectsImportError[] = collectedObjects.reduce(
       (errors: SavedObjectsImportError[], obj) => {
         if (obj.type === 'data-source' || isSavedObjectWithDataSource(obj.id)) {
           const error: SavedObjectsImportUnsupportedTypeError = { type: 'unsupported_type' };
@@ -121,7 +121,7 @@ export async function importSavedObjectsFromStream({
 
   // Validate references
   const validateReferencesResult = await validateReferences(
-    collectSavedObjectsResult.collectedObjects,
+    collectedObjects,
     savedObjectsClient,
     namespace
   );
@@ -129,11 +129,9 @@ export async function importSavedObjectsFromStream({
 
   if (isCopy) {
     // Data sources can only be assigned to workspaces and can not be copied between workspaces.
-    collectSavedObjectsResult.collectedObjects = collectSavedObjectsResult.collectedObjects.filter(
-      (obj) => obj.type !== 'data-source'
-    );
+    collectedObjects = collectedObjects.filter((obj) => obj.type !== 'data-source');
     const validateDataSourcesResult = await validateDataSources(
-      collectSavedObjectsResult.collectedObjects,
+      collectedObjects,
       savedObjectsClient,
       errorAccumulator,
       workspaces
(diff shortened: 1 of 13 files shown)


angle943:
**The most security-sensitive new code is untested.** The capability-derivation glue in `routes/import.ts` and `routes/resolve_import_errors.ts`:

```ts
let canImportConfig = false;
const resolver = getCapabilities?.();
if (resolver) {
  const capabilities = await resolver(req);
  canImportConfig = capabilities.advancedSettings?.save === true;
}
```

has no route-level test. The unit tests exercise `importSavedObjectsFromStream({ canImportConfig })` directly, but nothing verifies that:

- resolver present + `advancedSettings.save === true` → `canImportConfig` true
- resolver present + `save` false/absent → false
- resolver absent → false (the fail-safe)

This is exactly the boundary an authz regression would slip through. I'd ask for explicit coverage here before merge.

src/core/server/saved_objects/import/import_saved_objects.ts
```diff
@@ -74,21 +75,20 @@ export async function importSavedObjectsFromStream({
...
+        error: { type: 'unsupported_type' } as SavedObjectsImportUnsupportedTypeError,
+        type: obj.type,
+        id: obj.id,
+        title: obj.id,
+        meta: { title: obj.id },
+      }));
+    errorAccumulator = [...errorAccumulator, ...configErrors];
+    collectSavedObjectsResult.collectedObjects = collectSavedObjectsResult.collectedObjects.filter(
+      (obj) => obj.type !== 'config'
+    );
```
angle943: nit: this mutates the returned `collectSavedObjectsResult` object in place, whereas `resolve_import_errors.ts` does the equivalent cleanly with a local (`let objectsToResolve = collectedObjects; ... objectsToResolve = objectsToResolve.filter(...)`).

Reassigning a property on the result object is a side effect that's easy for a future reader to miss. Prefer the local-variable form here too so the two call sites stay consistent and `collectSavedObjectsResult` is left untouched.

src/core/server/saved_objects/routes/resolve_import_errors.ts
```diff
@@ -125,6 +130,13 @@ export const registerResolveImportErrorsRoute = (router: IRouter, config: SavedO
         workspaces = [workspaces];
       }
 
+      let canImportConfig = false;
+      const resolver = getCapabilities?.();
+      if (resolver) {
+        const capabilities = await resolver(req);
+        canImportConfig = capabilities.advancedSettings?.save === true;
+      }
+
```
angle943, line 134: This route carries the same security-sensitive capability glue as `import.ts`, but unlike the import route it has no route-level integration test for the gating. The unit tests exercise `resolveSavedObjectsImportErrors({ canImportConfig })` directly, but nothing verifies the resolver→`canImportConfig` derivation here. Could we mirror the three integration cases added for the import route — resolver absent (fail-safe → blocked), `advancedSettings.save === false` (blocked), and `=== true` (allowed)? This is the same gap that was flagged and closed for `import.ts`; the resolve-errors path is equally exploitable.

src/core/server/saved_objects/routes/import.ts
```diff
@@ -119,6 +126,13 @@ export const registerImportRoute = (router: IRouter, config: SavedObjectConfig)
 
       const dataSourceEnabled = req.query.dataSourceEnabled;
 
+      let canImportConfig = false;
+      const resolver = getCapabilities?.();
+      if (resolver) {
+        const capabilities = await resolver(req);
+        canImportConfig = capabilities.advancedSettings?.save === true;
+      }
```
angle943: This exact capability-resolution block is duplicated in `resolve_import_errors.ts`. Since this is the authorization decision, I'd extract it into a shared helper (e.g. `resolveCanImportConfig(getCapabilities, req))` so the two routes can't drift — a future change to the capability name or check logic should only need to happen in one place.

sumukhswamy: Shared helper extracted resolveCanImportConfig(getCapabilities, req) in import.ts, imported and used by both routes. No more duplicated logic.

src/core/server/saved_objects/routes/resolve_import_errors.ts
sumukhswamy: Resolve route integration tests added 3 matching test cases (resolver absent, save=false, save=true) mirroring the import route tests.

angle943 (dismissed):

angle943 (approved):

LDrago27 (approved):

PR #12236, opened 2026-06-18
[Bug] Trace dataset auto creation

TackAdam:
### Description
The trace/log auto-creation flow (Workspace Creator + Explore/Agent Traces callouts) was creating index patterns with no fields, breaking the time picker and any aggregations that referenced the time field.

**Symptom:** `Could not locate that index-pattern-field (id: endTime)` thrown from `AggConfig.setParams` whenever a chart attempted to use an auto-created dataset.

Fixed auto create:

https://github.com/user-attachments/assets/35a8aaa6-0c79-4ba8-a4c8-7cca303a17dc

## Root Cause

PR #11653 (*Remove legacy write-on-read behavior from DataViewsService and IndexPatternsService*) removed the safety net that automatically fetched fields when legacy or empty index patterns were read.

The auto-creation flow in Explore and Agent Traces was creating index patterns directly through `savedObjectsClient.create('index-pattern', ...)`, which never triggers `_fields_for_wildcard`. As a result, newly-created patterns were persisted with `fields: undefined`.

Simply switching to `dataViews.createAndSave(...)` was not sufficient. Its internal `refreshFields()` call can silently fail in some contexts. This was verified through DevTools: clicking **Create Trace Datasets** produced no `_fields_for_wildcard` request at all.

The Dataset Management and Index Pattern Management wizards were unaffected because their `step_time_field` UI calls `getFieldsForWildcard()` before creation in order to populate the time-field selector. That side effect effectively "warms" the field list before `createAndSave()` executes.

The auto-create flow skipped that step entirely.

## Fix

The auto-creation utility now mirrors the manual wizard flow explicitly.

### 1. Pre-fetch fields

Before creating the dataset, call:

`dataViews.getFieldsForWildcard({ pattern, dataSourceId })`

This is the same API used by `step_time_field.tsx`.

### 2. Embed fields directly into the DataView spec

The fetched field list is converted via `fieldArrayToMap()` and supplied directly to:

`dataViews.create(spec, /* skipFetchFields */ true)`

followed by:

`dataViews.createSavedObject(dataView)`

Going through `create()` + `createSavedObject()` instead of `createAndSave()` lets us:

* Pass `skipFetchFields: true` since fields were already pre-fetched into the spec, eliminating an unnecessary round-trip.
* Avoid `createAndSave()`'s internal `setDefault(id)` call, which would otherwise silently change the workspace's default index pattern whenever auto-creation ran in a workspace with no default configured.

Because the `DataView` constructor reads `spec.fields` and initializes `this.fields`, the saved object is persisted with a populated `fields` payload even if the in-flight `refreshFields()` operation never succeeds.

### 3. Post-create refresh

After creation, perform:

`clearCache() → get() → refreshFields() → updateSavedObject() → clearCache()`

This provides an additional recovery path for any dataset that still lands with empty fields.

### 4. Refresh reused patterns

The existing-pattern reuse path now performs the same refresh sequence.

This allows index patterns that were previously created by the buggy implementation to automatically recover the next time the auto-create flow executes.

### Failure handling

If all field-loading mechanisms fail (network issues, missing mappings, etc.), errors are caught and logged through `console.warn(...)`.

The dataset is still created and users can manually select **Refresh field list** later. No crash occurs.

### Indices do not exist / no OTel fields

Detection returns `tracesDetected: false`.

The callout is not shown and no dataset is created.

### Detection succeeds but field pre-fetch fails

The dataset is still created.

A warning is logged and the dataset displays **Fields (0)** until manually refreshed. No crash occurs.

### Existing empty pattern is reused

The pattern is automatically recovered through the refresh path added to the reuse flow.

#### Explore

**`src/plugins/explore/public/utils/create_auto_datasets.ts`**

* Reworked `createAutoDetectedDatasets`
* Accepts `DataViewsContract`
* Pre-fetches fields
* Embeds fields in the creation spec
* Uses `create()` + `createSavedObject()` (skips `setDefault()` and removes one redundant field fetch)
* Refreshes fields during both create and reuse flows

**`src/plugins/explore/public/components/trace_auto_detect_callout.tsx`**

* Passes `services.dataViews`
* No longer routes through `services.indexPatterns`

#### Agent Traces (single source of truth)

**`src/plugins/agent_traces/public/components/trace_auto_detect_callout.tsx`**

* Imports `createAutoDetectedDatasets` directly from `../../../explore/public`
* Matches the same pattern already used by `workspace_creator`

**Deleted:**

* `src/plugins/agent_traces/public/utils/create_auto_datasets.ts`
* `src/plugins/agent_traces/public/utils/create_auto_datasets.test.ts`

The previous implementation duplicated Explore logic and had begun to drift. Agent Traces now relies on the Explore implementation as the single source of truth.

**`src/plugins/agent_traces/public/index.ts`**

* No longer re-exports `createAutoDetectedDatasets`
* No external consumers remain

#### Workspace

**`src/plugins/workspace/public/components/workspace_creator/workspace_creator.tsx`**

* Passes `dataPlugin.dataViews`
* Removes an obsolete `@ts-expect-error` directive now flagged as unnecessary by type checking

### `create_auto_datasets.test.ts` (Explore)

Rewritten for the new function signature with coverage for:

* Pre-fetching fields and embedding them in the DataView spec
* Verifying `setDefault()` is never called
* Creating datasets when field pre-fetch returns no results
* Reusing existing index patterns without creating duplicates while still refreshing fields
* Returning the existing ID when refresh operations fail
* Falling back to `find()` when `createSavedObject()` throws `DuplicateDataViewError`

### Callout Component Tests

Both Explore and Agent Traces callout tests were updated to:

* Include `dataViews` in mocks
* Validate the updated function signature
* Mock `../../../explore/public` in Agent Traces to support the new direct import path

## Test Results

* ✅ 38/38 tests passing across all affected suites
* ✅ Type checking clean on all modified files

## Verification

Using the DevTools Network tab, clicking **Create Trace Datasets** now produces `_fields_for_wildcard?pattern=otel-v1-apm-span*` requests that were previously missing.

The resulting saved object contains a populated `attributes.fields` payload, and the trace page loads successfully without the `Could not locate that index-pattern-field (id: endTime)` error.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/plugins/explore/public/utils/create_auto_datasets.ts
```diff
@@ -13,11 +14,108 @@ export interface CreateDatasetsResult {
...
+  // refreshAndPersistFields below makes a second attempt.
+  const fields = await fetchFieldsForPattern(
+    dataViews,
+    spec.title as string,
+    effectiveDataSourceId
+  );
+
+  let createdId: string | null = null;
+  try {
+    const created = await dataViews.createAndSave({ ...spec, fields });
```
ps48: `createAndSave` internally calls `setDefault(id)` (data_views.ts:639), which writes `defaultIndex` whenever it's currently unset (data_views.ts:338). The previous `savedObjectsClient.create` path never touched `defaultIndex`. So in a workspace with no default yet, auto-creation will now set the trace/log dataset as the user's default index pattern. Flagging in case that side effect isn't intended here.

src/plugins/explore/public/utils/create_auto_datasets.ts
```diff
@@ -13,11 +14,108 @@ export interface CreateDatasetsResult {
...
+  }
+}
+
+// Force-refresh an existing index pattern's field list and persist it. Recovers index
+// patterns left empty by the prior buggy version of this code.
+async function refreshAndPersistFields(dataViews: DataViewsContract, id: string): Promise<void> {
+  try {
+    dataViews.clearCache(id);
+    const view = await dataViews.get(id);
+    await dataViews.refreshFields(view);
```
ps48, line 44: Worth noting the field list can be fetched up to three times per dataset: the explicit `fetchFieldsForPattern` pre-fetch, `createAndSave`'s own `refreshFields` (since `skipFetchFields` defaults to `false`), and this `refreshAndPersistFields` pass. Since fields are already embedded in the spec, passing `skipFetchFields=true` to `createAndSave` would drop one round-trip.

src/plugins/explore/public/utils/create_auto_datasets.ts
```diff
@@ -13,11 +14,108 @@ export interface CreateDatasetsResult {
...
+async function fetchFieldsForPattern(
+  dataViews: DataViewsContract,
+  pattern: string,
+  dataSourceId?: string
+) {
+  try {
+    const fields = await dataViews.getFieldsForWildcard({ pattern, dataSourceId });
+    if (!Array.isArray(fields) || fields.length === 0) {
+      // eslint-disable-next-line no-console
+      console.warn(`No fields returned for pattern "${pattern}" (dataSource: ${dataSourceId})`);
```
ps48, line 27: These field-fetch failures surface to the user as a silently empty field list. Routing through `notifications.toasts` or a debug logger would make them more discoverable than `console.warn` (matches existing code here, so non-blocking).

src/plugins/agent_traces/public/utils/create_auto_datasets.ts
```diff
@@ -13,11 +14,102 @@ export interface CreateDatasetsResult {
   correlationId: string | null;
 }
 
+// Pre-fetch the field list ourselves before calling createAndSave. PR #11653 removed the
```
ps48: This comment is truncated to a single line ending mid-clause ("...removed the"). The explore copy has the full 6-line version — worth syncing them.

src/plugins/agent_traces/public/utils/create_auto_datasets.ts
```diff
@@ -13,11 +14,102 @@ export interface CreateDatasetsResult {
...
+  }
+  return createdId;
+}
+
 /**
  * Create auto-detected trace and log datasets with correlation
  */
 export async function createAutoDetectedDatasets(
   savedObjectsClient: SavedObjectsClientContract,
+  dataViews: DataViewsContract,
```
ps48: This file and `explore/public/utils/create_auto_datasets.ts` are intended to be identical (test files too), and the truncated comment above shows they're already drifting. A shared helper — or importing from one plugin into the other, as `workspace_creator` already imports from `explore/public` — would prevent future skew.

TackAdam:
@ps48 
  Thanks for the review — addressed all of these:

  1. setDefault side effect — Switched from dataViews.createAndSave() to explicit dataViews.create() + dataViews.createSavedObject(). Auto-creation no longer touches defaultIndex. New test asserts setDefault is never invoked.

  2. 3× field fetch — Now passing skipFetchFields: true to create() since fields are already pre-fetched and embedded in the spec. Round-trips per dataset go from 3 → 2.

  3. Drift between explore/agent_traces — Eliminated. agent_traces/public/utils/create_auto_datasets.ts and its test file are deleted. agent_traces/public/components/trace_auto_detect_callout.tsx now imports createAutoDetectedDatasets directly from ../../../explore/public (same pattern workspace_creator already uses). Single source of truth, no shim file.

  Tests: 38/38 passing across the three remaining suites. Typecheck clean.

ps48 (approved):

Maosaic (approved):

PR #12255, opened 2026-06-18
feat(query-enhancements): PPL lint backend — feature flag + explain/calcite proxy routes

Hanyu-W:
### Description

Backend plumbing for the PPL linter, adding the feature flag and two read-only proxy routes. 

- **Feature flag** — `queryEnhancements.pplLint` capability, default `false`, resolved at runtime via a `DynamicConfigService` capability switcher (same pattern as `explore`). 
- **`POST /api/enhancements/ppl/explain`** — proxies a PPL query to `/_plugins/_ppl/_explain` and returns the Calcite execution plan. Validates a non-empty `query`; supports optional `dataSourceId`.
- **`GET /api/enhancements/ppl/calcite_settings`** — reads `/_cluster/settings` (scoped via `filter_path`) and returns `{ calciteEnabled, allJoinTypesAllowed }`. Fails open on errors so a settings-read failure never blocks the editor; logs 401/403 at `warn`.

### Issues Resolved

Backend plumbing for opensearch-project/sql#5405

## Screenshot

N/A — no UI changes.

## Testing the changes

1. Run unit tests for the new routes and capability switcher:
   ```bash
   node scripts/jest.js \
     src/plugins/query_enhancements/server/plugin.test.ts \
     src/plugins/query_enhancements/server/routes/ppl_calcite_settings.test.ts \
     src/plugins/query_enhancements/server/routes/ppl_explain.test.ts
   ```
2. Run the full plugin server suite to confirm no regressions:
   ```bash
   node scripts/jest.js src/plugins/query_enhancements/server
   ```
3. Point the dev server at a live `opensearch-sql` cluster and hit `GET /api/enhancements/ppl/calcite_settings` — verify the scoped `filter_path` response matches the full unfiltered `/_cluster/settings` output for `calciteEnabled` and `allJoinTypesAllowed`.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`
- [ ] New functionality has been documented


Swiddis (approved):

src/plugins/query_enhancements/server/routes/ppl_calcite_settings.ts
```diff
@@ -0,0 +1,74 @@
...
+        // Normalize to string so a typed-boolean value (e.g. JSON `false` from a
+        // future transport) compares the same as today's string `"false"`.
+        const resolveValue = (key: string): string | undefined => {
+          const raw = body?.transient?.[key] ?? body?.persistent?.[key] ?? body?.defaults?.[key];
+          return raw === undefined || raw === null ? undefined : String(raw);
+        };
+
+        return res.ok({
+          body: {
+            calciteEnabled: resolveValue('plugins.calcite.enabled') !== 'false',
```
ps48: ### p0: backward-compat issue with calciteEnabled default

`resolveValue(...) !== 'false'` returns `true` when the key is absent (`undefined !== 'false'`). On an older opensearch-sql cluster that predates `plugins.calcite.enabled`, the key is genuinely absent and the settings read succeeds (so we are on this success path, not the catch block), yet we report `calciteEnabled: true` for a cluster that has no Calcite engine at all.

The request already sends `include_defaults=true`, so any cluster that knows the setting surfaces it in the `defaults` bucket even when unset. Absence therefore reliably indicates that the cluster does not have Calcite, and the safe interpretation is the opposite of the current default. Since `calciteEnabled` gates the explain-based lint rules, the current behavior makes the editor fire `_explain` calls on every keystroke against clusters that cannot support them.

Suggested fix is to invert the success-path default:

```ts
calciteEnabled: resolveValue('plugins.calcite.enabled') === 'true',
```

New clusters still resolve correctly (the default is surfaced); old clusters correctly resolve to `false`. The catch-block fail-open (`true`) is a separate decision and can stay as-is. Worth confirming the intended old-cluster behavior with the SQL team either way.

src/plugins/query_enhancements/server/routes/ppl_calcite_settings.test.ts
```diff
@@ -0,0 +1,266 @@
...
+    const res = createResponse();
+
+    await handler()(context, req, res);
+
+    expect(context.dataSource.opensearch.getClient).not.toHaveBeenCalled();
+    expect(requestMock).toHaveBeenCalledWith({
+      method: 'GET',
+      path: EXPECTED_PATH,
+    });
+    // Absent setting => calcite treated as enabled (not 'false'), join types not allowed.
```
ps48: ### p0: test locks in the absent-setting default

This test currently encodes the behavior flagged on `ppl_calcite_settings.ts` (`// Absent setting => calcite treated as enabled`, asserting `calciteEnabled: true`). If the success-path default is inverted, this should flip to expect `false`, and it would be worth renaming or adding a case explicitly framed as an older cluster without `plugins.calcite.enabled` resolving to calcite disabled, so the backward-compat intent is captured in the test name.

src/plugins/query_enhancements/server/routes/index.ts
```diff
@@ -17,16 +19,40 @@ import { registerQueryAssistRoutes } from './query_assist';
...
+export const DATASOURCE_UNAVAILABLE_MESSAGE =
+  'dataSourceId is not supported because data source plugin is unavailable';
+
+/**
+ * Resolves the OpenSearch client for an optional dataSourceId. Returns the
+ * data source's client when a dataSourceId is given, the current-user client
+ * otherwise, or `null` when a dataSourceId is requested but the data source
+ * plugin is unavailable (the caller should respond 400 in that case).
+ */
+export async function resolveOpenSearchClient(
```
ps48, line 43: ### p1: add a direct multi-data-source test for resolveOpenSearchClient

This helper is the seam for the multi-data-source requirement in the parent RFC (sql#5405 section 2.9 plumbs `dataSourceId` end-to-end), but it is only tested indirectly through the two routes, and no test exercises two different `dataSourceId` values resolving to distinct clients. `getClient` is a single mock, so we currently prove that the id is forwarded, not that the right client comes back per id.

Suggested coverage in a dedicated unit test for this helper: `ds-1` resolves to `getClient('ds-1')` and `ds-2` resolves to `getClient('ds-2')` returning distinct clients, no id resolves to `asCurrentUser`, and an absent `context.dataSource` resolves to `null`. Cheap insurance given the rest of the lint feature builds on this.

src/plugins/query_enhancements/server/routes/ppl_explain.ts
```diff
@@ -0,0 +1,62 @@
...
+ * The response is the unwrapped transport body (`result.body ?? result`), which
+ * matches `definePPLBundleRoute`. The client parser must validate the plan shape
+ * before reading it rather than assume a fixed envelope.
+ */
+export function definePPLExplainRoute(logger: Logger, router: IRouter) {
+  router.post(
+    {
+      path: API.PPL_EXPLAIN,
+      validate: {
+        body: schema.object({ query: schema.string({ minLength: 1 }) }),
```
ps48: ### p1: confirm request guarding for a per-keystroke route

This proxies arbitrary PPL to `_explain` with only `minLength: 1` on `query`. It is read-only and auth is enforced by the downstream client, so the risk is low, but per the RFC this route is hit on every keystroke (debounced). Worth confirming the debounce and abort logic lives client-side, and whether a minimal server-side guard (max query length) is wanted. Non-blocking, flagging for the follow-up that wires the editor.

src/plugins/query_enhancements/common/config.ts
```diff
@@ -31,6 +31,11 @@ export const configSchema = schema.object({
       }),
     }),
   }),
+  // PPL linter feature flag, read at runtime via DynamicConfigService and
+  // surfaced as the queryEnhancements.pplLint capability. Disabled by default.
+  pplLint: schema.object({
```
joshuali925: can this be `ppl.lint.enabled`, or `lintEnabled: ["ppl"]`? i'm thinking if we add more languages in the future, can the config path be more structured?

src/plugins/query_enhancements/server/routes/ppl_calcite_settings.ts
joshuali925: +1

src/plugins/query_enhancements/common/config.ts
Hanyu-W: Good call — restructured to queryEnhancements.ppl.lint.enabled in 031498a (config schema now nests ppl: { lint: { enabled } }) so that future languages/features extend the same shape.

src/plugins/query_enhancements/server/routes/index.ts
Hanyu-W: Added a describe('resolveOpenSearchClient') to index.test.ts covering: two distinct dataSourceIds resolve to distinct clients (not just id-forwarding), no id resolves to asCurrentUser, and a dataSourceId with context.dataSource absent resolves to null.

src/plugins/query_enhancements/server/routes/ppl_calcite_settings.test.ts
Hanyu-W: Done — flipped 'uses the core client …' to expect calciteEnabled: false and added an explicitly-named case ('reports calciteEnabled:false for a cluster missing plugins.calcite.enabled') so the backward-compat intent is self-documenting. The catch-path tests ('swallows transport errors', 'logs auth failures at warn') keep asserting the fail-open true, which documents the success-vs-error asymmetry side by side.

src/plugins/query_enhancements/server/routes/ppl_calcite_settings.ts
Hanyu-W: Good catch. Confirmed that plugins.calcite.enabled is registered (with default true) only when the SQL plugin loads. I inverted the check to === 'true', left the catch-block fail-open (true) intact and added a comment: a 200 with the key absent is definitive "no Calcite," whereas an error can't distinguish "no plugin" from "transient failure." Side effect: calciteEnabled now matches the === 'true' idiom already used by allJoinTypesAllowed.

src/plugins/query_enhancements/server/routes/ppl_explain.ts
Hanyu-W: Confirmed — the keystroke throttling lives client-side and lands with the editor-wiring follow-up PR, not this backend slice. Specifically: a 500 ms trailing-edge debounce per model, plus an ExplainCache that dedups in-flight requests and LRU-caches results per (dataSourceId, query), so repeated lint passes over the same text issue at most one _explain call. There's no explicit AbortController on the explain request — the debounce + dedup make a superseded response a harmless cache write, so cancellation wasn't needed (easy to add later if we want it). On the server side, I added the maxLength: 65536 guard on the query schema here in 4ac79ae as the minimal request bound you flagged, independent of the global server.maxPayload.

ps48 (approved):
Thanks for the fixes

mengweieric (approved):

PR #12258, opened 2026-06-19
fix(data_source): prevent privilege escalation via unauthorized data …

ZilongX:
Closes a privilege escalation vulnerability in the multi-data-source feature where any authenticated user with tenant read access could use another user's stored data source credentials to bypass all OpenSearch security controls.

Root cause: getClient(dataSourceId) passed the caller's scoped SavedObjects client to configureClient/configureLegacyClient, which fetched and decrypted stored credentials with no authorization check on whether the caller was permitted to USE that data source. The console proxy route (POST /api/console/proxy) passed dataSourceId directly from the query string with no permission validation.

Fix (two layers):

Layer 1 — Strip credentials from read APIs in DataSourceSavedObjectsClientWrapper. The wrapper previously only intercepted write operations (create/update). get(), find(), and bulkGet() now strip auth.credentials from data-source saved objects before returning to any caller, so encrypted credential material is never exposed through the SavedObjects API.

Layer 2 — Internal repository for credential access in getClient. DataSourcePlugin.start() creates an internal SavedObjects repository that bypasses the credential-stripping wrapper. configureClient and configureLegacyClient now perform a two-step fetch: 1. getDataSource(id, scopedClient) — access check; throws 404/Forbidden if the calling user's tenant/workspace does not include this data source 2. getDataSourceInternal(id, internalRepo) — fetches with full encrypted credentials only after step 1 passes

internalSavedObjects is optional in DataSourceClientParams for backward compatibility; callers that do not provide it (e.g. test-connection routes, existing unit tests) retain the previous single-fetch behavior with no change in call count or semantics.

Affected: Self-managed deployments with data_source.enabled: true.

Issues Resolved

CVE pending — CWE-863: Incorrect Authorization (GitHub Security Advisory in progress)

Screenshot

N/A — backend-only change, no UI modifications.

Testing the changes

* Unauthorized user calling POST /api/console/proxy with another user's dataSourceId receives 404
* Authorized user can use their own data sources without disruption
* auth.credentials no longer appears in SavedObjects API responses for data sources
* Existing unit tests pass; new tests cover credential-stripping wrapper behavior

Check List

* [x] All tests pass
* [x] yarn test:jest
* [x] yarn test:jest_integration
* [x] New functionality includes testing.
* [ ] New functionality has been documented.
* [x] Commits are signed per the DCO using --signoff

### DIFF
--- a/src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
+++ b/src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
@@ -155,7 +155,7 @@ export class DataSourceSavedObjectsClientWrapper {
     };
 
     const bulkGetWithCredentialsStripping = async <T = unknown>(
-      objects?: Array<SavedObjectsBulkGetObject>
+      objects?: SavedObjectsBulkGetObject[]
     ) => {
       const result = await wrapperOptions.client.bulkGet<T>(objects);
       return {
--- a/src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
+++ b/src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
@@ -5,6 +5,7 @@
 
 import {
   SavedObjectsBulkCreateObject,
+  SavedObjectsBulkGetObject,
   SavedObjectsBulkResponse,
   SavedObjectsBulkUpdateObject,
   SavedObjectsBulkUpdateOptions,
@@ -12,6 +13,8 @@ import {
   SavedObjectsClientWrapperFactory,
   SavedObjectsClientWrapperOptions,
   SavedObjectsCreateOptions,
+  SavedObjectsFindOptions,
+  SavedObjectsFindResponse,
   SavedObjectsUpdateOptions,
   SavedObjectsUpdateResponse,
 } from 'opensearch-dashboards/server';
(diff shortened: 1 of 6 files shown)


Maosaic:
Thanks for tackling this — the two-layer design (strip credentials from external read APIs + bypass via an internal repository for legitimate decryption) is the right approach. I went through it against the surrounding code. One latent bug worth fixing, plus minor notes.

> _Edit: removed an earlier finding about test-connection on existing data sources. Requiring credential re-entry on test-connection is by design, so the stripped-credential path in the `!credentials?.password` branch is intended behavior, not a regression._

## 🟠 `bulkGetWithCredentialsStripping` drops the `options` argument

`SavedObjectsClient.bulkGet(objects, options)` takes a second `SavedObjectsBaseOptions` (namespace) argument, but the new wrapper only accepts `objects`:

```ts
const result = await wrapperOptions.client.bulkGet<T>(objects); // options lost
```

The wrapper sits in the chain for any `bulkGet` that includes the data-source type, so a mixed-type `bulkGet` with a `namespace` option would silently lose it. `data-source` is `namespaceType: 'agnostic'` so it's unaffected itself, but co-requested types in the same call could be. `get` and `find` correctly forward their options — only `bulkGet` regressed. Please restore the second param.

## 🟡 Minor
- `any` usage in `getWithCredentialsStripping` (`options?: Record<string, any>`) and `stripCredentials<T>(obj: any): any` — repo guidance is to avoid `any`; `SavedObjectsBaseOptions` and a typed SO shape would be cleaner.
- `stripCredentials` sets `credentials: undefined` rather than deleting the key, so `'credentials' in auth` still reports true. Functionally fine for the security goal, just noting.

## ✅ Looks correct
- Access-check-then-internal-fetch ordering: `getDataSource(scoped)` enforces tenant/workspace permission and throws before `getDataSourceInternal` runs.
- `getDataSourceInternal` fetching without namespace is fine since `data-source` is `namespaceType: 'agnostic'`.

Once `bulkGet` options are restored, this looks good to me.

abbyhu2000 (approved):

ZilongX:
> Thanks for tackling this — the two-layer design (strip credentials from external read APIs + bypass via an internal repository for legitimate decryption) is the right approach. I went through it against the surrounding code and have one blocking regression and one latent bug, plus minor notes.
> ## 🔴 1. Test-connection / fetch-metadata break for existing data sources
> 
> `routes/test_connection.ts` and `routes/fetch_data_source_metadata.ts` call `getDataSourceClient` **without** passing `internalSavedObjects`. When a user edits an existing data source and tests the connection **without re-entering the password**, `configure_client.ts` takes this branch:
> 
> ```ts
> if (dataSourceId && ((type === UsernamePasswordType && !credentials?.password) || ...)) {
>   dataSource = await getDataSource(dataSourceId, savedObjects); // now STRIPPED by Layer 1
>   if (internalSavedObjects) { ... }  // undefined here → not re-fetched
> }
> // requireDecryption stays true → downstream tries to decrypt undefined credentials
> ```
> 
> Because Layer 1 now strips credentials from the scoped client globally, `getDataSource` returns `auth.credentials: undefined` and decryption fails. The PR description says these callers "retain the previous single-fetch behavior with no change in semantics" — but the single fetch itself changed behavior, so the "test connection / fetch metadata on an existing source without retyping credentials" flow now errors.
> 
> **Suggested fix:** plumb `internalSavedObjects` into `getDataSourceClient` from `test_connection.ts` and `fetch_data_source_metadata.ts` as well, and add a test covering the non-credential-field-change branch.
> ## 🟠 2. `bulkGetWithCredentialsStripping` drops the `options` argument
> 
> `SavedObjectsClient.bulkGet(objects, options)` takes a second `SavedObjectsBaseOptions` (namespace) argument, but the new wrapper only accepts `objects`:
> 
> ```ts
> const result = await wrapperOptions.client.bulkGet<T>(objects); // options lost
> ```
> 
> The wrapper sits in the chain for any `bulkGet` that includes the data-source type, so a mixed-type `bulkGet` with a `namespace` option would silently lose it. `data-source` is `namespaceType: 'agnostic'` so it's unaffected itself, but co-requested types in the same call could be. `get` and `find` correctly forward their options — only `bulkGet` regressed. Please restore the second param.
> ## 🟡 Minor
> 
>     * `any` usage in `getWithCredentialsStripping` (`options?: Record<string, any>`) and `stripCredentials<T>(obj: any): any` — repo guidance is to avoid `any`; `SavedObjectsBaseOptions` and a typed SO shape would be cleaner.
> 
>     * `stripCredentials` sets `credentials: undefined` rather than deleting the key, so `'credentials' in auth` still reports true. Functionally fine for the security goal, just noting.
> 
> 
> ## ✅ Looks correct
> 
>     * Access-check-then-internal-fetch ordering: `getDataSource(scoped)` enforces tenant/workspace permission and throws before `getDataSourceInternal` runs.
> 
>     * `getDataSourceInternal` fetching without namespace is fine since `data-source` is `namespaceType: 'agnostic'`.
> 
> 
> Once `internalSavedObjects` is plumbed into the two route handlers and `bulkGet` options are restored, this looks good to me.

This is by design to resolve the very security concern, for existing behavior user would have to re-authenticate.

Maosaic (approved):

src/plugins/data_source/server/saved_objects/data_source_saved_objects_client_wrapper.ts
```diff
@@ -123,15 +126,55 @@ export class DataSourceSavedObjectsClientWrapper {
...
+          ),
+        };
+      }
+      return result;
+    };
+
+    const bulkGetWithCredentialsStripping = async <T = unknown>(
+      objects?: SavedObjectsBulkGetObject[]
+    ) => {
+      const result = await wrapperOptions.client.bulkGet<T>(objects);
```
joshuali925, line 160: does bulkget contain options?

yujin-emma (approved):

PR #12260, opened 2026-06-19
[Bug] Side-nav applications

TackAdam:
### Description

Corrected Behavior no longer goes to hidden page.

https://github.com/user-attachments/assets/55f9825a-910e-4c4a-834a-840eed7b0ce6

## Summary

When a workspace was created with `opensearchDashboards.enableIconSideNav: true`, the post-creation redirect could land on the wrong app — most visibly, on the alerting app when `observability.alertManager.enabled: false`, or on Metrics after that was filtered out.

The redirect now picks the first feature whose nav-link is not feature-flag-disabled, falling back to `features[0]` when no better candidate exists.

## Why this was specific to icon side nav

The two nav-registration paths in dashboards-observability use different orderings:

- Default side nav (`registerDefaultNavGroups`): Alerting is registered with `order: 30`, ahead of nearly everything else.
- Icon side nav (`registerIconSideNavGroups`): Alerting is registered with `order: 50`, also ahead of Notebooks, APM services, and so on.

In both modes, `features[0]` was Alerting whenever its nav link was registered. With `enableIconSideNav: true`, this combined with the icon-side-nav-specific category and ordering decisions to make the problem visible and reproducible. The fix is mode-agnostic but the bug was first reported in icon-side-nav environments.

## Why `features[0]` was wrong

dashboards-observability uses a register-then-hide pattern: the alerting app and its nav-group entry are registered unconditionally at `setup()`, then hidden in `start()` via `AppNavLinkStatus.hidden` when `observability.alertManager.enabled` is false. This is the supported OSD pattern for dynamic config flags.

But `getUseCaseUrl` and the post-create redirect in `workspace_creator.tsx` both consume `useCase.features[0]` directly, with no awareness of which features are actually visible. So a hidden-but-registered feature could still be picked as the landing app.

## Why the filter has to inspect both `navLinkStatus` and `status`

A naive filter on `navLinkStatus === hidden` or on the computed `ChromeNavLink.hidden` over-filters: it also drops `WorkspaceAvailability.insideWorkspace` apps when read from the workspace creator page, which runs outside any workspace.

The workspace plugin's global app updater marks those apps inaccessible while you're outside a workspace; that inaccessibility flips back to accessible the moment the redirect lands inside the new workspace.

We separate the two using both fields together:

| navLinkStatus | status | Meaning | Skip? |
| --- | --- | --- | --- |
| `hidden` | `accessible` | Feature-flag-disabled by the plugin | Yes |
| `hidden` | `inaccessible` | Transiently hidden because we're outside a workspace | No |
| `visible/default` | `inaccessible` | Same — workspace plugin marked it inaccessible | No |
| `visible/default` | `accessible` | Normal case | No |

`isFeatureFlagDisabled = navLinkStatus === hidden && status === accessible` is the discriminator.

A feature-flagged-off app is still notionally accessible — the plugin just hides the UI — so the status stays accessible. A workspace-availability-restricted app has its status flipped to inaccessible, which we treat as a transient state and ignore.

### `src/plugins/workspace/public/utils.ts`

- New exported helper `pickUseCaseLandingAppId(features, apps)` that scans features for the first whose registered app is not feature-flag-disabled, falling back to `features[0]` if none qualify or no apps snapshot is available.
- `getUseCaseUrl` now reads a synchronous snapshot from `application.applications$` — safe because the upstream pipe uses `shareReplay(1)` — and delegates to `pickUseCaseLandingAppId`.

### `src/plugins/workspace/public/components/workspace_creator/workspace_creator.tsx`

- Post-create redirect uses `pickUseCaseLandingAppId` against a synchronous `applications$` snapshot instead of indexing `features[0]` directly.

### `src/plugins/workspace/public/utils.test.ts`

- 5 unit tests for `pickUseCaseLandingAppId` covering: empty/undefined inputs, missing apps snapshot, feature-flag-disabled skip, transient-outside-workspace keep, all-disabled fallback.
- 1 integration test on `getUseCaseUrl` verifying the filter is applied end-to-end through the public API.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/workspace/public/components/workspace_creator/workspace_creator.tsx
+++ b/src/plugins/workspace/public/components/workspace_creator/workspace_creator.tsx
@@ -10,7 +10,7 @@ import { BehaviorSubject } from 'rxjs';
 import { useLocation } from 'react-router-dom';
 
 import { useOpenSearchDashboards } from '../../../../opensearch_dashboards_react/public';
-import { PermissionModeId, PublicAppInfo } from '../../../../../core/public';
+import { PermissionModeId } from '../../../../../core/public';
 import {
   CURRENT_USER_PLACEHOLDER,
   WORKSPACE_COLLABORATORS_APP_ID,
@@ -28,7 +28,11 @@ import { WorkspaceClient } from '../../workspace_client';
 import { DataSourceManagementPluginSetup } from '../../../../../plugins/data_source_management/public';
 import { DataPublicPluginStart } from '../../../../data/public';
 import { WorkspaceUseCase } from '../../types';
-import { getFirstUseCaseOfFeatureConfigs, pickUseCaseLandingAppId } from '../../utils';
+import {
+  getApplicationsSnapshot,
+  getFirstUseCaseOfFeatureConfigs,
+  pickUseCaseLandingAppId,
+} from '../../utils';
 import { useFormAvailableUseCases } from '../workspace_form/use_form_available_use_cases';
 import { NavigationPublicPluginStart } from '../../../../../plugins/navigation/public';
 import { DataSourceConnectionType } from '../../../common/types';
@@ -145,11 +149,7 @@ export const WorkspaceCreator = (props: WorkspaceCreatorProps) => {
             const newWorkspaceId = result.result.id;
             const useCaseId = getFirstUseCaseOfFeatureConfigs(attributes.features);
             const matchedUseCase = availableUseCases?.find(({ id }) => useCaseId === id);
-            let apps: ReadonlyMap<string, PublicAppInfo> | undefined;
-            const appsSub = application.applications$.subscribe((value) => {
-              apps = value;
-            });
-            appsSub.unsubscribe();
+            const apps = getApplicationsSnapshot(application);
             const useCaseLandingAppId = pickUseCaseLandingAppId(matchedUseCase?.features, apps);
 
             // For observability workspaces, run trace detection and create datasets if found
--- a/src/plugins/workspace/public/components/workspace_creator/workspace_creator.tsx
+++ b/src/plugins/workspace/public/components/workspace_creator/workspace_creator.tsx
@@ -10,7 +10,7 @@ import { BehaviorSubject } from 'rxjs';
 import { useLocation } from 'react-router-dom';
 
 import { useOpenSearchDashboards } from '../../../../opensearch_dashboards_react/public';
-import { PermissionModeId } from '../../../../../core/public';
+import { PermissionModeId, PublicAppInfo } from '../../../../../core/public';
 import {
   CURRENT_USER_PLACEHOLDER,
   WORKSPACE_COLLABORATORS_APP_ID,
@@ -28,7 +28,7 @@ import { WorkspaceClient } from '../../workspace_client';
 import { DataSourceManagementPluginSetup } from '../../../../../plugins/data_source_management/public';
 import { DataPublicPluginStart } from '../../../../data/public';
 import { WorkspaceUseCase } from '../../types';
-import { getFirstUseCaseOfFeatureConfigs } from '../../utils';
+import { getFirstUseCaseOfFeatureConfigs, pickUseCaseLandingAppId } from '../../utils';
 import { useFormAvailableUseCases } from '../workspace_form/use_form_available_use_cases';
 import { NavigationPublicPluginStart } from '../../../../../plugins/navigation/public';
 import { DataSourceConnectionType } from '../../../common/types';
@@ -144,8 +144,13 @@ export const WorkspaceCreator = (props: WorkspaceCreatorProps) => {
           if (application && http) {
             const newWorkspaceId = result.result.id;
             const useCaseId = getFirstUseCaseOfFeatureConfigs(attributes.features);
-            const useCaseLandingAppId = availableUseCases?.find(({ id }) => useCaseId === id)
-              ?.features[0].id;
+            const matchedUseCase = availableUseCases?.find(({ id }) => useCaseId === id);
+            let apps: ReadonlyMap<string, PublicAppInfo> | undefined;
+            const appsSub = application.applications$.subscribe((value) => {
+              apps = value;
+            });
+            appsSub.unsubscribe();
+            const useCaseLandingAppId = pickUseCaseLandingAppId(matchedUseCase?.features, apps);
 
             // For observability workspaces, run trace detection and create datasets if found
             const isObservabilityWorkspace = useCaseId === 'observability';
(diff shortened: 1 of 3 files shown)


src/plugins/workspace/public/utils.ts
```diff
@@ -526,13 +527,52 @@ export function prependWorkspaceToBreadcrumbs(
...
 ): string => {
-  const appId = useCase?.features?.[0]?.id || WORKSPACE_DETAIL_APP_ID;
+  let apps: ReadonlyMap<string, PublicAppInfo> | undefined;
+  // `applications$` is wrapped in `shareReplay(1)` upstream, so a
+  // synchronous subscribe-then-unsubscribe yields the current snapshot
+  // without leaking the subscription.
+  const sub = application.applications$.subscribe((value) => {
+    apps = value;
+  });
+  sub.unsubscribe();
```
ps48: This subscribe-then-unsubscribe snapshot block is duplicated almost verbatim in `workspace_creator.tsx`. Since both sites just need "the current `applications$` value," a tiny helper (e.g. `getApplicationsSnapshot(application)`) would keep the `shareReplay(1)` assumption in one place — if it ever changes upstream, there's a single spot to fix.

src/plugins/workspace/public/utils.ts
```diff
@@ -526,13 +527,52 @@ export function prependWorkspaceToBreadcrumbs(
...
+): string | undefined => {
+  if (!features?.length) {
+    return undefined;
+  }
+  if (!apps) {
+    return features[0].id;
+  }
+  const isFeatureFlagDisabled = (app: PublicAppInfo | undefined) =>
+    app?.navLinkStatus === AppNavLinkStatus.hidden && app?.status === AppStatus.accessible;
+  const firstSelectable = features.find((feature) => !isFeatureFlagDisabled(apps.get(feature.id)));
```
ps48, line 583: Worth noting this is load-bearing: the snapshot read depends on `applications$` emitting synchronously on subscribe (true today via `shareReplay(1)` at `application_service.tsx:268,321`). If a future refactor drops `shareReplay`, `apps` is silently `undefined` and the picker falls back to `features[0]` — i.e. it quietly regresses to exactly the bug this PR fixes. The current approach is fine given the explanatory comment; just calling out the silent-failure mode.

src/plugins/workspace/public/utils.ts
```diff
@@ -526,13 +527,52 @@ export function prependWorkspaceToBreadcrumbs(
...
+  features: WorkspaceUseCaseFeature[] | undefined,
+  apps: ReadonlyMap<string, PublicAppInfo> | undefined
+): string | undefined => {
+  if (!features?.length) {
+    return undefined;
+  }
+  if (!apps) {
+    return features[0].id;
+  }
+  const isFeatureFlagDisabled = (app: PublicAppInfo | undefined) =>
```
ps48, line 581: `isFeatureFlagDisabled(undefined)` returns `false`, so a feature whose `id` has no entry in the `apps` map is treated as selectable. That's the right call for the transient-load case, but it's a third state beyond the two in the PR's truth table: a genuinely unregistered/typo'd feature id could be picked and produce a dead link. Likely fine in practice since feature ids come from `convertNavGroupToWorkspaceUseCase` mapping real nav links — just flagging it.

src/plugins/workspace/public/components/workspace_creator/workspace_creator.tsx
```diff
@@ -144,8 +144,13 @@ export const WorkspaceCreator = (props: WorkspaceCreatorProps) => {
...
             const newWorkspaceId = result.result.id;
             const useCaseId = getFirstUseCaseOfFeatureConfigs(attributes.features);
-            const useCaseLandingAppId = availableUseCases?.find(({ id }) => useCaseId === id)
-              ?.features[0].id;
+            const matchedUseCase = availableUseCases?.find(({ id }) => useCaseId === id);
+            let apps: ReadonlyMap<string, PublicAppInfo> | undefined;
+            const appsSub = application.applications$.subscribe((value) => {
+              apps = value;
+            });
+            appsSub.unsubscribe();
```
ps48: Minor: this inline `let apps; subscribe(...); unsubscribe();` reads a little awkwardly and is the duplicate of the block in `utils.ts`. If you extract the helper suggested over there, this collapses to `const apps = getApplicationsSnapshot(application);`.

src/plugins/workspace/public/utils.test.ts
```diff
@@ -967,6 +969,66 @@ describe('workspace utils: mergeDataSourcesWithConnections', () => {
...
+    const apps = new Map<string, PublicAppInfo>([
+      ['dashboards', outsideWorkspaceHidden],
+      ['explore/logs', outsideWorkspaceHidden],
+    ]);
+    expect(pickUseCaseLandingAppId([{ id: 'dashboards' }, { id: 'explore/logs' }], apps)).toBe(
+      'dashboards'
+    );
+  });
+
+  it('falls back to features[0] when every feature is feature-flag-disabled', () => {
```
ps48, line 1023: Nice table-driven coverage of the four states. One case not covered: a feature whose id is **absent from the `apps` map** (the `undefined` lookup path). Since treating that as selectable is load-bearing for the transient-load scenario, a test asserting it would lock the behavior in.

TackAdam:
Thanks for the review — addressed all five comments in the latest push:

### 1. Duplicate snapshot block (`utils.ts` + `workspace_creator.tsx`) — ✅ Fixed

Extracted `getApplicationsSnapshot(application)` into `utils.ts`.

Both call sites now use the helper; the inline `let apps; subscribe; unsubscribe` blocks are gone. The `shareReplay(1)` assumption now lives in exactly one place.

### 2. Load-bearing `shareReplay(1)` dependency — ✅ Documented

The new `getApplicationsSnapshot` helper includes a doc comment that explicitly:

- Names the assumption
- Points to the upstream definition in `application_service.tsx`
- Calls out the silent-regression failure mode if `shareReplay` is ever removed

Centralizing the snapshot read means that regression now has a single place to surface rather than several.

### 3. Third state: feature ID absent from the apps map — ✅ Documented (behavior unchanged)

Added a paragraph to the `pickUseCaseLandingAppId` doc comment explaining why an absent lookup is intentionally treated as selectable.

Feature IDs come from `convertNavGroupToWorkspaceUseCase` over real registered nav links, so an absent lookup means the snapshot has not propagated yet—not that the app does not exist.

Skipping such features would silently skip the entire list during early page load.

### 4. Inline awkwardness in `workspace_creator.tsx` — ✅ Fixed

Resolved via (1).

The block now collapses to:

```ts
const apps = getApplicationsSnapshot(application);
```

Also removed the now-unused `PublicAppInfo` import from the file.

### 5. Test coverage for the absent-from-apps-map path — ✅ Added

Added a new test:

> treats a feature id absent from the apps map as selectable

This locks in the intentional **"absent ≠ disabled"** behavior.

If anyone later changes the picker to skip unknown IDs, this test will catch the regression.

### Validation

- **60 tests passing**
- `src/plugins/workspace/public/utils.test.ts`

ps48 (approved):

Maosaic (approved):

PR #12267, opened 2026-06-23
[Enhancement] Sort datasets alphabetically in dataset pickers

Maosaic:
## Description

The dataset pickers in both the **Explore** (`DatasetSelect`) and **Discover** (`DatasetSelector`) plugins previously rendered datasets in whatever order the saved-objects `find` query returned them (no `sortField` is passed, so effectively backend/relevance order — not alphabetical). This makes datasets hard to scan as the list grows.

This PR sorts datasets alphabetically across all three surfaces. All comparisons are case-insensitive (`localeCompare` with `sensitivity: 'base'`).

### Sort keys (intentionally differ per picker)

The two pickers sort on different keys **by design**, because their option labels are constructed differently:

- **Explore dropdown** and **Explore "View datasets" modal** — sort by the displayed name `displayName || title`. Explore renders the data source in a *subtitle*, not in the label, so the sort key is just the name. (The modal sorts a copy so the `datasets` prop isn't mutated.)
- **Discover dropdown** — Discover's option label *is* the prefixed string `dataSource.title::name`, and we sort on that full label. The effect is that datasets are grouped by **data source first, then by name** — the intended grouping for Discover's multi-cluster list. The **Recently selected** group intentionally keeps its recency order (only the **Index patterns** group is sorted).

In short: Explore orders purely by display name; Discover orders by data-source-prefixed label (cluster-grouped). This divergence is deliberate and matches how each picker presents the data source.

## Issues Resolved

N/A

## Testing

- Added a unit test asserting the "View datasets" modal renders rows in alphabetical order regardless of fetch order.
- `yarn test:jest src/plugins/data/public/ui/dataset_select/dataset_select.test.tsx` — 29 passed, 1 skipped.
- `yarn test:jest src/plugins/data/public/ui/dataset_selector/dataset_selector.test.tsx` — 3 passed.

## Check List

- [x] All tests pass
- [x] New functionality includes unit tests

### DIFF
--- a/src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
+++ b/src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
@@ -132,26 +132,32 @@ export const DatasetSelector = ({
     const buildDatasetOptions = (
       groupLabel: string,
       ds: Dataset[],
-      selectedDatasetId: string | undefined
+      selectedDatasetId: string | undefined,
+      sort: boolean = false
     ): EuiSelectableOption[] => {
-      const datasetOptions: EuiSelectableOption[] = [
-        {
-          label: groupLabel,
-          isGroupLabel: true,
-        },
-      ];
-      ds.forEach(({ id, title, type, dataSource, displayName }) => {
-        const displayLabel = displayName || title;
-        const label = dataSource ? `${dataSource.title}::${displayLabel}` : displayLabel;
-        datasetOptions.push({
-          label,
-          checked: id === selectedDatasetId ? 'on' : undefined,
-          key: id,
-          prepend: <EuiIcon type={datasetService.getType(type)!.meta.icon.type} />,
-          'data-test-subj': `datasetSelectorOption-${id}`,
-        });
-      });
-      return datasetOptions.length > 1 ? datasetOptions : [];
+      const groupLabelOption: EuiSelectableOption = {
+        label: groupLabel,
+        isGroupLabel: true,
+      };
+      const datasetOptions: EuiSelectableOption[] = ds.map(
+        ({ id, title, type, dataSource, displayName }) => {
+          const displayLabel = displayName || title;
+          const label = dataSource ? `${dataSource.title}::${displayLabel}` : displayLabel;
+          return {
+            label,
+            checked: id === selectedDatasetId ? 'on' : undefined,
+            key: id,
+            prepend: <EuiIcon type={datasetService.getType(type)!.meta.icon.type} />,
+            'data-test-subj': `datasetSelectorOption-${id}`,
+          };
+        }
+      );
+      if (sort) {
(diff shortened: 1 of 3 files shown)


src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
```diff
@@ -132,26 +132,32 @@ export const DatasetSelector = ({
...
+            checked: id === selectedDatasetId ? 'on' : undefined,
+            key: id,
+            prepend: <EuiIcon type={datasetService.getType(type)!.meta.icon.type} />,
+            'data-test-subj': `datasetSelectorOption-${id}`,
+          };
+        }
+      );
+      if (sort) {
+        datasetOptions.sort((a, b) =>
+          a.label.localeCompare(b.label, undefined, { sensitivity: 'base' })
```
angle943, line 157: Inconsistent sort key between the two pickers. Here in `buildDatasetOptions` the sort is on `a.label`, which is `${dataSource.title}::${displayLabel}` when a data source is present — so datasets get ordered by **data-source prefix first**, then display name. By contrast, `ViewDatasetsModal` (in `dataset_select.tsx`) sorts purely on `displayName || title`.

The PR title says "by display name," so this divergence is a little surprising. Two questions:
1. Is the data-source-prefix-first ordering here intentional (i.e., group by cluster)?
2. If the goal is true display-name ordering in both places, consider sorting on `displayName || title` rather than the prefixed `label`.

Either way is fine — just want the behavior to be deliberate and consistent across both pickers (and ideally noted in the description).

angle943 (approved):

src/plugins/data/public/ui/dataset_selector/dataset_selector.tsx
Maosaic: Good catch, and thanks for the detailed read. The divergence is intentional:

1. **Yes**, the data-source-prefix-first ordering in Discover is deliberate. Discover's option `label` *is* the prefixed string (`dataSource.title::name`), so sorting on `label` groups datasets by cluster, then by name — which is the intended grouping for Discover's multi-cluster list.
2. Explore is different by design: it renders the data source in a *subtitle* rather than the label, so its label (and sort key) is just `displayName || title`. That's why `ViewDatasetsModal` and the Explore dropdown sort purely on the display name.

So the two pickers sort on different keys because their labels are constructed differently. I've updated the PR description to spell this out explicitly so the behavior is documented and unambiguous. Let me know if you'd prefer both pickers to use pure display-name ordering instead — happy to switch Discover to sort on `displayName || title` if cross-picker consistency is preferred over cluster grouping.

abbyhu2000 (approved):

TackAdam (approved):

tomkdgun (approved):

PR #12270, opened 2026-06-24
chore: use major version in .nvmrc to avoid requiring exact Node.js version

ruanyl:
### Description
Relax `.nvmrc` to major version only for development flexibility. `.node-version` pins the exact runtime version; `.nvmrc` only needs to guide developers to the correct major version.

- Change `.nvmrc` from exact version to major version (22)
- Update build config to read `.node-version` for exact version

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


.nvmrc
```diff
@@ -1 +1 @@
-22.23.0
+22
```
tomkdgun, line 1: @ruanyl Probably src/dev/node_versions_must_match.test.ts will fail

ruanyl: Right, updated, thanks!

SuZhou-Joe (dismissed):

wanglam (dismissed):
Nit: Do we need to update some docs to clarify which files should be changed when upgrading major/minor nodejs versions?

ruanyl:
> Nit: Do we need to update some docs to clarify which files should be changed when upgrading major/minor nodejs versions?

That's a good point, I will add this to `DEVELOPER_GUIDE.md`

wanglam (approved):

SuZhou-Joe (approved):

PR #12274, opened 2026-06-24
feat(ppl-lint): add PPL lint engine with head-without-sort and division-by-zero rules

Hanyu-W:
### Description

Add a PPL lint engine that catches common query pitfalls directly in the editor before execution. Ships two rules as a vertical slice:

- **head-without-sort** (info): flags `head N` with no preceding `sort`, which returns nondeterministic rows depending on shard layout.
- **division-by-zero** (warning): flags `/ 0` literals that silently produce null instead of raising an error.

The implementation spans two layers:

**Lint engine (`@osd/monaco`):** A catalog-driven framework where each rule is a JSON entry (id, severity, docUrl, version constraints) mapped to a detector function that walks the ANTLR parse tree. Includes pipeline-shape analysis, version filtering (minVersion/maxVersion/Calcite predicates), a hover card system with structured Markdown (engine behavior, failure class explanation, safe-to-ignore guidance), and a detector registry extensible via `registerDetector()`.

**Client wiring (`data` + `query_enhancements` plugins):** A lint bridge (`globalThis`-pinned for bundle dedup) connects the editor to either the runtime grammar (hot path — reuses the cached grammar from autocomplete) or a Web Worker fallback (compiled grammar). Context flows from the dataset's data-source version and index fields into the engine. Per-rule overrides (enabled/severity) are read from uiSettings at editor mount. A 500ms trailing-edge debounce prevents redundant lint passes on rapid typing.

### Issues Resolved

Two Rules For https://github.com/opensearch-project/sql/issues/5405

## Screenshot
image
Head-Without-Sort
image
Division-By-Zero
image
Settings

https://github.com/user-attachments/assets/b566bbe6-eaee-494e-b67a-8d07e4f5e370

Quick demo

## Testing the changes

1. **Enable the feature:**
   Add to `opensearch_dashboards.yml`:
   ```yaml
   queryEnhancements.ppl.lint.enabled: true
   ```
   Restart the server.

2. **Trigger head-without-sort:**
   In Explore, select a dataset and enter:
   ```
   source=accounts | head 5
   ```
   Expect an info-level squiggle on `head 5`. Hover shows: "Add sort before head to get stable results." with engine behavior and safe-to-ignore guidance.

3. **Trigger division-by-zero:**
   ```
   source=accounts | eval x = balance / 0
   ```
   Expect a warning squiggle on `0`. Hover shows: "Dividing by zero returns null silently — guard with coalesce() or a conditional."

4. **Verify suppression:**
   ```
   source=accounts | sort age | head 5
   source=accounts | eval x = balance / 2
   ```
   No squiggles.

5. **Run tests:**
   ```bash
   yarn test:jest packages/osd-monaco/src/ppl/lint --no-coverage
   yarn test:jest src/plugins/data/public/ppl_lint --no-coverage
   yarn test:jest src/plugins/data/public/antlr/opensearch_ppl/runtime_lint --no-coverage
   yarn test:jest src/plugins/query_enhancements/server --no-coverage
   yarn test:jest src/plugins/query_enhancements/public/ppl_lint --no-coverage
   ```
   All 161 test cases pass across 16 test files.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest` (lint-related suites)


Hanyu-W:
Updated PR to fix IT failures + github bot issues

packages/osd-monaco/src/ppl/lint/version_filter.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+    return true;
+  }
+
+  const coercedVersion = coerce(dataSourceVersion);
+  if (!coercedVersion) {
+    return appliesTo(rule, undefined, isCalcite, knownVersion);
+  }
+
+  if (isCalciteGated && isCalcite !== true) {
```
RyanL1997, line 46: The `engine: 'calcite'` gate looks like it can never actually fire on a real cluster. When the version is known, this returns `false` unless `isCalcite === true` — but I can't find anywhere that populates `context.isCalcite`. `deriveIsCalcite()` was added to `ppl_grammar_cache.ts` and exported, but it's never called and `buildPPLLintContext` doesn't set it. So today a calcite-gated rule would only fire when the version is *unknown*, which is backwards.

The two shipped rules use `appliesTo: {}` so nothing's broken right now — is the plan to wire `deriveIsCalcite` into the lint context in a follow-up?

packages/osd-monaco/src/ppl/worker/ppl.worker.ts
```diff
@@ -10,30 +10,27 @@ import {
...
 
   async validate(content: string): Promise<PPLValidationResult> {
-    if (!this.analyzer) {
-      this.analyzer = getPPLLanguageAnalyzer();
-    }
     return this.analyzer.validate(content);
   }
+
+  async lint(content: string, overrides?: BundleRuleOverrides): Promise<LintResult> {
+    return this.analyzer.lint(content, overrides ? { overrides } : undefined);
```
RyanL1997: Heads up that the fallback path only forwards `overrides` here — `dataSourceVersion`, `isCalcite`, `fields`, and `typeMap` all get dropped when we fall back to the compiled worker (runtime grammar off / cache miss). Fine for the two version-agnostic rules in this PR, but it means version-gated or `needsContext` rules would silently behave differently on the fallback path vs. the runtime bridge.

Can we either thread the context through?

packages/osd-monaco/src/ppl/lint/rules/head_without_sort.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+    if (sawSort && !ORDER_PRESERVING_COMMANDS.has(stage.command)) {
+      sawSort = false;
+    }
+    if (stage.command === 'headCommand') {
+      if (!sawSort) {
+        diagnostics.push({
+          ruleId: config.id,
+          severity: config.severity,
+          message: 'Add sort before head to get stable results.',
```
RyanL1997: Small inconsistency: this hardcodes the message string, but `division_by_zero.ts` uses `config.message`. They happen to match the catalog today, so an override or catalog edit to this rule's message just won't take effect. Suggest `message: config.message` to match the other detector.

packages/osd-monaco/src/ppl/lint/pipeline_shape.ts
```diff
@@ -0,0 +1,183 @@
...
+      const children = node.children ?? [];
+      for (const child of children) {
+        visit(child);
+      }
+    }
+  };
+  visit(tree);
+
+  for (const stage of stages) {
+    collectCreatedFields(stage, ruleNameToIndex, createdFields);
```
RyanL1997: `createdFields` gets computed on every lint pass, but the only caller (`head_without_sort`) destructures `{ stages }` and ignores it, and nothing else reads it. It's an extra full subtree walk for a currently-unused result. Could we make it lazy (or drop it until a rule needs it)? Not urgent given the 500ms debounce, just flagging the wasted work.

packages/osd-monaco/src/ppl/lint/range_utils.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export function rangeFromContext(ctx: ParserRuleContext): DiagnosticRange {
+  if (!ctx.start) {
+    return { startLine: 1, startColumn: 0, endLine: 1, endColumn: 1 };
+  }
+  return rangeFromTokens(ctx.start, ctx.stop ?? ctx.start);
+}
+
+export function rangeWithinToken(
```
RyanL1997, line 27: Minor: `rangeWithinToken` is exported but isn't used by either shipped detector. Fine as scaffolding for upcoming rules — just flagging it's dead code for now in case it was meant to be wired up.

packages/osd-monaco/src/ppl/lint/hover/hover_card.ts
```diff
@@ -0,0 +1,147 @@
...
+  'engine-throw': 'the engine rejects the query, so it will not run.',
+  nondeterministic:
+    'the query runs, but the rows it returns are not stable across identical re-runs.',
+  fallback:
+    'the primary engine cannot run this natively and falls back to a secondary engine — it succeeds, but on a slower path.',
+  advisory:
+    'the query runs and may return data, but the command can behave differently than intended on this input — this is a heads-up, not a guaranteed outcome.',
+};
+
+function escapeInline(text: string): string {
```
RyanL1997, line 44: Minor: `escapeInline` doesn't escape `(`, `)`, `#`, or `!`. Harmless here given `isTrusted: false` and the inline-only context, but a one-line comment noting the intended scope would save the next reader from wondering whether it's an oversight.

packages/osd-monaco/src/ppl/language.ts
```diff
@@ -10,37 +10,39 @@ import { getPPLLanguageAnalyzer, PPLValidationResult } from './ppl_language_anal
...
-// PPL worker proxy service for worker-based syntax highlighting
 const pplWorkerProxyService = new PPLWorkerProxyService();
 
-// PPL analyzer for synchronous tokenization (lazy initialization)
 let pplAnalyzer: ReturnType<typeof getPPLLanguageAnalyzer> | undefined;
 
-/**
- * Map PPL Language Analyzer tokens to Monaco editor token classes
- * Based on ANTLR-generated token types from OpenSearchPPLLexer
- */
```
RyanL1997, line 26: why we are removing these comments?

packages/osd-monaco/src/ppl/language.ts
```diff
@@ -168,111 +155,177 @@ const processSyntaxHighlighting = async (model: monaco.editor.IModel) => {
 
       monaco.editor.setModelMarkers(model, OWNER, markers);
     } else {
-      // Clear markers if no errors
       monaco.editor.setModelMarkers(model, OWNER, []);
     }
-  } catch (error) {
-    // Silent error handling - continue without worker-based highlighting
+  } catch {
+    // best-effort
```
RyanL1997: can you elaborate more on this change?

RyanL1997 (commented):
Hi @Hanyu-W, thanks for the change. I just left some comments and also we may need to clean up some un related change.

packages/osd-monaco/src/ppl/language.ts
Hanyu-W: You're right, I was too overzealous with my refactoring - restored to keep the diff scoped to the linting.

packages/osd-monaco/src/ppl/language.ts
Hanyu-W: > You're right, I was too overzealous with my refactoring - restored to keep the diff scoped to the linting.

Same as above

packages/osd-monaco/src/ppl/lint/hover/hover_card.ts
Hanyu-W: Confirmed harmless. Added a comment documenting the intended scope.

packages/osd-monaco/src/ppl/lint/range_utils.ts
Hanyu-W: Yeah this is scaffolding for upcoming rules. I've added unit tests covering each helper to give them real callers now without contorting the shipped detectors.

packages/osd-monaco/src/ppl/lint/pipeline_shape.ts
Hanyu-W: I agree, dropping it for now. It will land in the next pr for the field-aware rule

packages/osd-monaco/src/ppl/lint/rules/head_without_sort.ts
Hanyu-W: Fixed. head_without_sort now reads config.message to match division_by_zero

packages/osd-monaco/src/ppl/worker/ppl.worker.ts
Hanyu-W: Fixed. The compiled-worker fallback now forwards the full serializable context (everything except http, which isn't structured-cloneable and isn't consumed by runLint). Sets/Maps are serialized to arrays/objects for postMessage and reconstituted on the worker side. Both paths now behave identically for any rule.

packages/osd-monaco/src/ppl/lint/version_filter.ts
Hanyu-W: Good catch. buildPPLLintContext never set isCalcite, so the version-filter gate was unwired. I've added isCalcite: deriveIsCalcite(dsVersion) and a unit test. No shipped rule is affected (both use appliesTo: {}), but this makes the gate correct for the first calcite-gated rule.

Hanyu-W:
Thanks Ryan for the review! Main changes in the follow-up:

1. Reverted all unrelated refactoring of pre-existing code (comment deletions, style compressions, guard removals across 7 files) — the diff now only touches what the lint feature needs
2. Restored all pre-existing comments (68 lines across language.ts, ppl.worker.ts, ppl_language_analyzer.ts, ppl_grammar_cache.ts, opensearch_ppl_autocomplete.ts, use_query_panel_editor.ts, ppl_error_listener.ts)
3. Wired deriveIsCalcite into buildPPLLintContext + test
4. Switched head_without_sort to config.message
5. Dropped unused createdFields walk
6. Added unit tests for range_utils helpers
7. Added a demo recording

packages/osd-monaco/src/ppl/lint/pipeline_shape.ts
```diff
@@ -0,0 +1,124 @@
...
+): Set<ParserRuleContext> {
+  const subtrees = new Set<ParserRuleContext>();
+
+  for (const node of findAllDescendantsByRule(tree, ruleNameToIndex, 'lookupCommand')) {
+    subtrees.add(node);
+  }
+
+  for (const node of findAllDescendantsByRule(tree, ruleNameToIndex, 'appendCommand')) {
+    if (findAllDescendantsByRule(node, ruleNameToIndex, 'searchCommand').length > 0) {
+      subtrees.add(node);
```
mengweieric, line 104: Since we add the whole appendCommand node to the alt-source set here, headWithoutSortDetector skips the outer append stage too, so it never resets sawSort. That means `source=logs | sort age | append [ search source=other ] | head 5` won't be flagged even though the result isn't the sorted pipeline anymore. Same thing for lookup (added unconditionally just above), so it's not append-specific. I think we want to prune the inner bracketed pipeline but still treat the outer command as order-destroying. If skipping nested pipelines entirely is intentional for this first PR, could we note it as top-level-only?

packages/osd-monaco/src/ppl/lint/rules/head_without_sort.ts
```diff
@@ -0,0 +1,68 @@
...
+  'parseCommand',
+  'grokCommand',
+  'rexCommand',
+  'spathCommand',
+  'fillnullCommand',
+  'expandCommand',
+  'flattenCommand',
+  'patternsCommand',
+  'renameCommand',
+]);
```
mengweieric, line 34: reverseCommand isn't in this set, so `source=logs | sort age | reverse | head 5` gets flagged even though the order after reverse is still deterministic. Should we add reverseCommand here, or is treating reverse as resetting the sort guarantee intentional? A comment either way would help.

packages/osd-monaco/src/ppl/lint/lint_runner.ts
```diff
@@ -0,0 +1,86 @@
...
+  const effectiveOverrides = bundleOverrides ?? context?.overrides;
+
+  for (const localConfig of catalog) {
+    const config = mergeConfig(localConfig, effectiveOverrides?.[localConfig.id]);
+
+    if (
+      !config.enabled ||
+      config.needsExplain ||
+      !appliesTo(config, dataSourceVersion, context?.isCalcite, knownVersion) ||
+      (config.needsContext && isContextEmpty(context))
```
mengweieric, line 65: Separate from the fallback-context fix: this skip block never looks at config.runtimeOnly, so a runtimeOnly rule would still run on the compiled-simplified surface where the parse shape can differ. The context already carries grammarSurface, so could we gate it here, something like `config.runtimeOnly && context?.grammarSurface !== 'runtime-bundle'` -> continue, with a test that a runtime-only rule runs on runtime-bundle but not on the compiled fallback? No shipped rule sets runtimeOnly today, so this is just to keep the flag honest before one does.

Hanyu-W: Done — used context?.grammarSurface !== 'runtime-bundle' so it also skips when no surface is set (safe default). Added a test with a synthetic runtimeOnly rule confirming it runs on runtime-bundle but not on the compiled fallback.

packages/osd-monaco/src/ppl/lint/rules/head_without_sort.ts
Hanyu-W: Agreed, reverse flips collation deterministically — added it. I also validated the full allowlist against a live 3.8 cluster and found streamstats, dedup, bin, and regex were missing (same class of false positive), so I folded those in with a comment explaining the principle. One surprise from the live run: eventstats … by <field> actually loses input sort order, so I left it out and added a test asserting it still flags.

packages/osd-monaco/src/ppl/lint/pipeline_shape.ts
Hanyu-W: Good catch — and you're right, lookup has the same issue. The bug is that isInsideAltSource starts its ancestor walk at the node itself, so a top-level append/lookup matches itself in altRoots and gets skipped before sawSort resets. Fix: start the walk at node.parent. The inner sub-pipeline is still pruned, but the outer command now participates in the order analysis. Both are genuinely order-destroying (append → UNION ALL with no post-union sort; lookup → LEFT JOIN with no collation), so sort | append [...] | head and sort | lookup | head correctly flag now. Existing appendcol tests still pass.

Hanyu-W:
@mengweieric thanks for the review! Here's what I changed based on your feedback:

1. append/lookup false negative: isInsideAltSource was starting its ancestor walk at the node itself, so top-level append/lookup matched themselves in altRoots and got skipped before sawSort could reset. Fixed by starting at node.parent — inner sub-pipelines are still pruned, but the outer command now participates in order analysis as it should.

2. reverse false positive: Added reverse to ORDER_PRESERVING_COMMANDS. While validating, I checked the full set against live clusters and found streamstats, dedup, bin, regex, and table were also missing (same bug class), so I folded those in too. Left eventstats out — live testing showed eventstats … by <field> actually loses sort order, so flagging it is correct. 

3. runtimeOnly guard: Added the check you suggested — config.runtimeOnly && context?.grammarSurface !== 'runtime-bundle' → skip. Used !== 'runtime-bundle' so it also skips when no surface is set. 

Please re-review when you get a chance!

joshuali925:
### Code review

Found 2 issues:

1. `createParserWithErrorHandling` returns a `lexerErrorListener` that is never attached to a lexer, so lexer/tokenization errors are silently dropped on the lint path. The JSDoc says it configures "error listeners" (plural) and the return type promises a `lexerErrorListener`, but the body only calls `parser.addErrorListener(parserErrorListener)` — the function receives a `tokenStream`, not a lexer, and `lexer.addErrorListener` is never called. The returned `lexerErrorListener.errors` is always empty dead state.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/baee682942fedab22e8544ef207c657a36714911/packages/osd-monaco/src/ppl/ppl_language_analyzer.ts#L73-L81

2. Stale comment: it claims "A follow-up PR will have the public plugin read `capabilities.queryEnhancements.pplLint` ... nothing consumes this capability yet" — but this same PR adds that consumer in `public/plugin.tsx`, so the comment is inaccurate as of this changeset.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/baee682942fedab22e8544ef207c657a36714911/src/plugins/query_enhancements/server/plugin.ts#L54-L61

The consumer added in this PR:

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/baee682942fedab22e8544ef207c657a36714911/src/plugins/query_enhancements/public/plugin.tsx#L338-L340

packages/osd-monaco/src/ppl/lint/catalog.ts
```diff
@@ -0,0 +1,97 @@
+/*
```
joshuali925, line 1: is there a reason that PPL rules are in monaco package, but not implemented separately and consumed by monaco?

Hanyu-W:
@joshuali925 thanks for the review! Changes in this revision:

1. Removed dead lexerErrorListener in ppl_language_analyzer.ts — it was constructed but never attached to the lexer, so it was always-empty state. Fixed the JSDoc to match. 
3. Replaced stale comment in server/plugin.ts that said "nothing consumes this capability yet" — this PR adds the consumer in public/plugin.tsx.
4. Added @osd/monaco/ppl-lint subpath (redirect-stub directory, same pattern as @osd/i18n/react) exposing a Monaco-free engine barrel. runtime_lint.ts now imports from this stable entry point instead of reaching into @osd/monaco/target/ppl/lint/*.

Please re-review when you get the chance.

packages/osd-monaco/src/ppl/lint/catalog.ts
Hanyu-W: The rules live in @osd/monaco for co-location with the ANTLR grammar and tree utilities — the lint detectors walk the same parse tree that syntax highlighting uses. rule_index, range_utils, and pipeline_shape are already here because that's where the PPL grammar lives. Putting rules in a separate package would mean either extracting the tree utilities into a shared @osd/ppl-ast package or duplicating them across packages.

That said, the engine itself is Monaco-free — only the adapter layer (diagnostic_to_marker.ts, hover provider/registry) touches Monaco. I've added a @osd/monaco/ppl-lint subpath (redirect-stub directory, same pattern as @osd/i18n/react) that exposes just the engine without the Monaco barrel, so non-Monaco consumers like the data plugin can import from a stable entry point.

joshuali925: > rule_index, range_utils, and pipeline_shape are already here because that's where the PPL grammar lives

aren't PPL grammar in the osd-antlr-grammar package?

but i got your point

joshuali925 (dismissed):
there are some conflicts

mengweieric (dismissed):
just pending resolving of conflicts.

Hanyu-W:
@mengweieric @joshuali925 Thank y'all so much for the approvals, I have fixed the merge conflicts, should be ready to merge now.

joshuali925 (approved):

mengweieric (approved):

Issue #11, opened 2021-02-03
[CI] Enable CI for project


mihirsoni:
- [x] Enable initial building on CI
- [x] Enable Unit test cases
- [x] Enable Integration tests
- [ ] Enable building distributions.

PR #12285, opened 2026-06-25
Add per-item hover navPopover system to the icon side navigation

### LINKED ISSUE #11545: RFC: Icon-based Side Navigation for OpenSearch Dashboards
## Problem Statement

The observability workspace sidebar has **30+ items across 8+ categories**, requiring scrolling to see everything. The nav suffers from:

- **No visual anchors** -- every item is plain text; users must read every label to find what they need
- **Collapsed state is blank** -- collapsing the nav shows an empty 48px column with zero navigation affordance
- **No dock/undock** -- the nav is either permanently pinned open (270px) or fully collapsed; no "peek and auto-close"
- **Redundant categories** -- groupings reflect plugin boundaries, not user workflows (e.g., "Metrics" appears in two places)

| Current nav (top) | Current nav (bottom) |
|:-:|:-:|
|Image|Image)  |

## Proposal

Add an icon-based side navigation mode for observability workspaces that gives every nav item an icon, restructures items into fewer meaningful categories, provides a usable collapsed icon strip with popover menus, and introduces dock/undock behavior with auto-close on navigation. All changes are feature-flagged and scoped exclusively to observability workspaces.

## Requirements

| Requirement | Detail |
|-------------|--------|
| Feature-flagged | New `home:enableIconSideNav` setting (default `false`, requires page reload) |
| Dual-gate activation | Both `enableIconSideNav` AND `useNewHomePage` must be `true` |
| Observability-only | Only activates when `currentNavGroup.id === 'observability'` |
| Other workspaces unchanged | SIEM, Essentials, Search, admin workspaces see zero differences |
| Flag-off = identical behavior | When disabled, nav renders exactly as it does today |
| Non-breaking API changes | All type extensions are optional fields on existing interfaces |

### A1: Dedicated Feature Flag

**Decision**: New `home:enableIconSideNav` boolean, separate from `useNewHomePage`.

**Why**: Icon nav is a significant UX change that should be independently toggleable. Coupling it with `useNewHomePage` would force an all-or-nothing rollout. Separate flags let us enable the new home page without icon nav, or vice versa, and provide a clean rollback path.

### A2: Observability-Only Scoping

**Decision**: The flag is a global boolean, but the visual behavior is scoped at the `header.tsx` consumption point via `currentNavGroup?.id === 'observability'`.

**Why**: Scoping at `header.tsx` (not deeper) is optimal because that's where the nav open/close state model lives. Computing `effectiveEnableIconSideNav` there means the state model, auto-close logic, and child component rendering all use the same scoped value. Other workspaces need their own icon/category design work before they can adopt this pattern.

### A3: Type Extensions

Two existing interfaces gain optional fields:

**`AppCategory`** (in `src/core/types/app_category.ts`):
- `euiIconType?: EuiIconType` -- icon for the category header
- `collapsible?: boolean` -- whether the category can be collapsed in the side nav
- `defaultOpen?: boolean` -- initial expand/collapse state (defaults to `true`)

**`ChromeRegistrationNavLink`** (in `nav_group_service.ts`):
- `euiIconType?: EuiIconType` -- per-registration icon (overrides app-level icon)
- `badge$?: Observable<number | string | undefined>` -- live badge observable

**Why per-registration, not per-app**: The same app may need different icons in different workspaces. Registration-level control gives plugins full flexibility without affecting other nav groups.

### A4: Collapsed Icon Strip

**Decision**: When the nav is collapsed, render a vertical strip of icon buttons instead of a blank column.

| Item type | Collapsed behavior |
|-----------|-------------------|
| Simple nav link (e.g., Dashboards) | Single icon button, direct click navigates |
| Parent link with children (e.g., Alerting) | Icon button + popover listing children on click |
| Non-collapsible category (e.g., App Monitoring) | Children flattened as individual icon buttons |
| Collapsible category (e.g., Settings) | Single icon button + popover listing children |

All icons show `EuiToolTip` on hover (position: right). Active item highlighted with `color="primary"`, `display="base"`.

### A5: Dock/Undock Model

**Decision**: Decouple nav visibility from the lock state using a new `isTempExpanded` state.

| State | Behavior |
|-------|----------|
| **Docked** (`isLocked=true`) | Nav stays expanded at 270px permanently. Pin icon shows as filled. |
| **Undocked** (`isLocked=false`) | Nav collapses to 48px icon strip. Hamburger or icon interaction sets `isTempExpanded=true`. |
| **Auto-close** | When undocked + temp-expanded, any app navigation (`currentAppId$` change) collapses back to icon strip. |

The formula: `isNavOpen = isLocked || isTempExpanded` (only when icon side nav is active).

### A6: Pin Button

**Decision**: Add a pin/unpin toggle (`pinFilled`/`pin` icon) in the expanded nav header, next to the existing collapse button.

Only visible when `enableIconSideNav` is active. Toggles `isLocked` state. Provides a discoverable way to switch between docked and undocked modes.

### A7: Category Restructuring

**Decision**: Introduce 3 new categories to replace the current 8+ fragmented groups:

| Category | Purpose | Behavior |
|----------|---------|----------|
| `applicationMonitoring` (order 4000) | Groups Traces + Services | Static section divider (not collapsible) |
| `observabilityTools` (order 5000) | Groups Notebooks, Anomaly Detection, Forecasting | Static section divider |
| `observabilitySettings` (order 9500) | Groups all workspace management items | **Collapsible, collapsed by default** |

Settings starts collapsed because users access it infrequently -- reduces noise for day-to-day workflows.

### A8: Plugin Registration Pattern

All plugins follow the same pattern: read the flag in `setup()`, conditionally register with `euiIconType`, modified `order`, and restructured `category` for the observability nav group only. Other nav groups are untouched.

```
// Pseudocode pattern (not actual implementation)
const enableIconSideNav = core.uiSettings.get('home:enableIconSideNav', false);
addNavLinksToGroup(observability, [{
  id: appId,
  order: enableIconSideNav ? newOrder : existingOrder,
  category: enableIconSideNav ? newCategory : existingCategory,
  ...(enableIconSideNav ? { euiIconType: 'someIcon' } : {}),
}]);
```

### A9: Per-Registration Icons

**Decision**: Icons are assigned via `euiIconType` on `ChromeRegistrationNavLink`, not on the app definition.

**Why**: This allows the same app (e.g., Explore) to register as "Logs" with `discoverApp` icon, "Metrics" with `stats` icon, and "Traces" with `apmTrace` icon -- all pointing to the same underlying app but appearing as distinct nav items in the observability workspace.

### A10: Badge Observable for Live Indicators

**Decision**: `badge$: Observable<number | string | undefined>` on `ChromeRegistrationNavLink`.

A `NavLinkBadge` component subscribes to the observable and renders an `EuiBadge` (color: accent). Suppresses render for `undefined`, `0`, and `''`. Primary use case: alerting plugin showing active alert count next to the Alerts nav item.

## New Visual Design POC

https://github.com/user-attachments/assets/39d2bc1b-9110-4096-80e1-3350693a3824

## Nav Structure

| Nav Item | Icon | Order | Category | Notes |
|----------|------|-------|----------|-------|
| Dashboards | `dashboardApp` | 100 | _(top-level)_ | |
| Alerts | `navAlerting` | 200 | _(top-level)_ | Parent link; badge$ for alert count |
| Logs | `discoverApp` | 300 | _(top-level)_ | Explore flavor or Discover fallback |
| Metrics | `stats` | 400 | _(top-level)_ | Explore flavor |
| Topology Map | `globe` | 500 | _(top-level)_ | dashboards-observability plugin |
| Agent Traces | `uptimeApp` | 600 | _(top-level)_ | |
| Traces | `apmTrace` | 4050 | Application Monitoring | Explore flavor |
| Services | `graphApp` | 4200 | Application Monitoring | dashboards-observability plugin |
| Notebooks | `notebookApp` | 5100 | Tools | dashboards-observability plugin |
| Anomaly Detection | `machineLearningApp` | 5200 | Tools | Parent link with children |
| Forecasting | `visLine` | 5300 | Tools | |
| _(settings items)_ | `gear` | 9500+ | Settings | Collapsed by default |

---

## Plugin Impact Summary

| Plugin | Change |
|--------|--------|
| **dashboard** | Add `euiIconType: 'dashboardApp'`, reorder to 100 in observability |
| **discover** | Add `euiIconType: 'discoverApp'`, rename to "Logs" (when explore not present) |
| **explore** | Split into 3 flavor registrations: Logs, Metrics, Traces with distinct icons |
| **visualize** | Remove from observability nav (replaced by Notebooks + dedicated tools) |
| **data_explorer** | Remove from observability nav |
| **agent_traces** | Add `euiIconType: 'uptimeApp'`, reorder to 600 |
| **workspace** | Use `observabilitySettings` category for management items in observability |
| **alerting** _(external)_ | Add icon, badge$, move to top-level order 200 |
| **dashboards-observability** _(external)_ | Add icons to Topology Map, Services, Notebooks; use new categories |
| **anomaly-detection** _(external)_ | Add icons; move to `observabilityTools` category |

---

## What's NOT Changing

- **Legacy `CollapsibleNav`** -- not touched; only `CollapsibleNavGroupEnabled` is modified
- **Non-observability workspaces** -- SIEM, Essentials, Search, admin: zero visual changes
- **`useNewHomePage: false` mode** -- identical behavior to today
- **Existing `ChromeNavLink` or `AppCategory` consumers** -- new fields are optional; no existing code breaks
- **Core API compatibility** -- optional field additions are non-breaking; API review files updated via `yarn docs:acceptApiChanges`
- **localStorage schema** -- new keys use `core.navGroup.<id>` prefix, no collision
- **Plugin registrations for non-observability groups** -- only the observability nav group registrations change

---

## Open Questions

1. What are the right set of icons to be used for all the plugins in the observability workspace?
2. Is a popover container the right UX pattern hierarchical flows?
3. Does the current implementation of extending the current side-nav registration align with maintainers?

ps48:
### Description

This change adds a generic, plugin extensible hover popover system (navPopover) to the icon side navigation. Any plugin can attach quick actions and contextual content to a nav item without the item losing its normal click to navigate behavior.

A nav item can now declare a navPopover with two optional parts. The first is a list of declarative action buttons, each with an id, label, icon, and an onClick callback. The second is a render escape hatch for custom content such as a list of recently visited items. Callbacks receive a small set of core services (navigateToApp, basePath, http, recentlyAccessed) so plugins do not need to re-resolve them.

The popover renders in both the expanded and the collapsed (icon rail) states. In the collapsed state a category flattens into a popover, and a nav item with children renders a cascading secondary popover that sits flush against the primary panel. The selected item is highlighted consistently in both states, and the trail to the active page stays highlighted when its sub popover is open.

Popover actions only receive navigateToApp, not overlays, so to open an in app overlay (for example a saved search flyout) an action navigates with a transient hash marker and the destination page opens the overlay on arrival. This is encapsulated in a shared useOpenOnUrlMarker hook in opensearch_dashboards_utils, which reads the marker, opens the overlay, and strips the marker, with a short cooldown to absorb the app's own hash re-sync.

Other changes in this PR:

- Wire navPopovers for Dashboards (create, view all, recent dashboards) and the explore Logs, Traces, and Metrics flavors (new search, browse saved, recent searches), plus Agent Traces and Agent Spans.
- Move the keyboard shortcuts entry into the header Help menu instead of a standalone footer rail icon.
- Use the dockedLeft icon for both the expand and collapse nav controls, and refine the chrome control grouping.
- Add the alwaysUseDefaultOpen category option so the Manage workspace section starts collapsed on each load.
- Keyboard accessibility: the popover opens on focus and closes on Escape, so its actions are reachable without a mouse.

### Issues Resolved

#11545 

## Screenshot

To be added.

## Testing the changes

1. Start the dev server with the icon side navigation enabled.
2. Hover each nav item with a registered popover and confirm its actions render in both the expanded and the collapsed states.
3. Click each action and confirm it navigates or opens the expected overlay.
4. Confirm Browse saved searches opens the saved search flyout, and that leaving the app and returning via its nav link does not reopen the flyout.
5. Tab to a nav item and confirm the popover opens on focus.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


TackAdam:
Review
## 🟠 High

### 1. Expanded-nav parent links: non-first children become unreachable without hover

**File:** `src/core/public/chrome/ui/header/expanded_side_nav.tsx` (~L95–145, `CollapsibleNavItem`)

The old `CollapsibleNavItem` rendered an in-place `useState`-driven accordion (`.obs-nav-children`) listing all child links. The new version removes that: the parent row now navigates to `firstChildId` on direct click, and the other children are reachable only via the hover popover.

- **Failure scenario:** On a touch device (no hover) a user can no longer see or reach any child except the first — clicking the parent jumps straight to `firstChildId`, and the popover never opens. Keyboard users are OK (focus opens the popover), but touch users lose access.
- **Proposed fix:** Either keep an accessible disclosure for touch (e.g. open the popover on click/tap for parent rows instead of navigating), or confirm the observability nav has no multi-child parent links in the expanded state and document that constraint.

---

## 🟡 Medium

### 2. `buildPopoverTitle` / `toSentenceCase` lowercase acronyms inside titles

**Files:**

- `collapsed_side_nav.tsx` L561-567 (`buildPopoverTitle`)
- `expanded_side_nav.tsx` L2244-2246 (`toSentenceCase`)

```ts
return phrase.charAt(0) + phrase.slice(1).toLowerCase();
```

This lowercases everything after the first char, mangling intentional acronyms/capitals.

- **Failure scenario:** A leaf `"APM Traces"` under `"Application Performance"` → popover title `"Application performance apm traces"`; category labels like `"ML Commons"` / `"PPL"` → `"Ml commons"` / `"Ppl"`.
- **Proposed fix:** Only lowercase a token when it isn't an all-caps acronym, e.g. lowercase per word with `if (word === word.toUpperCase())` keep, or restrict the transform to the first word of the category label rather than the whole phrase.

### 3. `metricsPageMode` is read by substring regex over the rison `_a` value

**File:** `src/plugins/explore/public/application/pages/metrics/metrics_page.tsx` L4304-4316

```ts
if (/metricsPageMode:query/.test(aParam)) {
  dispatch(setMetricsPageMode('query'));
} else if (/metricsPageMode:explore/.test(aParam)) {
  ...
}
```

`aParam` is the whole rison `_a` blob, so the regex can match the token embedded in user data.

- **Failure scenario:** A saved query/filter whose text contains `metricsPageMode:query` (persisted into `_a`) force-flips the page into query mode on mount/hashchange, against the user's actual selection.
- **Proposed fix:** Rison-decode `_a` and read `ui.metricsPageMode` as a structured field rather than substring-matching the serialized string.

### 4. Renamed `data-test-subj`s will break external functional tests

**File:** `collapsible_nav_group_enabled_top.tsx` L107-160

`collapsibleNavToggleButton` was split into `collapsibleNavCollapseButton` (expanded) and `collapsibleNavExpandButton` (collapsed). The shared `collapsibleNavToggleButton` handle no longer exists in the icon-side-nav path.

- **Failure scenario:** Cypress/FTR specs in `opensearch-project/opensearch-dashboards-functional-test` selecting `collapsibleNavToggleButton` will fail to find the element.
- **Proposed fix:** Grep the functional-test repo for `collapsibleNavToggleButton` / `collapsibleNavLockButton` and open a companion PR, or note the rename in the PR description for the FTR maintainers.

### 5. Collapsed parent icon now navigates on click; collapsed category icon still does nothing

**File:** `collapsed_side_nav.tsx`

`CollapsedParentIcon` (L748) added:

```ts
onClick={() => navTargetId && navigateToApp(navTargetId)}
```

while `CollapsedCategoryIcon` (L826+) has no `onClick`.

- **Failure scenario:** Two visually-similar collapsed icons behave differently — clicking a parent icon navigates to its first child (a behavior change from the old inert hover-only trigger), clicking a category icon is a no-op. Inconsistent and possibly surprising.
- **Proposed fix:** Make the two consistent — either both navigate to their first child on click, or both are pure popover triggers.

### 6. `useOpenOnUrlMarker` cooldown flag can latch true (latent)

**File:** `src/plugins/opensearch_dashboards_utils/public/react/use_open_on_url_marker.ts` L48-49, 58, 62-64, 94

The reset timer is the only thing that sets `cooldownRef.current = false`. The effect cleanup unconditionally `clearTimeout`s it, and on re-run `openIfMarked()` early-returns once the marker is stripped — without re-arming the reset.

- **Verdict:** PLAUSIBLE. Not reachable via the current explore flow (the strip uses `replaceState`, which react-router doesn't observe, so `locationKey` doesn't change inside the cooldown; `onOpen` is a stable `useCallback`). But if any consumer triggers a react-router navigation within `cooldownMs` of an open (e.g. an overlay that syncs state to the URL), the flag sticks true and "Browse saved searches" silently stops opening for the rest of the session.
- **Proposed fix:** Re-arm defensively — in the cleanup, reset `cooldownRef.current = false` when clearing the timer, or move the cooldown to a timestamp comparison (`lastOpenedAt`) instead of a boolean + timer that the cleanup can orphan.

---

## 🟢 Low

### 7. Recent-items filter uses substring `includes(appPath)` — prefix collision

**Files:**

- `explore/public/nav_popover.tsx` L4488-4503
- `dashboard_nav_popover.tsx` L4784

Uses:

```ts
item.link.includes(appPath)
```

A future sibling app whose path is a superstring (e.g. `/app/explore/logspatterns` vs `/app/explore/logs`) would surface under the wrong flavor.

- **Proposed fix:** Match on a path boundary (`/app/${appId}#` or `/app/${appId}/`).

### 8. Popover action `onClick` silently swallowed when `services` is undefined

**File:** `nav_item_popover.tsx` L3107-3110

```ts
onClick={() => services && action.onClick(services)}
```

If actions are rendered without `services`, clicks do nothing (no nav, no error, popover lingers).

- **Proposed fix:** Don't render action buttons when `services` is absent, or assert/log.

### 9. `metricsHashPath(mode?: 'query')` ignores the argument's value

**File:** `explore/public/nav_popover.tsx` L4540-4552

The truthy branch hardcodes the literal `metricsPageMode:query` instead of interpolating `mode`; correct only because the sole truthy value equals the literal.

- **Proposed fix:** Use `metricsPageMode:${mode}`, or rename the parameter to a boolean `isQueryMode` to match what it actually is.

### 10. Dev Tools button `aria-label` changed `go-to-dev-tools` → localized `"Developer tools"`

**Files:**

- `dev_tools_icon.tsx`
- `dev_tools_icon.test.tsx` L40

The old hardcoded label doubled as a stable, locale-independent automation/a11y handle; queries keyed on it (or expecting a fixed string in non-English locales) break.

- **Proposed fix:** Keep a stable `data-test-subj` for automation (`openDevToolsModal` already exists, so likely fine) — just confirm no external test relies on the `aria-label`.

### 11. Non-null assertions in new production code (`CLAUDE.md`)

**File:** `nav_item_popover.tsx`

- L3066 (`item.children!`)
- L3128 (`childItems!`)

`CLAUDE.md` states: **"Avoid non-null assertions (!)."**

Both are directly preceded by length guards, so they're removable by destructuring the guarded local first.

> **Note:** `popover.actions!`, `.parentElement!`, and `@ts-expect-error` also appear, but only in test files, and `CLAUDE.md` specifically names `@ts-ignore`, so they're not flagged.

### 12. Duplicated logic worth sharing (reuse)

- `flattenChildItems` (expanded) vs `buildChildItems` / `actionsAsChildren` (collapsed) — near-identical `LinkItem → NavPopoverChildItem` flattening in two files.
- Recent-items `subscribe → filter → slice(0,5) → map → state` effect duplicated in `dashboard_nav_popover.tsx` and `explore/nav_popover.tsx` (same magic `5`).
- Hash-query parsing (`indexOf('?') + URLSearchParams`) reimplemented in `use_open_on_url_marker.ts`, `metrics_page.tsx`, and `explore/plugin.ts`.

- **Proposed fix:** Extract a shared `buildPopoverChildItems(...)`, a `useRecentItems(recentlyAccessed$, { linkMatch, limit, map })` hook, and a `getHashQueryParam(hash, key)` helper in `opensearch_dashboards_utils`.

### 13. `NavPopoverConfig.render` invoked unmemoized on every popover render

**File:** `nav_item_popover.tsx` L3096

Documented in the JSDoc, and bounded today because plugin implementations defer fetching to effects — but the contract offloads memoization onto every consumer.

- **Proposed fix (optional):** `useMemo` the `render(services)` result keyed on `services`.

TackAdam (dismissed):

ps48:
Thanks for the thorough review, @TackAdam. Pushed `48c05f5` addressing the findings. Point by point:

**Addressed in code**

- **`#2` (acronym casing):** `toSentenceCase` now preserves all-caps tokens (APM/PPL/ML) instead of lowercasing the whole phrase, and `buildPopoverTitle` reuses it so the collapsed popover title is acronym-safe too. Added a test asserting "APM Services" renders as "APM services".
- **`#3` (metricsPageMode substring over rison `_a`):** confirmed reproducible (a query value containing the literal `metricsPageMode:query` flipped the page mode). Now rison-decodes `_a` and reads `ui.metricsPageMode` as a structured field. Added tests for the structural read, the embedded-token false-positive, and a malformed `_a`.
- **`#6` (cooldown latch):** the effect cleanup now re-arms `cooldownRef = false` when clearing the pending timer, so a re-run within the cooldown window can no longer latch it permanently. Added a regression test.
- **`#7` (recent-items prefix collision):** both filters (explore + dashboard) now match the app path on a boundary (followed by end, `#`, `/`, or `?`) so a superstring sibling app id can't surface under the wrong flavor.
- **`#9` (`metricsHashPath` ignores its arg):** now interpolates `metricsPageMode:${mode}` instead of the hardcoded literal.
- **`#11` (non-null assertions):** dropped both `item.children!` / `childItems!` in favor of guarded `?? []`.

**Responses (no code change)**

- **`#1` (touch users can't reach non-first children in expanded nav):** the icon side nav targeted here has no multi-child parent links in the expanded state, so every nav item is directly reachable; the cascading popover is an enhancement on top, not the only path. Happy to add a tap-to-disclose affordance if/when a multi-child parent is introduced.
- **`#4` (renamed `data-test-subj`s):** intentional split (`collapsibleNavCollapseButton` / `collapsibleNavExpandButton`) since the two states are distinct controls. The functional tests live in `opensearch-project/opensearch-dashboards-functional-test`; flagging here so a companion update can land alongside.
- **`#5` (collapsed parent navigates, category icon inert):** this is by design. A parent link is a real registered app (navigating to its first child on click is reasonable), whereas a category is a visual grouper with no destination, so a no-op click is correct. The popover is the discovery path for both.
- **`#8` (`services && onClick`):** `services` is always provided by the chrome layer in practice; the guard is defensive. Can tighten to not render the buttons without `services` if you'd prefer.
- **`#12` (duplicated flatten / recent-items logic):** fair. The marker logic was already extracted into the shared `useOpenOnUrlMarker` hook; consolidating the `LinkItem -> NavPopoverChildItem` flatteners and the recent-items effect is a reasonable follow-up that I would rather do as a separate cleanup to keep this PR scoped.

Let me know if any of the "no code change" items above should be addressed in this PR instead.

joshuali925:
### Code review

Found 1 issue:

1. The `shift+/` keyboard shortcut deactivates when the icon side nav is expanded, contradicting the comment that it "stays active everywhere". This PR drops the `registerLeftBottom` registration of `KeyboardShortcutIcon` (which rendered in the expanded footer via `navControlsLeftBottom$`) and keeps only `registerIconSideNavFooter` → `KeyboardShortcutListener`. That footer slot (`navControlsIconSideNavFooter$`) is rendered only in the **collapsed** branch of the `isNavOpen` ternary in [`collapsible_nav_group_enabled.tsx#L395-L418`](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/48c05f5a6f9b7e93f579b1a6df5a48a87145eee3/src/core/public/chrome/ui/header/collapsible_nav_group_enabled.tsx#L395-L418). When the nav is expanded, the listener unmounts and `useKeyboardShortcut` unregisters `shift+/`. The new Help-menu `KeyboardShortcutHelpModal` does not compensate, since `EuiPopover` only mounts its children while open. (bug due to footer-only mount of the trigger-less listener)

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/48c05f5a6f9b7e93f579b1a6df5a48a87145eee3/src/plugins/management/public/plugin.ts#L344-L355

TackAdam (approved):

joshuali925 (approved):

PR #12296, opened 2026-06-29
feat(osd-mfe): add Module Federation micro-frontend support

zengyan-amazon:
## Description

This PR adds **Module Federation–based micro-frontend (MFE) support to OpenSearch Dashboards**, enabling plugins to be built, signed, deployed to a CDN, and loaded at runtime by a thin OSD host — independently of the OSD release cycle. Driven by the new `--mfe` server flag; without the flag, OSD remains **byte-for-byte identical** to upstream 3.5.

The work spans 16 design phases (1-14 + a Phase-15 spike + Phase-16 follow-on), all summarized in `docs/01-MFE-DESIGN.md` through `docs/19-PHASE16-RESULTS.md` in this branch.

### Motivation

Today, every OSD plugin is bundled into the OSD release: deploying or updating a single plugin requires rebuilding and shipping all of OSD. This PR introduces an opt-in alternative where each plugin is a Module Federation remote loaded at runtime from a registry-pointed URL. Operators get:

- **Independent plugin deployment** — ship one plugin without re-deploying OSD.
- **Per-customer / canary rollouts** — registry resolution picks plugin versions per `{customerId, userBucket}`.
- **Production-safe failure isolation** — one failed remote does NOT block core boot or other plugins (Phase 4 invariant).
- **Supply-chain hardening** — every remote is SRI-pinned (Phase 12), registries are HMAC-signed.
- **Visible degradation** — failed plugins surface a graceful "unavailable" UX (Phase 14).
- **CDN-served core + themes + shared-deps** — the OSD server becomes an ultra-thin shim (Phase 16).

### Architecture in brief

1. **Server-side `--mfe` flag** gates everything. When off, OSD HTML, CSS, JS, and CSP are byte-for-byte unchanged.
2. **Browser bootstrap** (`packages/osd-mfe/src/bootstrap/`) seeds the Module Federation share scope from `__osdSharedDeps__`, fetches a per-customer manifest injected by the server, loads each remote with SRI verification, registers it into the existing `window.__osdBundles__` shim so `plugin_reader` resolves synchronously **unchanged**, then invokes core boot.
3. **Registry-driven** — a JSON registry (v1/v2/v3 schemas, auto-migrated) names every plugin's remoteEntry URL + integrity + version. Authoring is an offline data edit (`scripts/update_registry`); the registry is signed (Phase 12) and validated server-side (Phase 13). The browser never fetches the registry directly — the OSD server injects the resolved manifest.
4. **Phase 16** extends registry-managed delivery to core + orchestrator + themes + shared-deps CSS, completing the server↔CDN separation.

### Locked invariants (verified by CI gates in this PR)

| Invariant | Verifier (in this PR) |
|---|---|
| No-flag path byte-for-byte unchanged | `harness/verify_noflag_diff.js` reports EMPTY diff |
| `packages/osd-optimizer` + `src/core/public/plugins/plugin_reader.ts` UNTOUCHED | `git diff <base>..HEAD` on those two paths = 0 lines |
| One failed remote does NOT block core boot (Phase 4) | `verify_phase14` cases A-F |
| SRI fail-closed on every fetched remote (Phase 12) | `verify_sri` 6/6 |
| Lazy chunks per app (Phase 8) | `measure_lazy.js` LAZY verdict |
| Browser makes ZERO `/registry` HTTP fetches in `--mfe` mode (Phase 13) | `verify_phase13` case G |
| Per-customer manifest resolution (rollouts, tenant overrides) | `verify_phase13` cases A-F |
| Compat env-keyed policy (`block` in dev / `skip` in prod) | `verify_compat`, Phase 9 unit tests |

### Scope

**In scope** (this PR):
- 16 design phases of MFE infrastructure + the merged Phase-16 CDN-separation work.
- New `packages/osd-mfe` package (~14k LOC including tests).
- Build, deploy, registry-CLI scripts (`scripts/build_mfe.js`, `scripts/deploy_mfe.js`, `scripts/update_registry.js`).
- 623 jest tests pass; ~12 harness verifiers (workspace-only, not in the OSD repo per project convention).
- The MFE flow has been deployed end-to-end against a real CloudFront distribution and verified in real Chromium.

**NOT in scope** (recorded as follow-up work, NOT in this PR):
- Production registry SERVICE (HTTP-backed registry as a deployable service). This PR ships a FileRegistryReader reference implementation + a swappable `RegistryReader` interface for a future service.
- Asymmetric registry signing + KMS integration (Phase 12 currently ships HMAC; the verification API supports asymmetric via a future signer impl).
- Out-of-tree plugin SDK + the build pipeline for plugins outside the OSD monorepo. A Phase-15 spike (`docs/14-EXTERNAL-PLUGIN-SPIKE.md`) validated that real external plugins (e.g., `index-management-dashboards@3.5`) DO load as MFE remotes with no plugin-source patches, but ergonomic tooling for out-of-tree authoring is future work.
- Several `update_registry --merge` / `--plugin` flows are not yet v3-schema-aware (the v3-specific `--update-core / --update-orchestrator / --update-theme / --update-shared-deps-css` flags ARE v3-aware).

## Issues Resolved

This PR is a foundational architecture contribution; no specific issue tracker entries to link. The full design + per-phase rationale lives in this branch under `docs/01-MFE-DESIGN.md` … `docs/19-PHASE16-RESULTS.md`.

## Screenshot

The visible user-facing additions (off by default in production):

- **MFE Inspector panel** (dev-only, opt-in via `?inspect=true` URL query + `mfe.allowOverride=true` config): a small fixed-position panel listing every micro-frontend, its resolved source (registry/CDN vs override), and an editable input to repoint a single plugin at a local dev server. Persisted in `localStorage` + URL query. Source: `packages/osd-mfe/src/bootstrap/inspector.tsx`.
- **Disabled-plugin degraded app stub** (only renders for the affected plugin's `/app/<id>`): a pure-DOM status component with the plugin id, an `errorClass` badge, and a human-readable reason. Replaces OSD's default 404 for plugins that failed to load. Source: `packages/osd-mfe/src/bootstrap/disabled_plugin.ts`.
- **Inspector "Disabled plugins" section**: when the inspector is visible AND any plugin failed to load, an additional section lists the failures with their `errorClass` + human-reason. Source: `inspector.tsx::DisabledPluginsSection`.

Screenshots can be added if requested.

### Automated

- **Unit tests**: 623 jest tests passing across 33 suites in `packages/osd-mfe`. Coverage includes:
  - Bootstrap sequence (`bootstrap_mfe.test.ts`, 53 tests)
  - Registry schemas v1, v2, v3 + migrators (`schema*.test.ts`)
  - Compat classifier + Phase 9 policy (`compat_classifier.test.ts`)
  - SRI fail-closed + chunk error surface (Phase 12)
  - Server-side manifest resolution (`resolve_v2.test.ts`)
  - Telemetry dispatcher + visible degradation UX (Phase 14)
  - v3 asset build + apply (Phase 16)
- **Type-check**: `yarn tsc -p packages/osd-mfe --noEmit` clean.
- **The 12 harness verifiers** (in the source branch's workspace, not committed to this PR per convention — these are dev tools, not shipped tests):
  - `verify_baseline.js` 8/8
  - `verify_noflag_diff.js` EMPTY (the byte-for-byte guardrail)
  - `verify_csp.js` ALL PASS
  - `verify_sri.js` 6/6
  - `verify_mfe_coverage.js` 12/12 pages + 58/58 remotes
  - `verify_phase13.js` 7/7
  - `verify_phase14.js` 6/6
  - `verify_phase16.js` 8/8
  - `measure_lazy.js` LAZY
- **End-to-end browser**: the entire 16-phase sequence has been deployed to a real CloudFront distribution and verified in real Chromium against a v3 registry. The user-facing UX (login, navigation, dashboard, discover, console) works identically to non-MFE mode.

### Manual

To boot OSD in `--mfe` mode locally:

```bash
yarn osd bootstrap
yarn build:plugins   # populate the local origin's plugin remotes
node scripts/build_mfe.js --all --dist   # build every plugin as a Module Federation remote
# Start a local origin (e.g., the workspace harness on :8080)
yarn start --mfe --no-base-path -c <mfe.yml>   # OSD launches in --mfe mode on :5602
```

Visit `http://localhost:5602/app/home`. Open DevTools Network tab → every plugin loads its `remoteEntry.js` from the configured origin.

## Suggested review strategy

This PR is **72 commits / ~32k lines additive** and spans 16 design phases. Given its size, GitHub's Code-Diff-Analyzer already flagged it for manual review. We **strongly recommend reviewing this PR by logical phase boundary**, not as a single mega-diff.

We have a per-phase merge plan that proposes splitting this work into **6 sequential follow-up CRs** (Foundation / Resilience / Build-Deploy-Lazy / Compat-SRI-Signing / Server-Resolved-Registry / Observability-Plus-CDN-Separation), each ~3-8 KLOC, each with its own acceptance verifier. The plan is in `docs/20-NEO-MERGE-PLAN.md` in this branch (originally written for an Amazon-internal fork merge but the strategy applies equally well to upstream landing). Maintainers should feel free to **request that this PR be closed and re-opened as the 6-PR series** if that fits the OSD review process better.

Alternatively, the per-phase results docs (`docs/0N-PHASE*-RESULTS.md`) document each phase's design + acceptance independently and can be reviewed phase-by-phase within this single PR.

## Known caveats + follow-ups

1. **`packages/osd-apm-topology`** is referenced in some Amazon-internal forks but is NOT in this PR. The MFE work has no dependency on it.
2. **OSD `data` plugin's `QueryStringManager` constructor** has a pre-existing bug (not introduced by this PR): it tries to subscribe to `application.currentAppId$` during plugin setup phase, but that observable is only on `ApplicationStart`. Surfaces as a "Could not subscribe to application.currentAppId$" warning in both legacy and `--mfe` modes. Worth a separate upstream OSD fix (defer the subscription to start phase).
3. **`update_registry.js --from-manifest --merge`** is currently v1-validator-only when reading the input registry. v3 registries can be patched via the v3-specific `--update-*` flags (which ARE v3-aware) or via direct jq edit until the legacy CLI flows are updated. Not a runtime bug; only affects the operator authoring workflow.
4. **`harness/*.js` test/verifier files** live in the source branch's workspace, NOT in the OSD repo (consistent with project convention for dev tooling). If maintainers want them committed, they could be moved into `test/` or `cypress/` directories — happy to do that on request.
5. **The first 6 commits in this PR** are unrelated to the MFE work — they add cypress coverage to discover/dashboard/dev_tools, pre-dating the start of MFE work. Could be split into a separate PR if reviewers prefer.

## Check List

- [x] All tests pass
  - [x] `yarn test:jest` — 623/623 passing in `packages/osd-mfe`
  - [ ] `yarn test:jest_integration` — pending CI run on this branch
- [x] New functionality has been documented (`docs/01-MFE-DESIGN.md` through `docs/19-PHASE16-RESULTS.md`).
- [x] Commits are signed per the DCO using `--signoff`.

---

**Recommendation to maintainers**: this PR exists primarily as a **complete reference of the MFE work** for upstream review. Maintainers should feel empowered to close this PR and request a re-submission as the 6-PR series described in `docs/20-NEO-MERGE-PLAN.md` if that workflow better fits the OSD review process. The author is happy to do that restructuring on request.


src/core/server/utils/mfe_boot_manifest.ts
```diff
@@ -0,0 +1,656 @@
...
+
+/**
+ * Read the v2 (or v1, auto-migrated) registry document at `path` and resolve
+ * it against `dimensions`. Throws a descriptive Error on file/JSON errors, on
+ * an unsupported schemaVersion, or on a structurally malformed v2 doc.
+ *
+ * Cached by mtime in-process; dimensions vary per call so the pure resolution
+ * always runs even on a cache hit.
+ */
+export function readMfeBootManifest(
```
zengyan-amazon, line 529: **Merge-blocker: the registry signature is never verified on the live boot-manifest path.** `readMfeBootManifest` (and the `FileRegistryReader` reference impl) read → migrate → resolve the registry with no call to `verifyRegistrySignature`. Only the legacy `FileRegistryProvider` verifies, and the browser explicitly *skips* verification whenever a boot manifest is injected (`bootstrap_mfe.ts:702`: `if (!bootManifest && registryVerification && registryVerification.key)`) — which is exactly the Phase 13/16 path production uses (`ui_render_mixin.js` reads the registry server-side and injects `mfeBootManifest`). Net: the entire HMAC signing scheme (`signing.ts` / `verify_registry_web.ts`) is dead code on the shipped path.

**Impact:** anyone who can write the file the server reads from `mfe.registryPath` (a CDN-synced path, shared volume/mount, or a future HTTP-backed reader) can repoint every plugin's `remoteEntry` — and the v3 core/orchestrator — at attacker-controlled code, accepted with no signature check. Combined with the SRI gap (see `load_remote.ts`), that's arbitrary code execution in the OSD origin.

**Fix:** verify the signature server-side in `readMfeBootManifest` before resolving (the server is allowed to depend on the pure `signing_common` canonicalizer), and fail closed when `mfe.registryPath` points at a registry whose signature is absent/invalid.

src/core/server/utils/mfe_boot_manifest.ts
```diff
@@ -0,0 +1,656 @@
...
+  cache.set(path, { mtimeMs: stat.mtimeMs, doc });
+  return doc;
+}
+
+function coerceToV2DocLite(parsed: unknown): V2DocLite {
+  if (!isPlainObject(parsed)) {
+    throw new Error('MFE registry document must be a JSON object');
+  }
+  const sv = parsed.schemaVersion;
+  if (sv === 3) {
```
zengyan-amazon: **Drop v1 & v2 entirely — ship v3-only for the first release.** Since this is the first release there are no v1/v2 registries in the wild to stay backward-compatible with, so this `schemaVersion` auto-detect + `migrateV1ToV2Lite` migration is unnecessary complexity. That versioning is also duplicated across `schema.ts` / `schema_v2.ts` / `schema_v3.ts` plus parallel `resolve.ts` / `resolve_v2.ts` and the package-side reader.

Dropping v1/v2 would: (a) delete both migration paths and one of the two near-identical entry validators (`schema_v2.validateMfeEntry` is a verbatim copy of `schema.ts`'s); (b) remove the deliberately-laxer v3 projection in this file that drops the `sha384-` integrity check; and (c) shrink the registry attack surface and the drift risk between this hand-written server copy and `@osd/mfe`'s resolver.

**Recommendation:** keep only the v3 schema + a single resolver, and make `schemaVersion !== 3` a hard error here.

packages/osd-mfe/src/bootstrap/load_remote.ts
```diff
@@ -0,0 +1,149 @@
...
+ * @param url absolute URL of the script to inject
+ * @param integrity optional SRI hash (`sha384-…`); enables `crossorigin` + integrity
+ */
+export function loadScript(url: string, integrity?: string): Promise<void> {
+  return new Promise<void>((resolve, reject) => {
+    const el = document.createElement('script');
+    el.src = url;
+    el.type = 'text/javascript';
+    el.async = false;
+    if (integrity) {
```
zengyan-amazon, line 51: **Merge-blocker: SRI is fail-open.** `loadScript` only sets `integrity`/`crossorigin` inside this `if (integrity)` branch; when `integrity` is absent/empty the script loads with no hash check and `onload` resolves normally. `integrity` is *optional* in every schema (`schema.ts`, `schema_v2.ts`, `boot_manifest.ts`, `schema_v3` `V3AssetDescriptor`), and nothing requires it to be present before loading.

**Impact:** an attacker tampering with the registry keeps `remoteEntry` on the allow-listed CDN origin but simply deletes the `integrity` field — the entry still validates and the browser executes unverified remote JS. The SRI guarantee evaporates by omitting the one field the attacker controls.

**Fix:** on the prod path, make missing integrity fail closed (reject the entry rather than load without a hash). The `schema_v3` header already says integrity "MUST be set" for CDN URLs but notes it's enforced "by deployment, not by the schema" — enforce it in code.

src/legacy/ui/ui_render/bootstrap/startup.js.hbs
```diff
@@ -118,3 +118,65 @@ themedFontDom.innerHTML = `
...
+//
+// SAFETY: this block is INERT when the META is absent — `metaTag` is null,
+// the parse short-circuits, and nothing is appended to `<head>`. The
+// no-flag served HTML never contains this META (it's gated server-side by
+// `injectedMetadata.mfe?.themes`), so the no-flag path is byte-for-byte
+// unchanged at the served-bytes level; this added JS runs in both paths but
+// only TAKES EFFECT in the --mfe path with v3 themes — verify_noflag_diff
+// compares HTML, not /startup.js bytes (which already vary per-user via
+// `theme:darkMode` user setting).
+(function loadMfeTheme() {
```
zengyan-amazon, line 139: **The "off-flag output is byte-for-byte identical to upstream 3.5" invariant doesn't hold for `/startup.js`.** This `loadMfeTheme()` IIFE is ungated — only a runtime `if (!metaTag) return;`, not a handlebars `{{#if}}` — and `startup.js.hbs` is the always-served `/startup.js` template (`ui_render_mixin.js` builds `AppBootstrap(…, 'startup')` with no mfe branch). So the served `/startup.js` bytes differ from upstream for 100% of users even with `--mfe` off. `verify_noflag_diff.js` only diffs the `/app/home` HTML, never `/startup.js` (the comment just above concedes this).

**Fix:** either gate this block out at compile time when mfe is off (so the bytes truly match), or soften the headline claim to "HTML/CSP byte-for-byte; `/startup.js` gains an inert MFE theme loader" and extend `verify_noflag_diff` to cover `/startup.js`.

src/optimize/optimize_mixin.ts
```diff
@@ -33,18 +33,182 @@ import Hapi from '@hapi/hapi';
...
+        // the underlying parse error.
+        return false;
+      }
+    };
+  })();
+
+  if (mfeThemeRefuser) {
+    const LEGACY_THEME_PATH_RE = /^\/ui\/legacy_([^/]+)_theme\.css$/;
+    server.ext('onRequest', (request, h) => {
+      const match = LEGACY_THEME_PATH_RE.exec(request.path);
```
zengyan-amazon, line 207: **This refuser silently never fires under a non-root `server.basePath`.** With the default `rewriteBasePath: false`, no basePath strip is installed (`http_server.ts` `setupBasePathRewrite` returns early), so `request.path` keeps the prefix (e.g. `/dashboards/ui/legacy_dark_theme.css`) and the `^/ui/` anchor in `LEGACY_THEME_PATH_RE` never matches. The guarantee documented just above — that the legacy same-origin theme route is 404'd so a stale-cache copy can't downgrade the SRI gate — doesn't hold behind any basePath.

**Fix:** strip the configured basePath before matching (or match the basePath-relative path), the way the router does. The `core`/`sharedDepsCss` refusers are unaffected because they rely on Hapi route matching rather than raw `request.path`.

packages/osd-mfe/src/bootstrap/bootstrap_mfe.ts
```diff
@@ -0,0 +1,1087 @@
...
+      console.error(
+        `[mfe] Failed to load OSD core (core.entry.js) from ${options.core.url}; refusing to ` +
+          `invoke core boot. App cannot proceed on a tampered or unreachable core ` +
+          `(errorClass=${errorClass}).`,
+        err
+      );
+      // Re-throw to fail-close: bootstrapMfe's Promise rejects, the chunk-error
+      // surface catches the unhandled rejection, and the rest of the load
+      // sequence (shared-deps / plugins / invokeCoreBootstrap) is never run.
+      throw err;
```
zengyan-amazon, line 642: **Core load failure *without* an integrity hash hangs silently — contradicting the fail-closed comment above.** When `options.core` has no integrity (the legitimate same-origin dev fallback — see the `errorClass` ternary at line 620), a failed load rejects with `"Failed to load script: <url>"`. That's re-thrown here, the `.hbs` caller `window.__osdBootstrapMfe__({…})` attaches no `.catch`, and `CHUNK_FAILURE_PATTERNS` in `chunk_error_surface.ts` doesn't match that string — so `onRejection` ignores it. `invokeCoreBootstrap` never runs and the page hangs on the loading screen with only a `console.error`. The fail-closed comment holds only when core carries integrity.

**Fix:** attach a `.catch` at the `__osdBootstrapMfe__` call site (or here in `bootstrapMfe`) that renders a visible error banner for *any* core-boot rejection, not just integrity-pattern matches.

packages/osd-mfe/src/bootstrap/chunk_error_surface.ts
```diff
@@ -0,0 +1,295 @@
...
+    // capture phase with the failing element as the target — they do NOT bubble,
+    // and they are not `ErrorEvent`s. Only treat a <script> whose URL looks like a
+    // remote chunk as a chunk failure, so unrelated errors are ignored.
+    const el = event.target as HTMLScriptElement | null;
+    if (
+      el &&
+      typeof el.tagName === 'string' &&
+      el.tagName.toLowerCase() === 'script' &&
+      typeof el.src === 'string' &&
+      (/\.chunk\./.test(el.src) || /\/mfe\//.test(el.src))
```
zengyan-amazon, line 276: **This fires a false "compromised CDN / MITM" alarm for ordinary network blips.** The handler flags any `<script>` whose `src` matches `/\.chunk\./` and renders an unconditional integrity-failure banner ("failed its integrity check … a possible compromised CDN / MITM", lines 142/149) — but it never inspects `el.integrity` and can't distinguish a real SRI mismatch from a generic fetch failure (the browser fires the same `error` event for both). Core lazy chunks are named `core.chunk.<id>.js` by osd-optimizer and carry no SRI, so a transient network hiccup on a same-origin core chunk shows users a security-tampering banner.

**Fix:** only use the integrity/MITM wording when the failing element actually had an `integrity` attribute; otherwise use neutral "failed to load" wording.

zengyan-amazon (commented):
Automated code review focused on where the **shipped path diverges from the PR's stated guarantees**. Overall the architecture is solid and the crypto primitives themselves are correct (constant-time HMAC, canonicalized/key-sorted payload, prod-gated dev overrides, wildcard-rejecting origin parsing). But two of the marquee security controls — registry **signing** and **SRI** — are not actually enforced on the boot-manifest path that production uses, and the "byte-for-byte off-flag" claim doesn't hold for `/startup.js`. I'd treat the first four inline comments as merge-blockers.

**One finding that couldn't be anchored inline** (the line is outside this PR's diff hunks): `src/legacy/ui/ui_render/utils.js` `applyCspModifications` lowercases the rule text (L17) and every modification value (L33). CSP `'nonce-…'` / `'sha256-…'` source expressions are base64 and **case-sensitive**, so an operator-supplied `csp-modifications` entry containing a nonce/hash gets silently corrupted — legit scripts blocked, or an `'unsafe-inline'` workaround that downgrades CSP. OSD's own shipped CSP emits no nonce/hash sources, so this only bites operator-introduced ones (lower severity), but the function should preserve case for the token classes (`'nonce-`, `'sha256-`, `'sha384-`, `'sha512-`).

Cleanup notes (non-blocking): `readMfeBootManifest` re-runs the full `resolveOnce` on every call (up to ~5×/render) even on an mtime cache hit — memoize the resolved manifest by `(mtime, customerId, userBucket)`; and the three refuser IIFEs in `optimize_mixin.ts` are copy-paste that should collapse to one parameterized helper so the fail-soft policy is single-sourced.

packages/osd-mfe/dev/verify_baseline.js
```diff
@@ -0,0 +1,65 @@
+/*
```
zengyan-amazon, line 1: Minor: missing the Apache-2.0 SPDX header that its 6 sibling files in this directory carry (added in this same PR). `scripts/ci/mfe_harness` is in `.eslintignore`, so the header lint won't auto-catch this.

PR #12298, opened 2026-06-29
PPL Linter Field Validation + Command Typo Fix

### LINKED ISSUE #5558: [Bug] PPL syntax highlighting does not flag invalid source fields in grok, parse, patterns, spath
### What is the bug?

The Explore editor shows no error highlighting when the user writes an invalid `<name>=<name>` form as the source field of `grok`, `parse`, `patterns`, or `spath` (e.g. `grok field=body "..."`, `grok foo=bar "..."`). The grammar bundle served by `/_plugins/_ppl/_grammar` accepts these as valid because the source-field slot uses the broad `expression` rule, and `<name>=<name>` matches the `compareExpr` alternative.

The documented form is positional: `grok <field> <pattern>`.

### How can one reproduce the bug?

1. Open Explore with PPL on a >= 3.6 cluster (which uses the runtime grammar bundle).

2. Type any of these — all are accepted with no error highlighting:

        source=myindex | grok field=body '%{COMMONAPACHELOG}'
        source=myindex | grok foo=bar '%{WORD}'
        source=myindex | parse max_match=raw '(?<x>\w+)'
        source=myindex | spath input=a=b

3. Expected: red squiggle showing the source field is invalid (`field=body`, `foo=bar`, etc.).

4. Actual: no highlighting — the grammar accepts it as a comparison expression.

### What is the expected behavior?

The editor should flag `<name>=<name>` source fields as syntax errors. The valid positional forms like `grok body '%{COMMONAPACHELOG}'` and `grok log.message '%{WORD}'` should continue to show no error.

### What is your host/environment?

Any cluster >= 3.6 with the Explore editor using the runtime grammar bundle.

### Do you have any additional context?

The root cause is that `grokCommand`/`parseCommand`/`patternsCommand` use `expression` for `source_field` and `spathParameter` uses `expression` for `input=`/`output=`. The `compareExpr` alternative of `expression` matches any `<name>=<name>`.

Hanyu-W:
### Description

Adds field-aware validation and command-typo quick-fixes to the PPL lint engine from #12274. Both reuse the existing catalog-driven framework — no new engine or server routes — and stay behind the `queryEnhancements.ppl.lint.enabled` flag.

**1. Unknown-field rule (error, on by default)**

Flags field references not found in the dataset's index fields or fields created upstream in the pipeline (`eval`/`rename`/`stats` outputs, join aliases). When a close match exists, it offers a one-click fix:

> Unknown field "respose". Did you mean "response"? → *Replace with "response"*

Covers field references in `where`, `eval` RHS, `sort`, and comparable `fieldExpression` positions. Self-suppresses when no field context is available (no dataset selected). Also catches Splunk-style `grok field=x` / `parse field=x` / `patterns field=x` syntax (which PPL silently misparses) with a *Remove "field="* fix.

**2. Command-typo quick-fix**

Flags command typos, and turns ANTLR's raw `mismatched input 'wherre' expecting {…40 keywords…}` into:

> Unknown command "wherre". Did you mean "where"? → *Replace with "where"*

This provides a clearer and more actionable error message. The candidate set is derived from the grammar's `commands` rule via the ATN, so grammar renames propagate automatically. Gives a quick-fix to the nearest match using Damerau–Levenshtein.

**Implementation notes:**

- `code_action_provider` + `fix_registry` surface quick-fixes for both `ppl-lint` and `ppl-syntax` markers. Fixes are stored in a side table keyed by marker fields Monaco preserves (Monaco rebuilds markers and drops custom properties).
- Shared Damerau-Levenshtein distance + nearest-candidate sweep for both field and command suggestions.
- `GeneralErrorListener` extended with optional `code`/`fix` fields — backward-compatible with existing SQL/DQL/autocomplete callers.
- Command Suggestion will always be error severity, as that only fires with a confirmed syntax issue
- Field Validation will only run with the runtime grammar bundle enabled.

### Issues Resolved

Two more rules for https://github.com/opensearch-project/sql/issues/5405
Resolves https://github.com/opensearch-project/sql/issues/5558

## Screenshot

image
Command-Validation

image
Field-Validation

image
Field-Validation

image
UI settings

## Testing

1. Enable in `opensearch_dashboards.yml` and restart:
   ```yaml
   queryEnhancements.ppl.lint.enabled: true
   ```

2. **Unknown field** — select a dataset in Explore and enter:
   ```
   source=accounts | where respose > 1
   ```
   Expect a warning on `respose` with a "Did you mean response?" fix. Confirm real fields produce no warning.

3. **Eval RHS / sort** — `source=accounts | eval x = nonexistent + 1` flags `nonexistent` (not `x`); `source=accounts | sort agee` flags `agee`.

4. **Field-slot shape** — `source=accounts | grok field=email "%{...}"` flags `field=` with a remove fix.

5. **Command typo** — `source=accounts | wherre status = 1` shows "Did you mean where?" with a replace fix. A 1-char stray token (`| a`) keeps the raw ANTLR message.

6. **Unit tests:**
   ```bash
   yarn test:jest packages/osd-monaco/src/ppl --no-coverage
   yarn test:jest src/plugins/data/public/antlr/opensearch_ppl --no-coverage
   yarn test:jest src/plugins/data/public/ppl_lint --no-coverage
   yarn test:jest src/plugins/query_enhancements/server --no-coverage
   ```

### Check List

- [x] All tests pass (`yarn test:jest` lint-related suites)
- [ ] New functionality has been documented


packages/osd-monaco/src/ppl/language.ts
```diff
@@ -150,7 +161,34 @@ const processSyntaxHighlighting = async (model: monaco.editor.IModel) => {
...
+      // a fixed field list and drops custom properties — same constraint the
+      // lint path handles via setModelFixes.
+      const syntaxFixes = new Map<string, MarkerFix>();
+
+      // Command-typo suggestion is a UX layer on the syntax channel, toggleable
+      // via the same PPL-lint rules uiSetting (id `command-suggestion`). When it
+      // is disabled we revert the friendly rewrite: use ANTLR's raw message and
+      // drop the quick-fix, leaving the plain syntax error. Defaults to enabled
+      // when the flag is unset (no context / older config).
+      const commandSuggestionEnabled = getPPLLintContext(model)?.commandSuggestionEnabled !== false;
```
mengweieric: The command suggestion UX isn't behind the global PPL lint flag. `processSyntaxHighlighting` gates it only on `commandSuggestionEnabled !== false`, and because of the `!== false` default it fires even when there's no lint context pushed at all (the disabled state). The quick fix in `code_action_provider.ts` serves off the same syntax markers with no global check either, so the friendly "Unknown command ... Did you mean ...?" rewrite plus its lightbulb show up when the capability is meant to be off. The PR description says everything stays behind `queryEnhancements.ppl.lint.enabled`, so this path is a gap. Could you also check `isPPLLintEnabled()` here (or push the enabled bit down so the error listeners don't build the suggestion when lint is off), and add a test for suppression when the global capability is off? There's a per context toggle test but none for the global gate.

src/plugins/data/public/ppl_lint/lint_context_builder.ts
```diff
@@ -10,27 +10,62 @@ import {
...
 export function buildPPLLintContext(
   dataset: LintContextDataset | undefined,
+  lintFields: LintFieldsCache,
   services: { uiSettings: IUiSettingsClient; http: HttpSetup }
 ): PPLLintContext {
   const dsId = dataset?.dataSource?.id;
   const dsVersion = dataset?.dataSource?.version;
+  // Only apply cached fields when they belong to the active dataset; otherwise
+  // leave them undefined so field-validation self-suppresses (no false unknowns).
+  const cacheMatchesDataset = lintFields.datasetId === dataset?.id;
```
mengweieric: The field cache keys on dataset id only. `LintFieldsCache` carries just `datasetId`, `cacheMatchesDataset` is `lintFields.datasetId === dataset?.id`, and both editor reload effects stamp only `{ datasetId, fields }` and depend on `dataset?.id`. But the context also carries `dataSourceId`/`dataSourceVersion`. If someone switches data source while the dataset id stays the same, the old field set gets reused, so you get false unknown field flags or missed ones. Suggest adding `dataSourceId` to the cache stamp and the reload deps, and only applying cached fields when both match. A regression test for same dataset id with a different data source and field list would cover it.

mengweieric (commented):
Reviewed this alongside #12319 and #12339. Skipping the bot's low level notes. Two things worth a look before this lands.

mengweieric:
Please also address CI failures.

src/plugins/data/public/ppl_lint/lint_context_builder.ts
Hanyu-W: Addressed in `84beb49a20`.
`LintFieldsCache` now carries both `datasetId` and `dataSourceId`, and `buildPPLLintContext` applies cached fields only when both values match the active dataset:

```ts
 lintFields.datasetId === dataset?.id &&
  lintFields.dataSourceId === dataset?.dataSource?.id
```

Both field-loading paths now capture and stamp `dataSourceId`: the data plugin query editor and the Explore query panel editor. Their effect dependencies also include `dataset?.dataSource?.id`, so changing the data source under an unchanged dataset ID triggers a reload instead of retaining the previous source's fields.

I added regression coverage in `lint_context_builder.test.ts` for the exact failure case: the dataset IDs match but the data source IDs differ. The test verifies that cached fields are discarded so field validation self-suppresses until the correct metadata is loaded. There is also coverage for local datasets where both data source IDs are undefined.

packages/osd-monaco/src/ppl/language.ts
Hanyu-W: Addressed in `84beb49a20`.

`processSyntaxHighlighting` now requires both conditions before enabling the command-suggestion enhancement:
 ```ts
 isPPLLintEnabled() &&
 getPPLLintContext(model)?.commandSuggestionEnabled !== false
```

When the global capability is off, an `UNKNOWN_COMMAND` error now uses its raw ANTLR message and drops the associated fix. Because no syntax fix is registered in the model's fix table, `code_action_provider.ts` also has nothing to serve, so the quick-fix lightbulb is suppressed without requiring a second global check in the provider.

I added coverage in `language_command_suggestion_toggle.test.ts` that keeps the per-context toggle enabled while disabling the global capability. The test verifies both the raw ANTLR message and the absence of a registered quick-fix.

Hanyu-W:
Thanks for the PR review @mengweieric! Addressed all inline comments in `84beb49a20`:

- Command suggestions now respect the global PPL lint flag as well as the per-rule toggle. When lint is off, the editor shows the raw ANTLR error and does not register a quick-fix.
- Field metadata is now cached by both `datasetId` and `dataSourceId` in the data and Explore editors. Both IDs participate in cache matching and reload dependencies, preventing stale fields after a data source switch.
- Reran CIs, all of them pass now. CI failure was due to a flaky test, not a code issue.

Both changes include regression coverage for the cases raised in review.

Also included three follow-up correctness fixes:
- `084b28ad9a`: Runs field-slot shape validation on the compiled grammar path used by pre-3.6 clusters.
- `87975e96fb`: Correctly extracts named captures that follow a lookbehind, avoiding false unknown-field diagnostics.
- `87975e96fb`: Rejects unknown severity overrides so they cannot bypass configured severity floors.

Please re-review when you get the chance. Thanks!

mengweieric (dismissed):

joshuali925:
### Code review

Found 2 issues:

1. Backtick-quoted created fields produce false "Unknown field" diagnostics. Field *references* are normalized per dotted segment via `unquoteIdent` (backticks stripped)...

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/87975e96fbad810b4df3728a39e1ecfc356af322/packages/osd-monaco/src/ppl/lint/rules/field_validation.ts#L186-L188

...but created fields are registered with their raw `getText()` in `collectCreatedFields`, keeping the backticks — both for the `... AS <name>` alias slot and the eval-LHS slot (note the pattern/spath OUTPUT slots do call `unquote`, so the asymmetry is specific to these two):

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/87975e96fbad810b4df3728a39e1ecfc356af322/packages/osd-monaco/src/ppl/lint/pipeline_shape.ts#L150-L156

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/87975e96fbad810b4df3728a39e1ecfc356af322/packages/osd-monaco/src/ppl/lint/pipeline_shape.ts#L177-L181

So the lookup `createdFields.has('total')` fails for a stored `` `total` ``. Queries like ``source=t | eval `total` = bytes * 2 | where `total` > 100`` or ``... | stats count() as `cnt` | where `cnt` > 0`` get a spurious "Unknown field" (the rule is on by default). Applying the same per-segment `unquoteIdent` normalization when adding to `createdFields` would fix it.

2. `field-validation` ships at `severity: "error"`, but the file's own docstring and the PR description describe it as a `warning`. This is a Damerau-Levenshtein "did you mean" heuristic that can false-positive (see issue 1), yet it defaults to the harshest severity — while sibling rules default to `info`/`warning`. The docstring:

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/87975e96fbad810b4df3728a39e1ecfc356af322/packages/osd-monaco/src/ppl/lint/rules/field_validation.ts#L464-L466

vs the catalog:

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/87975e96fbad810b4df3728a39e1ecfc356af322/packages/osd-monaco/src/ppl/lint/rules_catalog.json#L23-L26

Behavior is unaffected (PASS 3 suppression is range-based, not severity-based), but the code/docstring/description disagreement suggests the `error` default may be unintended. Worth confirming the intended severity.

joshuali925 (dismissed):
i only ran through claude code, let me know if there's anything you want manual review on

Hanyu-W:
Thanks @joshuali925 for the review! I addressed both findings in 3ef7f40.

 1. Created fields and references now use the same normalizeFieldName helper. This fixes both reported eval/stats cases and also covers rename targets, join aliases, and spath outputs. Regression tests cover quoted and unquoted combinations.
 
 2. Good catch, the severity is supposed to be an error, because the query will not run. I corrected the stale source documentation and PR wording.
 
 Please re-review when you get the chance. Thanks!

joshuali925 (dismissed):

Hanyu-W (commented):
@dcoapp recheck

joshuali925 (approved):

mengweieric (approved):

PR #12300, opened 2026-06-30
feat(chat): hide /investigate suggestion card when tool is unavailable

wanglam:
### Description

Conditionally show/hide the `/investigate` starter suggestion card based on whether the `create_investigation` tool is registered by the dashboards-investigation plugin.

### Changes

- Added `requiredTool?: string` field to `SuggestionItem` interface — a generic mechanism to gate any suggestion card on tool availability.
- The `/investigate` card entry now declares `requiredTool: 'create_investigation'`.
- `ChatMessages` subscribes to `toolDefinitions` from `AssistantActionService` and filters `STARTER_SUGGESTIONS` reactively.
- When the investigation plugin is not installed or disables the tool (e.g. on the search-relevance page), the card is hidden.

### Testing

- 3 unit tests added covering: card hidden by default, shown when tool registered, hidden when tool unregistered.
- All 47 tests in `chat_messages.test.tsx` pass.
- Manual testing with TDD stack (investigation plugin enabled/disabled).

### Issues Resolved

N/A

### Check List

- [x] Commits are signed off (DCO).
- [x] All tests pass locally.

### DIFF
--- a/src/plugins/chat/public/components/chat_messages.tsx
+++ b/src/plugins/chat/public/components/chat_messages.tsx
@@ -79,6 +79,8 @@ interface SuggestionItem {
   text: string;
   prompt?: string;
   action?: () => void;
+  /** When set, this suggestion is only shown if the named tool is registered. */
+  requiredTool?: string;
 }
 
 const STARTER_SUGGESTIONS: SuggestionItem[] = [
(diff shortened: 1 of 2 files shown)


src/plugins/chat/public/components/chat_messages.tsx
SuZhou-Joe, line 93: We may need a registry for the starter suggestions.

SuZhou-Joe (approved):

src/plugins/chat/public/components/chat_messages.tsx
wanglam: Sure. I think the investigation card better to register from the investigation plugin. Let's add a registry and make them has priority and only display only top three in the welcome screen. Want to do some optimizations in the future.

yubonluo (approved):

PR #12319, opened 2026-07-01
feat(ppl-lint): add 6 structural PPL linting rules

Hanyu-W:
### Description

Adds 6 structural PPL lint rules to the lint engine landed in #12274. Each rule is registered in the detector registry, the rules catalog (`rules_catalog.json`), and the `query_enhancements` uiSettings defaults so it can be toggled from Advanced Settings.

**General Rules:**

- `unsupported-window-function-in-eventstats` (error, minVersion 3.4.0) — flags `rank`/`dense_rank`/`ntile`/etc. in `eventstats`/`streamstats` (only `row_number` is supported).
- `disabled-join-type` (warning) — flags `right`/`full`/`cross` join types that are disabled unless `plugins.calcite.all_join_types.allowed` is set; suppressed when the `allJoinTypesAllowed` setting is on. On the compiled surface this is reached through the `type=<keyword>` option form (the SQL-prefix form is only parsed by the runtime grammar).
- `dedup-consecutive-unsupported` (warning, Calcite, minVersion 3.3.0) — flags `dedup … consecutive=true`, which is not natively supported on Calcite and relies on engine fallback.

**Runtime-only rules (no-op on the compiled grammar; fire once the runtime grammar bundle loads):**

- `multisearch-min-subsearch` (error, minVersion 3.4.0) — flags `multisearch` with fewer than two subsearches.
- `union-min-datasets` (error, Calcite, minVersion 3.7.0) — flags `union` with fewer than two datasets.
- `replace-wildcard-asymmetry` (error, Calcite, minVersion 3.4.0) — flags a `replace` whose pattern and replacement have asymmetric, non-zero wildcard counts (escaped `\*` is not counted).

All rules are diagnostic-only (no quick fixes). We have the split between runtime grammar-only rules and general rules because the compiled grammar lacks newer commands like the `unionCommand` and `multisearchCommand`. We query for the runtime bundle using the _grammar endpoint, if it does not exist then we fall back to the compiled grammar.

### Issues Resolved

Provides 6 structural rules for SQL [RFC 5405](https://github.com/opensearch-project/sql/issues/5405)

## Screenshot

image
unsupported-window-function-in-eventstats
image
disabled-join-type
image
dedup-consecutive-unsupported
image
multisearch-min-subsearch
image
replace-wildcard-asymmetry
image
union-min-datasets
image
uiSettings

## Testing the changes

```
# All PPL lint unit tests (rules + engine)
yarn test:jest packages/osd-monaco/src/ppl/lint/

# Typecheck the package
cd packages/osd-monaco && npx tsc --noEmit

# Lint the changed files
node scripts/eslint packages/osd-monaco/src/ppl/lint/rules/
```

Each new rule has a dedicated test file. The compiled-surface rules assert the diagnostic fires; the runtime-only rules assert they stay silent under the compiled analyzer even when given a matching context.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest` (123/123 PPL lint tests pass)


RyanL1997 (approved):

mengweieric (commented):
Reviewed with #12298 since both touch the same catalog and uiSettings surface. Three higher level items.

src/plugins/query_enhancements/server/ui_settings.ts
```diff
@@ -16,6 +16,12 @@ const PPL_LINT_RULE_DEFAULTS: ReadonlyArray<{
 }> = [
   { id: 'head-without-sort', enabled: true, severity: 'info' },
   { id: 'division-by-zero', enabled: true, severity: 'warning' },
+  { id: 'unsupported-window-function-in-eventstats', enabled: true, severity: 'error' },
```
mengweieric, line 18: This and #12298 both branch off main and both insert into `PPL_LINT_RULE_DEFAULTS` right after `division-by-zero` (and into `rules_catalog.json` at the same spot). #12298 adds `field-validation` there. Merging them independently will conflict on both files, and a careless resolution could drop one side's entries. Worth rebasing one on the other, or explicitly folding in #12298's entries here.

packages/osd-monaco/src/ppl/lint/rules/dedup_consecutive_unsupported.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const dedupConsecutiveUnsupportedDetector: Detector = (
+  tree,
+  config,
+  context,
+  ruleNameToIndex
+) => {
+  // Calcite gating is also enforced by the version filter, but guard here too
+  // so a direct detector invocation respects the engine predicate.
+  if (context.isCalcite !== true) {
```
mengweieric, line 36: This rule (and `union_min_datasets` and `replace_wildcard_asymmetry`) requires `context.isCalcite === true`. But `buildPPLLintContext()` derives `isCalcite` only from `dataset?.dataSource?.version`. On the local/default cluster that version comes from the grammar warmup path via `/api/status`, which is stored only to gate the grammar bundle fetch and never flows back into `isCalcite`. So on local, `isCalcite` stays undefined and all three rules quietly return nothing, even where the engine would support them. Suggest propagating the resolved local version into the context (or exposing it from the grammar cache after warmup), plus a test proving a local dataset yields `isCalcite`.

packages/osd-monaco/src/ppl/lint/rules/disabled_join_type.ts
```diff
@@ -0,0 +1,83 @@
...
+        return { keyword: tok, node: joinTypeNode };
+      }
+    }
+  }
+
+  return undefined;
+}
+
+export const disabledJoinTypeDetector: Detector = (tree, config, context, ruleNameToIndex) => {
+  if (context.settings?.allJoinTypesAllowed === true) {
```
mengweieric, line 73: This suppresses on `context.settings?.allJoinTypesAllowed === true` and the unit test passes that in directly, but nothing in product populates `context.settings`. `buildPPLLintContext()` returns no `settings` field at all. There is a server route (`ppl_calcite_settings.ts`) that already resolves `plugins.calcite.all_join_types.allowed` from cluster settings, but it has no client consumer, so the value never reaches the context. Net effect: the rule warns even on clusters where the admin enabled all join types. Suggest wiring that route through to `PPLLintContext` (with a context builder test), or documenting the rule as warning only until it's wired.

Also worth noting: the positive detection path for `union-min-datasets`, `replace-wildcard-asymmetry`, and `multisearch-min-subsearch` has no test. The tests only assert silence on the compiled surface, and per the point above they can't fire on local, so a regression that breaks detection would still pass CI.

src/plugins/query_enhancements/server/ui_settings.ts
Hanyu-W: Good point. I’ll keep the PRs independent for now, since #12298 is still open and folding its rule entries into this PR would couple the scopes. Whichever PR merges second will be rebased and I’ll explicitly verify
both field-validation and these structural rules remain in PPL_LINT_RULE_DEFAULTS and rules_catalog.json.

packages/osd-monaco/src/ppl/lint/rules/dedup_consecutive_unsupported.ts
Hanyu-W: Fixed in 9eb66235. The grammar cache now exposes the resolved version via getResolvedVersion(), buildPPLLintContext() uses that as the effective version when the dataset metadata has no version, and the editor refreshes lint context when the version resolves. I also added a context-builder test proving a local dataset without version metadata gets isCalcite === true once the resolved version is available.

packages/osd-monaco/src/ppl/lint/rules/disabled_join_type.ts
Hanyu-W: Fixed in 9eb66235. I added a client-side calciteSettingsCache backed by the existing /api/enhancements/ppl/calcite_settings route, warm it alongside the PPL grammar warmup, inject settings.allJoinTypesAllowed into PPLLintContext, and refresh/revalidate when settings arrive. I also added positive-path detector tests union-min-datasets, replace-wildcard-asymmetry, multisearch-min-subsearch, plus join-type suppression when allJoinTypesAllowed is true.

Hanyu-W:
Thanks @mengweieric for the review! Just pushed a fix to address your feedback. Highlights:

1. Will rebase the PR that merges second to avoid the potential merge conflict
2. The grammar cache now exposes the resolved datasource version, and `buildPPLLintContext()` uses it when
  dataset version metadata is missing. Added coverage for the local-cluster case so Calcite-gated rules get
  `isCalcite` correctly.
3. Added `calciteSettingsCache` wiring for `/api/enhancements/ppl/calcite_settings`, inject
  `allJoinTypesAllowed` into `PPLLintContext`, and refresh lint context when the settings resolve.
4. Added positive-path tests for `union-min-datasets`, `replace-wildcard-asymmetry`, and `multisearch-min-
  subsearch`, plus coverage that `disabled-join-type` is suppressed when `allJoinTypesAllowed` is true.
  
 Please re-review when you get the chance.

joshuali925:
### Code review

Found 3 issues:

1. All 6 new detectors hardcode their `message` string instead of reading `config.message`, so the catalog `message` field is dead config — and for several rules the two texts diverge, meaning any catalog edit or override silently has no effect. This is the same inconsistency [RyanL1997 flagged and got fixed on #12274](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/12274) (where `head_without_sort` was switched to `message: config.message` to match `division_by_zero`). Example: `dedup-consecutive-unsupported` hardcodes "...on Calcite; the query relies on engine fallback." while the catalog says "...on Calcite and relies on engine fallback."

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/51d80a572389e2381089ef5e95c40a59b4dd2009/packages/osd-monaco/src/ppl/lint/rules/dedup_consecutive_unsupported.ts#L46-L51

vs the catalog `message` that never reaches the editor:

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/51d80a572389e2381089ef5e95c40a59b4dd2009/packages/osd-monaco/src/ppl/lint/rules_catalog.json#L55-L57

Static-message rules (`multisearch-min-subsearch`, `union-min-datasets`, `dedup-consecutive-unsupported`) can use `message: config.message` directly, matching [`head_without_sort.ts`](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/51d80a572389e2381089ef5e95c40a59b4dd2009/packages/osd-monaco/src/ppl/lint/rules/head_without_sort.ts#L73); the interpolated ones should reconcile with the catalog text.

2. `disabled-join-type` uses a descendant search for the `type=` keyword, so nested joins produce duplicate and misattributed diagnostics. The outer loop collects every `joinCommand` via `findAllDescendantsByRule`, and then for each one `detectJoinTypeKeyword` again searches *descendants* for `joinType` — which reaches into the right-hand subsearch's nested `joinCommand` (`joinCommand → tableOrSubqueryClause → subSearch → commands → joinCommand`).

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/51d80a572389e2381089ef5e95c40a59b4dd2009/packages/osd-monaco/src/ppl/lint/rules/disabled_join_type.ts#L48-L54

For `source=a | join type=inner b [ source=c | join type=cross d ] on a.id=b.id`, the outer valid `type=inner` join gets a warning pointing at the nested `cross`, and the nested `type=cross` is flagged a second time (diagnostics are not deduped downstream). Reading the keyword only from the join's own direct `joinOption` children — as the `sqlLikeJoinType` path already does with `findChildByRule` — would fix this.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/51d80a572389e2381089ef5e95c40a59b4dd2009/packages/osd-monaco/src/ppl/lint/rules/disabled_join_type.ts#L66-L71

3. The runtime-only rules assume their target nodes are *direct* children of the command node, but these grammar rules (`unionDataset`, `subSearch`, `replacePair`/`stringLiteral`) do not exist in either committed grammar, so the assumption can't be verified against the actual runtime-bundle grammar and the stub-based tests won't catch a mismatch. If the runtime grammar wraps these in an intermediate list rule (or orders the `replace` literals differently), the counts would always be `< 2` and the rule would fire an error on every valid query.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/51d80a572389e2381089ef5e95c40a59b4dd2009/packages/osd-monaco/src/ppl/lint/rules/union_min_datasets.ts#L24-L27

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/51d80a572389e2381089ef5e95c40a59b4dd2009/packages/osd-monaco/src/ppl/lint/rules/replace_wildcard_asymmetry.ts#L52-L62

Worth confirming the node shapes against the real runtime grammar bundle.

joshuali925 (commented):
i only ran through claude code, let me know if there's anything you want manual review on

Hanyu-W:
Thanks @joshuali925 for the review! I addressed all three findings in 000f6f5.
 
 1. Good catch, all six detectors now use config.message. Dynamic details such as join type, window function, and wildcard counts moved into hoverFacts, keeping the catalog message authoritative.

 2. disabled-join-type now reads only the current join’s direct joinOption children. Tests verify nested disabled
     joins are reported exactly once and do not mask the outer join.
  
 3. I verified the assumptions by parsing representative queries with the real OpenSearch 3.7 runtime grammar
     bundle:
       - multisearchCommand directly contains subSearch
       - unionCommand directly contains unionDataset
       - replacePair directly contains the pattern and replacement stringLiterals in that order
  
   This also exposed a union semantic issue: a mid-pipeline union has the upstream pipeline as its implicit first
   dataset. The rule now applies only to query-initial unions; the real grammar distinguishes these by a pplCommands
   versus commands parent.
   
   Please re-review when you get the chance. Thanks!

joshuali925 (approved):

RyanL1997 (approved):

mengweieric (approved):

PR #12327, opened 2026-07-06
[Workspace] Combine workspace switcher + manage links in the footer menu

ps48:
### Description

Reworks the footer **"Manage workspace"** menu (used by both the icon side navigation and the classic left nav) so it is never empty and always offers workspace navigation.

Previously, opening the footer menu **outside a workspace** showed an empty popover, and there was no workspace picker available outside a workspace at all (the top-of-nav `WorkspaceSelector` is only mounted when inside one, so users had to return to Home to switch).

Changes:
- **Embed the existing `WorkspaceSelector`** at the top of the footer popover, so switching/creating workspaces behaves identically to the nav header (same picker, same recent list, same Manage/Create actions).
- **Add a `flush` variant to `WorkspaceSelector`** that renders its trigger as a flat, full-width menu row — both inside a workspace and outside it ("Select a workspace") — so it visually matches the sibling context-menu rows instead of the boxed nav-header button.
- **Keep the manage-workspace links** (Workspace details, Data sources, Index patterns, Datasets, Assets, Sample data) below the switcher; they render only when present (i.e. inside a workspace).
- **One-time first-visit `EuiTour`** anchored to the footer icon so users learn workspace management moved here. Dismissed permanently via `localStorage` (or automatically once the user opens the menu), mirroring the existing `no_data_popover` pattern.

The default (non-`flush`) `WorkspaceSelector` behavior in the nav header is unchanged.

### Issues Resolved

Part of the icon side navigation RFC: opensearch-project/OpenSearch-Dashboards#11545

Follow-up UX fix for the footer "Manage workspace" menu introduced in #12285.

## Screenshot
* New button tour
Screenshot 2026-07-06 at 12 04 09 AM

* workspace switcher within a workspace 
Screenshot 2026-07-06 at 12 03 29 AM

* workspace switcher in admin pages
Screenshot 2026-07-06 at 12 04 30 AM

## Testing the changes

1. Enable the icon side nav (`opensearchDashboards.enableIconSideNav: true`) and open an Observability workspace.
2. Click the footer **Manage workspace** icon → the popover shows a flat workspace-switcher row on top, then the manage-workspace links. The switcher row style matches the links below it.
3. Navigate **outside any workspace** → open the footer menu → it is not empty; the switcher row shows "Select a workspace" and opens the picker.
4. Classic nav (flag off) → the same footer menu renders correctly.
5. First load with a cleared `localStorage` → the "Workspace management moved" tour points at the footer icon; dismissing it (or opening the menu) prevents it from reappearing.
6. `yarn test:jest src/plugins/workspace/public/components/manage_workspace_menu src/plugins/workspace/public/components/workspace_selector` — 24 tests pass.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest` (affected suites)

### DIFF
--- a/src/plugins/workspace/public/components/manage_workspace_menu/manage_workspace_menu.tsx
+++ b/src/plugins/workspace/public/components/manage_workspace_menu/manage_workspace_menu.tsx
@@ -3,16 +3,21 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import React, { useCallback, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import {
+  EuiButtonEmpty,
   EuiButtonIcon,
   EuiContextMenuItem,
   EuiContextMenuPanel,
+  EuiHorizontalRule,
   EuiPopover,
   EuiPopoverTitle,
+  EuiText,
   EuiToolTip,
+  EuiTourStep,
 } from '@elastic/eui';
 import { i18n } from '@osd/i18n';
+import { BehaviorSubject } from 'rxjs';
 import { useObservable } from 'react-use';
 import {
   CoreStart,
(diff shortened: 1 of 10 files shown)


src/plugins/workspace/public/components/manage_workspace_menu/manage_workspace_menu.tsx
```diff
@@ -60,6 +74,16 @@ export const ManageWorkspaceMenu = ({ coreStart }: Props) => {
 
   const [isPopoverOpen, setIsPopoverOpen] = useState(false);
 
+  const [isTourDismissed, setIsTourDismissed] = useState(() =>
+    Boolean(localStorage.getItem(MANAGE_WORKSPACE_TOUR_STORAGE_KEY))
+  );
+  const [isTourOpen, setIsTourOpen] = useState(false);
```
SuZhou-Joe: This state is more like a derived state to me, shall we use isTourOpen = !isTourDismissed?

ps48: Good call you're right, it was derived state. Replaced with `const isTourOpen = !isTourDismissed` and removed the `useState`/`useEffect` (and the now-unused `useEffect` import). Done in 97c9aef.

TackAdam (approved):

joshuali925 (approved):

Issue #8884, opened 2024-11-18
DarkMode settings apply globally per tenant and not per user


billal280:
In the lastest version of OpenSearch Dashboards (2.18.0), darkmode settings is user-specific but only applies globally within the "global" tenant. This means that if a user in tenant X activates dark mode, all other users within that same tenant are affected, and vice versa.
[#2159](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/2159)

Similarly, users must manually switch to dark mode for each tenant they have access to. However, logically, when a user decides to enable dark mode, it should apply across all the tenants they access.

This creates inconsistencies in the user experience when navigating between different tenants.

PR #12329, opened 2026-07-06
Feat: add user/workspace setting page

yubonluo:
### Description
Adds dedicated User settings and Workspace settings pages to left navigation, so a UI setting registered with `USER` / `WORKSPACE` scope can be edited per-user or per-workspace, while the existing Application page continues to edit the global scope.

If you want to see the user/workspace level setting and admin setting, you should enable these configs:
```
uiSettings:
  overrides:
    "home:useNewHomePage": true,

# enable security plugin to enable user level setting

# enable workspace level setting
workspace.enabled: true

# enable admin setting in advance setting page
```

**Key changes:**

**Scoped settings pages**

- `mountManagementSection` takes an optional `UiSettingScope`; `plugin.ts` registers `user_settings` (outside-workspace) and `workspace_settings` (inside-workspace) apps, gated by the `userSettings` / `workspaces` capabilities. The workspace nav link is added under "Manage workspace".

- On a scoped page, each field shows its provenance: a source badge ("User value" / "Workspace value" vs "Application"), a "Use Application value" action to clear the per-scope override and inherit from global, and an inline hint describing whether a pending change will override or inherit.
- `AdvancedSettings` fetches per-scope values (`getUserProvidedForScope`) instead of subscribing to the merged client cache, since the merged cache cannot express which layer a value comes from. Adds explicit loading and error states (full-page error + reload when any scope read fails, rather than showing partial data).

**Admin control of global settings**

- New admin-only setting `enableGlobalSettingControl` (`DASHBOARD_ADMIN` scope). When enabled, non-admins see global settings as read-only. This replaces the previous `DynamicConfigControlledUiSettingsWrapper` (removed); its behavior is merged into `PermissionControlledUiSettingsWrapper`, with the admin toggle as the source of truth and the legacy `uiSettings.globalScopeEditable` dynamic config as a backward-compatible fallback.
- Server reports `advancedSettings.permissionControlEnabled` and enables `userSettings` only when an authenticated user can be resolved.
- `getUserProvided` skips the `DASHBOARD_ADMIN` scope when permission control is disabled.

**Save behavior**

- `saveConfig` writes each change with an explicit scope and folds only the succeeded keys into the optimistic view; `Form.saveAll` clears the unsaved marker only for succeeded keys, names failed settings in the error toast, and shows the reload prompt for any successfully-saved key that requires it — independent of other failures.

User settings: 
Key | Description | Rec. level | Reason | Source | Priority
-- | -- | -- | -- | -- | --
theme:darkMode | Render the whole OSD UI in dark mode; a page refresh applies it. | USER, GLOBAL | Purely visual, affects only the viewer; OSD community issue #8884 explicitly asks for per-user (not per-tenant) dark mode. | core theme.ts | P1
theme:version | Which theme version the UI renders with (v7 / v9 / Next). | USER, GLOBAL | Visual variant only; no data, security, or cluster impact. | core theme.ts | P2
accessibility:disableAnimations | Turn off non-essential UI animations. | USER, GLOBAL | Accessibility/comfort choice that should differ per person. | core accessibility.ts | P1
pageNavigation | Side-navigation style (modern vs legacy). | USER, GLOBAL | Layout/appearance preference, individual. | core navigation.ts | P2
dateFormat | Moment.js pattern used to pretty-print dates. | USER, GLOBAL | Display preference; org may set a default, individuals reasonably override their own view. | core date_formats.ts | P3
dateFormat:tz | Timezone used to render dates (Browser = auto-detect). | USER, GLOBAL | A user's own timezone is inherently personal; display only. | core date_formats.ts | P1
dateFormat:dow | Which day weeks start on. | USER, GLOBAL | Locale/personal display preference. | core date_formats.ts | P3
defaultWorkspace | The workspace a user lands in by default. | USER ✅ | Personal. | workspace | P1
searchWorkspace:dismissGetStarted | Dismiss the Search overview get-started card. | USER ✅ | Personal onboarding state. | home | P2
essentialWorkspace:dismissGetStarted | Dismiss the Essentials overview get-started card. | USER ✅ | Personal onboarding state. | workspace | P2
analyticsWorkspace:dismissGetStarted | Dismiss the Analytics overview get-started card. | USER ✅ | Personal onboarding state. | workspace | P2

Workspace settings:

Key | Description | Rec. level | Reason | Source
-- | -- | -- | -- | --
defaultIndex | Index pattern used when none is explicitly selected. | WORKSPACE ✅ | Directly tied to the workspace's data. | data
defaultDataSource | Data source used when none is set. | WORKSPACE ✅ [WS,GLOBAL] | Workspace data context. | workspace
observability:defaultDashboard | Default Observability dashboard. | WORKSPACE ✅ | Data/workspace context. | dashboards-observability
observability:overviewCardsDisplay | Whether overview cards are shown. | WORKSPACE ✅ | Workspace display. | dashboards-observability

## Screenshot

https://github.com/user-attachments/assets/1a506052-ba5c-4fda-8957-26c41d3d3ca1

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


src/core/public/ui_settings/types.ts
```diff
@@ -59,6 +59,11 @@ export interface IUiSettingsClient {
    */
   getUserProvidedWithScope: <T = any>(key: string, scope: UiSettingScope) => Promise<T>;
 
+  /**
+   * Asynchronously fetches all user-provided values for a single scope.
+   */
+  getUserProvidedForScope: (scope: UiSettingScope) => Promise<Record<string, UserProvidedValues>>;
```
SuZhou-Joe: I'd prefer to name it as `getAllUserProvidedWithScope`.

yubonluo: Updated

src/core/server/ui_settings/settings/accessibility.ts
```diff
@@ -46,6 +46,7 @@ export const getAccessibilitySettings = (): Record<string, UiSettingsParams> =>
       category: ['appearance'],
       requiresPageReload: true,
       schema: schema.boolean(),
+      scope: [UiSettingScope.USER, UiSettingScope.GLOBAL],
```
SuZhou-Joe, line 49: Will there be case that UiSettingScope.USER is disabled?

src/core/server/ui_settings/ui_settings_client.ts
```diff
@@ -175,9 +187,12 @@ export class UiSettingsClient implements IUiSettingsClient {
       // Otherwise, read from all scopes and merge.
       const scopeValue = key ? this.defaults[key]?.scope : undefined;
       const keyScopes = scopeValue ? (Array.isArray(scopeValue) ? scopeValue : [scopeValue]) : null;
-      const scopesToRead = keyScopes
+      const scopesToRead = (keyScopes
         ? UiSettingScopeReadOptions.filter((opt) => keyScopes.includes(opt.scope as UiSettingScope))
-        : UiSettingScopeReadOptions;
+        : UiSettingScopeReadOptions
+      ).filter(
+        (opt) => this.permissionControlEnabled || opt.scope !== UiSettingScope.DASHBOARD_ADMIN
```
SuZhou-Joe, line 200: Could you please elaborate more on why we have to filter dashboard admin config out?

src/core/server/ui_settings/ui_settings_service.ts
```diff
@@ -160,6 +156,7 @@ export class UiSettingsService
         defaults: mapToObject(this.uiSettingsDefaults),
         overrides: this.overrides,
         log: this.log,
+        permissionControlEnabled: this.permissionControlEnabled,
```
SuZhou-Joe, line 151: We'd better expose this in saved objects plugin, not ui_settings_service.

src/core/server/ui_settings/ui_settings_client.ts
yubonluo: When `permissionControlEnabled` control is off, we no longer register any admin-scoped setting (`enableGlobalSettingControl` is only registered when `permission.enabled` is true), so there is nothing meaningful in the `DASHBOARD_ADMIN` scope to read.

src/core/server/ui_settings/settings/accessibility.ts
yubonluo: Yes, the User settings page is controlled by the `userSettings.enabled` capability.

src/core/server/ui_settings/ui_settings_service.ts
yubonluo: Yes, but the workspace plugin already reads this the same way `globalConfig.savedObjects.permission.enabled` in `workspace/server/plugin.ts`. `permission.enabled` is a public config field, so I think reading it directly from config is consistent with existing usage.

src/plugins/advanced_settings/public/management_app/advanced_settings.tsx
```diff
@@ -164,51 +238,83 @@ export class AdvancedSettingsComponent extends Component<
...
 
   mapConfig(config: IUiSettingsClient) {
     const all = config.getAll();
     const userSettingsEnabled = config.get('theme:enableUserControl');
-    const isDashboardAdmin = !!this.props.application.capabilities.dashboards?.isDashboardAdmin;
-
-    const isGlobalScopeEditable =
-      isDashboardAdmin || !!this.props.application.capabilities.globalScopeEditable?.enabled;
+    const isDashboardAdmin =
+      this.props.application.capabilities.dashboards?.isDashboardAdmin !== false;
```
SuZhou-Joe, line 248:
```suggestion
    const isDashboardAdmin =
      !!this.props.application.capabilities.dashboards?.isDashboardAdmin;
```

If the isDashboardAdmin is undefined, the resolved value will be true.

src/plugins/advanced_settings/public/management_app/advanced_settings.tsx
```diff
@@ -128,13 +207,8 @@ export class AdvancedSettingsComponent extends Component<
   }
 
   componentDidMount() {
-    this.uiSettingsSubscription = this.props.uiSettings.getUpdate$().subscribe(() => {
-      const { query } = this.state;
-      this.init(this.props.uiSettings);
-      this.setState({
-        filteredSettings: this.mapSettings(Query.execute(query, this.settings)),
-      });
-    });
```
SuZhou-Joe, line 137: Are we safe to remove the subscription?

src/core/server/ui_settings/settings/accessibility.ts
SuZhou-Joe: Get it, so the settings page works as expected even with userSettings.enabled is false?

src/core/server/ui_settings/ui_settings_service.ts
SuZhou-Joe: Then workspace is a bad example, saved objects core / plugin should be the source of truth for permission related flag.

src/core/server/ui_settings/saved_objects/permission_controlled_ui_settings_wrapper.ts
```diff
@@ -28,11 +30,79 @@ export class PermissionControlledUiSettingsWrapper {
...
+
+      if (isDashboardAdmin) {
+        return false;
+      }
+
+      let adminConfigAttributes: Record<string, unknown> | undefined;
+      try {
+        // The toggle lives in the admin-scoped config doc, which is readable by
+        // all users (ACL grants read to '*'). `get` is unwrapped here, so no recursion.
+        const adminConfig = await wrapperOptions.client.get<Record<string, unknown>>(
```
SuZhou-Joe: Nit: use ui_settings client to get admin config, this call touches the implementation detail of admin config.

src/plugins/advanced_settings/public/management_app/components/field/field.tsx
```diff
@@ -126,6 +130,12 @@ export class Field extends PureComponent<FieldProps> {
     this.props.handleChange(this.props.setting.name, unsavedChanges);
   };
 
+  inheritsFromApplication() {
```
wanglam: Nit: Better renaming to something like `canInheritsFromApplication`, I previous think this method is for checking the current value is inherits from the application.

src/plugins/advanced_settings/public/management_app/advanced_settings.tsx
yubonluo: `undefined` will happen when the workspace plugin isn't enabled (no dashboards capability). There we intentionally want admin rather than read-only, so !== false is correct

src/plugins/advanced_settings/public/management_app/advanced_settings.tsx
yubonluo: Yes. Its only job was to re-render on `uiSettings` updates, now covered explicitly: initial load via `componentDidMount` → `fetchScopedValues`, and local saves via `saveConfig` (calls `init()` + `setState()`). I verified no other component depends on this page reacting to external `set()` calls. The only behavior dropped is live-reflecting an external `set()` while the page is open, which doesn't happen in this in-page editing flow.

src/core/server/ui_settings/saved_objects/permission_controlled_ui_settings_wrapper.ts
yubonluo: Good point, I'll read the value through the ui settings client (scoped to DASHBOARD_ADMIN), so the wrapper no longer depends on how admin config is stored.

src/plugins/advanced_settings/public/management_app/components/field/field.tsx
yubonluo: Got it, will update the name to `canInheritsFromApplication `

src/core/server/ui_settings/settings/accessibility.ts
yubonluo: Yes. userSettings.enabled: false only hides the User settings page itself, it doesn't affect the global setting elsewhere.

src/core/server/ui_settings/ui_settings_service.ts
yubonluo: Get it, will use saved objects core as the source of truth

SuZhou-Joe (approved):

wanglam (approved):

PR #12331, opened 2026-07-07
fix(explore): Hide disabled query panel actions and remove duplicate …

riysaxen-amzn:
…metrics button

The QueryPanelActions component previously rendered all registered actions and used the disabled prop to grey out entries whose getIsEnabled returned false. Since the registry has no isVisible hook, this caused duplicate labels when multiple plugins register context-specific actions for different page flavors (e.g. two "Create alert rule" buttons — one for Logs/PPL and one for Metrics/PromQL — where one is greyed out on each page).

Changes:
- Filter out disabled actions entirely before rendering in QueryPanelActions, treating getIsEnabled as a visibility gate.
- Remove the hardcoded "Create alert rule" button from the Metrics page tab bar (MetricsPageTabs). The observability plugin now owns this action via the queryPanelActionsRegistry, so the built-in button was causing a duplicate on the Metrics page.

## Screenshot
Before the changes
Image 7-6-26 at 5 51 PM

Image 7-6-26 at 5 31 PM

After the changes

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/src/plugins/explore/public/application/pages/metrics/metrics_page_tabs.tsx
+++ b/src/plugins/explore/public/application/pages/metrics/metrics_page_tabs.tsx
@@ -4,13 +4,12 @@
  */
 
 import './metrics_page_tabs.scss';
-import React, { useMemo, useState } from 'react';
+import React from 'react';
 import {
   EuiTabs,
   EuiTab,
   EuiHorizontalRule,
   EuiIcon,
-  EuiButtonEmpty,
   EuiFlexGroup,
   EuiFlexItem,
 } from '@elastic/eui';
@@ -23,10 +22,6 @@ import { DatasetSelectWidget } from '../../../components/query_panel/query_panel
 import { RootState } from '../../utils/state_management/store';
 import { setMetricsPageMode } from '../../utils/state_management/slices/ui/ui_slice';
 import { MetricsPageMode, MetricsPageModeContext } from './metrics_page_mode_context';
-import { CreateMetricsRuleFlyout } from './create_metrics_rule_flyout';
-import { splitMultiQueries } from '../../utils/multi_query_utils';
-import { useOpenSearchDashboards } from '../../../../../opensearch_dashboards_react/public';
-import { ExploreServices } from '../../../types';
 
 export { useMetricsPageMode } from './metrics_page_mode_context';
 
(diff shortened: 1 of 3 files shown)


lezzago:
## Review — coupled with dashboards-observability#2773

(Reviewed together with dashboards-observability#2773, which registers the observability-owned metrics "Create alert rule" action that this PR's changes support. The full cross-PR review, including a **blocker** on the observability side — the replacement metrics flyout never persists a rule — is on that PR.)

One finding on this PR:

### `query_panel_actions.tsx:88` — adding `dependencies` to the `useMemo` deps defeats the memoization the adjacent comment documents

`useQueryPanelActionDependencies()` returns a **fresh object literal on every render** (`use_query_panel_action_dependencies.ts:43` — the return value is not memoized). So changing the split to `useMemo(..., [registry, dependencies])` means `dependencies` is a new reference each render, and the inline/overflow split now recomputes on **every** render.

That directly contradicts the comment right above it, which claims "The slice is memoized so the underlying buttons keep stable identity across renders when the registry contents haven't changed." With an always-changing dep, the buttons no longer keep stable identity.

Suggested fix: either wrap the return of `useQueryPanelActionDependencies` in a `useMemo` (keyed on its real inputs) so `dependencies` has stable identity, or filter/slice from a value with stable identity. If recomputing per render is acceptable here, at least update the comment so it doesn't claim a stability guarantee that no longer holds.

The rest of this PR (treating `getIsEnabled` as a visibility gate, removing the duplicate built-in Metrics button, and the corresponding test update) looks correct and is the right enabling change for the observability action.

riysaxen-amzn:
> ## Review — coupled with dashboards-observability#2773
> (Reviewed together with dashboards-observability#2773, which registers the observability-owned metrics "Create alert rule" action that this PR's changes support. The full cross-PR review, including a **blocker** on the observability side — the replacement metrics flyout never persists a rule — is on that PR.)
> 
> One finding on this PR:
> 
> ### `query_panel_actions.tsx:88` — adding `dependencies` to the `useMemo` deps defeats the memoization the adjacent comment documents
> `useQueryPanelActionDependencies()` returns a **fresh object literal on every render** (`use_query_panel_action_dependencies.ts:43` — the return value is not memoized). So changing the split to `useMemo(..., [registry, dependencies])` means `dependencies` is a new reference each render, and the inline/overflow split now recomputes on **every** render.
> 
> That directly contradicts the comment right above it, which claims "The slice is memoized so the underlying buttons keep stable identity across renders when the registry contents haven't changed." With an always-changing dep, the buttons no longer keep stable identity.
> 
> Suggested fix: either wrap the return of `useQueryPanelActionDependencies` in a `useMemo` (keyed on its real inputs) so `dependencies` has stable identity, or filter/slice from a value with stable identity. If recomputing per render is acceptable here, at least update the comment so it doesn't claim a stability guarantee that no longer holds.
> 
> The rest of this PR (treating `getIsEnabled` as a visibility gate, removing the duplicate built-in Metrics button, and the corresponding test update) looks correct and is the right enabling change for the observability action.

Thanks @lezzago — addressed in the latest push:

useMemo dependency regression ✅
Good catch. useQueryPanelActionDependencies() returns a fresh object each render (the hook doesn't memoize its return value), so putting dependencies in the useMemo deps made it recompute every render — defeating the memoization entirely.

Fix: Removed the useMemo wrapper and replaced it with direct computation (filter + slice). The operation is trivial (a handful of registered actions at most), so there's no performance concern. Added a comment explaining why it's intentionally not memoized.

Semantics change (getIsEnabled as visibility gate)
This is intentional. The registry has no isVisible hook, and adding one would be a larger API change. In practice, the only consumers registering actions with getIsEnabled: () => false on certain pages are the observability plugin's Logs/Metrics actions — which are explicitly designed to show on one page and hide on the other. No existing action uses getIsEnabled to mean "temporarily unavailable" (e.g. waiting for a query result) — they all gate on page context (language type). If a proper isVisible hook is added later, we can migrate to it.

lezzago:
## Re-review (updated PR)

✅ The `useMemo` finding is resolved. The split is no longer wrapped in a `useMemo` keyed on the always-fresh `dependencies` object; it's now a plain `filter` + `slice`, and the new comment correctly documents *why* it's not memoized (`dependencies` is a fresh object each render and the computation is trivial over a handful of actions). The stale "keeps stable identity across renders" claim is gone. Good fix.

The rest — treating `getIsEnabled` as a visibility gate, removing the duplicate built-in Metrics button, and the updated test asserting the button is hidden (not just disabled) — all looks correct.

(Coupled with dashboards-observability#2773, whose blocker — the replacement metrics flyout not persisting rules — has now also been fixed. These two should still land together.)

lezzago (approved):

TackAdam (approved):

PR #12333, opened 2026-07-07
[Chat] Improve tool result rendering with table support

wanglam:
### Description

Renders tool call results with a shape-detection pipeline instead of dumping raw text.

#### Rendering workflow

```
renderContent(text)
  ├── tryParseJSON succeeds?
  │   ├── string → recurse (unwrap double-encoded, max depth 3)
  │   ├── array, single item, only key is "text" → renderExtractedText(item.text)
  │   ├── object, only key is "text" → renderExtractedText(parsed.text)
  │   └── other object/array → CodeBlock json={value}
  ├── {text=[...]} wrapper matched?
  │   ├── single item, only key "text" → renderExtractedText(item.text)
  │   └── other → CodeBlock json={array}
  ├── {text={...}} wrapper matched?
  │   ├── inner only key "text" → renderExtractedText(inner.text)
  │   └── other → CodeBlock json={inner}
  └── no match → renderExtractedText(text)

renderExtractedText(text)
  ├── valid JSON object/array → CodeBlock json={parsed}
  ├── "prefix\n{json}" → EuiText(prefix) + CodeBlock text={json}
  ├── pipe-delimited table → MarkdownBlock(converted GFM table)
  ├── has markdown features → MarkdownBlock(text)
  └── fallback → CodeBlock text={text}
```

#### Cases handled

| Input shape | Rendered as |
|---|---|
| `{"key":"value"}` | Pretty-printed JSON in CodeBlock |
| `[{"text":"..."}]` (single, sole key) | Extracted text → shape pipeline |
| `[{"text":"a"},{"text":"b"}]` (multi) | JSON array in CodeBlock (no join) |
| `[{"text":"...","id":1}]` (sibling keys) | JSON in CodeBlock (no data loss) |
| `{text=[{"text":"..."}]}` Java wrapper | Same as bare `[{text:"..."}]` |
| `{text={"key":[...]}}` Java wrapper | Pretty JSON in CodeBlock |
| `"\"encoded string\""` (double-encoded) | Unwrapped then re-dispatched (max 3 levels) |
| `prefix text\n{"json":...}` | Prefix as plain text + JSON in CodeBlock |
| `index \| shard\nlogs \| 0` (pipe table) | Converted to GFM markdown table |
| `# Heading\n- list item` | Rendered as markdown |
| Plain text | CodeBlock (plain, no language) |

#### Truncation

- Lives **only** inside CodeBlock (bounds EuiCodeBlock tokenizer cost)
- Line-based first (20 lines), then character-cap fallback for single-line content
- MarkdownBlock: CSS overflow:hidden cap only, fullscreen always available
- Copy and fullscreen always operate on the full, untruncated content

### Issues Resolved

N/A

## Screenshot

- Before

https://github.com/user-attachments/assets/b6148c39-440f-4bde-b96c-19a774d464f8

- After

https://github.com/user-attachments/assets/661f5307-03ef-4314-a0e1-f4534f6a536f

## Testing the changes

58 unit tests in `tool_result_renderer.test.tsx` covering each rendering path and edge cases (single-key extraction, sibling-key preservation, line + char-cap truncation, fullscreen/copy behavior). Manually verified in a local dev environment.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


src/plugins/chat/public/components/tool_result_renderer.tsx
SuZhou-Joe, line 1: We'd better use a registry for different tool result card renderer so that in the future frontend tool / backend tool can customize their tool result behavior, like most of the Chat framework does. Can be done in a follow up PR.

SuZhou-Joe (approved):

Hailong-am (approved):

ruanyl (approved):

Issue #4445, opened 2023-06-29
[BUG] Modifying the cloned dashboard affects the original dashboard


bbarani:
**Describe the bug**
Whenever I modify the cloned dashboard its updating the original dashboard as well

**To Reproduce**
1.) Create a dashboard
2.) Add visualizations to it
3.) Clone the dashboard and provide it a new name
4.) Edit the visualization in cloned dashboard and save it
5.) The original dashboard visualizations are modified based on the changes made to cloned dashboard visualization.

**Expected behavior**
The cloned dashboard changes doesn't impact the original dashboard

**OpenSearch Version**
Please list the version of OpenSearch being used.

**Dashboards Version**
Please list the version of OpenSearch Dashboards being used.

**Plugins**

Please list all plugins currently enabled.

**Screenshots**

https://github.com/opensearch-project/OpenSearch-Dashboards/assets/70038446/2fd60866-abe7-47f0-9765-323374e08ede

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

AMoo-Miki:
Dashboards are simply containers to show a bunch of visualizations together. When you create a Dashboard, you link a visualization to it. When you clone a Dashboard, it clones that relationship; it does not clone the visualizations in them.

For your desired behavior, you would want to clone the visualizations as well.

ashwin-pc:
@bbarani This is a good issue to bring up and would like to know from product/UX how we should handle something like this.  what Bharani describes is a behavior that a user might want, but from what @AMoo-Miki describes is not what the clone button is meant to do. Having some sort of toggle when a user clones a dashboard to also clone the visualizations is one way i can see his being addressed.

cc: @dagneyb @KrooshalUX @kgcreative 

Rephrasing this to be an enhancement instead of a bug though since the current behavior as Miki mentioned is expected

KrooshalUX:
Overall, we need to improve upon communicating to the end user the relationship between visualizations and dashboards. For example - given the recent exploration on color selection of visualizations, it has become clear - that its not clear - what modifications a user can make on a dashboard to a visualization actually change the visualization itself vs just within the dashboard itself. Ex: coloring via the legend doesn't change the base visualization - and can be different between dashboards. This sounds like a good thing - having customization per dashboard vs modifying the visualization, allowing everyone to do what they need at the dashboard level - unless - the user that made those customization and expected them to appear across dashboards. 

[...]

PR #12339, opened 2026-07-08
fix(ppl-lint): Allow ppl queries to have spaces

AjimelecGonzalez:
### Description

Allows PPL queries to be able to run with spaces.

### Issues Resolved

Resolves the problem where the input is a fixed format and does not allow for spaces in the ppl query if the spaces are set then `source=opensearch_dashboards_sample_data_ecommerce, opensearch_dashboards_sample_data_flights`

Returns:
`No Results`

## Testing the changes

Added 4 unit test `~/get_query_with_source.test.ts` covering the pattern of adding spaces to ppl query.

Tested locally on the Logs page `source=opensearch_dashboards_sample_data_ecommerce, opensearch_dashboards_sample_data_flights`.

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


Swiddis (commented):
suggestion (if minor): it might be worth deduplicating?

chore: needs DCO

AjimelecGonzalez:
> suggestion (if minor): it might be worth deduplicating?
> 
> chore: needs DCO

Sorry confused on what the suggestion is, are you taking about the bot suggestions?  
And will add DCO, forgot.

Swiddis:
just deduplicating the logic for this ppl transformation since the tests are exactly equivalent, if it's easy to do so. if there's weird dependency mess that makes it nontrivial to share the logic it's fine

Swiddis (approved):

src/plugins/explore/public/application/utils/languages/ppl/get_query_string_with_source/get_query_with_source.ts
```diff
@@ -24,7 +24,15 @@ export const addPPLSourceClause = (query: Query): QueryWithQueryAsString => {
   if (query.dataset && ['INDEXES', 'INDEX_PATTERN'].includes(query.dataset.type)) {
     if (hasSource) {
       // Replace source=anything with source=`anything` (handling spaces around = and ensuring anything is not already backticked)
-      const updatedQuery = queryString.replace(/(\bsource\s*=\s*)([^`\s][^\s|]*)/gi, '$1`$2`');
+      // Supports comma-separated sources with optional spaces (e.g., "source=index1, index2")
+      // Normalizes spaces around commas to prevent leading spaces in index names
+      const updatedQuery = queryString.replace(
```
mengweieric: The guard `hasSource` only checks the prefix before the first pipe (`/^[^|]*\bsource\s*=/`), but this `.replace(...)` runs globally (`/gi`) over the whole query string. So a later field comparison named `source` gets rewritten too: `source=index1 | where source=prod` becomes `source=\`index1\` | where source=\`prod\``, which changes the user's filter. To be fair the global replace predates this PR, but this PR edits the exact line and broadens what it captures, so it's a natural place to fix it. Suggest splitting at the first pipe, normalizing only the prefix, and leaving the rest untouched, with a regression test asserting `where source=prod` after a pipe is left alone.

One more the bot didn't flag: the value pattern stops at the first space, so a quoted source with a space gets mangled, e.g. `source="my index"` turns into `source=\`"my\` index"`. Worth a test for quoted names too.

src/plugins/agent_traces/public/application/utils/languages/ppl/get_query_string_with_source/get_query_with_source.ts
```diff
@@ -3,49 +3,5 @@
...
-    queryStringWithSource = `source = ${datasetTitle} ${queryString}`;
-  }
-
-  return {
-    ...query,
-    query: queryStringWithSource,
-  };
-};
+// Re-export from explore plugin to avoid code duplication
+export { addPPLSourceClause } from '../../../../../../../explore/public';
```
mengweieric, line 7: Not blocking, since `agent_traces` declares `explore` as a required plugin so this won't break at runtime. But re-exporting `addPPLSourceClause` from `explore/public` still feels like the wrong home for a generic PPL string utility. A neutral module (the `data` plugin, or a common query util) would be a cleaner owner and avoid the cross plugin coupling.

mengweieric (commented):
The user problem makes sense. One correctness issue on the rewrite scope.

src/plugins/agent_traces/public/application/utils/languages/ppl/get_query_string_with_source/get_query_with_source.ts
AjimelecGonzalez: I can open another PR after this one to do this.

AjimelecGonzalez:
> > 🔗 [Workflow run](https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/29262748767) · commit `964ca2b77685bd2ff127d3d7baaf4254e4ecc47e`
> 
> ## ❌ 2 Jest Test Failure(s)
> ### 📄 `junit-jest-integration-Windows/TEST-Jest Integration Tests.xml`
> **❌ applies filter function specified** _(0.025s)_
> 
> > Jest Integration Tests.src\dev\build\lib\integration_tests
> 
> ```
> TypeError: Cannot read properties of undefined (reading 'coreStart')
>     at Object.<anonymous> (D:\a\OpenSearch-Dashboards\OpenSearch-Dashboards\src\plugins\workspace\server\saved_objects\integration_tests\workspace_ui_settings_wrapper.test.ts:64:51)
>     at Object.asyncJestTest (D:\a\OpenSearch-Dashboards\OpenSearch-Dashboards\node_modules\jest-jasmine2\build\jasmineAsyncInstall.js:173:37)
>     at D:\a\OpenSearch-Dashboards\OpenSearch-Dashboards\node_modules\jest-jasmine2\build\queueRunner.js:89:12
>     at new Promise (<anonymous>)
>     at mapper (D:\a\OpenSearch-Dashboards\OpenSearch-Dashboards\node_modules\jest-jasmine2\build\queueRunner.js:72:19)
>     at D:\a\OpenSearch-Dashboards\OpenSearch-Dashboards\node_modules\jest-jasmine2\build\queueRunner.js:119:41
> ```
> 
> **❌ workspace ui settings saved object client wrapper should get and update global ui settings when currently not in a workspace [Attempt #2]** _(0.001s)_
> 
> > Jest Integration Tests.src\plugins\workspace\server\saved_objects\integration_tests
> 
> ```
> TypeError: Cannot read properties of undefined (reading 'get')
>     at Object.<anonymous> (D:\a\OpenSearch-Dashboards\OpenSearch-Dashboards\src\plugins\workspace\server\saved_objects\integration_tests\workspace_ui_settings_wrapper.test.ts:106:41)
>     at Object.asyncJestTest (D:\a\OpenSearch-Dashboards\OpenSearch-Dashboards\node_modules\jest-jasmine2\build\jasmineAsyncInstall.js:173:37)
>     at D:\a\OpenSearch-Dashboards\OpenSearch-Dashboards\node_modules\jest-jasmine2\build\queueRunner.js:89:12
>     at new Promise (<anonymous>)
>     at mapper (D:\a\OpenSearch-Dashboards\OpenSearch-Dashboards\node_modules\jest-jasmine2\build\queueRunner.js:72:19)
>     at D:\a\OpenSearch-Dashboards\OpenSearch-Dashboards\node_modules\jest-jasmine2\build\queueRunner.js:119:41
> ```
> 
> _2 failure(s) across 1 suite(s). Full XML reports are in the `junit-jest-*` artifacts._


Two integration tests are failing due to timeout, can be skipped or rerun.

src/plugins/explore/public/application/utils/languages/ppl/get_query_string_with_source/get_query_with_source.ts
```diff
@@ -23,9 +23,35 @@ export const addPPLSourceClause = (query: Query): QueryWithQueryAsString => {
...
+      const prefix = pipeIndex === -1 ? queryString : queryString.slice(0, pipeIndex);
+      const suffix = pipeIndex === -1 ? '' : queryString.slice(pipeIndex);
+
+      // Supports:
+      // - Unquoted source names: source=index1
+      // - Comma-separated sources with optional spaces: source=index1, index2
+      // - Quoted source names: source="my index" or source='my index'
+      // Normalizes spaces around commas to prevent leading spaces in index names
+      const updatedPrefix = prefix.replace(
+        /(\bsource\s*=\s*)(`[^`]*`|"[^"]*"|'[^']*'|[^\s|,]+(?:\s*,\s*[^\s|,]+)*,?)/gi,
```
joshuali925: does PPL support `source='my index'`?

if not, why do we need to handle them here?

also should it be + instead of *? for each things like `"[^"]*"`

joshuali925: and why do we need `/g` flag?

AjimelecGonzalez: Double checking the syntax ppl supports only backticks will correct the regex. And the /g is not necessary removing as well.

joshuali925:
### Code review

Found 1 issue:

1. Only backtick-wrapping the source clause before the first pipe means a `source=<index-pattern>` inside a subsearch/join (which PPL supports, e.g. `source=logs | where id in [source=other | head 10]`) no longer gets backticked, so the temporary sql#4444/#4445 workaround is skipped for post-pipe sources. This is a deliberate trade-off to avoid mangling field comparisons like `| where source=prod`, but the post-pipe case is not handled or called out. Worth confirming whether backticking post-pipe index-pattern sources still matters while those SQL issues are open.

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/964ca2b77685bd2ff127d3d7baaf4254e4ecc47e/src/plugins/explore/public/application/utils/languages/ppl/get_query_string_with_source/get_query_with_source.ts#L26-L30

AjimelecGonzalez:
> ### Code review
> Found 1 issue:
> 
> 1. Only backtick-wrapping the source clause before the first pipe means a `source=<index-pattern>` inside a subsearch/join (which PPL supports, e.g. `source=logs | where id in [source=other | head 10]`) no longer gets backticked, so the temporary sql#4444/[[Feature Request] Modifying the cloned dashboard should not affect the original dashboard #4445](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4445) workaround is skipped for post-pipe sources. This is a deliberate trade-off to avoid mangling field comparisons like `| where source=prod`, but the post-pipe case is not handled or called out. Worth confirming whether backticking post-pipe index-pattern sources still matters while those SQL issues are open.
> 
> https://github.com/opensearch-project/OpenSearch-Dashboards/blob/964ca2b77685bd2ff127d3d7baaf4254e4ecc47e/src/plugins/explore/public/application/utils/languages/ppl/get_query_string_with_source/get_query_with_source.ts#L26-L30
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

Out of scope for this fix. Telling apart `source=<index>` in a subsearch from `| where source=prod` would need a Full PPL parser for this.

joshuali925 (dismissed):

mengweieric (dismissed):

src/plugins/explore/public/application/utils/languages/ppl/get_query_string_with_source/get_query_with_source.ts
```diff
@@ -23,9 +23,27 @@ export const addPPLSourceClause = (query: Query): QueryWithQueryAsString => {
...
+
+      // Supports:
+      // - Backtick-quoted source names: source=`index1`
+      // - Unquoted source names: source=index1
+      // - Comma-separated sources with optional spaces: source=index1, index2
+      // Normalizes spaces around commas to prevent leading spaces in index names
+      const updatedPrefix = prefix.replace(
+        /(\bsource\s*=\s*)(`[^`]+`|[^\s|,]+(?:\s*,\s*[^\s|,]+)*,?)/i,
+        (_match, srcPrefix, sourceValue) => {
+          if (sourceValue.startsWith('`')) return _match;
```
joshuali925: one more thing i just saw, what if there's an index with backtick but not the first index?
```
source=index1,`index2`
```
This is valid syntax, but seems it would become invalid after this processing?

joshuali925 (approved):

mengweieric (approved):

PR #12345, opened 2026-07-09
feat(dashboard variables): Support custom variable label

ruanyl:
### Description
Adds display labels for dashboard variable options and extends query variables so users can choose which query result field supplies option values and which field supplies optional labels.

## Screenshot

https://github.com/user-attachments/assets/feb28427-0fa7-4347-aab8-0c83cf238959

https://github.com/user-attachments/assets/05492e45-a710-44d5-8d88-a855038db449

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


yubonluo:
Currently, user need to manually select the value field after running query, could user can select other field as value field? If not, can we only use the the field written in query as the value field?

src/plugins/dashboard/public/variables/variable_query_utils.ts
```diff
@@ -85,72 +134,105 @@ function extractValues(value: unknown): string[] {
...
+
+  if (!selectedValueField || !result.fields.includes(selectedValueField)) {
+    return { options: [], optionType: undefined };
+  }
+
+  const selectedLabelField =
+    labelField && result.fields.includes(labelField) ? labelField : undefined;
+  const optionMap = new Map<string, NormalizedVariableOption>();
+
+  result.rows.forEach((row) => {
```
yubonluo, line 193: Should we validate against empty/duplicate labels

yubonluo (approved):

Qxisylolo (approved):

SuZhou-Joe (approved):

PR #12340, opened 2026-07-08
chore(deps): eslint 10.6.0, prettier 3.9.4

tomkdgun:
### Description

- Upgrade eslint from 8.57.1 to 10.6.0 + plugins
- Upgrade prettier from 2.1.1 to 3.9.4

New prettier 3.x has different default formatting rules so some minor reformats were applied on --fix.
New eslint discover more issues mostly with `no-unused-vars` which require minor manual corrections but on many files.

This is one of the steps to resolve #12316 

Dependency Hierarchy:

eslint-8.57.1.tgz (Root Library)
 config-array-0.13.0.tgz
  minimatch-3.1.5.tgz
  ❌ brace-expansion-1.1.13.tgz (Vulnerable Library)

Because OpenSearch Dashboard is using old jest-runtime 27.x which doesn't support ESM modules with import.meta, the three tests are currently excluded, check src/dev/jest/config.js L196-L200

### Issues Resolved

closes #12338 

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`
  - [x] `yarn test:jest_integration`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [x] Commits are signed per the DCO using --signoff

PR #12362, opened 2026-07-10
fix: plugin lint adoption for ESLint 10 (concurrency warning + shared plugin registration)

SuZhou-Joe:
### Description

Two follow-ups to the ESLint 10 / Prettier 3 upgrade (#12340), both aimed at making downstream OpenSearch Dashboards plugins able to adopt the flat-config setup cleanly.

**1. Don't terminate lint on the ESLint 10 concurrency warning**

ESLint 10 emits an advisory `ESLintPoorConcurrencyWarning` when `concurrency: 'auto'` (set in `src/dev/run_eslint.js`) selects multithreading for a small workload — e.g. linting a single plugin directory. The process-warning handler in `src/setup_node_env/exit_on_warning.js` treats any unlisted Node process-warning as fatal and calls `process.exit(1)`, so `yarn lint` exited `1` **even with zero lint errors**, spuriously failing CI. `ESLintPoorConcurrencyWarning` is added to the existing ignore list alongside the other advisory warnings.

**2. Share plugin registration via `@elastic/eslint-config-kibana`**

In flat config, `plugins` registration is scoped per matched file and is **not** inherited between config objects. A plugin that spread the shared config and then referenced, say, `@typescript-eslint/no-explicit-any` in its own override block would fail with `Could not find plugin "@typescript-eslint"` unless it re-`require`d and re-registered the plugin. That forced every plugin to know which plugin owns each rule.

This registers every bundled plugin once in an unscoped config object (`globalPluginRegistration`), placed first in the exported array, so downstream rule references resolve at any file scope with no per-plugin boilerplate:

- `plugins.js` — single source of truth for plugin instances + global registration.
- `eui.js` — opt-in EUI config block (EUI is intentionally not part of the base config; OSD core does not lint against EUI rules).
- `extras.js` — aggregates `{ eui, plugins, restrictedGlobals }` behind a single `/extras` subpath.
- `eslint.config.js` — sources instances from `plugins.js`; global registration first. The default export stays a pure spreadable array.
- `package.json` — adds an `exports` map (`.`, `./eslint.config.js`, `./extras`); aligns peer deps.
- plugin generator template + README — use the plain spread pattern, no registration.

A plugin's `eslint.config.js` now looks like:

```js
const osdConfig = require('@elastic/eslint-config-kibana');
const { eui } = require('@elastic/eslint-config-kibana/extras');

module.exports = [
  ...osdConfig,
  ...eui,
  { files: ['**/*.{js,mjs,ts,tsx}'], rules: { /* references any plugin rule, no registration */ } },
];
```

### Issues Resolved

Follow-up to #12340.

## Screenshot

N/A — no UI change.

## Testing the changes

1. Clean lint no longer terminates:
   - From a plugin directory that lints cleanly, run `node ../../scripts/eslint .`
   - Before: prints `Node.js process-warning detected ... Terminating process...` and exits `1` despite `0 errors`.
   - After: no termination message; exits `0`.
2. Real lint errors still exit `1` (verified with a file containing a missing semicolon).
3. Shared config verified against a real plugin (dashboards-investigation) using only `...osdConfig` + `...eui` + local rule overrides, with no plugin `require`/registration:
   - `@osd/eslint/require-license-header`, `@typescript-eslint/no-unused-vars` (incl. `caughtErrorsIgnorePattern`), `@typescript-eslint/no-explicit-any`, `@osd/eslint/no-restricted-paths`, `jest/expect-expect`, and `@elastic/eui/*` recommended rules all fire as expected.
   - `ignores` (migrated from `.eslintignore`) correctly skips `**/*.d.ts`.
4. OSD root config still loads and self-lints clean; `examples/expressions_example` (which imports `@elastic/eslint-config-kibana/eslint.config.js`) still resolves.

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/packages/opensearch-eslint-config-opensearch-dashboards/plugins.js
+++ b/packages/opensearch-eslint-config-opensearch-dashboards/plugins.js
@@ -1,0 +1,72 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+/* eslint-disable import/no-unresolved */
+
+// Single source of truth for the ESLint plugin instances used across the
+// OpenSearch Dashboards flat config. Both the shared config (`eslint.config.js`)
+// and the plugin helper (`create_config.js`) consume these so that plugin
+// registration lives in exactly one place.
+//
+// In ESLint flat config, `plugins` registration is scoped per matched file, not
+// inherited between config objects. Downstream plugins that only want to tweak a
+// few rules should not have to know which plugin owns a rule or re-register it.
+// `globalPluginRegistration` below registers every plugin in a single unscoped
+// config object, making every rule referenceable from any later config object
+// regardless of its `files` scope.
+
+const tsPlugin = require('@typescript-eslint/eslint-plugin');
+const eslintCommentsPlugin = require('@eslint-community/eslint-plugin-eslint-comments');
+const importPlugin = require('eslint-plugin-import-x');
+const jestPlugin = require('eslint-plugin-jest');
+const jsxA11yPlugin = require('eslint-plugin-jsx-a11y');
+const _mochaRaw = require('eslint-plugin-mocha');
+const mochaPlugin = _mochaRaw.default || _mochaRaw;
+const noUnsanitizedPlugin = require('eslint-plugin-no-unsanitized');
+const reactPlugin = require('eslint-plugin-react');
+const reactHooksPlugin = require('eslint-plugin-react-hooks');
+const prettierPlugin = require('eslint-plugin-prettier');
+const osdEslintPlugin = require('@osd/eslint-plugin-eslint');
+
+const plugins = {
+  '@typescript-eslint': tsPlugin,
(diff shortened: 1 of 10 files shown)


SuZhou-Joe:
> @elastic/eslint-plugin-eui@^2.14.0, @osd/eslint-plugin-eslint@1.0.0, eslint-config-prettier@^10.1.8, eslint-plugin-prettier@^5.5.6

These dependencies are already installed inside OSD root.

wanglam (approved):

packages/opensearch-eslint-config-opensearch-dashboards/plugins.js
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * Any modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+/* eslint-disable import/no-unresolved */
+
+// Single source of truth for the ESLint plugin instances used across the
+// OpenSearch Dashboards flat config. Both the shared config (`eslint.config.js`)
+// and the plugin helper (`create_config.js`) consume these so that plugin
```
FriedhelmWS, line 16: Nit: do we really have this `create_config.js`?

FriedhelmWS (approved):

packages/opensearch-eslint-config-opensearch-dashboards/plugins.js
SuZhou-Joe: I think so since it is used for plugins creation helper.

PR #12372, opened 2026-07-13
Fix error toast showing 'Error' without error message in production

Hailong-am:
### Description

When HTTP errors are thrown and displayed as toast notifications, the error message was being lost. The toast would show the generic title (e.g., \"Error\") with no descriptive text because `\`${error}\`` on a structured HTTP error object produces an unhelpful empty or generic string rather than the actual error message.

This fix extracts the human-readable message from `error.body.message` or `error.message` before falling back to string coercion, ensuring the actual error reason is shown to the user.

Affected locations:
- `table_list_view.tsx` — delete item error toast
- `saved_objects_table.tsx` — fetch, find, delete, and relationship-fetch error toasts

## Screenshot

**Before**

image

**After**

image

## Testing the changes

1. Start OSD against an OpenSearch cluster
2. Trigger an API error in Saved Objects Management (e.g., attempt an operation that fails server-side)
3. Observe the toast now displays the actual error message instead of showing only the toast title with no body text
4. Verify that plain string errors still display correctly (fallback path)

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`

### DIFF
--- a/src/plugins/opensearch_dashboards_react/public/table_list_view/table_list_view.tsx
+++ b/src/plugins/opensearch_dashboards_react/public/table_list_view/table_list_view.tsx
@@ -204,7 +204,7 @@ class TableListView extends React.Component<TableListViewProps, TableListViewSta
             values={{ entityName: this.props.entityName }}
           />
         ),
-        text: `${error}`,
+        text: error?.body?.message || error?.message || `${error}`,
       });
     }
     this.fetchItems();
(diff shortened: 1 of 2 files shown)


SuZhou-Joe (approved):
Nit: possible to add some UT?

wanglam (approved):

Hailong-am:
> Nit: possible to add some UT?

Thanks, will add UT in follow up PRs.

PR #12379, opened 2026-07-13
[Explore] Gate SQL in dataset creation under feature flag

ritvibhatt:
### Description

Fix SQL language appearing in Explore's dataset selector and query editor even when the explore.sqlSupport.enabled feature flag is disabled. Previously, SQL was always registered as a supported language for Explore regardless of the flag. Now, when the flag is disabled (default), SQL is removed from Explore's supported languages by re-registering an updated config via registerLanguage.

## Screenshot

Before (feature flag off):
Screenshot 2026-07-14 at 11 24 54 AM

After (feature flag off):
Screenshot 2026-07-14 at 11 21 01 AM

Default language behavior

https://github.com/user-attachments/assets/68b2cc98-5735-4eef-b19e-c942297241a3

## Testing the changes

1. Go to explore logs with sql support feature flag not enabled
2. Try to create dataset
3. Should not see OpenSearch SQL as a possible language to select

### Check List

- [ ] All tests pass
  - [ ] `yarn test:jest`


ps48:
@ritvibhatt when the SQL flag is enabled do the users see PPL as the default option?

src/plugins/explore/public/plugin.ts
```diff
@@ -700,6 +700,26 @@ export class ExplorePlugin implements Plugin<
...
+      const sqlConfig = languageService?.getLanguage('SQL');
+      if (sqlConfig) {
+        const updated = {
+          ...sqlConfig,
+          supportedAppNames: (sqlConfig.supportedAppNames ?? []).filter(
+            (name) => name !== 'explore'
+          ),
+          editorSupportedAppNames: (sqlConfig.editorSupportedAppNames ?? []).filter(
+            (name) => name !== 'explore'
+          ),
```
joshuali925: is it possible to not add the explore app if sql support is disabled, instead of adding and then removing?

ritvibhatt: inverted to only add explore if the sql support feature flag is enabled

ritvibhatt:
> @ritvibhatt when the SQL flag is enabled do the users see PPL as the default option?

@ps48 When creating a dataset in explore, if SQL is the current selected language in explore the dataset creation will default to SQL and if PPL is selected it'll default to PPL.
From the dataset creation page it now defaults to PPL

ps48 (approved):

joshuali925 (approved):

PR #12393, opened 2026-07-15
feat: add support for range field types in Discover

### LINKED ISSUE #11053: [BUG] Support all range field types
**Describe the bug**

Opensearch Dashboard currently does no support range field types.
There was an issue for it at Kibana https://github.com/elastic/kibana/issues/12751 which got fixed with version 7.13.
As opensearch was forked from 7.12 it was never supported.

**To Reproduce**
Steps to reproduce the behaviour:
Create an index with a [range type](https://docs.opensearch.org/latest/mappings/supported-field-types/range/) and try to create an index pattern for it. The range field will be displayed as `unkown`.

**Expected behaviour**
All range field types should be supported, so they can be used in the Discover function.

**OpenSearch Version**
All

**Dashboards Version**
All

**Plugins**

**Additional context**

This is a more general issue to combine two open issues.
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5445
https://github.com/opensearch-project/OpenSearch-Dashboards/issues/2665

mmustafasenoglu:
## Description

Add support for OpenSearch range field types (`integer_range`, `float_range`, `long_range`, `double_range`, `date_range`, `ip_range`) in Discover. These types were missing since OpenSearch was forked from Elasticsearch 7.12, causing range fields to display as `unknown`.

## Approach

Range fields are interval/object values (`{gte, lte}`), not scalar values. Mapping them to scalar OSD types (`number`, `date`, `ip`) would enable unintended behaviors:

- scalar formatters applied to `{gte, lte}` objects
- fields becoming `filterable`/`sortable` through scalar type flags
- filter editor exposing scalar operators (`is`, `is not`, `is one of`)
- Discover "filter for value" creating phrase filters for range objects

Instead, range fields are modeled with **distinct OSD field types**:

| OpenSearch field type | OSD field type |
|---|---|
| `integer_range` | `NUMBER_RANGE` |
| `float_range` | `NUMBER_RANGE` |
| `long_range` | `NUMBER_RANGE` |
| `double_range` | `NUMBER_RANGE` |
| `date_range` | `DATE_RANGE` |
| `ip_range` | `IP_RANGE` |

These types default to `sortable: false, filterable: false`, preventing invalid sort/filter/aggregation operations in Discover.

## Changes

- **types.ts**: Added `NUMBER_RANGE`, `DATE_RANGE`, `IP_RANGE` to `OSD_FIELD_TYPES` enum
- **osd_field_types_factory.ts**: Added 3 new `OsdFieldType` entries with range-specific esTypes mappings
- **osd_field_types.test.ts**: Updated cast and type name tests to expect new range types

## Related issue

Closes #11053


ruanyl:
Hi @mmustafasenoglu, thanks for submitting the PR, would you mind to fix the failed DCO CI?

ruanyl:
Thanks for the PR. I agree range fields should not show up as `unknown`, but I do not think we should map OpenSearch range field types directly to scalar OSD field types (`number`, `date`, `ip`).

Range fields are interval/object values, not scalar values. Mapping them to scalar OSD types enables scalar behavior:

- scalar field formatters are applied to `{ gte, lte }` objects
- fields become `filterable` / `sortable` through the scalar type flags
- the filter editor exposes scalar operators like `is`, `is not`, `is one of`
- Discover “filter for value” can create phrase filters for range objects
- visualization/aggregation field selection may offer unsupported scalar aggs

The filter UI is a concrete example: `ip_range` is treated like `ip`, so the editor shows scalar operators. Only range-style filtering is appropriate for these fields.

I think the safer approach is to model range fields distinctly, what do you think?

mmustafasenoglu:
Hi @ruanyl, thanks for the detailed feedback — you raise valid points about scalar mapping enabling unintended behaviors (filter operators, sortable flags, etc.).

You're right that range fields are inherently interval/object values. I'd be happy to revise the approach to model them distinctly.

A couple of questions before I update:

1. Should range fields get their own `OSD_RANGE_FIELD_TYPE` entries (e.g., `integer_range`, `float_range`, `date_range`, `ip_range`, `long_range`, `double_range`) with dedicated rendering/filtering logic, or
2. Should they remain as a generic `range` type with a sub-type property?

Would appreciate your guidance on the preferred direction before I proceed with the rewrite.

ruanyl:
@mmustafasenoglu Thanks for checking. I think we can follow the existing OSD field type categorization instead of adding one OSD type per OpenSearch range mapping.

| OpenSearch field type | Proposed OSD field type |
|---|---|
| `integer_range` | `OSD_FIELD_TYPES.NUMBER_RANGE` |
| `float_range` | `OSD_FIELD_TYPES.NUMBER_RANGE` |
| `long_range` | `OSD_FIELD_TYPES.NUMBER_RANGE` |
| `double_range` | `OSD_FIELD_TYPES.NUMBER_RANGE` |
| `date_range` | `OSD_FIELD_TYPES.DATE_RANGE` |
| `ip_range` | `OSD_FIELD_TYPES.IP_RANGE` |

This keeps the model close to the existing `number` / `date` / `ip` categories, but avoids inheriting scalar behavior by accident. Does it make sense?

mmustafasenoglu:
Hi @ruanyl, updated the approach based on your feedback.

**Changes made:**
- Added 3 distinct OSD field types: `NUMBER_RANGE`, `DATE_RANGE`, `IP_RANGE`
- Removed range types from scalar `NUMBER`, `DATE`, `IP` mappings
- Range types are `sortable: false, filterable: false` by default
- Updated tests to reflect new type casting

This prevents scalar behavior (sort, filter operators, formatters) from leaking into range fields while keeping the categorization close to the existing `number`/`date`/`ip` pattern.

Also fixed the DCO sign-off on the commits. Could you take another look when you get a chance? Thanks!

ruanyl:
Thanks for updating the PR @mmustafasenoglu 

I see there are some test failure related to test snapshots, you can run the test locally and fix the snapshots:
```
yarn test:jest <path> --watch
```

And the DCO check also failed, would you mind to take a look and get the commits properly signed? thanks!

mmustafasenoglu:
Closing in favor of new PR #12394 with squashed commits and DCO sign-off. Snapshot test fix included.

Hanyu-W:
Hey @mmustafasenoglu, I think there is a typo in PR number: #12394 is my unrelated PR. Did you mean #12430?

PR #12394, opened 2026-07-15
feat(ppl-lint): add three index-pattern metadata lint rules + backend context plumbing

Hanyu-W:
### Description

Builds on #12298 (field-validation + command-typo). This PR adds **three index-pattern metadata lint rules** —  which read field *types* off the loaded data view, not just field *names*. Each runs entirely from the parse tree plus the already-loaded index-pattern metadata, so no new request is made:

| Rule | What it flags 
|---|---|
| `agg-on-text` | An aggregation over a `text` field (needs a `keyword`/numeric) |
| `flat-object-subfield` | A subfield reference under a `flat_object` (not independently queryable) |
| `type-mismatch-numeric` | A numeric field compared against a non-numeric literal |

`agg-on-text` and `type-mismatch-numeric` carry a `warning` severity floor (user-downgradable, but not below warning). A field whose type is ambiguous across backing indices is omitted from the type map, so a rule self-suppresses rather than acting on an arbitrary type.

### Supporting change: shared lint context

Added more shared lint context — field types, engine type, and a trustworthy source classification — so this PR also threads that through the lint run, the web-worker boundary, and both editor hosts (Discover + Explore). It's plumbing, kept deliberately behavior-neutral so it can also serve the later rule PRs (source-metadata, explain-based performance, telemetry, AI-fix) instead of each re-implementing it. Nothing here fires a rule, request, or event on its own:

- `field_path`: quote-aware `parseFieldPath`/`normalizeFieldPath`/`findLongestTypedPrefix`, fixing the backtick-quoted-dot mis-split in the old `normalizeFieldName`.
- `extractFieldMetadata`: builds a field name → unambiguous `esType` map from the loaded data view.
- Provenance-checked context: `typeMap`, `selectedSourcePattern`, and `engineType` are exposed only when the field cache matches the active dataset id **+** data source id **+** dataset type.
- `top_level_source`: parser-neutral source classification + `isPipeFirst`, working on both the runtime and compiled-simplified grammar surfaces.
- `events`: optional no-op `PPLLintEventSink` (telemetry registers a real sink later).
- `diagnostic_action`: contributor registry consumed by the code-action and hover providers, plus `CatalogEntry.aiFixable`
- `isPipeFirst`, `selectedSourcePattern`, and `engineType` carried through the worker serialization.

### Correctness fixes carried in

- Prefer cached `calciteEnabled` as the authoritative engine gate, with the semver heuristic only as the no-settings fallback.
- Resolve the local-cluster grammar version from the engine, not the OSD package version.
- Key fix-registry entries by rule id so two same-span/same-message diagnostics with different fixes don't collide.
- Reject a prototype-chain name (e.g. `toString`) as a severity via an own-property guard.
- Offer the field-slot quick fix only for an AST-confirmed bare-field RHS (not a computed expression).

### Issues Resolved

Supports [opensearch-project/sql#5405](https://github.com/opensearch-project/sql/issues/5405).

## Screenshot
image
Agg-on-text
image
type-mismatch-numeric
image
flat-object-subfield

## Testing the changes

```bash
yarn test:jest packages/osd-monaco/src/ppl/
yarn test:jest src/plugins/data/public/ppl_lint/
yarn test:jest src/plugins/explore/public/components/query_panel/query_panel_editor/use_query_panel_editor/
cd packages/osd-monaco && npx tsc --noEmit
```

### Check List

- [x] All tests pass
  - [x] `yarn test:jest`


Hanyu-W:
I investigated the failing check:

1. osd:ciGroup2 appears to be a transient Cypress CI flake. It failed in the external table-visualization, which this PR does not touch

All other checks, including the merge gate, are passing.

packages/osd-monaco/src/ppl/lint/rules/agg_on_text.test.ts
```diff
@@ -0,0 +1,162 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { CharStream, CommonTokenStream, ParserRuleContext } from 'antlr4ng';
```
joshuali925, line 6: not blocking but i think most of OSD puts test files next to the source file, do we need separate `__tests__` dir?

packages/osd-monaco/src/ppl/lint/rules/agg_on_text.ts
```diff
@@ -0,0 +1,99 @@
...
+import { Diagnostic } from '../diagnostic';
+import { Detector } from '../types';
+import { findAllDescendantsByRule, findChildByRule } from '../rule_index';
+import { rangeFromContext } from '../range_utils';
+import { parseFieldPath } from '../field_path';
+
+// Engine ground truth (verified live, OpenSearch 3.7): a numeric aggregation
+// (avg/sum/stddev/var/median) on a `text`/`keyword` field returns null with a
+// `double` schema type and no error — a silent failure. This is a warning and
+// self-suppresses without a typeMap.
```
joshuali925: do we need comments like this? and in the future when engine updates, how do we sync the list?

packages/osd-monaco/src/ppl/lint/code_action_provider.ts
joshuali925: should we remove many of the comments? they are becoming stale and it's extra effort to maintain. here it says position + message, but in language.ts it is range + message + rule id. should they be different or comment is outdated?

packages/osd-monaco/src/ppl/lint/lint_runner.ts
```diff
@@ -36,6 +36,7 @@ export function mergeConfig(local: CatalogEntry, override?: Partial<CatalogEntry
 function isContextEmpty(context: LintRunContext | undefined): boolean {
   return (
     (!context?.fields || context.fields.size === 0) &&
+    (!context?.typeMap || context.typeMap.size === 0) &&
```
mengweieric, line 40: The metadata rules are gated only by whether the context is non empty (isContextEmpty here), never by whether the query's source matches the dataset the types came from. Since execution keeps an explicit source (addPPLSourceClause only adds one when absent), selecting orders and typing `source=returns | stats avg(note)` lints against orders rather than returns, which yields false findings and unsafe fixes. classifyTopLevelSource already exists, so it would be good to gate these rules on a source match. Could you confirm this against what you see in Explore?

packages/osd-monaco/src/ppl/lint/rules_catalog.json
```diff
@@ -95,5 +95,35 @@
...
+    "detector": "agg-on-text",
+    "enabled": true,
+    "severity": "warning",
+    "message": "Numeric aggregation on a text/keyword field returns null rather than a numeric result.",
+    "docUrl": "https://docs.opensearch.org/latest/sql-and-ppl/ppl/commands/stats/",
+    "needsContext": true,
+    "appliesTo": {}
+  },
+  {
+    "id": "flat-object-subfield",
```
mengweieric, line 115: These three rules are default on with an empty appliesTo, so they appear to fire on every engine and version, while the sibling rules above carry engine and minVersion gates and the rule comments verify only recent Calcite OpenSearch. flat-object-subfield is an error. The context already carries engineType and isCalcite, so could we gate them to the surface actually validated?

mengweieric (commented):
Two things I think are worth resolving before merge, both from a close read of the code rather than a run, so I would appreciate you verifying the behavioral points (especially the source mismatch). Inline below.

packages/osd-monaco/src/ppl/lint/rules/agg_on_text.ts
```diff
@@ -0,0 +1,78 @@
...
+const TEXT_TYPES: ReadonlySet<string> = new Set(['text', 'keyword']);
+
+export const aggOnTextDetector: Detector = (tree, config, context, ruleNameToIndex) => {
+  const typeMap = context.typeMap;
+  if (!typeMap || typeMap.size === 0) {
+    return [];
+  }
+
+  const diagnostics: Diagnostic[] = [];
+  const statsFunctions = findAllDescendantsByRule(tree, ruleNameToIndex, 'statsFunction');
```
mengweieric, line 40: This resolves node names (statsFunction, functionArgs, fieldExpression) against whichever grammar is active, and since these rules are not runtimeOnly they run on both surfaces: the backend-fetched runtime ATN when the cache is warm, and the compiled-simplified bundled grammar otherwise (lint_bridge falls back to the analyzer when getCachedGrammar returns null, e.g. during the initial fetch window or the 30s cooldown after a failed grammar fetch). So which grammar produces the tree here is a timing and network artifact that is invisible to the user. Could you verify these three rules give identical results on both surfaces, i.e. that these rule names and the tree shape match between the runtime bundle and the compiled grammar? If they can diverge, the same query would lint differently depending only on whether the grammar fetch had completed.

mengweieric (commented):
One follow-up on the grammar-surface dependency, as a verify-please rather than a blocker.

packages/osd-monaco/src/ppl/lint/rules/agg_on_text.test.ts
Hanyu-W: Done in 0abcc99. All eight test files introduced by this PR now live next to their source files. Pre-existing __tests__ files were left unchanged, will move those on a follow-up pr

packages/osd-monaco/src/ppl/lint/rules/agg_on_text.ts
Hanyu-W: Addressed in 0abcc99 and 77a57bc. I removed the pasted grammar and self-evident commentary, retaining only the rationale for the observed engine behavior and exclusions. For future engine updates, I am developing an engine-backed lint-rule validation workflow (https://github.com/Hanyu-W/sql/pull/2) that will detect behavioral changes and identify when NUMERIC_ONLY_AGGS, its tests, or version gates need updating.

packages/osd-monaco/src/ppl/lint/code_action_provider.ts
Hanyu-W: I agree, did an aggressive comment prune. This specific comment was outdated, the comment now says range + message + rule id, matching markerFixKey and language.ts.

packages/osd-monaco/src/ppl/lint/rules/agg_on_text.ts
Hanyu-W: Added a new test in 8c6b112, confirmed there is no divergence those rules give identical results on both surfaces.

packages/osd-monaco/src/ppl/lint/lint_runner.ts
Hanyu-W: Confirmed. Fixed in b5869b4: added sourceConflictsWithDataset plus a sourceScoped flag, set on all four metadata rules — field-validation had the same exposure. Suppresses only on a proven mismatch, fails open otherwise.

packages/osd-monaco/src/ppl/lint/rules_catalog.json
Hanyu-W: Fixed in b5869b4 + 28bcad6. Verified on 3.8 that all three are Calcite-specific — with Calcite off, v2 raises a hard error rather than the silent failure the messages describe. agg-on-text and type-mismatch-numeric gated to Calcite ≥3.7, flat-object-subfield to Calcite ≥3.8.

Hanyu-W:
Thank you @joshuali925 @mengweieric for the reviews! Highlights of the changes:

 - Source/dataset mismatch (b5869b4) — sourceConflictsWithDataset + sourceScoped flag on all four metadata rules (field-validation had it too). Suppresses only on a proven mismatch, fails open otherwise. Verified in Explore.
 - Engine/version gates (b5869b4, 28bcad6) — verified on 3.8 that all three are Calcite-specific. agg-on-text/type-mismatch-numeric → Calcite ≥3.7, flat-object-subfield → ≥3.8; catalog.test.ts asserts each appliesTo.
 - Grammar surfaces (8c6b112) — 13 queries through a real 3.8 engine ATN vs the simplified parser: identical diagnostics and ranges.
 - Test co-location (0abcc99) — this PR's eight files moved; pre-existing __tests__/ in a follow-up.
 - Comments (0abcc99, 77a57bc) — ~74 lines cut across 20 files; fixed the stale code_action_provider.ts one.
 - NUMERIC_ONLY_AGGS sync — engine-backed validation workflow in progress so engine changes fail a check.
 - CI — None of the red checks come from this PR. All of them were green on 349b785; the only change since is the main merge, which touched and no PPL Linter code at all.
Please re-review when y'all get a chance.

joshuali925 (dismissed):

Hanyu-W:
Pushed 091cc034c7 — version/engine detection fixes:

 - Unknown-version branch ignored isCalcite and never checked minVersion. Now one conservative branch, keyed on catalog (not user-overridden) severity.
 - isCalcite: true wasn't always measured — the route fails open to true and that was cached as authoritative. Route now reports calciteMeasured; only measured readings count. Added the missing detector self-checks.
 - Pipeline stages matched a stale name list, so rareTopCommand + 16 engine commands were invisible — which also made field-validation flag valid queries as unknown fields at error severity. Replaced with an explicit order classification for both grammars + a census test. Fixes timechart too.
 - Lint-rules uiSetting read no longer throws when queryEnhancements is off.

joshuali925 (dismissed):

Hanyu-W:
All test CI failures are now resolved.

The CI Group 4 failure was caused by a semantic conflict introduced when main was merged. #11753 changed useQueryPanelEditor() to require QueryEditorProps, but the PR-only test helper still called it without arguments. I updated the tests to use buildProps() and override queryState.dataset with mdsDataset. Both Linux and Windows Group 4 jobs now pass. 

PR is merge-ready.

mengweieric (approved):

joshuali925 (approved):

